PR #420, opened 2015-04-22
Use datalib

domoritz:


src/Encoding.js
```diff
@@ -194,14 +194,6 @@ var Encoding = module.exports = (function() {
     return sort;
   };
 
-  proto.any = function(f) {
```
kanitw: Have you check that these `Encoding.all/any` aren't being used in any dependent projects?

domoritz: A little bit. If I notice they are needed, I'll add them to datalib.

domoritz: Once you merged this, I'll port polestar and voyager.

src/util.js
```diff
@@ -1,175 +1,13 @@
...
-
-util.isArray = Array.isArray || function(obj) {
-  return toString.call(obj) == '[object Array]';
-};
-
-util.array = function(x) {
-  return x ? (util.isArray(x) ? x : [x]) : [];
-};
-
 util.forEach = function(obj, f, thisArg) {
```
kanitw: shouldn't this be included in datalib too?

src/vl.js
```diff
@@ -4,7 +4,10 @@ var globals = require('./globals'),
     util = require('./util'),
     consts = require('./consts');
 
-var vl = util.merge(consts, util);
+var vl = {};  // util.merge(consts, util);
```
kanitw: This will break dependent projects ...

src/util.js
domoritz: Not really. I think this is a bad implementation and would rather delete it.

src/Encoding.js
kanitw: compass too.

domoritz: I can try.

src/vl.js
domoritz: Will fix those. The problem is that if everything is in the same namespace it's really hard to find anything.

src/util.js
kanitw: Why is this bad?   We don't have a way to call foreach for object and this is it.   
Lodash and underscore provide similar method.

src/vl.js
kanitw: I think splitting might be better, but then you should submit a concurrent pr in the other 3 projects for these then.

test/data.spec.js
```diff
@@ -13,11 +14,7 @@ describe('getSchema', function() {
       {a: '', b: null, d: undefined},
       {a: 1, b: 's', d: 'Jan 11, 2014'}
     ];
-    var s = data.getSchema(d);
```
kanitw: are these two exactly the same? .... not sure.

src/vl.js
domoritz: If you reject this, I do all the work for nothing.

src/util.js
domoritz: The array version should use `for(var i; i<n;i++)` for performance reasons. Anyway, I'm okay with keeping this here for now but let's not push it into datalib.

test/data.spec.js
domoritz: Should be. With the sorting below.

src/vl.js
kanitw: I'm not rejecting, I am just saying please fix dependencies downstream.

domoritz: Ok

kanitw:
Please merge master into this branch too.  (I added `vl.version` and a few more things earlier today.)

domoritz:
> Please merge master into this branch too. (I added vl.version and a few more things earlier today.)

Done

src/util.js
kanitw: Array has `foreach`, so it would call the native’s foreach anyway.

kanitw: In any case, we would never call this method, if the collection is an array.

src/vl.js
kanitw: Thinking more about it.  This is a bad idea. I’ll revert this part — see 536b3d5e26b7070a33d1a6906a71418eb888e2a2

src/util.js
```diff
@@ -180,62 +18,6 @@ util.forEach = function(obj, f, thisArg) {
...
-    return obj.map.call(thisArg, f);
-  } else {
-    var output = [];
-    for (var k in obj) {
-      output.push( f.call(thisArg, obj[k], k, obj));
-    }
-  }
-};
-
-util.any = function(arr, f) {
```
kanitw: @domoritz — FYI, you broke compass by removing this ...

domoritz: Hmm, I guess, we can either move it into datalib or compass, since it was unused in vl.

PR #423, opened 2015-05-08
remove getSchema & update test

kanitw:
I’m moving getSchema to data-ui’s `dataset.js` instead.


domoritz:
And use datalib. Makes sense.

PR #426, opened 2015-05-09
update editor to use datalib.

kanitw:
As we need to make it work on gh-pages, I add lib/datalib.js and gulp task for copying it from the installed node_module.

### DIFF
--- a/editor/editor.js
+++ b/editor/editor.js
@@ -1,6 +1,6 @@
 'use strict';
 
-/*global location, d3, vl, vg, docCookies, document, $, alert */
+/*global location, d3, vl, vg, docCookies, document, $, alert, dl */
 
 var DATASETS = [
   {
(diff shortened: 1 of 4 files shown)


editor/editor.js
```diff
@@ -154,8 +154,10 @@ vled.datasetChanged = function(dataset, callback) {
 
   d3.json(dataset.url, function(err, data) {
     if (err) return alert("Error loading data " + err.statusText);
-
-    dataset.stats = vl.data.getStats(data);
+    dataset.stats = dl.summary(data).reduce(function(s, p) {
```
domoritz: are you sure this is correct?

kanitw: Hmmm I think we need to add `,{}`  add the end

domoritz: Yep, makes sense.

PR #427, opened 2015-05-09
Embedded data

domoritz:


kanitw:
I suspect that when we implement data management layer for voyager, we might need to bring back the preAggregate flag.  But since I’m not really sure.  I’m gonna merge this.

The only change I made is making stats calculation for values using `reduce` too, for consistency.

PR #428, opened 2015-05-09
Gridlines

domoritz:
Infinitely better looking charts.


kanitw:
While I agree that this looks much better, I don’t think it’s completely solves the layering issue though.

PR #439, opened 2015-05-22
bar positioning 

### LINKED ISSUE #388: Log Broken with Bar
should fix by either fixing min for log scale at 1   or must set x/y value to the min value of the scale (is there a way to do that in vega?)

pole

kanitw:
- fixes #388  -- log broken with bar (set `y2 = {group:'height'}`, and `x2= {value:0}`) + unit test
- add offset for single bar + unit test
- also set serve task to use google chrome

### DIFF
--- a/gulp/serve.js
+++ b/gulp/serve.js
@@ -4,9 +4,10 @@ var gulp = require('gulp');
 var browserSync = require('browser-sync');
 
 gulp.task('serve', ['copydl', 'bundle', 'watch-schema', 'watch-test'], function() {
-    browserSync({
-        server: {
-            baseDir: './'
-        }
-    });
+  browserSync({
+    server: {
+        baseDir: './'
+    },
+    browser: 'google chrome'
+  });
 });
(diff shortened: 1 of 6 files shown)


domoritz:
You should use the latest master. It fixes the build.

domoritz:
Still not perfect. 

```json
{
  "marktype": "bar",
  "enc": {
    "x": {"type": "O","name": "Origin"},
    "y": {"scale": {"type": "log"},"type": "Q","name": "Acceleration","aggr": "avg"}
  },
}
```

screen shot 2015-05-21 at 23 34 54

kanitw:
@domoritz  What do you mean still not perfect?   This solves the problem in https://github.com/uwdata/vega-lite/issues/388.

domoritz:
I should have been re specific. What I mean is that the y range should only go to 20, not 100.

kanitw:
That's a different issue.

kanitw:
Also I think it goes to 100 due to `nice` function's behavior so I don't think it's an issue we need to fix here  (unless you come up with a more objective `nice` function).

domoritz:
Agree

PR #440, opened 2015-05-22
Fixes 389

kanitw:
**Please merge #439  first**

- fixes #389 (range incorrect for binning with color)

PR #441, opened 2015-05-22
Eliminate bitcode type 

kanitw:
**Please merge #439 and #440 first**

- always use string type ('O', 'Q', 'T') instead of bitcode (1,2,4)


domoritz:
This makes the code so much simpler. I love it.

Issue #436, opened 2015-05-14
Discussion: full names for keys and values


ajturner:
In general it would be an improvement for developer experience to use full names when appropriate. for example `"type": "O"` is not clear to someone reading or working on Vega or Vega-Lite documents that this means `Ordinal`. Or `"aggr": "avg"` is `"aggregate": "average"`.

There would be a minimal JSON size growth by using full names but would dramatically improve legibility and editing.

kanitw:
For "O" vs "ordinal", I think this is a reasonable point, we will reconsider when it's closer to 1.0 release. 

For `avg`, I think it's a pretty common function name in other software including SQL, Tableau and Qlikview.
`aggr` is less common but I think it's very easy to guess.  QlikView also use it too.

ajturner:
Thanks for the response.

I generally treat API and specs like this as Developer UX. Avoiding having to guess - or remember shorthand - is preferable.

Similarly the `enc` at the root. Is that 'encoding'?

kanitw:
We actually planned to change `enc` to `encoding`.
Thinking more about your earlier feedback, we will change `aggr` to `aggregate`.

For aggregation function names, we would make it consistent with https://github.com/uwdata/vega.

Issue #20, opened 2014-12-02
Formula Transform (akin to Calculated Field in Tableau)


kanitw:
- Numeric Transform – formula e.g., log, etc.

kanitw:
Duplicate with #451

PR #447, opened 2015-05-23
Syntax Refactor

kanitw:
fixing #436 

- [x] `.enc` => `.encoding`
- [x] `.aggr` => `.aggregate`
- [ ] `.fn` => ???

~~`aggregate: 'avg'` => `aggregate: ‘average’`~~ — We will change this once we migrate to Vega2.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -9,7 +9,7 @@
   "predef": [
     "TABLE", "RAW", "STACKED", "INDEX",
     "X", "Y", "ROW", "COL", "SIZE", "SHAPE", "COLOR", "ALPHA", "TEXT", "DETAIL",
-    "O", "Q", "T"],
+    "O", "Q", "T", "N"],
   "globalstrict": true
 }
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Use Vega-lite in the [online editor](https://uwdata.github.io/vega-lite/).
 If you are using Vega-lite for your project(s), please let us know by emailing us at [Vega-lite \[at\] cs.washington.edu](mailto:vega-lite@cs.washington.edu).  Feedbacks are also welcomed.
 If you find a bug or have a feature request, please [create an issue](https://github.com/uwdata/vega-lite/issues/new).
 
-The complete schema for specifications as [JSON schema](http://json-schema.org/) is at [spec.json](https://uwdata.github.io/vega-lite/spec.json). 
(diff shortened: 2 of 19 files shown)


kanitw:
@jheer  @domoritz   currently `.fn` refers to time abstract function (year, month, day, etc.).

How should we name it?   
`.function` is a candidate although it’s bad due to conflict with reserved `function` keyword. 

Also, we might want to consider how to support formula transform #20 in the future.
(For example, deriving new field.) 

Should deriving new field be just a part of `.data` akin to vega?  (using Formula transform). 

Should time abstraction function still be part of field?

kanitw:
@domoritz  This PR is blocking #450 feature.   Maybe we can punt on `fn` for now?

kanitw:
I was looking at R and  maybe `as` is not a bad alternative to `fn`

```
{
name: ‘Release_Date’
type: ’T’,
as: ‘year’
```

??

domoritz:
Alternatives to `as` are `map`, `get`, or `accessor`. Another option is passing down expressions to vega. What I am thinking is that instead of `"x": {"type": "T","name": "foo", "fn": "year"}` you would write `"x": {"type": "T","value": "year(foo)"}`. Thoughts?

kanitw:
The question here is do we treat time abstraction similar to other value derivation?

In my head right now,  time abstraction is a common task and should be supported via a simple mean like `as: 'year'` and easily name the derived field `year_Release_Date` (akin to naming aggregated fields)

However, more complex formula for value derivation should be supported as part of `.data` as free form formula would require naming.   For example, suppose we have some finance data,  `earning / no_of_share`  should be manually named `earning_per_share`.   

For this, maybe (a) we can simply use vega's formula transform syntax?

```js
{
  data: { 
    url/values: ...,
    transforms: [
      {type: 'formula', field: 'earning_per_share', expr: `d.earning / d.no_of_share` // no d.data in vg2? 
    ]   
  }
}
```

or (b) add `derive` property so we only support formula transform
```
{
  data: { 
    url/values: ...,
    derive: [
      {field/name: 'earning_per_share', expr: `d.earning / d.no_of_share`}
    ]   
  }
}
```

**In any case, please merge this first if you agree with other parts as this is blocking other PR.**

domoritz:
I agree with the rest of the pr but not with the function stuff. I don't see why there should me multiple ways to express the same thing and `value: "year(foo)"` seems easy and intuitive. The more I think about this, the less I understand why we have a `name` property instead of `value`. I'll create a new issue for this.

PR #455, opened 2015-05-27
Test coverage using istanbul

kanitw:
Now we can look at test coverage online at https://coveralls.io/r/uwdata/vega-lite.  

Right now, we’re at 63% — quite low, but honestly surprisingly high given how much test we wrote. 

I am figuring out about local/offline line coverage reporter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ vega-lite*.map
 spec.json
 npm-debug.log
 instance.json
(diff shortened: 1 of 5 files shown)


kanitw:
Actually coverage report are automatically included by default in `coverage/lcov-report/index.html`. 

This is so easy!

PR #470, opened 2015-06-08
Datalib related improvements

domoritz:


kanitw:
We shouldn’t remove `lib/datalib.js` because `index.html` is referring to it.  

We can refer to `datalib.js` via `node_modules/datalib/datalib.js` but we also need to modify publish.sh to publish `node_modules/datalib/` to gh-pages.

kanitw:
Other than that.  This PR looks good.

domoritz:
@kanitw True but I removed that it actually needs it. Fixed in d7d43e5

kanitw:
Okay.  Make sense.  Thanks.

PR #473, opened 2015-06-18
Use the newly baked d3-color and the shim colorbrewer module

### LINKED ISSUE #456: Use d3.color once d3 becomes modular
Right now I manually extracted d3’s color modules and put in `src/lib`

kanitw:
- Use the newly baked d3-color 

fixes #456


kanitw:
I also make another commit use the shim [colorbrewer module](https://github.com/saikocat/colorbrewer) 

This is not the official module but I think it’s cleaner than having `src/lib/colorbrewer` directory inside our module.

PR #476, opened 2015-06-26
Fix wrong data schema order 

kanitw:
- fixes https://github.com/uwdata/vega-lite-ui/issues/37
- use c10/20palette config

### DIFF
--- a/src/field.js
+++ b/src/field.js
@@ -110,7 +110,9 @@ vlfield.order.type = function(field) {
 };
 
 vlfield.order.typeThenName = function(field) {
-  return vlfield.order.type(field) + '_' + field.name.toLowerCase();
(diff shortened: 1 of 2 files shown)


src/field.js
```diff
@@ -110,7 +110,9 @@ vlfield.order.type = function(field) {
 };
 
 vlfield.order.typeThenName = function(field) {
-  return vlfield.order.type(field) + '_' + field.name.toLowerCase();
+  return vlfield.order.type(field) + '_' +
+    (field.aggregate === 'count' ? '~' : field.name.toLowerCase());
```
domoritz: This will break if a field is called `~`. Just sayin.

kanitw: Feel free to make it better.  This is  a bad fix just for screenshot sake.

PR #479, opened 2015-06-27
title offset fixes

domoritz:
Fixes a bunch of issues. See commits for details.


kanitw:
I have similar fix that I don’t have to commit.  Also I don’t want to change `Horse Power` => `HorsePower` in this review round as we are running short of time and I prefer to read through the whole paper rather than making this fix.  

If I’m done with typos/grammars fix I’ll take a look at this again.

domoritz:
Can you merge?

kanitw:
I might override ur getMaxLength later

PR #482, opened 2015-06-30
Fixed array iteration bug

vivekratnavel:
Faced an issue with array iteration while using vega-lite inside an Ember.js project. 

The code below iterates all the properties attached to an array like "_super", "contains", "getEach" etc.

```javascript
for (var i in filters) {
  ...
}
```

Fixed it by checking if the property is the array's own property using "hasOwnProperty()" method.

### DIFF
--- a/src/compiler/filter.js
+++ b/src/compiler/filter.js
@@ -21,39 +21,43 @@ filter.addFilters = function(rawTable, encoding) {
 
   // add custom filters
   for (var i in filters) {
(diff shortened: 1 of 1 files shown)


kanitw:
@vivekratnavel   Thank you for catching the bug.   

I think this is better fixed using [the classic for loop](http://bonsaiden.github.io/JavaScript-Garden/#array.general), so I created #483 to fix this instead.

PR #483, opened 2015-07-01
fix array iteration in filter #482

kanitw:
Better alternative for #482


vivekratnavel:
@kanitw : You need to change the "for loop" to classic in the below code also (Line 44)

```javascript
// expects a number of fields
  for (var j in operands) {
  ...
  }
```

Thank you.

vivekratnavel:
@kanitw @domoritz I have created another pull request with desired changes - #484

PR #489, opened 2015-07-09
Feeding scale with raw data for supporting aligning axis with same domain 

### LINKED ISSUE #475: Option to feed scale with raw data for aggregate plot

kanitw:
fixes #475 

To support reading across chart in Voyager, we need an option to use RAW table as the data source for a scale’s domain.  

**TODO**
- [x] Unit Tests

Should we find some way to align scale for `COUNT`


kanitw:
@domoritz  ready to be reviewed now!

PR #499, opened 2015-07-14
Kw/enc field

kanitw:
- refactoring `encoding.field` => `encoding.fieldRef`
- refactor `encoding.enc()` to `encoding.field()` because it actually returns vega-lite field
- replace `encoding.legend|font|text()` with `encoding.field(et).legend|font|text
- Set config for `text` field, collapse `text.text.properties` to be just `text.properties`, add more properties
- make `default: undefined` explicit for `band.size`

+ migrate palette configs into color.scale 
+ add explicit `default:undefined` and description for color.scale.range
+ remove .role() as it is confusing with role in schema and it is only used once in compass -- so I will adding it to compass directly


domoritz:
LGTM but we will need to test extensively to make sure that we didn't miss to rename a function.

domoritz:
@kanitw Please test the editor. I cannot rest atm.

kanitw:
What do you mean "test the editor".   (I'm testing by looking if voyager output correctly.)

Issue #497, opened 2015-07-12
Number format for axis label and text mark


kanitw:
Many cases to handle:

- [ ] raw
  - [ ] integer vs floating point
    - From Vega’s data parsing, we do not know the difference between integer and real, but they should have different numberFormat setting.
    - [ ] provide a syntax for distinguishing integer/discrete from real/continuous – maybe `numberType=discrete|continuous`
    - [ ] float: big number vs small number?
- [ ] aggregate 
  - [ ] avg (or other things with division)
  - [ ] min / max (semantic still based on raw)
  - [ ] sum (semantic still based on raw — but could be very large) 
  - [ ] count, distinct() -- always integer

kanitw:
Actually, currently we only call `dl.summary`, which in turn only calls `dl.type()` but not `dl.infer.type()`.   

Therefore, we do not really know if a number field is an integer field or not.  
For vega-lite-ui, the question is whether we should call inferAll for each imported dataset.  
For small dataset, this should be fine.   For a large dataset, this could be dumb.   
Inferring type of samples might be a compromise but I'm not sure if that is the ideal solution.

kanitw:
A bigger issue is how should the syntax look like if we allow users to provide both information about type as level of measurement (N, O, Q, T) and type as primitive data type (integer, number, string, ...)

kanitw:
Using a better setting of numberFormat, this becomes less important -- so I'm closing this for now.

kanitw:
#500 is a reasonable good step, but we should create a set of visual test to test this.

kanitw:
![vega-lite_editor](https://cloud.githubusercontent.com/assets/111269/8818058/f2cf03d8-3009-11e5-8ac3-eac2045e021d.png)
For mean value in `TEXT` , `’’` is a bad formatter.   

```
{
  "marktype": "text",
  "encoding": {
    "row": {"name": "Name","type": "N"},
    "col": {"name": "Cylinders","type": "O"},
    "color": {"name": "Acceleration","type": "Q","aggregate": "avg"},
    "text": {"name": "Acceleration","type": "Q","aggregate": "avg"}
  },
  "data": {"url": "data/cars.json"}
}
```

kanitw:
We are currently doing pretty bad with integer, so this property is definitely needed

[...]

PR #500, opened 2015-07-14
Axis label format (1st version)

kanitw:
#497 Dealing with axis label format without knowledge about integer.


kanitw:
@domoritz ready to be merge so I can test and create new figures.

PR #501, opened 2015-07-14
Eliminate redundant help in encoding.

kanitw:
@domoritz   

I kinda feel annoy to have helpers in `encoding` for every single property of a field.   
That said, we still need some of them.  So I'm not sure if we should do this.  
That said, I created this branch on the plane, so would love to know what you think.


domoritz:
I guess the hope is that you can rename something without having to rename it everywhere. But since we rename all the time, let's get rid of this indirection.

PR #516, opened 2015-07-19
Axis label offset

kanitw:
Please merge #507 first.  [See Branch Specific Diff](https://github.com/uwdata/vega-lite/compare/kw/useRawDomain-config...kw/axisLabelOffset)

- make better offset for timeAxis 
   - `month`/`day`
   - for raw time —  take `timeFormat` into account
   - Q
   - special case for ordinal based on number
- remove redundant `numberFormat` config in axis and add `format` to textMixin
- add `dayScaleLabel`, `monthScaleLabel`

- refactor `time.js`

- refactor layout.offset

- add `vl.data.stats`

- add some tests

### DIFF
--- a/src/schema/schema.js
+++ b/src/schema/schema.js
@@ -661,20 +661,20 @@ var config = {
     },
     maxSmallNumber: {
       type: 'number',
-      default: 1000,
+      default: 10000,
       description: 'maximum number that a field will be considered smallNumber.'+
                    'Used for axis labelling.'
     },
     smallNumberFormat: {
       type: 'string',
-      default: ',g',
-      description: 'Number format for axis labels and text tables '+
+      default: '',
+      description: 'D3 Number format for axis labels and text tables '+
                    'for number <= maxSmallNumber. Used for axis labelling.'
     },
     largeNumberFormat: {
       type: 'string',
       default: '.3s',
-      description: 'Number format for axis labels and text tables ' +
+      description: 'D3 Number format for axis labels and text tables ' +
                    'for number > maxSmallNumber.'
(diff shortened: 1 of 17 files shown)


src/schema/schema.js
```diff
@@ -634,7 +639,24 @@ var config = {
...
-      minimum: 0
+      minimum: 0,
+      description: 'Max length for values in dayScaleLabel and monthScaleLabel.  Zero means using full names in dayScaleLabel/monthScaleLabel.'
+    },
+    dayScaleLabel: {
+      type: 'array',
+      items: {
+        type: 'string'
+      },
+      default: ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday'],
```
domoritz: How do you know that the week starts on a Monday?

kanitw: Excellent question -- you just catched a bug!!!  It actually started on Sunday.     

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay

domoritz: I learned that they start on Sunday the hard way when I scheduled a few meetings on Sundays ;-)

domoritz:
Hmm, this broke the histogram in the gallery. `d3_format is not defined`

kanitw:
I will take a look at that then.

kanitw:
Rebuild vega-lite  and my recent commit 7c627ff9dbc7e2289b8ddef300f108ea04ca3db0 would eliminate the error.

PR #524, opened 2015-07-20
bar actually only requires x or y

kanitw:
This makes Compass recommends bar for single aggregated value instead.  :)

data_voyager


domoritz:
@kanitw can't merge

PR #527, opened 2015-07-21
Make Legends as similar as marks as possible

### LINKED ISSUE #490: Legends should look as similar as marks as possible
- [x] If alpha is applied to point, it should be applied to legends as well 
- [x] symbols in legends should look similar to point style

kanitw:
fixes #490	

- [x] If alpha is applied to point, it should be applied to legends as well 
- [x] symbols in legends should look similar to point style

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -9,7 +9,7 @@
   "quotmark": "single",
   "predef": [
     "TABLE", "RAW", "STACKED", "INDEX",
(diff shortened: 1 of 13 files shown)


kanitw:
vega-lite_editor

Legends looks like the shape!

kanitw:
Bars’s legend is square.  

```
{
  "marktype": "bar",
  "encoding": {
    "x": {"name": "Acceleration","type": "Q","aggregate": "avg"},
    "y": {"name": "Cylinders","type": "O"},
    "color": {"name": "Origin","type": "O"}
  }
```

pole

domoritz:
This is cool.

PR #530, opened 2015-07-22
Fix Missing text tables, simplify a few things

kanitw:
- Remove `mdef` from `scales.def` and  `facet`
- fixes missing text tables
- simplify marks.def


src/compiler/compiler.js
```diff
@@ -88,12 +87,16 @@ compiler.compileEncoding = function (encoding, stats) {
     mdef.from.transform = [{type: 'sort', by: '-' + encoding.fieldRef(f)}];
   }
 
+  var singleScaleNames = [].concat.apply([], mdefs.map(function(markProps) {
```
domoritz: WTF does this code do...

PR #535, opened 2015-07-30
Better ordinal scale

kanitw:
@domoritz  this is probably good enough for the video - but need some clean-up after deadline — especially fixing tests.


kanitw:
Just fixes the test.

PR #543, opened 2015-08-08
A simple logger that preserves line number

kanitw:
Current logger function, which follow [vega-logging](https://github.com/vega/vega-logging)’s pattern does NOT show the right line number which is not particularly useful.  

Borrowing some ideas from stack overflow and some blog posts, this is a simple logger wrapper that still output the correct line number when `log`, `warn`, `error`, `info` or `debug` are called.  

cc: @arvind

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,0 +1,19 @@
+'use strict';
+
+// TODO(kanitw): chat with Vega team and possibly move this to vega-loggger
+module.exports = function(prefix) {
+  // Borrowed some ideas from http://stackoverflow.com/a/15653260/866989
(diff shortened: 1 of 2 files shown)


kanitw:
Note that there are people chatting about different console.log wrapper that perfectly works in older browser.  (Like IE9 doesn’t have `console.log` properly implemented.)
- http://www.briangrinstead.com/blog/console-log-helper-function
- http://patik.github.io/console.log-wrapper/
- http://benalman.com/projects/javascript-debug-console-log/

but for now, this is suffice for my own debugging.

PR #545, opened 2015-08-08
eliminiate unnecessary complexity in `encoding.data`

kanitw:
eliminiate unnecessary complexity in `encoding.data`

Please merge #544 first!


domoritz:
I like this.

PR #547, opened 2015-08-09
Decouple template.js

kanitw:
- Add data.js and move data related part in template.js to data.js
- Just create spec template and append mark group directly from compiler.js so we can remove template.js
- migrate test from template.spec.js to data.spec.js and add more tests

(Things will be further migrated to data.js in the next PRs.)
 
**Please merge #546 and #542 first !**  (See [Branch Specific Diff](https://github.com/uwdata/vega-lite/compare/kw/de-sorting...kw/decouple))


src/compiler/data.js
```diff
@@ -0,0 +1,58 @@
...
+
+module.exports = data;
+
+var vlfield = require('../field');
+
+function data(encoding) {
+  var def = [data.raw(encoding)];
+
+  // TODO(kanitw): Aug 8, 2015 - if aggregate add TABLE
+  def.push(data.aggregated(encoding));
```
domoritz: Won't some of this become easier with vega 2? Should we wait for it?

kanitw: With Vega 2, we still need to split the two .. But we  have to do less work in terms of sorting

PR #549, opened 2015-08-10
Add legend.title property and axis.labelAngle property

osnr:
See https://github.com/uwdata/vega-lite/issues/181.

Needed these for a project:
- had a computer-generated field with an ugly title that we didn't want in the legend, so added `legend.title`
- date labels were hard to read vertically so we wanted to force them horizontal, so added `axis.labelAngle`

Feedback welcome. Should maintain compatibility with existing specs.


kanitw:
Looks good to me.  Thanks for your contribution! @osnr

PR #556, opened 2015-08-17
Added unit tests for point and line

vivekratnavel:
I have added unit tests for point and line. I am planning to add more tests for point and line as well other mark types. But before that wanted to check if I am going in the right direction.

### DIFF
--- a/test/compiler/marks.spec.js
+++ b/test/compiler/marks.spec.js
@@ -71,5 +71,140 @@ describe('compile.marks', function() {
     });
   });
 
+  describe('point', function() {
+    describe('1D, horizontal', function() {
+      var f = fixtures.points['1d_hor'],
+          e = Encoding.fromSpec(f),
+          def = marks.point.prop(e, mockLayout, {});
+      console.log("def: ");
+      console.log(def);
+      it('should end on y axis', function() {
+        expect(def.y).to.eql({value: e.bandSize(Y, mockLayout.y.useSmallBand) / 2});
+      });
+      it('should scale on x', function() {
+        expect(def.x).to.eql({scale: X, field: "data.year"});
+      });
+    });
+
+    describe('1D, vertical', function() {
+      var f = fixtures.points['1d_ver'],
+          e = Encoding.fromSpec(f),
+          def = marks.point.prop(e, mockLayout, {});
+      console.log("def: ");
+      console.log(def);
+      it('should end on x axis', function() {
+        expect(def.x).to.eql({value: e.bandSize(X, mockLayout.x.useSmallBand) / 2});
+      });
+      it('should scale on y', function() {
+        expect(def.y).to.eql({scale: Y, field: "data.year"});
+      });
+    });
+
+    describe('2D, x and y', function() {
+      var f = fixtures.points['x_and_y'],
+          e = Encoding.fromSpec(f),
+          def = marks.point.prop(e, mockLayout, {});
+      console.log("def: ");
+      console.log(def);
+      it('should scale on x', function() {
+        expect(def.x).to.eql({scale: X, field: "data.year"});
+      });
+      it('should scale on y', function(){
+        expect(def.y).to.eql({scale: Y, field: "data.yield"});
+      });
+    });
+
+    describe('3D', function() {
+      describe('size', function () {
+        var f = fixtures.points.size,
+            e = Encoding.fromSpec(f),
+            def = marks.point.prop(e, mockLayout, {});
+        console.log("def: ");
+        console.log(def);
+        it('should have scale for size', function () {
+          expect(def.size).to.eql({scale: SIZE, field: "data.count"});
+        });
+      });
+
+      describe('color', function () {
+        var f = fixtures.points.stroke,
+            e = Encoding.fromSpec(f),
+            def = marks.point.prop(e, mockLayout, {});
+        console.log("def: ");
+        console.log(def);
+        it('should have scale for color', function () {
+          expect(def.stroke).to.eql({scale: COLOR, field: "data.yield"});
+        });
+      });
+
+      describe('shape', function () {
+        var f = fixtures.points.shape,
+            e = Encoding.fromSpec(f),
+            def = marks.point.prop(e, mockLayout, {});
+        console.log("def: ");
+        console.log(def);
+        it('should have scale for shape', function () {
+          expect(def.shape).to.eql({scale: SHAPE, field: "data.bin_yield"});
+        });
+      });
+    });
+  });
+
+  describe('line', function() {
+    describe('1D, horizontal', function() {
+      var f = fixtures.lines['1d_hor'],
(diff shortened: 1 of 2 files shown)


test/fixtures.js
```diff
@@ -60,4 +60,100 @@ f.stack.binX = {
...
+};
+
+f.points['x_and_y'] = {
+  "marktype": "point",
+  "encoding": {"x": {"name": "year","type": "O"},"y": {"name": "yield","type": "Q"}},
+  "config": {"singleHeight": 400,"singleWidth": 400,"largeBandMaxCardinality": 20},
+  "data": {"url": "data/barley.json"}
+};
+
+f.points.size = {
```
kanitw: maybe `f.points[‘x,y,size’]`

test/fixtures.js
```diff
@@ -60,4 +60,100 @@ f.stack.binX = {
...
+};
+
+f.points['1d_ver'] = {
+  "marktype": "point",
+  "encoding": {"y": {"name": "year","type": "O"}},
+  "config": {"singleHeight": 400,"singleWidth": 400,"largeBandMaxCardinality": 20},
+  "data": {"url": "data/barley.json"}
+};
+
+f.points['x_and_y'] = {
```
kanitw: `‘x,y’` to make it consistent with my comment below

test/fixtures.js
```diff
@@ -60,4 +60,100 @@ f.stack.binX = {
...
+  "encoding": {
+    "x": {"name": "year","type": "O"},
+    "y": {"name": "yield","type": "Q"},
+    "color": {"name": "yield","type": "Q"}
+  },
+  "config": {"singleHeight": 400,"singleWidth": 400,"largeBandMaxCardinality": 20},
+  "data": {"url": "data/barley.json"}
+};
+
+f.points.shape = {
```
kanitw: `’x,y,shape’`

test/compiler/marks.spec.js
```diff
@@ -71,5 +71,120 @@ describe('compile.marks', function() {
     });
   });
 
+  describe('point', function() {
+    describe('1D, horizontal', function() {
+      var f = fixtures.points['1d_hor'],
+          e = Encoding.fromSpec(f),
+          def = marks.point.prop(e, mockLayout, {});
+      it('should end on y axis', function() {
```
kanitw: 'should be centered'

test/compiler/marks.spec.js
```diff
@@ -71,5 +71,120 @@ describe('compile.marks', function() {
...
+      it('should scale on x', function() {
+        expect(def.x).to.eql({scale: X, field: "data.year"});
+      });
+    });
+
+    describe('1D, vertical', function() {
+      var f = fixtures.points['1d_ver'],
+          e = Encoding.fromSpec(f),
+          def = marks.point.prop(e, mockLayout, {});
+      it('should end on x axis', function() {
```
kanitw: ‘should be centered’

test/fixtures.js
```diff
@@ -60,4 +60,100 @@ f.stack.binX = {
...
+  "encoding": {
+    "x": {"name": "year","type": "O"},
+    "y": {"name": "yield","type": "Q"},
+    "size": {"name": "*","type": "Q","aggregate": "count"}
+  },
+  "config": {"singleHeight": 400,"singleWidth": 400,"largeBandMaxCardinality": 20},
+  "data": {"url": "data/barley.json"}
+};
+
+f.points.stroke = {
```
kanitw: f.points[‘x,y.color’]

test/compiler/marks.spec.js
```diff
@@ -71,5 +71,120 @@ describe('compile.marks', function() {
...
+            def = marks.point.prop(e, mockLayout, {});
+        it('should have scale for shape', function () {
+          expect(def.shape).to.eql({scale: SHAPE, field: "data.bin_yield"});
+        });
+      });
+    });
+  });
+
+  describe('line', function() {
+    describe('1D, horizontal', function() {
```
kanitw: I don’t think we should test 1D line as it is not a sensible encoding.  
(line should have both x and y).

kanitw:
@vivekratnavel 

Thank you for the PR.  I’m really glad that you’re helping us improving test coverage. 

I think you’re starting with the right direction.  Also thank for sending a small and quite atomic PR, which makes reviewing very easy.  

I added a few comments and would be happy to merge this after you address those comments (which shouldn’t take too much time).   
If you find my older code, which is written in a more prototyping fashion, inconsistent with my comment please feel free to bug me to refactor them as well.  


For testing other marks, I think the code for marks should be mostly reasonable (maybe 95%) but if you find a part where the logic might not make sense, please feel free to send us questions.  
You might also find running `gulp coverage` useful for seeing how much have you help us cover :)

vivekratnavel:
@kanitw 

Thank you for your comments. I will make required changes and let you know.

vivekratnavel:
@kanitw 

I made all the necessary changes and added a couple of new tests for mark type 'area'. Please review the changes and let me know.

Also, when will this part of the code get executed in marks.js
```
if (e.isMeasure(X)) {
    p.x = {scale: X, field: e.field(X)};
    if (e.isDimension(Y)) {
      p.x2 = {scale: X, value: 0};
      p.orient = {value: 'horizontal'};
    }
  }
```
How to test this?

test/compiler/marks.spec.js
vivekratnavel: Makes sense. Removed the 1D tests for line and committed.

test/fixtures.js
```diff
@@ -60,4 +60,109 @@ f.stack.binX = {
...
 };
 
+// POINT
+
+f.points = {};
+
+f.points['1d_hor'] = {
+  "marktype": "point",
+  "encoding": {"x": {"name": "year","type": "O"}},
+  "config": {"singleHeight": 400,"singleWidth": 400,"largeBandMaxCardinality": 20},
```
kanitw: Reading this one more time, I am guessing that we actually do not need the custom config for any of these fixtures.  

(I saw that my old code has some of these leftover as well, so I probably I need clean to them up.)

kanitw:
Re: 
```
if (e.isMeasure(X)) {
    p.x = {scale: X, field: e.field(X)};
    if (e.isDimension(Y)) {
      p.x2 = {scale: X, value: 0};
      p.orient = {value: 'horizontal'};
    }
  }
```

To test it, you probably want to read how we define `isDimension()` (and `isMeasure()`) in `field.js`.  
and test by providing measure as x and dimension as y and vice versa.   
Since area marks with only x or y do not make sense, I will have it logs error when only x or y is provided for area mark type (#558).

To make this PR small, maybe let’s just remove “config” from those fixtures so I can merge and you can send in another PR for other tests you want to help write? :)

vivekratnavel:
Makes sense. Removed "config" from all fixtures.

kanitw:
Thank you so much :)

Issue #533, opened 2015-07-24
sorting breaks for line and area marks


richardhundt:
Hi - you can observe this in Polestar as well - if you choose line or area types and try to sort, it breaks in visually interesting ways. It looks like the axes are sorted, but the vectors connecting the points aren't aware of the sorting.

I'm doing the following, and the axes are sorted correctly and for bar and point types it works (all my aggregations are done server-side, hence the `aggregate: 'max'`):

```JavaScript
        spec.encoding.x.sort = [{
           name: 'y',
           aggregate: 'max',
           reverse: true
         }];
```

The problem seems to relate to the `sort` attribute on the scales overriding the data sorted via a transform. It also looks like Vega itself sorts ordinal scales even if the `sort` field is not set, and since nominal maps to ordinal scales, it gets sorted lexicographically. While this makes sense for ordinal values, it's a bit arbitrary for nominals. In particular, when plotting quantitative against nominal, I think that what you'd normally want is to sort the nominal by the values of the quantitative scale.

kanitw:
Thank you for reporting.  We plan to fix this after we migrate to use Vega 2 :)

richardhundt:
I noticed some traces of the migration in the vega-lite code, thanks. Are there any blockers for the migration? If not then I can take a crack at it.

kanitw:
@richardhundt  we wanted to create more tests, refactor code and fix some bugs before migrating.

kanitw:
@richardhundt Do you have an example spec / data that I can test this?

kanitw:
Sort property is totally rewritten based on changes in Vega 2.  So I do not think this issue is no longer applicable.  If you find similar problems with the current version of Vega-Lite, please re-create another issue.

Issue #341, opened 2015-03-08
Refactor/extend sort


kanitw:
Forked from #27  

### Summary 

I try to summarize what we can do and we we need to do with sorting here. 
We can only sort nominal/ordinal fields.   We can sort them in the following ways. 

- [ ] temporal *** 

#### 1) manual order
- setting domain directly (and maybe add filter)
- [ ] support manual order
  - [ ] test what happens if we specify domain and there is a value not exist 
-  we might need to filter.
  - [ ] extend the scale — (append that at the end)

#### 2) alphabetical 
Vega 1: `sort:true`

- [x] `dl.cmp` — should be what vega already used. 
- [ ] support `reverse`  (now it doesn’t do anything)

#### 3) original order of the data 
Vega 1:  `sort: false`

#### 4) sort by aggregated value.  

There are a few cases:

4-a) **raw plot**  
We can sort by aggregated value of any column (not matter it’s included in the encoding or not). 
In Vega 1, this **requires** generating a separate aggregated table.  
because there might be multiple values for each O value of the column. 
However, Vega 2 we can use scale’s `sort` property directly. 

4-b) **aggregate plot with one dimension**.  
This does **not require** generating a separate aggregated table in both vega 1 and 2 as 
we can augment the generated aggregated table with the measure sorted on.

4-c) **aggregate plot with multiple dimensions** 

We can sort an ordinal column by aggregated value of another column.  This is quite complicated because the original aggregated table isn’t sufficient since there are still multiple aggregated value for each value of the column.   There are two ways to handle this: (1) create a separate aggregated table from the source (2) aggregate the aggregated table.   We should support (1) for now as (2) doesn’t make much sense. 

- [ ] support 1.

- [ ] support filter in the sort by "aggregated with filter” 

In addition, a commonly used sort includes filtering. 

For example if we have two ordinals O<sub>1</sub> and O<sub>2</sub> and one Q.    
We can sort O<sub>1</sub> by `aggr(Q)` where O<sub>2</sub> = `<value>`

![sort__product__and_tableau_-_book2](https://cloud.githubusercontent.com/assets/111269/6313041/699f7cd4-b94c-11e4-8609-3f0c4f776ecf.png)

----------------

### Other TODO

- [x] refactor sort to be consistent with vega2’s sort

- [ ] add `type` to sort’s schema

- [ ] auto sort that include ascending / descending  -- right now toggle sort has fixed ascending/descending https://github.com/uwdata/vegalite/pull/343/files

[...]

PR #557, opened 2015-08-18
Allow field sort to be explicitly disabled in line chart.

osnr:
related to #533 and #341? not sure if this is best way to do it (uneasy about mixing object/bool in `encoding.sort`), but I *really needed* this functionality.

I think sorting is a bad default, actually -- it's not hard to sort the data upfront before giving it to vega-lite, but it's impossible to go the other way if vega-lite forces it to be sorted.

### DIFF
--- a/src/Encoding.js
+++ b/src/Encoding.js
@@ -203,6 +203,8 @@ module.exports = (function() {
       isTypes = vlfield.isTypes;
 
     if ((!sort || sort.length===0) &&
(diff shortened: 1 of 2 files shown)


kanitw:
Please see if #559 that works for you.   This does not work because we does not work because we declare sort properties to be `Array` type in the [JSON schema](https://github.com/uwdata/vega-lite/blob/master/src/schema/schema.js#L215) and therefore it can’t be `Boolean`.

PR #559, opened 2015-08-19
allow disabling auto-sort for line

kanitw:
@osnr   Since you need it, I’m adding this as a config.    

Note that your PR #557 does not work because we declare sort properties to be `Array` type in the [JSON schema](https://github.com/uwdata/vega-lite/blob/master/src/schema/schema.js#L215) and therefore it can’t be `Boolean`.  

Please let me know if this works for you.  

Note that this is subject to change once we migrate to Vega 2.

### DIFF
--- a/src/compiler/compiler.js
+++ b/src/compiler/compiler.js
@@ -84,8 +84,7 @@ compiler.compileEncoding = function (encoding, stats) {
   }
 
   // auto-sort line/area values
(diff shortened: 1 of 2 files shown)


kanitw:
Basically provide `“config”: {“autoSort”: false}` should disable the auto-sort.   

I’m considering to change the default but will think more when I work on migrating this part to vega 2.   

Hope this is sufficient for now.

osnr:
looks great -- haven't gotten to try it yet but thanks so much for doing this!
will let you know soon

PR #589, opened 2015-10-03
Eliminate group.js

kanitw:
Simplify the code base by eliminating group.js


src/compiler/compiler.js
```diff
@@ -52,29 +51,29 @@ compiler.compileEncoding = function (encoding, stats) {
       padding: 'auto',
       data: compiler.data(encoding),
       // global scales contains only time unit scales
-      scales: compiler.time.scales(encoding)
+      scales: compiler.time.scales(encoding),
+      marks: [{
+        _name: 'cell',
```
arvind: `name` is an officially supported property for marks with Vega 2.0, and can be useful when adding mark-specific interaction. E.g., `@cell:mousedown` references `mousedown` events that occur in the mark named `cell`.

kanitw: Thanks, great that you comment.  I forgot to jot this down in the issue tracker!  
Now fixed in #596

PR #591, opened 2015-10-03
Instantiate LONG_DATE constant using Date.UTC

kanitw:
so that a test case doesn't simply fail because I'm in Thailand!


jheer:
Note that for test cases you can explicitly set the time zone. See datalib's package.json for an example. Using UTC rather than local time internally may have unintended consequences, so tread carefully.

kanitw:
Thank you.  For here, I only create the time to determine max string length for temporal value.
This is actually not the best way to be implemented and should be reimplemented as noted in #600. 

But at least, this change will make all test passes for me while I am in Thailand.

Issue #307, opened 2015-03-01
Bug in rendering stack area — probably caused by empty groups


kanitw:
`mark=area|x=Effect__Amount_of_damage,O|y=avg_Cost__Total_$,Q|color=When__Phase_of_flight,O`

![shelfviz](https://cloud.githubusercontent.com/assets/111269/6428991/ef916b7e-bf6c-11e4-82a0-04658fbe3163.png)

`mark=area|x=Effect__Amount_of_damage,O|y=sum_Cost__Total_$,Q|color=When__Phase_of_flight,O`

![shelfviz](https://cloud.githubusercontent.com/assets/111269/6428994/2fced104-bf6d-11e4-94a5-671e2b2609e9.png)

The problem is probably values in B only has value for take-off-run.  
Maybe we need to fix at Vega level — aggregate transform should have option to fill zero values for empty group?

kanitw:
This should be handled by vega!

kanitw:
![slack](https://cloud.githubusercontent.com/assets/111269/10566855/12018e14-75a7-11e5-8fa5-fe62a05ea4f8.png)

Our stack area example is suffering from this example, I’m moving this to 0.8.x.

kanitw:
This will require Vega to add `impute` transform to add missing values.

kanitw:
Fixed in #696

PR #595, opened 2015-10-03
[WIP-0.8] Vega 2 migration

domoritz:
- update vega dependency in editor and gallery to Vega 2

### Syntax Change
- axis
  - only include `axis.orient` when necessary
- legends
  - set legend's orient to right by default (similar to Vega)
  - Vega-lite now supports multiple legends 
- filter 
  - move `filter` to `data.filter` and takes Vega expression as parameter instead of a complicated predicate object.
  - Null filter is added before binning and timeUnit conversion
- scale
  - `scale.reverse` is removed as it is redundant with `sort`. 
- stack
  - add stack control to `color` and `detail` encoding
- sort — now `sort` property can be either `undefined`, a __String__ or an __Object__
  - `undefined` - the field is unsorted.
  - (as __String__) `'ascending'` or `'descending'` – the field is sort by the field's value in ascending or descending order.
  - (as __Object__ — for ordinal/nominal fields only) A sort field object - for sorting the field by an aggregate calculation (`sort.op`) over a specified sort field (`sort.field`) in a specified order (`sort.order`)

### Vega 2 Migration Refactor

- ValueRef
  - replace `d.data` with `datum` in filter
  - add `datum.` option to fieldRef and use it in time transform and template
  - eliminate outdated `fieldRef` options
  - Replace the eliminated `group` mark properties with `field.group`
  - Refer to `{group: 'mark.group.width’}` with 

- Transforms
  - Aggregation, Bin
  - Stack
    - Use vega2 stack's parameter and rename it to `<field_name>_start|end`, remove the need to add facet transform before stacking
    - Make facet comes after stack for stacked area chart  (This makes stacked area chart looks almost right but still looks weird due to #307 – see [example screenshot] (https://www.evernote.com/l/AAi8LLKSH3lDgZYIOTW39BQDzY4As6Vy2hk))
  - Facet
    - use field name reference instead of using `keys.0` (#639)

- Aggregation (#621)
  - `avg` => `mean`
  - add all supported aggregation in Vega 2
- Sorting (#612)
  - remove `sort.js` and no longer use `sort-<field_name>` tables as data source
  - modify `scale.domain()` to include DataRef’s `.sort` object
  - modify `sort` property to be consistent with Vega 2 (but still leave it as a top-level vega-lite property)
    - `sort.name` => `sort.field`
    - `sort.aggregate` => `sort.op`

### API Change
- `vl.field` => `vl.encDef`, field.js => encDef.js

#### Variable Name
- `fieldName` => `field`
- `field` => `encDef`

#### Other 
- removed `_vega2` flag 
- extract filter non-positive value when log are presented from normal filter
- eliminate `encoding.fieldName()` and just use `encoding.field(...).name`

- only use `encDef.name` where applicable rather than calling `fieldRef`
- local variable refactor in `scale.js`
- move stack’s stacked data transform to data.stack
- add encoding.stack so we do not have to pass stack around
- remove Encoding.toggleFilterNullO, which should just be a part of vega-lite-ui

### Examples
- added Aggregate Bar Chart, Stacked bar chart, Trellis Stacked Bar Chart
- both vl input and vg output are validated in the gallery and editor


src/compiler/axis.js
```diff
@@ -101,9 +101,9 @@ axis.grid = function(def, name, encoding, layout) {
           value: def.offset
         },
         x2: {
+          field: {group: 'mark.group.width'},
```
arvind: Is this referencing a grandparent's width? I'd recommend `field: {group: 'width', level: 2}` in this case, rather than manually walking up the scenegraph structure. See [value refs](https://github.com/vega/vega/wiki/Marks#value-references) documentation for more detail.

kanitw:
Great.  I was thinking about always create a dev version branch and having a PR that we append all changelog before merging to master like this.    (I’ve just appended the merged logs above and add 0.8 prefix to the title)

So when we release, we just copy the description above to the release note. :)

kanitw:
@domoritz   Thanks for merging.   I just appended changes to `0.8` branch.

@arvind  thanks

arvind:
:tada:

kanitw:
@arvind  — not yet but pretty close!

src/compiler/axis.js
kanitw: @arvind  Thanks for your comment.  

Note that this is under `axis.properties.grid` and I try to reference the `group`’s width here.  

From testing both `field: {group: ‘mark.group.width’}` and `field: {group: ‘width’, level: 2}` works correctly but I kinda feel that the former reads better given that there is no documentation for how to refer to group’s width in axis’s properties.   At least the former give better clue that it is using parent mark’s group’s width.  

A few comments:
- People will have no clue what going up 2 levels in `axis.properties.grid` refers to the parent's `mark`.  If you think this is a preferred method, it should be documented in `axis.properties`’s docs and I'd be happy to change the output to use `field: {group: ‘width’, level: 2}`.  
- The `level` property looks confusing with `group`.  When using `level` with `parent` like shown in the example  `"field": {"parent": "f", "level": 2}`, it makes sense.  But `field: {group: ‘width’, level: 2}` looks confusing what `level` means.  That said, it might be too verbose and too late to name it `parentLevel`.

kanitw:
All issues in the milestone are fixed!  Let’s release `0.8`.

arvind:
:tada:

domoritz:
:balloon:

PR #599, opened 2015-10-05
Extract `data.raw.transform.nullFilter` from `data.raw.transform.filter`

kanitw:
Extract `data.raw.transform.nullFilter` from `data.raw.transform.filter` so that we can perform `nullFilter` before binning and timeUnit calculation but perform general filter afterward.

We can further optimize by only apply filter based on bin/time unit afterward so that we perform less transforms, but I do not think it is that important — at least for now.  

**Please merge #591 first!**

### DIFF
--- a/editor/bower.json
+++ b/editor/bower.json
@@ -15,6 +15,6 @@
     "tests"
   ],
   "dependencies": {
-    "vega": "^1.5.4"
+    "vega": "^2.2.6"
   }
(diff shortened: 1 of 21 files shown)


kanitw:
@domoritz   should be ready to be merged!

kanitw:
This should be merged to 0.8 instead.

PR #601, opened 2015-10-05
Move filter to data.filter and use Vega expression

### LINKED ISSUE #588: Revise Filter: Move filter to data.filter and use Vega expression
- [ ] move `filter` to `data.filter`
- [ ] use Vega expression instead of a custom predicate object

kanitw:
Fixes #588  Move filter to data.filter and use Vega expression.

Now filter simply accepts a string, which should be a Vega expression.  

Please merge #599 first !  (See [Branch Specific Diff](https://github.com/uwdata/vega-lite/compare/kw/filter-null...kw/588-data.filter))

## TODO after merge
- [ ] update docs

### DIFF
--- a/src/Encoding.js
+++ b/src/Encoding.js
@@ -100,25 +100,7 @@ module.exports = (function() {
   };
 
   proto.filter = function() {
-    var filterNull = [],
-      fields = this.fields(),
-      self = this;
-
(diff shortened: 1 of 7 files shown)


kanitw:
This actually should be merged to 0.8 as it uses Vega expression — will create a new PR.

PR #602, opened 2015-10-05
[WIP] Extract `data.raw.transform.nullFilter` from `data.raw.transform.filter`

### LINKED ISSUE #588: Revise Filter: Move filter to data.filter and use Vega expression

kanitw:
Extract `data.raw.transform.nullFilter` from `data.raw.transform.filter` so that we can perform `nullFilter` before binning and timeUnit calculation but perform general filter afterward.

We can further optimize by only apply filter based on bin/time unit afterward so that we perform less transforms, but I do not think it is that important — at least for now.


kanitw:
Looks like this is still incorrect.

kanitw:
Looks like it is correct for inline data but not for url. 
Not sure if I overlook anything — or if it is a Vega’s bug.  

For example, 

```
{
  "data": {
    "values": [
      {"x": "A", "y": 28},
      {"x": "B", "y": null},
      {"x": "C", "y": 43},
      {"x": "D", "y": 91},
      {"x": "E", "y": null},
      {"x": "F", "y": 53},
      {"x": "G", "y": 19},
      {"x": "H", "y": 87},
      {"x": "I", "y": 52}
    ]
  },
  "marktype": "bar",
  "encoding": {
    "y": {
      "type": "Q",
      "name": "y"
    },
    "x": {
      "type": "O",
      "name": "x"
    }
  }
}
``` 

renders correctly.  

However, when loading data from url like the following spec, null values are not filtered correctly.
```
{
  "marktype": "point",
  "encoding": {
    "x": {"bin": true,"name": "Displacement","type": "Q"},
    "y": {"bin": true,"name": "Miles_per_Gallon","type": "Q"},
    "size": {
      "name": "*",
      "aggregate": "count",
      "type": "Q",
    }
  },
}
```

(inspect `vled.vis._model._data.raw._data.filter(function(d) { return !d.Miles_per_Gallon; })` in the console returns 8 objects that has `d.Miles_per_Gallon = null`)

arvind:
Assuming `vled.vis` refers to a Vega `View` instance, the access path used above refers to _input_ data (i.e., data before transforms are run) rather than _output_ data (i.e., after transforms are run). Whenever possible, I would recommend using [the documented API](https://github.com/vega/vega/wiki/Streaming-Data) to inspect data values. In this case, `vled.vis.data('raw').values()`.

kanitw:
@arvind  thanks.

kanitw:
@arvind  so if the access path above refers to input data -- why does it contains variable with `bin` prefix?  -- is that because the transform cause side effect to the original data object?

arvind:
That's right. To reduce Vega's memory footprint, we modify/augment/reuse existing tuples wherever possible, rather than produce wholly new ones like a true functional programming system.

domoritz:
I got your examples working (no errors, size is not used correctly yet) in https://github.com/vega/vega-lite/pull/608 but there is a lot of potential for refactoring.

kanitw:
@domoritz  I tested your #608 and still don’t see it correctly bin stuff.  See #608 for more info.

kanitw:
@arvind  I have another question here. 

What’s the easiest way to inspect how data goes through a particular transform / inspect a transform’s I/O?  

I’m trying to see why `bin` outputs all `NaN` for me.  I’m guessing that the min/max calculation returns `NaN` and thus makes all the binned values all becomes NaN.  However, I can’t find how to correctly adding breakpoint yet.  I tried to add breakpoint in `Bin()`’s `prototype.transform` but did not succeed having the breakpoint triggered.  


I think answer to this question and the following answer from previous question should be included in the Vega tutorial or maybe a separate “How to debug” page.  

> Whenever possible, I would recommend using the documented API to inspect data values.

arvind:
That's odd! Adding breakpoints to `prototype.transform` (or `prototype.evaluate` for other `Node` subclasses) is exactly how I've been debugging. Breakpoints trigger correctly for me with http://vega.github.io/vega-editor/?spec=crossfilter and allow me to inspect the parameters, etc. I wonder if something is off with your console workspace setup (e.g., vega/vega#376)?

domoritz:
I'll go ahead and merge this since all coming PRs depend on this and I don't want to do redundant work when I fix something on this branch.

kanitw:
@arvind  That’s still good to know so at least I eliminate a few cases :)

PR #604, opened 2015-10-05
Aggregate, Facet, Subfacet, and Bin for Vega 2

kanitw:
- Migrate aggregate and bin transform to Vega 2.
- Add Aggregate Bar Chart to the gallery

### DIFF
--- a/gallery/gallery.js
+++ b/gallery/gallery.js
@@ -54,6 +54,16 @@ var EXAMPLES = [
       data: {'url': 'data/cars.json'}
     }
   },{
+    title: 'Aggregate Bar Chart',
+    spec: {
+      "marktype": "bar",
+      "encoding": {
+        "x": {"name": "Cylinders","type": "O"},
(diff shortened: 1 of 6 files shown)


kanitw:
Note aggregation is now working — for example the following spec works 

```
{
  "marktype": "point",
  "encoding": {
    "x": {"name": "Cylinders","type": "O"},
    "y": {"name": "Acceleration","type": "Q","aggregate": "mean"}
  },
}
```

kanitw:
Looks like binning is still broken because `bin` transform now requires `min` and `max` parameters while these were optional in Vega 1.

kanitw:
@domoritz  just got `bin` working by using stats.  
Stack is still incorrect — but should be addressed in a separate PR.
Feel free to merge this for now.

kanitw:
Thanks.  I just modified the changelog in #595 to reflect this merge .

PR #608, opened 2015-10-06
Fix facets and aggregates

domoritz:


kanitw:
This is redundant with #604 ...

kanitw:
(The aggregate part)

kanitw:
I tested this branch and still do not get the correct binning (See below), which is not surprising since you haven’t changed any of the binning logic. 

vega-lite_editor

That said, your changes in facet, stack, and subfacet seems useful so I’m merging them into #604 and will further test/modify/refactor it.

Issue #276, opened 2015-02-25
refactor: Reduce side effects and ambiguity in the compile flow


kanitw:
- [x] split marks into multiple files
- [x] rename `src/compiler` => `src/compile`,  `compiler.ts` => `compile.ts`  and update `CONTRIBUTING.md`

# 1.0

- [x] remove constants in schema and place them where they belongs
  - [x] vl.schema.aggregate
  - [x] `supportedMarkTypes` in encoding schema 
  - [x] supportedTypes
  - [x] supportedEnums 

- [x] check if these are still correct with multiple details
  - [x] `model.isMeasure` / `model.isDimension`  -- if there is a way to remove these two methods, remove them

# as of 0.9

- [x] remove isTypes()
- [x] remove model.facets()
- [x] `encoding.fields`
- [x] rename `encoding.fieldRef` back to `encoding.field` 

# Older Stuff 

- [x] throwing `spec` around

#### Encoding

- [x] eliminate `encoding.field` and use `encoding.fieldRef` instead. 

- [x] simplify `encoding.data`

#### Data 

- [x] eliminate `getDataset()` from `sort.js`
- [x] decouple data generation from `template.js` and add `compile/data.js`
- [x] group `bin,filter,aggregate` in `compile/data`
- [x] #432 Merge “raw” and “table” table in vega where feasible #432

#### Template

- [x] Split template into template and data

#### Marks
- [x] decouple marks from `template.js`
- [x] anywhere with  `marks[0]`
- [x] `mdef`

#### names(), defs(), def()
- [x] Eliminate `axis.names()`,  `scale.names()` and `singleScaleNames`

#### Scales
- [x] revise `scale.domain` to see if all conditions are correct
- [x] split scale.range

#### Misc 
- [x] rename `time` to `timeUnit`
- [x] vl.format?

#### Stack
- [x] reduce side effect
- [x] consider caching stack as encoding’s property to avoid redundant computation
- [x] remove `stack.js` and move the same logic to `marks.js` to reduce side effect
- [x] consider if there is a way to avoid creating stacked data, which perform redundant calculation with scales in each facet.  (maybe signal could work)

#### Subfacet (Line & Area)
- [x] eliminate

#### Facet
- [x] simplify

kanitw:
@domoritz   

We need to establish a  variable naming convention for what’s input (vega-lite spec/definition) and what’s output (vega spec/definition). 

[...]

PR #626, opened 2015-10-08
Simplify stack code

kanitw:
#276

- move stack’s stacked data transform to `data.stack`
- add `encoding.stack` so we do not have to pass `stack` around

- TODO in future PRs 
  - consider caching stack as encoding’s property to avoid redundant computation
  - remove `stack.js` and move the same logic to `marks.js` to reduce side effect
  - consider if there is a way to avoid creating stacked data, which perform redundant calculation with scales in each facet.  (maybe signal could work)

### DIFF
--- a/src/Encoding.js
+++ b/src/Encoding.js
@@ -260,10 +260,37 @@ module.exports = (function() {
       spec.encoding.color;
   };
 
-  proto.isStack = function() {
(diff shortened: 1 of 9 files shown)


domoritz:
Tests are broken.

domoritz:
Updated 0.8 pr change log.

Issue #613, opened 2015-10-07
Validate Vega-lite output against Vega’s Json schema


kanitw:
(forked from #459)

Validating vega-lite output against Vega’s Json schema will help us catch our migration bugs.  That said, validating this everytime vega-lite is used might be an overkill — so we should find a middle ground.  

Maybe enable that just in `0.8` branch and then we remove it before we merge `0.8` to `master`?

jheer:
Sounds like this should be part of the test suite, not normal Vega-Lite use?

domoritz:
I agree with Jeff. Let's make it part of the tests and maybe the editor.

kanitw:
That makes sense to me.  Maybe we have to refactor gallery.js and verify that all example specs produced legitimate vega specs.

domoritz:
Not fixed in da4198b

kanitw:
Partially addressed in #628 -- we should add a unit test that loads all the examples and verify if it produces sensible vega spec too.  Therefore, whenever we break the output a test is failed -- rather than having to rely on obscure `console.error` in the gallery.

kanitw:
Done in #641

PR #628, opened 2015-10-08
Dom/validate vg schema

domoritz:


domoritz:
Fixes #613

editor/validate.js
```diff
@@ -0,0 +1,36 @@
...
+
+  if (!valid) {
+    console.error(validator.getLastErrors());
+  }
+};
+
+var validateVg = function(vgspec) {
+  var validator = new ZSchema();
+
+  var cb = function() {
```
kanitw: `callback`?

kanitw:
This is a step toward #613, but I think we add a unit test that loads all the examples and verify if it produces sensible vega spec too.  Therefore, whenever we break the output a test is failed -- rather than having to rely on obscure `console.error` in the gallery.

kanitw:
I'll merge this for now as it's still useful.

PR #629, opened 2015-10-08
Dom/axis refactor

domoritz:
- only add axis orientation if we have to

contains #628


kanitw:
merged & appended comment to 0.8

Issue #451, opened 2015-05-25
Calculated Field / Formula Transform


domoritz:
The syntax for expressions is odd at the moment. There are multiple options to refactor

currently:

```js
{
name: ‘Release_Date’
type: ’T’,
fn: ‘year’
}
```

rename fn to `function` or `as`

```js
{
name: ‘Release_Date’
type: ’T’,
as: ‘year’
}
```

rename to `map`, `accessor` and allow more functions.

```js
{
name: ‘Release_Date’
type: ’T’,
map: ‘year’
}
```

or change how fields are mapped to encodings entirely. This also allows other expressions such as rations and such. 

```js
{
value: ‘year(Release_Date)’
type: ’T’
}
```

See discussion in #447

kanitw:
Oh now I get what you mean.   So when we use `value:`, we eliminate `name:` for that field, right?   
If yes, I think that's an interesting idea.   A couple of interesting decision to make.

- do we only always use `value:` then?
- do we want to come up with Vega-like expression parser to eliminate `d.` in the value?

domoritz:
Yes, we would get rid of `name`. The issue with `d.` would not be a problem in vega 2 but not vega 1, which is a bit annoying. I was thinking only about the ideal API and not so much the implementation at this point. 

As an intermediate step, we could say that we only support simple function calls and maybe ratios and write a simple regex parser for it. However, I'm not really sure about all the implications of my suggestion yet.

kanitw:
It's worth noting that `T` (temporal/`Date`) and `Q` (quant/`number`) should support different set of operations. 

For T, most of the time, you don't want to support complex calculation.  (even `+` doesn't make much sense) You just want to abstract the time value.   And we need to know function name to predict cardinality, etc.   (This might change a lot with vega2)

For Q, it's more free form and most of the time, we don't care much about cardinality (unless you cast it to be O).


But yes, from the user perspective, the distinction might make things more complicated.

kanitw:
Another thought. 

Is it kinda weird that `aggregation` are not included in the expression `expr`?

For example,
```js
{'aggregation':'min', value:'abs(foo)'}
```

domoritz:
I also see how this would make it harder for polestar because you cannot have a dropdown for all available functions.

domoritz:
> Is it kinda weird that aggregation are not included in the expression expr?

[...]

PR #631, opened 2015-10-09
Use vega-datasets

domoritz:
* use vega-dataset
* add formula transform


domoritz:
Need to fix date parsing for years.

domoritz:
Works

```json
{
  "marktype": "line",
  "encoding": {
    "x": {"name": "Year","type": "T","timeUnit": "year"},
    "y": {"name": "Miles_per_Gallon","type": "Q","aggregate": "median"}
  }
}
```

Doesn't work
```json
{
  "marktype": "line",
  "encoding": {
    "x": {"name": "Year","type": "T","timeUnit": "month"},
    "y": {"name": "Miles_per_Gallon","type": "Q","aggregate": "median"}
  }
}
```

domoritz:
I decided to add formula transforms (https://github.com/vega/vega-lite/issues/451). So this should work

```json
{
  "data": {
    "formula": {
      "field": "year",
      "expr": "new Date(datum)"
    }
  },
  "marktype": "line",
  "encoding": {
    "x": {"name": "Year","type": "T","timeUnit": "month"},
    "y": {"name": "Miles_per_Gallon","type": "Q","aggregate": "median"}
  },
}
```

jheer:
Somewhat orthogonal to the general issue of using formulas, but relevant here: if we specify the type as `"T"` shouldn't we be able to appropriately "cast" behind the scenes? Relying on the expression language to get date/time objects may not play well across different implementations of vega-lite parsers/renderers. Expression language support should be an "add-on", not required.

domoritz:
Goo points. My take on this is that I want to get the examples working first. Automatically parsing dates for example seems to be a can of worms that I don't want to open yet. In js vl we could automatically add a sensible formula transform if the user specifies the type `T` but I don't think that there is a great default parsing that I would recommend for other programming languages (e.g. Python).

domoritz:
I just noticed that we automatically add a formula transform for time already. See https://github.com/vega/vega-lite/blob/master/src/compiler/time.js#L35

domoritz:
The problem is that we add a utcmonth transform for the example from above. 

```
{
          "type": "formula",
          "field": "month_Year",
          "expr": "utcmonth(datum.Year)"
        }
```

arvind:
I have a question after seeing the above addition of formula transforms to vega-lite. What is the guiding principle for where data transformation operators live in a vega-lite specification? Looking through the schema, I count:

1. `formulas` that appear in the `data` property. 
2. `filter` is a top-level property. 
3. `aggregate`, `bin`, and `sort` are specified within the visual encoding channels. 

I ask because this has implications for how the specification of interaction techniques can augment/override data transformation properties.

domoritz:
We moved `filter` into `data`. `formulas` will also be in `data`. 

`aggregate`, `bin`, and `sort` are specifies in the encDef (visual encoding channels) because they make less sense in the global context. I would like to have formula in the encDef as well but @jheer and @kanitw decided against it as mentioned in https://github.com/vega/vega-lite/issues/451.

arvind:
Cool, that dichotomy makes sense to me. Formulas within the visual encoding channels would dramatically restrict the expressivity of interactions.

domoritz:
I think I understand the reasoning now. One of the goals of vl is that it can be generated automatically and that it can be supported by other languages (e.g. python). If we have expressions in fields in the encDef, this would be much harder. If we keep it in the data, we can say that we only support formula transforms in js.

domoritz:
Okay, there seems to be some funny business going on when I choose month that are not related to the changes in this pr (see blow). This is ready to be merged. 

@kanitw Let me know if you want a separate pr for 2fe16cc. 

screen shot 2015-10-12 at 17 09 20

kanitw:
Why do we need 7zip.png?

kanitw:
Oh i see, it's a part of vega-datasets.  Sorry!

kanitw:
@domoritz  Actually, it would be great if we can separate 2fe16cc if not too difficult.  

We can first fix all examples in the gallery we can publish a working `0.8.1` release (and ignore the month example) for now -- while adding formula transform (with a solution for its stats) to future `0.8.x` branches.  

We can also discuss more about roadmap in person tomorrow.

domoritz:
I think the formula transform is simple enough that we can release it with 0.8.1.

domoritz:
@kanitw resolved conflicts

PR #638, opened 2015-10-13
Fix remaining basic migration issues

### LINKED ISSUE #459: Vega 2 migration basics
- [x] use the new `name` mark property in groups

### Field 

- [x] remove `.data` direct access — use `fieldRef` instead
- [x] remove `.data` from `field` 
- [x] `index` => `_id` in `field`

- [x] replaced mark’s `group` property with `field.group` #593

### Scale

- [x] use scale’s sort for sorting (Done in #612)

### Transform

- [x] stats => aggregate (Done in #604)

- Facet
  - [x] `keys` => `groupby` (Done in #604)
  - [x] Eliminate `keys.0` reference

- [x] Stack #412

- [x] Check if we use any of removed transforms: array, copy, flatten, slice, truncate, unique, window (__We did not__)

- [x] Table Heatmap

- [ ] ~~Valid test cases against Vega 2 JSON Schema~~ (forked to #613)

kanitw:
Fixes #459 

- eliminate `keys.0` reference
- add `datum.` in template reference

After merging this to `0.8`, we should test this with voyager a bit and consider releasing `0.8.1`.


kanitw:
wrong merge target

PR #642, opened 2015-10-13
From master

domoritz:
Had to redo some commits that I had to undo earlier.


domoritz:
This should allow us to merge 0.8 into master.

kanitw:
@domoritz  -- should we finish fixing validate errors first?

PR #647, opened 2015-10-13
Make `stack` either boolean or object

### LINKED ISSUE #634: Use stack’s `sortby` property to order values in the stack
forked from #39

kanitw:
Make `stack` either boolean or object and accept `reverse` and `offset` properties.  -- fixes #634


kanitw:
@domoritz okay — this is a working version of #644 provided that we defer vega schema reference for later.

PR #657, opened 2015-10-15
add explicit .npmignore because we want to include vega-lite

domoritz:


kanitw:
> add explicit .npmignore because we want to include vega-lite

you mean include __only__ `vega-lite*` ?

domoritz:
To make this more clear. npm uses .gitignore if no .npmignore is present.

.gitignore
```diff
@@ -7,4 +7,3 @@ vega-lite*.map
 spec.json
 npm-debug.log
 instance.json
-coverage
```
kanitw: Why not exclude coverage?

.npmignore
```diff
@@ -0,0 +1,8 @@
+test/log
+test/log/difflist.json
+node_modules
+bower_components
+data
+editor
```
kanitw: I guess we’ll have to remove “editor” once we migrate.

.gitignore
domoritz: I never have coverage locally. Reverted.

.npmignore
domoritz: I don't think it hurts, though.

kanitw: Sure – just a note for cleanliness.

.gitignore
kanitw: lol

PR #670, opened 2015-10-18
remove Encoding.toggleFilterNullO

kanitw:
Remove `Encoding.toggleFilterNullO`, which should just be a part of `vega-lite-ui`

Address part of #598


domoritz:
Is this ready? Since we are missing the implementation in vlui.

kanitw:
ok ready now https://github.com/vega/vega-lite-ui/pull/127

PR #671, opened 2015-10-18
Make `sort` support `undefined`, String (for all types), or Object (for ordinal/nominal only) 

### LINKED ISSUE #663: Should we add `order` property to `encDef.sort`?
Currently Vega-lite’s `encDef.sort` looks exactly like vega Scale's `DataRef.sort`, which only contains `aggr` and `op` properties while the `order` field is deprecated as same mechanism can be achieved by using `reverse` property of the scale.  

However, in vega-lite’s encoding definition, `sort` is another top-level object, not a part of `scale`.  If we do not have `order` property.  Specifying descending sort would require specifying both `sort` and `scale` like the following Barley example, which is kind of miss the point for extracting `sort` from `scale`. 

```js
{
      data: {url: 'data/barley.json'},
      marktype: 'point',
      encoding: {
        x: {type: 'Q',name: 'yield'},
        y: {
          name: 'variety',
          type: 'O',
          sort: {field: 'yield', op: 'median'},
          __scale: {reverse: true}__
        },
        row: {
          type: 'O',
          name: 'site',
          sort: {field: 'yield', op: 'median'},
          __scale: {reverse: true}__
        },
        color: {type: 'N',name: 'year'}
      }
    }
``` 

By having `order` property, the spec can be a bit nicer:

```
{
  data: {url: 'data/barley.json'},
  marktype: 'point',
  encoding: {
    x: {type: 'Q',name: 'yield'},
    y: {
      name: 'variety',
      type: 'O',
      sort: {field: 'yield', op: ‘median’, __order:’desc'__}
    },
    row: {
      type: 'O',
      name: 'site',
      sort: {field: 'yield', op: 'median', __order:’desc’__},
    },
    color: {type: 'N',name: 'year'}
  }
}
```

kanitw:
Fixes #663 

## Change List

- now `sort` property can be either `undefined`, a __String__ or an __Object__
  - `undefined` - the field is unsorted.
  - (as __String__) `'ascending'` or `'descending'` – the field is sort by the field's value in ascending or descending order.
  - (as __Object__ — for ordinal/nominal fields only) A sort field object - for sorting the field by an aggregate calculation (`sort.op`) over a specified sort field (`sort.field`) in a specified order (`sort.order`)

## Internal 

- add `scale.sort` that correct determine `sort` output
- add `scale.reverse` and determine the value from `sort`
- update docs regarding `sort`

### DIFF
--- a/src/schema/schema.js
+++ b/src/schema/schema.js
@@ -231,9 +231,15 @@ var sortMixin = {
   type: 'object',
   properties: {
     sort: {
-      type: ['object', 'boolean'],
-      default: true,
+      default: undefined,
       supportedTypes: toMap([N, O]),
+      oneOf: [
+        {
+          type: 'string',
+          enum: ['ascending', 'descending']
+        },
+        { // sort by aggregation of another field
+          type: 'object',
       required: ['field', 'op'],
       properties: {
         field: {
@@ -244,8 +250,15 @@ var sortMixin = {
           type: 'string',
           enum: VALID_AGG_OPS,
           description: 'The field name to aggregate over.'
+            },
+            order: {
+              type: 'string',
+              enum: ['ascending', 'descending']
         }
       }
+    }
+      ]
+
     }
   }
 };
--- a/src/schema/schema.js
+++ b/src/schema/schema.js
@@ -99,11 +99,6 @@ var typicalField = merge(clone(schema.field), {
       properties: {
         /* Common Scale Properties */
         type: schema.scale_type,
-        reverse: {
-          type: 'boolean',
-          default: false,
-          supportedTypes: toMap([Q, T])
-        },
 
         /* Quantitative Scale Properties */
         nice: {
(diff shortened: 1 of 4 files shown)


src/schema/schema.js
```diff
@@ -231,21 +226,34 @@ var sortMixin = {
...
   properties: {
     sort: {
-      type: ['object', 'boolean'],
-      default: true,
+      default: undefined,
       supportedTypes: toMap([N, O]),
-      required: ['field', 'op'],
-      properties: {
-        field: {
+      oneOf: [
```
domoritz: How does the instantiation handle this?

kanitw: Good point.  I think our instantiation doesn't support `oneOf`?

kanitw: Actually if `default` value is provided — this is fine as we [simply return default value](https://github.com/vega/vega-lite/blob/master/src/schema/schemautil.js#L19).

PR #673, opened 2015-10-18
Migrate `Encoding.toggleSort` to `vega-lite-ui` and fix domain when sort by aggregation of a sort field.

kanitw:
- remove Encoding.toggleSort and migrate to vega-lite-ui
- only output vega `scale.reverse` when it is `true`
- Make sure that If sort by aggregation of a specified sort field,  use RAW table so we can aggregate value for the scale independently from the main aggregation. 

Please merge #670, #671, #672 first!

### DIFF
--- a/src/Encoding.js
+++ b/src/Encoding.js
@@ -354,38 +354,5 @@ module.exports = (function() {
     return spec;
   };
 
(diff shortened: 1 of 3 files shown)


kanitw:
@domoritz  ready for review.

PR #679, opened 2015-10-18
Make `grid` value automatically determined

### LINKED ISSUE #378: Have grid: “auto” ...
data_voyager

bar don’t need grid on that ordinal axis

let’s punt on this for now...

kanitw:
Fixes #378

If `grid` is unspecified, the default value is `true` for ROW and COL. For X
and Y, the default value is `true` for quantitative fields and `false`
otherwise.

### DIFF
--- a/docs/Encoding-Mapping.md
+++ b/docs/Encoding-Mapping.md
@@ -102,8 +102,8 @@ Scales of quantitative fields are linear scales by default, but users can specif
 
 ## axis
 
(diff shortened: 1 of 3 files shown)


kanitw:
Wait, want to test a few case …. will ping again when ready

kanitw:
Just made `grid=true` by default for time fields too —now ready for review.

PR #681, opened 2015-10-18
Rename `spec.json` -> `vega-lite-spec.json` for consistency with vega

kanitw:
Based on @arvind's suggestion


domoritz:
I literally implemented the same.

PR #682, opened 2015-10-18
Rename spec.son -> vega-lite-spec.json

domoritz:
Better version of https://github.com/vega/vega-lite/pull/681


domoritz:
Will rename to schema

PR #691, opened 2015-10-23
Remove Vega-Lite Editor

domoritz:


kanitw:
- I don’t think gallery should become index.html.  index.html should be the Vega-lite’s project page, which we can easily copy vega’s project page and have 15 nice examples there that links to editor.  
Meanwhile, our gallery is a testing gallery, not example.   

- I think we should copy the functionalities for (1) save code in cookie (2) browser sync to vega-editor, it’s quite annoying when forgetting to save script there. 

- instead of having `gallery/bower.json` can we move those dependency to `devDependencies` in the main `bower.json`.  Or do something so we do not have to do `cd gallery && bower install` every time we check out new branches.

kanitw:
No hurry on doing any of this though, we should do it when other more important things are done.

domoritz:
Agree with all you points but certainly not today.

domoritz:
Done. I haven't ported the cookie feature over to the editor yet. What are your thoughts on a landing page for vl?

kanitw:
@domoritz  We can initially copy Vega's landing page, and replace vega examples with vega-lite examples.  That could be done in another PR though.

kanitw:
LGTM except one thing:

Why do we need to validate in the gallery?  We already added validation to the unit test.

domoritz:
Done

PR #693, opened 2015-10-23
Reimplement bin using `bin_start`, `bin_end`

kanitw:
- Reimplement binning to use linear scale instead of ordinal scale based on @jheer’s example in https://github.com/vega/vega/pull/388 using - use the new bin transform’s `bin_end`

- Make `scale.zero` automatically determined rather than always `true`
  - always returns `false` for bin and time
  - Otherwise, return true for `X`, `Y` and `SIZE`

## TODO
- use `bin_mid` instead

### DIFF
--- a/gallery/examples.js
+++ b/gallery/examples.js
@@ -1,4 +1,5 @@
 'use strict';
+// jshint quotmark: false
 
 var EXAMPLES = [
   {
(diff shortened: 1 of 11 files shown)


kanitw:
Okay.  Things looks good now for histograms (and when replacing `bar` with `point` and `line`). 
The `area` one still looks odd, but I am not sure how to make it looks better so I defer optimizing using bin with area for #694 .  

vega-lite_gallery

gallery/examples.js
```diff
@@ -252,6 +253,81 @@ var EXAMPLES = [
       },
       'data': {'url': 'data/cars.json'}
     }
+  },{
```
domoritz: I'll put these in separate files some time later.

kanitw: I think refering for exemplar ones in vega-editor make sense.  For ones that are more visual test (like histogram with point, line -- we should make sure that it is easy to create ones)

Having it here, reduce the need to name the file nicely (and burden of renaming if the original name sucks, etc.)  That said, this examples.js can become too damn long so I'm not against splitting it.  Just wanna express potential downside of that here.

domoritz: I have similar concerns. We will see after next week.

Issue #181, opened 2015-02-03
Support all applicable Vega mark, scale, legend, and axis properties


kanitw:
## Axis

- [x] `axis.tick`
- [x] `title`
- [x] `title` ’s maxlength and offset
- [x]  `labelAngle|Align/Baseline`

### Common Scale Properties
- [x] round
- [x] domain, ~~domainMin, domainMax~~
- [x] range, ~~rangeMin, rangeMax~~
- [] ~~reverse (for ordinal — time/quantitative are already supported)~~  _we have `order` instead. 

### Ordinal Scale Properties
- [x] `domain.sort`

### Time/Quantitative Scale Properties
- [x] clamp

### Quantitative Scale Properties
- [x] exponent

## [Legends](https://github.com/vega/vega/wiki/Legends)**
( has to modify `legendMixin.legend` to be an object that contains properties )

- [x] title
- [x] format
- [x] orient
- [x] properties (need to be merged)

(`size, shape, fill, stroke, values` shouldn’t be overrides.)

## Marks

- [ ] List!

kanitw:
Done for everything besides marks as of #729, which could be done later - so forking #730 just for marks.

PR #701, opened 2015-10-31
Major scale revision

kanitw:
# Syntax
- Add all of vega's scale properties that should be supported #181 
- add scale property to `row`, `column` and `shape`
- remove `encDef.band`
  - `band.size` => `scale.bandWidth`
  - `band.padding` => `scale.padding`

# Internal
- add `encoding.padding()` helper
- Add separate methods for 'bandWidth', 'nice', 'padding', 'points', 'range', 'reverse', 'round', ‘zero’
- revise default valus for `nice`, `points` and `zero`
- revise `useRawDomain` to support only aggregate functions that produces values ranging in the domain of the source data (`'mean'`, `'average'`, `'stdev'`, `'stdevp'`, `'median'`, `'q1'`, `'q3'`, `'min'`, `'max'`)
- rename
  - `bandSize` => `bandWidth`
  - `scale.sort()` => `scale.domain.sort()`

### DIFF
--- a/src/schema/schema.js
+++ b/src/schema/schema.js
@@ -171,7 +171,7 @@ var typicalScaleMixin = {
     useRawDomain: {
       type: 'boolean',
       default: undefined,
-      description: 'Use the raw data range as scale domain instead of ' +
+      description: 'Uses the raw data range as scale domain instead of ' +
                    'aggregated data for aggregate axis. ' +
                    'This option does not work with sum or count aggregate' +
                    'as they might have a substantially larger scale range.' +
--- a/src/schema/schema.js
+++ b/src/schema/schema.js
@@ -84,17 +84,7 @@ var bin = {
   supportedTypes: toMap([Q]) // TODO: add O after finishing #81
 };
 
-var typicalField = merge(clone(schema.field), {
-  type: 'object',
-  properties: {
-    type: {
-      type: 'string',
-      enum: [N, O, Q, T]
-    },
-    aggregate: schema.aggregate,
-    timeUnit: schema.timeUnit,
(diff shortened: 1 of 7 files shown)


kanitw:
@domoritz  I got some major revision of scale done on the plane.  This is ready for review.

src/schema/schema.js
```diff
@@ -28,7 +28,7 @@ schema.aggregate = {
     Q: VALID_AGG_OPS,
     O: ['median','min','max'],
     N: [],
-    T: ['mean', 'median', 'min', 'max'],
+    T: ['mean', 'median', 'min', 'max'], // TODO: revise what should time support
```
domoritz: why?

src/schema/schema.js
```diff
@@ -84,6 +84,105 @@ var bin = {
...
+
+    /* Vega-lite only Properties */
+    useRawDomain: {
+      type: 'boolean',
+      default: undefined,
+      description: 'Use the raw data range as scale domain instead of ' +
+                   'aggregated data for aggregate axis. ' +
+                   'This option does not work with sum or count aggregate' +
+                   'as they might have a substantially larger scale range.' +
+                   'By default, use value from config.useRawDomain.'
```
domoritz: uses

src/schema/schema.js
kanitw: There are so many aggregate function that Vega now supports ... but not all of them will be supported by time.  
For example, `q1`, `q3` should be supported by time.

domoritz: Okay. Noted.

PR #708, opened 2015-11-07
WIP convert to typescript

domoritz:
## TODO before merging

- [x] get rid of all `.default`
- [ ] ~~add TODO for all `:any`~~ (we can enable tslint rule later to prevent any)
- [x] eliminate weird method map
- [x] namespace in  data.ts
- [ ] make output js  tabbed by 2
- [x] make mocha know about TS line no. 
- [x] make `countDisplayName` a constant 
- [x] rename module usages from `axis` to `vlAxis` to avoid name conflicts
- [x] get rid of all `exports.`
- [x] use an enum for X, Y, COL, ROW, ...
- [x] use an enum types
- [x] get correct son schema
- [ ] code coverage in ts?

## TODO if we merge
- [ ] Describe the schema in TS

- [ ] enable more tslint rules (and fix issues)
  - [ ] replace `var` with `let` and `const`

- [ ] fix isType in Compass (and vlui, ps, vy if applicable)


src/Encoding.ts
```diff
@@ -1,31 +1,31 @@
-'use strict';
```
kanitw: tslint?   

One annoying thing about tslint is that it prefers double quote by default ...

src/Encoding.ts
```diff
@@ -167,110 +166,85 @@ module.exports = (function() {
...
   };
 
-  proto.value = function(et) {
+  value(et) {
     return this._enc[et].value;
   };
 
-  proto.numberFormat = function(fieldStats) {
+  numberFormat(fieldStats) {
     var formatConfig = fieldStats.max > this.config('maxSmallNumber') ?
```
kanitw: outdated

src/Encoding.ts
kanitw: Can be done in another commit though

domoritz: What do you mean? Should I keep the 'use strict'?

src/Encoding.ts
domoritz: I just merged.

src/compiler/scale.ts
```diff
@@ -1,26 +1,26 @@
-'use strict';
-require('../globals');
-var util = require('../util'),
-  time = require('./time'),
-  colorbrewer = require('colorbrewer'),
-  interpolate = require('d3-color').interpolateHsl;
+/// <reference path="../colorbrewer.d.ts"/>
```
kanitw: Should put in README when do we use import, when do we use require and when <reference>

src/compiler/scale.ts
```diff
@@ -1,26 +1,26 @@
...
   return names.reduce(function(a, name) {
-    var scaleDef = {};
-
-    scaleDef.name = name;
-    var type = scaleDef.type = scale.type(name, encoding);
-    scaleDef.domain = scale.domain(encoding, name, type, facet);
+    var scaleDef = {
+      name: name,
+      type: type(name, encoding),
+      domain: domain(encoding, name, scaleDef.type, facet)
```
kanitw: this gonna lead to scaleDef is undefined (for `scaleDef.type`)

src/compiler/scale.ts
domoritz: If I knew... 

I will finish this and see what is the best way.

src/compiler/scale.ts
domoritz: What has changed?

src/compiler/scale.ts
```diff
@@ -30,7 +30,8 @@ scale.defs = function(names, encoding, layout, stats, facet) {
       ];
 
     properties.forEach(function(property) {
-      var value = scale[property](encoding, name, type, layout, stats);
+      // TODO: check if this works
+      var value = this[property](encoding, name, type, layout, stats);
```
kanitw: scale isn’t a class, so using `this` is extremely weird.

src/compiler/scale.ts
```diff
@@ -364,54 +365,55 @@ scale.color = function(encoding, name, scaleType, stats) {
...
 };
 
-scale.color.palette = function(range, cardinality, type) {
-  // FIXME(kanitw): Jul 29, 2015 - check range is string
-  switch (range) {
-    case 'category10k':
-      // tableau's category 10, ordered by perceptual kernel study results
-      // https://github.com/uwdata/perceptual-kernels
-      return ['#2ca02c', '#e377c2', '#7f7f7f', '#17becf', '#8c564b', '#d62728', '#bcbd22', '#9467bd', '#ff7f0e', '#1f77b4'];
+export var colors = {
```
kanitw: this looks weird to me too, this is a submodule?

src/compiler/scale.ts
kanitw: Why not using `module`?

src/compiler/stack.ts
```diff
@@ -26,9 +20,17 @@ function stacking(encoding, mdef, stack) {
     });
   }
 
+  interface StackTransform {
```
kanitw: Should we put Vega interfaces all in one place?

src/compiler/scale.ts
domoritz: Ha, that is what I was looking for. Thx.

src/Encoding.ts
kanitw: ‘use strict’ is a thing for jshint?  So that reminds me that we need a linter for ts instead.

src/Encoding.ts
```diff
@@ -1,31 +1,31 @@
-'use strict';
+import schema = require('./schema/schema')
```
kanitw: Why are we missing `require` with another type of import?

src/encdef.ts
```diff
@@ -126,21 +120,21 @@ function isDimension(field) {
...
 };
 
-vlEncDef.count = function() {
-  return {name:'*', aggregate: 'count', type: Q, displayName: vlEncDef.count.displayName};
+export function count() {
+  return {name:'*', aggregate: 'count', type: Q, displayName: countDisplayName};
 };
 
-vlEncDef.count.displayName = 'Number of Records';
+export var countDisplayName = 'Number of Records';
```
kanitw: Constant => use `const COUNT_DISPLAYNAME` ?

src/Encoding.ts
domoritz: Cause wip

src/compiler/scale.ts
kanitw: Isn’t the diff here obviously different?  
On the left I declare `scaleDef` first, so it’s always defined.  
On the right, scaleDef is undefined.  

You also remove my `var type = scaleDef.type`

src/Encoding.ts
domoritz: Added tslint. We can add more rules later. I already found a few bugs but don't want to refactor right now.

src/compiler/stack.ts
domoritz: Later. For now I mostly added `any` types and this is one of the few places where I didn't do it.

src/encdef.ts
domoritz: I am not going to refactor now. We can use lint rules later.

domoritz:
All tests passing. There are some refactorings/ annotations we can do but I'll leave those for later. Let's make a decision whether we want to follow this typescript path or not.

src/compiler/scale.ts
domoritz: Kind of. I'm sure there are better ways to do this but I don't want to spend time on it until we decide to go with it.

src/encdef.ts
```diff
@@ -77,15 +72,16 @@ vlEncDef.fromShorthand = function(shorthand) {
     }
   }
 
+  // TODO
```
domoritz: @kanitw Cannot compile with this code because we reference `schema.timefns`. One bug discovered by ts :-)

src/compiler/marks.ts
```diff
@@ -1,13 +1,9 @@
-'use strict';
-
-require('../globals');
-
-var marks = module.exports = {};
-
-marks.def = function(encoding, layout, style) {
+import {X, Y, TEXT, COLOR, SIZE, SHAPE} from '../consts';
```
kanitw: yeah, I like this part more than what we had.

src/consts.ts
```diff
@@ -0,0 +1,28 @@
+export const SUMMARY = 'summary';
+export const SOURCE = 'source';
+export const STACKED = 'stacked';
+export const INDEX = 'index';
+
+export const X = 'x';
```
kanitw: make enum ‘CHANNEL`?

src/consts.ts
```diff
@@ -0,0 +1,28 @@
...
+export const Y = 'y';
+export const ROW = 'row';
+export const COL = 'col';
+export const SIZE = 'size';
+export const SHAPE = 'shape';
+export const COLOR = 'color';
+export const TEXT = 'text';
+export const DETAIL = 'detail';
+
+export const N = 'N';
```
kanitw: make enum `TYPE`?

src/compiler/marks.ts
domoritz: Totally agree. We don't clutter the global namespace any more and the imports are explicit. When we enable more typescript lint rules it will also tell us about unused imports.

src/consts.ts
domoritz: Then we have to replace all `X` with `Channel.X`, right?

kanitw: Yep.  But that sounds like a good thing too.

domoritz: Definitely. Noted in TODO.

src/compiler/axis.ts
```diff
@@ -188,62 +183,64 @@ axis.title = function (def, encoding, name, layout) {
...
-/**
- * Determine number format or truncate if maxLabel length is presented.
- */
-axis.labels.format = function (def, encoding, name) {
-  if (encoding.axis(name).format) {
-    def.format = encoding.axis(name).format;
-  } else if (encoding.isType(name, Q)) {
-    def.format = encoding.numberFormat(name);
-  } else if (encoding.isType(name, T)) {
+export var labels = {
```
kanitw: I hate this part!   Still looks like a JS namespace hack …

src/consts.ts
domoritz: Noted

src/compiler/compiler.ts
```diff
@@ -1,40 +1,38 @@
...
-compiler.subfacet = require('./subfacet');
-compiler.time = require('./time');
-
-compiler.compile = function (spec, stats, theme) {
-  return compiler.compileEncoding(Encoding.fromSpec(spec, theme), stats);
+
+import {summary} from '../util';
+import {Encoding} from '../Encoding';
+
+import * as axis from './axis';
```
kanitw: Should they be `vlAxis`, `vlLegend`, …?  (Not sure)

src/compiler/data.ts
```diff
@@ -82,18 +86,18 @@ data.source.formatParse = function(encoding) {
...
-data.source.transform = function(encoding) {
+export function transform(encoding) {
   // null filter comes first so transforms are not performed on null values
   // time and bin should come before filter so we can filter by time and bin
-  return data.source.transform.nullFilter(encoding).concat(
-    data.source.transform.formula(encoding),
-    data.source.transform.time(encoding),
-    data.source.transform.bin(encoding),
-    data.source.transform.filter(encoding)
+  return nullFilterTransform(encoding).concat(
```
kanitw: `sourceTransformNullFilter` , ... are better?   (so it’s clear that it’s `source`’s)

src/compiler/compiler.ts
domoritz: The thing is that we often use local variables and call functions with the same name. We should rename the functions/modules as you suggest.

src/compiler/data.ts
domoritz: I think there should be a better way to do nesting. For example a module within a module.

src/compiler/marks.ts
```diff
@@ -37,13 +33,35 @@ marks.def = function(encoding, layout, style) {
   return defs;
 };
 
-marks.bar = {
+function switchMark(mark: string): any {
+  // TODO: use constants
```
kanitw: “enum"

src/compiler/marks.ts
```diff
@@ -37,13 +33,35 @@ marks.def = function(encoding, layout, style) {
...
+  switch (mark) {
+    case 'bar':
+      return bar;
```
kanitw: should this just return 

```
{
   type: 'rect'
   prop: bar_props,
   supportedEncoding: {row: 1, col: 1, x: 1, y: 1, size: 1, color: 1}		        
 };		 
```

rather than having 
```
export var bar = ...
```

below?

src/compiler/data.ts
kanitw: I think module might work.

src/compiler/legend.ts
```diff
@@ -1,44 +1,38 @@
...
-legend.def = function(name, encoding, def, style) {
+export function def(name, encoding, def, style) {
   var timeUnit = encoding.encDef(name).timeUnit;
 
-  def.title = legend.title(name, encoding);
+  def.title = title(name, encoding);
   def.orient = encoding.encDef(name).legend.orient;
 
-  def = legend.style(name, encoding, def, style);
+  def = exports.style(name, encoding, def, style);
```
kanitw: why do we need `exports.`  prefix here 
isn't `style(...)` sufficient?

kanitw: oh I know now.... conflicting with local variable... ><"""

domoritz: Because there is a parameter called style. Can fix later by renaming style to vlStyle.

kanitw: I don't think renaming to `vlStyle` will work here ... 

Otherwise, you get `legend.vlStyle` which is kinda weird.  

But assuming we remove `style` parameter, this should be fine

src/compiler/time.ts
```diff
@@ -86,11 +87,11 @@ time.range = function(timeUnit, encoding) {
  * @param  {Object} encoding
  * @return {Array}  scales for time unit names
  */
-time.scales = function(encoding) {
+export function scales(encoding) {
   var scales = encoding.reduce(function(scales, encDef) {
     var timeUnit = encDef.timeUnit;
     if (encDef.type === T && timeUnit && !scales[timeUnit]) {
-      var scale = time.scale.def(encDef.timeUnit, encoding);
+      var scale = exports.scale.def(encDef.timeUnit, encoding);
```
kanitw: we can rename local `scale` variable to `scaleDef` and use `exports.`

src/compiler/scale.ts
kanitw: I think this pattern is important.  We should figure this out before merging.

src/Encoding.ts
kanitw: SGTM.

src/encdef.ts
kanitw: This branch is a refactor branch ...

src/encdef.ts
kanitw: hmm, can you explain more what's wrong with that?

src/encdef.ts
domoritz: Where I avoid merge conflicts :-P

src/encdef.ts
domoritz: timefns is undefined

src/encdef.ts
kanitw: I don't think that this particular one will lead to more lead conflict thought.

package.json
```diff
@@ -16,7 +16,7 @@
     "build": "gulp build",
     "cover": "gulp coverage",
     "deploy": "npm run lint && npm run test && scripts/deploy.sh",
-    "lint": "gulp jshint",
+    "lint": "gulp jshint && gulp tslint",
```
kanitw: are we leaving any js code behind.  If not, why still running jshint?

Also a minor point that can be done later, if we can just run tslint directly, better run it directly?

src/compiler/scale.ts
```diff
@@ -1,46 +1,51 @@
...
     // Add optional properties
-    var properties = ['range', 'reverse', 'round',
-        'clamp', 'nice', // quantitative / time
-        'exponent', 'zero', // quantitative
-        'bandWidth', 'outerPadding', 'padding', 'points' // ordinal
-      ];
-
-    properties.forEach(function(property) {
-      var value = scale[property](encoding, name, type, layout, stats);
+    var properties = {
```
kanitw: This part looks weird also

src/compiler/style.ts
```diff
@@ -1,10 +1,8 @@
-'use strict';
+import * as vlEncDef from '../encdef';
 
-require('../globals');
+import {X, Y, ROW, COL} from '../consts';
 
-var vlEncDef = require('../encdef');
-
-module.exports = function(encoding, stats) {
+export function def(encoding, stats) {
```
kanitw: style doesn't produce a part of vega and thus shouldn't be named `def`.

I don't know how to just export in a way that compiler.ts can call `vlstyle(...)` rather than `vlstyle.def(...)`

package.json
domoritz: We still have the tests. But they will go at some point.

src/encdef.ts
domoritz: Done

src/compiler/style.ts
domoritz: Renamed it to style. You cannot export an anonymous function directly.

src/compiler/axis.ts
```diff
@@ -0,0 +1,299 @@
+import {getter, setter} from '../util';
+import {Encoding} from '../Encoding';
+import * as util from '../util';
+import {X, Y, COL, ROW} from '../consts';
+import {Q, O, N, T} from '../consts';
+
+import * as time from './time';
+
+export default function(name: string, encoding: Encoding, layout, stats) {
```
kanitw: export default = replacing namespace?  

In any case, this should be `axis.def`

domoritz: This is as intended. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import

src/compiler/compiler.ts
```diff
@@ -1,40 +1,37 @@
...
-compiler.subfacet = require('./subfacet');
-compiler.time = require('./time');
-
-compiler.compile = function (spec, stats, theme) {
-  return compiler.compileEncoding(Encoding.fromSpec(spec, theme), stats);
+
+import {summary} from '../util';
+import {Encoding} from '../Encoding';
+
+import * as scale from './scale';
```
kanitw: we probably should make these `vlScale, vlTime, vlAxis, vlSubfacet` for consistency?

domoritz:
> get rid of all .default

You mean in the tests?

kanitw:
> You mean in the tests?
In the main code for sure.   And in the tests (if possible).

domoritz:
@kanitw I think we are ready. Opinions?

domoritz:
Yaay

PR #710, opened 2015-11-07
Refactor axis to minimize side effects

kanitw:
- All produced properties that have special rules are organized in methods with the following patterns: `axis.<propertyName>` and `axis.properties.<propertyGroup>`
- #181 now support explicit specification of all Vega’s __axis properties__ (`'format', 'grid', 'offset', 'orient', 'tickSize', 'ticks', 'title', ‘titleOffset', 'layer', 'tickPadding', 'tickSize', 'tickSizeMajor', 'tickSizeMinor', 'tickSizeEnd',
    'values', ‘subdivide’`) and __property groups__ (`'axis', 'grid', 'labels', 'title', 'ticks', 'majorTicks', 'minorTicks’`)
- Output spec should remain the same except
  - `titleOffset` is set to `undefined` (auto) for `x` and `y` — so we can just rely on the [new Vega’s axis auto title offset](https://github.com/vega/vega/pull/421) — __This mean we need to wait until Vega releases new version before we can release one.__

### DIFF
--- a/src/compiler/axis.js
+++ b/src/compiler/axis.js
@@ -20,9 +20,9 @@ axis.def = function(name, encoding, layout, stats) {
 
   // Add optional properties
   [
-    // has special rules (so it has axis[property] methods)
+    // properties with special rules (so it has axis[property] methods) -- call rule functions
     'format', 'grid', 'orient', 'tickSize', 'ticks', 'title', 'titleOffset',
-    // only produce default values in the schema, or explicit value if specified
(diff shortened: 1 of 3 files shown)


domoritz:
@kanitw Since we have to wait for a new vega feature, should we wait or merge now?

kanitw:
I think just merge it.  The current master doesn't always have nice `titleOffset` anyway so this doesn't make it worse.  Also, this will allow you to merge this change into the TS branch.

PR #714, opened 2015-11-10
Enums

domoritz:
https://basarat.gitbooks.io/typescript/content/docs/enums.html


src/Encoding.ts
```diff
@@ -1,18 +1,17 @@
...
 import * as vlEncDef from './encdef';
 import * as vlEnc from './enc';
 import * as schema from './schema/schema';
 import * as schemaUtil from './schema/schemautil';
 
-export class Encoding {
+export default class Encoding {
   _data: any;
-  _marktype: any;
+  _marktype: string;  // TODO: replace with enum
```
kanitw: This should refer to MarkType interface for the schema

test/Encoding.test.js
```diff
@@ -1,15 +1,11 @@
-'use strict';
-
-var expect = require('chai').expect;
-
-var Encoding = require('../src/Encoding').Encoding;
-
+var chai_1 = require('chai');
```
kanitw: Why this weird `chai_1` pattern?

domoritz: Compiled from ts

kanitw: So I guess we check in both ts, js because we haven't migrated test to run on TS?

domoritz: No, just this one because I forgot to delete it. Sorry.

domoritz:
@kanitw ready to merge

PR #716, opened 2015-11-10
Tests

domoritz:
* Convert tests
* Use source maps to map from js to ts


src/compiler/time.ts
```diff
@@ -39,7 +39,7 @@ export function formula(timeUnit, fieldRef) {
   return fn + '(' + fieldRef + ')';
 };
 
-export function maxLength(timeUnit, encoding: Encoding) {
+export function maxLength(timeUnit, encoding: any) {
```
kanitw: why `:any` here?

test/compiler/marks.test.ts
```diff
@@ -1,18 +1,10 @@
-'use strict';
-
 var expect = require('chai').expect;
```
kanitw: why still require?  (isn’t there a way for node require for external module?)

test/compiler/scale.test.ts
```diff
@@ -30,7 +25,7 @@ describe('vl.compile.scale', function() {
             x: {name: 'x', type: 'O'},
             color: {name: 'color', type: 'O'}
           }
-        }), 'y', 'linear', {}, true);
+        }), 'y', 'linear', {});
```
kanitw: why changing this?

domoritz: Because vlscale.domain does not support that many params.

kanitw:
wrong merge target?

typings/node.d.ts
```diff
@@ -524,41 +524,41 @@ declare module "http" {
...
-    /**
-     * Maximum number of sockets to allow per host. Default for Node 0.10 is 5, default for Node 0.12 is Infinity
-     */
-    maxSockets?: number;
-    /**
-     * Maximum number of sockets to leave open in a free state. Only relevant if keepAlive is set to true. Default = 256.
-     */
-    maxFreeSockets?: number;
-  }
+	export interface AgentOptions {
```
kanitw: why this weird tabbing?

kanitw:
Replaced with #718

PR #719, opened 2015-11-11
fix for charts with rows with bin aggregation

jpocom:
There was an error on this type of vlSpec:

```json
{
 "encoding": {
  "row": {"name": "Acceleration","type": "Q","bin": true},
  "text": {
   "name": "*",
   "aggregate": "count",
   "type": "Q",
  },
  "color": {
   "name": "*",
   "aggregate": "count",
   "type": "Q",
  }
 },
 "marktype": "text",
 "data": {
  "formatType": "json",
  "url": "data/cars.json"
 }
}
```

I don't think this is the correct fix but at least it is working for me now.

### DIFF
--- a/src/compiler/scale.js
+++ b/src/compiler/scale.js
@@ -292,6 +292,8 @@ scale.range = function (encoding, name, type, layout, stats) {
       return 'shapes';
     case COLOR:
       return scale.color(encoding, name, type, stats);
+    case ROW:
+      return [0, layout.height];
   }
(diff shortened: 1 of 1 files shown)


kanitw:
@jpocom  Thanks for trying.

Please try #720 and see if that fixes all the cases you need.

If you run into other Vega-Lite, feel free to ping us on slack, or report a bug in the issue tracker :)

Issue #703, opened 2015-11-05
Split Schema.js and consider TypeScript


kanitw:
Right now schema.js is almost 1000 lines, we should 
- [x] at least split them off into multiple files
- [ ] consider if we would like to migrate to TypeScript.   Basically, the schema is describing the Vega-Lite object and if we’re doing that anyway, why don’t we declare these using Typescript’s interface and leverage it to prevent type error in the whole code base?   
  - There is a tool for generating JSON schema from TypeScript interface. However, we need to handle
    - `default`: Currently we generate default values by augmenting json schema with `default` property and used in our instantiate method.  
    - `role`: Currently we add `role` property to guide how UI can be automatically generated in Polestar.

domoritz:
We use typescript and split the schema.

kanitw:
we haven't declared interface...

domoritz:
The issue is not clear about that. Also, we have the same in #723.

kanitw:
Fair enough.   Closing then :)

PR #721, opened 2015-11-11
Break Schema into multiple files + Simplify Confusing parts + Move constants to const

kanitw:
part of #703


src/consts.ts
```diff
@@ -27,12 +27,27 @@ export enum Type {
   T,  // Time
 };
 
+export const TimeUnits = [
```
domoritz: why not all caps?

src/schema/encdef.scale.schema.ts
```diff
@@ -0,0 +1,119 @@
+import {toMap, duplicate as clone} from '../util';
+import {merge} from './schemautil';
+import {Type} from '../consts';
+
+var Q = Type[Type.Q];
```
domoritz: Like

domoritz:
Merge when you have looked at the all caps const comment.

PR #722, opened 2015-11-12
A step toward making row's `axis.orient = 'right'` correct

kanitw:
As @jpocom reported, that this spec
```
{
 "encoding": {
   "row": {
     "name": "Year",
     "type": "T",
     "timeUnit": "year",
     "axis": {
        "orient": "right"
     }
   },
   "text": {
     "name": "*",
     "aggregate": "count",
     "type": "Q",
   },
   "color": {
     "name": "*",
     "aggregate": "count",
     "type": "Q",
   }
 },
 "marktype": "text",
 "data": {
   "formatType": "json",
   "url": "data/cars.json"
 }
}
```
produces:
vega_editor

With this fix, now it is slightly better: 

vega_editor

### DIFF
--- a/src/compiler/axis.js
+++ b/src/compiler/axis.js
@@ -214,6 +214,8 @@ axis.properties.grid = function(encoding, name, spec, layout, def) {
     } else if (name == ROW) {
       var xOffset = encoding.config('cellGridOffset');
 
+      var sign = encoding.encDef(name).axis.orient === 'right' ? -1 : 1;
+
       // TODO(#677): this should depend on orient
       // set grid property -- put the lines on the top
       return util.extend({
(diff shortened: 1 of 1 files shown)


jheer:
See also vega/vega#422. Definitely a Vega-level issue involved here.

kanitw:
@jheer  Thank you.  

I will fix the grid offset for column and rely on Vega fix for legends then.

domoritz:
Later we should remove the grid lines and change the style so heat maps look more like https://vega.github.io/vega-editor/?mode=vega&spec=heatmap

kanitw:
Agreed.  I would like to revisit the "text" encoding before we release 1.0
but this is good for now.

PR #728, opened 2015-11-15
Add first sketch of vl.validate

kanitw:
- move `requiredEncodings` and `supportedEncodings` to `vl.validate`
- add `vl.validate.getEncodingMappingError()`


domoritz:
Why not `vl.validate()` or `vl.validateSpec()`?

kanitw:
I like `vl.validate()` as the unified method, but I also want to be able to call a method for particular rule e.g., `getEncodingMappingError()`.  

Should we make `vl.validate` as namespace and have `vl.validate.spec()`?   (Will be done in later PR though)

PR #740, opened 2015-11-18
simplify schema by extracting supportedRole to vl.Enctype

kanitw:


domoritz:
Does the schema change? Can you compare the output?

PR #745, opened 2015-11-18
Remove enter group 

kanitw:


kanitw:
> I have some code to remove enter and use width/height signals btw.

sounds like different part of "enter"

PR #751, opened 2015-11-19
Use datalib's month/day of week templates

domoritz:
Depends on https://github.com/vega/datalib/pull/44


src/compiler/axis.ts
```diff
@@ -268,9 +268,11 @@ namespace properties {
   export function labels(encoding: Encoding, name: string, spec, layout, def) {
     let fieldDef = encoding.fieldDef(name);
     var timeUnit = fieldDef.timeUnit;
-    if (fieldDef.type === Type.TEMPORAL && timeUnit && (time.hasScale(timeUnit))) {
+    if (fieldDef.type === Type.TEMPORAL && timeUnit && time.labelTemplate(timeUnit)) {
+      var filterName = fieldDef.axis.abbreviatedTimeNames ? 'abbrev-' : '';
```
kanitw: `abbreviatedTime`?

src/compiler/time.ts
```diff
@@ -40,21 +40,21 @@ export function formula(timeUnit, fieldRef) {
...
+export function maxLength(fieldDef, encoding: Encoding) {
+  switch (fieldDef.timeUnit) {
     case 'seconds':
     case 'minutes':
     case 'hours':
     case 'date':
       return 2;
     case 'month':
+      if (fieldDef.axis.abbreviatedTimeNames)
+        return 3;
```
kanitw: is it always 3?

src/compiler/axis.ts
domoritz: I cannot abbreviate time :-P

Yeah, finding a good term here is tricky.

src/compiler/time.ts
```diff
@@ -154,12 +122,13 @@ export namespace scale {
   }
 }
 
-/** whether a particular time function has custom scale for labels implemented in time.scale */
-export function hasScale(timeUnit) {
+/** returns the teplate name used for axis labels for a time function */
```
kanitw: typos:
“template” 
“time unit"

src/compiler/time.ts
domoritz: I guess so. We don't really have a way to do this properly because it may depend on the locale setting.

src/compiler/time.ts
domoritz: Done

gulp/lint.js
```diff
@@ -7,17 +7,10 @@ var $ = require('gulp-load-plugins')();
 gulp.task('tslint', function() {
   return gulp.src([
       paths.src + '/**/*.ts',
+      paths.test + '/**/*.ts',
```
kanitw: you should just cherry pick this to master

domoritz:
screen shot 2015-11-18 at 21 49 19

domoritz:
Ready when @arvind releases new vega with latest datalib.

domoritz:
@kanitw why did you delete my branch?

kanitw:
Misclick when I was removing merged branches

domoritz:
Maybe not a bad idea to remove as we need to redo it pretty much at this point :-P

PR #754, opened 2015-11-19
Encoding => Model

kanitw:
rename internal `Encoding` class to be `Model` (rather than `Specification`).

This PR should not change the external API yet.  

So specification would only refer to Vega-Lite's JSON spec and model means the internal version that we use in compiler.   

Note that this would cause conflict with the pending PR, but I can help merge that.


src/vl.ts
```diff
@@ -14,20 +14,19 @@ import {format as d3Format} from 'd3-format';
 export * from './util';
 export * from './consts';
 
-// TODO(#727) split this into two namespace
+// TODO() rename this
+export var Encoding = vlModel;
```
domoritz: This breaks the linter. Please use the export from the code you deleted.

kanitw: I do not know what you mean

domoritz: Run `npm run lint`

PR #758, opened 2015-11-20
Add interfaces

kanitw:
## TODO: later
- [ ] things in schema are actually constants — we should eventually use ALL CAPS?
- [ ] declare interface for Config (once config is more stable)


src/schema/axis.schema.ts
```diff
@@ -1,3 +1,18 @@
+export interface Axis {
```
domoritz: FYI, we should call this IAxis. Seems to be the standard.

domoritz: `interface-name enforces the rule that interface names must begin with a capital 'I'`

https://github.com/palantir/tslint

domoritz: Discussed offline.

domoritz:
fielddef.schema missing @kanitw

kanitw:
@domoritz Can you clarify that?

domoritz:
The tests are failing and this PR is not marked WIP.

kanitw:
If you test that offline, do you have the same problem?   I don't know how to reproduce that.  Test are passing in my machine.

domoritz:
Hmm, odd. No, don't see that issue.

src/schema/schema.ts
```diff
@@ -1,305 +1,35 @@
...
+export interface Spec {
+  data?: Data;
+  marktype?: string;
+  encoding?: Encoding;
+  config?: any; // FIXME: declare
+}
 
-export {aggregate} from './fielddef.schema';
+// TODO remove this
+export {aggregate} from './fieldDef.schema';
```
domoritz: OSX is weird and does not respect capitalization. You need to fix this!

kanitw:
Done

PR #765, opened 2015-11-22
Fix stack 

### LINKED ISSUE #683: Revising stack syntaxes
- [x] 1) __Revise if we should add “-“ to stack’s sort order__

We automatically  add `’-‘` to the output Vega’s `sortby` property of `stack` transform to make legends consistent with stack order for vertical stacked bar chart.

vega_editor

However, this might not make sense for horizontal stacked bar.

Should we only add `’-'` for vertical stacked bar chart?

Also, if a field’s `sort` is `’descending’`, we might need to revert the `'-‘` appended. 

----

- [ ] 2) __Do we only allow stack if `aggregate` is `sum` ?__
Stack of `mean`, `median`, `min`, `max` certainly do not make sense.   
What shall we do when people specify stack of mean / median?  

- Do we throw error/warning?
- Do we should layered bar chart for these by default?

-----

- [ ] 3) __Should stack be a part of `x`,`y` instead of `color`/`detail`?__

`stack.offset` actually applies to the quantitative encoding channel (either `x` or `y`) rather than `color`/`detail`.

However, stack is only applied when either `color` or `detail` is specified. 

——

Any thoughts on these 3 issues?

kanitw:
__Please merge #764 first !  (See [Branch Specific Diff](https://github.com/uwdata/vega-lite/compare/kw/model...kw/stack))__

Fixes #683 

## syntax 
- move `stack` property from `color`/`detail` to `config`
  - make `stack` config supports an array of fields. 

## output 
- make stack uses descending by default 

## API 
- update `vl.spec.isStack()`

## Internal
- add StackDef interface
- calculate `_stack` only once in model 

## Docs
- Add docs for stack in `Config.md` 
——

### DIFF
--- a/src/compiler/data.ts
+++ b/src/compiler/data.ts
@@ -17,7 +17,7 @@ import {NOMINAL, ORDINAL, QUANTITATIVE, TEMPORAL} from '../type';
  *                 If the encoding contains aggregate value, this will also create
  *                 aggregate table as well.
  */
-export default function(model: Model) {
+export function def(model: Model) {
   var def = [source.def(model)];
 
   var summaryDef = summary.def(model);
--- a/src/compiler/data.ts
+++ b/src/compiler/data.ts
@@ -2,6 +2,7 @@ import * as vlFieldDef from '../fielddef';
 import * as util from '../util';
 import {Model} from './Model';
 import {FieldDef} from '../schema/fielddef.schema';
+import {StackDef} from './stack';
 
 import {MAXBINS_DEFAULT} from '../bin';
 import {Channel} from '../channel';
@@ -32,9 +33,9 @@ export function def(model: Model) {
   filterNonPositive(def[def.length - 1], model);
 
   // Stack
-  var stackCfg = model.stack();
-  if (stackCfg) {
-    def.push(stack.def(model, stackCfg));
+  var stackDef = model.stack();
+  if (stackDef) {
+    def.push(stack.def(model, stackDef));
   }
 
   return def;
@@ -254,9 +255,9 @@ export namespace stack {
   /**
    * Add stacked data source, for feeding the shared scale.
    */
-  export function def(model: Model, stackCfg):VgData {
-    var dim = stackCfg.groupby;
-    var val = stackCfg.value;
+  export function def(model: Model, stackDef: StackDef):VgData {
+    var groupbyChannel = stackDef.groupbyChannel;
+    var fieldChannel = stackDef.fieldChannel;
     var facets = model.facets();
 
     var stacked:VgData = {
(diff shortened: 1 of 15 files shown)


src/compiler/data.ts
```diff
@@ -18,7 +19,7 @@ import {NOMINAL, ORDINAL, QUANTITATIVE, TEMPORAL} from '../type';
  *                 If the encoding contains aggregate value, this will also create
  *                 aggregate table as well.
  */
-export default function(model: Model) {
+export function def(model: Model) {
```
domoritz: What is the reason for this?

src/compiler/data.ts
```diff
@@ -18,7 +19,7 @@ import {NOMINAL, ORDINAL, QUANTITATIVE, TEMPORAL} from '../type';
...
   var def = [source.def(model)];
```
domoritz: def used again

src/compiler/data.ts
```diff
@@ -254,9 +255,9 @@ export namespace stack {
   /**
    * Add stacked data source, for feeding the shared scale.
    */
-  export function def(model: Model, stackCfg):VgData {
-    var dim = stackCfg.groupby;
-    var val = stackCfg.value;
+  export function def(model: Model, stackDef: StackDef):VgData {
```
domoritz: and again

src/compiler/data.ts
kanitw: good catch ...

src/compiler/data.ts
kanitw: We shouldn't use `default` because it creates weird `.default` when you do require in the compiled javascript.  But I do agree that `def` isn't the right way to name this.

domoritz: I disagree that default is weird. It is in the es6 standard https://people.mozilla.org/~jorendorff/es6-draft.html#sec-exports. I think we shouldn't change this for the wrong reasons.

kanitw: While default is in the ES6 standard, that doesn't mean that it is a good practice.   

By using default, that means you always create alias with arbitrary names every time you import and there is no mechanism for enforcing that you will name the alias similarly.  Therefore, unless there is a good reason.  Providing method name such as `compileData()` and do `import {compileData} from './data.ts'` is a much better way.   On that note, I'm creating a PR to change all confusing `def()`, `defs()` to `compileXXX()` ... will send you when I'm done.

domoritz: I still disagree that `default` is always a bad idea but agree that we should use `compileXXX()`.

kanitw: I'm not saying it is __always__ bad, but avoid it

>  unless there is a good reason

src/compiler/data.ts
kanitw: This doesn't create redundant variable name though.  

That said, I don't like the `stackDef` name.   But can't think of a better name.  
It combines some properties and config in the same object so I can't call it `StackConfig`.  

Any suggestion?

src/compiler/data.ts
domoritz: Fair enough. Do you think it is sensible to say that we only use default if 1) there is only one function exported from a module 2) that function pretty much resembles what you would expect the module to do.

src/compiler/data.ts
domoritz: stackProps maybe

kanitw: or stackModel? lol

PR #766, opened 2015-11-22
Replace def(), defs() with compileXXX() methods and remove export default where it’s not necessary 

kanitw:
Note that I leave `facet`, `layout`, `stack`, `style`, `subfacet`, and `time` since these will be further refactored to simplify flow (#276) and some of them might be removed so I don’t want to waste time refactoring them now.


kanitw:
Note that I’m merging this into `kw/stack`

domoritz:
Sorry, already merged stack.

Issue #37, opened 2014-12-06
Generated Vega spec can show fractional widths


kanitw:
Generated Vega spec can show fractional widths (e.g., 1300.0000000000002). These should be integers.

kanitw:
Since we dynamically calculate layout, the spec would no longer include fractional width.

PR #774, opened 2015-11-22
Deal with floating point instability in javascript

kanitw:
Part of #37

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -8,6 +8,14 @@ export * from 'datalib/src/stats';
 import dlBin = require('datalib/src/bins/bins');
 export var bin = dlBin;
 
+/*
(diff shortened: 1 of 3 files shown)


src/util.ts
```diff
@@ -8,6 +8,14 @@ export * from 'datalib/src/stats';
 import dlBin = require('datalib/src/bins/bins');
 export var bin = dlBin;
 
+/*
+ * Deal with floating point instabiility in javasscript to avoid weird .000001 output
+ * http://stackoverflow.com/questions/11695618/dealing-with-float-precision-in-javascript
+ */
+export function roundFloat(x: number, digit:number = 4) {
+  return +x.toFixed(digit);
```
domoritz: We should use http://herbie.uwplse.org/ :-P

PR #780, opened 2015-11-23
Builds

domoritz:


kanitw:
LGTM.   Btw, do we still plan to eliminate gulp in longer term?

kanitw:
Changes in this PR leads to infinite loop in the build system!

My build keep running all the time which is not the best way to use my CPU cycle, esp there is no file change!

domoritz:
Ups, I'm looking into it.

PR #789, opened 2015-11-24
Make marks.ts far less confusing + Use Update instead of enter for root 

kanitw:
- Make `marks.ts` far less confusing 
  - refactor the weird pattern on the top
  - `e` => `model`
- Use `update` instead of `enter` for root 
- Use `{group: "width|height"}` for cell if there is no facet

### DIFF
--- a/src/compiler/compiler.ts
+++ b/src/compiler/compiler.ts
@@ -45,11 +45,11 @@ export function compile(spec, stats, theme?) {
         name: 'cell',
         type: 'group',
         properties: {
-          enter: {
-            width: layout.cellWidth ?
+          update: {
+            width: model.has(COLUMN) ?
                      {value: layout.cellWidth} :
                      {field: {group: 'width'}},
-            height: layout.cellHeight ?
+            height: model.has(ROW) ?
                     {value: layout.cellHeight} :
                     {field: {group: 'height'}}
           }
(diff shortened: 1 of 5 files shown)


src/compiler/compiler.ts
```diff
@@ -45,11 +45,11 @@ export function compile(spec, stats, theme?) {
         name: 'cell',
         type: 'group',
         properties: {
-          enter: {
-            width: layout.cellWidth ?
+          update: {
```
domoritz: why

kanitw: For consistency.  (Otherwise, a part of the spec would use enter, another part would use update)

src/compiler/marks.ts
```diff
@@ -1,111 +1,84 @@
...
-  prop: point_props
-};
-
-export const text = {
-  type: 'text',
-  prop: text_props
-};
-
-function bar_props(e: Model, layout, style) {
+export namespace properties {
```
kanitw: @domoritz  I start wondering if we should organize this in the opposite way e.g.,
```ts
export namespace bar  {
   export function properties() {
   }
}
```

This way, we can easily add `label` to each mark's namespace.

domoritz: I'm having a hard time following this argument.

kanitw: nvm.  Let's ignore it.

kanitw:
@domoritz  If you don’t have other objection — merge this?

PR #790, opened 2015-11-24
[WIP] Reduce side effect and simplify flow

kanitw:
- remove `compileScaleNames`
- merge stack and subfacet's logic into `marks.ts`
- Replace `autoSortLine` with `sortLineBy`  in case we can to sort line's order with something rather than the dimension.  This option is only enabled for area chart though.  

(This addresses part of #276)

Note: flow for facet will be further simplified in another PR.


domoritz:
@kanitw still WIP?

kanitw:
Done :)

Issue #648, opened 2015-10-14
Replace Stats with Vega 2 signals and data flow


kanitw:
We use limit the use of stats, which is found in the following places

- `axis.js`
  - [x] using cardinality to determine axis’s `orient` (__This is a specific hack for voyager — should be only done for voyager__)
  - [x] using field’s primitive data type to determine number `format`

- [x] `compiler.js` (just for passing stats around)

- `data.js`
  - [x]  used as  `min`, `max` parameters of `bin` transform

- `layout.js`
  - [x] using x,y cardinality (from `distinct` and `null`) to determine layout width / height 
    - [x] scale.range()
    - [x] scale.bandWidth
    - [x] axis.offset
    - [x] axis.grid (grid for small multiples) 
  - [x] use cardinality to determine bandSize
  - [x] use `max` value as a heuristic to determine maxTextLength, which is used for assigning axisOffset. (This is totally a dumb heuristic and should be replaced anyway!)

- `marks.js`
  - [x] using ??? (in `encoding.numberFormat`) to determine text table’s number format 

- `scale.js`
  - [x] using `min`, `max` (`util.getbins()`) to hard code bin’s domain — see #614 for explanation
  - [x] using `distinct` and `null` in `scale.color` to pick color palette from cardinality

- `style.js`
  - [x] using `max` and `cardinality` to determine opacity

- `time.js`
  - [x] use `distinct` and `nulls` value to estimate cardinality (this should totally be removed!)

- [ ] `data.js`  (just the wrapper for calling stats in datalib)

- `Encoding.js`
  - [x] used for calling `Encoding.alwaysNoOcclusion` in `Encoding.toggleSort.support`
  - [x] used `nulls` in `Encoding.toggleFilterNullO.support`

kanitw:
stats removed from main compiler from in #800, so I’m closing this.  

Leftover stuff are noted in #276.

PR #792, opened 2015-11-25
remove style.ts and automatic style calculation

kanitw:
So we do not have to pass `style` around and simplify the flow (#276) and help reduce reliance on stats (#648)

Even if we decided to set `opacity = 0.7` for `point` by default, this should still be done.   (We can add `opacity(channel)` helper in `Model.ts` later)


src/compiler/marks.ts
```diff
@@ -506,8 +512,10 @@ export function text(model: Model, layout, style) {
   // color should be set to background
   p.fill = {value: fieldDef.color};
 
-  var opacity = model.fieldDef(COLOR).opacity  || style.opacity;
-  if(opacity) p.opacity = {value: opacity};
+  var opacity = model.fieldDef(COLOR).opacity;
```
domoritz: Can you add a note here that this is where we decide to set or not to set the opacity?

kanitw: We can just look at this PR.

PR #793, opened 2015-11-25
remove stats from axis

kanitw:
This removes logic for automatically putting x-axis on top if y has high cardinality.

We should move this logic to `vlplot.js` so it’s still applied in both voyager and polestar.


domoritz:
Also don't forget to edit the gallery example.

Issue #741, opened 2015-11-18
Faceting should have "cross" semantic


domoritz:
Currently faceting a chart by both row and column do not produce cross results between two fields.

kanitw:
I will remove the grid as mentioned in #822, but I think imputing cell for background can be a post-1.0 patch.

kanitw:
Example:
```js
{
  "data": {"url": "data/movies.json"},
  "mark": "point",
  "encoding": {
    "column": {"field": "MPAA_Rating","type": "ordinal"},
    "row": {"field": "Major_Genre","type": "ordinal"},
    "x": {"field": "Worldwide_Gross", "type": "quantitative"},
    "y": {"field": "US_DVD_Sales", "type": "quantitative"}
  },
  "config": {
    "cell": {"fill": "yellow"} // I’m just adding this to demonstrate a minor bug 
  }
}
```

Current do not have background for all cells

![vega-lite_composition_-_google_docs](https://cloud.githubusercontent.com/assets/111269/13062288/4fc081f2-d3f2-11e5-9376-a69ec9901661.png)

kanitw:
Instead of producing just the facet transform.  We might have to: 

1. Produce unique values of column field and row field.  
2. Cross the unique value of both fields.  (and project both fields to the top-level again)
3. Produce faceted table
4. Join the crossed table (2.) with faceted table.  

(Or alternatively, we can extend Vega facet transform) 

In the case of faceting the same field by both row and column.  We can use the same step with minor modification. 
1. Produce unique values of **the field**
2. **Self cross the output from 1. (without projecting the field to the top-level as it is ambiguous)**  
3. Produce facet table
4. Join the crossed table (2.) with faceted table **where `row.<field> == <field> && column.<field> == <field>`**

kanitw:
Here is an example spec that facets by Origin for both row and column.  

```js
{
  "description": "A scatterplot showing horsepower and miles per gallons for various cars.",
  "data": {"url": "data/cars.json"},
  "mark": "point",
  "encoding": {
    "row": {"field": "Origin","type": "ordinal"},
    "column": {"field": "Origin","type": "ordinal"},
    "size": {"aggregate": "count", "field": "*","type": "quantitative"}
  },
  "config": {
    "cell": {"fill": "yellow"}
  }
}
```

Ideally, we should see yellow background in all cells.  

[...]

PR #800, opened 2015-11-27
Remove stats from layout

kanitw:
### Syntax 
- remove `col.width`, `row.height` => use `config.cellWidth`, `config.cellHeight`

### API
- __`stats` parameter removed from `vl.compile`__

### Output 
- migrate the old layout code base to do dynamically in Vega and simplify 
  - Create new data table `layout` that calculates `cellWidth`, `cellHeight`, `width`, `height` when necessary 
  - Move Row’s label to the right side instead so we do not have to guess `axisLabelOffset`
  - Remove axis grid for `row` and `column` and add `rule` marks (`row|column-rules`) instead for drawing  rules/grid between rows / columns  

- try to use `{field: {group: ‘width|height’}}` where possible

- simplify COLOR scale’s rule and remove stats #648 
  - remove `c10palette`, `c20palette`, `ordinalPalette`, `quantitativeRange` from scale.

Please merge #798  first !


domoritz:
:sparkles:

kanitw:
@domoritz  Okay.  This “remove stats” PR is ready for review.  
Now text table heat map looks slight better and we can further improve it later since we have #741 open.  

Here are also some notes for the future:

#### TODO after this PR 
- Optimize Data Sources 
- consider removing leftover stats stuff #276 
  - vl.fieldDef.cardinality
  - time.cardinality 
  - vl.data.stats()

- [ ] Ping Vega team to add “auto” mode for top-level width/height so we can use it.  
- [ ] Line for Text Table 
- [ ] Consider if we want to put commonly used constant in signal for output readability

PR #808, opened 2015-11-30
Dates

domoritz:


gallery/examples.js
```diff
@@ -361,6 +361,22 @@ var EXAMPLES = [
         "x": {"type": "ordinal","field": "dataset"}
       }
     }
+  }, {
+    title: "Birdstrikes throughout the year",
```
kanitw: "Number of"

src/compiler/axis.ts
```diff
@@ -85,9 +85,9 @@ export function grid(model: Model, channel: Channel) {
   }
 
   // If `grid` is unspecified, the default value is `true` for
-  // - X and Y that have (1) quantitative fields that are not binned or (2) time fields.
+  // - X and Y that have (1) quantitative fields that are not binned or (2) time fields that don't have a timeUnit.
```
kanitw: what if you have timeUnit but the timeUnit leads to linear scale?  (e.g., minutes / seconds)?

src/compiler/axis.ts
```diff
@@ -208,9 +208,11 @@ namespace properties {
   export function labels(model: Model, channel: Channel, spec, def) {
     let fieldDef = model.fieldDef(channel);
     var timeUnit = fieldDef.timeUnit;
-    if (fieldDef.type === TEMPORAL && timeUnit && (time.hasScale(timeUnit))) {
+    var filterName = time.labelTemplate(timeUnit, fieldDef.axis.shortTimeNames);
+    if (fieldDef.type === TEMPORAL && timeUnit && time.labelTemplate(timeUnit)) {
```
kanitw: && filterName?

src/compiler/axis.ts
```diff
@@ -208,9 +208,11 @@ namespace properties {
   export function labels(model: Model, channel: Channel, spec, def) {
     let fieldDef = model.fieldDef(channel);
     var timeUnit = fieldDef.timeUnit;
-    if (fieldDef.type === TEMPORAL && timeUnit && (time.hasScale(timeUnit))) {
+    var filterName = time.labelTemplate(timeUnit, fieldDef.axis.shortTimeNames);
```
kanitw: what happens with this line when there is no `timeUnit`?

kanitw:
I have a few comments — otherwise LGTM.  

Can you also provide a summary of this PR, so I can copy to the release note?

PR #810, opened 2015-11-30
FieldDef type

domoritz:


src/compiler/legend.ts
```diff
@@ -43,7 +43,7 @@ export function compileLegend(model: Model, channel: Channel, def) {
   });
 
   // 2) Add mark property definition groups
-  let props = legend.properties || {};
+  const props = typeof legend === 'Legend' && legend.properties || {};
```
kanitw: This line is terrible!  Nobody on earth remember precedence of these two operators. 
I also do not get what you try to do here?

src/schema/sort.schema.ts
```diff
@@ -8,6 +8,11 @@ export interface Sort {
   order?: string;
 }
 
+export function isSort(object: any): object is Sort {
```
kanitw: I don't understand why do you need to add this?

domoritz: You cannot do runtime type checking.

kanitw: First, I still do not get it. 
Second, if there is no way to understand this easily, should we ignore it?

domoritz: A good explanation is https://github.com/Microsoft/TypeScript/issues/1007

src/compiler/legend.ts
```diff
@@ -58,9 +60,12 @@ export function compileLegend(model: Model, channel: Channel, def) {
 }
 
 export function title(model: Model, channel: Channel) {
-  let leg = model.fieldDef(channel).legend;
+  // https://github.com/Microsoft/TypeScript/issues/5842
+  const legend: any = model.fieldDef(channel).legend;
 
-  if (leg.title) return leg.title;
+  if (isLegend(legend) && legend.title) {
```
kanitw: What's the problem here?

domoritz: `legend` is of type `Legend | boolean` so it does not necessarily have a property `title`.

kanitw: But if there is no property title, this predicate is then false, so that's fine?

domoritz: Yes, but not for the type checker. 

```js
var a = true;
a.title  // you expect this to crash
```

Issue #666, opened 2015-10-16
Revise Config


kanitw:
- [x] add top-level’s `name`, `padding`, `background`, `description`, `scene`

- leave as top-level as-is: 
  - `sortLineBy`, `stack`
  

- Group things so we can easy map values to vega and help organize config UI for polestar / voyager 
  - `cell`
    - [x]  `background` (formerly `cellBackgroundColor`),  `gridColor`, `gridOffset`,  `gridOpacity`, `padding`, `width`, `height` and other group mark’s properties 
  - `marks`: 
    - [x] `filled` (formerly `encoding.shape.filled`) 
    - [x] opacity (formerly `encoding.color.opacity`) 
    - [x] move all text properties that are currently in `encoding.text`
    - [x] support other non-channel properties (#730) except `shapeOpacity` and `filledOpacity` (users should simply use `opacity`) 
      - [x] text’s dx, dy, radius, theta, angle 
    - [x] `strokeWidth`
  
- remove 
  - [x] `useRawDomain`, `padding`, `bandWidth` — just set default value in scale rather than having two configs. 
  - [x] `gridColor`, `gridOpacity` (unused)
  - [x] `singleHeight`, `singleWidth` — simply always use `cell.width|height`
- need further revision before deciding what to do 
  - [x] ~~`numberFormat`, `timeFormat`~~ has its own #497
  - [ ] ~~filterNull~~ has its own #597
  - [x] `singleBarOffset`
  - [x] `characterWidth`  — move to axis

domoritz:
Let's rename `shapeFilled` to just `filled` because otherwise there is confusion with the shape channel.

kanitw:
- [ ] `config.marks` => `config.mark` ? 

(To be consistent with `config.cell`, which is singular)

kanitw:
- [ ] @domoritz How should we name `cell.gridColor|opacity` – should it still be called `grid` because we actually implement it using rules marks.  But semantically it’s not a rule either.

domoritz:
Let's use the term grid. It doesn't matter how we implemented it.

PR #811, opened 2015-12-01
Revise config 

kanitw:
Most important part of #666 for 0.9.0 — still left a few to be removed/added later.  

### Config 
- remove `single/cell Width/Height`, always use `cell.width|height`
- remove unused `gridColor`, `gridOpacity`
- remove `config.bandWidth|padding` and `model.bandWidth|padding()` and set different default values for different channels instead!
- remove config.useRawDomain

- group configs for a particular part
  - `config.cell`
    - move `cell` configs to be under `cell`:  `width`, `height`, `padding`, `gridColor`, `gridOpacity`, `gridOffset`
    - add group marks properties to `config.cell`: `fill`, `fillOpacity`, `stroke`, `strokeWidth`, `strokeOpacity`, strokeDash, strokeDashOffset`
  - `config.marks`
    - move `baseline`, `font`, `fontStyle`, `fontWeight` from `encoding.text` to `config.marks`
    -  `encoding.shape.filled` => `config.marks.filled`, `encoding.color.opacity` => `config.marks.opacity`
    - move all text properties from `encoding.text` to `config.marks` except renaming `encoding.text.placeholder` to `encoding.text.value`
    - `config.strokeWidth` => `config.marks.strokeWidth`

### output 
- set width and height to 1 if it's dynamically calculated

### DIFF
--- a/src/schema/encoding.schema.ts
+++ b/src/schema/encoding.schema.ts
@@ -39,7 +39,6 @@ var x = merge(duplicate(typicalField), requiredNameType, {
   }
 });
 
-console.log('xAxis', x.properties.axis);
 var y = duplicate(x);
 
 var facet = merge(duplicate(onlyOrdinalField), requiredNameType, {
--- a/src/schema/encoding.schema.ts
+++ b/src/schema/encoding.schema.ts
@@ -127,10 +127,7 @@ var text = merge(duplicate(typicalField), {
       type: 'string',
       default: 'right'
     },
-    baseline: {
-      type: 'string',
-      default: 'middle'
-    },
+
     color: {
       type: 'string',
       role: 'color',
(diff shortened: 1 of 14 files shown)


src/schema/encoding.schema.ts
```diff
@@ -28,10 +28,17 @@ var requiredNameType = {
 
 var x = merge(duplicate(typicalField), requiredNameType, {
   properties: {
-    axis: axis,
+    scale: {// replacing default values for just these two axis
```
domoritz: axes

timelyportfolio:
I like these changes, but how now shall we try to fill width, since `config.bandWidth` is no longer available?

kanitw:
@timelyportfolio   Use `encoding.<channel>.scale.bandWidth` :) 
We remove the config one because it’s redundant!

timelyportfolio:
oh great, thanks.  I thought I looked there but missed it.

PR #814, opened 2015-12-01
Simplify type checking

kanitw:
If you can't type check the interface object, check if it's not the primitive and you can get away with it without introducing complex `isX` methods

### DIFF
--- a/src/compiler/legend.ts
+++ b/src/compiler/legend.ts
@@ -3,7 +3,6 @@ import {COLOR, SIZE, SHAPE, Channel} from '../channel';
 import {Model} from './Model';
 import * as time from './time';
 import {TEMPORAL} from '../type';
-import {isLegend} from '../schema/legend.schema';
 
 export function compileLegends(model: Model) {
   var defs = [];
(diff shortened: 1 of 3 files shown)


src/compiler/legend.ts
```diff
@@ -45,7 +44,7 @@ export function compileLegend(model: Model, channel: Channel, def) {
   });
 
   // 2) Add mark property definition groups
-  const props =  isLegend(legend) && legend.properties || {};
+  const props = (typeof legend !== 'boolean' && legend.properties) || {};
```
domoritz: Can you use `isinstance`?

domoritz: https://stackoverflow.com/questions/899574/which-is-best-to-use-typeof-or-instanceof

kanitw: `true instanceof Boolean`  is `false`

`instanceof` is for checking prototype chain.  `typeof` is the right way to test primitive type.

kanitw: The top vote answer is the article you provided also says

> Use typeof for simple built in types:

domoritz: Ok.

PR #818, opened 2015-12-02
scale documentation and minor refactor

kanitw:
I think we should extract “transformation”, “axis”, “scale” and “legend” as their own page in a later PR.

### DIFF
--- a/src/compiler/time.ts
+++ b/src/compiler/time.ts
@@ -52,27 +52,31 @@ export function range(timeUnit, model: Model) {
   return;
 }
 
-function isOrdinalFn(timeUnit) {
-  switch (timeUnit) {
-    case 'seconds':
-    case 'minutes':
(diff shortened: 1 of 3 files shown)


src/compiler/time.ts
```diff
@@ -52,27 +52,31 @@ export function range(timeUnit, model: Model) {
...
+      return 'ordinal';
+    }
+
+    switch (timeUnit) {
+      case 'hours':
+      case 'day':
+      case 'date':
+      case 'month':
+        return 'ordinal';
+      case 'year':
```
domoritz: year is linear too?

kanitw: You think it should be "time", right?

domoritz: Probably. But since you are not changing anything, let's discuss that later.

PR #820, opened 2015-12-02
Remove gulp

domoritz:
Boom, deleted 200 lines of code and removed a ton of deps.


test/compiler/scale.test.ts
```diff
@@ -2,7 +2,6 @@
 
 import {expect} from 'chai';
 
-var d3 = require('d3');
```
domoritz: Not sure what this was doing... adding d3 as a req.

arvind:
:tada:

test/compiler/scale.test.ts
kanitw: Wait I think you just remove it?

domoritz: Yes, I just removed it because it wasn't doing anything. It just surprised me that it was there.

package.json
```diff
@@ -13,14 +13,20 @@
     "test": "test"
   },
   "scripts": {
-    "build": "gulp build",
-    "cover": "gulp coverage",
-    "clean": "gulp clean",
+    "build": "browserify src/vl.ts -p tsify -d -s vl | exorcist vega-lite.js.map > vega-lite.js",
+    "postbuild": "uglifyjs vega-lite.js -cm --source-map vega-lite.min.js.map > vega-lite.min.js",
```
kanitw: why are there two `"postbuild"`?

PR #824, opened 2015-12-02
Some minor revisions

kanitw:
- remove weird aggregate count from only ordinalField
- remove deprecated rules for offset (we no longer use grid for row and column))
- comments

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -21,6 +21,12 @@ interface SupportedMarktype {
   [marktype: string]: boolean;
 };
 
+/**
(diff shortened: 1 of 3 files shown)


src/channel.ts
```diff
@@ -21,6 +21,12 @@ interface SupportedMarktype {
   [marktype: string]: boolean;
 };
 
+/**
+ * Return whether a channel supports a particular mark type.
+ * @param  {string} channel  channel name
+ * @param  {string} marktype
+ * @return {boolean}
```
domoritz: you can omit the type annotations

PR #826, opened 2015-12-03
Enums

domoritz:


domoritz:
Note that the only changes related to enums are in 51df2c640d1657499c9c2ff092852f5. The other commit makes sense regardless.

domoritz:
Okay, did the same for other types as well and fixed a bunch of bugs along the way. 

The cool thing is that nothing changes for external users. The only thing that changes is that we can rely on correct types.

src/shorthand.ts
```diff
@@ -26,7 +27,7 @@ export function parse(shorthand: string, data?, config?) {
     encoding = parseEncoding(split.join(DELIM));
 
   let spec:Spec = {
-    marktype: marktype,
+    marktype: Marktype[marktype],
```
domoritz: This is slightly ugly because we go from string to Marktype type but I think that's okay.

src/type.ts
```diff
@@ -28,10 +41,10 @@ export const TYPE_FROM_SHORT_TYPE = {
...
- * @param  {String} type
- * @return {String} Full type name.
+ * @param  type
+ * @return Full type name.
  */
-export function getFullName(type: string): string {
-  return TYPE_FROM_SHORT_TYPE[type.toUpperCase()] || // short type is uppercase by default
-         type.toLowerCase();
+export function getFullName(type: Type): Type {
+  return TYPE_FROM_SHORT_TYPE[Type[type].toUpperCase()] || // short type is uppercase by default
```
domoritz: I admit it looks a bit ugly but we somehow have to make sure that if we get garbage in that we get the right thing out. I think it's not too bad.

kanitw: I don't think we gain that much from adding this complexity ....

domoritz: Well, this is just needed because we support `Q` and `quantitative`. And the complexity is very well contained in this function.

src/compiler/Model.ts
```diff
@@ -207,7 +207,7 @@ export class Model {
     return fieldDef && (
       contains([NOMINAL, ORDINAL], fieldDef.type) ||
       (fieldDef.type === TEMPORAL && fieldDef.timeUnit &&
-        time.scale.type(fieldDef.timeUnit, channel) === ORDINAL)
+        time.scale.type(fieldDef.timeUnit, channel) === <any>ORDINAL)
```
kanitw: having to put `<any>` when you have to do this kind of comparison is already a sign of weakness of this enum approach!

src/type.ts
kanitw: This introduce one more level of complexity!  so I don't think that's a fair argument.

kanitw:
If say, I migrate Compass to TS, then I have another headache to reason about what's enum and what's not, so I prefer to do it.

src/compiler/layout.ts
```diff
@@ -3,6 +3,7 @@ import * as time from './time';
 
 import {FieldDef} from '../schema/fielddef.schema';
 import {COLUMN, ROW, X, Y, TEXT, Channel} from '../channel';
+import {TEXT as TEXT_MT} from '../marktype';
```
kanitw: TEXT_MARK?

src/compiler/scale.ts
```diff
@@ -41,7 +41,7 @@ export function compileScales(names: Array<string>, model: Model) {
   }, []);
 }
 
-export function type(channel: Channel, model: Model) {
+export function type(channel: Channel, model: Model): Type | string {
```
kanitw: this returns scale’s type, not the data `Type`, so only `string` is the correct semantic.

src/type.ts
kanitw: I read the code and also doesn’t understand why it works for short type and from pulling and testing — this does not work for short type!

I found this extremely confusing and if the one who proposes it also get confused by it, I don’t think this is worth it.

src/compiler/layout.ts
domoritz: Done.

src/compiler/scale.ts
domoritz: Done

src/type.ts
domoritz: Fixed and I wrote a test for it.

src/compiler/Model.ts
domoritz: Fixed.

domoritz:
> If say, I migrate Compass to TS, then I have another headache to reason about what's enum and what's not, so I prefer not to do it now.

The cool thing with this change is that we can easily go back to strings. Simply create an alias from the enum type to string and remove the enum and you're done. 

Since the cost of undoing this change is very low but the benefits are clearly there (fixed a couple of bugs  in commits) and the additional complexity is very localized, I'd say let's go with this and remove if we deem so necessary when you may or may not convert compass to TS. Sounds good?

kanitw:
After you clean up, I think this is now much less confusing and I agree with you.

PR #829, opened 2015-12-04
Clean up

domoritz:
Clean up types and unused functions.


src/util.ts
```diff
@@ -30,13 +30,13 @@ export function reduce(obj, f, init, thisArg?) {
...
 
-export function map(obj, f, thisArg?) {
+export function map(obj, f: (a, d, k, o) => any, thisArg?) {
   if (obj.map) {
     return obj.map.call(thisArg, f);
   } else {
     var output = [];
     for (var k in obj) {
-      output.push( f.call(thisArg, obj[k], k, obj));
+      output.push(f(thisArg, obj[k], k, obj));
```
kanitw: `f.call`  is correct!  -- we bind `this` argument to the function

domoritz: Reverted. The change did not break any tests, though.

domoritz: All these functions are used in compass, right? Or can I delete them?

src/compiler/Model.ts
```diff
@@ -185,20 +185,20 @@ export class Model {
...
   }
 
-  numberFormat = function(channel?: Channel) {
+  numberFormat(channel?: Channel): string {
     // TODO(#497): have different number format based on numberType (discrete/continuous)
     return this.config('numberFormat');
   };
 
-  map(f) {
+  map(f: (...any) => any) {
```
kanitw: f should have same signature as vlEncoding.reduce ... same below

PR #830, opened 2015-12-04
Config 

### LINKED ISSUE #730: Add config for marks properties that are not mapped to encoding channels
## Easy 

- [x] name, description should be added to the top level of the specification?  (name will be particular useful when we support interaction)
- [x] properties
  - [x] general:  `fillOpacity`/`strokeOpacity`, `strokeDash`, `strokeDashOffset`, `strokeWidth`, …
  - [x] line/area: `interpolate`, `tension`
  - [x] area/bar: `orient`
  - [x] text: `align`, `baseline`, `dx`, `dy`, `radius`, `theta`, `angle`, `font`, `fontSize`, `fontWeight`, `fontStyle` 

- [x] support specifying group mark’s properties (e.g., background, border, etc.)

## Hard 
- [ ] How should we deal with key, delay, ease?

kanitw:
#730 

- Add top-level `name`, `description`
- Add config: `background`, `scene` and `viewport` properties and reorder these top-level properties in the Vega output
- Split the config schema files, add interfaces
- Add config for mark properties


src/compiler/compiler.ts
```diff
@@ -22,21 +22,24 @@ export function compile(spec, theme?) {
...
-               {field: layout.width.field} :
-               {value: layout.width},
-        height: layout.height.field ?
-                {field: layout.height.field} :
-                {value: layout.height}
+  let rootGroup:any = extend({
+      name: spec.name ? spec.name + '_root' : 'root',
+      type: 'group',
+    },
+    spec.description ? {description: spec.description} : {},
```
domoritz: can you add more let? This is hard to read

src/compiler/compiler.ts
```diff
@@ -70,14 +73,26 @@ export function compile(spec, theme?) {
...
-  var output = {
+  var output = extend(
+    spec.name ? {name: spec.name} : {},
+    {
       width: layout.width.field ? FIT : layout.width,
       height: layout.height.field ? FIT : layout.height,
-      padding: 'auto',
+      padding: 'auto'
+    },
+    ['viewport', 'background', 'scene'].reduce(function(topLevelConfig, property) {
```
domoritz: can you put that in another variable?

src/compiler/compiler.ts
kanitw: I disagree.  The goal of this chunk of code is to optimize order of the vega output. 
While stylistically a little uglier, putting it like this allows you to determine order of the output in one glance.

src/compiler/compiler.ts
kanitw: I disagree.  The goal of this chunk of code is to optimize order of the vega output. 
While stylistically a little uglier, putting it like this allows you to determine order of the output in one glance.

PR #831, opened 2015-12-04
rename `marktype` to `mark`

kanitw:
Please merge #830 first!


kanitw:
> move instead of delete + create
I thought clicking "rename" in atom would automatically do `git mv` since it's an editor by github.  
That said, is there an easy way to do that.  If there is, please help do it and let me know how for next time since I have to leave soon and we should release this morning.
Apologies!

Issue #431, opened 2015-05-13
Support adding title to a plot 


kanitw:
- [ ] ~~Need to add another text mark group to the top-level marks~~
- [ ] ~~The first step is to always add title to the bottom.~~  
- [ ] ~~Then we might add mode for adding title to the top (and thus offset the main mark group)~~

jowens:
I'd also like this capability.

kanitw:
We're punting on this to see if we should support this by composition instead.

kanitw:
Note: @pteehan asks for this in https://github.com/altair-viz/altair/issues/258.

kanitw:
Now that Vega has `title` directive for group mark, we can easily add this.

kanitw:
We can add `title` to `BaseSpec` with properties similar to https://vega.github.io/vega/docs/title/. 

(Maybe except `encode` for now.)

kanitw:
Fixed in #2779

PR #832, opened 2015-12-04
Example Gallery 

kanitw:
- extract examples to be in json files so we can copy to editor (or export via npm) easily!
- rename json files, conform description / title
- rewrite gallery with d3

@domoritz  — I haven’t modified script to export this with NPM, but if you think that’s the way to sync this to vega-lite editor.  Please go ahead and do it!

merge #831 first!


domoritz:
Fix tests :-P

examples/vlexamples.js
```diff
@@ -0,0 +1,94 @@
+'use strict';
+// jshint quotmark: false
+
+var VL_SPECS = {
+  'Basic': [
+    {
+      "name": "bar",
+      "title": "Simple Bar Chart"
```
domoritz: do we want to move title into the spec?

kanitw: That's a good question because I don't know if we should add "title" as top-level property.  
(Maybe we do, and let's support #431?)   That said, this might has some conflict with the vega-embed spec format.

PR #837, opened 2015-12-04
More lint rules

domoritz:


src/compiler/facet.ts
```diff
@@ -90,7 +90,7 @@ export function facetMixins(model: Model, marks) {
     // TODO: add properties to make rule optional
     rootMarks.push(getColumnRulesGroup(model, cellWidth));
   } else { // doesn't have column
-    if (model.has(Y)) { //keep y axis in the cell
+    if (model.has(Y)) { // keep y axis in the cell
```
kanitw: This rules looks really annoying unless there is a auto-fixer.

kanitw: I mean it's a good thing but I don't think it is very important.

domoritz: It's not but there were only 3 cases where this happened. I'm happy to
remove the rule.

src/compiler/marks.ts
```diff
@@ -31,7 +31,7 @@ export function compileMarks(model: Model): any[] {
       type: MARKTYPES_MAP[mark],
       from: {
         // from.data might be added later for non-facet, single group line/area
-        transform: [{type: 'sort', by: sortBy}]
+        transform: [{ type: 'sort', by: sortBy }]
```
kanitw: This rules is nice if computer automatically do for us, again if there is no nice way to fix this without screwing up other style (Running formatter for the whole file is generally risky), let's turn this off.

src/compiler/facet.ts
kanitw: Thinking more about this -- keep it then.

src/compiler/marks.ts
kanitw: (This one is more annoying than the other one) -- no need to revert the change you already made -- it looks nicer but I just don't want to enforce it.  :)

domoritz: I did this automatically. I just right clicked and said "beautify content". It's not enforced, I believe and I definitely don't want this enforced.

kanitw: Sounds good then!   

Btw, do you mean Atom's "Format Code"?

domoritz: screen shot 2015-12-04 at 14 54 34

PR #838, opened 2015-12-04
 add `npm run site` , rename `watch:all` to just `watch`

kanitw:
add `npm run site` , rename `watch:all` to just `watch` + update README.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,12 +19,13 @@
     "clean": "rm -f vega-lite.* vega-lite-schema.json & find src -name '*.js*' -type f -delete & find test -name '*.js*' -type f -delete",
     "deploy": "npm run clean && npm run lint && npm run test && scripts/deploy.sh",
     "lint": "tslint -c tslint.json src/**/*.ts test/**/*.ts",
-    "start": "npm run watch:build & browser-sync start --server --files 'vega-lite.js' --index 'gallery.html'",
     "schema": "tsc && node src/schema/schemagen.js > vega-lite-schema.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,12 +19,13 @@
     "clean": "rm -f vega-lite.* vega-lite-schema.json & find src -name '*.js*' -type f -delete & find test -name '*.js*' -type f -delete",
     "deploy": "npm run clean && npm run lint && npm run test && scripts/deploy.sh",
     "lint": "tslint -c tslint.json src/**/*.ts test/**/*.ts",
-    "start": "npm run watch:build & browser-sync start --server --files 'vega-lite.js' --index 'gallery.html'",
     "schema": "tsc && node src/schema/schemagen.js > vega-lite-schema.json",
+    "site": "bundle exec jekyll serve",
+    "start": "npm run watch & browser-sync start --server --files 'vega-lite.js' --index 'gallery.html'",
```
domoritz: Hmm, I'm not happy with this. If you use this and make a change in the code, it takes a long time before the gallery refreshes (10 seconds).

kanitw: Okay.  10 seconds sound terrible to me too.  Sounds like we should find a way to make this work. 
I don't think having two watchers is the ideal solution for longer term as well.

domoritz: Try it. I got annoyed and added the fast build command.

kanitw: no need to try -- 10 secs is ridiculous, I'm just saying that there must be a better set of tools to do this.

PR #843, opened 2015-12-05
Bin

kanitw:
- fixes #816 -- add missing bin parameter
- [x] Auto maxbins
- [x] docs
- [x] unit-tests

PR #848, opened 2015-12-06
Kw/examples

### LINKED ISSUE #835: Need a way to make vlexamples.ts work with both node and browser without error
Right now, we get 

```
Uncaught ReferenceError: exports is not defined
```

in the gallery!

kanitw:
- fix #835 - change vlexamples list to json
- also fix filter example

Please merge #843 first!  (See [Branch Specific Diff](https://github.com/uwdata/vega-lite/compare/kw/bin...kw/examples))


domoritz:
Can you remove angular from bower?

PR #853, opened 2015-12-06
Hide the binned scatterplot again since Vega hasn't updated…

kanitw:
- hide the binned scatterplot again since Vega hasn't released an update that uses datalib `1.5.7`.    
- remove datalib from bower dependency as datalib is embedded as vg.util

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -29,7 +29,6 @@
     "example/_diff/*"
   ],
   "dependencies": {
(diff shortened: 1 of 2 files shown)


kanitw:
@domoritz I’m gonna merge this and release so I can resolve issue in Compass.  Feel free to post follow-up comments.

PR #856, opened 2015-12-07
Add vl2png and vl2svg

### LINKED ISSUE #34: Script for generating image files
create cli for generating image file from spec

basically just wiring `vl2vg` and vega's `vg2png` and `vg2svg` should work

yhoonkim:
Fixes #34

### DIFF
--- a/bin/vl2png
+++ b/bin/vl2png
@@ -1,0 +1,66 @@
+#!/usr/bin/env node
+// Render a Vega specification to PNG, using node canvas
+
+var helpText =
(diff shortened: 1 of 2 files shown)


kanitw:
As a good practice, you should mention "fixes #34" and rename title to "Add vl2png and vl2svg".

This looks acceptable for using as standalone.  But as said in #34, it might be better to modify vg2png and vg2svg to support unix pipe and these two scripts just call them.   

That said, this comment is optional.  So if you do not want to do so, feel free to merge this.

PR #860, opened 2015-12-07
Update util import from datalib

kanitw:
- be more explicit about methods we import from datalib
- remove datalib/src/stats dependency


kanitw:
@domoritz I’m gonna merge this and release so I can resolve issue in Compass.  Feel free to post follow-up comments.

PR #866, opened 2015-12-12
Add name as prefix to mark group and scale’s name 

### LINKED ISSUE #819: Add name for a spec & use spec name as prefix for scale name.
Name should be a top-level property.  

This would be extremely important for composition.  (In fact, composition charts should automatically generate names for each subspace.)  

Currently scale name are just channel names, but with compositions - we can have multiple Xs and multiple Ys, etc.   Using each sub-spec’s name as prefix for scale would help distinguish different scale for the same channel of different subspecs.

kanitw:
See log for each commit for more details. 

Fixes #819

### DIFF
--- a/src/encoding.ts
+++ b/src/encoding.ts
@@ -12,18 +12,14 @@ export function countRetinal(encoding: Encoding) {
 }
 
 export function channels(encoding: Encoding) {
-  var arr = [];
-  CHANNELS.forEach(function(channel) {
-    if (has(encoding, channel)) {
-      arr.push(channel);
-    }
+  return CHANNELS.filter(function(channel) {
+    return has(encoding, channel);
   });
-  return arr;
 }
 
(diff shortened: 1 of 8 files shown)


domoritz:
Cannot merge.

kanitw:
Stacking PR like Google does will avoid this problem.... so there's a trade-off between the two approach.

kanitw:
Anyway, mergable now.

src/encoding.ts
```diff
@@ -11,6 +11,16 @@ export function countRetinal(encoding: Encoding) {
   return count;
 }
 
+export function channels(encoding: Encoding) {
+  var arr = [];
+  CHANNELS.forEach(function(channel) {
```
domoritz: Want to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter ?

PR #867, opened 2015-12-12
Auto drop unsupported channel from the spec 

### LINKED ISSUE #205: Using "text" channel with non-text mark is equivalent to details
If we fix this, we should fix #11 too.

kanitw:
fixes #205

### DIFF
--- a/src/compiler/Model.ts
+++ b/src/compiler/Model.ts
@@ -1,7 +1,7 @@
 import {Spec} from '../schema/schema';
 import {FieldDef} from '../schema/fielddef.schema';
 
-import {COLUMN, ROW, X, Y, COLOR, DETAIL, Channel} from '../channel';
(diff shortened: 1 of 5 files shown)


src/compiler/Model.ts
```diff
@@ -41,18 +41,25 @@ export class Model {
...
     this._spec = schemaUtil.merge(defaults, theme || {}, spec);
 
-    // convert short type to full type
+
     vlEncoding.forEach(this._spec.encoding, function(fieldDef: FieldDef, channel: Channel) {
+      if (!supportMark(channel, this._spec.mark)) {
+        // Drop unsupported channel
+
+        // FIXME consolidate warning method
+        console.warn(channel, 'dropped as it is incompatible with', this._spec.mark);
```
domoritz: Yep, we should use the warning and error outputs soon.

PR #872, opened 2015-12-15
Use vega embed in gallery

domoritz:


site/gallery.js
```diff
@@ -21,19 +21,24 @@ d3.json('examples/vlexamples.json', function(VL_SPECS) {
...
-        });
-        view.update();
+      var embedSpec = {
+        mode: 'vega-lite',
+        spec: vlSpec,
+        actions: {
+          export: false
+        }
+      };
+      vg.embed('.viz#'+ example.name + '> div.view', embedSpec, function(view, vega_spec) {
```
kanitw: According to Jeff message in slack, `vega_spec` here is an object with both `view` and spec`?  Also, why underscore rather than camelCase?

site/gallery.js
```diff
@@ -21,19 +21,24 @@ d3.json('examples/vlexamples.json', function(VL_SPECS) {
...
-        var view = chart({
-          el: d3.select('.viz#'+ example.name + '> div.view').node(),
-          renderer: 'svg'
-        });
-        view.update();
+      var embedSpec = {
+        mode: 'vega-lite',
+        spec: vlSpec,
+        actions: {
+          export: false
```
kanitw: What is `export: false`?  — This seems undocumented in Vega as well?

site/gallery.js
domoritz: Just copied. Also, we don't actually use it.

site/gallery.js
domoritz: Look at the actions in the code. It is only partially documented. I'll update the docs.

domoritz: https://github.com/vega/vega/wiki/Embed-Vega-Web-Components#configuration-propeties

kanitw:
I got this error in the gallery 
```
[Vega Err] TypeError: (spec || []).forEach is not a function
```

Based on my inspection, somehow vega-lite spec, instead of vega spec, is parsed through `vg.parse.spec`.

site/gallery.js
domoritz: Fixed after merge

domoritz:
The error was because you still use vega-embed 1.x. Fixed.

kanitw:
Umm, that version change should be in be pr though :p

domoritz:
Well, you merged it. Vega-embed was released after I fixed this branch and you merged it before I got the chance to fix it, lol.

kanitw:
Hmm, then it should me marked WIP -- and note that depends on pending
release? Lol :p

PR #882, opened 2015-12-19
Clean up npm

domoritz:


.npmignore
```diff
@@ -1,11 +1,22 @@
 node_modules
 bower_components
 data
-examples/_original
-examples/_output
-examples/_diff
-index.html
-script
+docs
+examples
```
kanitw: If we were to import examples to vl-editor, what I did with examples is already correct.

domoritz: Good point. Fixed.

domoritz:
Look at what gets bundled with `tar -tf $(npm pack)`.

PR #886, opened 2015-12-20
make `config.marks.filled` applied to other applicable mark types

### LINKED ISSUE #812: Make config.marks.filled work with other mark types
(Currently it only works for `point`)

kanitw:
- make `config.marks.filled` applied to other applicable mark types (`bar` and `area`) 
- consolidate rules for determining default config values. 

fixes #812.

### DIFF
--- a/src/compiler/Model.ts
+++ b/src/compiler/Model.ts
@@ -248,29 +248,35 @@ export class Model {
     return vals && vals.length;
   }
 
-  public config(name: string) {
-    return this._spec.config[name];
+  public config(name: string, prop?: string) {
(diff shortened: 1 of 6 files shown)


src/compiler/Model.ts
```diff
@@ -248,8 +248,30 @@ export class Model {
     return vals && vals.length;
   }
 
-  public config(name: string) {
-    return this._spec.config[name];
+  public config(name: string, prop?: string) {
```
domoritz: so this function returns bool, number, or value? Maybe document what it does.

PR #891, opened 2015-12-21
Remove time.ts

kanitw:
time.ts is kinda of weird at this point. 

I’m moving:
- `time.formula` to `data.ts`
- `time.scale` to `scale.ts`
- `time.cardinality` to be part of `fielder.cardinality`
- `time.labelTemplate` => `model.labelTemplate` + add missing `legend.shortTimeLabels`


kanitw:
I’ll just merge since @domoritz is in Japan and this PR doesn’t change logic —  only refactoring / moving things around.

domoritz:
LGTM

Issue #771, opened 2015-11-22
Allow hiding axis (axis: false)


kanitw:
Sometimes, hiding axis just make things more beautiful. 
For example, see http://bl.ocks.org/mbostock/3887051.

domoritz:
You can hide legends. So we only need to add support to hide axes.

PR #893, opened 2015-12-24
labels:true|false

domoritz:


kanitw:
Why do we need this?   (When do we want axis without labels?)  

Also, if this is really needed, it’s better done at the vega level?  (Do not producing label objects in the scenegraph at all instead of producing a group of blank labels.

domoritz:
> Why do we need this? (When do we want axis without labels?)

If you are over encoding. For example in the grouped bar chart example. 

> Also, if this is really needed, it’s better done at the vega level? (Do not producing label objects in the scenegraph at all instead of producing a group of blank labels.

Yes, I created https://github.com/vega/vega/issues/480. It would be more semantically meaningful but we still need some option in vl that we can pass down to vega. For now, we can compile to `text: ""`.

kanitw:
>  grouped bar chart example
Good point!  
1. We should then modify group bar chart examples to use this property.
2. I guess you would want to hide ticks as well.  Hiding axis labels without hiding ticks is weird. 
3. Should we set text to "" or set opacity to 0?   I use the latter method for hiding ticks and axis lines for ROW and COLUMN.

domoritz:
> 1. We should then modify group bar chart examples to use this property.

Will do

> 2. I guess you would want to hide ticks as well. Hiding axis labels without hiding ticks is weird. 

Good point. I'll fix that as well. 

> 3. Should we set text to "" or set opacity to 0? I use the latter method for hiding ticks and axis lines for ROW and COLUMN.

Nether of them are semantically correct. I prefer setting the text to "" because this way it implies that no space has to be reserved for the axis label.

kanitw:
I tried hiding both labels and tick and turns out to need to hide axis title too.   

vega_editor

After hiding axis title, it looks like this:

vega_editor

which makes me wonder if the thing we really need is simply hiding axis (#771)

vega_editor

domoritz:
I think you want the baseline. I did the same steps in https://github.com/vega/vega-lite/commit/537630b6dfa1d7b0e85ccdc8c482e2bf13e93d8e

kanitw:
Maybe, but ideally I would want a single, long baseline.. which is something more like this:

vega_editor

Or even better without the rules:

kanitw:
(and more padding) ...

vega_editor

domoritz:
I can add an option to remove the axis entirely. But let's merge as is for now because removing the axis requires rewriting the instantiate.

kanitw:
This branch has conflicts that must be resolved.

PR #895, opened 2015-12-25
Move characterWidth to axis

kanitw:


docs/encoding.md
```diff
@@ -219,11 +219,10 @@ Moreover, Vega-Lite supports the following additional axis properties.
 
 | Property        | Type          | Description    |
 | :------------   |:-------------:| :------------- |
+| characterWidth  | Integer       | Character width for automatically determining title max length. |
```
domoritz: title max length -> max title length

Issue #64, opened 2014-12-16
Supporting labeling for all basic charts


kanitw:
Just always put numbers outside like I add 28, 55 below (same for all bars — I’m just adding this by hand so I added only the first two.)

![vega_editor](https://cloud.githubusercontent.com/assets/111269/11292824/01a2310c-8f08-11e5-8c0a-53b613c63c11.png)
- [ ] try extending vega output for the “bar” example with another marks group for showing the label with some offset 
- [ ] modify the code (more details soon)
  - [ ] add `label` encoding channel to `channel.ts` and the schema
  - [ ] extend `marks.ts`

kanitw:
@willium   I actually start to think that this one might be a better to begin with.  :)  
Start with trying to see how to extend Vega spec first, more details about where do you have to extend the code will follow.  We can chat more tomorrow evening or next week if you’re around as well.

domoritz:
@kanitw just wondering what should happen if the bar goes all the way to the top. If the label is outside, we may have to change the layout.

kanitw:
@domoritz  That's a good corner case, but let will begin with putting it on the top / on the right first.  

- [ ] when axis `orient` is `top` or `right`
- [ ] maybe worth adding an option for location of the label (but we need to think very hard about how this works for other mark type as well)

willium:
@kanitw seems relatively straight forward -- To clarify, this is different than a tooltip in that you want the text to always show above __all__ bars  (or below, or on the sides depending orient) without hover. Essentially an extension of https://github.com/vega/vega/wiki/Tutorial#bar-chart-specification and the code that makes that work. 
So:
1) implement it in vega
2) write the vega-lite for it

@domoritz 
This text overflow case seems to be broken in the vega example https://github.com/vega/vega/wiki/Tutorial#bar-chart-specification as well 
![overflow text](https://cloud.githubusercontent.com/assets/356976/11318241/5281acf0-8ffe-11e5-9410-a0a6a49219cd.png)

kanitw:
@willium yep. You're right.  It's basically should be the same as "label" shelf in Tableau. 

I think currently Vega does not have a way to add additional offset to the scale when drawing domain from data, but this feature might come in handy in this case and #374
By this I mean, if the data ranges from [0, 100] if we can add property like `domain.upperBoundOffset` and thus main the scale's domain become [0, 120].  That might be useful.   But since we don't have it yet, let's not worry about it yet.

willium:
[...]

PR #897, opened 2015-12-26
[WIP] Bar Label

kanitw:
@willium let’s create a feature branch “1.x-label” so that we can merge label related stuff into this branch since we’re not sure if all labels will be done by the 1.0 release.


src/compiler/marks.ts
```diff
@@ -357,8 +357,53 @@ export namespace bar {
...
+    // - color support
+    // - container height/stretching consideration
+    // - consider defaults for offset
+
+    var l: any = {};
+    const orient = model.marksConfig('orient');
+
+    l.align = { value: (orient !== 'horizontal') ? "center" : "left" };
+
+    l.text = { field: model.has(LABEL) ? model.field(LABEL) : ((orient !== 'horizontal') ? model.field(Y) : model.field(X)) };
```
kanitw: if there is not `label` channel, this function would not be evaluated at all.   

See [this line](https://github.com/vega/vega-lite/blob/master/src/compiler/marks.ts#L104)

willium: right, I meant to try to pick up on whether or not the VL explicitly assigns a __field__ (see below) to the label, but you're right the way I wrote it is redundant. 

     "encoding": {
                "x": {"type": "quantitative", "field": "amount"},
                "y": {"type": "ordinal", "field": "category"},
                "label": {
                  "field": "amount",
                  "type": "quantitative"
                }
            }

willium:
I'm curious if you could point me to where I could pick up the model's colors (bg, where steelblue comes from, axes). I figure the labels are better default colored to the axes, though I imagine there might be an algorithm for finding the best color (you probably already have it implemented somewhere)

There are likely some bigger conversations to be had or that you've already had, but It'd be cool to work on some of the other labels and variations. 
Some thoughts I had:
- inset (negatively colored) labels could be cool, but would cause problems with stacked bars (unless there was a label inset for each stacked bar, but then you'd have a width/height limitation
- the fix you mentioned in your (comment about intelligently scaling the bounds)[https://github.com/vega/vega-lite/issues/64#issuecomment-158663494] probably wouldn't work for relative bar charts once they're maxed out to 100%
- it'd be cool to have a short hand for specific points in scatterplots: outliers, critical points and summary numbers, or intersections with line charts, but all these labels would probably look different and be more complicated than just a floating annotation

src/compiler/marks.ts
kanitw: If you look at `model.has()` in `Model.ts` -- `has` actually check existence of the field.   So we can be sure about that. 

(I start wondering if we should rename `has` to `hasField`..)

willium: ah you're right. I should've looked more closely. I wonder if there's a case for wanting to apply a label without being explicit with field. In the case of a bar chart with a quantitative variable and a ordinal/categorical variable, I can't imagine the label would be of anything but the quantitative var. Either way, totally makes sense, I'll remove that conditional.

kanitw:
> where I could pick up the model's colors (bg, where steelblue comes from, axes).

It's actually very simple.   If color channel is not mapped to a field, we draw value from `encoding.color.value`.  If `encoding.color.value` is not explicitly specified, we use the default value from [this line](https://github.com/vega/vega-lite/blob/master/src/schema/encoding.schema.ts#L74).  (`vl.schema.instantiate` uses `default` property in the schema to set default values for properties.) 


> - inset (negatively colored) labels could be cool, but would cause problems with stacked bars (unless there was a label inset for each stacked bar, but then you'd have a width/height limitation)

This is definitely a good thing to do in the future since there are many corner cases like you said. 
I think it's worth do labels for other mark types first before introducing advanced features for one mark type (bar).   

Other mark types will have complicated issues too, but maybe we can start with basic supports for all mark types before upgrading rules for each specific ones.  For example, scatterpoint's labels can lead to a lot of occlusion, but we might assume that the first version we are okay with occlusion.  In a later version, we can then introduce a Vega transform for hiding occluded labels.  

> the fix you mentioned in your [comment about intelligently scaling the bounds](https://github.com/vega/vega-lite/issues/64#issuecomment-158663494) probably wouldn't work for relative bar charts once they're maxed out to 100%

Good point.  In that case, we may just add offset and hope that Vega will support [padding at the group level soon](https://github.com/vega/vega/issues/472).   Once we have inset labels, we may be have relative bar charts use inset labels by default.  

> - it'd be cool to have a short hand for specific points in scatterplots: outliers, critical points and summary numbers, or intersections with line charts, but all these labels would probably look different and be more complicated than just a floating annotation

Yeah, these would be useful in the future, but is much more complicated to do now.  We have to implement Vega transforms for detecting outliers before we can highlight/visualize them.  

>  intersections with line charts
(I'm not sure what you mean "intersections with line charts" though)

kanitw:
FYI, I just added more comment about axis’s color in the comment above.

willium:
- [x] I fixed the conditional

Because the axis color is set by Vega, how should we handle the coloring of the label? I'm not sure if I can piggy back on some other logic (the grid perhaps?), or should write something more custom. Not sure what the desired behavior is either for labels.

I hardcoded in offsets of 6 because its the same as the grid offset, but wasn't sure if I should directly pull from that or not... Also, for the horizontal bars, the __l.dy__'s _mult_ and _band_ parameters don't seem to do anything. If I hardcode in half the grid offset as a y offset, it does the trick but -- I'm not sure we can be sure that the bar's widths are only 6, especially in such a hardcoded way. Is there something to _mult_ and _band_ not working that I should realize, or is that a bug? I'm not sure what numbers I can rely on in this case.

kanitw:
> Because the axis color is set by Vega, how should we handle the coloring of the label?

This is marks label, so we shouldn't worry too much about axis label.   That said, I think having label in black by default is a reasonable option.  We can then provide a set of label configs ( `config.label.<property>` for customizing marks property of the label marks and other label options).  
And maybe add a special `config.label.sameColorAsMark` to make the label use the same color as mark (no matter it's from `encoding.color.value` or `encoding.color.field`)

I just send a PR to provide scaffolding for `config.label`.  

> I hardcoded in offsets of 6
Eventually we might add `config.label.offset` as part of `config.label`.  
we might try to add label for other mark type first -- just to see if they can share same set of config properties? 

> I hardcoded in offsets of 6 because its the same as the grid offset, but wasn't sure if I should directly pull from that or not.

I'm changing gridOffset to 0 and they are different things -- so definitely not! 

>  l.dy's mult and band parameters don't seem to do anything
Maybe @arvind  will know better about this.

src/compiler/marks.ts
```diff
@@ -357,8 +357,48 @@ export namespace bar {
...
+      l.y.offset = 3;
+      l.dy = {
+        scale: model.scale(Y),
+        band: true,
+        mult: 0.5
+      }
+    }
+
+    // fill
+    extend(l, colorMixins(model));
```
willium: `applyColorAndOpacity(l, model)`

src/compiler/marks.ts
```diff
@@ -357,8 +357,48 @@ export namespace bar {
...
 
   export function labels(model: Model) {
-    // TODO(#64): fill this method
-    return undefined;
+    // TODO(#64):
+    // - adjust for container bounds
+    // - consider defaults for offset and color
+
+    var l: any = {};
+    const orient = model.marksConfig('orient');
```
willium: model.config().mark.orient

kanitw:
@willium 

A few notes from today’s discussion 
- VL’s bar use `xc` and `width` instead of  `x` and `x2` with offset.  
- We talked about reactive geometry, which means you can drive mark with another mark like shown in the [text group in the group bar example](http://vega.github.io/vega-editor/?mode=vega&spec=grouped_bar).    There is no doc for reactive geometry yet but the note in https://github.com/vega/vega/issues/497 should be useful.

PR #898, opened 2015-12-27
add config.label scaffold

kanitw:
@willium 
I quickly added config.label scaffold, which would be useful for your `bar-label` branch.

- Add JSON schema and TS interface for `config.label`
- Add model.labelConfig and `applyLabelConfig()` inside `compiler/marks.ts`.  `applyLabelConfig()` is not used yet but should be used in a similar fashion to `applyMarksConfig()`.  

For list of label properties, I just copied it directly from `config.marks.schema.ts` and add `fontSize`.  (And since I’m doing this really quick I might have overlooked other important properties that should be included as well.


src/compiler/Model.ts
```diff
@@ -246,6 +246,13 @@ export class Model {
   }
 
   /**
+   * @return Label config value from the spec, or a default value if unspecified.
+   */
+  public labelConfig(name: string) {
```
kanitw: Remove this!

src/compiler/marks.ts
```diff
@@ -168,6 +168,16 @@ function applyMarksConfig(marksProperties, model: Model, propsList: string[]) {
   });
 }
 
+// TODO: use this in label methods
+function applyLabelConfig(marksProperties, model: Model, propsList: string[]) {
+  propsList.forEach(function(property) {
+    const value = model.labelConfig(property);
```
willium: `model.config().label[property]`

src/schema/config.label.schema.ts
```diff
@@ -0,0 +1,125 @@
...
+    },
+    strokeDashOffset: {
+      type: 'array',
+      default: undefined,
+      description: 'The offset (in pixels) into which to begin drawing with the stroke dash array.'
+    },
+    // text-only
+    align: {
+      type: 'string',
+      default: 'right',
```
kanitw: @willium you might wanna change this default to `undefined` since you have a rule for determining it.

kanitw:
@willium  This should be now ready for merging into your branch.  Note that I now call the `applyLabelConfig()` inside your `label()` method. See 94d0e3a2c5915b0fff333248028a0eb64a307053.

PR #906, opened 2016-01-10
[WIP] extract shorthand from Vega-Lite

kanitw:
move to https://github.com/vega/vega-lite-shorthand

Need to finish https://github.com/vega/vega-lite-shorthand/pull/1 first!


domoritz:
Probably need to just do this one again.

PR #914, opened 2016-01-11
Refactor STACKED_SCALE (originally STACKED)

kanitw:
- Rename STACKED to STACKED_SCALE to avoid confusion since the data source sums data for scale while the actual stack layout is applied inline.

- Remove the unnecessary max calculation in STACKED_SCALE since scale would automatically calculate max anyway.  

@arvind @domoritz


kanitw:
Just resolved conflicts… now mergable

PR #923, opened 2016-01-12
Refactor to avoid overloading the term spec / def.

kanitw:
- `model.axisDef` => `model.axis()`
- spec always refers to input
- def generally refers to output except `fieldDef`, which is too late to change.

- part of #276


domoritz:
`fieldSpec`, lol

PR #925, opened 2016-01-12
Add null filter and refactor config

domoritz:


src/compiler/compiler.ts
```diff
@@ -32,15 +32,15 @@ export function compile(spec, theme?) {
       padding: 'auto'
     },
     ['viewport', 'background'].reduce(function(topLevelConfig, property) {
-      const value = model.config(property);
+      const value = model.config()[property];
```
kanitw: should `model.config()` and `model.sceneConfig()` be cached in consts above?  

(No worry if you prefer this)

domoritz: Nah, it's only used in one line.

src/compiler/data.ts
```diff
@@ -175,9 +182,10 @@ export namespace source {
    * @return An array that might contain a filter transform for filtering null value based on filterNul config
    */
   export function nullFilterTransform(model: Model) {
-    const filterNull = model.config('filterNull');
+    const filterNull = model.config().filterNull;
     const filteredFields = keys(model.reduce(function(aggregator, fieldDef: FieldDef) {
-      if (fieldDef.field && fieldDef.field !== '*' && filterNull[fieldDef.type]) {
+      if (filterNull === true ||
+        filterNull !== false && fieldDef.field && fieldDef.field !== '*' && DEFAULT_NULL_FILTERS[fieldDef.type]) {
```
kanitw: - `filterNull === true` => `filterNull` ?
- `filterNull !== false` => `filterNull === undefined` – clearer
- Please add parentheses for readability!

src/schema/config.schema.ts
```diff
@@ -57,13 +57,9 @@ export const config = {
...
     // TODO(#597) revise this config
     filterNull: {
-      type: 'object',
-      properties: {
-        nominal: {type:'boolean', default: false},
-        ordinal: {type:'boolean', default: false},
-        quantitative: {type:'boolean', default: true},
-        temporal: {type:'boolean', default: true}
-      }
+      type: 'boolean',
```
kanitw: Please copy this to `config.md`

src/schema/config.schema.ts
```diff
@@ -57,13 +57,9 @@ export const config = {
     // filter null
     // TODO(#597) revise this config
```
kanitw: remove this line.

src/compiler/data.ts
domoritz: Hmm, I wonder whether 'auto' should also work. That's why I wrote it the way it is. But we should be consistent.

kanitw:
Please also modify VLUI’s `toggleNull` button in `vlplot.js` based on this change.

src/compiler/compiler.ts
kanitw: Fine

src/compiler/data.ts
kanitw: Nope.  This is boolean, `'auto'` is not boolean.

domoritz: This is javascript. Anything can happen :-P

kanitw: It should pass the schema validator from the beginning once we have it in place.

kanitw:
LGTM, feel free merge once you solve all issue and add VLUI fixes

src/compiler/Model.ts
```diff
@@ -232,17 +232,17 @@ export class Model {
...
+    return this._spec.config;
   }
 
   /**
-   * @return Cell config value from the spec, or a default value if unspecified.
+   * Get the spec configuration for cells.
    */
-  public cellConfig(name: string) {
-    return this._spec.config.cell[name];
+  public cellConfig() {
```
kanitw: Actually, should we just remove `scene|cellConfig()` and use `config().scene|cell` ?

domoritz: Thought about it as well. Yes.

Issue #145, opened 2015-01-24
Extensive Time Support


kanitw:
in #48  we support basic time binning  (`second`, `minute`, `hour`, `dayofweek`, `date`, `month`, year`)

- [ ] add century, decade, quarter

- [x] other time “flooring” function — we need to implement floor instead of transforming to ordinal in order to utilize the time scale.   And thus, we should have the following functions:

- `yearmonth` or `%y-%m` in [`d3.time.format()`](https://github.com/mbostock/d3/wiki/Time-Formatting)
- `yearmonthday` or `%y-%m-%d` 
- `yearmonthdayhour` or `%y-%m-%d-%H` 
- `yearmonthdayhourminute` or  `%y-%m-%d-%H-%M` 
- `hourminute` or  `%H-%M` 
- `hourminutesecond` or  `%H-%S`

Note that using the the format like above would simplify sorting order of the data.  (For labeling, we can use %b instead of %m but it would make sorting more difficult)

Even better, we could support
- [ ]  using custom label on the axis (later)

——

In any case, make sure we update
- [ ] `axisTitleOffset()` to provide correct estimator for length of labels

- [ ] `scale.type` add add default timeUnit

domoritz:
Support mapping numbers to day of week string as in examples from @depstein. 

`0` -> `Monday` ...

depstein:
Unsure if you want to support this, but I map hours to 12-hour time as well.

`14` -> `2 PM` (`%I %p`)

domoritz:
@depstein thank for the suggestion. The challenge there is that you want to parse the number before you use a formatter (like https://github.com/mbostock/d3/wiki/Time-Formatting). Unfortunately, we won't support parsing arbitrary formats a la http://momentjs.com/docs/#/parsing/ soon. We have special functions for month and day of week names (https://github.com/vega/datalib/blob/master/test/template.test.js#L159), though and I might add support for times at some point.

domoritz:
I think I'm wrong. At least d3 supports parsing with formats (https://github.com/mbostock/d3/wiki/Time-Formatting#format) so we should be able to add that easily.

depstein:
That would be great! I've worked around it for the time being, but it sounds useful long-term. Thanks for the good work.

domoritz:
I should see what works now.

kanitw:
@domoritz bump

kanitw:
Seems like there are 3 cases for timeUnit scales:
[...]

PR #937, opened 2016-01-14
More time units

domoritz:
Addresses parts of #145

* Adds additional time units such as `yearmonth`, `yearmonthday`, `hoursminutes`, ...
* Use timestamps internally for temporal data even if the data is being aggregated.
* When time unit is `month`, `day`, `hours`, `minutes`, or `seconds`, vl automatically adds missing points to the domain. Previously the domain was set directly. Now, vl creates a new data source. 
* Allow customization of time format.

### DIFF
--- a/src/compiler/axis.ts
+++ b/src/compiler/axis.ts
@@ -17,10 +17,13 @@ export function compileAxis(channel: Channel, model: Model) {
     scale: model.scale(channel)
   };
 
-  // 1. Add properties
+  // 1.1 Add properties with special rules
+  model.format(channel, model.axis(channel).format, def);
+
+  // 1.2. Add properties
   [
     // a) properties with special rules (so it has axis[property] methods) -- call rule functions
-    'format', 'grid', 'layer', 'orient', 'tickSize', 'ticks', 'title',
+    'grid', 'layer', 'orient', 'tickSize', 'ticks', 'title',
     // b) properties without rules, only produce default values in the schema, or explicit value if specified
     'offset', 'tickPadding', 'tickSize', 'tickSizeMajor', 'tickSizeMinor', 'tickSizeEnd',
     'titleOffset', 'values', 'subdivide'
@@ -40,7 +43,7 @@ export function compileAxis(channel: Channel, model: Model) {
   var props = model.axis(channel).properties || {};
 
   [
-    'axis', 'labels',// have special rules
+    'axis', 'labels', // have special rules
     'grid', 'title', 'ticks', 'majorTicks', 'minorTicks' // only default values
   ].forEach(function(group) {
     var value = properties[group] ?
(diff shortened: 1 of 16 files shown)


kanitw:
Once no longer WIP, please better describe what's now supported since #145 might not mention what exactly you support here.  (So we can copy the summary to release note :D)

src/compiler/data.ts
```diff
@@ -52,10 +51,16 @@ export function compileData(model: Model): VgData[] {
     def.push(stack.def(model, stackDef));
   }
 
+  // Time domain tables
+
+  dates.defs(model).forEach(dateDef => {
```
kanitw: For multiple line function, should we still use classic function style for consistency? 

(I found this shorter format super useful to do one-liner lambda like in python though)

src/compiler/scale.ts
```diff
@@ -67,7 +68,6 @@ export function type(fieldDef: FieldDef, channel: Channel, mark: Mark): string {
       if (fieldDef.scale.type !== undefined) {
         return fieldDef.scale.type;
       }
-      // TODO: add timeUnit for other timeUnit once added
       switch (fieldDef.timeUnit) {
```
kanitw: better put switch inside the `if(fieldDef.timeUnit)` below and `return 'ordinal';` for `default:` case (yearmonth, monthday, ...)

domoritz:
Done. Addressed your comments and added changelog.

src/compiler/data.ts
kanitw: ping

src/compiler/time.ts
```diff
@@ -1,11 +1,134 @@
...
+  if (timeUnit.indexOf('year') > -1) {
+    out += get('year');
+  } else {
+    out += '2006, '; // January 1 2006 is a Sunday
+  }
+
+  if (timeUnit.indexOf('month') > -1) {
+    out += get('month');
+  } else {
+    out += '0, ';
```
kanitw: // month starts at 0 in javascript

src/compiler/data.ts
```diff
@@ -410,6 +415,37 @@ export namespace stack {
   };
 }
 
+export namespace dates {
+  /**
+   * Add data source for with dates for all months, days, hours, ... as needed.
+   */
+  export function defs(model: Model) {
+    let defs = [];
```
kanitw: You never change reference to this array, so this should be `const`.
Another note is that we commonly use `reduce` for this kind of thing elsewhere in this codebase.  Better be consistent.

domoritz: I am pushing to the array. Anyway, changed to reduce.

kanitw: I know that but if you push to the array, the reference to the array doesn't change.   Doesn't matter anymore anyway.

src/compiler/axis.ts
```diff
@@ -187,13 +170,6 @@ export namespace properties {
       }, labelsSpec);
     }
 
-    let filterName = model.labelTemplate(channel);
```
kanitw: What if the case where scaleType is ordinal that you mentioned last night?

PR #938, opened 2016-01-14
Refactor Test

### LINKED ISSUE #936: Split the validate test from unit test
Currently, we have example validation test (validate input VL spec and output VG spec) in `test/examples.test.ts` and other unit tests under `test/`.  

Having both kind of test is great. However, combining validation test in the same infrastructure as unit test gives us misleading unit test coverage.

We should either (a) call them separately or (b) make `npm run cover` run only non-validation test.

kanitw:
- Add `parseModel()` to `model.ts` and prepend `/* tslint:disable:quotemark */` to all files under `test/compiler` so that we can use pure JSON spec (rather than requiring enums) to instantiate model and thus facilitate easier unit-test writing.  (Just copy spec, and allow copy spec to editor for easy debugging.)
- Replace `test.fixtures.ts` with locally placed fixtures and fixtureFactory where applicable
- separate example validation test from unit tests to avoid overly high test coverage - fixes #936


test/compiler/marks.test.ts
```diff
@@ -32,9 +45,12 @@ describe('compile.marks', function() {
     });
 
     describe('1D, vertical', function() {
-      var f = bars['1d_ver'],
-          e = new Model(f),
-          def = marks.bar.properties(e);
+      const e = parseModel({
+          "mark": "bar",
+          "encoding": {"y": {"type": "quantitative", "field": 'US_Gross', "aggregate": "sum"}},
+          "data": {"url": 'data/movies.json'}
```
domoritz: Use "

test/compiler/marks.test.ts
```diff
@@ -20,9 +27,15 @@ describe('compile.marks', function() {
...
 
     describe('horizontal, with log', function() {
-      var f = bars.log_hor,
-          e = new Model(f),
-          def = marks.bar.properties(e);
+      const e = parseModel({
+        "mark": "bar",
+        "encoding": {
+          "y": {"bin": true, "type": "quantitative", "field": "IMDB_Rating"},
+          "x": {"scale": {"type": 'log'}, "type": "quantitative", "field": 'US_Gross', "aggregate": "mean"}
```
domoritz: use "

src/compiler/Model.ts
```diff
@@ -22,9 +22,16 @@ import {compileStackProperties, StackProperties} from './stack';
 import {type as scaleType} from './scale';
 
 /**
- * Internal model of Vega-Lite specification for the compiler.
+ * Call new Model without worrying about type.
+ * We use this in test to allow using raw JSON. 
  */
+export function parseModel(spec) {
```
domoritz: Move to tests. We shouldn't export this in the public API.

.gitignore
```diff
@@ -15,5 +15,6 @@ test/log/difflist.json
 examples/_original/*
 examples/_output/*
 examples/_diff/*
+examples/*.test.js
```
domoritz: not needed

PR #944, opened 2016-01-15
Revise cell config

### LINKED ISSUE #822: heatmap is much less dense in color than it used to be
https://gist.github.com/jowens/04e37346513ac9b8570e

vega_scaffold

For a heatmap, all that extra padding does little good; I think you'd like much more color than white space. Previous VL builds were mostly color, not so much white space. Now it's much more whitespacey.

### LINKED ISSUE #917: Consider if cell.padding redundant with scale.padding

### LINKED ISSUE #943: Add properties for hiding row/column’s grid group
For example, the [trellis_barley example](http://vega.github.io/vega-editor/?mode=vega-lite&spec=trellis_barley&showEditor=1) would produce a group of rule marks called `row-grid`.  

Currently, there is no way to hide this.  

Maybe we never apply row/column’s `axis.grid` to the output `axis.grid`, but apply it for hiding this instead?

kanitw:
- Apply `axis.grid` of `row` and `column` to our customized grid built by a rule mark group instead of passing it to the row’s scale. Also remove the redundant `config.cell.gridShow`.  Fixes #943
- Apply `scale.padding` of `row` and `column` to our cell padding layout calculation & remove the redundant `config.cell.padding`. Fixes #917
- Set ROW and COLUMN’s padding to 0 if Y and X are not mapped respectively. Fixes #822

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -38,7 +38,6 @@ the following grid properties:
 
 | Property      | Type          | Description    |
 | :------------ |:-------------:| :------------- |
-| gridShow      | Boolean       | Whether to show facet grid.  (True by default)  |
 | gridColor     | Color         | Color of the grid between facets. |
 | gridOpacity   | Number        | Opacity of the grid between facets. |
 | gridOffset    | Number        | Offset for grid between facets.  |
--- a/src/compile/Model.ts
+++ b/src/compile/Model.ts
@@ -52,6 +52,14 @@ export class Model {
       if (fieldDef.axis === true) {
         fieldDef.axis = instantiate(axisSchema);
       }
+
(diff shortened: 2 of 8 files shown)


docs/config.md
```diff
@@ -38,7 +38,6 @@ the following grid properties:
 
 | Property      | Type          | Description    |
 | :------------ |:-------------:| :------------- |
-| gridShow      | Boolean       | Whether to show facet grid.  (True by default)  |
```
domoritz: "(grid shown by default)"

kanitw: I removed this, so why are you worrying about a removed property?

domoritz: Lol, ups.

PR #946, opened 2016-01-15
Revise rules for text 

### LINKED ISSUE #724: Revise rules for marks
- [x] __text__: 
  - [x] `x` and `y` should work with `text`.   Currently text is only used with `row` and `col`.   — Related to #822, #741
   - [x] Property for applying color to background?
   - [x] text also current has no fill applied
 - [x] remove `textCellWidth`
- [x] __bar, area__: orientation
- [x] behavior of __size__ for different marks 
- [x] ~~square with OxO~~
- [x] `bar`’s scale (see note below)

#205, #185

kanitw:
* make text support X, Y
* center align axis when x exists
* fix missing text fill color

Fixes #724


kanitw:
Ready for review, with 100% statement test coverage.  :)

src/compile/mark-text.ts
```diff
@@ -58,13 +59,16 @@ export namespace text {
...
     }
 
-    // FIXME applyColorAndOpacity
-    // fill
-    // TODO: consider if color should just map to fill instead?
+    if (model.config().mark.applyColorToBackground && !model.has(X) && !model.has(Y)) {
+      p.fill = {value: 'black'}; // TODO: add rules for swapping between black and white
+
+      // opacity
+      var opacity = model.config().mark.opacity;
```
domoritz: const

Issue #930, opened 2016-01-12
Make Vega-Lite Gallery work in Safari


kanitw:

domoritz:
Chrome 47 has similar problems. We suspect that the issue is non-prefixed flexbox attributes.

PR #954, opened 2016-01-16
Fix gallery flexbox in Safari and Chrome

willium:
Quick fix for #930. 

Tested in:
- [x] Safari Version 9.0.2 (11601.3.9)
- [x] Chrome Version 47.0.2526.111
- [x] Firefox 33.1.1
- [x] Firefox 43.0.4

### DIFF
--- a/site/gallery.css
+++ b/site/gallery.css
@@ -1,88 +1,88 @@
 body {
-    margin: 20px;
-    padding: 15px;
-    font-family: Avenir, Helvetica Neue, Helvetica, Arial;
+  margin: 20px;
+  padding: 15px;
+  font-family: Avenir, Helvetica Neue, Helvetica, Arial;
 }
 
 strong {
-    font-weight: 400;
+  font-weight: 400;
 }
 
 a {
-    color: #8A5ED3;
-    text-decoration: none;
-    font-weight: 500;
+  color: #8A5ED3;
+  text-decoration: none;
+  font-weight: 500;
 }
 
 a:hover {
-    color: #d5a928;
+  color: #d5a928;
 }
 
 .logo {
-    font-weight: 600;
-    font-size: 60px;
-    line-height: 72px;
-    margin-bottom: 20px;
-    margin-top: 0;
+  font-weight: 600;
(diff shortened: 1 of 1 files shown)


willium:
Looks like Trellis plots are broken. I'm not sure if that's because of my fix or they were broken before. I'll see if I can fix it.

screen shot 2016-01-15 at 5 06 58 pm

domoritz:
It looks like some flexbox issue. Do you see the same issue in chrome 48?

willium:
This is fixed in the latest, but will break as long as it's fixed height (based on current implementation). What happens if we do fixed height and a vertical trellis is too tall?

domoritz:
> What happens if we do fixed height and a vertical trellis is too tall?

The content should scroll. 

odd

domoritz:
It looks like we are not using the flex box alignment any more. 

screen shot 2016-01-15 at 21 10 59

willium:
Fixed the trellis/max-height.

Re: alignment, which is desired:
screen shot 2016-01-17 at 2 13 28 pm

OR

screen shot 2016-01-17 at 2 14 26 pm

domoritz:
I prefer the second option. Looks cleaner. Although I'm surprised that the heatmap is not aligned on top with the other charts.

willium:
More like this then:
screen shot 2016-01-17 at 2 23 16 pm

domoritz:
Almost ;-) I liked that the "View Source" and "Open in Vega Editor" are aligned at the bottom.

kanitw:
LGTM and agree with aligned "view source" and "open in editor" if it's not
too hard.

willium:
Great. I believe the latest works as desired. I also formatted the CSS because indentation was variable and it made it difficult to read.

willium:
Also rebased because I figured it made more sense to only 1 commit for this in master

domoritz:
Excellent. Works perfectly. I just changed scroll -> auto so that you only get the scrollbar when it's needed and switched to 2 spaces indentation. We use 2 spaces everywhere else so I wanted to make it consistent.

PR #956, opened 2016-01-17
Draft docs with rendered examples for point

kanitw:


domoritz:
Sweet

domoritz:
Just a thought. Right now the examples are not tested. If we had a way to load the examples from our examples directory, we could make sure they are tested. Shouldn't be too hard and maybe it's worth to do now.

kanitw:
> f we had a way to load the examples from our examples directory, we could make sure they are tested

What do you mean tested?   We have `example.test.ts` that validates all of them.   

If you want to do unit test for all of them, I think that's an overkill and is philosophically not "unit test".

Using some of them as spec in unit test is more palatable and you can also do json loading in unit test rather than putting specs in place.  However, you still need to generate some less polished spec to cover all the cases.  So having some specs pull from `examples/` would make the unit test code look less consistent.

kanitw:
Oh you mean examples in the tutorial?

domoritz:
> Oh you mean examples in the tutorial?

Yes.

kanitw:
> Shouldn't be too hard and maybe it's worth doing now.

I don't think it is that important.  You need to preview the doc when you write tutorial anyway.

Moreover, Making rendered correctly with markdown block is probably impossible so I guess we need other way for syntax highlighting.   

There are couple of more important tasks we should do. 

- Finish writing content of the docs
- Make docs's UI / navigation easy.  
- Finish all leftover 1.0 tasks
- And yes, we should do research stuff as well.

kanitw:
That said, I agree that it's a good thing to do.  (maybe later)

domoritz:
> You need to preview the doc when you write tutorial anyway.

I'm more worried about the other direction. If we change something in the library and forget to update the examples. But yes, not necessary now but it is worth thinking about.

PR #958, opened 2016-01-17
[WIP] Transformation Docs

kanitw:
So we can mention about what's the order of execution for all these transforms


kanitw:
Replaced with #1001

PR #959, opened 2016-01-18
Don’t add format to legends if we use bin because we have a special f…

domoritz:
…ormat that displays the range.

### DIFF
--- a/src/compile/legend.ts
+++ b/src/compile/legend.ts
@@ -37,7 +37,7 @@ export function compileLegend(model: Model, channel: Channel, def) {
 
   // 1.1 Add properties with special rules
   def.title = title(fieldDef);
-  model.format(channel, (typeof legend !== 'boolean' && legend.format) || undefined, def);
+  format(model, channel, def);
 
   // 1.2 Add properties without rules
   ['orient', 'values'].forEach(function(property) {
@@ -69,6 +69,16 @@ export function title(fieldDef: FieldDef) {
   }
 
   return fieldTitle(fieldDef);
+}
+
+export function format(model: Model, channel: Channel, def) {
+  const fieldDef = model.fieldDef(channel);
+  const legend = fieldDef.legend;
+
+  // If the channel is binned, we should not set the formatter because we have a range label
+  if (!fieldDef.bin) {
+    model.format(channel, (typeof legend !== 'boolean' && legend.format) || undefined, def);
+  }
 }
 
 namespace properties {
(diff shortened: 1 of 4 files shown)


kanitw:
The change LGTM. 

But I prefer would like to have `model.format()` either
- return `def` so it's clear that the method cause side effect on the input `def`
- or do something like

```js
extend(def, model.format()); 
```
where `model.format()` only returns mix-ins part of the format.   (And thus, possibly rename it to `model.formatMixins()`.)    

The latter is consistent with `compile/util.ts`'s `applyColorMixins` and scale's `rangeMixins`.  

This is slightly slower, but is clearer that we are modifying def.

domoritz:
> This is slightly slower, but is clearer that we are modifying def. 

Performance doesn't matter for small things like this, imho. I'll change this.

kanitw:
Cool.  And maybe unit test? :p

domoritz:
> return def so it's clear that the method cause side effect on the input def

It's not clear because you'd modify and return the def. Making a copy is so painful is js. 

> The latter is consistent with `compile/util.ts`'s `applyColorMixins` and scale's `rangeMixins`.

`compile/util.ts` doesn't have`applyColorMixins`. And `applyColorAndOpacity` actually writes directly.

kanitw:
I didn’t check that when I write lol … it’s purely from my bad memory.  

Thinking more about what I did, I think with the name `applyColorAndOpacity()` then it makes sense not to return `mixin`.  (Otherwise it should be named `colorAndOpacityMixins()`.)

> scale's `rangeMixins`.
This is still true though.

domoritz:
Agreed. We can easily make this a mixing, though by not taking the p parameter and just returning and incomplete object.

kanitw:
> We can easily make this a mixing, though by not taking the p parameter and just returning and incomplete object.

Yeah, but given that `applyColorAndOpacity()` is in `util.ts`, not `model.ts`.  The original `apply` method already make sense IMHO.

domoritz:
Done.

src/compile/legend.ts
```diff
@@ -71,6 +71,18 @@ export function title(fieldDef: FieldDef) {
...
+export function format(model: Model, channel: Channel) {
+  const fieldDef = model.fieldDef(channel);
+
+  // If the channel is binned, we should not set the formatter because we have a range label
+  if (!fieldDef.bin) {
+    return {};
+  }
+
+  const legend = fieldDef.legend;
+  return model.format(channel, (typeof legend !== 'boolean' && legend.format) || undefined);
```
kanitw: should we introduce `model.legend()` here just like how we have `model.axis()` for consistency?

kanitw:
Don’t we have to apply `model.format()` to `mark-text.ts` as well?

PR #962, opened 2016-01-18
By default, don’t pass a number format to vega.

domoritz:
With this change, by default we don't see the legend label issues from #497.

### DIFF
--- a/test/compile/Model.test.ts
+++ b/test/compile/Model.test.ts
@@ -62,10 +62,10 @@ describe('Model', function() {
           x: {field:'a', type: "quantitative"}
         },
         config: {
-          numberFormat: 'foo'
+          numberFormat: 'd'
         }
       }).format(X, undefined), {
-        format: 'foo'
+        format: 'd'
       });
     });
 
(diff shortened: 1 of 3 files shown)


test/compile/Model.test.ts
```diff
@@ -60,9 +60,12 @@ describe('Model', function() {
         mark: "point",
         encoding: {
           x: {field:'a', type: "quantitative"}
+        },
+        config: {
+          numberFormat: 'foo'
```
kanitw: Although `’foo’` is fine placing here for a purpose of unit testing, I prefer to a value that is a valid in real life setting as well.

kanitw:
Another general question that I should have noted in #497 is whether we should keep the `config.numberFormat` and `config.timeFormat`?       

If yes, we should remove `FIXME` comment in `config.schema.ts`.

domoritz:
I think we will always need a number format. Even if we add different number formats for special cases, we still need a config option.

domoritz:
Done

PR #969, opened 2016-01-18
format => formatMixins & provide scaffold for unit test

kanitw:
- the refactor part should provide some structure for our junior colleague’s unit-test exercises


domoritz:
I'm not sure whether I like the *mixin names but it's consistent.

kanitw:
> I'm not sure whether I like the *mixin names but it's consistent.
I think, at least, it suggests that it is more than just the  `format` property.

PR #970, opened 2016-01-19
Add missing shortTimeLabel for text marks, support formatting for tex…

domoritz:
…t marks, simplify logic.

Follow up for #937


src/compile/Model.ts
```diff
@@ -233,48 +233,88 @@ export class Model {
     return 30;
   }
 
-  /** @return an object with format and formatType properties. */
+  /**
+   * Returns an object with format and formatType properties.
```
kanitw: @param format explicitly specified format?

src/compile/Model.ts
```diff
@@ -233,48 +233,88 @@ export class Model {
...
+      const filter = def.formatType + (def.format ? ':\'' + def.format + '\'' : '');
+      return {
+        text: {
+          template: '{{' + this.field(channel, {datum: true}) + ' | ' + filter + '}}'
         }
-        break;
+      };
+    }
+
+    if (def.formatType === 'number') {
```
kanitw: Why setting it first and remove later?  Better not set it at all from the beginning.

src/compile/Model.ts
```diff
@@ -233,48 +233,88 @@ export class Model {
...
+      case COLOR:
+      case SHAPE:
+      case SIZE:
+        const legend = fieldDef.legend;
+        return (typeof legend !== 'boolean' ? legend.shortTimeLabels : false);
+      case TEXT:
+        return this.config().mark.shortTimeLabels;
+      case LABEL:
+        // TODO(#897): implement when we have label
+      case DETAIL:
```
kanitw: Detail doesn't have scale and thus shouldn't be called at all.  No need to add this condition.

src/compile/Model.ts
domoritz: See the check in :273

src/compile/Model.ts
domoritz: I like to have exhaustive switch statements.

src/compile/Model.ts
```diff
@@ -233,48 +233,88 @@ export class Model {
     return 30;
   }
 
-  /** @return an object with format and formatType properties. */
+  /**
+   * Returns an object with format and formatType properties.
+   */
   public formatMixins(channel: Channel, format: string) {
```
kanitw: Now that I add util.ts -- should this be moved to util.ts?   

One possible way to make distinction is things `util.ts` might either contains mixins or slightly complicated logic, while methods in model could be something more like helper?

src/compile/Model.ts
```diff
@@ -233,48 +233,88 @@ export class Model {
...
+    if (def.formatType === 'number') {
+      // no need to set vega default
+      delete def.formatType;
     }
 
     return def;
   }
 
-  /** returns the time format used for axis labels for a time unit */
+  private isAbbreviated(channel: Channel, fieldDef: FieldDef) {
```
kanitw: util.ts?

src/compile/legend.ts
```diff
@@ -81,7 +81,7 @@ export function formatMixins(model: Model, channel: Channel) {
   }
 
   const legend = fieldDef.legend;
-  return model.formatMixins(channel, (typeof legend !== 'boolean' && legend.format) || undefined);
+  return model.formatMixins(channel, (typeof legend !== 'boolean' && legend.format));
```
kanitw: formatMixins accept string as second parameter  
 should it be 
`typeof legend !== 'boolean' ? legend.format : undefined`  

although it pretty much gives same result.

test/compile/mark-text.test.ts
```diff
@@ -23,7 +23,43 @@ describe('Mark: Text', function() {
...
+    let spec = {
+      "mark": "text",
+      "encoding": {
+        "text": {"field": "foo", "type": "temporal"}
+      }
+    };
+    const model = parseModel(spec);
+    const props = text.properties(model);
+
+    it('should use dtae template', function() {
```
kanitw: date

test/compile/mark-text.test.ts
```diff
@@ -23,7 +23,43 @@ describe('Mark: Text', function() {
     it('should have placeholder text', function() {
       assert.deepEqual(props.text, {value: "Abc"});
     });
-  })
+  });
+
+  describe('with quantitative and format', function() {
+    let spec = {
```
kanitw: why `let` if you didn't modify it?

test/compile/mark-text.test.ts
```diff
@@ -23,7 +23,43 @@ describe('Mark: Text', function() {
...
+    const model = parseModel(spec);
+    const props = text.properties(model);
+
+    it('should use number template', function() {
+      assert.deepEqual(props.text, {template: '{{datum.foo | number:\'d\'}}'});
+    });
+  });
+
+  describe('with temporal', function() {
+    let spec = {
```
kanitw: why `let`?

src/compile/Model.ts
```diff
@@ -233,48 +233,88 @@ export class Model {
...
     if (format !== undefined) {
       def.format = format;
-      return def;
+    } else {
+      switch (fieldDef.type) {
+        case QUANTITATIVE:
+          def.format = this.numberFormat(channel);
+          break;
+        case TEMPORAL:
+          const f = this.timeFormat(channel);
```
kanitw: `def.format = this.timeFormat(channel) || this.config().timeFormat`
is more concise?

src/compile/mark-text.ts
```diff
@@ -72,18 +73,9 @@ export namespace text {
...
-        p.text = {
-          template: '{{' + model.field(TEXT, { datum: true }) +
-          ' | number:\'' + numberFormat + '\'}}'
-        };
-      } else {
-        p.text = { field: model.field(TEXT) };
-      }
+      p.text = { field: model.field(TEXT) };
+      const format = model.config().mark.format;
+      extend(p, model.formatMixins(TEXT, format));
```
kanitw: We generally avoid overriding and `p.text` is overridden for `Q` and `T`.

Better split this by `fieldDef.type`.   (And just call `formatMixins` for `Q` and `T`)

src/compile/Model.ts
kanitw: Better have 

`if ( contains([QUANTITATIVE, TEMPORAL], fieldDef.type) {`  

surrounds all logic in this function.  (`format` should be dropped for nominal and ordinal anyway!)

and do
```
   def.formatType = 'time'; 
}
```

src/compile/Model.ts
domoritz: I moved everything

src/compile/legend.ts
domoritz: done

test/compile/mark-text.test.ts
domoritz: my atom has become extremely sluggish recently

test/compile/mark-text.test.ts
domoritz: just copied

src/compile/Model.ts
domoritz: Done

src/compile/mark-text.ts
domoritz: What if the format is specified for ordinal?

src/compile/util.ts
```diff
@@ -52,3 +55,84 @@ export function applyMarkConfig(marksProperties, model: Model, propsList: string
...
+
+    const filter = def.formatType + (def.format ? ':\'' + def.format + '\'' : '');
+    return {
+      text: {
+        template: '{{' + model.field(channel, { datum: true }) + ' | ' + filter + '}}'
+      }
+    };
+  }
+
+  if (def.formatType === 'number') {
```
kanitw: Still have this add and delete...

domoritz: Still because of line 92. I can drop this but it makes the logic hard to change.

src/compile/util.ts
```diff
@@ -52,3 +55,84 @@ export function applyMarkConfig(marksProperties, model: Model, propsList: string
...
+
+  if (format !== undefined) {
+    def.format = format;
+  } else {
+    switch (fieldDef.type) {
+      case QUANTITATIVE:
+        def.format = model.numberFormat(channel);
+        break;
+      case TEMPORAL:
+        def.format = timeFormat(model, channel) || model.config().timeFormat;
```
kanitw: I guess `timeFormat` is only used here -- remove `timeFormat()` and move the logic here?

src/compile/util.ts
```diff
@@ -52,3 +55,84 @@ export function applyMarkConfig(marksProperties, model: Model, propsList: string
...
+      def.formatType = 'time';
+      break;
+  }
+
+  if (format !== undefined) {
+    def.format = format;
+  } else {
+    switch (fieldDef.type) {
+      case QUANTITATIVE:
+        def.format = model.numberFormat(channel);
```
kanitw: I guess `numberFormat` is only used here -- remove `model.numberFormat()` and move the logic here?

domoritz: done

src/compile/util.ts
domoritz: it's being tested as a whole

domoritz:
Done, now format only works for Q and T.

kanitw:
:tada:

src/schema/config.marks.schema.ts
```diff
@@ -204,9 +205,13 @@ export const markConfig = {
     // text-only & VL only
     format: {
       type: 'string',
-      default: '',  // auto
-      description: 'The formatting pattern for text value.'+
-                   'If not defined, this will be determined automatically. '
+      default: undefined,  // auto
+      description: 'The formatting pattern for text value. If not defined, this will be determined automatically. '
+    },
+    shortTimeLabels: {
```
kanitw: ah, should also add this to `config.md`!

PR #972, opened 2016-01-19
Test “fill” of square and circle mark

kanitw:


kanitw:
Actually I forgot to add two more cases in the scaffolding:
- “It should have correct shape”  (test props.shape.value === ‘circle’ / ‘square’ for each mark type.

domoritz:
When you're done, remove the [WIP] from the pr title.

kanitw:
And tag us in the comment :)

test/compile/mark-point.test.ts
```diff
@@ -114,7 +114,15 @@ describe('Mark: Square', function() {
   });
 
   it('should be filled by default', function() {
-    // TODO
+    const model = parseModel({
```
ayush29feb: Shouldn't the cost model and props be defined outside it()? Looking at the rest of the tests I think it should be defined inside the describe()

domoritz: Here, we want to test different models parsed from different specs. If we can extract some common functionality, we can of course do that.

ayush29feb:
So both circle and square have this
```javascript
it('should return the correct mark type', function() {
    // TODO call square.markType()
});
```
but atom says `Supplied parameter do not match any signature of call target`

I have finished the rest of the test cases

domoritz:
How do you call `sqare.markType`?

ayush29feb:
I think it figured it out. Please check the commit I made in ayush/test-square branch

domoritz:
e1f514f0e240f319aa57f56d5713f277af51bc55

domoritz:
Instead of creating a model, you can just test `point.markType() === "symbol"`.

ayush29feb:
`circle.markType()` & `sqaure.markType` doesn't seem to be working.

domoritz:
Fixed in https://github.com/vega/vega-lite/pull/977. I'm glad you caught the bug!

Issue #888, opened 2015-12-20
Supports streamgraph


kanitw:
Here is an [streamgraph example](http://bl.ocks.org/mbostock/4060954) from d3:  http://bl.ocks.org/mbostock/4060954.  

![streamgraph](https://cloud.githubusercontent.com/assets/111269/12273872/0b77d972-b91d-11e5-83ef-74a8d2c80391.png)

To get a vega example, try adding  "offset": “center” to the stack transform in the [Vega stack area example](http://vega.github.io/vega-editor/?mode=vega&spec=stacked_area).  ([Modified spec](http://pastebin.com/2a8fxkJu))

I believe this is pretty much supported in Vega-Lite, but we need to
- [ ] create a good Vega-Lite example. 
   - look at area chart example
   - set `config.stack.offset` to center
   - hide y-axis by setting `axis` = `false`
   - set `config.mark.area` to some nice one

Some more notes about stream graph:
- https://hrbrmstr.github.io/streamgraph/#fn5
- http://www.visualisingdata.com/2010/08/making-sense-of-streamgraphs/

kanitw:
Almost done in https://github.com/vega/vega-lite/issues/973, let’s continue conversation there.

PR #973, opened 2016-01-19
[WIP] Added Streamgraph example

ayush29feb:
Added the streamgraph example (showInEditor : false)

### DIFF
--- a/examples/streamgraph.json
+++ b/examples/streamgraph.json
@@ -1,0 +1,18 @@
+{
+  "data": {"url": "data/cars.json"},
+  "mark": "area",
+  "encoding": {
+    "x": {"field": "Year","type": "temporal","timeUnit": "year"},
+    "y": {
+      "field": "Weight_in_lbs",
+      "type": "quantitative",
+      "aggregate": "sum",
+      "axis": {"labels": false,"grid": false,"tickSize": 0}
+    },
+    "color": {"field": "Cylinders","type": "nominal"}
+  },
+  "config": {
+    "mark": {"interpolate": "cardinal"},
+    "stack": {"offset": "center"}
+  }
+}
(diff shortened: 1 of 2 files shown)


domoritz:
Maybe we can replace this with a movie example.

kanitw:
You can say "fixes #888" to suggest that this fixes an issue.  ;)

examples/streamgraph.json
```diff
@@ -0,0 +1,18 @@
...
+      "field": "Weight_in_lbs",
+      "type": "quantitative",
+      "aggregate": "sum",
+      "axis": {"labels": false,"grid": false,"tickSize": 0}
+    },
+    "color": {"field": "Cylinders","type": "nominal"}
+  },
+  "config": {
+    "mark": {"interpolate": "cardinal"},
+    "stack": {"offset": "center"}
```
kanitw: Actually after thinking more about it, the offset should be `”wiggle”`, so it’s similar to the d3 example: http://bl.ocks.org/mbostock/4060954.

kanitw:
@domoritz  I don’t know if this really needs to become the movie example.

domoritz:
I think the move dataset leads to an infinitely more interesting visualization. Seeing the weights of cars by # of cylinders over time is not that exciting.

ayush29feb:
Hey, so I have been trying some examples for the movies data set. However, I am not really able to get a good looking visualization. I tried stuff like `"x": "Release_Date"`, `y: US_Gross` and `color : Major_Genre`
I am guessing this is because the data is not very consistent and contains a lots of `null`

image

domoritz:
Marked as WIP. @ayush29feb change title when you're done.

kanitw:
supersedes by https://github.com/vega/vega-lite/pull/1138

PR #978, opened 2016-01-22
[WIP] Bar Label

kanitw:


src/compile/mark-bar.ts
```diff
@@ -159,7 +159,35 @@ export namespace bar {
   }
 
   export function labels(model: Model) {
-    // TODO(#64): fill this method
-    return undefined;
+    let p: any = {};
+
+    const orient = model.config().mark.orient;
+    const horizontal = (orient === 'horizontal');
```
kanitw: isHorizontal? :)

willium:
This will close #64 when merged

src/compile/mark-bar.ts
willium: good idea. Fixed

src/compile/mark.ts
```diff
@@ -69,13 +69,13 @@ export function compileMark(model: Model): any[] {
         );
 
       return [{
-        name: (name ? name + '-' : '') + mark + '-facet',
+        name: (specName ? specName + '-' : '') + mark + '-facet',
```
domoritz: This one is very different from the other ones. Can you extract this into a helper function of the model?

src/compile/mark-bar.ts
```diff
@@ -159,7 +159,33 @@ export namespace bar {
...
   export function labels(model: Model) {
-    // TODO(#64): fill this method
-    return undefined;
+    let p: any = {};
+
+    const orient = model.config().mark.orient;
+    const isHorizontal = (orient === 'horizontal');
+
+    // for reactive geometry
+    let datumField = 'datum.' + (isHorizontal ? model.field(X) : model.field(Y));
```
kanitw: you can also do

`const datumField = model.field(isHorizontal ? X :Y, {datum: 1});`

- Should be `const` because you never modify it.  
- See `FieldRefOption` object which is an optional parameter for `model.field`.

- To make it work correct with stacked bar, need to add suffix `’_end’`.   
See `mark-bar.ts` line 27.   
  - (We should make `suffix` another parameter in `FieldRefOption` though! — feel free to do that in a separate PR — since that refactor should be merged to the master)

willium:
This branch got slightly convoluted as I have been keeping it up to date with will/fieldDef-suffix (and master) as I am using what I built their within this branch. See the PR for that one over here https://github.com/vega/vega-lite/pull/983

src/compile/mark.ts
willium: Something like this? https://github.com/vega/vega-lite/blob/will/bar-label/src/compile/Model.ts#L189-L192

willium:
Latest commits partially implements text formatting for labels. However, I’m not able to get formatting to work for labels (or perhaps its text in the properties)… I did some reading into Vega/D3 and I see that s is proper. From the example that uses formatting

```
{
  "description": "This example shows how to use timeUnit and shortTimeNames.",
  "data": {
    "url": "data/birdstrikes.json"
  },
  "mark": "line",
  "encoding": {
    "x": {
      "axis": {
        "shortTimeNames": true
      },
      "field": "Flight_Date",
      "type": "temporal",
      "timeUnit": "month"
    },
    "y": {
      "field": "*",
      "type": "quantitative",
      "aggregate": "count"
    }
  }
}
```

 its only on an axis. When I implement the formatting I get NaNundefined for all the labels as seen below

screen shot 2016-01-23 at 3 29 10 am

This can be reproduced with the following Vega spec (which is generated by my code, see latest commit)

```
{
  "width": 200,
  "height": 1,
  "padding": "auto",
  "data": [
    {
      "name": "source",
      "values": [
        {"category": "A","amount": 28},
        {"category": "B","amount": 55},
        {"category": "C","amount": 43},
        {"category": "D","amount": 91},
        {"category": "E","amount": 81},
        {"category": "F","amount": 53},
        {"category": "G","amount": 19},
        {"category": "H","amount": 87},
        {"category": "I","amount": 52}
      ],
      "format": {"type": "json","parse": {"amount": "number"}},
      "transform": [{"type": "filter","test": "datum.amount!==null"}]
    },
    {
      "name": "layout",
      "source": "source",
      "transform": [
        {
          "type": "aggregate",
          "summarize": [{"field": "category","ops": ["distinct"]}]
        },
        {
          "type": "formula",
          "field": "cellHeight",
          "expr": "(datum.distinct_category + 1) * 21"
        }
      ]
    }
  ],
  "marks": [
    {
      "name": "root",
      "type": "group",
      "from": {"data": "layout"},
      "properties": {
        "update": {
          "width": {"value": 200},
          "height": {"field": "cellHeight"}
        }
      },
      "marks": [
        {
          "name": "rect",
          "type": "rect",
          "from": {"data": "source"},
          "properties": {
            "update": {
              "x": {"scale": "x","field": "amount"},
              "x2": {"scale": "x","value": 0},
              "yc": {"scale": "y","field": "category"},
              "height": {"value": 20},
              "strokeWidth": {"value": 2},
              "fill": {"value": "#4682b4"}
            }
          }
        },
        {
          "type": "text",
          "from": {"mark": "rect"},
          "properties": {
            "update": {
              "text": {"template": "{{datum.amount | number:'s'}}"},
              "x": {"field": "x2","offset": 5},
              "y": {"field": "yc"},
              "baseline": {"value": "middle"},
              "align": {"value": "left"},
              "fill": {"value": "#000000"},
              "strokeWidth": {"value": 2}
            }
          }
        }
      ],
      "scales": [
        {
          "name": "x",
          "type": "linear",
          "domain": {"data": "source","field": "amount","sort": true},
          "rangeMin": 0,
          "rangeMax": 200,
          "round": true,
          "clamp": true,
          "nice": true
        },
        {
          "name": "y",
          "type": "ordinal",
          "domain": {
            "data": "source",
            "field": "category",
            "sort": true
          },
          "bandWidth": 21,
          "round": true,
          "clamp": true,
          "padding": 1,
          "points": true
        },
        {
          "name": "label",
          "type": "linear",
          "domain": {"data": "source","field": "amount"},
          "clamp": true,
          "zero": false
        }
      ],
      "axes": [
        {
          "type": "x",
          "scale": "x",
          "format": "s",
          "grid": true,
          "layer": "back",
          "ticks": 5,
          "title": "amount"
        },
        {
          "type": "y",
          "scale": "y",
          "grid": false,
          "title": "category",
          "properties": {
            "labels": {
              "text": {"template": "{{ datum.data | truncate:25}}"}
            }
          }
        }
      ]
    }
  ]
}
```

From my research into Vega and D3 formatting, this line 
```
              "text": {"template": "{{datum.amount | number:'s'}}"},
```

looks correct. Hopefully you can point out what I'm missing!

willium:
I resolved this by making 
```
 "text": {"template": "{{datum.amount | number:'s'}}"},
```
into
```
 "text": {"template": "{{datum.datum.amount | number:'s'}}"},
```

This makes the labels into 12.0000000 instead of NaN which is good, but I'm not sure how to truncate those trailing 0s as I believe that's handled by Vega itself. Unfortunately a double datum prefix is both difficult to read/comprehend, and it breaks the factory function of fieldef, and might require adding a place to add prefixes on top of datum (or in this case, two datums). I think this is because of reactive geometry needing that datum prefix and formatting needing one as well. Yikes.

kanitw:
@willium This is quite a tricky issue.  

We can of course customize `numberFormat` from the spec, but another related issue that we should fix separately is that default `numberFormat` should be different based on aggregation function like discussed in part of #497 

@domoritz What do you think?

domoritz:
The formatting is an unrelated issue and has to be fixed at the vega level. For this issue, you should just use the format function from the model (maybe with modifications to get the `datum.datum`). 

We should talk to @arvind about how we correctly use the reactive geometry.

src/compile/Model.ts
```diff
@@ -186,6 +186,11 @@ export class Model {
     return this._spec.data;
   }
 
+  public name(suffix: string): string {
```
domoritz: Describe the method.

src/compile/mark-bar.ts
```diff
@@ -165,7 +166,47 @@ export namespace bar {
...
+
+    const stack = model.stack();
+
+    let fieldOptions: any = { datum: true };
+
+    if (stack && X === stack.fieldChannel) {
+      fieldOptions.suffix = '_end';
+    }
+
+    // const labelField = (model.fieldDef(LABEL).field === model.fieldDef(X).field) ? X : Y;
```
domoritz: delete

src/compile/util.ts
```diff
@@ -57,13 +58,24 @@ export function applyMarkConfig(marksProperties, model: Model, propsList: string
   });
 }
 
+export function applyLabelConfig(marksProperties, model: Model) {
```
domoritz: add comment

src/compile/mark.ts
domoritz: Yes

src/compile/util.ts
```diff
@@ -89,14 +101,16 @@ export function formatMixins(model: Model, channel: Channel, format: string) {
...
     // text does not support format and formatType
     // https://github.com/vega/vega/issues/505
 
     const filter = (def.formatType || 'number') + (def.format ? ':\'' + def.format + '\'' : '');
+    const options = fieldOptions || {};
+
     return {
       text: {
-        template: '{{' + model.field(channel, { datum: true }) + ' | ' + filter + '}}'
+        template: '{{datum.' + model.field(channel, options) + ' | ' + filter + '}}'
```
domoritz: If we need the `datum.datum`, we should probably add a fieldOption for it.

src/schema/config.label.schema.ts
```diff
@@ -0,0 +1,129 @@
...
+      default: undefined,
+      description: 'The rotation angle of the text, in degrees.'
+    },
+    baseline: {
+      type: 'string',
+      default: undefined,
+      enum: ['top', 'middle', 'bottom'],
+      description: 'The vertical alignment of the text. One of top, middle, bottom.'
+    },
+    dx: {
```
domoritz: Do you think it makes sense to combine the offset properties and automatically set it to dx or dy? In vega-lite, the config should be the same for multiple charts. If you switch from a horizontal to a vertical bar chart with labels, you'd also have to change this property.

domoritz: Actually, this is passed to vega so never mind. But a general offset property in vega-lite would be nice, don't you think?

src/schema/config.label.schema.ts
```diff
@@ -0,0 +1,129 @@
...
+    },
+    theta: {
+      type: 'number',
+      default: undefined,
+      description: 'Polar coordinate angle, in radians, of the text label from the origin determined by the x and y properties. Values for theta follow the same convention of arc mark startAngle and endAngle properties: angles are measured in radians, with 0 indicating "north".'
+    },
+    format: {
+      type: 'string',
+      default: 's',
+      enum: ['e', 'f', 'g', 'r', 's', '%', 'p', 'b', 'o', 'd', 'x', 'X', 'c', 'n'],
```
domoritz: I didn't know you can do that ;-)

However, this is not complete because you can have numbers and . as well. Maybe remove it for now and not worry too much about it. We can add a regex for it later (here and in a lot of other places).

src/schema/config.label.schema.ts
willium: Thats a good idea. I think it might be something good add to 1.x-label because it might vary based on the way different charts use labels?

kanitw: To make it generalize, `dx`, `dy` already mean offset-x/y.  
But I guess Dom's `offset` means offset that aligns with the orientation, which is only sensible for some tick, bar, and area (and maybe line) -- but not for point.  Not sure adding `offset` is worth the complexity though.

domoritz: I think this property can also make sense for points. In that case the offset describes the distance between the point and the label in any direction that is sensible.

kanitw: How would  you determine "direction that is sensible" for points?

domoritz: In the simplest case it can be anything to the right. Once we have a smart algorithm, it is the direction that minimized overlap.

kanitw: I think a smart algorithm might adjust both offset-x and offset-y (`dx` / `dy`).  Unless we're certain that it will always adjust one direction, I wouldn't add `offset` as we might want to remove it in the future.

domoritz: I don't understand. Exactly when we have a smart algorithm, then we need a single offset property because the algorithm would split it into dx and dy.

willium: dx and dy could relate to offset with pythagorean theorem, dx^2+dy^2=offset^2

kanitw: Fair enough, I think I misunderstood when I first read it.  We can have `offset` then.

domoritz:
Let's close this pr for now since @willium will redo it with all the new understanding of the problem.

PR #979, opened 2016-01-22
Transparent layer bar is added to examples

bobocandys:
@kanitw @domoritz

### DIFF
--- a/examples/transparent_layer_bar.json
+++ b/examples/transparent_layer_bar.json
@@ -1,4 +1,5 @@
 {
+  "description": "A bar chart showing the US population distribution of age groups and gender in 2000.",
   "data": { "url": "data/population.json",
     "filter": "datum.year == 2000",
     "calculate": [{"field": "Gender", "expr": "datum.sex == 2 ? \"Female\" : \"Male\""}]
--- a/examples/transparent_layer_bar.json
+++ b/examples/transparent_layer_bar.json
@@ -1,0 +1,20 @@
+{
+  "data": { "url": "data/population.json",
+    "filter": "datum.year == 2000",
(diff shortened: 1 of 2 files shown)


examples/transparent_layer_bar.json
```diff
@@ -0,0 +1,20 @@
+{
+  "data": { "url": "data/population.json",
```
kanitw: Maybe add description:

``"description": "..."`

(Think if you're putting this on a newspaper or something, what would the caption of the output figure.)

examples/transparent_layer_bar.json
```diff
@@ -0,0 +1,20 @@
...
+    "filter": "datum.year == 2000",
+    "calculate": [{"field": "Gender", "expr": "datum.sex == 2 ? \"Female\" : \"Male\""}]
+  },
+  "mark": "bar",
+  "encoding": {
+    "y": {"field": "people","type": "quantitative", "aggregate": "sum"},
+    "x": {"field": "age", "type": "ordinal"},
+    "color": {"field": "Gender","type": "nominal",
+      "scale": {"range": ["red","blue"]}
```
kanitw: Is red blue better than the original color?

bobocandys: I think this is easier for people to understand the ratio because females are often associated with red-ish colors, males are with blue-ish colors.

examples/transparent_layer_bar.json
bobocandys: Added.

examples/transparent_layer_bar.json
kanitw: okay.

kanitw:
Awesome :)

domoritz:
Whoo. Looks great!

screen shot 2016-01-22 at 17 39 10

I wonder whether we should switch back to the `s` number formatter.

PR #983, opened 2016-01-23
add suffix to fieldDefOption for stack usage

willium:
Quick fix allows for suffix to be added to fields as an option, for use with stacked bar. As proposed by @kanitw here: https://github.com/vega/vega-lite/pull/978#discussion_r50604826.


kanitw:
Awesome :)

PR #989, opened 2016-01-23
Add description for name

domoritz:


src/schema/schema.ts
```diff
@@ -29,7 +29,8 @@ export var schema = {
   required: ['mark', 'encoding'],
   properties: {
     name: {
-      type: 'string'
+      type: 'string',
+      description: 'A name for the specification. The name is used to annotate marks.'
```
kanitw: Not just marks, scale name, etc.

PR #1001, opened 2016-01-26
Decouple data source from data transform: add transform property to the top-level 

### LINKED ISSUE #994: Separate `data` source from data `transform`?
Based on my experience writing docs, `data` is mixing two things right now.  
Some properties describe source (`url`, `values`), some describe transformations (`calculate`, `filter`). 

I think we should separate `transform` from `data` just like GoG.  
Basically, this will make a clear separation that all properties in `data` only refers to data source.  Meanwhile, everything under `transform`  (e.g,. `calculate`, `filter`, `nullFilter` and more in the future) refers to transformation.

This is much easier to teach than that this `data` object does a few things.  

Compare these two alternatives:
- __Keeping transform in data__: Transformation are specified either from some properties inside `data` (which describe both source and some transform) or from inline transformation inside `encoding`.
- __Having transform object__:  Transformation are specified either from `transform` or from inline transformation inside `encoding`.

I would rather teach the latter to people.  While in Vega, transform is inside data — it is always encapsulated inside one `transform` property rather than multiple properties so it is much easier to learn.  

Any thoughts?

----------
Another relevant issue (no need to do now) — we can imagine that in the future, the top-level `transform` can accept an array that simply takes Vega transform as well.

kanitw:
Fix #994.  Also, see discussion behind the rationale for this move in #994.  

- Move `calculate` and `filter` to the new top-level `transform` property so we have a clean separation between `data` source and data `transform`.
- Swap order of transform so that transforms in `transform` happens before `encoding`'s inline transforms
- Move `filterNull` to be inside the top-level `transform`
- Update transform docs + Move detail about inline transforms from encoding.md to transform.md


docs/config.md
```diff
@@ -66,8 +64,6 @@ and the following fill and stroke properties:
 | Property      | Type          | Description    |
 | :------------ |:-------------:| :------------- |
 | filled        | Boolean        | Whether the shape\'s color should be used as fill color instead of stroke color.  This is only applicable for `bar`, `point`, and `area`.  All marks except `point` marks are filled by default. |
-| sortBy        | Field &#124; Field[] | Data field(s) for sorting layer of marks.  The first mark will placed on the bottom.  `"-"` prefix can be added to each field to set descending order. |
```
kanitw: I’m hiding this because they will be moved anyway and just in case I don’t finish that before wednesday — better hide it rather than causing confusion.

docs/transform.md
```diff
@@ -0,0 +1,303 @@
...
+    "encoding": {
+      "y": {"type": "quantitative","field": "b2"},
+      "x": {"type": "ordinal","field": "a"}
+    }
+  }
+});
+</script>
+<div id="bar_filter_calc"></div>
+
+## Inline Transforms inside each `encoding` channel
```
domoritz: Hmm, maybe this should be a different file.

kanitw: Arguably, currently I put on top that it can be done in two places ..

Issue #726, opened 2015-11-14
Validate & Auto-modify Spec


kanitw:
To truly validate a Vega-Lite spec, we have to validate 

1) if the spec matches Vega-Lite JSON schema

2) Some other rules:
- [ ] **Relationship between the encoding mapping and the mark type**
  - all the required encoding channels for the mark type are specified
  - (and) all the specified encoding channels are supported by the mark type
- **channel and functions**
  - [ ] aggregating fields on `row`, `column`, `shape`, or `detail`
  - [ ] bin on `shape`
- **data type based constraints**
  - [ ] `timeUnit` without type T
    - [ ] auto-drop timeUnit?
  - [ ] `aggregate` with ordinal / nominal 
    - [ ] auto-drop `aggregate`
  - [ ] `bin` without Q
    - [ ] auto-drop
  - [ ] using line without O, T (just warning)
  - [ ] use `scale.type` = `log`, `sqrt`, etc.  for non-‘Q’ type
    - [ ] auto-drop
  - [ ] use `scale.type` = `time` for non-`T` type
    - [ ] auto-drop
  - [ ] use `Q` as dimension for aggregate plot
    - [ ] auto-convert to `O` #688
- [ ] using `order` and `path` with `nominal` field (Nominal by default should not have order)
- **role-based constraints (dimension / measure)**
  - [ ] use two dimension with `line` / `area`
  - [ ] use two measure with `bar`  (Using two dimension with `bar` is fine for us though! — That will produce a heat map.)
- **other rules**
  - [ ] warning people use `bar` with `scale.zero=false`  (If scale doesn’t start at zero, bar is very misleading.)
  - [ ] use summative aggregation (’sum’, ‘count’, etc.) for stack
    We probably should have one `vl.validate()` method for people to call, but that also requires some JSON schema validator. 
  - [ ] use `config.mark.orient` when orientation is unambiguous  (e.g., when x is dimension and y is measure — the orientation should be vertical for sure.  Specifying `horizontal` isn’t a good idea.) 
  - [ ] path can only be used with marktype line
  
  We have a few options:
  (1) implement a method that both calls JSON schema validate and our special rules, and shim ZSchema
  (2) implement only a method that validate our special rules.  

Other suggestions are welcomed!

@domoritz, @yhoonkim  any thoughts?

domoritz:
I think we should offer a utility method. This allows us to change its implementation without breaking other peoples code.

kanitw:
[...]

PR #1002, opened 2016-01-26
Add `path` channel for sorting line's path + Replace `config.mark.sortLineBy`

kanitw:
Note: Documentation will be written in another huge doc writing PR.

### DIFF
--- a/src/schema/encoding.schema.ts
+++ b/src/schema/encoding.schema.ts
@@ -15,6 +15,7 @@ export interface Encoding {
   color?: FieldDef;
   size?: FieldDef;
   shape?: FieldDef;
+  path?: FieldDef;
   detail?: FieldDef | FieldDef[];
   text?: FieldDef;
   label?: FieldDef;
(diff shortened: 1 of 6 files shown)


domoritz:
We should add a warning that path can only be used with line. Should be done in https://github.com/vega/vega-lite/issues/726

kanitw:
Yeah it is a part of 

> all the specified encoding channels are supported by the mark type 

in #726

kanitw:
I already added line to path's supported mark type in channel.ts, so that should be done automatically with 

> all the specified encoding channels are supported by the mark type

src/schema/encoding.schema.ts
```diff
@@ -15,6 +15,7 @@ export interface Encoding {
   color?: FieldDef;
   size?: FieldDef;
   shape?: FieldDef;
+  path?: FieldDef;
```
domoritz: | [FieldDef]?

kanitw: DOne

domoritz:
Fix conflicts

kanitw:
Fixed

PR #1007, opened 2016-01-27
Rename examples (to be feature oriented name), remove redundant ones.

kanitw:


domoritz:
I would like our examples to be cool visualizations and not just "feature demos". 

Maybe we need a two categories
* Cool complex visualizations that actually show something
* One visualization with each marktype, plain (feature demo)

kanitw:
This tension is coming back again.  Features demos also serves well as a visual unit tester, but yeah we need cool vis to serve as showcase as well.  

That said, I think we can just add “showcase” category but I don’t think that’s the scope of this PR.

PR #1016, opened 2016-01-27
Missing supported mark types in documentation page

emilygu:
Mark page shows 'circle', 'square' and 'tick' as possible mark types but the table in the main documentation page leaves it out.


domoritz:
Point can be circle, square, ...

But tick is definitely missing

PR #1030, opened 2016-01-29
Add script for adding example specs from `examples/`

kanitw:
This will simplify the documentation a lot


_layouts/default.html
```diff
@@ -12,14 +12,20 @@
...
     <script src="{{site.baseurl}}/vega-lite.min.js"></script>
     <script src="{{site.baseurl}}/bower_components/vega-embed/vega-embed.js" charset="utf-8"></script>
+    <script src="{{site.baseurl}}/bower_components/highlightjs/highlight.pack.js" charset="utf-8"></script>
+
+    <script>var BASEURL = "{{site.baseurl}}";</script>
+    <script src="{{site.baseurl}}/site/static/docs.js" charset="utf-8"></script>
 
     <title> {{site.title}}: {{page.title}} </title>
     <link rel=stylesheet href="{{site.baseurl}}/site/static/main.css">
     <link rel=stylesheet href="{{site.baseurl}}/site/static/syntax.css">
```
domoritz: So if we use highlight, we don't need this any more, right? Doesn't it share the same classes.

site/docs/config.md
```diff
@@ -207,38 +179,10 @@ vg.embed('#horizontal_line', {
...
-    "config": {
-      "mark": {"interpolate":"monotone"}
-    }
-  }
-});
-</script>
-<div id="line_interpolate"></div>
+
+<div id="ex-line_monotone" class="side"></div>
+<script>example("line_monotone", "docs")</script>
```
domoritz: Hmm, I actually liked that the examples were inline. Makes it much easier to change.

site/static/main.css
```diff
@@ -192,3 +192,48 @@ h6 {
...
+  flex-direction: row-reverse;
+}
+.example .example-vis {
+  flex-grow: 1;
+  align-self: center;
+  min-width: 300px;
+  text-align: center;
+}
+.example .example-code {
+  width: 550px;
```
domoritz: Note that the width of the page is not fixed any more! Check out on small screens.

site/static/main.css
```diff
@@ -192,3 +192,48 @@ h6 {
...
+  margin: 0;
+  align-self: flex-start;
+}
+.example .example-code pre {
```
domoritz: combine classes

_layouts/default.html
kanitw: I only call highlight.js on examples, so they are not the same class.

site/docs/config.md
kanitw: That's not true because a lot of examples are in `examples/` as well, so when you want to change one line - you have to change in multiple places (3 times in the old code).

site/static/main.css
kanitw: Too bad, should have committed this earlier so you have to fix this for me lol

site/static/main.css
kanitw: Done

site/static/main.css
kanitw: Ok.  I make it slightly more responsive.

domoritz: Lol, already in master. So your call to fix it :-P

site/static/main.css
```diff
@@ -192,3 +191,52 @@ h6 {
...
+  flex-direction: column;
+  margin: 24px 0;
+}
+.example.side {
+  flex-direction: row-reverse;
+}
+
+@media (max-width: 640px) {
+  .example.side {
+    flex-direction: column;
```
domoritz: I like

PR #1039, opened 2016-01-31
Describe the purpose of marks.

domoritz:


site/docs/mark.md
```diff
@@ -5,8 +5,7 @@ title: Mark
 permalink: /docs/mark.html
 ---
 
-Marks are the basic visual building block of a visualization.
-In a bar chart, the marks are bars. In a scatterplot, the marks might be circles or squares.
+Marks are the basic visual building block of a visualization: they are graphical primitives whose properties (such as position, size, and color) can be used to visually encode data. In a bar chart, the marks are bars. In a scatterplot, the marks might be circles or squares.
```
kanitw: I like this!   Except that scatterplot is normally a point for us (and circle and square are special case of point).  Maybe mention line chart instead of scatter?

domoritz: Lol, that's the only part I copied from the old description :-P

kanitw: lol

PR #1040, opened 2016-01-31
Refactor FieldDef Schema

kanitw:
- Declare schema for all types of fields inside field.schema.ts
- Make all channels support aggregate


domoritz:
Does this affect the docs?

kanitw:
It simplifies the docs (that we don't have to write about special case for aggregate)

PR #1042, opened 2016-02-01
Use time scale for composite time units. Fixes #1013

### LINKED ISSUE #1013: Shouldn’t “yearmonth" use time scale by default?
```
{
  "transform": {"filter": "datum.symbol==='GOOG'"},
  "mark": "line",
  "encoding": {
    "x": {"field": "date", "type": "temporal", "timeUnit": "yearmonth"},
  }
}
```

vega_editor

domoritz:


src/compile/scale.ts
```diff
@@ -65,31 +65,28 @@ export function type(fieldDef: FieldDef, channel: Channel, mark: Mark): string {
...
           case 'month':
             return 'ordinal';
-          case 'date':
-          case 'year':
-          case 'second':
-          case 'minute':
-            // Returns ordinal if (1) the channel is X or Y, and
-            // (2) is the dimension of BAR or TICK mark.
-            // Otherwise return linear.
-            return contains([BAR, TICK], mark) &&
```
domoritz: Tbh, I don't understand this logic so I changed it.

src/compile/scale.ts
```diff
@@ -65,31 +65,28 @@ export function type(fieldDef: FieldDef, channel: Channel, mark: Mark): string {
         // Also, if we support color ramp, this should be ordinal too.
         return 'time'; // time has order, so use interpolated ordinal color scale.
       }
-      if (contains([ROW, COLUMN, SHAPE], channel)) {
```
domoritz: Seems odd that we override the user preference. If they really want to use something wrong, why not?

src/compile/scale.ts
```diff
@@ -65,31 +65,28 @@ export function type(fieldDef: FieldDef, channel: Channel, mark: Mark): string {
...
+
       if (fieldDef.scale.type !== undefined) {
         return fieldDef.scale.type;
       }
 
+      // Returns ordinal if
+      // (1) the channel is ROW, COLUMN, or SHAPE, or
+      // (2) uses a BAR or TICK mark
+      // because in that case we should not use a time scale.
+      if (contains([ROW, COLUMN, SHAPE], channel) || contains([BAR, TICK], mark)) {
```
domoritz: This seems to cover the cases I could think of.

src/compile/scale.ts
kanitw: Please undo this.  This will not output any thing reasonable.  
When would you want a value of shape to be linear value except than you're making mistake? 
When would you map a facet to a quantitative scale?

src/compile/scale.ts
kanitw: I think the make sense for mark's condition to stay down here.  

But I start wondering if we should have special case for bar and tick here...

kanitw: I'll test if it makes sense when I arrive at the apartment, gotta go to post office first.

kanitw: So I think this really depends on how we think about it.

If we don't assume that users are design-savvy.   Bar/Tick should use ordinal only if the cardinality of the dimension is low.  Otherwise, use linear/temporal.  Since we don't know cardinality, I am leaning toward making it consistent with other marks. 

However, if we assume that users are **often** design-savvy and they would only use bar/tick for this if they know that the cardinality would make sense.  In this case, defaulting BAR and TICK might make sense for temporal with `timeUnit`.  I think that for quantitative dimension like in QxQ if we make one dimension ordinal, but not another that's super weird.  (I don't know about TxT -- would people ever make that?)

kanitw: In any case, this is breaking `bin` for sure.  (Bin should use linear scale even for `bar`/`tick`)

src/compile/scale.ts
```diff
@@ -65,31 +69,24 @@ export function type(fieldDef: FieldDef, channel: Channel, mark: Mark): string {
...
       }
-      if (contains([ROW, COLUMN, SHAPE], channel)) {
-        return 'ordinal';
-      }
+
       if (fieldDef.scale.type !== undefined) {
         return fieldDef.scale.type;
       }
 
+      if (contains([BAR, TICK], mark)) {
```
kanitw: What if you don't have `timeUnit`?

domoritz: if you use bar or tick, then ordinal, otherwise time.

kanitw:
To be honest I'm leaning toward not having ordinal by default for bar.
This really depends on how we think about it.

If we don't assume that users are design-savvy.   Bar/Tick should use `ordinal` only if the cardinality of the dimension is low.  Otherwise, use `time`.  Since we don't know cardinality, I am leaning toward making it consistent with other marks. 

However, if we assume that users are **often** design-savvy and they would only use bar/tick for this if they know that the cardinality would make sense.  In this case, defaulting BAR and TICK might make sense for temporal with `timeUnit`.  (But what if it is TxT? -- do we make it ordinal for both axes?)

kanitw:
And since I don't want to assume that users are design savvy, I would prefer to remove this complexity.  (So we don't have to explain our debatable rationale in the docs.)

domoritz:
Hmm, the issue is that we use the type to generate the right parser and formatter. To do what you suggested, we would need to a) add a `rawType` property or b) add a `ordinal-time` type.

kanitw:
Can you explain _where_ and _why_ we would need that ?

- Begin with "we use the type:" -- scale type or data type? 
- I don't know what you mean `rawType`.  For `ordinal-time`, isn't checking that the scale type is `ordinal` and the data type is `temporal` sufficient?

domoritz:
Sorry, I totally messed up scale types and data types.

domoritz:
Done

kanitw:
:tada:

kanitw:
Fixes #1013

PR #1043, opened 2016-02-01
Add a scatter_aggregate example

ZeningQu:
Add a scatter_aggregate example: one axis is aggregated, the other axis is binned.


kanitw:
@ZeningQu  Thanks for your contribution.  However, we're becoming more selective in terms of adding examples as we're closer to release.  If you have bandwidth for contribution, ping me on slack and I can point you to issues that we need help :)

ZeningQu:
@kanitw Yes please! Which other issues would help better? Is there a slack team I could join?

PR #1044, opened 2016-02-01
Add scatter_aggregate_axis_color example

ZeningQu:


kanitw:
This visualization tries to show two different things at the same time.  It is doubtful if this is the most effective way to show this information.  Therefore, I don't think it should be a part of the example gallery.

Issue #1051, opened 2016-02-03
Support value for all channels


kanitw:
Based on conversation in #1029, fieldDef's `value` refers to a visual domain value.  
Meanwhile, once we have composition, 
- [ ] ~~add fieldDef's  `datum` that refers to a data domain value.  (Therefore produces `value` without `scale` in Vega)~~
- [ ] make `x` and `y` supports value as well.

kanitw:
(adding this to description)
- [ ] make `x` and `y` supports value as well.

kanitw:
Since we plan to do #1340, we don't have to add `datum` anymore.

kanitw:
Done as a part of mark refactor to use common valueref.ts

PR #1052, opened 2016-02-03
Bring back `encoding.<channel>.value`

### LINKED ISSUE #1029: Moving default values of `value` to `config`
The reason that we have color and size channel automatically maps to `fill`/`stroke` and `size` / `width` / `height` of tick and bar / `fontSize` is that we allow seamless transition between mark types.

Suppose a field is mapped to `bar`'s `color` (practically to bar’s `fill`), if users change the mark type to line, this allow the field to automatically be mapped to line’s `stroke`.  

Similarly, suppose a field is mapped to `point`'s `size` (practically to point's `size`), if users change the mark type to `text`, this allow the field to automatically be mapped to text's `fontSize`.  

And then we added `value` field to allow setting constant values for these fields.  

However, having `value` property that seamlessly work across mark type is not always sensible.  For example, if user explicitly set `size` to 96 when using `point` mark.  This does not mean that user would want `fontSize=96` if the user changes mark to `text`.  

Therefore, we should instead move all `value` to `config.mark` so that we can have separate config for `config.size`, `config.barWidth`, `config.tickWidth`, `config.fontSize` respectively.  For `config.mark.color`, we can maintain similar behavior to `encoding.channel`.  Basically, if `config.mark.color` is set but `config.mark.fill` and `config.mark.stroke` is not,  then the color config is automatically applied to both fill and stroke.   (Explicitly specifying `fill` or `stroke` config will override `color` config.)

In addition, as @mcorrell point out, user might wonder when would a value be in encoding channel and when would it be in config.   This move will allow us to add other visual properties as an encoding channel in the future without deprecating their config.  

This proposal could work perfectly under the current model. 

One exception that we might need to bring value back in the future is only for `x` and `y`.  With layering, one sensible thing we might do is to add a reference line.  For a horizontal line, `y` channel should have a constant value.   I don’t know if it makes sense to set the reference line’s y value in `config.mark.y` because this doesn’t feel like config to me. 

@domoritz thoughts?

kanitw:
I add separate case for value instead of using `||` with config because in the future, value might refer to parent.  

Fixes  #1029.  (For `datum`, I'm adding #1051 for doing later with composition)

### DIFF
--- a/site/docs/encoding.md
+++ b/site/docs/encoding.md
@@ -69,7 +69,7 @@ Here is a list of properties for the field definition object:
 | Property      | Type          | Description    |
 | :------------ |:-------------:| :------------- |
 | field         | String        | Name of the field from which to pull a data value.    |
-| value         | String &#124; Number | A constant value. |
+| value         | String &#124; Number | A constant value in visual domain. |
 | type          | String        | Data type of the field.  This property accepts both a full type name (`"quantitative"`, `"temporal"`, `"ordinal"`,  and `"nominal"`), or an initial character of the type name (`"Q"`, `"T"`, `"O"`, `"N"`).  This property is case insensitive.|
 
 {:#types}
--- a/src/compile/Model.ts
+++ b/src/compile/Model.ts
@@ -227,6 +227,10 @@ export class Model {
   }
 
   public sizeValue(channel: Channel = SIZE) {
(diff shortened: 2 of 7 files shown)


arvind:
🎉

kanitw:
@arvind Feel free to review this PR as well :)

PR #1055, opened 2016-02-03
Flexible x-scale's `bandWidth` for text on ordinal + remove `config.textCellWidth`

### LINKED ISSUE #724: Revise rules for marks

kanitw:
- Make `x`'s `bandWidth` 90 if `mark` = `text` and `x` is ordinal scale by default  (otherwise `bandWidth` = `21` for `x` and `y`)
- Do not refer to x&y's `bandWidth` when it doesn't exist.  (And add TODO refer to the future `config.scale.bandWidth` / `config.scale.textBandWidth`)
- Remove `config.textCellWidth`, which is out of place

Fix #724

### DIFF
--- a/src/compile/layout.ts
+++ b/src/compile/layout.ts
@@ -41,9 +41,9 @@ function getCellWidth(model: Model): LayoutValue {
     return model.config().cell.width;
   }
   if (model.mark() === TEXT_MARK) {
-    return model.config().textCellWidth;
+    return 90; // TODO: config.scale.textBandWidth
   }
-  return model.fieldDef(X).scale.bandWidth;
+  return 21; // TODO: config.scale.bandWidth
 }
 
 function getWidth(model: Model, cellWidth: LayoutValue): LayoutValue {
@@ -61,7 +61,7 @@ function getCellHeight(model: Model): LayoutValue {
       return model.config().cell.height;
     }
   }
-  return model.fieldDef(Y).scale.bandWidth;
+  return 21 /* config.scale.bandWidth */;
 }
 
 function getHeight(model: Model, cellHeight: LayoutValue): LayoutValue {
--- a/src/compile/Model.ts
+++ b/src/compile/Model.ts
@@ -58,6 +58,18 @@ export class Model {
         fieldDef.legend = instantiate(legendSchema);
       }
 
+      // set default bandWidth for X and Y
+      if (channel === X && fieldDef.scale.bandWidth === undefined) {
+        // This should be zero for the sake of text table.
+        fieldDef.scale.bandWidth = this.isOrdinalScale(X) && this.mark() === 'text' ?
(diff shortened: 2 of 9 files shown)


domoritz:
Cool. Will merge when tests pass.

src/compile/layout.ts
```diff
@@ -61,7 +61,7 @@ function getCellHeight(model: Model): LayoutValue {
       return model.config().cell.height;
     }
   }
-  return model.fieldDef(Y).scale.bandWidth;
+  return 21 /* config.scale.bandWidth */;
```
domoritz: You're going to change all these constants, I assume.

kanitw: They should refer to config.scale.width in the future.   Now it doesn't make sense to add config.scale unless we want to migrate all properties there.

domoritz: Wait, so what happened to `model.fieldDef(Y).scale.bandWidth`?

kanitw: It was original set in combination with positionFieldDef's definition.   (Have to be the same for both x and y) 
  
But now x scale has flexible bandWidth and going forward, we should move these default values to `config.scale.*` anyway so I decide not to bother with `encoding.x|y.scale` `s default.

kanitw:
@domoritz  test now pass.  :) 

(And yes, I amend and force push.  Let's do this going forward when make sense.  Apologies for being dumb about this for a year!)

src/compile/layout.ts
domoritz: Let's move them now. Having the same constant everywhere really scares me.

kanitw: If that really scares you, it should still be done in a separate PR as that involves a lot more things.  
See #1056 about facet's scale/axis.  I don't think that belongs here and it's much messier PR than here.  

If you take a look closely, it's just 21 everywhere except a few places where it should be 90.  So it is not that many numbers.  (Just two)

domoritz: Oh, I'm aware that it's only 2 numbers but they are in 16 different places. Will you send a follow up PR?

kanitw: I also added comment everywhere so everyone is warned not to change them!

domoritz: Yep, saw that. Do you want to send a follow up PR?

kanitw: No time to do it yet...

Issue #992, opened 2016-01-24
Generate reference to unknown dataset


domoritz:
```json
{
  "description": "A scatterplot showing horsepower and miles per gallons.",
  "data": {"url": "data/cars.json"},
  "mark": "point",
  "encoding": {
    "x": {"field": "Horsepower","type": "quantitative"},
    "y": {"field": "Miles_per_Gallon","type": "quantitative"},
    "color": {"field": "Origin", "type": "ordinal"}
  }
}
```

The generated vega code references layout but layout does not exist. This should be easy to fix by adding the reference only when it is needed. The vega spec actually never used the referenced dataset.

domoritz:
Started at https://github.com/vega/vega-lite/compare/dom/fix-992

kanitw:
lol is "started at" a new pattern?  I think normally you'd create a PR and says WIP for this issue.

domoritz:
Came out of a conversation with @willium. He's adding tests for it.

kanitw:
Pretty much done in #1059 -- only minor comment left.

PR #1059, opened 2016-02-03
Fix incorrect assignment of layout as a data source (Issue #992)

willium:
Fixes bug in #992 where non-ordinal marks are incorrectly set to use layout as a data source in the root group

### DIFF
--- a/test/compile/compile.test.ts
+++ b/test/compile/compile.test.ts
@@ -1,0 +1,56 @@
+/* tslint:disable:quotemark */
+
+import {assert} from 'chai';
+import {parseModel} from '../util';
+import * as compile from '../../src/compile/compile';
+
+describe('Compile', function() {
+
+  describe('compileRootGroup()', function() {
+
+    describe('non-ordinal', function() {
+      const model = parseModel({
+        "description": "A scatterplot showing horsepower and miles per gallons.",
+        "data": {"url": "data/cars.json"},
+        "mark": "point",
+        "encoding": {
+          "x": {"field": "Horsepower","type": "quantitative"},
+          "y": {"field": "Miles_per_Gallon","type": "quantitative"},
+        }
+      });
+
+      const rootGroup = compile.compileRootGroup(model);
+
+      it('should not refer to layout data', function() {
+        assert.notDeepEqual(rootGroup.from, {"data": "layout"});
+      });
+
+    });
+
+    describe('ordinal', function() {
+      const model = parseModel({
+        "description": "A simple bar chart with embedded data.",
+        "data": {
+          "values": [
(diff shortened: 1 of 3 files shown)


test/compile/compile.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        "encoding": {
+          "x": {"field": "Horsepower","type": "quantitative"},
+          "y": {"field": "Miles_per_Gallon","type": "quantitative"},
+          "color": {"field": "Origin", "type": "ordinal"}
+        }
+      });
+
+      const rootGroup = compile.compileRootGroup(model);
+
+      it('should not have layout dataref', function() {
```
kanitw: very minor:  maybe  "should refer to layout data" ?

test/compile/compile.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        "mark": "bar",
+        "encoding": {
+          "x": {"field": "a", "type": "ordinal"},
+          "y": {"field": "b", "type": "quantitative"}
+        }
+      });
+
+      const rootGroup = compile.compileRootGroup(model);
+
+      it('should have layout dataref', function() {
```
kanitw: Same here (maybe  "should refer to layout data" ?)

kanitw:
LGTM, feel free to merge after addressing minor comments.

test/compile/compile.test.ts
willium: done.

willium:
These tests could easily be more thorough and theres other things to test in compile (it's probably a good idea to do), but I didn't want to chase down every edge case when we're fairly sure it works and this is relatively harmless compared to other potential tests. We can improve coverage later?

test/compile/compile.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    describe('non-ordinal', function() {
+      const model = parseModel({
+        "description": "A scatterplot showing horsepower and miles per gallons.",
+        "data": {"url": "data/cars.json"},
+        "mark": "point",
+        "encoding": {
+          "x": {"field": "Horsepower","type": "quantitative"},
+          "y": {"field": "Miles_per_Gallon","type": "quantitative"},
+          "color": {"field": "Origin", "type": "ordinal"}
```
domoritz: color channel is not needed, right?

domoritz:
Yeah, I think that's fine. To make this function more testable, we should probably extract the lines that get the width and height anyway. but let's defer that.

kanitw:
Yeah, don't test the top-level compile too much.   This will change vastly in the upcoming month to support composition.

test/compile/compile.test.ts
willium: removed and rebased

PR #1065, opened 2016-02-04
Will/docs

willium:
Fixes syntax highlight coloring:
- [x] matched styles between **Pygments** and **Highlight.js**
- [x] converted all pygments code blocks to JSON type
- [x] created explicit `.suppress-errors` class for **Pygments** snippets to be used when adding invalid JSON (`...`, `//comments`, etc)
- [x] converted examples in *encoding.md* to new embed script method (**Highlight.js**) used by all other docs pages


domoritz:
Excellent. Can you make sure we can merge?

examples/scatter_color_shape_constant.json
```diff
@@ -0,0 +1,10 @@
+{
```
kanitw: examples/docs?  This one probably isn't interesting enough to be in galllery?

examples/bar_aggregate_size.json
```diff
@@ -0,0 +1,19 @@
+{
```
kanitw: examples/docs

PR #1066, opened 2016-02-04
Definition lists

domoritz:


site/static/main.css
```diff
@@ -147,6 +150,10 @@ h2:hover .anchor .octicon-link, h3:hover .anchor .octicon-link, h4:hover .anchor
   visibility: visible;
 }
 
+dt {
```
domoritz: @kanitw repeated from above

kanitw: Why not just fix it then?

domoritz: Wasn't sure whether you were done.

PR #1068, opened 2016-02-04
Add `order` channel + `config.mark.stacked`

### LINKED ISSUE #993: Revise Sorting Syntax
Both actually encodes map data to the encoding so I think I was wrong i have them as config.  

- `config.mark.sortBy` —> This is essentially the `layer`/`z` channel.  
- `config.mark.sortLineBy` —> This is essentially the `path` channel.   (For a reference point, Tableau has `path` channel for exactly this reason. )

### LINKED ISSUE #1032: Stack config could become an enum?
Since stack's`sort` should be moved into encoding (#993), we only have one property left in the stack (`offset`).  

Should we just make `config.stack` an enum:  'normalize', 'zero', 'center', 'none'?

kanitw:
- Add `order` channel for sorting layer/stack order,  remove `detail.sort`, `config.stack.sort`  Fix #993 

- Add `config.mark.stacked`,  remove `config.stack`.   Fix #1032


src/schema/config.marks.schema.ts
```diff
@@ -4,6 +4,7 @@ export interface MarkConfig {
   color?: string;
   barWidth?: number;
   tickWidth?: number;
+  stacked?: string;
```
domoritz: Could use string literal here :-P

kanitw: true.   Later :)

PR #1071, opened 2016-02-04
Consolidate the submenu logic

domoritz:


domoritz:
I think this should be merged as is. Let's make a new PR for sub sub menus or side menus.

willium:
LGTM!

PR #1073, opened 2016-02-05
Ordinal color scales

domoritz:
screen shot 2016-02-06 at 14 10 05
screen shot 2016-02-06 at 14 10 12
screen shot 2016-02-08 at 17 51 26

### DIFF
--- a/src/compile/data.ts
+++ b/src/compile/data.ts
@@ -218,22 +218,17 @@ export namespace source {
   }
 
   export function rankTransform(model: Model) {
-    const fieldDef = model.fieldDef(COLOR);
-    const rankColor = model.has(COLOR) && fieldDef.type === ORDINAL;
-    const binColor = model.has(COLOR) && fieldDef.bin;
-    const field = model.field(COLOR);
+    const rankColor = model.has(COLOR) && model.fieldDef(COLOR).type === ORDINAL;
+
     return rankColor ? [{
       type: 'sort',
-      by: field
+      by: model.field(COLOR)
     },{
       type: 'rank',
-      field: field,
+      field: model.field(COLOR),
       output: {
         rank: model.field(COLOR, {prefn: 'rank_'})
       }
-    }] : binColor ? [{
-      type: 'sort',
-      by: model.field(COLOR, {suffix: '_start'})
     }] : [];
   }
 }
--- a/src/compile/data.ts
+++ b/src/compile/data.ts
@@ -218,7 +218,8 @@ export namespace source {
     }, []);
   }
 
-  // We need to add a rank transform as a work around for color ramps
+  // We need to add a rank transform so that we can use the rank value as
+  // input for color ramp's linear scale.
   export function rankTransform(model: Model) {
     const rankColor = model.has(COLOR) && model.fieldDef(COLOR).type === ORDINAL;
 
--- a/src/compile/data.ts
+++ b/src/compile/data.ts
@@ -166,6 +166,7 @@ export namespace source {
         }
 
         transform.push(binTrans);
+        // color ramp has type linear or time
         if (scaleType(fieldDef, channel, model.mark()) === 'ordinal' || channel === COLOR) {
           transform.push({
             type: 'formula',
@@ -218,6 +219,8 @@ export namespace source {
   }
 
   export function rankTransform(model: Model) {
+    // We need to add a rank transform as a work around for color ramps
+
     const rankColor = model.has(COLOR) && model.fieldDef(COLOR).type === ORDINAL;
 
     return rankColor ? [{
--- a/src/compile/data.ts
+++ b/src/compile/data.ts
@@ -19,8 +19,6 @@ const DEFAULT_NULL_FILTERS = {
   temporal: true
 };
 
-export const COLOR_LABEL_FIELD = '_color_label';
-
 /**
  * Create Vega's data array from a given model.
  *
@@ -168,12 +166,12 @@ export namespace source {
         }
 
         transform.push(binTrans);
-        if (scaleType(fieldDef, channel, model.mark()) === 'ordinal') {
+        if (scaleType(fieldDef, channel, model.mark()) === 'ordinal' || channel === COLOR) {
           transform.push({
             type: 'formula',
             field: field(fieldDef, {binSuffix: '_range'}),
             expr: field(fieldDef, {datum: true, binSuffix: '_start'}) +
-                  '+ \'-\' +' +
+                  ' + \'-\' + ' +
                   field(fieldDef, {datum: true, binSuffix: '_end'})
           });
         }
@@ -223,20 +221,19 @@ export namespace source {
     const fieldDef = model.fieldDef(COLOR);
     const rankColor = model.has(COLOR) && fieldDef.type === ORDINAL;
     const binColor = model.has(COLOR) && fieldDef.bin;
-    const field = model.field(COLOR, {noRank: true});
+    const field = model.field(COLOR);
     return rankColor ? [{
       type: 'sort',
       by: field
     },{
       type: 'rank',
-      field: field
+      field: field,
+      output: {
+        rank: model.field(COLOR, {prefn: 'rank_'})
+      }
     }] : binColor ? [{
       type: 'sort',
-      by: vlFieldDef.field(fieldDef, {suffix: '_start'})
-    },{
-      type: 'formula',
-      field: COLOR_LABEL_FIELD,
-      expr: vlFieldDef.field(fieldDef, {suffix: '_start', datum: true}) + '+ " - " + ' + vlFieldDef.field(fieldDef, {suffix: '_end', datum: true})
+      by: model.field(COLOR, {suffix: '_start'})
     }] : [];
   }
 }
--- a/src/compile/data.ts
+++ b/src/compile/data.ts
@@ -5,10 +5,10 @@ import {FieldDef} from '../schema/fielddef.schema';
 import {StackProperties} from './stack';
 
 import {autoMaxBins} from '../bin';
-import {Channel, X, Y, ROW, COLUMN} from '../channel';
+import {Channel, X, Y, ROW, COLUMN, COLOR} from '../channel';
 import {SOURCE, STACKED_SCALE, LAYOUT, SUMMARY} from '../data';
 import {field} from '../fielddef';
-import {QUANTITATIVE, TEMPORAL} from '../type';
+import {QUANTITATIVE, TEMPORAL, ORDINAL} from '../type';
 import {type as scaleType} from './scale';
 import {parseExpression, rawDomain} from './time';
 
@@ -18,6 +18,8 @@ const DEFAULT_NULL_FILTERS = {
   quantitative: true,
   temporal: true
 };
+
+export const COLOR_LABEL_FIELD = '_color_label';
 
 /**
  * Create Vega's data array from a given model.
@@ -124,7 +126,8 @@ export namespace source {
       formulaTransform(model),
       filterTransform(model),
       binTransform(model),
-      timeTransform(model)
+      timeTransform(model),
+      rankTransform(model)
     );
   }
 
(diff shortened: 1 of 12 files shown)


src/compile/legend.ts
```diff
@@ -6,13 +6,16 @@ import {AREA, BAR, TICK, TEXT, LINE, POINT, CIRCLE, SQUARE} from '../mark';
...
+import {INVERSE_RANK, COLOR_LABEL} from './scale';
 
 export function compileLegends(model: Model) {
   var defs = [];
 
   if (model.has(COLOR) && model.fieldDef(COLOR).legend) {
+    const fieldDef = model.fieldDef(COLOR);
     defs.push(compileLegend(model, COLOR, {
-      fill: model.scaleName(COLOR)
+      fill: fieldDef.type === ORDINAL || fieldDef.bin ? INVERSE_RANK : model.scaleName(COLOR)
```
kanitw: Put parentheses to highlight precedence   (I don't know if `||` or `?` has higher precedence.)

src/compile/scale.ts
```diff
@@ -347,11 +373,10 @@ export function rangeMixins(model: Model, channel: Channel, scaleType: string):
       return {range: 'shapes'};
     case COLOR:
       if (fieldDef.type === NOMINAL
-        || fieldDef.type === ORDINAL // FIXME remove this once we support color ramp for ordinal
       ) {
```
kanitw: remove line break :)

src/fielddef.ts
```diff
@@ -20,6 +20,8 @@ export interface FieldRefOption {
   binSuffix?: string;
   /** append suffix to the field ref (general) */
   suffix?: string;
+  /** return the base field, not rank */
+  noRank?: boolean;
```
kanitw: I'm curious if we could simply do `prefn: 'rank_'`  for the case where we need rank 
or at least do `rank` rather than `noRank`.   That should easily solve the problem I mentioned earlier.

domoritz: Hmm, the problem I see is that we have way more places where we want rank by default and only a handful of places where we do not want the rank.

kanitw: But when people read the code for `model.field`, they would not expect that color returns `rank`.   If we use `prefn:` then it is explicit that in that case it returns something different.   And in the normal case (despite used less frequently), it returns the field normally.

domoritz:
All working. Yaay.

src/compile/data.ts
```diff
@@ -215,6 +216,21 @@ export namespace source {
       return transform;
     }, []);
   }
+
+  export function rankTransform(model: Model) {
```
kanitw: Add comment that this add rank transform for the color ramp as a workaround (until the future color ramp support in Vega).

src/compile/legend.ts
```diff
@@ -6,13 +6,16 @@ import {AREA, BAR, TICK, TEXT, LINE, POINT, CIRCLE, SQUARE} from '../mark';
...
+import {INVERSE_RANK, COLOR_LABEL} from './scale';
 
 export function compileLegends(model: Model) {
   var defs = [];
 
   if (model.has(COLOR) && model.fieldDef(COLOR).legend) {
+    const fieldDef = model.fieldDef(COLOR);
     defs.push(compileLegend(model, COLOR, {
-      fill: model.scaleName(COLOR)
+      fill: (fieldDef.type === ORDINAL || fieldDef.bin) ? INVERSE_RANK : model.scaleName(COLOR)
```
kanitw: I assume that since temporal uses time scale, this will be correct for them.  

(If so, I guess this will still be incorrect for month, day, hour which has ordinal scale and thus need the same hack?)

src/compile/mark-text.ts
```diff
@@ -15,7 +15,10 @@ export namespace text {
       y: { value: 0 },
       width: { field: { group: 'width' } },
       height: { field: { group: 'height' } },
-      fill: { scale: model.scaleName(COLOR), field: model.field(COLOR) }
+      fill: {
+        scale: model.scaleName(COLOR),
+        field: model.field(COLOR, model.fieldDef(COLOR).type === ORDINAL ? {prefn: 'rank_'} : {})
```
kanitw: Just curious .. why isn't `bin` treated like ordinal in this case?

src/compile/scale.ts
```diff
@@ -43,8 +47,35 @@ export function compileScales(channels: Channel[], model: Model) {
         }
       });
 
-      return scaleDef;
-    });
+      if (channel === COLOR && (fieldDef.type === ORDINAL || fieldDef.bin)) {
+        scales.push({
```
kanitw: Add comment that this scale provides mapping for the rank field back to original data for legend labels.

src/compile/scale.ts
```diff
@@ -43,8 +47,35 @@ export function compileScales(channels: Channel[], model: Model) {
...
+          name: INVERSE_RANK,
+          type: ORDINAL,
+          domain: {data: model.dataTable(), field: model.field(COLOR, fieldDef.bin ? {} : {prefn: 'rank_'}), sort: true},
+          range: {data: model.dataTable(), field: model.field(COLOR), sort: true}
+        });
+
+        if (fieldDef.bin) {
```
kanitw: Why does bin require two scales?  (Add comment to explain)

src/compile/scale.ts
```diff
@@ -66,11 +97,12 @@ export function type(fieldDef: FieldDef, channel: Channel, mark: Mark): string {
     case NOMINAL:
       return 'ordinal';
     case ORDINAL:
+      if (channel === COLOR) {
+        return 'linear'; // time has order, so use interpolated ordinal color scale.
```
kanitw: The comment here looks out of place.  

Better comment that this return scale type for one of a few scales (2 for ordinal, 3 for bin) that color ramp scale uses  + why should `scale.type` returns the linear one.

kanitw: (`scale.type` is currently being called by other files to satisfy some condition, can't remember what are those, but what this method returns might have implication for other parts of the codebase.

src/compile/data.ts
```diff
@@ -165,12 +166,12 @@ export namespace source {
         }
 
         transform.push(binTrans);
-        if (scaleType(fieldDef, channel, model.mark()) === 'ordinal') {
+        if (scaleType(fieldDef, channel, model.mark()) === 'ordinal' || channel === COLOR) {
```
kanitw: Probably worth add comment that `COLOR` ramp has `linear` type for the main scale.

src/compile/scale.ts
kanitw: + .. while the main scale provides mapping from ... to ...

kanitw:
Yaay.  I think we should add more comments because this is so complicated and without comment, only you would understand all of it.

src/compile/legend.ts
kanitw: Maybe the right condition if the scale type is ordinal or it is binned.

domoritz: This logic is getting ridiculous.

src/compile/mark-text.ts
domoritz: i'm glad you asked. For bin, I actually generate 3 scales. try the examples to see it. We don't generate a rank.

src/compile/legend.ts
domoritz: > Maybe the right condition if the scale type is ordinal or it is binned.

Definitely not because we use a linear scale for colors.

domoritz:
All working. Let me know if we need more comments.

src/compile/data.ts
kanitw: Here?

src/compile/scale.ts
```diff
@@ -43,8 +47,43 @@ export function compileScales(channels: Channel[], model: Model) {
...
+        });
+
+        // bin needs an additional scale for labels because we want labels in legends
+        if (fieldDef.bin) {
+          scales.push({
+            name: COLOR_LABEL,
+            type: ORDINAL,
+            domain: {
+              data: model.dataTable(),
+              field: model.field(COLOR,  (fieldDef.bin || fieldDef.timeUnit) ? {} : {prefn: 'rank_'}), sort: true
```
kanitw: It is always `fieldDef.bin` on this line.

src/compile/scale.ts
```diff
@@ -43,8 +47,43 @@ export function compileScales(channels: Channel[], model: Model) {
...
-      return scaleDef;
-    });
+      // add additional scales needed to support ordinal color scales also for legends
+      if (channel === COLOR && fieldDef.legend && (fieldDef.type === ORDINAL || fieldDef.bin || fieldDef.timeUnit)) {
+        scales.push({
+          name: INVERSE_RANK,
+          type: ORDINAL,
+          domain: {
+            data: model.dataTable(),
+            field: model.field(COLOR, (fieldDef.bin || fieldDef.timeUnit) ? {} : {prefn: 'rank_'}), sort: true
```
kanitw: // use `rank_<field>` for ordinal type, for bin and timeUnit use default field

src/compile/scale.ts
```diff
@@ -43,8 +47,43 @@ export function compileScales(channels: Channel[], model: Model) {
...
+          },
+          range: {data: model.dataTable(), field: model.field(COLOR), sort: true}
+        });
+
+        // bin needs an additional scale for labels because we want labels in legends
```
kanitw: "because we need to map bin_start to bin_range in legends"

test/compile/scale.test.ts
```diff
@@ -258,6 +258,86 @@ describe('Scale', function() {
...
+
+    it('should sort range of color labels', function() {
+      assert.deepEqual(scales[1].range, {
+        data: 'source',
+        field: 'bin_origin_range',
+        sort: {"field": "bin_origin_start","op": "min"}
+      });
+    });
+  });
+
```
kanitw: Missing test for timeUnit.   I'd say add least let's add test scaffolding, so later someone can add additional test for it.

src/compile/data.ts
domoritz: See one line lower :-P

test/compile/scale.test.ts
domoritz: I added the tests

src/compile/scale.ts
domoritz: Done

domoritz:
Done

src/compile/scale.ts
```diff
@@ -43,8 +51,46 @@ export function compileScales(channels: Channel[], model: Model) {
...
+      if (channel === COLOR && fieldDef.legend && (fieldDef.type === ORDINAL || fieldDef.bin || fieldDef.timeUnit)) {
+        scales.push({
+          name: INVERSE_RANK,
+          type: ORDINAL,
+          domain: {
+            data: model.dataTable(),
+            // use rank_<field> for ordinal type, for bin and timeUnit use default field
+            field: model.field(COLOR, (fieldDef.bin || fieldDef.timeUnit) ? {} : {prefn: 'rank_'}), sort: true
+          },
+          range: {data: model.dataTable(), field: model.field(COLOR), sort: true}
```
kanitw: I'm now confused.   For bin and timeUnit case, isn't domain and range the same now?  (No parameter passed to model.field for both of them.)

src/compile/scale.ts
kanitw: Hmm instead of because .... because ... 

can you have one "because": 

"because we need to map bin_start to bin_range in legend's "labels

src/compile/scale.ts
```diff
@@ -11,10 +11,18 @@ import {SOURCE, STACKED_SCALE} from '../data';
 import {NOMINAL, ORDINAL, QUANTITATIVE, TEMPORAL} from '../type';
 import {Mark, BAR, TEXT as TEXT_MARK} from '../mark';
 import {rawDomain} from './time';
+import {field} from '../fielddef';
+
+// scale from rank to field for ordinal color scale,
+// ordinal field to field mapping for binned quantitative and temporal scale with time unit
```
kanitw: This line reads confusing.

src/compile/scale.ts
kanitw: Looks like you never use INVERSE_RANK for bin and timeUnit case except in legend.ts  but I don't think it is necessary.

src/compile/scale.ts
kanitw: "ordinal field to field"

domoritz: So for ordinal color scale: rank to field
For bin and temporal: field to field

src/compile/scale.ts
domoritz: What do you mean? What is not necessary?

kanitw: This INVERSE_RANK scale for bin and timeUnit is not necessary because you do not have rank for these two cases so you do not have to invert. 

In fact, your current code for INVERSE_RANK produces identity scale (`f(x)=x`) for bin & timeUnit cases so I think you can simplify a lot of code here.

domoritz: I admit that the name is misleading and I can change it to identity for bin and time unit. However, i'd love to know how you add legends without an identity scale. The code would get more complicated if I had separate cases for ordinal and bin/time unit but I could rename INVERSE_RANK to INVERSE to make it less confusing.

kanitw:
@domoritz  I added clarifying comments and rename these scales.  Feel free to merge if you like them.

PR #1081, opened 2016-02-07
fixed broken link for tutorials

ayush29feb:


kanitw:
Thank you for catching this error.  

Could you also please submit a new PR for merging into master?  (That way we don't lose the change when we publish gh-pages next time. :)

PR #1083, opened 2016-02-07
Add missing examples to encoding.md + Minor Editing

kanitw:
It is worth noting that using `detail` doesn't make much sense without labeling / tooltips.   (No way to know which mark represent which group).  Similarly, we don't know which part of the connected scatter plot is the beginning of the path.  

So I don't know if actually should hide the docs for these until we ship either labeling or tooltips?

### DIFF
--- a/examples/docs/scatter_color_order.json
+++ b/examples/docs/scatter_color_order.json
@@ -5,6 +5,6 @@
     "x": {"field": "Horsepower", "type": "quantitative"},
     "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
     "color": {"field": "Origin", "type": "nominal"},
-    "order": {"field": "Origin", "type": "nominal", "sort": "descending"}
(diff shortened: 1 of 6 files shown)


examples/docs/scatter_color_order.json
```diff
@@ -0,0 +1,10 @@
+{
+  "data": {"url": "data/cars.json"},
+  "mark": "point",
+  "encoding": {
+    "x": {"field": "Horsepower", "type": "quantitative"},
+    "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
+    "color": {"field": "Origin", "type": "nominal"},
+    "order": {"field": "Origin", "type": "nominal", "sort": "descending"}
```
domoritz: Wait, how can you order nominal? This should not be allowed.

kanitw:
@domoritz this branch is done ... I think

Issue #817, opened 2015-12-02
Add “Unit” so we automatically append units to axis & legend title.


kanitw:
In an attempt to make axis title nicer, maybe we add `fieldDef.unit` or allow describing unit in the schema once we support it (#512).  

For example,  if `fieldDef.unit` is `’$’` for mean of `’cost`’,   the axis title could be `’MEAN(cost) (in $)’`.

ayush29feb:
@kanitw is this what you want to?
![image](https://cloud.githubusercontent.com/assets/8674376/12534195/d307ad12-c202-11e5-923e-ba41212588bf.png)

domoritz:
We already use a similar syntax for aggregation functions. I think there are two options here. One is to write "Year in years" and the other is to show the unit with the data on the axis (acceptable for short units such as $).

kanitw:
Good question! Here is what I think:

- `unit` as part of a field can be considered equivalent to having `unit` included from the schema.
- We could introduce `showUnit` as a property to axis and have these possible values: `'none'`, `'title'` (Showing "Year in years" for the axis title), `'label-prefix'` (Showing $100), 'label-suffix` ('Show 100 kg).

@domoritz -- what do you think?

domoritz:
Sounds reasonable.

kanitw:
Cool.  To summarize, 

- [ ] add `unit` to `fieldDef` (interface & schema)
- [ ] add `showUnit` as enum to `axis` (interface & schema).  This should be an enum with four values above.  The default should be `'title'` IMHO.  See `fieldDef.schema.ts`'s `type` for an example of how we declare enum and see `config.schema.ts` for how to declare default. 
   - [ ] For `'title'` case, modify  `model.fieldTitle()`  
   - [ ] For `label-prefix/suffix`, need to modify `axis.ts`'s  `properties.labels`.  Probably need to use output's `text`'s `template` property.

ayush29feb:
I am a little confused why we have the `showUnit` in `fieldDef.axis` and `unit` in `fieldDef`
```js
"encoding": {
    "y": {
      "field": "b2",
      "type": "quantitative",
      "unit": "$",
      "axis": {"showUnit": "label-prefix"}
    },
    "x": {"field": "a","type": "ordinal"}
  }
```

it seems a little weird to define that the `unit` would be in `title` inside `axis`

ayush29feb:
[...]

PR #1089, opened 2016-02-10
[WIP] Working on #817 : Custom Units

ayush29feb:
Okay, so far I have implemented
- Axis Unit
   - Title
   - Label Suffix/Prefix (messes up the formatting)
- Legend Unit
   - Title

The axis suffix/prefix messes up with the date/time formating for the labels. My guess is to do some research in vega again to see how it uses the `text.template`

The Legend labels are not generated in `vega-lite` but it is done by `vega`, and I need to do some reading to understand that. Working on it.

TODO
- [ ] Fix Date/Time Formating for axis label-suffix/prefix units
- [ ] Add labe-suffix/prefix to legend
- [ ] Update the docs


src/compile/axis.ts
```diff
@@ -177,22 +178,32 @@ export namespace properties {
...
+    let textTemplate = '{{ datum.data }}';
+
     if (contains([NOMINAL, ORDINAL], fieldDef.type) && axis.labelMaxLength) {
       // TODO replace this with Vega's labelMaxLength once it is introduced
-      labelsSpec = extend({
-        text: {
-          template: '{{ datum.data | truncate:' + axis.labelMaxLength + '}}'
-        }
-      }, labelsSpec || {});
+      textTemplate = '{{ datum.data | truncate:' + axis.labelMaxLength + '}}';
```
ayush29feb: I need to figure out how vega uses these text templates. Because the text right now looses the date/time formatting.

Also on a side note shouldn't there be a unit test that checks if the units are being formatted for date/time

src/fielddef.ts
```diff
@@ -112,14 +112,20 @@ export function cardinality(fieldDef: FieldDef, stats, filterNull = {}) {
     (stat.missing > 0 && filterNull[type] ? 1 : 0);
 }
 
-export function title(fieldDef: FieldDef) {
+export function title(fieldDef: FieldDef, showUnit: boolean) {
   if (isCount(fieldDef)) {
     return COUNT_DISPLAYNAME;
   }
+
+  var unit = '';
```
ayush29feb: Let me know if this is good style. I can try to do something more clean, or if you have any suggestions

src/compile/axis.ts
kanitw: There should eventually be. :)

kanitw: As mentioned in a slack conversation, Vega might use not use text template but rather calling formatting function directly.    

First, take a look why I think that to be the case.  :)

src/fielddef.ts
kanitw: Normally I prefer:
```
const unit =  fieldDef.unit && showUnit ? ' in ' + fieldDef.unit : '';
```

src/compile/axis.ts
```diff
@@ -137,7 +138,7 @@ export function title(model: Model, channel: Channel) {
   }
 
   // if not defined, automatically determine axis title from field def
-  const fieldTitle = model.fieldTitle(channel);
+  const fieldTitle = fieldDefTitle(model.fieldDef(channel), axis.showUnit === 'title' || !axis.showUnit);
```
kanitw: showUnit is always a string, and the default value is `'title'`. 

I think you don't need `!axis.showUnit` here.

src/compile/axis.ts
kanitw: I mean, if you really can't find where .. ask me.  But it's good to practice venturing into other codebase such as Vega. :)

src/compile/axis.ts
ayush29feb: Right, I think I added it later because it wasn't working. I will take another look.

src/fielddef.ts
ayush29feb: I like that, will change it

kanitw: Yeah.   Also, don't use `var` unless you have compelling reason.   Use `let` only when you need to mutate the variable.   Always try to use `const` when possible.  :)

src/compile/axis.ts
kanitw: Hint: scales.js and legends.js  and obviously you are thinking about "format".

ayush29feb:
Does it make sense to leave out units when there is timeUnit set?

kanitw:
Good question. After thinking about it, i d say no.
In general people wouldnt specify unit with time.

In fact, year's unit is implicitly AD. But there are cases where people
might wanna use other unit such as BE (Buddhist Era). So it is probably
better not to add this condition.

Issue #948, opened 2016-01-15
Clean up styling for facet


kanitw:
We should provide good default before 1.0 release and may provide customizable property later

- [ ] For example, do we add offset between shared axis and the graphs.  This will cause problem in some case like shown in #523 

- [x] Unconnected grid like in #920

kanitw:
#920 is done.   The only question left is whether we always add offset to axis of facet?

Right now we only add offset to the shared axis like y-axis in the figure below.  

![vega_editor](https://cloud.githubusercontent.com/assets/111269/12997554/531fa18a-d0ef-11e5-89de-ee5690ee5ccf.png)

It could be nice to do so for the x-axis as well for symmetry.

kanitw:
Oops. Although I think it is nice to do so, there is a Vega bug with grid when we add `offset` to `axis`.  https://github.com/vega/vega/issues/519

Let's punt this until it is solved in Vega then.  (Maybe 1.0.1)

domoritz:
@arvind closed vega/vega#519 in https://github.com/vega/vega/commit/96c539f831da716641436baaaf7d1ada10f4ba09

PR #1098, opened 2016-02-12
Facet Styling

### LINKED ISSUE #803: Add additional "rule" marks when COLUMN/ROW is mapped, but X/Y is not
For example, to complete text table's table border.  
vega_editor

(Whether we add table border by default is orthogonal -- if the borders are added, it should be completed.

### LINKED ISSUE #920: Draw x,y grid inside facet
Current faceted charts have shared grid across multiple like this:

vega_editor

However, having separate grid for different singles like an example below is nicer.

——

## TODO
- [ ] add a property for `split` axis mode (Need to think what should be the name though)
  - In this mode
    - [ ] hide grid from the outside shared x/y axis
    - [ ] add inner x/y axis (inside each cell group) that shows only grid 
    - [ ] decide which properties get applied to the outer axis (e.g., ticks)

kanitw:
- Drawing grid for facet inside each cell and draw border around each cell.  Fix #920 
vega_editor

- Add extra rule mark to the manually drawn grid to close the table when faceting without inner x/y axis. Fix #803
vega_editor

- always add offset to facet's axis
- Don't use band of row/column scales to make sure the border of the cell works correctly

### DIFF
--- a/src/compile/axis.ts
+++ b/src/compile/axis.ts
@@ -7,6 +7,28 @@ import {formatMixins} from './util';
 // https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#11-ambient-declarations
 declare let exports;
 
+export function compileGridOnlyAxis(channel: Channel, model: Model) {
+  const isCol = channel === COLUMN,
+    isRow = channel === ROW,
+    type = isCol ? 'x' : isRow ? 'y': channel;
+
+  // TODO: replace any with Vega Axis Interface
+  return {
+    type: type,
+    scale: model.scaleName(channel),
+    grid: true,
+    tickSize: 0,
+    properties: {
+      labels: {
+        text: {value:''}
+      },
+      axis: {
+        stroke: {value: 'transparent'}
+      }
+    }
+  };
+}
+
 export function compileAxis(channel: Channel, model: Model) {
   const isCol = channel === COLUMN,
     isRow = channel === ROW,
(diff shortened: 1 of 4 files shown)


kanitw:
- Drawing grid for facet inside each cell and draw border around each cell.  Fix #920 
vega_editor

- Add extra rule mark to the manually drawn grid to close the table when faceting without inner x/y axis. Fix #803
vega_editor

domoritz:
Sweeeet

domoritz:
The thick line issue is gone with https://github.com/vega/vega/pull/521. No need to add offset to background.

src/compile/axis.ts
```diff
@@ -7,6 +7,47 @@ import {formatMixins} from './util';
 // https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#11-ambient-declarations
 declare let exports;
 
+/**
+ * Make an inner axis for showing grid for shared axis.  
+ */
+export function compilerInnerAxis(channel: Channel, model: Model) {
```
domoritz: compile or compiler?

domoritz:
There is something funny things going on with the offsets. Look at the examples and change row to column.

screen shot 2016-02-11 at 22 33 25
screen shot 2016-02-11 at 22 33 51
screen shot 2016-02-11 at 22 34 17

kanitw:
> Also it seems odd to me that there is space between the axis and the grid only for the y axis.

See my note in #948 about vega/vega#519.  You're welcome to fix that.  Shouldn't be too hard.

kanitw:
Re: stacked bar.  Great that you're finding this bug.   I need to think about how to solve it.  One thing that's terrible about d3 ordinal is that padding is a ratio, not integer.   ><"

kanitw:
Ok. The other issue fixed in https://github.com/vega/vega-lite/commit/4a9185885d6617d6ad6258935ca7fc5dc681e465

domoritz:
https://github.com/vega/vega/issues/519 has been fixed.  So we should be able to add offsets now.

PR #1099, opened 2016-02-12
Don’t output empty shape

domoritz:


kanitw:
Great!  This has been bugging me too.

PR #1100, opened 2016-02-12
Use format instead of role. See http://json-schema.org/latest/json-schema-validation.html#anchor104

domoritz:


test/schema.test.ts
```diff
@@ -2,12 +2,20 @@ const assert = require('assert'),
...
-  specSchema = require('../src/schema/schema').schema;
+const schema = require('../lib/schema.json');
+
+import * as schemautil from '../src/schema/schemautil';
+import {schema as specSchema} from '../src/schema/schema';
 
 describe('Schema', function() {
   it('should be valid', function() {
+    zSchema.registerFormat('color', function (str) {
+      return false;
```
kanitw: can you add comment why this `return false` ?

domoritz: Well, the formatter is not actually ever used because we only validate that the schema is valid.

kanitw: Ok.  Please add the comment and feel free to merge. :)

PR #1102, opened 2016-02-13
[WIP] Add body css properties

ZeningQu:
@domoritz These are the changes to `gallery.css` that you made today


domoritz:
Let's wait until your changes are in. Otherwise this will break the current gallery.

PR #1110, opened 2016-02-14
Add rank transform after aggregation, simplify legend logic.

### LINKED ISSUE #1104: Stacked Area has turned to the dark side
vega_editor

```
{
  "mark": "area",
  "encoding": {
    "x": {"timeUnit": "year", "field": "Year", "type": "temporal"},
    "y": {"aggregate": "sum", "field": "Weight_in_lbs", "type": "quantitative"},
    "color": {"field": "Cylinders", "type": "ordinal"}
  },
  "config": {"mark": {"interpolate": "monotone"}}
}
```

domoritz:
Fixes #1104.

### DIFF
--- a/src/compile/data.ts
+++ b/src/compile/data.ts
@@ -36,6 +36,9 @@ export function compileData(model: Model): VgData[] {
     def.push(summaryDef);
   }
 
+  // add rank to the last dataset
+  rankTransform(def[def.length-1], model);
+
   // append non-positive filter at the end for the data table
   filterNonPositiveForLog(def[def.length - 1], model);
 
@@ -124,8 +127,7 @@ export namespace source {
       formulaTransform(model),
       filterTransform(model),
       binTransform(model),
-      timeTransform(model),
-      rankTransform(model)
+      timeTransform(model)
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/compile/data.ts
```diff
@@ -462,6 +447,23 @@ export namespace dates {
   }
 }
 
+// We need to add a rank transform so that we can use the rank value as
+// input for color ramp's linear scale.
+export function rankTransform(dataTable, model: Model) {
+  if (model.has(COLOR) && model.fieldDef(COLOR).type === ORDINAL) {
+    dataTable.transform = dataTable.transform.concat([{
```
kanitw: I actually prefer not causing side effect.   
(And yes, I also don't like how I wrote `filterNonPositiveForLog` earlier)

src/compile/legend.ts
```diff
@@ -103,58 +103,46 @@ namespace properties {
...
       case AREA:
-        // set stroke to transparent by default unless there is a config for stroke
-        symbols.stroke = {value: 'transparent'};
-        applyMarkConfig(symbols, model, FILL_STROKE_CONFIG);
-
-        // TODO use shape here after implementing #508
         break;
     }
 
+    // fill or stroke
```
kanitw: You shouldn't have done these two irrelevant things in one commit.

I only have the "filled" logic for CIRCLE, SQUARE, POINT because only these three mark types we can adjust shape / fill style for legend.

(We don't adjust shape for BAR, TICK, TEXT, LINE, AREA, so applying fill here do not always make sense.)   

It is possible that there are bugs for these 5 marks.  But treating them all the same is likely even more incorrect.

domoritz: Well, it's not irrelevant. The problem was that points don't get colored when we have an ordinal color field. 

I could add complicated logic to only add the fill if we have an ordinal scale (which is likely to have bugs now or become buggy in the future) or just set the stroke color for all marktypes instead of also the fill.

PR #1113, opened 2016-02-14
Refactor legends to only produce necessary properties and match the l…

domoritz:
…egend property (fill, stroke) with what we actually care about.


src/compile/legend.ts
```diff
@@ -92,6 +90,11 @@ export function formatMixins(legend: LegendProperties, model: Model, channel: Ch
   return utilFormatMixins(model, channel, typeof legend !== 'boolean' ? legend.format : undefined);
 }
 
+// we have to use special scales for ordinal or binned fields for the color channel
+export function specialScale(fieldDef: FieldDef) {
```
kanitw: `useColorLegendScale` maybe a better name?

src/compile/legend.ts
```diff
@@ -105,40 +108,36 @@ namespace properties {
...
+        // use default circle
         break;
     }
 
-    // fill or stroke
-    if (model.config().mark.filled) { // filled
-      // set stroke to transparent by default unless there is a config for stroke
-      symbols.stroke = { value: 'transparent' };
-      applyMarkConfig(symbols, model, FILL_STROKE_CONFIG);
+    applyMarkConfig(symbols, model, FILL_STROKE_CONFIG);
```
kanitw: If a fill / stroke is specified, it might override what we put in fill/stroke up there (while in marks, the config has lower precedence).

src/compile/legend.ts
domoritz: Done

src/compile/legend.ts
domoritz: Fixed. This works correctly now

```json
{
  "mark": "point",
  "encoding": {
    "x": {"aggregate": "sum", "field": "yield", "type": "quantitative"},
    "y": {"field": "variety", "type": "nominal"},
    "color": {"field": "site", "type": "nominal"}
  },
  "config": {
    "mark": {
      "stroke": "green"
    }
  }
}
```

However, there is https://github.com/vega/vega-lite/issues/1114

PR #1117, opened 2016-02-15
Add examples to mark.md

kanitw:
- Add text right after points due to its similarity
  -  I'm not feeling that great about adding text table to the tutorial because I think it will change a lot after we support labeling.  

- Add other missing examples 

- The only part that is missing is the area.  We need a good dataset for stacked area / normalized stacked area / streamgraph.   @domoritz  if you have an idea, feel free to take it over.


domoritz:
Great. Our example tests found a bug ;-)

domoritz:
I sent a pr your way https://github.com/vega/vega-lite/pull/1118

PR #1118, opened 2016-02-15
Fix

domoritz:


examples/specs/bar_grouped_population.json
```diff
@@ -25,5 +25,5 @@
       "scale": {"range": ["#EA98D2", "#659CCA"]}
     }
   },
-  "config": {"cell": {"stroke": null}}
+  "config": {"cell": {"strokeWidth": 0}}
```
kanitw: The annoying part of this is default stroke value is still produced atlhough it is hidden.  

Is there away to override default config to be undefined?

domoritz: I don't think so, unfortunately.

PR #1120, opened 2016-02-15
Extract `mainScale()`, `colorLegendScale()`, `binColorLegendLabelScale()`

kanitw:


src/compile/scale.ts
```diff
@@ -63,40 +63,9 @@ export function compileScales(channels: Channel[], model: Model) {
       // Add additional scales needed to support ordinal legends (list of values)
       // for color ramp.
       if (channel === COLOR && model.legend(COLOR) && (fieldDef.type === ORDINAL || fieldDef.bin || fieldDef.timeUnit)) {
```
kanitw: @domoritz  Is there any reason why colorLegend scales have to be pushed before normal `scaleDef`?   It looks like it's fragmenting the logic for main scale and colorLegendScale

domoritz: Nope, no reason. Can be moved outside.

PR #1121, opened 2016-02-15
Revise scale properties rules & Refactor & Scale Docs

kanitw:
- Refactor scale property signature & check if the property is applicable to a scaleType
- Remove unused `outerPadding`  (We never use band scale, so `outerPadding` is never used.)
- `scale.nice` using smallest timeUnit for time scale
- 	Do not allow customized range for x, y, row & column
- Update Scale's docs

### DIFF
--- a/src/compile/time.ts
+++ b/src/compile/time.ts
@@ -49,6 +49,38 @@ export function format(timeUnit, abbreviated = false): string {
   return out.length > 0 ? out.join(' ') : undefined;
 }
 
+/** returns the smallest nice unit for scale.nice */
+export function smallestUnit(timeUnit): string {
+  if (!timeUnit) {
+    return undefined;
+  }
+
+  if (timeUnit.indexOf('second') > -1) {
+    return 'second';
+  }
+
+  if (timeUnit.indexOf('minute') > -1) {
+    return 'minute';
+  }
+
+  if (timeUnit.indexOf('hour') > -1) {
+    return 'hour';
+  }
+
+  if (timeUnit.indexOf('day') > -1 || timeUnit.indexOf('date') > -1) {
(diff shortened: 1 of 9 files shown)


src/compile/time.ts
```diff
@@ -49,6 +49,38 @@ export function format(timeUnit, abbreviated = false): string {
...
+
+  if (timeUnit.indexOf('minute') > -1) {
+    return 'minute';
+  }
+
+  if (timeUnit.indexOf('hour') > -1) {
+    return 'hour';
+  }
+
+  if (timeUnit.indexOf('day') > -1 || timeUnit.indexOf('date') > -1) {
```
domoritz: We should not treat date as day. They are different.

kanitw: yeah, but there is no nice value for date in Vega?  What else would you put here?

domoritz: I don't want to have confusion between https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate

domoritz: My main concern is that there are many days in the month so day for nice may be too detailed. What exactly happens when we set nice to day anyway? Maybe I'm seeing a problem where no problem is.

kanitw: When the nice is at "day" level, it's rounding things to `day` level (cutting all hours and minutes).  So I don't know why is that too detailed.

domoritz: Okay, I misunderstood. Then this is fine. You said on slack that you have concerns now?

PR #1122, opened 2016-02-15
Move layout data to layout.ts 

kanitw:
Its flow would be different from normal data in composition.

Normal data will be populated with pre-order traversal while layout data will have post-order traversal.


src/schema/vega.schema.ts
```diff
@@ -0,0 +1,8 @@
+export interface VgData {
```
domoritz: We should move this to the vega typings at some point.

Issue #1103, opened 2016-02-13
Add name prefix for every named component


kanitw:
- [x] ~~add `model.prefix()` that returns `spec.name + '-'`  for simplicity~~
- [ ] make `model.name()` takes a string input
       
- [ ] use `model.name()` in the declarations of every named instance -- Basically search for `name:` inside `compiler/` and most (if not all) of them should call `model.name()`.  On the top of my head, at least it should be:
  - [ ] data  
     - [ ] data declaration
     - [ ] data references (search for `data:`)
  - [ ] fields in layout  (e.g., `cellWidth`) 
  - [ ] scale
  - [ ] mark 

- [ ] For all the renamed instances, make sure to edit when we refer to them.

kanitw:
@bobocandys  Looks like you can consolidate  `model.scaleName()` with `model.name()` too.  (They basically do the same thing, just make `localName` takes `string | Channel`. )

kanitw:
Pro-tip:  modify one type of instance at a time and make sure all of its references are correct first before modifying next type of object :)

kanitw:
@bobocandys   I hope you aren't too far on this branch.  
Actually I was wrong about consolidating `model.scaleName()` with `model.name()`. 

In fact, the correct thing to do is actually the opposite.  Basically, we should have different name methods for different types of named instance.  
- `model.layoutName()`
- `model.scaleName()`
- `model.dataName()`
- `model.markName()`

In Vega-Lite 1.0, all of these methods will behave the same.    But with composition in the very near future, some name reference will reference to a slightly different name.  (You don't have to worry about this, but make sure to split these methods.  Otherwise, we will have to split them later anyway.)

kanitw:
handled as part of #1192

PR #1128, opened 2016-02-16
Add title to plot with fixed y coordinate of the root group.

bobocandys:


src/compile/compile.ts
```diff
@@ -33,10 +33,10 @@ export function compile(spec) {
     config.viewport ? { viewport: config.viewport } : {},
     config.background ? { background: config.background } : {},
     keys(config.scene).length > 0 ? scene(config) : {},
-    {
-      data: compileData(model).concat([compileLayoutData(model)]),
-      marks: [compileRootGroup(model)]
-    });
+    { data: compileData(model).concat([compileLayoutData(model)]) },
```
domoritz: Can you reformat this so it is easier to read?

src/compile/compile.ts
```diff
@@ -57,6 +57,26 @@ function scene(config) {
   }, {});
 }
 
+export function compileTitle(model: Model) {
+  const spec = model.spec();
+  return {
```
domoritz: no need to put `'` around keys

src/compile/compile.ts
```diff
@@ -57,6 +57,26 @@ function scene(config) {
...
+      'name': 'title',
```
domoritz: Please use the method in model to generate names

src/compile/compile.ts
```diff
@@ -57,6 +57,26 @@ function scene(config) {
...
+      'type': 'text',
+      'from': {'data': 'layout'},
+      'properties': {
+        'update': {
+          'x': {'field': 'width', 'mult': 0.5},
+          'y': {'value': 0},
+          'text': {'value': spec.title},
+          'fill': {'value': 'black'},
```
domoritz: All of these things should probably become configs. Can do later.

test/compile/compile.test.ts
```diff
@@ -33,4 +33,38 @@ describe('Compile', function() {
     });
 
   });
+
+  describe('compileTitle()', function() {
+    describe('ordinal', function() {
+      const model = parseModel({
+        "description": "A simple bar chart with embedded data.",
```
domoritz: Make the spec much smaller. Only add what you need.

test/compile/compile.test.ts
```diff
@@ -33,4 +33,38 @@ describe('Compile', function() {
...
+      const titleGroup = compile.compileTitle(model);
+
+      it('should refer to layout data', function() {
+        assert.deepEqual(titleGroup.from, {"data": "layout"});
+      });
+
+      it('should have type of text', function() {
+        assert.deepEqual(titleGroup.type, "text");
+      });
+
```
domoritz: Please tests that it actually sets the title ;-)

src/compile/compile.ts
bobocandys: Can you explain more on this? Use the method in model instead of giving `name: 'title` ?

src/compile/compile.ts
kanitw: One readable option is:

```
{ 
   data: compileData(model).concat([compileLayoutData(model)]),
   marks: [].concat(
     (spec.title ? [compileTitle(model)] : []),
     [compileRootGroup(model)]
   )
}
```

src/compile/compile.ts
kanitw:
```js
'name':   (model.name() ? model.name() + '-' : '') + 'title'
```

domoritz: Hmm, I'm thinking whether we should change the `model.name` function to take a string parameter instead. I can do that outside this pr.

kanitw: Yeah. Let's do that as another PR (as part of fixing #1103).

src/compile/compile.ts
```diff
@@ -35,8 +35,12 @@ export function compile(spec) {
     keys(config.scene).length > 0 ? scene(config) : {},
     {
       data: compileData(model).concat([compileLayoutData(model)]),
-      marks: [compileRootGroup(model)]
-    });
+      marks: [].concat(
+       (spec.title ? [compileTitle(model)] : []),
```
kanitw: For all `spec.title` reference, could you please add `model.title()` and call it?   (We plan to get rid of `model.spec()` so we don't want to call it in too many more places)

test/compile/compile.test.ts
bobocandys: I checked if it actually sets the correct title name. Is this okay?

src/compile/compile.ts
bobocandys: Done. It makes sense to do this.

src/compile/compile.ts
```diff
@@ -68,10 +92,13 @@ export function compileRootGroup(model: Model) {
     {
       from: {data: LAYOUT},
       properties: {
-        update: {
-          width: {field: 'width'},
-          height: {field: 'height'}
-        }
+        update: extend(
+          model.title() !== undefined ? {y: {'value': 55} }: {},
```
domoritz: 55 actually looked way to large for the example @kanitw showed today. Can you try different examples and see whether we need a more complex formula here?

domoritz:
Closing since we will use composition.

PR #1129, opened 2016-02-16
[WIP] Docs redesign and updates

willium:
Not a lot to show yet, most of my work for the redesign has been in sketches or partial implementations. Not a big fan of any of the versions I've built yet, and they all break at some point. Opening this up for discussion -- will update this description with more information.


domoritz:
Can you delete the .jekyll-metadata

willium:
what still needs to be done @domoritz?

Issue #1011, opened 2016-01-27
Reduce redundancy in generated schema


mprudhom:
Some structures in the generated `vega-lite-schema.json` are repeated multiple times. For example, the `timeUnit` enumeration is repeated verbatim 13 times throughout the schema. Detecting this redundancy and emitting repeated structures to a shared `defs` section of the schema would make it much more compact.

TODO:

- [x] add enum support to typescript-json-schema
- [ ] change build process to generate schema using new typescript-json-schema
- [ ] remove old schema generation process

mprudhom:
I would be interested in working on this, but I was first wondering if it should be part of #927.

kanitw:
Yes, this should be a part of #927.    In addition, our `vl.schema.instantiate` method, which instantiate default values for the spec,  also uses `default` property in the schema.   Current vl.schema.instantiate does not support traversing `defs` section of the schema.  So that's another part of work to be done for this to work.

domoritz:
@mprudhom It would be awesome if you could look into this! In addition to using `defs`, we think that the json schema should be generated from typescript rather than manually by hand. In fact, I worked a bit on https://github.com/YousefED/typescript-json-schema to add support for union types and it's almost ready to be used for our purposes. What's missing right now is support for enums. 

I suggest that we first look into the ts -> json schema generation (which can use `defs`) and then modify the instantiate method to use references. 

If you want to talk about this in more detail, we could discuss on skype.

mprudhom:
@domoritz Are you sure that `typescript-json-schema` supports storing shared schemas in a separate definitions section of the schema? Glancing over the code I don't see anywhere that will generate any `$ref` schema attributes.

domoritz:
Sorry, I didn't realize this hasn't been merged yet. There is a branch referenced from https://github.com/YousefED/typescript-json-schema/issues/3. It only needs a few more changes to be ready.

domoritz:
Relevant #927

domoritz:
@mprudhom would you like to work on this? It would be incredibly helpful for us!

mprudhom:
@domoritz I'll take a stab at it. Will it be OK for vega-lite to rely on my own fork of typescript-json-schema? As you mention, it will at least need support for enums.

domoritz:
Sure! Although @ YousefED has been extremely responsive so I think we will be able to merge quickly. 

[...]

Issue #927, opened 2016-01-12
Automatically generate JSON Schema from TS Interfaces


kanitw:

mprudhom:
I can try working on this. To clarify: is the goal that the `axis.schema.ts` file will contain only the `interface Axis` section and not the extra somewhat-redundant `export var axis` section?

domoritz:
Exactly. We should be able to generate the json scheme from the typescript interfaces.

One thing to keep in mind is that we need to sometimes instantiate parts of the schema for instance when we want to replace `bin: true`. We were hoping that if we use classes with the defaults in addition to interfaces, we could simply instantiate an object of the class. The ts to json schema converter can handle classes.

domoritz:
We cold also instantiate from a #ref or when we merge the defaults into the incoming spec, we could be aware of the fact that we have to replace booleans with objects sometimes. 

There are a lot of options here but we will have to try them out to see which one works (best). 

I'm happy to discuss options on Skype.

kanitw:
>  replace booleans with objects sometimes.

In that case, the signature would be   `boolean | InterfaceName` (e.g.,  `boolean | Axis`) anyway so if the TS-> schema generator handles union type (`A | B`) and interface extension (`A extends B`) correctly.  We should be fine.  

(Default value instantiation should be all moved into `config` soon anyway.)

domoritz:
Done in #1131

PR #1130, opened 2016-02-16
[WIP] Generate schema from TypeScript classes (#1011 & #927)

mprudhom:
Schema is now generated directly from using the TypeScript classes `typescript-json-schema`. The classes have been refined just enough to pass the schema validation tests.

The old schema is still being generated as well; this redundancy should be eliminated. Both the old and new schemas are being validated in `examples.test.ts` (see `isVlValid` and `isVlValid2`).


domoritz:
Superseded by #1131

domoritz:
I merged everything and removed everything about the old schema in #1131.

PR #1131, opened 2016-02-17
Generate schema from TypeScript classes (#1011 & #927)

domoritz:
See  #1011


src/schema/scale.schema.ts
```diff
@@ -60,125 +30,56 @@ export const defaultFacetScaleConfig: FacetScaleConfig = {
...
+  /**
+   * @minimum 0
+   */
   bandWidth?: number;
+  /**
+   * Applies spacing among ordinal elements in the scale range. The actual effect depends on how the scale is configured. If the __points__ parameter is `true`, the padding value is interpreted as a multiple of the spacing between points. A reasonable value is 1.0, such that the first and last point will be offset from the minimum and maximum value by half the distance between points. Otherwise, padding is typically in the range [0, 1] and corresponds to the fraction of space in the range interval to allocate to padding. A value of 0.5 means that the range band width will be equal to the padding width. For more, see the [D3 ordinal scale documentation](https://github.com/mbostock/d3/wiki/Ordinal-Scales).
+   */
   padding?: number;
 
   // typical
```
kanitw: Maybe we need to make the HEADER for semantic grouping clearer?

kanitw:
LGTM except minor comment about comment header.  We can take care of that later too.

src/schema/scale.schema.ts
domoritz: Not sure what you mean. But seems to be something we can do over time.

PR #1134, opened 2016-02-17
Dom/image scripts

domoritz:


kanitw:
Faster is great, can you provide a way to know which one throws error as well?

domoritz:
Already done. The script aborts when there is an error and it shows where it erred.

kanitw:
Great!

kanitw:
@domoritz  Can you add to contribute.md that people have to install parallel as well?

PR #1135, opened 2016-02-17
Migrate scale default value to config + remove schema.instantiate

kanitw:
- Add scale config, replace constants with config, remove default values from config
- remove schema.instantiate + perform necessary fixes to eliminate null pointer errors


domoritz:
Sweet. Ping me when it's passing.

PR #1136, opened 2016-02-17
Home page examples

kanitw:
vega-lite

This is not perfect yet.  Further edits are welcomed!


kanitw:
Note that the grouped bar chart should actually look like this but the images script doesn't finish runnning for me ... 

vega_editor

PR #1147, opened 2016-02-18
Formalize string enums

mprudhom:
This PR creates formal enum types for various string constants like `TimeUnit`, `StackOffset`, `ShapeKind`, `FontStyle`, etc.

It also changes the enums 'Type' to 'DataType' and 'Mark' to 'MarkType' to disambiguate them from other enums and built-in typescript keywords.


src/mark.ts
```diff
@@ -1,4 +1,4 @@
-export enum Mark {
+export enum MarkType {
```
kanitw: Please do not change Mark to MarkType as the property name is Mark.

mprudhom: OK, reverted.

src/enums.ts
```diff
@@ -0,0 +1,80 @@
...
+    LEFT = 'left' as any,
+    BOTTOM = 'bottom' as any
+}
+
+export enum FontWeight {
+    NORMAL = 'normal' as any,
+    BOLD = 'bold' as any
+}
+
+export enum ShapeKind {
```
kanitw: Simply `Shape` is better.

kanitw:
Thanks for the PR, besides minor comments.  I'll need to discuss with @domoritz whether we want to eventually use string literals or Enums.   That said, I am thinking that even if we want to use string literals in the future merging this for now would be useful.   So very likely we’ll merge it soon, but let's see what @domoritz  thinks.  

In any case, in the future, if you plan to modify a lot of files like this (which is greatly helpful and we really appreciate your help), it might be worth discussing with us first just to make sure that you won't put effort into a PR that might not get merged.

kanitw:
Okay.  Just chatted with @domoritz.   We'll merge this and fix the ShapeKind for you :) 

As said, it's a very good thing to do anyway and we can consider migrating to string literals later. 

Thanks a lot for this! :)

mprudhom:
Thanks for merging, and in the future I'll discuss ahead of time if I have any more sweeping PRs to propose.

Regarding enums vs. string literals, I assume you are referring to typescript's constrained string literals like:

```
fontWeight?: "normal" | "bold";
```

The reason I think enums are superior is that even though the typescript compiler validates assignment to constrained string fields (e.g., `fontWeight = "nermal"` would fail to compile), it does not validate comparison to constrained string fields (e.g., `if (fontWeight == "nermal")` passes even through the state is impossible).

PR #1149, opened 2016-02-18
Default to 'json' type if data format is not set

mprudhom:
Whenever setting vega data, ensure that the data.type is set to something (otherwise is introduces ambiguity into the resulting vega spec, since data's oneOf will match multiple schemas when 'type' is unset).


kanitw:
Thanks for your PR.  However, I think the correct fix is to fix the schema because we should not generate {type: 'json'} when we intended to use a Vega-Lite spec for binding data at runtime (using Vega runtime).

PR #1151, opened 2016-02-18
Scale's bandWidth => bandSize

kanitw:
Note that textBandWidth remains "width" because it is only applied to x-scale :)

### DIFF
--- a/site/docs/config.md
+++ b/site/docs/config.md
@@ -198,7 +198,7 @@ vg.embed('#horizontal_line', {
 
 | Property      | Type          | Description    |
 | :------------ |:-------------:| :------------- |
-| barWidth      | Number        | The width of the bars. <span class="note-line">__Default value:__  `bandSize-1` for the bar's dimension is an ordinal scale. (This provides 1 pixel offset between bars.) and `2` for if bar's dimension has linear scale.  </span>  |
+| barSize      | Number        | The size of the bars (width for vertical bar charts and height for horizontal bar chart). <span class="note-line">__Default value:__  `bandSize-1` for the bar's dimension is an ordinal scale. (This provides 1 pixel offset between bars.) and `2` for if bar's dimension has linear scale.  </span>  |
 
 
 ### Point Config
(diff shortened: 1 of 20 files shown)


site/docs/config.md
```diff
@@ -198,7 +198,7 @@ vg.embed('#horizontal_line', {
 
 | Property      | Type          | Description    |
 | :------------ |:-------------:| :------------- |
-| barWidth      | Number        | The width of the bars. <span class="note-line">__Default value:__  `bandWidth-1` for the bar's dimension is an ordinal scale. (This provides 1 pixel offset between bars.) and `2` for if bar's dimension has linear scale.  </span>  |
+| barWidth      | Number        | The width of the bars. <span class="note-line">__Default value:__  `bandSize-1` for the bar's dimension is an ordinal scale. (This provides 1 pixel offset between bars.) and `2` for if bar's dimension has linear scale.  </span>  |
```
kanitw: @domoritz   Should this become barSize for consistency?

site/docs/config.md
```diff
@@ -221,7 +221,7 @@ vg.embed('#horizontal_line', {
 
 | Property            | Type                | Description  |
 | :------------------ |:-------------------:| :------------|
-| tickWidth           | Number        | The width of the ticks. <span class="note-line">__Default value:__ `2/3*bandWidth` (This will provide offset between band equals to the width of the tick.) </span>|
+| tickWidth           | Number        | The width of the ticks. <span class="note-line">__Default value:__ `2/3*bandSize` (This will provide offset between band equals to the width of the tick.) </span>|
```
kanitw: @domoritz   Should this become tickSize for consistency?

site/docs/config.md
domoritz: Hmm. No? Because I consider the size of a bar to be it's length/height.

site/docs/config.md
domoritz: Yes.

site/docs/config.md
kanitw: But bar's size channel is always this width thing.   (Length is fixed)

But I was thinking about this the other day too.   (Should we allow using size for gantt bar?)  The current meaning of size for bar is mostly inspired from Tableau, but has no practical use.

domoritz: https://www.mathworks.com/help/matlab/ref/bar.html#bthxceu, maybe thickness. I think we can go either either. All choices have advantages and disadvantages. But I guess consistency trumps everything.

kanitw: `thickness` is worth considering, but it is not consistent with `tickThickness`  (thickness of the bar is orthogonal to thickness of thick)

barWidth can be confusing because it's the height of the bar

domoritz: > thick

lol, tick

domoritz: > barWidth can be confusing because it's the height of the bar

Depends on whether the bar is horizontal or vertical.

Issue #1156, opened 2016-02-19
Some example vega-lite specs do not produce valid vega specs


mprudhom:
`examples/examples.test.ts` is not actually validating that the compiled specs match `vega-schema.json`, since https://github.com/vega/vega-lite/blob/master/examples/examples.test.ts#L28:

```
  const isVgValid = validator.validate(vegaSpec, vgSchema);
```

should instead be:

```
  const isVgValid = validator.validate(vegaSpec.spec, vgSchema);
```

Once the above change is made, 23 of the vega specs fail schema validation:

1. area.json
2. area_vertical.json
3. bar_yearmonth.json
4. line.json
5. line_color.json
6. line_detail.json
7. line_monotone.json
8. line_month.json
9. line_step.json
10. point_dot_timeunit_color.json
11. point_ordinal_color.json
12. scatter_color_ordinal.json
13. scatter_color_ordinal_custom.json
14. stacked_area.json
15. stacked_area_normalize.json
16. stacked_area_ordinal.json
17. stacked_area_stream.json
18. trellis_bar.json
19. trellis_barley.json
20. trellis_row_column.json
21. trellis_scatter.json
22. trellis_scatter_binned_row.json
23. trellis_stacked_bar.json

domoritz:
Oh wow. Good catch. We moved the vega output spec into an object and forgot to update.

arvind:
Note, there have been some errors in the Vega schema that have been fixed in the master branch. For example, `area.json` validates for me.

PR #1157, opened 2016-02-19
Fix Schema validator in examples.test.ts

kanitw:


examples/examples.test.ts
```diff
@@ -33,7 +40,7 @@ function validateAgainstSchemas(vlspec) {
   assert(isVgValid);
 }
 
-function getExampleList(examples) {
+export function getExampleList(examples) {
```
domoritz: what's this function? Can you delete it?

kanitw: Good point.  I was a function that you should have removed when we changed it to read directly from directory.


Done.

domoritz: Hehe, yeah.

examples/examples.test.ts
```diff
@@ -1,13 +1,20 @@
...
-const ZSchema = require('z-schema');
+const zSchema = require('z-schema');
 const inspect = require('util').inspect;
 const dl = require('datalib');
 const fs = require('fs');
 const path = require('path');
 
-const validator = new ZSchema({
+zSchema.registerFormat('color', function (str) {
+  return false;
```
domoritz: Hmm, don't you have to actually validate the string? I'm surprised that the tests pass if the validator always returns false.

kanitw: I just copied what you did earlier.

domoritz: The difference is that from where you copied it, we only care that the validator exists. Here the format validator should actually check whether `red` is a color. 

I guess we never specify a color in an example?

test/util.ts
```diff
@@ -8,3 +9,14 @@ import {Spec} from '../src/schema/schema';
 export function parseModel(spec) {
   return new Model(spec as Spec);
 }
+
+export const zSchema = require('z-schema');
+
+zSchema.registerFormat('color', function (str) {
+  // valid colors are in list or hex color
+  return contains(['purple'], str) || /^#([0-9a-f]{3}){1,2}$/i.test(str);
```
kanitw: Why only purple?

kanitw: Better copy a list of valid svg & canvas color name from somewhere?

domoritz: Cause that's the only color we actually use :-P Feel free to add it. I don't think we need to be exhaustive.

domoritz:
Fixes https://github.com/vega/vega-lite/issues/1156

PR #1159, opened 2016-02-20
Basic facet documentation

domoritz:


kanitw:
@domoritz  just edited the page.   

Note that occlusion is one potential problem that a small multiple can solve, but not the main one, so I didn't include it in my edit.   Facilitate comparison–across category, time, etc.–is the key. 

If you like it, please feel free to merge :)

PR #1164, opened 2016-02-22
Convert all vars to const or let where applicable

### LINKED ISSUE #907: Replace var with let and const
(forked from #723)

willium:
Fixes #907. Just a general plumbing thing, I thought would be nice for v1.


kanitw:
Nice!

PR #1165, opened 2016-02-22
Quick-fixes to docs (sidebar)

willium:
- Increased click area
- added `break` instead of empty sidebar item

### DIFF
--- a/_layouts/page_submenu.html
+++ b/_layouts/page_submenu.html
@@ -25,6 +25,9 @@ layout: plain
   <aside class="page-sidebar">
     <ul class="sidebar-nav">
       {% for link in page.sidebar %}
+      {% if link.break %}
+      <br>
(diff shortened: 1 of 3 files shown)


_layouts/page_submenu.html
```diff
@@ -25,6 +25,9 @@
   <aside class="page-sidebar">
     <ul class="sidebar-nav">
       {% for link in page.sidebar %}
+      {% if link.break %}
```
domoritz: this is good to have but can you use a separator that we can resize in css?

willium: you can resize <br> tags, I'm pretty sure

domoritz: I just haven't used `br` in a decade because pure css seems cleaner but maybe it's not that important right now.

PR #1171, opened 2016-02-23
fix docs presentation and bugs in FF/Safari

willium:
Should fix FF bug, Safari bug, and italicized optional for readability

### DIFF
--- a/site/static/main.css
+++ b/site/static/main.css
@@ -267,7 +267,7 @@ nav a:hover {
   margin-right: 60px;
 }
 .lead-buttons {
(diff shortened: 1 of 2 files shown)


site/static/main.css
```diff
@@ -323,6 +322,12 @@ nav a:hover {
   background-repeat: no-repeat;
 }
 
+.showcase .image:before{
+	content: "";
```
domoritz: spaces, not tabs

what does this code do?

willium: fixes FF bug

PR #1172, opened 2016-02-23
Update for consistency with the Medium article

kanitw:
- update binned scatter plot and histogram to use the medium example.
- add binned scatter plot to the examples gallery.
- update bin.md accordingly

@domoritz  @willium

### DIFF
--- a/site/docs/bin.md
+++ b/site/docs/bin.md
@@ -42,23 +42,36 @@ If `bin` is `true`, default binning parameters are used.  To customize binning p
 
 Given a field with quantitative continuous data values
 
-<span class="vl-example" data-name="point_1d"></span>
-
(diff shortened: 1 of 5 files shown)


willium:
LGTM. We just need to make sure to deploy vega-editor soon after vega-lite

site/docs/bin.md
```diff
@@ -69,15 +82,19 @@ Setting the `maxbins` parameter changes the number of output bins.
 
 <div class="vl-example">
```
domoritz: why not use `examples/specs/histogram.json`

kanitw: This one has more number of bin.   

(I use `histogram`  right above here if you see the full file.)

PR #1176, opened 2016-02-24
Name root group mark as "rootCell"

arvind:
`root` is a reserved name corresponding to the actual root item of the Vega scenegraph (i.e., the element you can style with the `scene` property). This is relevant when we need to reference the `rootGroup` Vega-Lite adds during event handling via `eventGroup`.

### DIFF
--- a/src/compile/compile.ts
+++ b/src/compile/compile.ts
@@ -48,7 +48,7 @@ export function compileRootGroup(model: Model) {
   const spec = model.spec();
 
   let rootGroup:any = extend({
-      name: spec.name ? spec.name + '-root' : 'root',
+      name: spec.name ? spec.name + '-rootCell' : 'rootCell',
       type: 'group',
     },
     spec.description ? {description: spec.description} : {},


src/compile/compile.ts
```diff
@@ -48,7 +48,7 @@ export function compileRootGroup(model: Model) {
   const spec = model.spec();
 
   let rootGroup:any = extend({
-      name: spec.name ? spec.name + '-root' : 'root',
+      name: spec.name ? spec.name + '-rootCell' : 'rootCell',
```
kanitw: This is good to know haha.  Are there other reserved names that I should be aware of?  (Maybe we should put this in Vega's docs.) 

Btw, how about `layout-root`?  Because this isn't really a cell as in unit's cell.

arvind: Oh I see. I assumed that what is currently Vega-Lite's "root" group mark would become a unit cell. But I could also see introducing an additional group cell for each unit. Feel free to change as you see fit! No other reserved names that I can think of.

arvind: I had forgotten about this myself. Perhaps the root mark in Vega should be named `scene` instead to correspond to the `scene` property we introduced recently. Hmm.

kanitw: I think scene in Vega sounds right to me :)

kanitw:
We will rename `root` to `scene` in Vega :)

PR #1180, opened 2016-02-25
[WIP] Refactor Schema and Model to prepare for multiple types of them

kanitw:
- Schema 
  - Add schema for `BaseSpec`, `SingleSpec`, `FacetSpec`
  - make `(Extended)Encoding` extends `UnitEncoding`

- Rename Model to `UnitModel`
   - most of the marks should still call `UnitModel`  -- but need to refactor other files in a follow-up PR

- Split BaseModel from `UnitModel`  
 
cc: @arvind  – Also, FYI in this and the next few PRs,  I wanna make decompose facet and make it work within master first and then we can fork to extend it to support layering and concat.   (This way, other development by our junior colleagues for single spec wouldn't cause too much merge conflicts.)


kanitw:
superseded by #1192

PR #1182, opened 2016-02-25
Refactor scale to use scaleName

bobocandys:


kanitw:
@bobocandys  You actually did not update the scale reference like suggested.  

Anyway, since I would like to avoid conflict, I have just fixed it for you.

PR #1184, opened 2016-02-26
Make count op an array to conform to vega schema

mprudhom:
This fixes all but one of the failures in #1156


kanitw:
Thanks @mprudhom

PR #1185, opened 2016-02-26
Remove orient property from vega spec

mprudhom:
The "orient" mark property no longer exists in the Vega schema (https://github.com/vega/vega/issues/539), so do not generate it. This fixes the last failing example in https://github.com/vega/vega-lite/issues/1156.

### DIFF
--- a/src/compile/mark/area.ts
+++ b/src/compile/mark/area.ts
@@ -13,10 +13,6 @@ export namespace area {
     let p: any = {};
 
     const orient = model.config().mark.orient;
-    if (orient !== undefined) {
-      p.orient = { value: orient };
-    }
-
     const stack = model.stack();
     const xFieldDef = model.encoding().x;
     // x


kanitw:
Thanks for the help but this is not the right fix.  

For a vertical area chart, 
```
{
  "mark": "area",
  "encoding": {
    "y": {
      "scale": {"nice": "month"},
      "axis": {"axisWidth": 0, "format": "%Y", "labelAngle": 0}
    },
    "x": {
      "aggregate": "sum", "field": "count","type": "quantitative"
    }
  },
  "config": {"cell": {"width": 300, "height": 200}}
}
```

This correctly produces 

```
              "orient": {"value": "horizontal"}
```

in the Vega output.  If you remove this line from the compiled Vega spec, it no longer behaves correctly.

mprudhom:
Oops, right you are. Once "orient" gets back into the vega schema, #1156 should start passing.

kanitw:
Yep.  Thanks for figuring out the culprit in the schema validation though. :)

PR #1186, opened 2016-02-26
Validate generated vega spec against vega schema

mprudhom:
Fixes https://github.com/vega/vega-lite/issues/1156


kanitw:
Thanks.  I will merge this when Vega releases a new version then.  
(Otherwise, everything in Travis will always fail.)

kanitw:
We added another commit to vega version in #1209.  Will merge that soon! 
Thanks a lot :)

PR #1125, opened 2016-02-16
Image scripts

domoritz:
Use vega vg2png and vg2svg. Generate images in a different directory.    Fixes #34

PR #1188, opened 2016-02-26
Include command-line tools with package

sampsyo:
I was trying to use Vega-Lite locally after an `npm install -g vega-lite`; this should make that possible. See also #1125 and #34.

Perhaps these tools should be documented? Let me know if you'd like me to add something to the docs.

### DIFF
--- a/bin/vl2vg
+++ b/bin/vl2vg
@@ -9,7 +9,7 @@ var helpText =
 
 // import required libraries
 var fs = require('fs'),
-    vl = require('../vega-lite.js');
+    vl = require('vega-lite');
 
 // arguments
 var args = require('yargs')
--- a/package.json
+++ b/package.json
@@ -9,6 +9,11 @@
   ],
   "description": "Vega-lite provides a higher-level grammar for visual analysis, comparable to ggplot or Tableau, that generates complete Vega specifications.",
   "main": "src/vl.js",
+  "bin": {
+    "vl2png": "./bin/vl2png",
+    "vl2svg": "./bin/vl2svg",
+    "vl2vg": "./bin/vl2vg"
+  },
   "directories": {
     "test": "test"
   },


sampsyo:
I'm also not sure about the path to `vg2*` used in the `vl2*` scripts. I had to change it to use `$DIR`, i.e., the same bin directory as the Vega-Lite scripts, to make it work as an installed script. But I imagine you probably want to keep this working in an "uninstalled" location too.

bin/vl2vg
```diff
@@ -9,7 +9,7 @@ var helpText =
 
 // import required libraries
 var fs = require('fs'),
-    vl = require('../vega-lite.js');
+    vl = require('vega-lite');
```
domoritz: Unfortunately this breaks our image build script. We call vl2vg from package.json on `npm run build:images`. So this won't work unless we install vega lite globally. I'm happy to accept the change below but cannot merge unless we can call the bin scripts from local as well.

sampsyo:
Great!

Sorry for breaking your build scripts. I'm surprised a normal import doesn't work in a non-global `npm install`ed package; I guess I just don't understand npm as well as I thought I did. I'll see if there's a way to make these work both globally and locally and open another PR.

domoritz:
I guess I don't understand it enough either ;-) It might work if you install vega-lite globally but that is rarely the case. 

Thanks for the PR and let us know if there's anything you are missing from VL for your work. Are you building approximate visualizations?

sampsyo:
I think I might have a lead... I might be able to check back in tomorrow!

I was just hoping to draw some graphs for a paper. I hate all other plotting tools with intense passion, but I kinda like Vega! I tried using plain Vega for publishable plotting a couple of years ago ([notes](http://adriansampson.net/blog/vega.html)) but this stuff looks even cooler, so I'm just giving it another go.

domoritz:
> I hate all other plotting tools with intense passion, but I kinda like Vega! ... but this stuff looks even cooler, so I'm just giving it another go.

Fantastic. Note that you can edit the Vega that Vega-Lite generates to customize it if something is not quite the way you want it to be.

PR #1191, opened 2016-02-27
Make vl2svg and vl2png scripts actually work when installed

sampsyo:
A follow-up to #1188, where the goal is to get usable CLI tools after running `npm install -g vega-lite`.

This fixes two search-path issues that made `vl2svg` and `vl2png` work only when run from the Vega-Lite source directory. As suggested by @domoritz, I tried to avoid breaking any scripts that clearly *want* to run from inside the source directory, without a global install. Namely, `npm run build:images` still works.

Please let me know if I broke anything else!


domoritz:
Excellent. Thank you!

PR #1192, opened 2016-02-27
Decompose Compiler into compiling composition of unit spec

kanitw:
## Spec

- Add interface for
   - `UnitSpec` (Purely unit spec), 
   - `ExtendedUnitSpec` (Unit spec plus `row` and `column` channels), 
   - `FacetSpec` – all of them extend `BaseSpec` class.
- As we are merging this into `master`, currently the JSON schema produces `ExtendedUnitSpec`.  
- In our composition branch, we should change this to the `type Spec = ExtendedUnitSpec | FacetSpec`.   However, the TS schema converter doesn't work with `type` syntax.  

## Model

We now have two types of models: `UnitModel` and `FacetModel`, which both extends the base `Model` class.  (Obviously, we will add `LayerModel` and `ConcatModel` soon.)

## Compilation
- Compilation is now done in 4 steps:
  1. **Normalization** – Convert extended unit spec into a compositon of unit specs (normalized form) (Decompose all extended unit specs into composition of unit spec.)
  2. **Instantiate the model** with default properties
  3. **Parse** each part of the model to produce *components* that will be assembled later.  We traverse the tree to compile once for each type of components (e.g., data, layout, mark, scale).  Please see inside model.compile() for order for compilation.
  4. **Assemble** a Vega Spec from the parsed components in 3.


domoritz:
Decompose instead of Decompile?

src/compile/axis.ts
```diff
@@ -157,10 +159,6 @@ export function orient(model: Model, channel: Channel) {
   } else if (channel === COLUMN) {
     // FIXME test and decide
     return AxisOrient.TOP;
-  } else if (channel === ROW) {
```
kanitw: Note: Rule involves both Y and ROW -- moved to FacetModel's `_initConfig()`

src/compile/axis.ts
```diff
@@ -250,8 +251,6 @@ export namespace properties {
       // auto rotate for X and Row
       if (channel === X && (isDimension(fieldDef) || fieldDef.type === TEMPORAL)) {
         labelsSpec.angle = {value: 270};
-      } else if (channel === ROW && model.has(X)) {
```
kanitw: Note: Rule involves both X and COLUMN -- moved to FacetModel's `_initConfig()`

src/compile/data.ts
```diff
@@ -23,79 +23,228 @@ const DEFAULT_NULL_FILTERS = {
 };
 
 /**
- * Create Vega's data array from a given model.
+ * Composable component instance of a model's data.
+ */
+export interface DataComponent {
```
kanitw: Signature for each member of DataComponent might get revised more as we continue implementing composition.

src/compile/data.ts
```diff
@@ -23,79 +23,228 @@ const DEFAULT_NULL_FILTERS = {
...
+  dimensions: StringSet;
+
+  /** dictionary mapping field name to string set of aggregate ops */
+  measures: Dict<StringSet>;
+}
+
+// TODO: split this file into multiple files and remove this linter flag
+/* tslint:disable:no-use-before-declare */
+
+export function compileUnitData(model: UnitModel): DataComponent {
```
kanitw: Note: I'm not using for loop for `compileUnitData` and `compileFacetData` here like in axis/scale/legend so that we can ensure type check.

kanitw:
@domoritz   Now working for all 1.0 examples!  Ready for review + walkthrough.

kanitw:
- [ ] `compile` => `parse`

domoritz:
The enum at the root should work https://github.com/YousefED/typescript-json-schema/pull/32

package.json
```diff
@@ -32,7 +32,7 @@
     "start": "npm run watch & browser-sync start --server --files 'vega-lite.js' --index 'test-gallery.html'",
     "poststart": "rm examples/all-examples.json",
     "schema": "typescript-json-schema --required true src/spec.ts Spec > vega-lite-schema.json",
-    "presite": "tsc && npm run build && bower install && npm run data",
```
domoritz: Why?

PR #1194, opened 2016-02-28
Correct type for stacked bar horizontal order example

ZeningQu:
I was reading the docs and thought the `type` of `yield sum` should be `quantitative` instead of `nominal` in the `stacked bar horizontal order` example.


domoritz:
Good catch

PR #1205, opened 2016-03-01
Create LineInterpolation with valid interpolation options

mprudhom:
The interpolate values of "basis-closed", "bundle", and "cardinal-closed" are not valid in the vega spec (despite being valid D3 interpolations).

Should interpolate be an enum? I can implement that if you like.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -27,11 +27,7 @@ export const defaultCellConfig: CellConfig = {
   height: 200
 };
 
-export const defaultFacetCellConfig: CellConfig = {
-  stroke: '#ccc',
-  strokeWidth: 1
-};
-
+int
 export interface FacetConfig {
   scale?: FacetScaleConfig;
   axis?: AxisConfig;
@@ -95,6 +91,19 @@ export enum StackOffset {
     CENTER = 'center' as any,
     NORMALIZE = 'normalize' as any,
     NONE = 'none' as any,
+}
+
+export enum LineInterpolation {
+    LINEAR = 'linear' as any,
+    BUNDLE = 'bundle' as any,
+    MONOTONE = 'monotone' as any,
+    STEP_BEFORE = 'step-before' as any,
+    STEP_AFTER = 'step-after' as any,
+    BASIS = 'basis' as any,
+    BASIS_OPEN = 'basis-open' as any,
+    BASIS_CLOSED = 'basis-closed' as any,
+    CARDINAL = 'cardinal' as any,
+    CARDINAL_OPEN = 'cardinal-open' as any,
 }
 
 export interface MarkConfig {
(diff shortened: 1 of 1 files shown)


kanitw:
@mprudhom  Can you please help checking if putting these values in Vega produces desirable results?   

If so, we should close this issue and fix the schema inside Vega instead.

mprudhom:
They do in fact work in vega, and they could conceivably be used in a useful way in a vega spec. But I don't see any use cases for a closed interpolation in a line/area channel in vega-lite. 

Unless you can think of any scenario in which they would be useful, I propose they removed from vega-lite. E.g.:

```json
{
  "data": {
    "values": [
      {"a": "A","b": 1}, {"a": "B","b": 3}, {"a": "C","b": 2}, {"a": "D","b": 4}
    ]
  },
  "mark": "line",
  "config": {
    "mark": { "interpolate": "cardinal-closed" }
  },
  "encoding": {
  }
}
```

unknown


And:

```json
{
  "mark": "area",
  "encoding": {
    "x": {
      "scale": {"nice": "month"},
      "axis": {"axisWidth": 0, "format": "%Y", "labelAngle": 0, "tickSize": 0}
    },
    "y": {
      "aggregate": "sum", "field": "count","type": "quantitative",
      "axis": false
    },
    "color": {"field":"series", "type":"nominal", "scale":{"range": "category20b"}}
  },
  "config": {"cell": {"width": 300, "height": 200}, "mark": {"stacked": "center","interpolate": "basis-closed" }}
}
```

unknown-1

kanitw:
Thanks for checking.  This makes sense.  Please feel free to add a commit to convert this property into an enum without nonsense options.  

(Another PR to Vega for fixing the schema would be useful as well.)

domoritz:
I don't think that the interpolations never make sense and if we only need to add them to the enum, why omit some?

src/config.ts
```diff
@@ -177,9 +206,9 @@ export interface MarkConfig {
 
   // ---------- Interpolation: Line / area ----------
   /**
-   * The line interpolation method to use. One of linear, step-before, step-after, basis, basis-open, basis-closed, bundle, cardinal, cardinal-open, cardinal-closed, monotone.
+   * The line interpolation method to use. One of linear, step-before, step-after, basis, basis-open, cardinal, cardinal-open, monotone.
    */
-  interpolate?: string;
+  interpolate?: LineInterpolation;
```
kanitw: Let's name this `Interpolate`?  Because there is no other interpolate in the schema, no need to make different name.

kanitw: @mprudhom nvm.  I'll just do it myself.

mprudhom: Oops, sorry I missed that request.

PR #1207, opened 2016-03-01
Make strokeDashOffset a number

mprudhom:
Vega's strokeDashOffset is a number (or signal), not an array of numbers.


kanitw:
Thanks for catching this.

PR #1214, opened 2016-03-04
Support `scale.bandSize` = `"fit"`, Top-level `width` and `height`

### LINKED ISSUE #194: Responsive Ordinal Scale (Ordinal Fit Mode)
Currently  `config.cell.width`, `config.cell.height`  are ignored for x and y if they ordinal scales respectively.  
Basically, the cell width and height of an ordinal scale is determined by the `band size * cardinality`.   

- [ ] ~~Let's add `config.unit.fitOrdinalScale` as boolean to make the ordinal scale rely on the `config.unit.width/height` value instead of bandWidth.~~  See below
- [ ] Need to deal make the scale work with all mark type.  
- [ ] corner case when the cardinality of the ordinal field is too high
- [ ] Of course, add tests.

kanitw:
Fix #194


domoritz:
Nice!

kanitw:
We also need to decide whether we want to allow `padding` when `bandSize` is a number and thus the underlying ordinal scale is a point ordinal scale.

domoritz:
> We also need to decide whether we want to allow padding when bandSize is a number and thus the underlying ordinal scale is a point ordinal scale.

I'd say lat's reduce the attack surface unless we have a compelling example.

kanitw:
Yeah. In any case, I think it is hard to reason about this because when I experiment with properties, there is still this bug https://github.com/vega/vega/issues/502.   Thus, something I see isn't the correct behavior.

kanitw:
Before we commit to this syntax, another alternative to scale.bandWidth = `'fit'` is to use `points` property to determine whether the x/y ordinal scale is a point scale or band scale (band scale is when `points` = `false`), but I found that more confusing.

kanitw:
- [ ] Test Docs
- [ ] Update docs to warn about high cardinality case

domoritz:
@ellisonbg @jakevdp This is the fit feature we talked about today.

kanitw:
@domoritz  I've just updated the docs and should be in good shape.  Please let me know if you have any comment.  

I still have to change `round` to `false` by default when scale is `"fit"` as unrounded scale is more scalable for higher cardinality.

peller:
Looking forward to seeing this PR land.  However, I notice that the top-level width and height still relate to the plot area and not the overall chart with axes and labels.  The latter is what's needed for most apps doing layout.

kanitw:
> However, I notice that the top-level width and height still relate to the plot area and not the overall chart with axes and labels. The latter is what's needed for most apps doing layout.

Vega 2 doesn't support this well yet anyway.  However, Vega 3 will have better support for this and we're planning to use Vega 3 for Vega-Lite 2.

arvind:
> Vega 2 doesn't support this well yet anyway.

Vega support a `strict` padding option which lays out a visualization such that, after axes and legends have been added, the top-level `width` and `height` is respected.

kanitw:
@arvind  Oh this seems new to me.  Good to know then.   That said, to support it in Vega-Lite would be a separate issue from this PR.

peller:
Sorry, one more issue.  I tried using the [View Component API](https://github.com/vega/vega/wiki/Runtime#view-component-api) on the `kw/fit` branch to set width and height and it had no effect.  Is this a vega-lite bug?

kanitw:
> Is this a vega-lite bug?

It's implemented in a different way, so size of inner components especially when faceted are set manually so yeah, you can't use it (yet).

PR #1216, opened 2016-03-04
Don’t make properties of DataComponent optional

domoritz:


test/compile/data.test.ts
```diff
@@ -342,6 +343,7 @@ describe('data: timeUnitDomain', function() {
     });
 
     it('should be compiled into correct string set', function() {
+      model.component.data = {} as DataComponent;
```
kanitw: Having to do this for every line of test looks annoying.   I think it doesn't hurt to init `data: {}` above.

kanitw: above = in `model.ts` constructor

domoritz: The advantage is that we only have to do this in the tests and we can guarantee that the properties are not undefined in the code. So we eliminate possible bugs at the cost of failures in the tests if you forget to add this line.

PR #1222, opened 2016-03-05
Fix generation of wrong field reference in scale. Also add a test for it.

domoritz:


src/compile/scale.ts
```diff
@@ -122,7 +123,7 @@ function parseBinColorLegendLabel(model: Model, fieldDef: FieldDef) {
     type: 'ordinal',
     domain: {
       data: model.dataTable(),
-      field: model.field(COLOR,  {prefn: 'rank_'}),
+      field: model.field(COLOR, (fieldDef.bin || fieldDef.timeUnit) ? {} : { prefn: 'rank_' }),
```
kanitw: @domoritz   Now I know when this got wrong.  

On this line, `fieldDef.bin` is always `true`.  (See how `parseBinColorLegendLabel` gets called.)  So I asked you to remove `fieldDef.bin`, but I guess you put in the wrong condition.


I think the right fix here is `model.field(COLOR, {})`

test/compile/scale.test.ts
```diff
@@ -349,6 +349,11 @@ describe('Scale', function() {
       assert.deepEqual(scales[1].range, {
         data: 'source',
         field: 'bin_origin_range',
+        sort: { "field": "bin_origin_start", "op": "min" }
```
kanitw: why do we have to assert the `scales` twice?

domoritz: I meant domain. Ups.

domoritz:
Done

Issue #1174, opened 2016-02-23
Start page looks different in Firefox vs Chrome


domoritz:
The height of the images is different. 

<img width="1079" alt="screen shot 2016-02-23 at 00 05 11" src="https://cloud.githubusercontent.com/assets/589034/13245327/2cc9bb80-d9c1-11e5-8f53-5b8c92b8f1a3.png">
<img width="1070" alt="screen shot 2016-02-23 at 00 05 15" src="https://cloud.githubusercontent.com/assets/589034/13245331/30cc2aba-d9c1-11e5-8574-08a9b686cfbd.png">

willium:
This will need to redone. Probably okay for v1?

willium:
I'm not entire sure how, but this seems to work fine for me as-is. 
![screen shot 2016-03-06 at 2 17 49 am](https://cloud.githubusercontent.com/assets/356976/13553533/b7f7a2c4-e341-11e5-9555-e408b3bcefaa.jpg)

domoritz:
Yeah, something changed. The plots used to not be squares but rectangles that are slightly wider than tall.

willium:
Good to close?

domoritz:
Let me see at what point the rectangles became squares. I feel the showcase is too tall now and we could show more content.

domoritz:
Looks like it was changed in https://github.com/vega/vega-lite/commit/b72da4622bbb624fd2810cec6c075c00b5ca4c44. We were under time pressure to fix it so the patch was fine. I created this issue when we merged so we can clean things up now.

Issue #1213, opened 2016-03-02
Move usage and installation instructions


willium:
Currently the most primary information is buried, we should move it to the top or homepage

PR #1225, opened 2016-03-06
Small fixes throughout docs, re-ordering usage instructions

willium:
- [x] fix docs in smaller screen sizes
- [x] move usage up, rewrite
- [x] convert all double spaces to single spaces
- [x] responsive wide code blocks and wide visualizations


willium:
Also see: https://github.com/vega/vega-lite/issues/1226

willium:
Fixes #1174, #1213

site/docs/using.md
```diff
@@ -2,7 +2,7 @@
 layout: docs
 menu: docs
 title: Embedding Vega-Lite in a Web Page
```
kanitw: We can use Vega-Lite in command line via vl2png, vl2svg too?   

Should we change title to Using Vega-Lite and add more sections for those?   
(And possibly add in-page TOC on the top of the page?)

kanitw: or "Usage Instructions"

site/docs/docs.md
```diff
@@ -47,10 +47,3 @@ In Vega-Lite, a specification can have the following top-level properties.
 | [mark](mark.html)    | String        | The mark type.  One of `"bar"`, `"circle"`, `"square"`, `"tick"`, `"line"`, `"area"`, `"point"`, and `"text"`. |
 | [encoding](encoding.html) | Object   | A key-value mapping between encoding channels and definition of fields. |
 | [config](config.html)   | Object     | Configuration object. |
-
```
domoritz: Why delete and not move up?

willium: I don't think it's a good idea to have usage instructions partially in 3 places, but I can put it back (and up) if you think thats better.

kanitw: I think if we're moving the menu up already, it makes sense to remove it.

domoritz: So the overview page only has links to the spec docs? I have to see it to be convinced that it's a good idea.

willium:
I rewrote Usage Instructions, Changed Tutorials to Guides, and move it there. I know we discussed this and I believe @domoritz was against it. I can see why, but I think I like this more than before. Thoughts?

domoritz:
> I rewrote Usage Instructions, Changed Tutorials to Guides, and move it there. I know we discussed this and I believe @domoritz was against it. I can see why, but I think I like this more than before. Thoughts?

I can't remember whether we agreed to call the tutorials "guides"? I think I like tutorials better but we can still put a link to the usage section in the menu for tutorials. 

I'd also say that we should leave the usage.md under docs/ but just put a link to it in the tutorials menu.

* Tutorials -> shows getting started
  * Getting Started
  * Analyze data
  * Using Vega-Lite -> only a link
* Docs -> shows overview
  * Overview, needs toc
  * Using Vega-Lite
  * all the remaining docs for the spec

willium:
I don't really think usage really is API documentation, and I find it weird in the docs for that reason. I'm not sure I really like either, but slightly prefer Guides to Tutorials as it lends itself more to future uses of this section (integration of Vega-Lite with x, etc.) which are probably not semantically "tutorials." I suppose it could be either or.

For the TOC on Overview would it be for the entire docs? I'm not sure that's possible to generate like we do with Jekyll. How specific would be want to get

willium:
What about moving Usage to the highest level (next to Documentation, Examples, Tutorials in the header). To Save room we can rename Documentation to Docs, but I actually just made the header responsive so it shouldn't be too much an issue.

domoritz:
Usage is definitely part of docs. Docs are not only API docs imho. 

> For the TOC on Overview would it be for the entire docs? I'm not sure that's possible to generate like we do with Jekyll. How specific would be want to get

You can only generate TOC for one page so the TOC on the overview page probably has to be written by hand.

willium:
Okay, I'll work on the TOC. How detailed should it be?

kanitw:
> Usage is definitely part of docs. Docs are not only API docs imho.

Another way to see it is  the whole web is the docs.

Guide = Tutorial + Usage Guide
and we can rename docs to Syntax

kanitw:
This way, we don't overload the "docs" subtree with too many things.

site/static/main.css
```diff
@@ -245,6 +250,30 @@ nav a:hover {
   margin-top: -4px;
   /* hack */
   margin-right: 8px;
+  flex: 0 0 auto;
+  display: inline-block;
+  line-height: 3em;
```
kanitw: Why is the header now taller and not centered?

kanitw:
> I'd also say that we should leave the usage.md under docs/ but just put a link to it in the tutorials menu.
* Tutorials -> shows getting started
  * Getting Started
  * Analyze data
  * Using Vega-Lite -> only a link
* Docs -> shows overview
  * Overview, needs toc
  * Using Vega-Lite
  * all the remaining docs for the spec


To be honest,  this is a very confusing information architecture.    I strongly suggest

1) Make docs -> syntax

2a)  Make tutorial -> guide and put usage in guide
or 2b)  Put usage as "Usage Guide" at the top-level

kanitw:
Thinking more about it .... Tutorial is a kind of usage guide?  so 2a) make more sense.

domoritz:
Docs and tutorials is more common than guides and syntax. I feel it will me more confusing even if it is more consistent. 

Let's look at some other (please help!)

* angular has top level docs with tutorial, guide (note that this is not a tutorial, it's a complete walkthrough without a running example), and more
* react has docs with quick start, guides, and reference (that's a good term, much better than "syntax"), note that guides are again not a tutorial (guides here are similar to our usage guide)
* top level documentation, then "learning rust" and references
* top level docs with "installing go", "learning go", references, articles
* top level learn with tutorial, handbook (like guides in the previous examples), samples, language spec

domoritz:
After looking at the examples, I propose (if we have enough space at the top level)

* Tutorials
  * Getting Started
  * Exploring Data
* Guides
  * Usage Guide
  * Faceting
  * Customizing Size
* Reference
  * Data
  * Mark
  * Encoding
  * ...

The three top level options are all docs but it makes sense to expand it.

kanitw:
Maybe we should have a card-sorting session.  This is indeed not an easy problem.

- Reference sounds like a more common term than syntax,  but sounds like we have to rename overview to "Specification" and move the "overview" part to guide.
- Faceting is  a weird page.  It will have also its place in "reference" after we have composition.  
- Another problem is that we wrote some language page as if they are guide.  For example, mark page is sort of a guide to all mark types.  

- I think combining tutorials and guides and make all pages except home using left side menu is another option too.  We can group the menu for guides like this:

**Tutorial** (No Link -- just category)
Getting Started
Exploring Data

**Usage Guide** (No Link -- just category)
- Embed in Web Applications
- Using on the server

**Advanced Guide**
- Faceting
- Customizing Size
- (We can add more advanced guide here)

Let's chat more tmr

kanitw:
Now my best proposal is we can leave it as is.  Just remove that weird "Usage Guide" jumpy link from tutorial and it kinda make sense already.  

(This way, the overview might wanna have a paragraph hinting about usage guide page though.)

kanitw:
Let's add a space between "Usage Instruction" and Data to suggest grouping.  

Also, we might want to make the left side menu font slightly smaller so it's smaller than the header?

willium:
Think this is good to merge

domoritz:
LGTM

domoritz:
One last thing. I'm sure we link to the old embed page somewhere. Please check for dead links and fix them.

kanitw:
LGTM too.

PR #1228, opened 2016-03-07
Seperate tests for schema validation

willium:
This doesn't really serve a purpose other than semantics, but I am debugging a breaking test and its nice to know which compilation is breaking. This asserts validation for both vl and the compiled vg.


domoritz:
That serves a good purpose. A very welcome change!

PR #1229, opened 2016-03-07
Add row/column to the encoding docs

domoritz:
At least we should have a reference to not confuse people about why we did not put it there. This can be reverted once we have composition.

### DIFF
--- a/site/docs/encoding.md
+++ b/site/docs/encoding.md
@@ -21,7 +21,7 @@ An integral part of the data visualization process is encoding data with visual 
     "size": ...,
     "shape": ...,
     "text": ...,
-    "detail": ...,
+    "detail": ...
   },
   ...
 }
--- a/site/docs/encoding.md
+++ b/site/docs/encoding.md
@@ -15,11 +15,13 @@ An integral part of the data visualization process is encoding data with visual 
   "encoding": {     // Encoding
     "x": ...,
     "y": ...,
+    "column": ...,
+    "row": ...,
     "color": ...,
     "size": ...,
     "shape": ...,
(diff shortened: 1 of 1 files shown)


site/docs/encoding.md
```diff
@@ -15,6 +15,8 @@ An integral part of the data visualization process is encoding data with visual
   "encoding": {     // Encoding
     "x": ...,
     "y": ...,
+    "column": ...,
```
kanitw: column row should be on the top as default format.

site/docs/encoding.md
```diff
@@ -41,7 +43,8 @@ Mark properties channels map data fields directly to visual properties of the ma
 | Property      | Type          | Description    |
 | :------------ |:-------------:| :------------- |
 | x, y          | [ChannelDef](#def)| X and Y coordinates for `point`, `circle`, `square`, `line`, `text`, and `tick`. (or to width and height for `bar` and `area` marks). |
-| color         | [ChannelDef](#def)| Color of the marks – either fill or stroke color based on mark type. (By default, fill color for `area`, `bar`, `tick`, `text`, `circle`, and `square` /   stroke color for `line` and `point`.) (See [scale range](scale.html#range) for more detail about color palettes.)  |
+| column, row  | [ChannelDef](#def)| `row` and `column` are special encoding channels for [faceting](#facet). |
```
kanitw: For here, since they are **special encoding channels**, let's put them at the end.  
(I think it's okay to make this slightly different from the spec above.)

domoritz:
Okay, I moved things around

PR #1230, opened 2016-03-07
[WIP] Add properties to axis

YuhanLu:
#1158 addedadded properties to axis and matching test cases, docs not finished yet  @kanitw

### DIFF
--- a/src/compile/axis.ts
+++ b/src/compile/axis.ts
@@ -186,6 +186,14 @@ export function tickSize(model: Model, channel: Channel) {
   return undefined;
 }
 
+export function tickSizeEnd(model: Model, channel: Channel) {
+    const tickSizeEnd = model.axis(channel).tickSizeEnd;
+    if (tickSizeEnd !== undefined) {
+        return tickSizeEnd;
+    }
+    return undefined;
+}
+
 
 export function title(model: Model, channel: Channel) {
   const axis = model.axis(channel);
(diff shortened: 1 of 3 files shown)


src/compile/axis.ts
```diff
@@ -186,6 +186,14 @@ export function tickSize(model: Model, channel: Channel) {
   return undefined;
 }
 
+export function tickSizeEnd(model: Model, channel: Channel) {
```
kanitw: Please also move `tickSizeEnd` above from the group below
`    // b) properties without rules, only produce default values in the schema, or explicit value if specified
`

to 

`    // a) properties with special rules (so it has axis[property] methods) -- call rule functions
`

See around line 72.  (We sort each group by alphabetical order)

src/compile/axis.ts
```diff
@@ -210,18 +218,42 @@ export function title(model: Model, channel: Channel) {
   return maxLength ? truncate(fieldTitle, maxLength) : fieldTitle;
```
kanitw: Same as `tickSizeEnd`

src/compile/axis.ts
```diff
@@ -210,18 +218,42 @@ export function title(model: Model, channel: Channel) {
...
 }
 
+export function titleOffset(model: Model, channel: Channel) {
```
kanitw: Same as `tickSizeEnd`

src/compile/axis.ts
```diff
@@ -210,18 +218,42 @@ export function title(model: Model, channel: Channel) {
...
+    }
+    return undefined;
+}
+
 export namespace properties {
   export function axis(model: Model, channel: Channel, axisPropsSpec, def) {
     const axis = model.axis(channel);
 
     return extend(
+      axis.axisColor !== undefined ?
```
kanitw: Please use consistent tabbing style

src/compile/axis.ts
```diff
@@ -210,18 +218,42 @@ export function title(model: Model, channel: Channel) {
...
+      {stroke: {value: axis.axisColor}} : {},
       axis.axisWidth !== undefined ?
-        { strokeWidth: {value: axis.axisWidth} } :
-        {},
+        {strokeWidth: {value: axis.axisWidth }} : {},
+
+      axis.tickColor !== undefined ? { ticks: {stroke : {value: axis.tickColor} } } : {},
```
kanitw: Does this really work correctly?  This looks incorrect.

kanitw:
Please also rebase to master :)

src/compile/axis.ts
YuhanLu: Do you mean we need to group them in to a) and b) that all the functions in a) are above b) and in alphabetical order?

kanitw: I mean move `'tickSizeEnd'` from line 75 to line 73 and keep them in alphabetical order.  No need to move the whole method.  :)

PR #1234, opened 2016-03-08
Add Table of Contents in Docs overview

willium:
Add script to generate a global docs TOC file in markdown

added `npm run build:toc`


willium:
screen shot 2016-04-12 at 3 38 02 am

willium:
This is a little quick and dirty, the code's not beautiful, but hey it works.

domoritz:
Sweet. Could you change one thing and put the toc into a markdown file in `_includes` so that we can include it anywhere with `{% include doc_toc.md %}`? I think I prefer to have to just below the overview on the first page rather than on a separate page.

willium:
Sure. I thought that'd make it harder to notice the other sections on overview (they'd be pushed down the page)

domoritz:
Thanks! We can still move the include but it'll be much easier if the doc toc is an include.

kanitw:
Screenshot looks awesome!  Great work!

scripts/generate-toc
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env node
+'use strict';
+
+var fs = require('fs'),
+		cheerio = require('cheerio'),
+		yamlFront = require('yaml-front-matter');
+
+var URL = 'https://vega.github.io/vega-lite/docs/';
```
domoritz: Could you change this to `{{site.baseurl}}/docs/`?

willium:
Fixed @domoritz!

Issue #571, opened 2015-09-02
support "displayName" on encodings


jowens:
All of the gallery visualizations save one use the internal data field name as the axis label. It is likely that users will want to use a different axis label name as the internal data field name (no one really likes underscores in axis labels, do they?). 

In the gallery, the binned plot example has a "size" encoding, which has a "displayName" field, and the displayName is what's displayed atop the legend. However, displayName does not appear to work everywhere (try it on, say, the y axis of a bar plot). It'd be nice if it did.

kanitw:
We actually added   `axis.title` property for this.  

try
```
encoding: {
  x: { ..., axis: {title: "custom title"} }
}
```
(Again, apologies for lack of documentation. We're still in an alpha state but we'll surely have a detailed documentation when we release 1.0)

jowens:
Also I note that axis.title doesn't back-propagate to "shorthand". I like "shorthand"; it'd be nice if axis.title could be part of it.

PR #1238, opened 2016-03-10
Use displayName for field title if it is specified

mprudhom:
Fixes https://github.com/vega/vega-lite/issues/571

### DIFF
--- a/src/fielddef.ts
+++ b/src/fielddef.ts
@@ -175,6 +175,9 @@ export function cardinality(fieldDef: FieldDef, stats, filterNull = {}) {
 }
 
 export function title(fieldDef: FieldDef) {
+  if (fieldDef.displayName != null) {
+    return fieldDef.displayName;
+  }
   if (isCount(fieldDef)) {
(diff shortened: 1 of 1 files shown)


kanitw:
I'm not sure if `displayName` is the best property name, maybe we should consider other alternatives before sticking to this one. 

`title` or `fieldTitle` are also considerable options.    

@domoritz  thoughts?

mprudhom:
My vote would be for "title".

ZeningQu:
I'm using `fieldTitle` in the VL tooltip.

domoritz:
Wait, we already have axis.title and legend.title. I think an explicit title property that can be used for both is a good idea.

kanitw:
The only con of title is it is same name as the whole spec's title, which
is coming soon.

domoritz:
> The only con of title is it is same name as the whole spec's title, which
is coming soon.

But it's in the channelDef/fieldDef and not the first level of the spec. I don't think that's an issue.

PR #1240, opened 2016-03-13
Export timeunit.format() for VgTooltip

ZeningQu:
Export `timeunit.format(timeUnit, abbreviated)` because tooltip needs to translate VL timeUnit to the time string specifier.

### DIFF
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -6,7 +6,7 @@ import {contains} from '../util';
 
 import {FacetModel} from './facet';
 import {Model} from './model';
-import {format as timeFormatExpr} from './time';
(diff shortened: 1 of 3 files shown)


domoritz:
Do we have to move the function? It's very inconsistent now.

kanitw:
@domoritz  what's inconsistent?  (Ideally we shouldn't expose methods inside compile to the public?)

domoritz:
What's inconsistent is that other methods stayed in time.ts (https://github.com/vega/vega-lite/blob/zqu/export-format/src/compile/time.ts). What's different about format? Is it the fact that it is made available to the outside? I don't have strong opinions just try to understand the reasoning.

PR #1241, opened 2016-03-14
Layer

domoritz:
- [x] merge scales
- [x] layout
- [x] axes
- [x] legends
- [x] marks
- [x] data
  - [x] generate separate data source for different filters
  - [x] correctly set up bin

Need to get these working:

```json
{
  "data": {"url": "data/stocks.csv","formatType": "csv"},
  "layers": [
    {
      "mark": "line",
      "encoding": {
        "x": {"field": "date","type": "temporal"},
        "y": {"field": "price","type": "quantitative"},
        "color": {"field": "symbol","type": "nominal"}
      }
    },
    {
      "mark": "rule",
      "encoding": {
        "x": {
          "field": "date",
          "type": "temporal",
          "aggregate": "mean"
        },
        "color": {"value": "red"}
      }
    }
  ]
}
```

screen shot 2016-03-16 at 19 04 07

```json
{
  "data": {"url": "data/flights-2k.json"},
  "layers": [
    {
      "mark": "bar",
      "encoding": {
        "x": {
          "field": "distance",
          "type": "quantitative",
          "bin": true
        },
        "y": {
          "aggregate": "count",
          "field": "*",
          "type": "quantitative"
        }
      }
    },
    {
      "transform": {"filter": "datum.delay < 5"},
      "mark": "bar",
      "encoding": {
        "x": {
          "field": "distance",
          "type": "quantitative",
          "bin": true
        },
        "y": {
          "aggregate": "count",
          "field": "*",
          "type": "quantitative"
        },
      }
    }
  ]
}
```

screen shot 2016-03-17 at 17 12 51


domoritz:
somehow this doesn't work

```
{
      "mark": "rule",
      "encoding": {
        "y": {"field": "c", "type": "quantitative", "aggregate": "mean"}
      },
      "config": {"mark": {"stroke": "gray", "strokeWidth": 10}}
    }
```

src/compile/model.ts
```diff
@@ -231,27 +231,34 @@ export abstract class Model {
...
   public dataName(dataSourceType: DataTable): string {
-    return this._dataNameMap.get(this.name(dataSourceType + ''));
+    return this._dataNameMap.get(this.name(String(dataSourceType)));
+  }
+
+  /**
+   * Function to rename the dataset name using the name map. Used in the assemble step.
+   */
+  public dataRename(name: string): string {
+    return this._dataNameMap.get(name);
```
kanitw: This is `get`ting, not rename.   I also have the `renameData` method right above on line 226.

src/compile/data.ts
```diff
@@ -383,6 +440,11 @@ export namespace bin {
     return binComponent;
   }
 
+  export function parseLayer(model: LayerModel) {
+    // TODO
```
kanitw: TODO (just mark so we know we still have unfinished item)

kanitw: TODO (just mark so we know we still have unfinished item in the PR)

src/compile/data.ts
```diff
@@ -564,7 +667,16 @@ export namespace summary {
...
+  export function parseLayer(model: LayerModel): SummaryComponent[] {
+    // we create new data sources for each summary in a layer
+    return [];
+  }
+
+  /**
+   * Assemble the summary. Needs a rename function because we cannot guarantee that the
+   * parent data before the children data.
+   */
+  export function assemble(component: DataComponent, rename: (name: string) => string): VgData[] {
```
kanitw: As discussed, we should not do rename in assemble.

src/compile/data.ts
```diff
@@ -564,7 +667,16 @@ export namespace summary {
     return [];
   }
 
-  export function assemble(component: DataComponent): VgData[] {
+  export function parseLayer(model: LayerModel): SummaryComponent[] {
+    // we create new data sources for each summary in a layer
+    return [];
```
kanitw: The right thing to do is to move all summary from its children.  (remove from children) and merge summary with the same group by fields and do the rename in here.

src/compile/data.ts
```diff
@@ -650,6 +762,11 @@ export namespace stackScale {
     return null;
   }
 
+  export function parseLayer(model: LayerModel) {
+    // TODO
```
kanitw: //TODO (Just marking so we can see in the PR)

src/compile/data.ts
```diff
@@ -736,6 +860,11 @@ export namespace colorRank {
     return {} as Dict<VgTransform[]>;
   }
 
+  export function parseLayer(model: LayerModel) {
+    // TODO
```
kanitw: //TODO (Just marking so we can see in the PR)
 Add a line note

src/compile/data.ts
```diff
@@ -769,6 +898,11 @@ export namespace nonPositiveFilter {
     return {} as StringSet;
   }
 
+  export function parseLayer(model: LayerModel) {
+    // TODO
```
kanitw: //TODO (Just marking so we can see in the PR)

src/compile/layer.ts
```diff
@@ -0,0 +1,228 @@
...
+    // layer does not have any channels
+    return false;
+  }
+
+  public children() {
+    return this._children;
+  }
+
+  private hasSummary() {
+    // TODO: don't just use the first child
```
kanitw: Not sure why you don't just implement it correctly.

src/compile/data.ts
domoritz: I'll actually do this one because it's needed for an example.

src/compile/data.ts
domoritz: That's not what we discussed. We discussed that we should not set the names in parse ;-P

src/compile/data.ts
domoritz: Are you saying this is the right thing or the thing to do now? Or that I should just add a comment?

src/compile/layer.ts
```diff
@@ -0,0 +1,228 @@
...
+              fields: existingDomain.fields.concat([newScales[0].domain])
+            };
+            if (existingScales.length > 1) {
+              delete existingScales[0];
+              scaleComponent[channel] = newScales.concat(existingScales);
+            } else {
+              scaleComponent[channel] = newScales;
+            }
+          } else {
+            // Use the first quantitative scale
```
kanitw: Why is quantitative scale not unioned?

src/compile/layer.ts
```diff
@@ -0,0 +1,228 @@
...
+      if (true) { // if shared/union scale
+        keys(child.component.scale).forEach(function(channel) {
+          let newScales = child.component.scale[channel] as ScaleComponent[];
+          if (!newScales) {
+            return;
+          }
+
+          const existingScales = scaleComponent[channel];
+          if (existingScales && newScales[0].type === ScaleType.ORDINAL) {
+            // Ordinal scales are unioned by combining the domain of the first scale. Other scales are appended.
```
kanitw: Looks like we should write a test case for merging color

src/compile/layer.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    let scaleComponent = this.component.scale = {} as Dict<ScaleComponent[]>;
+
+    this._children.forEach(function(child) {
+      child.parseScale();
+
+      // TODO: correctly implement independent scale
+      if (true) { // if shared/union scale
+        keys(child.component.scale).forEach(function(channel) {
+          let newScales = child.component.scale[channel] as ScaleComponent[];
```
kanitw: `newScales`  => `childScale`?

src/compile/layer.ts
```diff
@@ -0,0 +1,228 @@
...
+          if (!newScales) {
+            return;
+          }
+
+          const existingScales = scaleComponent[channel];
```
kanitw: `existingScales` => `scales` or `modelScales` or something that suggests that it is this model's.

src/compile/layer.ts
domoritz: I removed it because we cannot merge ordinal so why bother now?

src/compile/layer.ts
```diff
@@ -0,0 +1,228 @@
...
+          if (existingScales && newScales[0].type === ScaleType.ORDINAL) {
+            // Ordinal scales are unioned by combining the domain of the first scale. Other scales are appended.
+            const existingDomain = existingScales[0].domain.fields ? existingScales[0].domain : { fields: [existingScales[0].domain] };
+            newScales[0].domain = {
+              // assumes that the new scale domain is not a list
+              fields: existingDomain.fields.concat([newScales[0].domain])
+            };
+            if (existingScales.length > 1) {
+              delete existingScales[0];
```
kanitw: better write comment what's happening here.  why is this guy removed.  

I assume this case is currently only for color? 
Maybe worth changing the order in `scale.ts` so that `scales[0]` is always the main color scale and `[1]` is `colorLegend` ...

src/compile/layer.ts
```diff
@@ -0,0 +1,228 @@
...
+          } else {
+            // Use the first quantitative scale
+            if (!scaleComponent[channel]) {
+              scaleComponent[channel] = newScales;
+            }
+          }
+
+          // for each new scale, need to rename old scales
+          newScales.forEach(function(scale) {
+            const scaleNameWithoutPrefix = scale.name.substr(child.name('').length);
```
kanitw: `child.name('').length`  can be called once outside `forEach`?

src/compile/layout.ts
```diff
@@ -72,7 +72,9 @@ export function parseUnitLayout(model: UnitModel): LayoutComponent {
 }
 
 function parseUnitSizeLayout(model: UnitModel, channel: Channel): SizeComponent {
-  const nonOrdinalSize = channel === X ? model.cellWidth() : model.cellHeight();
+  // TODO: think about whether this config has to be the cell or facet cell config
```
kanitw: always cell config, and unit inside facet should have its config overridden by facet?

src/compile/layout.ts
```diff
@@ -142,6 +144,39 @@ function facetSizeFormula(model: Model, channel: Channel, innerSize: string) {
...
+  return {
+    width: parseLayerSizeLayout(model, X),
+    height: parseLayerSizeLayout(model, Y)
+  };
+}
+
+function parseLayerSizeLayout(model: LayerModel, channel: Channel): SizeComponent {
+  if (true) {
+    // For shared scale, we can simply merge the layout into one data source
+    // TODO: don't just take the layout from the frist child
```
kanitw: Let's file an issue for this.

src/compile/mark/mark.ts
```diff
@@ -36,7 +38,7 @@ export function parseMark(model: UnitModel): any[] {
 function parsePathMark(model: UnitModel) { // TODO: extract this into compilePathMark
   const mark = model.mark();
   // TODO: replace this with more general case for composition
-  const hasParentData = !!model.parent();
+  const isFaceted = model.parent() && model.parent().isFacet();
```
kanitw: What happens if the facet is a grandparent (or great grand ..)

kanitw: This seems wrong.   Also what if it's a child of layer where layer has data.

src/compile/mark/mark.ts
```diff
@@ -94,7 +96,7 @@ function parsePathMark(model: UnitModel) { // TODO: extract this into compilePat
 
 function parseNonPathMark(model: UnitModel) {
   const mark = model.mark();
-  const hasParentData = !!model.parent();
+  const isFaceted = model.parent() && model.parent().isFacet();
```
kanitw: This seems wrong.   Also what if it's a child of layer where layer has data.

src/compile/data.ts
kanitw: I also say that rename should happen in parse, not in assemble.  assemble should just read the name.  That's it.  `rename` parameter here should be removed.

src/compile/data.ts
kanitw: We should do it now, so that we don't need `rename` in assemble as commented above.

kanitw: We should do it now, so that we don't need rename in assemble as commented below.

domoritz: Done

src/compile/data.ts
domoritz: The summary source will be set in assemble.

domoritz:
We get warnings when scales cannot be unioned

```
{
  "data": {"url": "data/stocks.csv","formatType": "csv"},
  "layers": [
    {
      "mark": "line",
      "encoding": {
        "x": {"field": "date","type": "temporal"},
        "y": {"field": "price","type": "quantitative"},
        "color": {"field": "symbol","type": "nominal"}
      }
    },
    {
      "mark": "rule",
      "encoding": {
        "y": {
          "field": "price",
          "type": "quantitative",
          "aggregate": "mean",
          "scale": {
            "domain": [1,2,3]
          }
        },
        "size": {"value": 2},
        "color": {"field": "symbol","type": "nominal"}
      },
      "config": {"mark": {"opacity": 0.5}}
    }
  ]
}
```

domoritz:
Test failures are due to https://github.com/vega/vega/issues/556

src/compile/layout.ts
domoritz: Should be confirmed so I'll leave the comment here.

src/compile/layout.ts
domoritz: https://github.com/vega/vega-lite/issues/1251

src/compile/mark/mark.ts
domoritz: Can you give me a concrete example?

src/compile/mark/mark.ts
domoritz: Can you give me a concrete example?

domoritz:
Ready for review @kanitw. 

@arvind I think most of the churn is over. I'll create a separate pr for facets over values and the lookup transform.

arvind:
Suhweet! Excellent timing because I just got done with a block of writing. Will dive into the code this weekend and start porting over my `as/selections` branch.

src/compile/layer.ts
```diff
@@ -225,4 +213,17 @@ export class LayerModel extends Model {
...
+  /**
+   * Returns true if the child either has no source defined or uses the same url.
+   * This is useful if you want to know whether it is possible to move a filter up.
+   * 
+   * This function can only be called once th child has been parsed.
+   */
+  public compatibleSource(child: UnitModel) {
+    const sourceUrl = this.data().url;
+    const childData = child.component.data;
+    const compatible = !childData.source || sourceUrl && sourceUrl === childData.source.url;
```
kanitw: @domoritz   Could you please really stop mixing up  `||` and `&&` without parentheses?

src/compile/mark/mark.ts
kanitw:
```
{
   data: ...,
   layer:  [{  
      // no data here
      mark: ...
      encoding: ...
   }, {  
      // no data here
      mark: ...
      encoding: ...
   }]
}
```

src/compile/mark/mark.ts
kanitw:
```
{
   data: ...,
   layer:  [{  
      // no data here
      mark: ...
      encoding: ...
   }, {  
      // no data here
      mark: ...
      encoding: ...
   }]
}
```

src/compile/data.ts
```diff
@@ -120,18 +117,22 @@ export function parseFacetData(model: FacetModel): DataComponent {
 
 export function parseLayerData(model: LayerModel): DataComponent {
   return {
-    source: source.parseLayer(model),
+    // filter and formatParse could cause us to not be able to merge into parent
```
kanitw: What's the rationale for reordering here?   

In any case, we should order this consistently for facet and unit for readability.

src/compile/layer.ts
domoritz: Can we add a lint rule so I don't forget?

src/compile/data.ts
```diff
@@ -263,9 +274,9 @@ export namespace source {
       sourceData.transform = [].concat(
         nullFilter.assemble(component),
         formula.assemble(component),
-        filter.assemble(component),
         bin.assemble(component),
-        timeUnit.assemble(component)
+        timeUnit.assemble(component),
+        filter.assemble(component)
```
kanitw: Do we really need filter to happen the last, because this is **very bad** for explaining the order of top-level `transform` and field inline transform.  (field transform came later, but with this change, this is no longer true.)

src/compile/data.ts
```diff
@@ -287,13 +298,11 @@ export namespace formatParse {
     // if there is no parse needed
     model.forEach(function(fieldDef: FieldDef) {
       if (fieldDef.type === TEMPORAL) {
-        parseComponent = parseComponent || {};
```
kanitw: Have you thoroughly tested that removing this doesn't cause problem?

domoritz: If you look above, you see that there is a `let parseComponent: Dict<string> = {};` so this is fine.

src/compile/data.ts
```diff
@@ -479,12 +506,13 @@ export namespace nullFilter {
   }
 
   export function parseLayer(model: LayerModel) {
+    // note that we run this before source.parseLayer
     let nullFilterComponent = parse(model);
```
kanitw: We might not even need to call `parse` here because layers doesn't have its own field.  This is always an empty object.

src/compile/data.ts
```diff
@@ -479,12 +506,13 @@ export namespace nullFilter {
...
 
     model.children().forEach((child) => {
       const childDataComponent = child.component.data;
-      if (!childDataComponent.source && childDataComponent.nullFilter) {
-        extend(nullFilterComponent || {}, childDataComponent.nullFilter);
+      if (model.compatibleSource(child) && !differ(childDataComponent.nullFilter, nullFilterComponent)) {
```
kanitw: If two children of the layer have the same `nullFilterComponent` they still don't get merge here if I read this correctly.
Because each of the child won't be equal to the layer's initial `nullFilterComponent`.

src/compile/data.ts
kanitw: Yep.  But before this PR, a part of `source` assemble code assumes that `formatParse` component either contains something or is null.    I'm not saying that this is worse, but need to make sure that this doesn't break those parts.

domoritz:
note to self: timeUnit should not be executed before source in data.ts

src/compile/mark/mark.ts
domoritz: The two examples I provided in the description work and have data in the layer only.

src/compile/mark/mark.ts
domoritz: same

src/compile/data.ts
domoritz: I fixed that part

src/compile/data.ts
```diff
@@ -667,16 +690,63 @@ export namespace summary {
...
+          // the key is a key of the dimensions
+          // we use it to find out whether we have a summary that uses the same group by fields
+          const key = stringSetHash(summary.dimensions);
+          if (key in summaries) {
+            // yes, there is a summary hat we need to merge into
+            // we know that the dimensions are the same so we only need to merge the measures
+            mergeMeasures(summaries[key].measures, summary.measures);
+          } else {
+            // give the summary a new name
+            summary.name = model.dataName(SUMMARY) + '_' + keys(summaries).length;
```
kanitw: > `keys(summaries).length`

Should we name it by the dimensions included instead of a summary counter?

src/compile/data.ts
```diff
@@ -667,16 +690,63 @@ export namespace summary {
     return [];
   }
 
+  function mergeMeasures(parentMeasures, childMeasures) {
+    for (var field in childMeasures) {
```
kanitw: This makes me wonder if we should (later) refactor all `Dict` to be a class.

src/compile/layer.ts
```diff
@@ -89,39 +87,62 @@ export class LayerModel extends Model {
...
+            // Scales are unioned by combining the domain of the first scale (the main scale).
+            // Other scales that are used for ordinal legends are appended.
+            const modelDomain = modelScales.main.domain;
+            const childDomain = childScales.main.domain;
+
+            if (isArray(modelDomain)) {
+              if (isArray(childScales.main.domain)) {
+                modelScales.main.domain = modelDomain.concat(childDomain);
+              } else {
+                model.addWarning('scales cannot be unioned with explicit domain');
```
kanitw: Do you mean

"custom domain scale cannot be unioned with default field-based domain?"

src/compile/layer.ts
```diff
@@ -89,39 +87,62 @@ export class LayerModel extends Model {
...
+              // TODO: if all domains use the same data, we can merge them
+              if (fields.length > 1) {
+                modelScales.main.domain = { fields: fields };
+              } else {
+                modelScales.main.domain = fields[0];
+              }
             }
+
+            // create color legend and color legend bin scales if we don't have them yet
+            modelScales.colorLegend = modelScales.colorLegend ? modelScales.colorLegend : childScales.colorLegend;
```
kanitw: Why does domain of the color field in `colorLegend` never get merged?   

It seems like there are 3 cases:

- **ordinal**:   this is kinda dumb to simply merge two ordinal fields together.  It seems like the right thing to do is to provide a custom domain for the union domain.   (Another VL 1.0 bug: I also notice that the ordinal color scale would behave incorrect if we provide explicit custom domain)
- **bin**:  This is tricky.   It seems like the field should be unioned before binning, I haven't tracked back the code to see how this should work.  
- **timeUnit**:  Similar to bin. 

Maybe we should file an issue for this and fix them later as it should not be show stopping for our paper's demo?

src/compile/mark/mark.ts
kanitw: Ok.  I think you're right about `layer` with children deriving data from parents because we don't add additional group.

However, this still seems wrong if the facet is a grandparent:

```
{
   data: ...,
   facet: ...,
   spec: {
    layer:  [{  
        // no data here and its parent is not a facet. 
        mark: ...
        encoding: ...
    }, {  
        // no data here
        mark: ...
        encoding: ...
    }]
  }
}
```

and is also wrong in the case of splom.

The right name should be `useFacetData`  (which can be parent / grandparent in the model tree + only refer to the case where the facet children doesn't have its own data).

src/compile/data.ts
kanitw: Why is this one ignored?

src/compile/data.ts
domoritz: Good point. I changed the way summaries are generated and now it would make more sense to use the fields (+the source name?). But I don't want to worry about all the edge cases (explicit values, ...)

domoritz: Ups, our key may be wrong in that case...

kanitw: Not sure what "explicit values" mean in this case?

kanitw: > Ups, our key may be wrong in that case...
?? Can you elaborate?

domoritz: `summary_bin_distance_start_bin_distance_mid_bin_distance_end_bin_distance_range` nope nope nope

src/compile/data.ts
domoritz: Not needed for examples

src/compile/layer.ts
domoritz: Fixed

src/compile/mark/mark.ts
domoritz: I'll fix the splom separately.

src/compile/data.ts
kanitw: Ok  let's use number for now although a semantically meaningful mean could just be `summary_bin_distance`

src/compile/data.ts
kanitw: @domoritz  This minor one got eaten too.

src/compile/data.ts
domoritz: maybe. I don't quite know what happens if we reference different datasets. I think we should forbid different datasets somehow!

src/compile/mark/mark.ts
kanitw: Sure, we should at least put the right name in and add TODO for the splom and grand parent case.

src/compile/data.ts
kanitw: Do we have an issue for this?  If not, please file one in the follow-up milestone

domoritz: https://github.com/vega/vega-lite/issues/1254

src/compile/mark/mark.ts
domoritz: Will do in follow up

src/compile/data.ts
domoritz: Oh yeah. I saw this comment and wanted so say YES! 
Maybe use this https://github.com/basarat/typescript-collections

src/compile/data.ts
```diff
@@ -416,9 +506,29 @@ export namespace nullFilter {
     return nullFilterComponent;
   }
 
+  export function parseLayer(model: LayerModel) {
+    // note that we run this before source.parseLayer
+
+    // FIXME: null filters are not properly propagated right now
```
kanitw: It's not only "not properly propagated".   The merge logic below is still incorrect.  I'll add this to #1254.

src/compile/data.ts
kanitw: I think forbidding simplify things for now, but not sure if we should forbid this in long run.

src/compile/data.ts
kanitw: SGTM especial if we can import only what we use.  We could consider that later.

PR #1259, opened 2016-03-20
[WIP] Repeat

### LINKED ISSUE #1257: Repeat
For cross filter we need something that resembles
```json
{
  "repeat": {
    "row": ["petalWidth","petalLength","sepalWidth","sepalLength"]
  },
  "spec": {
    "data": {"url": "data/iris.json"},
    "mark": "point",
    "encoding": {
      "x": {"field": "petalWidth","type": "Q"},
      "y": {"field": {"repeat": "row"},"type": "Q"},
      "color": {"field": "species","type": "N"}
    }
  }
}
```

Make sure this also works with binning! 

And splom
```json
{
  "repeat": {
    "row": ["petalWidth", "petalLength", "sepalWidth", "sepalLength"],
    "column": ["petalWidth", "petalLength", "sepalWidth", "sepalLength"]
  },
  "spec": {
    "data": {"url": "data/iris.json"},
    "mark": "point",
    "encoding": {
      "x": {"field": {"repeat": "column"}, "type": "Q"},
      "y": {"field": {"repeat": "row"}, "type": "Q"},
      "color": {"field": "species", "type": "N"}
    }
  }
}
```

The first one should compile to something like 

```json
{
  "width": 150,
  "height": 600,
  "data": [
    {"name": "iris","url": "data/iris.json"},
    {
      "name": "fields",
      "values": [
        "petalWidth",
        "petalLength",
        "sepalWidth",
        "sepalLength"
      ]
    }
  ],
  "scales": [
    {
      "name": "gy",
      "type": "ordinal",
      "range": "height",
      "round": true,
      "reverse": true,
      "domain": {"data": "fields","field": "data"}
    },
    {
      "name": "c",
      "type": "ordinal",
      "domain": {"data": "iris","field": "species"},
      "range": "category10"
    }
  ],
  "legends": [
    {
      "fill": "c",
      "title": "Species",
      "offset": 10,
      "properties": {
        "symbols": {
          "fillOpacity": {"value": 0.5},
          "stroke": {"value": "transparent"}
        }
      }
    }
  ],
  "marks": [
    {
      "name": "cell",
      "type": "group",
      "from": {"data": "fields"},
      "properties": {
        "enter": {
          "b": {"field": "data"},
          "y": {"scale": "gy","field": "data"},
          "width": {"value": 120},
          "height": {"scale": "gy","band": true,"offset": -35},
          "fill": {"value": "#fff"},
          "stroke": {"value": "#ddd"}
        }
      },
      "scales": [
        {
          "name": "x",
          "type": "linear",
          "domain": {"data": "iris","field": "petalWidth"},
          "range": "width",
          "zero": false,
          "round": true
        },
        {
          "name": "y",
          "type": "linear",
          "domain": {"data": "iris","field": {"parent": "data"}},
          "range": "height",
          "zero": false,
          "round": true
        }
      ],
      "axes": [
        {"type": "x","scale": "x","ticks": 5},
        {"type": "y","scale": "y","ticks": 5}
      ],
      "marks": [
        {
          "name": "point",
          "type": "symbol",
          "from": {"data": "iris"},
          "properties": {
            "enter": {
              "x": {
                "scale": "x",
                "field": {"datum": "petalWidth"}
              },
              "y": {
                "scale": "y",
                "field": {"datum": {"parent": "data"}}
              },
              "fill": {"scale": "c","field": "species"},
              "fillOpacity": {"value": 0.5},
              "size": {"value": 36}
            }
          }
        }
      ]
    }
  ]
}
```

The second example needs to use a cross operator.

domoritz:
Fixes #1257


src/compile/axis.ts
```diff
@@ -196,8 +196,9 @@ export function tickSize(model: Model, channel: Channel) {
 }
 
 
-export function title(model: Model, channel: Channel) {
+export function title(model: Model, channel: Channel): Field {
```
kanitw: Why can `title` be a `RepeatRef`?

src/compile/data/summary.ts
```diff
@@ -45,8 +46,11 @@ export namespace summary {
           meas['*']['count'] = true;
           /* tslint:enable:no-string-literal */
         } else {
-          meas[fieldDef.field] = meas[fieldDef.field] || {};
-          meas[fieldDef.field][fieldDef.aggregate] = true;
+          const field = fieldDef.field;
+          if (isString(field)) {
```
kanitw: `// FIXME  support iterator`

I guess we have to add another parameter as `iteratorMap` so we can replace repeatRef with current value of the iterator.

src/compile/layer.ts
```diff
@@ -233,6 +233,7 @@ export class LayerModel extends Model {
   }
 
   protected mapping() {
+    // TODO: what is this?
```
kanitw: This is null for layer, concat, repeat.   It should only exists for `facet` and `unit`.

PR #1262, opened 2016-03-24
Repeat

domoritz:
todo: clean up a bit and merge layouts if conventient. Maybe think about how to order transforms. 

SPLOM

```json
{
  "repeat": {
    "row": ["Horsepower","Miles_per_Gallon","Displacement"],
    "column": ["Displacement","Miles_per_Gallon","Horsepower"]
  },
  "spec": {
    "mark": "point",
    "encoding": {
      "x": {
        "type": "Q",
        "scale": {"zero": false}
      },
      "y": {
        "field": {"repeat": "row"},
        "type": "Q",
        "scale": {"zero": false}
      },
      "color": {"field": "Origin","type": "N"}
    }
  }
}
```

vega

Crossfilter

```json
{
  "repeat": {"column": ["hour", "delay", "distance"]},
  "spec": {
    "layers": [
      {
        "data": {"url": "data/flights-2k.json"},
        "transform": {"calculate": [{"field": "hour", "expr": "hours(datum.date)"}]},
        "mark": "bar",
        "encoding": {
          "x": {"field": {"repeat": "column"}, "type": "Q", "bin": true},
          "y": {"aggregate": "count", "field": "*", "type": "quantitative"},
          "color": {"value": "steelblue"}
        }
      },{
        "data": {"url": "data/flights-2k.json"},
        "transform": {
          "calculate": [{"field": "hour", "expr": "hours(datum.date)"}],
          "filter": "datum.delay < 5"
        },
        "mark": "bar",
        "encoding": {
          "x": {"field": {"repeat": "column"}, "type": "Q", "bin": true},
          "y": {"aggregate": "count", "field": "*", "type": "quantitative"},
        }
      }
    ]
  }
}
```

vega 1

burtin (had to hack the layout a bit)

```json
{
  "repeat": {
    "column": ["Penicilin", "Streptomycin", "Neomycin"]
  },
  "spec": {
    "data": {"url": "data/burtin.json"},
    "mark": "point",
    "encoding": {
      "x": {"field": {"repeat": "column"}, "type": "Q", "scale": {"type": "log"}},
      "y": {"field": "Bacteria", "type": "N", "axis": {"labelMaxLength": 20}},
      "color": {"field": "Gram_Staining", "type": "N"}
    }
  }
}
```

vega


.vscode/tasks.json
```diff
@@ -0,0 +1,23 @@
+{
```
kanitw: looks like it's useful to copy this to master too.

domoritz: Done. I cheery picked this commit.

.vscode/tasks.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "0.1.0",
+  "command": "npm",
+  "isShellCommand": true,
+  "args": [
+    "run"
+  ],
+  "showOutput": "silent",
+  "tasks": [
+    {
+      // "npm run build"
```
kanitw: Hmm.  JSON shouldn't have JS comment?

src/compile/model.ts
```diff
@@ -312,6 +355,19 @@ export abstract class Model {
     return this._legend[channel];
   }
 
+  public title(channel: Channel) {
```
kanitw: This method is like `field()`, which is based on the assumption that there is one field per channel, add thus will not work with multiple fields due to production rules.

Will note this in #1263.

src/compile/data/data.ts
```diff
@@ -83,8 +84,8 @@ export function parseUnitData(model: UnitModel): DataComponent {
   return {
     formatParse: formatParse.parseUnit(model),
     nullFilter: nullFilter.parseUnit(model),
-    filter: filter.parseUnit(model),
     nonPositiveFilter: nonPositiveFilter.parseUnit(model),
+    filter: filter.parseUnit(model),
```
kanitw: Are you swapping the order here?   (Just curious why?)

src/compile/data/filter.ts
```diff
@@ -41,6 +42,18 @@ export namespace filter {
     return filterComponent;
   }
 
+  export function parseRepeat(model: RepeatModel) {
+    // Note that this `filter.parseLayer` method is called before `source.parseLayer`
+    let filterComponent = parse(model);
+    model.children().forEach((child) => {
+      const childDataComponent = child.component.data;
+      if (childDataComponent.filter) {
+        delete childDataComponent.filter;
```
kanitw: Why always delete without even merging?

src/compile/model.ts
```diff
@@ -264,12 +286,24 @@ export abstract class Model {
     return this._transform || {};
   }
 
-  /** Get "field" reference for vega */
-  public field(channel: Channel, opt: FieldRefOption = {}) {
-    const fieldDef = this.fieldDef(channel);
+  /**
+   * Just the raw field. Get's the value from the iterator if the parent is iterating.
+   */
+  public fieldOrig(channel: Channel): string {
```
kanitw: looks like this method is simply  `field()` with  `opt.nofn = true`

src/compile/data/filter.ts
domoritz: Ups. Fixed

src/compile/model.ts
```diff
@@ -264,12 +286,24 @@ export abstract class Model {
...
+
+  /**
+   * Get the field reference for vega
+   */
+  public field(channel: Channel, opt: FieldRefOption = {}): string {
+    let fieldDef = this.fieldDef(channel);
 
     if (fieldDef.bin) { // bin has default suffix that depends on scaleType
-      opt = extend({
+      opt = extend({}, {
```
kanitw: why do you need to add this?

src/compile/repeat.ts
```diff
@@ -0,0 +1,295 @@
...
+  public has(channel: Channel): boolean {
+    return !!this._repeat[channel];
+  }
+
+  public children() {
+    return this._children;
+  }
+
+  public dataTable(): string {
+    return this.has(ROW) ? 'fields_row' : 'fields_column';
```
kanitw: Can you explain the rationale behind this line?

src/compile/data/filter.ts
domoritz: Fixed in 018f662

src/compile/model.ts
domoritz: Yes. I added a todo. I think we should use optional argument anyway instead of this option argument.

src/compile/model.ts
domoritz: we don't

src/compile/repeat.ts
domoritz: This returns the data table for the repeat. It's wrong now but it was right when we still had the cross. Good catch.

.vscode/tasks.json
kanitw: @domoritz

domoritz: I'll fix in on master

PR #1266, opened 2016-03-26
Hack to make all binning fields available in all sources for cross fi…

domoritz:
…lter


domoritz:
The clean solution is to figure out what the order of transforms should be and create additional data sources. But right now there is way to do this that doesn't break a bunch of assumption and thus examples.

PR #1283, opened 2016-04-05
Vega-Lite users

domoritz:


domoritz:
Once merged, I will add a link to https://vega.github.io/

site/examples/users.md
```diff
@@ -0,0 +1,21 @@
+---
```
kanitw: Users sounds a little weird... because using it in their webpages are users too.

domoritz: So?

PR #1286, opened 2016-04-06
Read bandSize config if the scale does not exist

### LINKED ISSUE #1244: Error with quantitative x and ordinal size
```json
{
  "mark": "point",
  "encoding": {
    "size": {"field": "Origin","type": "ordinal"}
  }
}
```

Safari console:

```error
[Error] TypeError: undefined is not an object (evaluating 'model.scale(xIsMeasure ? channel_1.Y : channel_1.X).bandSize')
	rangeMixins
	mainScale
	(anonymous function)
	reduce
	compileScales
	compileRootGroup
	compile
	parseVl
	(anonymous function)
	(anonymous function)
```

kanitw:
Fix #1244

Now the following no longer throws error

```js
{
  "mark": "point",
  "encoding": {
    "size": {"field": "Origin","type": "ordinal"}
  }
}
```

### DIFF
--- a/test/compile/scale.test.ts
+++ b/test/compile/scale.test.ts
@@ -389,11 +389,51 @@ describe('Scale', function() {
   });
 
   describe('rangeMixins()', function() {
-    // FIXME
+    describe('row', function() {
+      // TODO:
+    });
+
+    describe('column', function() {
+      // TODO:
+    });
+
+    describe('x', function() {
+      // TODO: X
+    });
+
+    describe('y', function() {
+      // TODO: Y
+    });
+
+    describe('size', function() {
+      describe('bar', function() {
+        // TODO:
+      });
+
+      describe('text', function() {
+        // TODO:
+      });
+
+      describe('rule', function() {
+        // TODO:
+      });
+
+      describe('point, square, circle', function() {
+        // TODO:
+      });
+    });
+
+    describe('shape', function() {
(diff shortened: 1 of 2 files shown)


domoritz:
Can you write a test for `pointBandSize`?

kanitw:
Sounds like a task for our junior colleagues :)

test/compile/scale.test.ts
```diff
@@ -389,11 +389,51 @@ describe('Scale', function() {
...
+
+    describe('column', function() {
+      // TODO:
+    });
+
+    describe('x', function() {
+      // TODO: X
+    });
+
+    describe('y', function() {
```
domoritz: I don't think this is testing the right thing. We should test what happens if there is only x and x and y.

kanitw: That's irrelevant for `rangeMixins()`. 

`rangeMixins()` of `x` and `y` scales are **independent**.

domoritz: Okay. Maybe I'm missing the point of this fix then. I thought this pr fixes the issue that we don't have a scale when we only define an x scale because a spec only defines y.

kanitw: Oh I see. That case is `size`'s scale, not `x`, `y` scale, and thus should be in line 422.

PR #1287, opened 2016-04-06
Add Anscombe Quartet Example

mcorrell:
Anscombe Quartet example


kanitw:
@mcorrell  Can you rename this to `trellis_anscombe` and add it to https://github.com/vega/vega-lite/blob/master/examples/vl-examples.json#L166 so that it shows up in the editor?

examples/specs/anscombe.json
```diff
@@ -0,0 +1,69 @@
+{
+  "description": "Anscombe's Quarter (https://en.wikipedia.org/wiki/Anscombe%27s_quartet)",
+  "data": {
+    "values": [
+      {"Series":"I", "X":10.0, "Y":8.04},
```
domoritz: I think it's worth putting this dataset in https://github.com/vega/vega-datasets

mcorrell:
Okay, will add to [vega-datasets](https://github.com/vega/vega-datasets ), and then change this example to use the url source + update name once that's done

mcorrell:
Okay, added a json of anscombe data to vega/vega-datasets#6

domoritz:
I just released a new version of vega-datasets with the data.

examples/vl-examples.json
```diff
@@ -164,6 +164,12 @@
     }
   ],
   "Trellis": [
+    { "name": "trellis_anscombe",
+      "title": "Anscombe's Quarter",
+      "galleryParameters": {
+        "backgroundSize": 130%"
```
domoritz: invalid json

mcorrell:
Should be "Quartet" rather than "Quarter." Also seems to not compile correctly in the gallery.

domoritz:
Yeah, we have to push a new version of the editor.

PR #1290, opened 2016-04-08
Axis properties

YuhanLu:
Support more properties for axis

### DIFF
--- a/site/docs/config.md
+++ b/site/docs/config.md
@@ -249,7 +249,7 @@ vg.embed('#horizontal_line', {
 | Property            | Type                | Description  |
 | :------------------ |:-------------------:| :------------|
 | font                | String  | The typeface to set the text in (e.g., `Helvetica Neue`).|
-  | fontSize            | Number  | The font size, in pixels. The default value is 10. |
+| fontSize            | Number  | The font size, in pixels. The default value is 10. |
 | fontStyle           | String  | The font style (e.g., `italic`).|
 | fontWeight          | String  | The font weight (e.g., `bold`).|
 
--- a/site/docs/config.md
+++ b/site/docs/config.md
@@ -249,7 +249,7 @@ vg.embed('#horizontal_line', {
 | Property            | Type                | Description  |
 | :------------------ |:-------------------:| :------------|
 | font                | String  | The typeface to set the text in (e.g., `Helvetica Neue`).|
-| fontSize            | Number  | The font size, in pixels. The default value is 10. |
+  | fontSize            | Number  | The font size, in pixels. The default value is 10. |
 | fontStyle           | String  | The font style (e.g., `italic`).|
 | fontWeight          | String  | The font weight (e.g., `bold`).|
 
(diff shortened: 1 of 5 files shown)


src/compile/axis.ts
```diff
@@ -195,6 +195,14 @@ export function tickSize(model: Model, channel: Channel) {
   return undefined;
 }
 
+export function tickSizeEnd(model: Model, channel: Channel) {
```
kanitw: You still haven't modify the top of this file (line 86) as suggested earlier.

src/compile/axis.ts
```diff
@@ -295,6 +326,41 @@ export namespace properties {
       }
     }
 
+    if (axis.tickLabelColor !== undefined) {
```
kanitw: Why is this method the only method that doesn't use `extend`

YuhanLu: labelsSpec is extend as well on line 280

kanitw: @YuhanLu  

Looking closer, it seems like this method here will be throw error if `labelSpec` is `undefined`  (my old code starting from line 287 has the same problem too.  

Can you go through this method and make sure if it's still correct even though the input `labelSpec` is undefined? 

Also, make sure that this method never returns empty object (`{}`).

src/compile/axis.ts
```diff
@@ -295,6 +325,41 @@ export namespace properties {
...
+    }
+
+    if (axis.tickLabelFontSize !== undefined) {
+        labelsSpec.fontSize = {value: axis.tickLabelFontSize};
+    }
+
     return labelsSpec || undefined;
   }
+
+  export function ticks(model: Model, channel: Channel, ticksPropsSpec, def) {
```
kanitw: @YuhanLu 

let's remove `def` from all of these methods if it's not used.  (If it's used like `labels()`, we can leave it in.)

YuhanLu:
did I press some wrong buttons? why is it closed?

YuhanLu:
All the properties are tested in editor

src/compile/axis.ts
YuhanLu: Done

kanitw:
LGTM.  One last thing.  Can you add documentation for these added properties?

YuhanLu:
Documentation added

site/docs/config.md
```diff
@@ -249,7 +249,7 @@ vg.embed('#horizontal_line', {
 | Property            | Type                | Description  |
 | :------------------ |:-------------------:| :------------|
 | font                | String  | The typeface to set the text in (e.g., `Helvetica Neue`).|
-| fontSize            | Number  | The font size, in pixels. The default value is 10. |
+  | fontSize            | Number  | The font size, in pixels. The default value is 10. |
```
kanitw: why are there whitespace in the front of this line?

YuhanLu: Sorry that is a typo, just fixed it.

domoritz:
🎉

PR #1291, opened 2016-04-08
Flip x and y axes for the scatter_binned example

ZeningQu:
**What**: Put Rotten Tomatoes ratings on the x axis and IMDB ratings on the y axis.

**Why**: From Jeff's class yesterday, Rotten Tomatoes ratings are more uniform while IMDB ratings are more like normal distribution. Putting Rotten Tomatoes ratings on the x axis makes it easier to examine correlations.

Before:
before

After:
after

### DIFF
--- a/examples/specs/scatter_binned.json
+++ b/examples/specs/scatter_binned.json
@@ -4,12 +4,12 @@
   "encoding": {
     "x": {
       "bin": {"maxbins": 10},
-      "field": "IMDB_Rating",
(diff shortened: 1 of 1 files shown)


kanitw:
Honestly I dont see difference here

domoritz:
Yeah, me neither. I'd even say that it was better before because large numbers were on y and thus horizontal (and legible).

PR #1294, opened 2016-04-11
Usage Instructions in docs

willium:
- [x] move usage instructions to top level
- [x] move applications under usage as 'integrations'
- [x] promote each sub-category

This hierarchy makes sense to me a little more than the current one. "Applications" don't really make sense to me under Examples, though it seems like you were originally going for "Users" which makes a bit more sense. Usage doesn't really feel like it falls under "Docs" to me either. We also wanted "Usage Instructions" on the homepage. If you add these three things all together, it's pretty clear that Usage Instructions of some kind makes sense to be at the top level (as they are with most other similar tools). This way we sort of get the best of all worlds.


_layouts/examples.html
```diff
@@ -1,12 +1,6 @@
 ---
 layout: page_submenu
```
domoritz: Why is this still submenu page and not just page?

site/usage/compile.md
```diff
@@ -0,0 +1,44 @@
+---
+layout: usage
+menu: usage
+title: Compiling Vega-Lite
```
domoritz: This is confusing. How about "compiling vl to vg"?

domoritz:
You need to fix the links to the example gallery (e.g. on the start page).

site/usage/compile.md
```diff
@@ -0,0 +1,44 @@
...
+
+```sh
+npm install vega-lite
+```
+
+Alternatively you can [download the latest Vega-Lite release](https://github.com/vega/vega-lite/releases/latest) and add it to your project manually.
+
+Then, create an HTML element that the visualization should be attached to.
+
+```html
```
domoritz: Can we distinguish the compile and embed pages a bit more distinct? Maybe the embed page should show all steps and the compile page only shows you how to generate vega and then we can link to the vega instructions. I don't think many people want to generate vega (and some might even want to do this on node). 

Let's target a different audience with this page.

willium:
Check out the new copy @domoritz. Think I fixed all the dead links too

PR #1299, opened 2016-04-13
Add tickSizeRange

### LINKED ISSUE #1298: Ticks with size channel are too large
When a `tick` mark has a size encoding, the default scaling makes them too big for the chart.

It might be nice if there was a mark config preference for size `magnitude` which would be a percentage multiple to apply to the size for `point`, `bar`, `tick`, `text`, `circle`, and `square` mark types. That way the size could be adjusted by the user after the default size scaling has been applied.

unknown

```json
{
  "mark": "tick",
  "data": {
    "values": [
    ]
  },
  "encoding": {
    "x": {"field": "b", "type": "quantitative"},
    "size": {"field": "b", "type": "quantitative"}
  }
}
```

mprudhom:
Fixes #1298


kanitw:
Thank you so much!

PR #1304, opened 2016-04-19
[Obsolete] Layering revisions

domoritz:
A clean version of layering
- Eliminate distinction between `SOURCE` and `SUMMARY` data source so we remove the weird `dataTable()` method in model.
- [x] reimplement scale merging
  how:  TODO
  why:  TODO
- [x] reimplement data merging
  how:  TODO
  why:  TODO
- [ ] reimplement layout merging
- [ ] Independent axes (support 2, then throw error)
- [ ] Clean up unnecessary data sources (can be done later)
- [ ] Some Unit Tests

Based on some changes from https://github.com/vega/vega-lite/compare/repeat3Selections3
# Ham's Review Checklist
- [ ] aggregateComponent as an array or single component -- this can be problematic with facet.
- [ ] facet component.  Discuss the following statement, which highly influence how dom structure facet code
  
  >  We only need to add the facet fields as dimensions to the summary component.  
- [ ] put `formula` and `calculate` side-by-side 
- [ ] `hash` => `key`
- [ ] parseUnit taking Model as input rather than UnitModel?
  - [ ] `data/source.ts`, `stackscale.ts`
- [ ] why some data component have `merge` and why some have `mergeIfCompatible`.  Can't we make their pattern more consistent?
## `data/aggregate.ts`
- [ ] `merge()` in `src/compile/data/aggregate.ts`  
- [ ](ham) understand why `parseFacet` is correct for `facet(field, layer(unit, unit))`
- [ ] Where does logic for renaming child data in  `parseFacet` in  `src/compile/data/summary.ts`   (now `aggregate.ts`) go?
## `data/data.ts`
- [x] `parseFacetData` very weird.  (Facet can have its own `bin`, `timeUnit`, `calculate`, `filter` but the code assume that's not true.)
  - [x](Facet should support its own `bin`, `timeUnit`, `calculate`, `filter`)
  - [ ] Why does the facet just take data from the child? weird.
- [ ] BUG? If the child data already have source, we should not override its source. Plus, we can't merge anything. Why isn't there an abort logic for source there.
- [x] weird location for `dataHash` method 
- [x] Why is time unit domain moved up in `parseLayerData()`, but not in `parseFacetData()`.
- [x] `assembleData` Probably need to create intermediate source if contains scale sort.
- [ ] why should `assembleRaw` be a part of `AggregateComponent`, not `SourceComponent`?
  - [ ] removing data source logic based on empty transform seems questionable https://github.com/vega/vega-lite/pull/1304/files#r66189288
- [ ] weird comment inside `parseFacetData()` https://github.com/vega/vega-lite/pull/1304/commits/3c6a4ecfbbb6f3252972b1d7f528c3bc2924d97e#diff-e8c47ddc5f23ed116bbdf06d4df1959fR130
## `data/formatparse.ts`
- [ ] `merge()` -- only returns the merge results, but does not return whether the merge completes.
  Don't you have to care if the something cannot be merged, then terminate the overall merge process?
  Also, unlike transform, parse is a part of data source. So if parse are not compatible, we cannot merge two data sources at all.
## `data/nonpositivenullfilter.ts`
- [ ] file name
- [ ] `mergeIfCompatible()` -- when would parent have nonPositiveFilter? This only comes from log scale and none of facet / concat / repeat / layer can have log scale.
## `data/nullfilter.ts`
- [ ] `mergeIfCompatible()`  -- why do we need `nullFilterComponent`
## `data/stackscale.ts`
- [ ] `parseFacet()` -- Would a stacked bar inside layer inside facet be still correct with this logic?
- [ ] Ham TODO: think if there is a case where stacked scale can have conflicting scale (Think `facet` of `layer`)
## `compile/layer.ts`
- `_initResolve` 
  - [ ]  not just type, but also scale signature?  
  - [ ] read explicit resolve property 
## `spec.ts`
- [ ] `ResolveStrategy` => `ResolveMode` ?
- [ ] `Resolve` => `ChannelResolve`
- [ ] `guide` => `axis` | `legend`
- [ ] `ResolveMapping` missing other channels (have only `x` and `y`)
## `model.ts`
- [ ] `rename` -- `console.error`
- [ ] searching and replacing is expensive
## `scale.ts`
- [ ] L246: why `SCALE`, not `model.dataName(SCALE)` ?
- [ ] L524 "range and other things will just be defined by the first scale"
  - This is not always true yet, given than you only determine whether to merge by type. We need to settle that first, and that will determine the correct rule here.
- [ ] L542: should not run through `colorLegend` / `binColorLegend` logic every time. 
- [ ] `normalize` -- incorrect facet logic
- [ ] would `x2`/`y2`from ranged mark cause any problem here?
## `src/vega.schema.ts`
- [ ] why do you have to make `name` optional and add TODO?
- [ ] L274:  why should model.dataName(SCALE) return something different from SCALE in this case?
## Question
- [ ] The other day, we discussed that some merge could be handled by `layer(layer(a,b), c)`  (if data of `a` and `b` should be merged).   –– but that requires supporting layer of layer (not just layer of unit)


src/compile/data/bin.ts
```diff
@@ -1,7 +1,7 @@
 import {autoMaxBins} from '../../bin';
 import {Channel, COLOR} from '../../channel';
 import {field, FieldDef} from '../../fielddef';
-import {extend, vals, flatten, hash, Dict} from '../../util';
+import {extend, vals, flatten, hash, Dict, isObject} from '../../util';
```
kanitw: `hash` => `key`?

src/compile/scale.ts
```diff
@@ -239,42 +241,43 @@ export function domain(scale: Scale, model: Model, channel:Channel): any {
   sort = domainSort(model, channel, scale.type);
 
   if (includeRawDomain) { // includeRawDomain - only Q/T
+    model.component.data.includeRawDomain = true;
     return {
-      data: SOURCE,
+      data: SCALE,
```
kanitw: why this one is not using `model.dataName`

src/compile/scale.ts
```diff
@@ -239,42 +241,43 @@ export function domain(scale: Scale, model: Model, channel:Channel): any {
...
         model.field(channel, { binSuffix: '_end' })
       ]
     };
   } else if (sort) { // have sort -- only for ordinal
     return {
-      // If sort by aggregation of a specified sort field, we need to use SOURCE table,
+      // If sort by aggregation of a specified sort field, we need to use SCALE table,
       // so we can aggregate values for the scale independently from the main aggregation.
-      data: sort.op ? SOURCE : model.dataTable(),
+      data: sort.op ? SCALE : model.dataName(SCALE),
```
kanitw: why should `model.dataName(SCALE)` return something different from `SCALE` in this case?

src/vega.schema.ts
```diff
@@ -1,8 +1,8 @@
-import {isArray} from './util';
+import {isArray, isObject} from './util';
 import {ScaleType, NiceTime} from './scale';
 
 export interface VgData {
-  name: string;
+  name?: string;  // TODO: make required
```
kanitw: why do you have to make it optional and add TODO?

src/compile/data/bin.ts
```diff
@@ -25,7 +25,7 @@ export namespace bin {
           }
         },
           // if bin is an object, load parameter here!
-          typeof bin === 'boolean' ? {} : bin
+          isObject(bin) ? bin : {}
```
kanitw: why do you need to change this line.  (FYI, below in this file, you use `typeof bin === 'boolean'`)

src/compile/model.ts
```diff
@@ -233,7 +233,11 @@ export abstract class Model {
   }
 
   public renameData(oldName: string, newName: string) {
-     this._dataNameMap.rename(oldName, newName);
+    if (oldName === newName) {
```
kanitw: why would this case happen?      Can we write code in a way that this does not happen?

src/compile/compile.ts
```diff
@@ -70,7 +70,8 @@ export function assembleRootGroup(model: Model) {
             width: {field: 'width'},
             height: {field: 'height'}
           },
-          model.assembleParentGroupProperties(model.config().cell)
+          // what is this?
```
kanitw: Adding cell config to the parent group?  
But I agree with you.  This looks doubtful to be here.   
Maybe file an issue and let me handle in a separate PR?

src/compile/data/aggregate.ts
```diff
@@ -0,0 +1,134 @@
...
+import {LayerModel} from './../layer';
+import {Model} from './../model';
+
+import {DataComponent, AggregateComponent} from './data';
+
+
+export namespace aggregate {
+  function addDimension(dims: { [field: string]: boolean }, fieldDef: FieldDef) {
+    if (fieldDef.bin) {
+      dims[field(fieldDef, { binSuffix: '_start' })] = true;
```
kanitw: `// To avoid confusion, group by all types of bin values so we always have them if we have to use`

src/compile/data/aggregate.ts
```diff
@@ -0,0 +1,134 @@
...
+    let dims: StringSet = {};
+
+    /* dictionary mapping field name => dict set of aggregation functions */
+    let meas: Dict<StringSet> = {};
+
+    model.forEach(function (fieldDef: FieldDef, channel: Channel) {
+      if (fieldDef.aggregate) {
+        if (fieldDef.aggregate === AggregateOp.COUNT) {
+          meas['*'] = meas['*'] || {};
+          /* tslint:disable:no-string-literal */
```
kanitw: Should we just do  `meas['*'].count = true`?

src/compile/data/aggregate.ts
kanitw: `// TODO: handle if bin parameters are different`

src/compile/data/aggregate.ts
```diff
@@ -0,0 +1,134 @@
...
+   * Add facet fields as dimensions.
+   */
+  export function parseFacet(model: FacetModel, aggregateComponent: AggregateComponent) {
+    aggregateComponent.dimensions = model.reduce(addDimension, aggregateComponent.dimensions);
+  }
+
+  /**
+   * Merges the aggregates from the children into the data component.
+   */
+  export function merge(dataComponent: DataComponent, childDataComponents: DataComponent[]) {
```
kanitw: Why is this function correct with facet? 

For facet, the dimension of the parent do not have to equal the child.  But all child's dimension (and source) have to be equal.  

If all children have the same dimensions, we can merge both dimensions and measures.  The new dimension = child's dimension union with the parent's dimension.  Basically, `facet` provides additional dimensions to the children.   

On the other hand, if the dimensions for children are not the same, we have to append the facet's dimensions into dimensions of each child that does not have its own data source so that data are correctly splitted across facets. 


For non-facet parents, I assume there should always be no dimensions for the parents?

src/compile/data/aggregate.ts
```diff
@@ -0,0 +1,134 @@
...
+        dataComponent.aggregate.dimensions = childDataComponents[0].aggregate.dimensions;
+      }
+      childDataComponents.forEach((data) => {
+        delete data.aggregate;
+      });
+    }
+  }
+
+  /**
+   * Assemble the aggregate. Needs a rename function because we cannot guarantee that the
```
kanitw: > "that the parent data before the children data."

Incomplete clause here.

src/compile/data/summary.ts
```diff
@@ -1,172 +0,0 @@
...
-  export function parseFacet(model: FacetModel): SummaryComponent[] {
-    const childDataComponent = model.child().component.data;
-
-    // If child doesn't have its own data source but has a summary data source, merge
-    if (!childDataComponent.source && childDataComponent.summary) {
-      let summaryComponents = childDataComponent.summary.map(function(summaryComponent) {
-        // add facet fields as dimensions
-        summaryComponent.dimensions = model.reduce(addDimension, summaryComponent.dimensions);
-
-        const summaryNameWithoutPrefix = summaryComponent.name.substr(model.child().name('').length);
```
kanitw: In the new `aggregate.ts` file, there is no longer logic for renaming data.  I guess this part is moved somewhere else.  Where does this go though?

src/compile/data/colorrank.ts
```diff
@@ -19,6 +19,7 @@ export namespace colorRank {
    * Return hash dict from a color field's name to the sort and rank transforms
    */
   export function parseUnit(model: Model) {
+    // FIXME: why is this an array?
```
kanitw: See 4 lines below this line.   It's obviously an array of `VgTransform`s.

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
+    aggregate.parseFacet(model, dataComponent.aggregate);
+  }
+  if (dataComponent.stackScale) {
+    stackScale.parseFacet(model, dataComponent.stackScale);
+  }
+
+  return dataComponent;
 }
 
+function mergeChildren(model: Model, dataComponent: DataComponent, children: UnitModel[]): DataComponent {
```
kanitw: Right now we employ all or nothing model.  I wonder if we can adjust a little bit by removing problematic `childDataComponent` from `childDataComponents`.  This way the code shouldn't be that much more complicated, but we can still reuse the logic for optimizing in many cases.  

(Maybe future work.)

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
+  if (children.length === 0) {
+    return dataComponent;
+  }
+
+  const childDataComponents = children.map((child: UnitModel) => child.component.data);
+
+  // set the parent source as child source
+  childDataComponents.forEach((data) => {
```
kanitw: `data` => `childData` to avoid confusion

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
+    if (data.source === undefined) {
+      // child does not define its own source
+      data.source = {
+        source: model.dataName(SOURCE)
+      };
+    } else {
+      data.source.source = model.dataName(SOURCE);
```
kanitw: BUG?  If the child data already have source, we should not override its source.   Plus, we can't merge anything.  Why isn't there an abort logic for source there.  

Side Note:  We really need to consider how to make things work with binding data at runtime.

src/compile/data/colorrank.ts
```diff
@@ -19,6 +19,7 @@ export namespace colorRank {
...
     let colorRankComponent: Dict<VgTransform[]> = {};
     if (model.has(COLOR) && model.fieldDef(COLOR).type === ORDINAL) {
```
kanitw: Note: Actually, if we include color rank as a part of the main transform, it's causing side effect to the data order.
I assume that this is fine because we never guarantee any particular data order.  (And if users want any specific order, 
they should use the `order` channel anyway.)  So this should be fine. 

(If we consider this problematic, this was a problem with my old code as well.)

src/compile/data/data.ts
```diff
@@ -58,17 +59,20 @@ export interface DataComponent {
   /** String set of time units that need their own data sources for scale domain */
   timeUnitDomain: StringSet;
 
-  /** Array of summary component object for producing summary (aggregate) data source */
-  summary: SummaryComponent[];
+  /** Array of summary component object for producing aggregates */
+  aggregate: AggregateComponent;
```
kanitw: Now I remember why I make my old `summary` an array of `SummaryComponent`s.

One major problem of "never" merging aggregate up like I did before is that now we have to recursively apply dimensions facet to more than just the child of current data. 

For example, if we have nesting like this.

```
facet
   layer
      unitA (aggregate - group by a)
      unitB (aggregate - group by b)
```

layer will not `aggregate` of `unitA` and `unitB`.  When you want to apply dimensions from the facet to `unitA` and `unitB`, you have to recursively apply it to deeper than one level.  (That's why original I always merge up the aggregate.) 

That said, we can keep your model if we find recursive application of facet not too terrible.  I guess you find my old array model is causing problem in other aspects too.  (In what aspects though?)

We should discuss about this in detail.

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
 
+/**
+ * Create components for data source and transforms/ aggregates.
+ */
+export function parseUnitData(model: UnitModel): DataComponent {
+  return parseData(model);
+}
+
+/**
+ * The data component for a facet is the data component of its child.
```
kanitw: I don't think this is true.  

Facet can have its own `bin`, `timeUnit`, `calculate`, `filter`.

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
+  childDataComponents.forEach((data) => {
+    if (data.source === undefined) {
+      // child does not define its own source
+      data.source = {
+        source: model.dataName(SOURCE)
+      };
+    } else {
+      data.source.source = model.dataName(SOURCE);
+    }
+    delete data.source.format;
```
kanitw: Why only removing property of `data.source` but not the entire `data.source`?

src/compile/data/aggregate.ts
```diff
@@ -0,0 +1,134 @@
...
+   */
+  export function parseFacet(model: FacetModel, aggregateComponent: AggregateComponent) {
+    aggregateComponent.dimensions = model.reduce(addDimension, aggregateComponent.dimensions);
+  }
+
+  /**
+   * Merges the aggregates from the children into the data component.
+   */
+  export function merge(dataComponent: DataComponent, childDataComponents: DataComponent[]) {
+    const dimensions = childDataComponents.reduce((collector, data) => {
```
kanitw: `data` => `childData`

kanitw:
@domoritz  Can you describe an overview why `merge` approach is better than parsing different types of construct seperately?  Does this always generalize for concat / repeat?  I'm worried that if this does not generalize, we might need to roll back a huge chunk of this PR.

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
+  // function to compare the data sources
+  function dataHash(component: DataComponent) {
+    const source = component.source;
+    if (source.values) {
+      return hash(source.values);
+    }
+    return hash(source.url + source.format);
+  }
+
+  // move time unit domain by moving it up
```
kanitw: Why is time unit domain moved up in `parseLayerData`, but not in `parseFacetData`.

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
-    timeUnit: timeUnit.parseLayer(model),
-    timeUnitDomain: timeUnitDomain.parseLayer(model),
-    summary: summary.parseLayer(model),
-    stackScale: stackScale.parseLayer(model),
-    colorRank: colorRank.parseLayer(model)
-  };
+  const dataComponent = parseData(model);
+
+  // function to compare the data sources
+  function dataHash(component: DataComponent) {
```
kanitw: Why should dataHash be inside here?  We will definitely need this for `parseConcatData` in the future PRs for sure.

src/compile/data/data.ts
```diff
@@ -149,46 +243,78 @@ export function parseLayerData(model: LayerModel): DataComponent {
...
+
+  dataSource.transform = [].concat(
+    calculate.assemble(component),
+    filter.assemble(component),
+    nullFilter.assemble(component),
+    bin.assemble(component),
+    timeUnit.assemble(component)
+  );
+
+  if (component.includeRawDomain) {
```
kanitw: Probably need to create intermediate source if contains scale `sort`.

https://github.com/vega/vega-lite/issues/1386

src/compile/data/filter.ts
```diff
@@ -12,33 +13,21 @@ export namespace filter {
...
-        childDataComponent.filter;
-      delete childDataComponent.filter;
+  /**
+   * Combines the filter in the data component if all child data components define the same filter.
+   */
+  export function mergeIfEqual(dataComponent: DataComponent, childDataComponents: DataComponent[]) {
+    const sameFilter = allSame(childDataComponents, (data) => data.filter);
+
+    if (sameFilter) {
+      // combine filter at parent
```
kanitw: Add `//TODO: compare AST` for merging?

src/compile/data/filter.ts
```diff
@@ -12,33 +13,21 @@ export namespace filter {
...
-      // merge by adding &&
-      filterComponent =
-        (filterComponent ? filterComponent + ' && ' : '') +
-        childDataComponent.filter;
-      delete childDataComponent.filter;
+  /**
+   * Combines the filter in the data component if all child data components define the same filter.
+   */
+  export function mergeIfEqual(dataComponent: DataComponent, childDataComponents: DataComponent[]) {
+    const sameFilter = allSame(childDataComponents, (childData) => childData.filter);
```
kanitw: Add `// TODO: compare AST` to be more precise?

src/compile/data/formatparse.ts
```diff
@@ -25,38 +26,45 @@ export namespace formatParse {
...
+      if (childData.formatParse) {
+        forEach(childData.formatParse, (parse, field) => {
+          if (parse === collector[field] || collector[field] === undefined) {
+            collector[field] = parse;
+            delete childData.formatParse[field];
+          }
+        });
+      }
+      return collector;
+    }, dataComponent.formatParse);
```
kanitw: This only returns the merge results, but does not return whether the merge completes.  

Don't you have to care if the something cannot be merged, then terminate the overall merge process?

src/compile/data/nonpositivenullfilter.ts
```diff
@@ -23,32 +23,24 @@ export namespace nonPositiveFilter {
     }, {} as Dict<boolean>);
   }
 
-  export function parseFacet(model: FacetModel) {
-    const childDataComponent = model.child().component.data;
+  export function mergeIfCompatible(dataComponent: DataComponent, childDataComponents: DataComponent[]) {
+    const nonPosComponent = childDataComponents.reduce((collector, childData) => {
+      extend(collector, childData.nonPositiveFilter);
+      return collector;
+    }, duplicate(dataComponent.nonPositiveFilter));
```
kanitw: when would parent have `nonPositiveFilter`?  This only comes from log scale and none of facet / concat / repeat / layer can have log scale.

src/compile/data/formatparse.ts
kanitw: Also, unlike transform, parse is a part of data source.  So if parse are not compatible, we cannot merge two data sources at all.

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
+
+  // set the parent source as child source
+  childDataComponents.forEach((childData) => {
+    if (childData.source === undefined) {
+      // child does not define its own source
+      childData.source = {
+        source: model.dataName(SOURCE)
+      };
+    } else {
+      childData.source.source = model.dataName(SOURCE);
```
kanitw: BUG?  If the child data already have source, we should not override its source.   Plus, we can't merge anything.  Why isn't there an abort logic for source there.  

Side Note:  We really need to consider how to make things work with binding data at runtime.  #1392

src/compile/data/nullfilter.ts
```diff
@@ -31,38 +31,28 @@ export namespace nullFilter {
     }, {});
   }
 
-  export const parseUnit = parse;
-
-  export function parseFacet(model: FacetModel) {
-    let nullFilterComponent = parse(model);
+  export function mergeIfCompatible(dataComponent: DataComponent, childDataComponents: DataComponent[]) {
+    const nullFilterComponent = childDataComponents.reduce((collector, childData) => {
```
kanitw: I don't understand why you have to merge this one first here.  

I think you only need to check if all of children are compatible.  If so, we can merge them into the parent.

src/compile/data/data.ts
kanitw: This comment actually doesn't make sense.  I'll remove it.

src/compile/data/stackscale.ts
```diff
@@ -13,16 +13,14 @@ import {DataComponent} from './data';
  * Stacked scale data source, for feeding the shared scale.
  */
 export namespace stackScale {
-  export function parseUnit(model: UnitModel): VgData {
+  export function parseUnit(model: Model): VgData {
```
kanitw: why parseUnit acceptModel instead of UnitModel?

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
-    timeUnitDomain: timeUnitDomain.parseFacet(model),
-    summary: summary.parseFacet(model),
-    stackScale: stackScale.parseFacet(model),
-    colorRank: colorRank.parseFacet(model)
-  };
+  // TODO: treat data in the facet and in the child differently
+
+  const child = model.child();
+  const dataComponent = child.component.data;
+  delete child.component.data;
```
kanitw: are you assuming we do not support data instead child for now?

src/compile/data/source.ts
```diff
@@ -10,17 +10,15 @@ import {DataComponent} from './data';
...
 import {filter} from './filter';
 import {bin} from './bin';
-import {formula} from './formula';
+import {calculate} from './calculate';
 import {timeUnit} from './timeunit';
+import {formatParse} from './formatparse';
 
 export namespace source {
-  function parse(model: Model): VgData {
+  export function parseUnit(model: Model): VgData {
```
kanitw: `parseUnit` taking `Model` as input rather than `UnitModel`?

src/compile/data/stackscale.ts
kanitw: `parseUnit` taking `Model` as input rather than `UnitModel`?

src/compile/data/stackscale.ts
```diff
@@ -13,16 +13,14 @@ import {DataComponent} from './data';
...
     const stackProps = model.stack();
 
     if (stackProps) {
       // produce stacked scale
       const groupbyChannel = stackProps.groupbyChannel;
       const fieldChannel = stackProps.fieldChannel;
       return {
-        name: model.dataName(STACKED_SCALE),
```
kanitw: Why are you removing rename?  Do you move it to somewhere?

src/compile/data/stackscale.ts
```diff
@@ -35,36 +33,15 @@ export namespace stackScale {
...
-
-  export function parseLayer(model: LayerModel) {
-    // TODO
-    return null;
+  /**
+   * Add facet fields as dimensions.
+   */
+  export function parseFacet(model: FacetModel, stackComponent: VgData) {
+    // Add more dimensions for row/column
+    stackComponent.transform[0].groupby = model.reduce(function (groupby, fieldDef) {
```
kanitw: Would a stacked inside layer inside facet be still correct with this logic?

src/compile/data/data.ts
```diff
@@ -79,61 +83,151 @@ export interface SummaryComponent {
...
-    timeUnit: timeUnit.parseFacet(model),
-    timeUnitDomain: timeUnitDomain.parseFacet(model),
-    summary: summary.parseFacet(model),
-    stackScale: stackScale.parseFacet(model),
-    colorRank: colorRank.parseFacet(model)
-  };
+  // TODO: treat data in the facet and in the child differently
+
+  const child = model.child();
+  const dataComponent = child.component.data;
```
kanitw: Why does the facet just take data from the child?  weird.

src/spec.ts
```diff
@@ -50,8 +50,27 @@ export interface FacetSpec extends BaseSpec {
   spec: LayerSpec | UnitSpec;
 }
 
+export enum ResolveStrategy {
```
kanitw: ResolveMode?  

Strategy generally refer to higher-level picture.  This is a very low level detail.

src/spec.ts
```diff
@@ -50,8 +50,27 @@ export interface FacetSpec extends BaseSpec {
...
+  INDEPENDENT = 'independent' as any,
+  SHARED = 'shared' as any
+}
+
+export const INDEPENDENT = ResolveStrategy.INDEPENDENT;
+export const SHARED = ResolveStrategy.SHARED;
+
+export interface Resolve {
```
kanitw: `ChannelResolve`?

src/spec.ts
```diff
@@ -50,8 +50,27 @@ export interface FacetSpec extends BaseSpec {
...
+  scale: ResolveStrategy;
+  guide?: ResolveStrategy;
+}
+
+export interface ResolveMapping {
+  x?: Resolve;
```
kanitw: why do we only have `x` and `y`?  What about other channels?

src/spec.ts
```diff
@@ -50,8 +50,27 @@ export interface FacetSpec extends BaseSpec {
...
+  INDEPENDENT = 'independent' as any,
+  SHARED = 'shared' as any
+}
+
+export const INDEPENDENT = ResolveStrategy.INDEPENDENT;
+export const SHARED = ResolveStrategy.SHARED;
+
+export interface Resolve {
+  scale: ResolveStrategy;
+  guide?: ResolveStrategy;
```
kanitw: We don't use the term `guide` anywhere else in the syntax. 

I think we should use `axis` and `legend` here.

src/compile/layer.ts
```diff
@@ -1,36 +1,69 @@
...
+        if (fd) {
+          return fd.type;
+        }
+        return undefined;
+      }).filter((type) => !!type));
+
+      if (types.length > 1) {
+        resolve[channel] = {
+          scale: INDEPENDENT,
+          guide: INDEPENDENT
```
kanitw: We don't use the term `guide` anywhere else in the syntax. 

I think we should use `axis` and `legend` here.   (depending on type of channel)

src/compile/layer.ts
```diff
@@ -1,36 +1,69 @@
...
+        }, resolve[channel]);
+      }
+    }
+
+    return resolve;
   }
 
   public has(channel: Channel): boolean {
     // layer does not have any channels
+    console.error('weird');
```
kanitw: Why do you need to output this error?   It will flood user's console.

src/compile/scale.ts
```diff
@@ -504,3 +506,78 @@ export function zero(scale: Scale, channel: Channel, fieldDef: FieldDef) {
   }
   return undefined;
 }
+
+/**
+ * Merge scales from children by unioning their domains
+ */
+export function mergeScales(model: LayerModel, channel: string) {
```
kanitw: it's so hard to diff across files.  In the future, can you avoid moving some part of the code across files in a big PR like this?  We can always add `//TODO: move to XXX.ts` and handle it later.

src/compile/layer.ts
```diff
@@ -1,32 +1,64 @@
...
 
-  private _initConfig(specConfig: Config, parent: Model) {
-    return mergeDeep(duplicate(defaultConfig), specConfig, parent ? parent.config() : {});
+  /**
+   * Figure out which fields have different field types and make those independent.
+   * Use resolve for all other fields with default shared.
+   */
+  private _initResolve(resolve: ResolveMapping): ResolveMapping {
+    CHANNELS.forEach((channel) => {
+      const types = unique(this.children().map((child: UnitModel) => {
```
kanitw: I think It's not just type.  It's scale signature.  

If two scales have different properties (e.g., different `bandWidth`) or different `clamp`, etc. 

I don't think they should be SHARED by default?  (Unless you think the right behavior should be choose one of the property, ignoring property conflicts.)

src/compile/scale.ts
```diff
@@ -504,3 +506,78 @@ export function zero(scale: Scale, channel: Channel, fieldDef: FieldDef) {
...
+
+  let domains = {};
+
+  model.children().forEach((child) => {
+    const childScales = child.component.scale[channel];
+    if (!childScales) {
+      return;
+    }
+
+    // range and other things will just be defined by the first scale
```
kanitw: This is not always true yet, given than you only determine whether to merge by `type`.   

We need to settle that first, and that will determine the correct rule here.

src/compile/scale.ts
```diff
@@ -504,3 +506,78 @@ export function zero(scale: Scale, channel: Channel, fieldDef: FieldDef) {
...
+    if (isUnionedDomain(childDomain)) {
+      // if the domain is itself a union domain, concat
+      childDomain.fields.forEach((domain) => {
+        domains[hash(domain)] = domain;
+      });
+    } else {
+      domains[hash(childDomain)] = childDomain;
+    }
+
+    // add color legend and color legend bin scales if we don't have them yet
```
kanitw: Don't you have to check if the channel is `color`?   

Why should we run through these logic for merging other channels?

src/compile/scale.ts
```diff
@@ -504,3 +506,78 @@ export function zero(scale: Scale, channel: Channel, fieldDef: FieldDef) {
...
+    if (childScales.colorLegend) {
+      scaleComp.colorLegend = scaleComp.colorLegend ? scaleComp.colorLegend : childScales.colorLegend;
```
kanitw: why not 
```
scaleComp.colorLegend = scaleComp.colorLegend || childScales.colorLegend;
```

kanitw:
Appending this to notes (along with other checklist)

## Question

- [ ] The other day, we discussed that some merge could be handled by `layer(layer(a,b), c)`  (if data of `a` and `b` should be merged).   –– but that requires supporting layer of layer (not just layer of unit)

src/compile/model.ts
```diff
@@ -41,24 +41,39 @@ export interface Component {
...
+    if (this._nameMap[newName] === oldName) {
+      console.error('Cannot rename ' + oldName + ' to ' + newName);
+      return;
+    }
+
+    if (newName in this._nameMap) {
+      // since we already have a new name for the target, we need to change it
+      newName = this._nameMap[newName];
+    } else {
+      // update existing renames
```
kanitw: The most optimized way to implement this is to update the map inside get before returning value. 

This requires searching over the map every time.

src/compile/model.ts
```diff
@@ -41,24 +41,39 @@ export interface Component {
...
+export class NameMap {
   private _nameMap: Dict<string>;
 
   constructor() {
     this._nameMap = {} as Dict<string>;
   }
 
   public rename(oldName: string, newName: string) {
+    if (this._nameMap[newName] === oldName) {
+      console.error('Cannot rename ' + oldName + ' to ' + newName);
```
kanitw: Do we really need this clause?  

In general, `console.error` doesn't really help us because we don't always see it.

src/spec.ts
```diff
@@ -105,25 +124,31 @@ export function normalize(spec: ExtendedSpec): Spec {
...
-      { data: spec.data },
-      spec.transform ? { transform: spec.transform } : {},
+      isFacet ? {} : data,
       {
         facet: extend(
           hasRow ? { row: spec.encoding.row } : {},
           hasColumn ? { column: spec.encoding.column } : {}
         ),
-        spec: {
+        spec: extend(isFacet ? data : {}, {
```
kanitw: I think the correct thing is still putting data outside because we facets the data.  
Unless you have good reason why, let's not change this.

src/spec.ts
kanitw: The rationale is that we don't have to teach that  guide collective means these two terms because practically guide also include annotations (in addition to `axis` and `legend`)

src/util.ts
```diff
@@ -17,6 +18,16 @@ export function hash(a: any) {
   return stringify(a);
 }
 
+export function empty(a: any) {
```
kanitw: Do we really need this mixed up utility?  

Looking at where you use this method, your `a` is always an object.  

Why don't you just call `keys(a).length === 0` just like everywhere else in the codebase?

test/compile/model.test.ts
```diff
@@ -0,0 +1,26 @@
+import {assert} from 'chai';
+
+import {NameMap} from '../../src/compile/model';
+
+
+describe('NameMap', function () {
+  it.only('should rename correctly', function () {
```
kanitw: it.only

src/compile/data/nonpositivenullfilter.ts
```diff
@@ -1,5 +1,5 @@
 import {ScaleType} from '../../scale';
```
kanitw: Maybe it's my own mistake, but I guess this should become `nonpositivefilter.ts`?

src/compile/data/stackscale.ts
kanitw: Ok.  It goes to `assembleData`.

src/compile/layer.ts
```diff
@@ -1,32 +1,64 @@
...
+        const fieldDef = child.fieldDef(channel);
+        if (fieldDef) {
+          return fieldDef.type;
+        }
+        return undefined;
+      }).filter((type) => !!type));
+
+      // TODO: read explicit resolve property
```
kanitw: I think reading explicit resolve property is easy.  We should just read it.

src/compile/model.ts
kanitw: Also, you already have `console.error` method inside `rename` method of the name map.  
Not sure why you have to add it twice here.

src/vega.schema.ts
```diff
@@ -50,6 +50,10 @@ export type VgScale = {
   zero?: boolean
 }
 
+export function isVgData(data: string | VgData): data is VgData {
```
kanitw: I don't understand this method..

kanitw:
why some data component have `merge` and why some have `mergeIfCompatible`.  Can't we make their pattern more consistent?

src/vega.schema.ts
domoritz: This method returns true if the variable you pass into it is a `VgData` and not a string. The nice thing is that you can use it as a type guard (hence the `data is VgData`). See how it's being used.

domoritz:
Quick status update. All specs work now except for layer inside facet where one of the layers defines a data source and the other one doesn't. I also have to check and implement layer inside layer (I'll see how difficult it is). Then the only remaining thing is a clean up pass to address @kanitw's comments.

src/compile/scale.ts
domoritz: Done

src/compile/data/bin.ts
domoritz: Done

src/compile/scale.ts
domoritz: Fixed with different implementation

src/vega.schema.ts
domoritz: Because we don't need the name until we actually output vega. The issue is that the data component != VgData.

domoritz: Fixed

src/compile/data/bin.ts
domoritz: The comment said object. I fixed the code below.

src/compile/model.ts
domoritz: These are two different cases with different error messages.

src/compile/data/aggregate.ts
domoritz: Can you add this?

src/compile/data/aggregate.ts
domoritz: Yes.

domoritz:
New status update. Things are looking good but need to implement solution outlined in #1407.

examples/specs/bar_include_raw.json
```diff
@@ -0,0 +1,13 @@
+{
+  "data": {"url": "data/cars.json"},
+  "mark": "bar",
+  "encoding": {
+    "y": {
+      "field": "Horsepower",
+      "type": "quantitative",
+      "aggregate": "mean",
+      "scale": {"includeRawDomain": true}
```
kanitw: this is now `useRawDomain` in the master

src/compile/model.ts
```diff
@@ -131,12 +131,12 @@ export abstract class Model {
     this.parseData();
     this.parseSelectionData();
     this.parseLayoutData();
-    this.parseScale(); // depends on data name
```
kanitw: why removing all these comments?

src/compile/model.ts
```diff
@@ -260,10 +274,17 @@ export abstract class Model {
    * Return the data source name for the given data source type.
    *
    * For unit spec, this is always simply the spec.name + '-' + dataSourceType.
-   * We already use the name map so that marks and scales use the correct data.
+   * The data name has not been renamed yet. Call `dataName` to get the final name.
    */
   public dataName(dataSourceType: DataTable): string {
-    return this._dataNameMap.get(this.name(String(dataSourceType)));
+    return this.name(String(dataSourceType));
```
kanitw: why `String(...)`, not `... as string`?  Our enum are `string` already so it seems unnecessary to call the function again.

src/compile/scale.ts
```diff
@@ -241,49 +252,54 @@ export function domain(scale: Scale, model: Model, channel:Channel): any {
   sort = domainSort(model, channel, scale.type);
 
   if (includeRawDomain) { // includeRawDomain - only Q/T
-    model.component.data.includeRawDomain = true;
+    // FIXME: might have already been merged up
```
kanitw: I guess we shouldn't merge with this FIXME leftover?

src/compile/data/data.ts
```diff
@@ -78,6 +72,12 @@ export interface AggregateComponent {
 
   /** dictionary mapping field name to string set of aggregate ops */
   measures: Dict<StringSet>;
+
+  /**
+   * Whether we need the unaggrgeated data source.
```
kanitw: typo -- maybe "pre-aggregated"

src/compile/data/data.ts
```diff
@@ -121,8 +121,8 @@ export function parseFacetData(model: FacetModel): DataComponent {
   const dataComponent = child.component.data;
   delete child.component.data;
 
-  [SOURCE, RAW, RANK, SCALE, STACKED_SCALE].forEach((dataSrcType) => {
-    model.renameData(child.dataName(dataSrcType), model.dataName(dataSrcType));
+  [SOURCE, RAW, STACKED_SCALE].forEach((data) => {
```
kanitw: why `data`?    `data` seems very unclear what's the variable is really for.

src/compile/data/data.ts
```diff
@@ -78,6 +72,12 @@ export interface AggregateComponent {
 
   /** dictionary mapping field name to string set of aggregate ops */
   measures: Dict<StringSet>;
+
+  /**
+   * Whether we need the unaggrgeated data source.
+   * Should be set when we parse the scales.
+   */
+  assembleRaw?: boolean;
```
kanitw: why should `assembleRaw` be a part of `AggregateComponent`, not `SourceComponent`?

src/compile/data/data.ts
```diff
@@ -316,5 +314,16 @@ export function assembleData(model: Model, data: VgData[]) {
     data.push.apply(data, tuDomain);
   }
 
+  // remove data sources that are not needed
```
kanitw: Is it possible to just not add them at all up front?  removing later is generally a bad pattern that we should avoid unless we really need to.   Plus, you are iterating over the entire `data` array that is an input parameter, which means you might remove data source that's not even generated by this `assembleData`.  This seems like a very risky pattern.  

I guess you're only adding this to remove raw?   I think even if raw doesn't have its own transform.  
There is a good reason that we might need it separately.  
For example, if source is aggregated by `O1` and `O2`.  To sort ordinal scale of `O1`, you still need the raw source even if you don't do any transform in raw. 

Even if I'm wrong about we shouldn't remove, it seems like if we move the posFilter up to before adding the raw one, we can check if data source already contains raw.

src/compile/model.ts
```diff
@@ -169,12 +183,12 @@ export abstract class Model {
     // FIXME: write assembleScales() in scale.ts that
     // help assemble scale domains with scale signature as well
     return flatten(vals(this.component.scale).map((scales: ScaleComponents) => {
-      let arr = [scales.main];
+      let arr = [renameScaleData(this, scales.main)];
```
kanitw: Maybe add comment that we need to rename because we already put these data name in the parse phase and the data source name might be renamed after that.   (Do I understand correctly?)

src/compile/model.ts
```diff
@@ -169,12 +183,12 @@ export abstract class Model {
...
       if (scales.colorLegend) {
-        arr.push(scales.colorLegend);
+        arr.push(renameScaleData(this, scales.colorLegend));
```
kanitw: looks like `renameScaleData` is only used here.  Should it be just a private method for model?

src/data.ts
```diff
@@ -23,16 +23,13 @@ export enum DataTable {
...
   LAYOUT = 'layout' as any,
-  RANK = 'rank' as any,
   SCALE = 'scale' as any,
 }
 
 export const RAW = DataTable.RAW;
 export const SOURCE = DataTable.SOURCE;
 export const STACKED_SCALE = DataTable.STACKED_SCALE;
 export const LAYOUT = DataTable.LAYOUT;
-export const RANK = DataTable.RANK;
```
kanitw: I'm quite confused.  Why does RANK disappear suddenly in this commit?

src/compile/data/data.ts
```diff
@@ -115,26 +115,42 @@ export function parseUnitData(model: UnitModel): DataComponent {
...
     model.renameData(child.dataName(data), model.dataName(data));
   });
 
-  if (dataComponent.aggregate) {
-    aggregate.parseFacet(model, dataComponent.aggregate);
-  }
-  if (dataComponent.stackScale) {
-    stackScale.parseFacet(model, dataComponent.stackScale);
-  }
+  // we should completely move the data component to the facet because
```
kanitw: not sure which part of code this comment try to address  (above? below?  or is it a TODO?)

src/compile/data/data.ts
```diff
@@ -149,46 +270,100 @@ export function parseLayerData(model: LayerModel): DataComponent {
...
     }
+  }
+
+  // time unit domain
+  const tuDomain = timeUnitDomain.assemble(component);
+  if (tuDomain) {
+    data.push.apply(data, tuDomain);
+  }
+
+  // remove data sources that are not needed
```
kanitw: Is it possible to just not add them at all up front?  removing later is generally a bad pattern that we should avoid unless we really need to.   Plus, you are iterating over the entire `data` array that is an input parameter, which means you might remove data source that's not even generated by this `assembleData`.  This seems like a very risky pattern.  

I guess you're only adding this to remove raw?   I think even if raw doesn't have its own transform.  
There is a good reason that we might need it separately.  
For example, if source is aggregated by `O1` and `O2`.  To sort ordinal scale of `O1`, you still need the raw source even if you don't do any transform in raw. 

Even if I'm wrong about we shouldn't remove, it seems like if we move the posFilter up to before adding the raw one, we can check if data source already contains raw.

src/data.ts
kanitw: NVM, it never exists in the master.

src/compile/data/data.ts
```diff
@@ -48,92 +50,211 @@ export interface DataComponent {
...
+  if (child.data() || !empty(child.transform())) {
+    model.addWarning('Facet child should not define data source or transform.');
+  }
+
+  // rename all references to child data, everything should have been merged up
+  [RAW, SOURCE, STACKED_SCALE].forEach((data) => {
+    model.renameData(child.dataName(data), model.dataName(data));
+  });
+
+  // we should completely move the data component to the facet because
```
kanitw: not sure which part of code this comment try to address (above? below? or is it a TODO?)

src/compile/data/nonpositivenullfilter.ts
```diff
@@ -14,7 +13,7 @@ export namespace nonPositiveFilter {
   export function parseUnit(model: Model): Dict<boolean> {
     return model.channels().reduce(function(nonPositiveComponent, channel) {
       const scale = model.scale(channel);
-      if (!model.field(channel) || !scale) {
+      if (!model.has(channel) || !scale) {
```
kanitw: this line is very interesting.  I don't know why it was `model.field`-- any idea?

src/compile/data/nullfilter.ts
```diff
@@ -31,6 +31,16 @@ export namespace nullFilter {
     }, {});
   }
 
+  export const parseUnit = parse;
+
+  export function parseFacet(model: FacetModel) {
+    const nullFilterComponent = parse(model);
+    const childDataComponent = model.child().component.data;
+
+    extend(nullFilterComponent, childDataComponent.nullFilter);
```
kanitw: why not removing the child's one?

src/compile/data/stackscale.ts
```diff
@@ -10,38 +10,45 @@ import {DataComponent} from './data';
 
 
 /**
- * Stacked scale data source, for feeding the shared scale.
+ * Stacked scale data transform, for feeding the shared scale.
  */
 export namespace stackScale {
-  export function parseUnit(model: Model): VgData {
+  export function parseUnit(model: Model) {
```
kanitw: why removing `: VgData`?

src/compile/data/timeunitdomain.ts
```diff
@@ -1,21 +1,21 @@
...
 
 import {FacetModel} from './../facet';
-import {LayerModel} from './../layer';
 import {Model} from './../model';
 import {parseExpression, rawDomain} from './../time';
 
 import {DataComponent} from './data';
 
-
+// should be similar to timeUnit
```
kanitw: why? See this comment https://github.com/vega/vega-lite/pull/1304/commits/3c6a4ecfbbb6f3252972b1d7f528c3bc2924d97e#r66189916

src/compile/facet.ts
```diff
@@ -184,32 +184,25 @@ export class FacetModel extends Model {
...
-              groupby: [].concat(
-                this.has(ROW) ? [this.field(ROW)] : [],
-                this.has(COLUMN) ? [this.field(COLUMN)] : []
-              )
-            }]
-          }
-        ),
-        properties: {
-          update: getFacetGroupProperties(this)
+    this.component.mark = [{
```
kanitw: I agree that you move `this.child.assembleGroup()` to `assembleMark()`.  But why does this need to become an array?

src/compile/facet.ts
```diff
@@ -275,7 +268,10 @@ export class FacetModel extends Model {
       // axisGroup is a mapping to VgMarkGroup
       vals(this.component.axisGroup),
       flatten(vals(this.component.gridGroup)),
-      this.component.mark
+      extend(
+        this.component.mark[0],
```
kanitw: Why not make mark an object [here](https://github.com/vega/vega-lite/pull/1304/commits/d96190407525bbe130ca584dcb7c6959365d3999#r66190592) and just do `this.component.mark`?

domoritz:
Fixes #1453

domoritz:
This branch is horribly outdated and I need to redo it. Probably it's best to get vl up to vega 3 first and then redo the changes from here.

domoritz:
Closing as we have redone the things in here.

Issue #947, opened 2016-01-15
Add ~~quarter, `year-quarter`,~~ century, decade to timeUnit


kanitw:
Forked from #145.  Currently we do not have support for `quarter`, `year-quarter`, `decade` and `century`.

- [x] `quarter`, `year-quarter`
- [ ] decade
- [ ] century

kanitw:
Since d3-format do not support quarter and decade directly, we probably have to add additional formula transform to calculate quarter / decade from month / year values.

kanitw:
Let's say they can always use `calculate` to make `decade` and `century`.  We have a lot other things to do.

domoritz:
👍

PR #1305, opened 2016-04-19
Adding quarter to timeUnit for #947

bobocandys:
Make axis compatible with quarter as timeUnit.


domoritz:
Looks like you are on the right track but there is still a lot missing. I suggest that you think of a few examples, write tests for those and see that you support them. For instance, the time unit can be `year`, `year-quarter`, `year-quarter-month`.

src/compile/common.ts
```diff
@@ -114,7 +117,10 @@ export function formatMixins(model: Model, channel: Channel, format: string) {
         def.format = model.config().numberFormat;
         break;
       case TEMPORAL:
-        def.format = timeFormat(model, channel) || model.config().timeFormat;
+        if ((!fieldDef.timeUnit) || (fieldDef.timeUnit &&
+              fieldDef.timeUnit.toString().indexOf('quarter') === -1)) {
```
domoritz: we have a util function for this. Also, use the constant for quarter instead of the string.

src/compile/time.ts
```diff
@@ -54,6 +60,8 @@ export function parseExpression(timeUnit: TimeUnit, fieldRef: string, onlyRef =
 
   if (timeString.indexOf('month') > -1) {
     out += get('month');
+  } else if (timeString.indexOf('quarter') > -1) {
```
domoritz: you can use the util function I mentioned here as well.

src/compile/common.ts
domoritz: the second argument should be a timeUnit so you will need to convert it to a string. Let me know if you can't find out how to convert from a ts enum value to a string.

src/compile/common.ts
```diff
@@ -103,7 +103,9 @@ export function formatMixins(model: Model, channel: Channel, format: string) {
   let def: any = {};
 
   if (fieldDef.type === TEMPORAL) {
-    def.formatType = 'time';
+    if ((!fieldDef.timeUnit) || !containsTimeUnit(fieldDef.timeUnit, TimeUnit.QUARTER)) {
```
kanitw: Can you comment why if there is no `timeUnit` or if the `timeUnit` is quarter we don't assign `formatType` here?

kanitw: I guess for quarter, we use format as a part of template?

src/compile/common.ts
```diff
@@ -103,7 +103,9 @@ export function formatMixins(model: Model, channel: Channel, format: string) {
   let def: any = {};
 
   if (fieldDef.type === TEMPORAL) {
```
kanitw: Another question is do we ever need `formatType` when `format` is not specified.

src/compile/time.ts
```diff
@@ -1,92 +1,100 @@
...
     return 'year';
   }
   return undefined;
 }
 
 export function parseExpression(timeUnit: TimeUnit, fieldRef: string, onlyRef = false): string {
   let out = 'datetime(';
-  let timeString = timeUnit.toString();
 
   function get(fun: string, addComma = true) {
```
kanitw: can we rename this to `func`?

src/compile/time.ts
```diff
@@ -1,92 +1,100 @@
...
     if (onlyRef) {
       return fieldRef + (addComma ? ', ' : '');
     } else {
-      return fun + '(' + fieldRef + ')' + (addComma ? ', ' : '');
+      let res = '';
+      if (fun === 'quarter') {
+        res = 'floor(month(' + fieldRef + ')' + '/3)*3';
```
kanitw: why do you have to multiply by 3 at the end?   (If there is a good reason, add comment.)

src/compile/common.ts
domoritz: Yes, if we have a default format. But in that case the formatType should be set automatically.

kanitw: The current code seems to output `formatType` no matter if it outputs `format` or not.

domoritz:
LGTM. @bobocandys can you merge master into this so I can do a final review and merge?

kanitw:
Test still failing

domoritz:
I hope they are failing because datalib has not been updated. It was odd that the tests passed.

kanitw:
@domoritz   FYI, It's passing now.

domoritz:
Simple test spec

```
{
  "data": {"url": "data/seattle-weather.csv", "formatType": "csv"},
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "date",
      "type": "temporal",
      "timeUnit": "quarter",
      "axis": {"title": "Month of the year"}
    },
    "y": {
      "field": "*",
      "type": "quantitative",
      "aggregate": "count"
    }
  }
}
```

domoritz:
@bobocandys There seems to be a bug with legends. See this bad example spec. 

```
{
  "description": "Stock price mean per quarter broken down by years.",
  "mark": "line",
  "encoding": {
    "x": {"field": "date", "type": "temporal", "timeUnit": "year"},
    "y": {"field": "price", "type": "quantitative", "aggregate": "mean"},
    "color": {"field": "date", "type": "temporal", "timeUnit": "quarter"}
  }
}
```

image

domoritz:
This looks wrong:

screen shot 2016-05-19 at 11 38 35

bobocandys:
It does look wrong. I will look into it later today.

domoritz:
Only need to adapt to latest datalib changes. Then we're ready to go.

kanitw:
@arvind  Do you plan to release a Vega patch sometimes soon?  (Two PR are blocked by the next Vega release.)

arvind:
There have been some features that have landed into Vega master, so I'm hesitant to cut a patch release. My goal is to release a minor version once Emily's themeing work has landed which should hopefully be within the week or two. If a patch release is urgent, I can create one by rolling back on master.

domoritz:
@bobocandys Let's wait for the vega release. Until then the test should fail in this pr.

kanitw:
Why is the test passing now?

domoritz:
I'm not sure. I thought the tests should be failing because we use an unreleased feature (quarter filters in expressions) but we can detect the error only if we parse a spec with vega (expecting `Error: Unrecognized template filter: quarter`). @bobocandys I was expecting `line_quarter.json` to cause test failures but tests pass. Can you look into why?

domoritz:
This should not be blocked any more.

src/compile/mark/text.ts
```diff
@@ -79,9 +79,14 @@ export namespace text {
 
     // text
     if (model.has(TEXT)) {
-      if (contains([QUANTITATIVE, TEMPORAL], model.fieldDef(TEXT).type)) {
+      if (QUANTITATIVE === model.fieldDef(TEXT).type) {
         const format = model.config().mark.format;
         extend(p, formatMixins(model, TEXT, format));
+      } else if (TEMPORAL === model.fieldDef(TEXT).type) {
+        p.text = {
+         // need to replace datum.data -> datum.FIELD_NAME
```
kanitw: @domoritz  this regexp looks complicated, not sure if we really need it.

src/timeunit.ts
```diff
@@ -92,5 +103,22 @@ export function format(timeUnit: TimeUnit, abbreviated = false): string {
...
+  // clean up quarter formatting remainders
+   const template = '{{datum.data | time:\'' + out.join(' ') + '\'}}';
+   return template.replace(new RegExp('{{datum.data \\| time:\'\'}}', 'g'), '');
+  } else {
+   return undefined;
+  }
+}
+
+/** Returns true if container contains the containee, false otherwise. */
+export function containsTimeUnit(container: TimeUnit, containee: TimeUnit) {
```
kanitw: should we rename `container` to `fullTimeUnit` and `containee` => `singleTimeUnit`  ?

src/timeunit.ts
```diff
@@ -92,5 +103,22 @@ export function format(timeUnit: TimeUnit, abbreviated = false): string {
...
+  let containerStr = container.toString();
+  let containeeStr = containee.toString();
+  if (containerStr.indexOf(containeeStr) > -1) {
```
kanitw: just return in one line?

src/timeunit.ts
```diff
@@ -92,5 +103,18 @@ export function format(timeUnit: TimeUnit, abbreviated = false): string {
     out.push(timeComponents.join(':'));
   }
 
-  return out.length > 0 ? out.join(' ') : undefined;
+  if (out.length > 0) {
+  // clean up quarter formatting remainders
+   const template = '{{' + field + ' | time:\'' + out.join(' ') + '\'}}';
+   return template.replace(new RegExp('{{' + field + ' \\| time:\'\'}}', 'g'), '');
```
kanitw: @domoritz   another regexp that I don't understsand ...

examples/specs/line_quarter.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "Stock price mean per quarter broken down by years.",
+  "data": {"url": "data/stocks.csv", "formatType":"csv"},
+  "mark": "point",
```
kanitw: @bobocandys  @domoritz   not sure if you want this to be line or point.   In any case, either rename the file name or the change the mark to make it consistent.  ;)

PR #1309, opened 2016-04-22
Remove example with log scale that is not meant to be a showcase

domoritz:


kanitw:
@domoritz   I actually meant for this to be a showcase, so people can easily found that they can do log scale from the gallery .

domoritz:
Then we need a more compelling example. We actually have one now ;-)

kanitw:
I agree that Gapminder one is a good example, but it introduces many things.  This example is a still a good example that just look at its name, you get right away that we can support log scale.

PR #1315, opened 2016-04-25
add opacity channel

YuhanLu:

### DIFF
--- a/examples/specs/stacked_bar_opacity.json
+++ b/examples/specs/stacked_bar_opacity.json
@@ -1,0 +1,25 @@
+{
+  "description": "A bar chart showing the US population distribution of age groups and gender in 2000.",
+  "data": { "url": "data/population.json"},
+  "transform": {
+    "filter": "datum.year == 2000",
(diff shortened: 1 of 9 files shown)


domoritz:
Can you add a test or at least an example to the test gallery that uses this feature?

kanitw:
Missing defaultScaleRange for opacity

kanitw:
- [ ] support stack with opacity (I guess stack should be `'none'` by default)

YuhanLu:
Where should I put `defaultScaleRange`? Is it better to hard code it into common.ts or somewhere else  like a `const defaultScaleRange`?

domoritz:
The default opacity range should go in https://github.com/vega/vega-lite/blob/master/src/scale.ts#L75

YuhanLu:
Documentation and example are added, ready for review.

kanitw:
Please let us know once you're done with solving conflicts.

YuhanLu:
Sorry I can't fix the conflict. Other than that this is good for review.

domoritz:
You have to make a new pr for this.

PR #1331, opened 2016-04-29
Rename `includeRawDomain` back to `useRawDomain`

### LINKED ISSUE #1330: Rename `includeRawDomain` to `useRawDomain` and set default to `true`
Because we want to see consistent scale domains across visualizations. Defaulting `useRawDomain` to `true` will give us more consistent scale domains when there is data aggregation.

ZeningQu:
Fix #1330

* Rename `includeRawDomain` to `useRawDomain`
* ~~Default `useRawDomain` to `true`~~

In the Vega-Lite test gallery, two visualizations are affected by this change:
userawdomain

The other visualizations stay the same.


kanitw:
@domoritz  any opinion about turning this on by default? 

I can't remember exactly why we make it `false` by default, but I think we discussed about it before.

@ZeningQu  In any case, you should update the docs, regarding changed default value.

site/docs/scale.md
```diff
@@ -191,25 +191,25 @@ For ordinal, quantitative, and time fields, `range` can be a two-element array d
...
+| nice          | Boolean       | If `true`, modifies the scale domain to use a more human-friendly number range (e.g., 7 instead of 6.96). <span class="note-line">__Default value:__ `true` only for quantitative x and y scales and `false` otherwise.</span> |
+| zero          | Boolean       | If `true`, ensures that a zero baseline value is included in the scale domain. <span class="note-line">__Default value:__ `true` if the quantitative field is not binned.</span> |
+| useRawDomain  | Boolean       | If `true`, set scale domain to the raw data domain. If `false`, use the aggregated data domain for scale. <span class="note-line">__Default value:__ `true`<br/>__Only valid for certain aggregations:__ This property only works with aggregate functions that produce values within the raw data domain (`"mean"`, `"average"`, `"stdev"`, `"stdevp"`, `"median"`, `"q1"`, `"q3"`, `"min"`, `"max"`). For other aggregations that produce values outside of the raw data domain (e.g. `"count"`, `"sum"`), this property is ignored. <br/>__Note:__ This property is ignored when the scale's `domain` is specified.</span>|
```
ZeningQu: I have a question about the supported aggregations. `stdev` and `stdevp` won't necessarily produce values that fall into the original data domain, right?

site/docs/scale.md
```diff
@@ -191,25 +191,25 @@ For ordinal, quantitative, and time fields, `range` can be a two-element array d
 
 | Property      | Type          | Description    |
 | :------------ |:-------------:| :------------- |
-| round         | Boolean       | If true, rounds numeric output values to integers. This can be helpful for snapping to the pixel grid (only available for `x`, `y`, `size`, `row`, and `column` scales). <span class="note-line">__Default value:__ derived from [scale config](config.html#scale-config) (`true` by default).</span> |
+| round         | Boolean       | If `true`, rounds numeric output values to integers. This can be helpful for snapping to the pixel grid (only available for `x`, `y`, `size`, `row`, and `column` scales). <span class="note-line">__Default value:__ derived from [scale config](config.html#scale-config) (`true` by default).</span> |
```
ZeningQu: I changed true to `true` because the other booleans in the doc have \` \` around them.

domoritz:
Why do we rename it back to useRawDomain? We should include the raw values in the domain but not use it entirely. I'm against changing the default to true because we released 1.0 and it is unexpected in tools other than voyager.

ZeningQu:
So the domain produced by the aggregated function can extend the the raw data domain?

kanitw:
> Why do we rename it back to useRawDomain?

We (I and @domoritz) already discussed this when we tried to solve the transform order than instead of unioning the raw domain and the domain produced by aggregate function, we should always use one of them.  Thus, the behavior we want is `useRawDomain`, not `includeRawDomain`.   So i'm surprised by your comment above.  


In fact, current implementation is `useRawDomain`, but we renamed it to `includeRawDomain` because we thought we would change the implementation, but we haven't and now we decide that the current implementation is good.   

> I'm against changing the default to true because we released 1.0 and it is unexpected in tools other than voyager.

I think a good question here is why is it unexpected in other tools?  In some sense, it might be a good default, but I don't have strong opinion here, but if you have good reason, please articulate.

domoritz:
> We (I and @domoritz) already discussed this when we tried to solve the transform order than instead of unioning the raw domain and the domain produced by aggregate function, we should always use one of them. Thus, the behavior we want is useRawDomain, not includeRawDomain. So i'm surprised by your comment above.

I remember that discussion but still think includeRawDomain should not be renamed. The behavior of unioning or using the raw domain directly (in case of aggregates like mean but not sum) should be the same so there is no reason to rename the property. 

>  think a good question here is why is it unexpected in other tools? In some sense, it might be a good default, but I don't have strong opinion here, but if you have good reason, please articulate.

I can't prove that it's unexpected without asking other people but I just cannot think of another library that has that behavior.

kanitw:
> The behavior of unioning 

Well, we discussed and agree that there should be no union between raw domain and aggregate domain for unit plot.   Simply we should pick one based on the type of aggregate function.  Therefore the behavior we want here is `useRawDomain` (if applicable), not `includeRawDomain` which implies union.  

> there is no reason to rename the property.

Given current agreement, actually it's the opposite, there is no reason to rename `useRawDomain` to `includeRawDomain`.  We renamed it back then based on the assumption that we will change the implementation to union.  SInce we no longer plan to do so, it doesn't make sense to rename it.   Note that I intentionally did not document about this property so we still have freedom to change the property name. 

>  I just cannot think of another library that has that behavior.

I don't think this is a good reason to argue for not doing it because Vega-Lite do a lot of smart defaults that other people do not do.  It's best to consider what's best from VL's perspective.

kanitw:
So @domoritz any more thoughts on this?

domoritz:
I still don't get how the observed behavior would be different and would like to keep the API stable.

site/docs/scale.md
kanitw: I think they are always within the range of `[0,   max(abs(domain_max), abs(domain_min))]` ?

src/scale.ts
```diff
@@ -144,5 +144,5 @@ export interface Scale {
   /**
    * Uses the source data range as scale domain instead of aggregated data for aggregate axis. This option does not work with sum or count aggregate as they might have a substantially larger scale range.
```
kanitw: Update to this option only works with ... <list of function or pointer to the const storing all valid aggregation function for use raw domain> ...

kanitw:
> I still don't get how the observed behavior would be different and would like to keep the API stable.

- Re: property name, our expect behavior is semantically `useRawDomain`, not `includeRawDomain`.  Since we have not documented this, I would say we better name it properly. 

- Re: the default value.   Thinking more about it, it seems like setting `useRawDomain` to `true` would always most of the charts of the same data have consistent scale, without real harm.  

So unless @domoritz has strong objection to these, maybe we could merge.

domoritz:
> Re: property name, our expect behavior is semantically useRawDomain, not includeRawDomain. Since we have not documented this, I would say we better name it properly.

Can you explain this? I still don't get how the semantics (external behavior, not implementation) is different.

kanitw:
- useRawDomain  = simply use raw values (min/max of the raw values) as the domain of the scale. 
- includeRawDomain = have two domain sources for the scale -- one is the aggregated data and one is raw.  

We want the former

domoritz:
I see. The observed behavior would still be the same (although the vega is different). If the aggregation function is sum we should definitely not make it true. 

Since we haven't documented this feature yet, I'm okay with renaming.

kanitw:
> If the aggregation function is sum we should definitely not make it true.

Currently, for sum and other aggregation function that doesn't make sense with `useRawDomain`, the property is simply ignored even if it's `true`.

domoritz:
Okay. Sounds good to me.

kanitw:
So merge?

domoritz:
Sure. I'll merge into my layer/data rewrite branch.

kanitw:
why that branch?

domoritz:
Well, cause I rewrite how the scales get data so I need to incorporate the changes from this branch. nothing to worry about for you.

src/scale.ts
domoritz: Missing before we can merge.

domoritz:
I can't merge this unless we disable `useRawDomain` for sum. Thinking more about this, it seems to be the case that charts could be unexpected if they use min aggregation with `useRawDomain` set to true. Do you still think that we should enable this option by default @kanitw?

kanitw:
I chat with Jeff the other day and seems like this shouldn't be true by default for similar reason like what you just mentioned. 

> I can't merge this unless we disable useRawDomain for sum. 

We already handled this outside of this PR.

kanitw:
@ZeningQu   Can you merge master into this,  reset default to false, and edit the PR title so we can merge?

domoritz:
So the only change is renaming? SGTM.

PR #1336, opened 2016-05-03
add new channel opacity

### LINKED ISSUE #1025: Adding Opacity as an encoding channel
Currently `opacity` is not an encoding channel.

YuhanLu:
Fix #1025 
redo #1315 
Add a new encoding channel opacity (with an example and doc).
Support stack for opacity.


YuhanLu:
It is ready for review

kanitw:
@YuhanLu  `value` property should also work too.  

```
{
  "mark": "point",
  "encoding": {
    "opacity": {"value": 0.1}
  }
}
```

YuhanLu:
Fix it. Now value works too.

kanitw:
Can you add test and example for all these cases?

- opacity.value
- opacity.field
- opacity's stack

site/docs/encoding.md
```diff
@@ -44,6 +45,7 @@ Mark properties channels map data fields directly to visual properties of the ma
 | :------------ |:-------------:| :------------- |
 | x, y          | [ChannelDef](#def)| X and Y coordinates for `point`, `circle`, `square`, `line`, `text`, and `tick`. (or to width and height for `bar` and `area` marks). |
 | color         | [ChannelDef](#def)| Color of the marks – either fill or stroke color based on mark type. (By default, fill color for `area`, `bar`, `tick`, `text`, `circle`, and `square` /   stroke color for `line` and `point`.) (See [scale range](scale.html#range) for more detail about color palettes.)  |
+| opacity         | [ChannelDef](#def)| Opacity of the marks – either can be a value or in a range. (By default, range of opacity is `[0.3, 0.8]`.)  |
```
kanitw: Inconsistent default value style.

YuhanLu:
Fix docs. Tests and examples are added.

examples/specs/stacked_bar_opacity.json
```diff
@@ -0,0 +1,25 @@
...
+    "y": {
+      "aggregate": "sum", "field": "people", "type": "quantitative",
+      "axis": {"title": "population"}
+    },
+    "opacity": {
+      "field": "gender", "type": "nominal"
+    }
+  },
+  "config": {
+    "mark": {"opacity": 0.6, "stacked" : "none"}
```
kanitw: This file is not a stacked bar.  Please correct the name.

examples/specs/stacked_bar_opacity2.json
```diff
@@ -0,0 +1,27 @@
+{
```
kanitw: rename this file to `stacked_bar_opacity` after correct the filename for `stacked_bar_opacity`.

examples/specs/stacked_bar_opacity.json
kanitw: Maybe `bar_layered_opacity`?

kanitw:
Please also update the description about to mention "fix #<issue-no>"

YuhanLu:
Renamed examples

examples/specs/stacked_bar_opacity.json
```diff
@@ -1,25 +1,27 @@
 {
-  "description": "A bar chart showing the US population distribution of age groups and gender in 2000.",
-  "data": { "url": "data/population.json"},
-  "transform": {
-    "filter": "datum.year == 2000",
-    "calculate": [{"field": "gender", "expr": "datum.sex == 2 ? \"Female\" : \"Male\""}]
-  },
+  "data": {"url": "data/seattle-weather.csv", "formatType": "csv"},
   "mark": "bar",
+  "config": {
```
kanitw: This config is unused.

examples/specs/bar_layered_opacity.json
```diff
@@ -0,0 +1,25 @@
...
+    "y": {
+      "aggregate": "sum", "field": "people", "type": "quantitative",
+      "axis": {"title": "population"}
+    },
+    "opacity": {
+      "field": "gender", "type": "nominal"
+    }
+  },
+  "config": {
+    "mark": {"opacity": 0.6, "stacked" : "none"}
```
kanitw: why do you need `mark.opacity` here since the opacity is mapped to a field.

YuhanLu: I think I will just change this example to demonstrate something else since the old one doesn't make sense.

YuhanLu:
Just fix those examples (again).

examples/specs/stacked_bar_opacity.json
```diff
@@ -0,0 +1,24 @@
...
+      "field": "date",
+      "type": "temporal",
+      "timeUnit": "month"
+    },
+    "y": {
+      "field": "*",
+      "type": "quantitative",
+      "aggregate": "count"
+    },
+    "color": {
```
kanitw: why do you need to add color in this example, seems like a distraction.

examples/specs/stacked_bar_opacity.json
```diff
@@ -0,0 +1,24 @@
...
+      "field": "weather",
+      "type": "nominal"
+    },
+    "opacity": {
```
kanitw: Also, encoding nominal field with opacity channel is a bad example, so I think we should change example here.

examples/specs/area_opacity_xfield.jason
```diff
@@ -0,0 +1,16 @@
+{
+  "data": {"url": "data/unemployment-across-industries.json"},
```
kanitw: this file is too large, taking a while to load.

examples/specs/area_opacity_xfield.jason
```diff
@@ -0,0 +1,16 @@
...
+  "encoding": {
+    "x": {
+      "timeUnit": "yearmonth", "field": "date", "type": "temporal",
+      "scale": {"nice": "month"},
+      "axis": {"axisWidth": 0, "format": "%Y", "labelAngle": 0}
+    },
+    "y": {
+      "aggregate": "sum", "field": "count","type": "quantitative"
+    },
+    "opacity": {"field": "date", "type": "temporal"}
```
kanitw: this opacity field is creating weird stacking if you look closely at the output.  

I would say that we better remove this example, since it's not an effectively designed example.

examples/specs/bar_sum_opacity.json
```diff
@@ -0,0 +1,28 @@
+{
```
kanitw: This bar is automatically stacked.  Thus, add `stacked_` prefix

YuhanLu:
Fixed

examples/specs/scatter_opacity_yfield.json
```diff
@@ -0,0 +1,9 @@
+{
```
kanitw: why do we need two of these?  

why not just one `scatter_opacity` example?

YuhanLu: I use two to show opacity with different fields (from x and y) and `scatter_opacity` is using value instead of scale.

kanitw: I see.

Let's pick one of these `scatter_opacity` and remove `scatter_opacity` because our unit test is covering this anyway.  

My rationale: the purpose of visual test is so that we can skim the gallery and check if everything is still in good shape.  If we add too many redundant examples, there would be too many things to look at anyway. 

That said, please search through all existing examples specs.  If we use `config.mark.opacity`, let's use `encoding.opacity.value` instead so it will cover the same test coverage.

YuhanLu: Duplicated examples are deleted.
I checked every existed examples with `config.mark.opacity` and changed them to `encoding.opacity.value` and they all work well.

PR #1344, opened 2016-05-10
Disable default strokeDash for legend

### LINKED ISSUE #1314: Legend should not use mark config strokeDash
The legend symbol can look funny or be completely invisible when it defaults to using the strokeDash from the mark config:

unknown

```json
{
  "mark": "line",
  "encoding": {
    "x": {"field": "date", "type": "temporal"},
  },
  "config": {
    "mark": {
      "strokeWidth": 1,
      "strokeDash": [ 24, 20 ] ,
      "strokeDashOffset": 24
    }
  }
}
```

YuhanLu:
Fix #1314 
Now legend won't take `strokeDash` and `strokeDashOffset` from config for all marks.

the following spec create the correct legend symbol now.
```
{
  "mark": "line",
  "encoding": {
    "x": {"field": "date", "type": "temporal"},
  },
  "config": {
    "mark": {
      "strokeWidth": 1,
      "strokeDash": [ 24, 20 ] ,
      "strokeDashOffset": 24
    }
  }
}
```
vega

### DIFF
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -30,7 +30,8 @@ export function buildModel(spec: Spec, parent: Model, parentGivenName: string): 
 }
 
 // TODO: figure if we really need opacity in both
-export const STROKE_CONFIG = ['stroke', 'strokeWidth', 'strokeOpacity', 'opacity'];
+export const STROKE_CONFIG = ['stroke', 'strokeWidth',
+  'strokeDash', 'strokeDashOffset', 'strokeOpacity', 'opacity'];
 
 export const FILL_CONFIG = ['fill', 'fillOpacity',
   'opacity'];
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -30,8 +30,7 @@ export function buildModel(spec: Spec, parent: Model, parentGivenName: string): 
 }
 
 // TODO: figure if we really need opacity in both
-export const STROKE_CONFIG = ['stroke', 'strokeWidth',
-  'strokeDash', 'strokeDashOffset', 'strokeOpacity', 'opacity'];
+export const STROKE_CONFIG = ['stroke', 'strokeWidth', 'strokeOpacity', 'opacity'];
 
 export const FILL_CONFIG = ['fill', 'fillOpacity',
   'opacity'];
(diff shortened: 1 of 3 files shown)


src/compile/common.ts
```diff
@@ -30,8 +30,7 @@ export function buildModel(spec: Spec, parent: Model, parentGivenName: string):
 }
 
 // TODO: figure if we really need opacity in both
-export const STROKE_CONFIG = ['stroke', 'strokeWidth',
-  'strokeDash', 'strokeDashOffset', 'strokeOpacity', 'opacity'];
+export const STROKE_CONFIG = ['stroke', 'strokeWidth', 'strokeOpacity', 'opacity'];
```
kanitw: I don't think this is the right fix.  This would make strokeDash unused for both legend and marks.  

Why don't you modify the Line 128 in legend.ts like I suggested?

YuhanLu: I did this before I got your suggestion. I will fix it.

kanitw: Oh I see.  Sounds good then ;)

YuhanLu:
Fixed it by eliminating the default strokeDash and strokeDashOffset for all marks

src/compile/legend.ts
```diff
@@ -146,9 +146,9 @@ export namespace properties {
     applyMarkConfig(symbols, model,
       channel === COLOR ?
         /* For color's legend, do not set fill (when filled) or stroke (when unfilled) property from config because the the legend's `fill` or `stroke` scale should have precedence */
-        without(FILL_STROKE_CONFIG, [ filled ? 'fill' : 'stroke']) :
+        without(FILL_STROKE_CONFIG, [ filled ? 'fill' : 'stroke', 'strokeDash', 'strokeDashOffset']) :
```
kanitw: I would organize this chunk of the code a bit so that the comments still make sense.

Ideally I would do something like:
```
let config = channel === COLOR ? 
  /* For color's legend, do not set fill (when filled) or stroke (when unfilled) property from config because the the legend's `fill` or `stroke` scale should have precedence */
  ... : 
  /* For other legend, no need to omit. */
  ...;
config = without(config, ['strokeDash', 'strokeDashOffset'])

applyMarkConfig(symbols, model, config)
```

YuhanLu:
Fixed the style.

kanitw:
Can you add test?  (and for future PRs, please add screenshot to help reviewers review PR easier.)

YuhanLu:
tests and screenshot are added

PR #1350, opened 2016-05-13
Update config.md

aishfenton:
JSON schema has strokeDashOffset as Number not array of Number (I assume the schema is current and the docs are wrong?)


domoritz:
Thank you! The docs were indeed incorrect. The schema is automatically generated from the code and thus usually correct.

PR #1362, opened 2016-05-18
Add gapminder health vs income example to galleries

ZeningQu:
Add gapminder health vs income example to vega-lite test gallery and site gallery.

test gallery look:
screen shot 2016-05-17 at 7 49 06 pm

site gallery look:
vega-lite_example_gallery

### DIFF
--- a/examples/specs/bubble_health_income.json
+++ b/examples/specs/bubble_health_income.json
@@ -1,0 +1,23 @@
+{
+  "description": "A bubble plot showing the correlation between health and income for 187 countries in the world.",
+  "data": {
+    "url": "data/gapminder-health-income.csv",
+    "formatType": "csv"
+  },
+  "mark": "circle",
+  "encoding": {
+    "y": {
+      "field": "health",
+      "type": "quantitative",
+      "scale": {"zero": false}
+    },
+    "x": {
+      "field": "income",
+      "type": "quantitative",
+      "scale": {"type": "log"}
+    },
+    "size": {"field": "population","type": "quantitative"},
+    "color": {"value": "#000"}
+  },
+  "config": {"cell": {"width": 500,"height": 300}}
+}
--- a/examples/vl-examples.json
+++ b/examples/vl-examples.json
@@ -59,6 +59,14 @@
       }
     },
     {
+      "name": "bubble_health_income",
+      "title": "Gapminder Bubble Plot",
+      "galleryParameters": {
+        "backgroundSize": "230%",
+        "backgroundPosition": "40% 20%"
+      }
+    },
+    {
       "name": "tick_strip",
       "title": "Strip Plot",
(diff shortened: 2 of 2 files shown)


examples/specs/bubble_health_income.json
```diff
@@ -0,0 +1,23 @@
+{
+  "description": "A bubble plot showing the correlation between health and income for 187 countries in the world.",
```
kanitw: Actually we copied this from that article, right?  If so, please credit the article.

domoritz: It's the default for gapminder.

kanitw: Yes, but if we copy the actual code from her, we should still credit her?

domoritz: @ZeningQu your call. The data is originally from http://www.gapminder.org/data/, right?

ZeningQu: I'm actually not sure where the original data comes from. I downloaded the dataset from this url (https://raw.githubusercontent.com/OpenNewsLabs/onechart-twelvechartinglibraries/master/Vega-Lite/data.csv) and added it to `vega-datasets` repo.

I think we should definitely credit gapminder -- perhaps by citing this url (http://www.gapminder.org/tools/bubbles#_)?

But then if we are claiming that we are recreating their graph, I guess we should make our example look more like their graph? I can do the following:

* change x from "income" to "GDP/capita in $"
* change y from "health" to "life expectancy in years"
* use color to encode continent

I may need to make some changes in `vega-datasets` too.

Should I do the above things?

kanitw: Still, we modified Lisa's code and even copied data from her copy.  Definitely doesn't hurt to credit her and her article.    In fact, it might be nice to ask her for a permission to copy.

kanitw:
Also, need to re-publish vega-editor to make sure the link is not broken.  

@ZeningQu can you also update the instruction in [this doc](https://docs.google.com/document/d/1u1T9HOTXABqusSX-n6aM8FSsXdhTnrdlaZ-pzNInNik/edit) regarding how to add an example.  (Adding dataset, publishing vega-editor, etc.)

ZeningQu:
> can you also update the instruction in [this doc](https://docs.google.com/document/d/1u1T9HOTXABqusSX-n6aM8FSsXdhTnrdlaZ-pzNInNik/edit) regarding how to add an example.  (Adding dataset, publishing vega-editor, etc.)

Done. 

I have one thing that I'm not sure... `vega-editor/app/vl-spec.js`, is this file updated automatically or manually? If I want to add an example to vega editor, do I manually edit `vega-editor/app/vl-spec.js` repo or do I only edit `vega-lite/examples/vl-examples.json`?

PR #1364, opened 2016-05-18
Update getting_started.md & demo.html to match vega-lite-demo and add missing stuff	

kanitw:


site/tutorials/gettings_started.md
```diff
@@ -315,3 +321,5 @@ Now you can create a website that embeds a Vega-Lite specification. If you want
 - See the [examples gallery]({{site.baseurl}}/examples/gallery.html).
 - Build your own visualizations in the [online editor](https://vega.github.io/vega-editor/?mode=vega-lite).
 - Browse through the [documentation]({{site.baseurl}}/docs/).
+- Learn [how to manually compile Vega-Lite specifications into Vega specifications](https://vega.github.io/vega-lite/usage/compile.html).
```
domoritz: Users very rarely have to compile vl to vg. It's a special case that we should not link to from this basic tutorial. Let's not include the link.

kanitw: But this is NOT the main part of the tutorial, this is saying what we didn't say up there.  
A lot of people want to minimize # of dependency in production, better let them aware of an option.

domoritz: I still disagree but think the rest of this pr should be merged.

kanitw: So you think people should not know about this at all?

domoritz: In the context of this tutorial, yes. It's an advanced topic that has it's place in a later tutorial.

kanitw: But that's not gonna happen anytime soon or may never happen.  This last section isn't really a part of the tutorial ...

Issue #1237, opened 2016-03-10
Cannot customize facet grid color


mprudhom:
Setting the "gridColor" property doesn't seem to have an effect.

```json
{
  "data": {"url": "data/cars.json"},
  "mark": "text",
  "encoding": {
    "row": {"field": "Origin", "type": "ordinal"},
    "column": {"field": "Cylinders", "type": "ordinal"},
    "color": {"aggregate": "mean", "field": "Horsepower", "type": "quantitative"},
    "text": {"aggregate": "count", "field": "*", "type": "quantitative"}
  },
  "config": {
    "mark": {"applyColorToBackground": true},
    "cell": {"stroke": "red", "gridColor": "red", "strokeWidth": 4},
    "axis": {"stroke": "red", "gridColor": "red", "strokeWidth": 4},
    "facet": {
      "stroke": "red", "gridColor": "red", "strokeWidth": 4,
      "cell": {"stroke": "blue", "gridColor": "red", "strokeWidth": 4},
      "grid": {"stroke": "red", "gridColor": "red", "strokeWidth": 4},
      "axis": {"stroke": "red", "gridColor": "red", "strokeWidth": 4}
    },
    "grid": {"stroke": "red", "gridColor": "red", "strokeWidth": 4}
  }
}
```

![unknown](https://cloud.githubusercontent.com/assets/659086/13657528/f2e9c750-e63d-11e5-8a2f-07cbd9f8fbbc.png)

mprudhom:
More generally, it seems like all axis grid style properties are ignored whenever there is a facet encoding, even through other properties (like labels) are still respected. For example, this non-faceted spec produces the expected output:

```json
{
  "data": {
    "values": [
      {"a": "A","b": 28,"c": "x"}, {"a": "B","b": 55,"c": "x"}, {"a": "C","b": 43,"c": "x"},
      {"a": "A","b": 28,"c": "y"}, {"a": "B","b": 55,"c": "y"}, {"a": "C","b": 43,"c": "y"}
    ]
  },
  "mark": "square",
  "encoding": {
    "x": {
      "field": "a", 
      "type": "ordinal",
      "axis": {
        "grid": true,
        "layer": "front",
        "properties": {
          "labels": {
            "fill": { "value": "red" }
          },
          "grid": {
            "stroke": { "value": "red" },
            "strokeWidth": { "value": 5 }
          }
        }
      }
    },
    "y": {
      "field": "b", 
      "type": "quantitative",
      "axis": {
[...]

PR #1367, opened 2016-05-18
Facet generates customized grid

YuhanLu:
Fix the second part of #1237.
Now the spec below generate the right output.
```
{
  "data": {
    "values": [
      {"a": "A","b": 28,"c": "x"}, {"a": "B","b": 55,"c": "x"}, {"a": "C","b": 43,"c": "x"},
      {"a": "A","b": 28,"c": "y"}, {"a": "B","b": 55,"c": "y"}, {"a": "C","b": 43,"c": "y"}
    ]
  },
  "mark": "square",
  "encoding": {
    "column": {
      "field": "c", 
      "type": "ordinal"
    },
    "x": {
      "field": "a", 
      "type": "ordinal",
      "axis": {
        "grid": true,
        "layer": "front",
        "properties": {
          "labels": {
            "fill": { "value": "red" }
          },
          "grid": {
            "stroke": { "value": "red" },
            "strokeWidth": { "value": 5 }
          }
        }
      }
    },
    "y": {
      "field": "b", 
      "type": "quantitative",
      "axis": {
        "grid": true,
        "layer": "front",
        "properties": {
          "labels": {
            "fill": { "value": "green" }
          },
          "grid": {
            "stroke": { "value": "green" },
            "strokeWidth": { "value": 5 }
          }
        }
      }
    }
  }
}
```
vega 3

### DIFF
--- a/src/compile/axis.ts
+++ b/src/compile/axis.ts
@@ -32,7 +32,7 @@ export function parseInnerAxis(channel: Channel, model: Model): VgAxis {
   // TODO: support adding ticks as well
 
   // TODO: replace any with Vega Axis Interface
-  let def = {
+  let def:any = {
     type: type,
     scale: model.scaleName(channel),
     grid: true,
@@ -58,6 +58,19 @@ export function parseInnerAxis(channel: Channel, model: Model): VgAxis {
                   axis[property];
     if (value !== undefined) {
       def[property] = value;
+    }
+  });
+
+  const props = model.axis(channel).properties || {};
+
+  // it might have more to be included here.
+  ['grid'].forEach(function(group) {
+    const value = properties[group] ?
+      properties[group](model, channel, props[group] || {}, def) :
+      props[group];
+    if (value !== undefined && keys(value).length > 0) {
+      def.properties = def.properties || {};
+      def.properties[group] = value;
     }
   });
 
--- a/src/compile/axis.ts
+++ b/src/compile/axis.ts
@@ -32,7 +32,7 @@ export function parseInnerAxis(channel: Channel, model: Model): VgAxis {
   // TODO: support adding ticks as well
 
   // TODO: replace any with Vega Axis Interface
-  let def:any = {
+  let def: any = {
     type: type,
     scale: model.scaleName(channel),
     grid: true,
@@ -63,7 +63,8 @@ export function parseInnerAxis(channel: Channel, model: Model): VgAxis {
 
   const props = model.axis(channel).properties || {};
 
-  // it might have more to be included here.
+  // For now, only need to add grid properties here because innerAxis is only for rendering grid.
+  // TODO: support add other properties for innerAxis
   ['grid'].forEach(function(group) {
     const value = properties[group] ?
       properties[group](model, channel, props[group] || {}, def) :
--- a/test/compile/axis.test.ts
+++ b/test/compile/axis.test.ts
@@ -29,6 +29,26 @@ describe('Axis', function() {
     });
   });
 
+  describe('parseInnerAxis', function() {
+    it('should produce a Vega inner axis object with correct type, scale and grid properties', function() {
+      const model = parseUnitModel({
+        mark: "point",
+        encoding: {
+          x:
+          {field: "a",
+           type: "quantitative",
+           axis: {grid: true, gridColor: "blue", gridWidth: 20}
+          }
+        }
+      });
+      const def = axis.parseInnerAxis(X, model);
+      assert.isObject(def);
+      assert.equal(def.type, 'x');
+      assert.equal(def.scale, 'x');
+      assert.deepEqual(def.properties.grid, {stroke: {value: "blue"}, strokeWidth: {value: 20}});
+    });
+  });
+
   describe('parseAxis', function() {
     it('should produce a Vega axis object with correct type and scale', function() {
       const model = parseUnitModel({


src/compile/axis.ts
```diff
@@ -61,6 +61,19 @@ export function parseInnerAxis(channel: Channel, model: Model): VgAxis {
     }
   });
 
+  const props = model.axis(channel).properties || {};
+
+  // it might have more to be included here.
+  ['grid'].forEach(function(group) {
+    const value = properties[group] ?
```
YuhanLu: I am not sure about if there's more properties needed to be added here so I leave here as a list.

YuhanLu:
The problem of this bug before the change in axis.ts, the vega-lite spec will generate this vega-spec

```
{
          "name": "cell",
          "type": "group",
          "from": {
            "data": "source",
            "transform": [{"type": "facet","groupby": ["c"]}]
          },
          "properties": {
            "update": {
              "x": {"scale": "column","field": "c","offset": 8},
              "y": {"value": 8},
              "width": {"field": {"parent": "child_width"}},
              "height": {"field": {"parent": "child_height"}},
              "stroke": {"value": "#ccc"},
              "strokeWidth": {"value": 1}
            }
          },
          "marks": [
            {
              "name": "child_marks",
              "type": "symbol",
              "properties": {
                "update": {
                  "x": {"scale": "x","field": "a"},
                  "y": {"scale": "y","field": "b"},
                  "size": {"value": 30},
                  "shape": {"value": "square"},
                  "opacity": {"value": 0.7},
                  "fill": {"value": "#4682b4"}
                }
              }
            }
          ],
          "axes": [
            {
              "type": "x",
              "scale": "x",
              "grid": true,
              "tickSize": 0,
              "properties": {
                "labels": {"text": {"value": ""}},
                "axis": {"stroke": {"value": "transparent"}}
              },
              "layer": "front",
              "ticks": 5
            },
            {
              "type": "y",
              "scale": "y",
              "grid": true,
              "tickSize": 0,
              "properties": {
                "labels": {"text": {"value": ""}},
                "axis": {"stroke": {"value": "transparent"}}
              },
              "layer": "front"
            }
          ]
        }

```
And not it will look like:
```
{
          "name": "cell",
          "type": "group",
          "from": {
            "data": "source",
            "transform": [{"type": "facet","groupby": ["c"]}]
          },
          "properties": {
            "update": {
              "x": {"scale": "column","field": "c","offset": 8},
              "y": {"value": 8},
              "width": {"field": {"parent": "child_width"}},
              "height": {"field": {"parent": "child_height"}},
              "stroke": {"value": "#ccc"},
              "strokeWidth": {"value": 1}
            }
          },
          "marks": [
            {
              "name": "child_marks",
              "type": "symbol",
              "properties": {
                "update": {
                  "x": {"scale": "x","field": "a"},
                  "y": {"scale": "y","field": "b"},
                  "size": {"value": 30},
                  "shape": {"value": "square"},
                  "opacity": {"value": 0.7},
                  "fill": {"value": "#4682b4"}
                }
              }
            }
          ],
          "axes": [
            {
              "type": "x",
              "scale": "x",
              "grid": true,
              "tickSize": 0,
              "properties": {
                "labels": {"text": {"value": ""}},
                "axis": {"stroke": {"value": "transparent"}},
                "grid": {
                  "stroke": {"value": "red"},
                  "strokeWidth": {"value": 5}
                }
              },
              "layer": "front",
              "ticks": 5
            },
            {
              "type": "y",
              "scale": "y",
              "grid": true,
              "tickSize": 0,
              "properties": {
                "labels": {"text": {"value": ""}},
                "axis": {"stroke": {"value": "transparent"}},
                "grid": {
                  "stroke": {"value": "green"},
                  "strokeWidth": {"value": 5}
                }
              },
              "layer": "front"
            }
          ]
        }
```

The difference is from `child.axis`, which generated by `axis.ts` `parseInnerAxis()`. `parseInnerAxis()` didn't update `child.axis.properties.grid` and not I add a parse for `properties.grid` so the under vega spec mark `cell`, it will include `grid`.

src/compile/axis.ts
kanitw: Good question.   I would examine properties of `'axis', 'labels', 'title', 'ticks', 'majorTicks', 'minorTicks'`  and see if they have similar problem.

YuhanLu: I tried and grid is only the special rule here. Should I remove the list and only leave grid here or keep it as a list?

YuhanLu:
Hmm actually it works for the old version of axis properties but not the new version. It cannot parse `axisColor: green` to be `axis: {stroke: {value: green}}`. I will fix it.

YuhanLu:
It was my mistake and it works for new version of axis properties, too.

The spec below using the new axis properties will generate the same output as above.
```
{
  "data": {
    "values": [
      {"a": "A","b": 28,"c": "x"}, {"a": "B","b": 55,"c": "x"}, {"a": "C","b": 43,"c": "x"},
      {"a": "A","b": 28,"c": "y"}, {"a": "B","b": 55,"c": "y"}, {"a": "C","b": 43,"c": "y"}
    ]
  },
  "mark": "square",
  "encoding": {
    "column": {
      "field": "c", 
      "type": "ordinal"
    },
    "x": {
      "field": "a", 
      "type": "ordinal",
      "axis": {
        "grid": true,
        "layer": "front",
        "gridColor": "red",
        "gridWidth": 5,
        "properties": {
          "labels": {
            "fill": { "value": "red" }
          }
        }
      }
    },
    "y": {
      "field": "b", 
      "type": "quantitative",
      "axis": {
        "grid": true,
        "layer": "front",
        "gridColor": "green",
        "gridWidth": 5,
        "properties": {
          "labels": {
            "fill": { "value": "green" }
          }
        }
      }
    }
  }
}
```

kanitw:
Haven't got a chance to look deeply, but could you add unit test first?

YuhanLu:
Unit test is added.

src/compile/axis.ts
kanitw: Let's leave it as is.   Currently innerAxis is only for grid, but we can support others in the future.  
(That's beyond scope of this PR.)

Issue #1372, opened 2016-05-19
1.0.9 npm package broke webpack build


domoritz:
Not sure what't going wrong yet. Have the contents of the npm package changed?

```
ERROR in ./~/vega-lite/src/compile/layer.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./data/data in /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile
resolve file
  /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data.webpack.js doesn't exist
  /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data doesn't exist
  /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data.web.js doesn't exist
  /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data.js doesn't exist
  /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data.json doesn't exist
resolve directory
  /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data doesn't exist (directory default file)
  /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data/package.json doesn't exist (directory description file)
[/Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data.webpack.js]
[/Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data]
[/Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data.web.js]
[/Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data.js]
[/Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data.json]
 @ ./~/vega-lite/src/compile/layer.js 9:13-35

ERROR in ./~/vega-lite/src/compile/unit.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./data/data in /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile
resolve file
  /Users/domoritz/Developer/UW/ipython-vega/node_modules/vega-lite/src/compile/data/data.webpack.js doesn't exist
[...]

PR #1373, opened 2016-05-19
Remove unused browserify-shim

kanitw:


jheer:
Just be careful that this does not break the build process for other tools
that package up vega-lite in a browserify project.

domoritz:
Such as the ipython vega project.

kanitw:
@domoritz   Do you mean this PR break ipython vega project?  

The motivation for this PR is actually making my CompassQL buildable.   (It's complaining about Vega-Lite's browserify-shim, which we actually do not use.

domoritz:
I have to try. Unfortunately the build is already broken as documented in https://github.com/vega/vega-lite/issues/1372\.

domoritz:
I don't have build problems in ipyvega, which uses webpack or vl-shorthand (on the new branch), which uses rollup. The only browserify project we tested with is compassql. 

@arvind Let us know if this causes build issues in lyra.

PR #1375, opened 2016-05-19
Legend supports opacity channel

### LINKED ISSUE #1349: Legend should support opacity
Legend should support the new channel opacity. Now the spec cannot generate a output with desired legend.

```
{
  "data": {"url": "data/seattle-weather.csv", "formatType": "csv"},
  "mark": "bar",
  "config": {
    "mark": { "opacity": 0.5 }
  },
  "encoding": {
    "x": {
      "field": "date",
      "type": "temporal",
      "timeUnit": "month"
    },
    "y": {
      "field": "*",
      "type": "quantitative",
      "aggregate": "count"
    },
    "opacity": {
      "field": "*",
      "type": "quantitative",
      "aggregate": "count"
    }
  }
}
```
vega 1

YuhanLu:
Fix #1349 
Now legend supports to show opacity channel

```
{
  "data": {"url": "data/seattle-weather.csv", "formatType": "csv"},
  "mark": "bar",
  "config": {
    "mark": { "opacity": 0.5 }
  },
  "encoding": {
    "x": {
      "field": "date",
      "type": "temporal",
      "timeUnit": "month"
    },
    "y": {
      "field": "*",
      "type": "quantitative",
      "aggregate": "count"
    },
    "opacity": {
      "field": "*",
      "type": "quantitative",
      "aggregate": "count"
    }
  }
}
```


src/compile/legend.ts
```diff
@@ -158,6 +160,10 @@ export namespace properties {
       symbols.strokeWidth = { value: 0 };
     }
 
+    if (channel === OPACITY) {
+      delete symbols.opacity;
```
kanitw: // Avoid override default mapping for opacity channel

YuhanLu: Do you have any suggestion how should we avoid this? Since `symbols` are initialized in `applyMarkConfig`, I think it's better to do any modification in `legend.ts` instead of other places.

kanitw: I think this is fine.  I mean add a comment
```
// Avoid override default mapping for opacity channel
```

Sorry for not being clear.

kanitw:
Still missing test  and I guess we need to wait for a new Vega release, which is coming soon?

PR #1382, opened 2016-05-22
Ranged Mark

ayush29feb:
- Add ranged `rule`, `bar` mark and `errorbar` composition mark
- normalize `x2` to `x` if there is no `x` (similar for `y`'s)
- example box plot, errorbar 
- remove unnecessary `isAbbreviated`


src/channel.ts
```diff
@@ -74,6 +78,8 @@ export function getSupportedMark(channel: Channel): SupportedMark {
   switch (channel) {
     case X:
     case Y:
+    case X2:
```
kanitw: `X2`, `Y2`  need their own group.   `point`, `tick`, `circle`, `square`, `line`, `text` should not support x2, y2.

src/channel.ts
```diff
@@ -113,6 +119,8 @@ export function getSupportedRole(channel: Channel): SupportedRole {
...
```
kanitw: I'm not 100% sure but I think X2 Y2 should not support `dimension`.  

(Can you think of a case where you want to map group-by variable of aggregate plot to X2/Y2?)

src/compile/mark/rule.ts
```diff
@@ -13,23 +13,29 @@ export namespace rule {
 
     // TODO: support explicit value
 
-    // vertical
-    if (model.has(X)) {
+    if (model.has(X) && model.has(Y) && model.has(Y2)) { // vertical range
```
kanitw: Right now, there are four cases.    Two of them vertical, two of them horizontal.  

Instead of breaking down like this,  can you:

1) Add rule for initialize `orient` for `rule` mark [here](https://github.com/vega/vega-lite/blob/master/src/compile/config.ts#L31). 

2) Have top-level `if` determine the `orient`. Then split case within each type of orientation.    
   - If horizontal, see if `x` and `x2` are specified.  Otherwise, use the default.  (If only `x` is specified, set `x2` to 0.) 
   - Similar for `y` except that `y2`'s default should be `{group: 'height'}`.

domoritz:
Sweet. The screenshot looks great.

kanitw:
Looking at the example spec,  I think this line is incorrect.

```
        "scale": {"bandSize": 10}
```

In fact, I'm surprised that it renders correctly.  It's probably a bug in our scale merging logic.  

@domoritz

domoritz:
I rewrote the scale merge logic in my coming layer pr.

src/compile/config.ts
```diff
@@ -39,12 +39,22 @@ export function initMarkConfig(mark: Mark, encoding: Encoding, config: Config) {
              } else {
                cfg[property] = 'horizontal'; // implicitly horizontal
              }
+             if (contains([RULE, BAR, AREA], mark)) {
```
kanitw: This if-clause is really weird because it can override what's set right above.

src/compile/config.ts
```diff
@@ -39,12 +39,22 @@ export function initMarkConfig(mark: Mark, encoding: Encoding, config: Config) {
...
+             } else {
+               cfg[property] = 'vertical';
+             }
            } else if (!xIsMeasure && yIsMeasure) {
              if (mark === TICK) {
                cfg[property] = 'horizontal';
              } else {
                cfg[property] = 'vertical';
              }
+             if (contains([RULE, BAR, AREA], mark)) {
```
kanitw: This if-clause is really weird because it can override what's set right above.

ayush29feb: Well, in this specific case, I don't think it will. But I see your point.

src/compile/mark/rule.ts
```diff
@@ -12,25 +12,34 @@ export namespace rule {
...
 
-    // vertical
-    if (model.has(X)) {
-      p.x = position(model, X);
-
-      p.y = { value: 0 };
-      p.y2 = {
-          field: {group: 'height'}
+    if (orient === 'vertical') {
+      if (model.has(X) && model.has(Y) && model.has(Y2)) {
```
kanitw: Why don't you split x and y inside this case?  

- You haven't handle the case 
  - where there is y2, but no y.  set `y` to `{group: 'height'}`)
  - both `y` and `y2` are not specified
  - where there is no x.  (Returns `{value: 0}`)

src/compile/mark/rule.ts
```diff
@@ -12,25 +12,34 @@ export namespace rule {
...
-      p.x = { value: 0 };
-      p.x2 = {
-          field: {group: 'width'}
+      } else if (model.has(X) && model.has(Y)) {
+          p.x = position(model, X);
+          p.y = { field: { group: 'height' } };
+          p.y2 = position(model, Y);
+      }
+    } else if (orient === 'horizontal') {
+      if (model.has(Y) && model.has(X) && model.has(X2)) {
```
kanitw: similiar to vertical case.

src/compile/config.ts
```diff
@@ -34,16 +34,23 @@ export function initMarkConfig(mark: Mark, encoding: Encoding, config: Config) {
...
            if (xIsMeasure && !yIsMeasure) {
-             if (mark === TICK) {
-               cfg[property] = 'vertical'; // implicitly vertical
+             if (contains([RULE, BAR, AREA], mark)) {
+               cfg[property] = 'horizontal';
+             } else if (mark === TICK) {
+               cfg[property] = 'vertical';
              } else {
-               cfg[property] = 'horizontal'; // implicitly horizontal
+               cfg[property] = 'vertical';  // implicitly vertical
```
kanitw: what's the rationale for changing the default to vertical?

src/compile/config.ts
```diff
@@ -34,16 +34,23 @@ export function initMarkConfig(mark: Mark, encoding: Encoding, config: Config) {
...
+             {
+               cfg[property] = 'vertical';
+             }
+             if (contains([RULE, BAR, AREA], mark)) {
+               cfg[property] = 'vertical';
+             } else if (mark === TICK) {
                cfg[property] = 'horizontal';
              } else {
-               cfg[property] = 'vertical';
+               cfg[property] = 'horizontal';  // implicitly horizontal
```
kanitw: what's the rationale for changing the default to horizontal?

src/compile/config.ts
```diff
@@ -34,16 +34,23 @@ export function initMarkConfig(mark: Mark, encoding: Encoding, config: Config) {
...
+               cfg[property] = 'horizontal';
+             } else if (mark === TICK) {
+               cfg[property] = 'vertical';
              } else {
-               cfg[property] = 'horizontal'; // implicitly horizontal
+               cfg[property] = 'vertical';  // implicitly vertical
              }
            } else if (!xIsMeasure && yIsMeasure) {
-             if (mark === TICK) {
+             {
```
kanitw: why `{`, `}`  here?

kanitw:
@ayush29feb   You will also have to modify `scale.ts` so that it takes both `y` and `y2` as part of the `y` scale.

kanitw:
re:

>         "scale": {"bandSize": 10}

the correct code should be using

> "size": {"value": 10}

ayush29feb:
> modify scale.ts so that it takes both y and y2 as part of the y scale.

Yeah, I just realized it.

> the correct code should be using

Okay, I think one of the examples was using that so I used it.

kanitw:
> Okay, I think one of the examples was using that so I used it.

Yes, but the example really modify the scale bandSize and thus the tick size is automatically adjusted.  
Here you want the same bandSize (but different tickSize) :)

kanitw:
@ayush29feb  I'm appending this to the description on the top

Ham's additional comment:
- [ ]  modify `scale.ts` so that it takes both `y` and `y2` as part of the `y` scale.    
- [ ] make sure there is only x-axis (for `x` and/or `x2`) -- same for y
- [ ] how should `scale` and `axis` properties in `x2` be taken care off.

kanitw:
@domoritz   I think we gonna not allow `scale` and `axis` inside `x2` and `y2` (because that's the same scale as `x` and `y`).  

Any objection?

domoritz:
> @domoritz I think we gonna not allow scale and axis inside x2 and y2 (because that's the same scale as x and y).

In that case I wonder whether whether we should just have `field1`, `field2` instead of two channels.

kanitw:
> In that case I wonder whether whether we should just have `field1`, `field2` instead of two channels.

That doesn't make sense because it's a fieldDef that can contain `aggregate`, `timeUnit`, `bin`.  
We have keep it as channel to maintain consistency (and make intermediate transition state in the UI reasonable.  For example, user would drag one field to `x` first, then `x2`).

ayush29feb:
@domoritz I don't know if you are thinking similar to what I was thinking. But I just thought about instead of doing `X` and `X2` we can do `X` as a cannel and inside of it we can have `x1` and `x2` which would basically be start and end props. I think thats what you meant by `field1` and `field2`.

ayush29feb:
@kanitw I do see your point that it wouldn't make complete sense. In that case we should maybe have some sort of overloading. Basically, both `X`/`Y` and `X2`/`Y2` can have `scale` or `axis` but `X` will override the props of `X2` if it exists otherwise `X` props will be used.

kanitw:
>  But I just thought about instead of doing X and X2 we can do X as a cannel and inside of it we can have x1 and x2 which would basically be start and end props.

If you think about the UI and drag and drop, this doesn't offer reasonable intermediate representation for UI.

> Basically, both X/Y and X2/Y2 can have scale or axis but X will override the props of X2 if it exists otherwise X props will be used.

In slack, we agree to do override and give  `x` stronger precedence than `x2`.
This means if someone stupidly set `x.scale.type='linear'`, `x2.scale.type='log'`,


I think the right solution for all these confusion is to separate `scale`, `axis`, `legend` mapping from encoding mapping because they do not always have 1:1 correspondence.  (scale/axis/legend has only `x` but encoding can have `x` and `x2`)

By this I mean,

```
...
encoding: {
  x: { ..., scale: XSCALE }
}
...
```

should be

```
...
encoding: {
  x: { ... }
},
scale: {
  x: XSCALE
}
...
```

so that even if one adds `x2` or make `x` become a production rule, the scale remains sensible.  

This will also solve problem when we have production rule (but different fieldDef in the production rule should share the same) 

But let's think about it more and we can handle that in the future.

src/compile/layout.ts
```diff
@@ -86,7 +86,7 @@ function parseUnitSizeLayout(model: UnitModel, channel: Channel): SizeComponent
 }
 
 function unitSizeExpr(model: UnitModel, channel: Channel, nonOrdinalSize: number): string {
-  if (model.has(channel)) {
+  if (model.has(channel) || (channel === X && model.has(X2)) || (channel === Y && model.has(Y2))) {
```
kanitw: If I understand correctly, even if `model.has(X)` is `false`, we would have initialized scale `x`.  

Thus, can you try if replacing the predicate inside `if` with  `model.scale(channel) !== null` working?  

(If it works, we can eliminate `model.has()` as well. )

ayush29feb: Yeah that makes sense. I tried it, its working. I will push that in the next commit

src/compile/config.ts
```diff
@@ -31,19 +31,21 @@ export function initMarkConfig(mark: Mark, encoding: Encoding, config: Config) {
          case 'orient':
            const xIsMeasure = isMeasure(encoding.x);
            const yIsMeasure = isMeasure(encoding.y);
+           const x2IsMeasure = isMeasure(encoding.x2);
```
kanitw: Maybe this is cleaner: 

```
const xIsMeasure = isMeasure(encoding.x) || isMeasure(encoding.x2);
```
So you don't need `x2IsMeasure`.

(same for y)

src/compile/mark/rule.ts
```diff
@@ -12,25 +12,48 @@ export namespace rule {
...
+        p.x = position(model, X);
+      } else {
+        p.x = { value: 0 };
+      }
 
-      p.x = { value: 0 };
-      p.x2 = {
-          field: {group: 'width'}
+      if (model.has(X2)) {
+        p.x2 = {
```
kanitw: Why does this one doesn't use the `position` method?  

(I actually prefer to get rid of the position method because it's not saving that many lines but requires me to jump back and forth to read the code.  That said, even if we keep it, it should be consistent?)

src/compile/mark/rule.ts
```diff
@@ -12,25 +12,48 @@ export namespace rule {
...
+        p.y = position(model, Y);
+      } else {
+        p.y = { field: { group: 'height' } };
+      }
 
-      p.y = { value: 0 };
-      p.y2 = {
-          field: {group: 'height'}
+      if (model.has(Y2)) {
+        p.y2 = {
```
kanitw: Why does this one doesn't use the `position` method?  

(I actually prefer to get rid of the position method because it's not saving that many lines but requires me to jump back and forth to read the code.  That said, even if we keep it, it should be consistent?)

src/compile/axis.ts
```diff
@@ -92,8 +92,14 @@ export function parseAxis(channel: Channel, model: Model): VgAxis {
   };
 
   // format mixins (add format and formatType)
-  extend(def, formatMixins(model, channel, model.axis(channel).format));
-
+  // In case we need to get the fielddef axis for X2 or Y2
```
kanitw: We can eliminate this.  See b3180305c79588a04717349998210921102353f6

src/compile/mark/rule.ts
ayush29feb: At first I was using different scales, I think I forgot to fix that. Although, I agree to get rid of `position()` because of the reasons you mentioned and its more clear with just specifying it there.

src/compile/scale.ts
```diff
@@ -68,15 +69,29 @@ export function parseScaleComponent(model: Model): Dict<ScaleComponents> {
...
     name: model.scaleName(channel),
     type: scale.type,
   };
 
-  scaleDef.domain = domain(scale, model, channel);
-  extend(scaleDef, rangeMixins(scale, model, channel));
+  // If channel is either X or Y then union them with X2 & Y2 if they exist 
+  if (channel === X && model.has(X2)) {
+    if (model.has(X)) {
+      scaleDef.domain = { fields : [domain(scale, model, X), domain(scale, model, X2)] };
```
kanitw: `field :` => `field:`

src/compile/scale.ts
```diff
@@ -68,15 +69,29 @@ export function parseScaleComponent(model: Model): Dict<ScaleComponents> {
...
+    } else {
+      scaleDef.domain = domain(scale, model, X2);
+    }
+  } else if (channel === Y && model.has(Y2)) {
+      if (model.has(Y)) {
+        scaleDef.domain = { fields : [domain(scale, model, Y), domain(scale, model, Y2)] };
```
kanitw: `field :` => `field:`

src/compile/scale.ts
```diff
@@ -68,15 +69,29 @@ export function parseScaleComponent(model: Model): Dict<ScaleComponents> {
...
-
   let scaleDef: any = {
     name: model.scaleName(channel),
     type: scale.type,
   };
 
-  scaleDef.domain = domain(scale, model, channel);
-  extend(scaleDef, rangeMixins(scale, model, channel));
+  // If channel is either X or Y then union them with X2 & Y2 if they exist 
+  if (channel === X && model.has(X2)) {
```
kanitw: I think the logic here is correct.  That said, I want to cc: @domoritz  as this might have implication for how we merge scale later down the stream.

src/compile/unit.ts
```diff
@@ -83,11 +83,18 @@ export class UnitModel extends Model {
...
+
+        let targetChannel = channel;
+        if (!vlEncoding.has(encoding, channel)) {
+          targetChannel = channel === X ? X2 : Y2;
+        }
+        const channelDef = encoding[targetChannel];
+        const scaleSpec = channelDef.scale || {};
 
-        const _scaleType = scaleType(scaleSpec, channelDef, channel, mark);
+        const _scaleType = scaleType(scaleSpec, channelDef, targetChannel, mark);
```
kanitw: If you read the logic inside `scaleType`, I don't think this should be `targetChannel` here.

kanitw: Another part is that tricky part is here `channelDef` of `x` and `x2`  should have the same `type`, `bin`, `timeUnit`.   

Maybe we have to generate warning if they do not.

src/compile/unit.ts
```diff
@@ -83,11 +83,18 @@ export class UnitModel extends Model {
...
+      if (vlEncoding.has(encoding, channel) ||
+          (channel === X && vlEncoding.has(encoding, X2)) ||
+          (channel === Y && vlEncoding.has(encoding, Y2))) {
+
+        let targetChannel = channel;
+        if (!vlEncoding.has(encoding, channel)) {
+          targetChannel = channel === X ? X2 : Y2;
+        }
+        const channelDef = encoding[targetChannel];
+        const scaleSpec = channelDef.scale || {};
```
kanitw: We haven't really merge the properties between `x` and `x2` if both exist yet.  
I would say just add `//TODO` for now and create a follow-up issue.  
No need to implement it yet.

src/compile/unit.ts
```diff
@@ -105,8 +112,15 @@ export class UnitModel extends Model {
...
-        const axisSpec = encoding[channel].axis;
+      if (vlEncoding.has(encoding, channel) ||
+          (channel === X && vlEncoding.has(encoding, X2)) ||
+          (channel === Y && vlEncoding.has(encoding, Y2))) {
+
+        let targetChannel = channel;
+        if (!vlEncoding.has(encoding, channel)) {
+          targetChannel = channel === X ? X2 : Y2;
+        }
+        const axisSpec = encoding[targetChannel].axis;
```
kanitw: We haven't really merge the axis properties between `x` and `x2` if both exist yet.  
I would say just add `//TODO` for now and create a follow-up issue.  
No need to implement it yet.

ayush29feb: Yeah, I know. Correct me if I am wrong but I think this essentially defaults the axis props to be `x` if `x` exist and takes `x2` if `x` doesn't exist

src/compile/mark/rule.ts
ayush29feb: in `position()` it sets `field` to be `model.field(channel, { binSuffix: '_mid' })`. Do we need the `binSuffix`?

src/compile/unit.ts
ayush29feb: > If you read the logic inside scaleType, I don't think this should be targetChannel here.

Ah you are right.

> Maybe we have to generate warning if they do not.
Yeah I think that would be a good idea. Although, it would just ignore the `x2`'s `type`, `bin`, `timeUnit`. That should make the spec work, but yeah a warning would be nice

src/compile/axis.ts
ayush29feb: not sure why would we eliminate it? we modified it in the refactoring.

src/compile/unit.ts
kanitw: This is correct.

src/compile/common.ts
```diff
@@ -130,65 +127,14 @@ export function formatMixins(model: Model, channel: Channel, format: string) {
         def.format = model.config().numberFormat;
         break;
       case TEMPORAL:
-        def.format = timeFormat(model, channel) || model.config().timeFormat;
+        def.format = timeFormatExpr(fieldDef.timeUnit, shortTimeLabels) || model.config().timeFormat;
```
kanitw: `timeFormatExpr` => `timeFormat`?

src/compile/layout.ts
```diff
@@ -86,7 +86,7 @@ function parseUnitSizeLayout(model: UnitModel, channel: Channel): SizeComponent
 }
 
 function unitSizeExpr(model: UnitModel, channel: Channel, nonOrdinalSize: number): string {
-  if (model.has(channel) || (channel === X && model.has(X2)) || (channel === Y && model.has(Y2))) {
+  if (model.scale(channel) !== null || (channel === X && model.has(X2)) || (channel === Y && model.has(Y2))) {
```
kanitw: Wait, I mean I think you don't even need 
`|| (channel === X && model.has(X2)) || (channel === Y && model.has(Y2))`  here.

src/compile/legend.ts
```diff
@@ -226,7 +229,7 @@ export namespace properties {
       } else if (fieldDef.timeUnit) {
         labelsSpec = extend({
           text: {
-            template: '{{ datum.data | time:\'' + timeFormat(model, channel) + '\'}}'
+            template: '{{ datum.data | time:\'' + timeFormat(fieldDef.timeUnit, model.legend(channel).shortTimeLabels) + '\'}}'
```
kanitw: why not `legend.shortTimeLabels`?

src/compile/mark/text.ts
```diff
@@ -81,7 +81,8 @@ export namespace text {
     if (model.has(TEXT)) {
       if (contains([QUANTITATIVE, TEMPORAL], model.fieldDef(TEXT).type)) {
         const format = model.config().mark.format;
-        extend(p, formatMixins(model, TEXT, format));
+        const def = formatMixins(model, fieldDef, format, model.config().mark.shortTimeLabels);
```
kanitw: Why don't put this line in `formatMIxinsText`:

```
const def = formatMixins(model, fieldDef, format, model.config().mark.shortTimeLabels);
```

We don't really need `def` anywhere else here.

test/compile/common.test.ts
```diff
@@ -3,94 +3,104 @@
 import {assert} from 'chai';
 import {parseUnitModel} from '../util';
 import {X} from '../../src/channel';
-import {timeFormat, formatMixins, applyColorAndOpacity} from '../../src/compile/common';
+import {formatMixins, applyColorAndOpacity} from '../../src/compile/common';
+import {format} from '../../src/timeunit';
 
 describe('Model', function() {
```
kanitw: This file isn't really testing model.

test/compile/common.test.ts
```diff
@@ -3,94 +3,104 @@
...
-  describe('timeFormat()', function() {
+  describe('format()', function() {
```
kanitw: `format()` is actually in timeunit.ts.  This test should be moved to `timeunit.test.ts`.

test/compile/common.test.ts
```diff
@@ -3,94 +3,104 @@
...
     it('should get the right time template', function() {
```
kanitw: I would break this `it` into multiple `it` since this is testing many aspects of the method.  
(So that you can use `const` with the `model` below instead of `let`.

src/compile/mark/rule.ts
kanitw: Yes.  It makes sure than for binned axis, we use middle of each bin value for the position.

src/compile/axis.ts
kanitw: I think I meant the `shortTimeLabels` part.  That was done!

src/compile/common.ts
ayush29feb: I think it will be a little confusing to have timeFormat function and timeFormat variable `model.config().timeFormat`

test/compile/common.test.ts
ayush29feb: Fair enough

src/compile/common.ts
kanitw: We do that all the time unless you need to make `timeFormat` a local variable, which isn't true in this case :)

test/compile/common.test.ts
ayush29feb: creating a new test file?

kanitw: Yep. If there isn't one yet.  

Make sure you add the new file's path to `tsconfig.json`  :)

examples/specs/errorbar_aggregate.json
```diff
@@ -0,0 +1,64 @@
...
+          "aggregate": "min",
+          "field": "people",
+          "type": "quantitative",
+          "axis": {"title": "population"}
+        },
+        "y2": {
+          "aggregate": "max",
+          "field": "people",
+          "type": "quantitative",
+          "axis": {"title": "population"}
```
kanitw: Should the example repeat `axis` twice?

src/compile/unit.ts
kanitw: From the meeting with Jeff today, let's make x2, y2 do not accept `scale` and `axis`.

examples/specs/errorbar_aggregate.json
ayush29feb: I don't think so. Will remove it

src/compile/unit.ts
ayush29feb: Oh okay. I will do that.

src/compile/mark/rule.ts
ayush29feb: Oh okay.

ayush29feb:
Box Plot Example from #1397 (To be merged)
image

ayush29feb:
I just discovered a bug while working on ranged `area` in `rule` & `bar`. Although this is not a correct spec. I think @kanitw and I decided we want to implicitly make it work. In the following spec. It is is a `horizontal` ranged bar/rule chart. However, the user instead of using `y` used `y2`.
```
{
  "description": "A bar chart showing the US population distribution of age groups in 2000.",
  "data": {"url": "data/population.json"},
  "mark": "rule" / "bar",
  "encoding": {
    "y2": {
      "field": "age",
      "type": "ordinal"
    },
    "x": {
      "aggregate": "min",
      "field": "people",
      "type": "quantitative"
    },
    "x2": {
      "aggregate": "max",
      "field": "people",
      "type": "quantitative"
    }
  }
}
```

Because of that the user ended up getting
```
"y": {"value": 0},
"x": {"scale": "x","field": "min_people"},
"x2": {"scale": "x","field": "max_people"},
```
In vega, and the following visualization.
image

examples/specs/box_plot.json
```diff
@@ -0,0 +1,77 @@
+{
+  "description": "A error bar plot showing mean, min, and max in the US population distribution of age groups in 2000.",
```
kanitw: This is a box plot.

src/compile/axis.ts
```diff
@@ -92,8 +92,14 @@ export function parseAxis(channel: Channel, model: Model): VgAxis {
...
+  // if X or Y doesn't exist in the model. Use targetChannel
+  // to access X2 or Y2.
+  let targetChannel = channel;
+  if (!model.has(channel)) {
+    targetChannel = channel === X ? X2 : Y2;
+  }
+  extend(def, formatMixins(model, model.fieldDef(targetChannel), axis.format, axis.shortTimeLabels));
```
kanitw: Check if we really need other properties in `fieldDef`  or we simply need `fieldDef.axis`.  If it's the latter case, remove this `targetChannel` thing here because we won't have `axis` in `x2` / `y2`.

src/compile/config.ts
```diff
@@ -29,21 +29,21 @@ export function initMarkConfig(mark: Mark, encoding: Encoding, config: Config) {
...
              } else {
-               cfg[property] = 'horizontal'; // implicitly horizontal
+               cfg[property] = 'horizontal';  // implicitly vertical
              }
            } else if (!xIsMeasure && yIsMeasure) {
              if (mark === TICK) {
                cfg[property] = 'horizontal';
              } else {
-               cfg[property] = 'vertical';
+               cfg[property] = 'vertical';  // implicitly horizontal
```
kanitw: why changing comment here?  The comment also seems to contradict with the code.

src/compile/config.ts
```diff
@@ -29,21 +29,21 @@ export function initMarkConfig(mark: Mark, encoding: Encoding, config: Config) {
...
+           const yIsMeasure = isMeasure(encoding.y) || isMeasure(encoding.y2);
 
            // When unambiguous, do not allow overriding
            if (xIsMeasure && !yIsMeasure) {
              if (mark === TICK) {
-               cfg[property] = 'vertical'; // implicitly vertical
+               cfg[property] = 'vertical';
              } else {
-               cfg[property] = 'horizontal'; // implicitly horizontal
+               cfg[property] = 'horizontal';  // implicitly vertical
```
kanitw: why changing comment here?  The comment also seems to contradict with the code.

src/compile/mark/bar.ts
```diff
@@ -17,6 +17,10 @@ export namespace bar {
 
     const stack = model.stack();
     const xFieldDef = model.encoding().x;
+    const x2FieldDef = model.encoding().x2;
+
+    const xIsMeasure = isMeasure(xFieldDef) || isMeasure(x2FieldDef);
```
kanitw: We still need to refactor `bar` and group output by vega properties, but maybe it's better to do this in a follow-up PR.

src/compile/mark/rule.ts
```diff
@@ -12,25 +12,60 @@ export namespace rule {
     let p: any = {};
 
     // TODO: support explicit value
+    if(model.config().mark.orient === 'vertical') {
```
kanitw: I guess you already extract methods here in some commits.  Cherry pick those?

src/compile/unit.ts
kanitw: let's make x2, y2 do not accept `scale` and `axis`.

src/encoding.ts
```diff
@@ -57,6 +59,11 @@ export function isAggregate(encoding: Encoding) {
   });
 }
 
+export function isRanged(encoding: Encoding) {
+  return (has(encoding, X) && has(encoding, X2)) ||
```
kanitw: I guess `(!!encoding.x  && !!encoding.x2) || (!!encoding.y  && !!encoding.y2)` should be sufficient.  (Or more specific  `(!!encoding.x.field || !!encoding.x.value) ...`)

Using `has` can also be incorrect since has only check `fieldDef.field`. 

If you specify `value` to `x` and `x2`, we should consider that "ranged" too.

src/spec.ts
```diff
@@ -120,18 +129,91 @@ export function normalize(spec: ExtendedSpec): Spec {
...
+export function normalizeRangedUnitSpec(spec: UnitSpec): Spec {
+  if (spec.encoding) {
+    const hasX = has(spec.encoding, X);
+    const hasY = has(spec.encoding, Y);
+    const hasX2 = has(spec.encoding, X2);
+    const hasY2 = has(spec.encoding, Y2);
+    if ((hasX2 && !hasX) || (hasY2 && !hasY)) {
+      let normalizedSpec = duplicate(spec);
+      if (hasX2 && !hasX) {
+        normalizedSpec.encoding.x = duplicate(normalizedSpec.encoding.x2);
```
kanitw: do we need `duplicate` on this line?

src/spec.ts
```diff
@@ -120,18 +129,91 @@ export function normalize(spec: ExtendedSpec): Spec {
...
+        delete normalizedSpec.encoding.x2;
+      }
+      if (hasY2 && !hasY) {
+        normalizedSpec.encoding.y = duplicate(normalizedSpec.encoding.y2);
```
kanitw: do we need `duplicate` on this line?

src/spec.ts
```diff
@@ -120,18 +129,91 @@ export function normalize(spec: ExtendedSpec): Spec {
...
+    spec.transform ? {transform: spec.transform} : {},
+    spec.config ? {config: spec.config} : {}, {layers: []}
+  );
+  if (!spec.encoding) {
+    return layerSpec;
+  }
+  if (spec.mark === ERRORBAR) {
+    const ruleSpec = {
+      mark: RULE,
+      encoding: extend(
```
kanitw: what about color, opacity?

src/spec.ts
```diff
@@ -120,18 +129,91 @@ export function normalize(spec: ExtendedSpec): Spec {
...
+        spec.encoding.x ? {x: duplicate(spec.encoding.x)} : {},
+        spec.encoding.y ? {y: duplicate(spec.encoding.y)} : {},
+        spec.encoding.x2 ? {x2: duplicate(spec.encoding.x2)} : {},
+        spec.encoding.y2 ? {y2: duplicate(spec.encoding.y2)} : {},
+        {})
+    };
+    const lowerTickSpec = {
+      mark: TICK,
+      encoding: extend(
+        spec.encoding.x ? {x: duplicate(spec.encoding.x)} : {},
```
kanitw: what about color, opacity?

src/spec.ts
```diff
@@ -120,18 +129,91 @@ export function normalize(spec: ExtendedSpec): Spec {
...
+        spec.encoding.y ? {y: duplicate(spec.encoding.y)} : {},
+        spec.encoding.size ? {size: duplicate(spec.encoding.size)} : {},
+        {})
+    };
+    const upperTickSpec = {
+      mark: TICK,
+      encoding: extend({
+        x: spec.encoding.x2 ? duplicate(spec.encoding.x2) : duplicate(spec.encoding.x),
```
kanitw: what about color, opacity?

src/spec.ts
kanitw: I wonder if you should duplicate the whole encoding and simply remove size for rule?

ayush29feb: Yeah, I am gonna redo the composite marks. I don't think we should keep it in this PR. Its pretty hacky

kanitw: If removing this chunk of code from thIs PR is troublesome, I'm fine with letting it be merged as is, as long as you send a follow up PR.  

If you want to do that, create another issue and assign to yourself, so that we can track and I'll ignore this part for now :)

ayush29feb:
@kanitw @domoritz I have merged the master branch into this branch but I wasn't too sure about the `normalizeOverlay` method I am not sure if the flow in `normalize` is perfect. I have added a TODO.

PR #1383, opened 2016-05-22
Refactor point by splitting properties into multiple methods

kanitw:
Refactor Point 

- Eliminate `model.has()`
- Split `properties()` into multiple methods that takes `fieldDef`, `scaleName`, `config` (and possibly `scale`) as input that returns Vega's `ValueRef`.    (TODO: declare interface)

### DIFF
--- a/src/compile/mark/point.ts
+++ b/src/compile/mark/point.ts
@@ -1,6 +1,10 @@
-import {UnitModel} from '../unit';
 import {X, Y, SHAPE, SIZE} from '../../channel';
+import {Config} from '../../config';
+import {ChannelDefWithLegend, FieldDef, field} from '../../fielddef';
+import {Scale} from '../../scale';
+
 import {applyColorAndOpacity} from '../common';
+import {UnitModel} from '../unit';
 
 export namespace point {
   export function markType() {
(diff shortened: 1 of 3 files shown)


src/compile/mark/point.ts
```diff
@@ -10,66 +14,77 @@ export namespace point {
...
-    // x
-    if (model.has(X)) {
-      p.x = {
-        scale: model.scaleName(X),
-        field: model.field(X, { binSuffix: '_mid' })
-      };
-    } else {
-      p.x = { value: model.config().scale.bandSize / 2 };
-    }
+    p.x = x(model.encoding().x, model.scaleName(X), config);
```
kanitw: For now, I'm trying not to pass `p` inside these methods and have side effect inside the methods.  
Let's see if this works well for other marks too.

src/compile/mark/point.ts
```diff
@@ -10,66 +14,77 @@ export namespace point {
...
   }
 
-  export function labels(model: UnitModel) {
-    // TODO(#240): fill this method
+  function size(fieldDef: ChannelDefWithLegend, scaleName: string, scale: Scale, config: Config): any {
+    if (fieldDef) {
+      if (fieldDef.field) {
+        return {
+          scale: scaleName,
+          field: field(fieldDef, {scaleType: scale.type})
```
kanitw: Note that for `model.field()` that does not have explicit `binSuffix` provided, we need to pass `scaleType` into the method too.

src/compile/mark/point.ts
```diff
@@ -10,66 +14,77 @@ export namespace point {
...
+      // TODO: fieldDef.value (for layering)
+    }
+    return { value: config.scale.bandSize / 2 };
   }
 
-  export function labels(model: UnitModel) {
-    // TODO(#240): fill this method
+  function size(fieldDef: ChannelDefWithLegend, scaleName: string, scale: Scale, config: Config): any {
+    if (fieldDef) {
+      if (fieldDef.field) {
```
kanitw: model.has is equivalent to `fieldDef && fieldDef.field`

PR #1384, opened 2016-05-22
Make setting `axis` or `legend` = `null` as the default way to disable them.

kanitw:
Remove the support for `axis: false` and just support `axis: null` and simplify the schema 
as briefly discussed in #1377.


kanitw:
Looks like we're blocked because our JSON schema validator doesn't support null yet.

domoritz:
I read the error message differently. I think we just have to add `null` as an option (null seems to be a type like number or object).

domoritz:
Looks like I'm right https://spacetelescope.github.io/understanding-json-schema/reference/null.html

domoritz:
Okay, issue is that there is no null type in typescript and we are blocked on https://github.com/YousefED/typescript-json-schema/issues/36 but nullable types are coming to ts soon (already merged) https://github.com/Microsoft/TypeScript/pull/7140/

aishfenton:
How will nullable be expressed in the scheme? In wrapper languages we'll need to handle this case explicitly (since not all types can be null, so will need to do something funky).

Json schema has "required" but that really means undefinable, and is already being used for that purpose. 

And in general in JavaScript is it ok to have null != undefined? I don't know since I'm not a JS guy, but in typed languages there's really no equivalent to this, so having them mean something different will be tricky to support (for example in the Json parser I'm using it assume they're equivalent).

javadba:
How does this work in practice? I can find zero documentation. After sprinkling liberally in a spec the legend is still there:

                "mark": "bar",
                "encoding": {
                    "x": {"field": "State", "type": "nominal", "legend": "null"},
                    "y": {"field": "State_Crime_Rate", "type": "quantitative", "scale": {"type": "log"}, "legend": "null"},
                    "color": {"field": "State", "type": "nominal"}
                },
                "config": {
                    "legend": "null"

javadba:
Saw an example with `null` (no quotes). But following also does not prevent the legend:

                    "x": {"field": "State", "type": "nominal", "legend": null},
                    "y": {"field": "State_Crime_Rate", "type": "quantitative", "scale": {"type": "log"}, "legend": null},
                    "color": {"field": "State", "type": "nominal"}
                },
                "config": {
                    "legend": null

domoritz:
@aishfenton We explicitly mark properties that can be `null`. In typescript, they have a `@nullable` annotation and in the schema the properties are an enum with `null` as a possible value.

PR #1397, opened 2016-05-31
added X2/Y2 to bar mark with tests. Added basic box plot example.

ayush29feb:
- [x] Added `X2`/`Y2`
- [x] Added test cases
- [x] Added Example

### DIFF
--- a/examples/specs/box_plot.json
+++ b/examples/specs/box_plot.json
@@ -15,8 +15,7 @@
         "y2": {
           "aggregate": "max",
           "field": "people",
-          "type": "quantitative",
-          "axis": {"title": "population"}
+          "type": "quantitative"
         }
       }
     },
@@ -27,8 +26,7 @@
         "y": {
           "aggregate": "min",
           "field": "people",
-          "type": "quantitative",
-          "axis": {"title": "population"}
+          "type": "quantitative"
         },
         "size": {"value": 5}
       }
@@ -40,8 +38,7 @@
         "y": {
           "aggregate": "max",
           "field": "people",
-          "type": "quantitative",
-          "axis": {"title": "population"}
+          "type": "quantitative"
         },
         "size": {"value": 5}
       }
@@ -70,8 +67,7 @@
         "y": {
           "aggregate": "mean",
           "field": "people",
-          "type": "quantitative",
-          "axis": {"title": "population"}
+          "type": "quantitative"
         },
         "size": {"value": 5},
         "color": { "value" : "white" }
--- a/examples/specs/box_plot.json
+++ b/examples/specs/box_plot.json
@@ -1,0 +1,81 @@
+{
+  "description": "A error bar plot showing mean, min, and max in the US population distribution of age groups in 2000.",
+  "data": {"url": "data/population.json"},
+  "layers": [
+    {
+      "mark": "rule",
+      "encoding": {
+        "x": {"field": "age","type": "ordinal"},
+        "y": {
+          "aggregate": "min",
+          "field": "people",
+          "type": "quantitative",
+          "axis": {"title": "population"}
+        },
+        "y2": {
+          "aggregate": "max",
+          "field": "people",
+          "type": "quantitative",
+          "axis": {"title": "population"}
+        }
+      }
+    },
+    {
+      "mark": "tick",
+      "encoding": {
+        "x": {"field": "age","type": "ordinal"},
+        "y": {
+          "aggregate": "min",
+          "field": "people",
+          "type": "quantitative",
+          "axis": {"title": "population"}
+        },
+        "size": {"value": 5}
+      }
+    },
+    {
+      "mark": "tick",
+      "encoding": {
+        "x": {"field": "age","type": "ordinal"},
+        "y": {
+          "aggregate": "max",
+          "field": "people",
+          "type": "quantitative",
+          "axis": {"title": "population"}
+        },
+        "size": {"value": 5}
+      }
+    },
+    {
+      "mark": "bar",
+      "encoding": {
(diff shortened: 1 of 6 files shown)


test/compile/mark/bar.test.ts
```diff
@@ -100,4 +100,40 @@ describe('Mark: Bar', function() {
       });
     });
   });
+  
+  describe('ranged bar', function() {
+    it('vertical bars should work with aggregate', function() {
```
kanitw: Incorrect description of what you're testing:  'vertical bars should work with aggregate'

test/compile/mark/bar.test.ts
```diff
@@ -100,4 +100,40 @@ describe('Mark: Bar', function() {
...
+        }
+      });
+
+      const props = bar.properties(model);
+      assert.deepEqual(props.xc, { scale: 'x', field: 'age'});
+      assert.deepEqual(props.y, { scale: 'y', field: 'q1_people' });
+      assert.deepEqual(props.y2, { scale: 'y', field: 'q3_people' });
+    });
+    
+    it('vertical bars should work with aggregate', function() {
```
kanitw: Incorrect description of what you're testing:  'vertical bars should work with aggregate'

This seems like horizontal

examples/specs/box_plot.json
```diff
@@ -0,0 +1,81 @@
...
+          "aggregate": "min",
+          "field": "people",
+          "type": "quantitative",
+          "axis": {"title": "population"}
+        },
+        "y2": {
+          "aggregate": "max",
+          "field": "people",
+          "type": "quantitative",
+          "axis": {"title": "population"}
```
kanitw: we'll remove this :)

examples/specs/box_plot.json
```diff
@@ -0,0 +1,81 @@
...
+    },
+    {
+      "mark": "tick",
+      "encoding": {
+        "x": {"field": "age","type": "ordinal"},
+        "y": {
+          "aggregate": "min",
+          "field": "people",
+          "type": "quantitative",
+          "axis": {"title": "population"}
```
kanitw: @domoritz  @arvind  We only need to say `axis.title` once for layers.  Should example specs for this kind of things repeat `axis.title` multiple times?  (I don't think so.)   Which layer should we add the `axis.title` to?

test/compile/mark/bar.test.ts
```diff
@@ -100,4 +100,40 @@ describe('Mark: Bar', function() {
...
+        }
+      });
+
+      const props = bar.properties(model);
+      assert.deepEqual(props.xc, { scale: 'x', field: 'age'});
+      assert.deepEqual(props.y, { scale: 'y', field: 'q1_people' });
+      assert.deepEqual(props.y2, { scale: 'y', field: 'q3_people' });
+    });
+    
+    it('horizontal bars should work with aggregate', function() {
```
kanitw: "work with ranged x"?

test/compile/mark/bar.test.ts
kanitw: "work with ranged y"?

src/compile/mark/bar.ts
```diff
@@ -17,6 +17,10 @@ export namespace bar {
 
     const stack = model.stack();
```
kanitw: @ayush29feb   Your question about stack is a very good one. 

In fact, the logic inside `mark/bar.ts` for stack should be correct here provided that the `stack` variable is correct. 
Thus, you have to prevent `stack` becomes truthy when `x2` or `y2` is specified, so that it doesn't go into that `if` clause.
You should to track down where this `stack` value comes from.  If you have questions, let me know.   
(Reading the docs about stack might help you understand when stack is turned on.)

test/compile/mark/bar.test.ts
ayush29feb: It says ranged in the describe

src/compile/mark/bar.ts
ayush29feb: Ah I see. Currently, it doesn't affect the truthy of `stack` when `x2` or `y2` is specified. It essentially ignores it. Although, the scales are made with a union of the two domains(`x` + `x2`). I am not sure how we should handle that. Maybe create the stack before scales and then check for the `stack` when creating `domain` for the scales. 

Also, I wanted to ask if we wanna allow to make stacked bar charts with `x2` instead of `x`. That might require a little work but if not then it seems to be okay.

kanitw: If you specify both `x` and `x2`, it shouldn’t be automatically stacked by default.

The question is what if user specify `stacked` property, and specify `x2` -- which are conflicting.  Which one of them get precedence.  I would say `x2` has stronger precedence because `x2` is closer to the output that we finally produce (If you think about it, stack also produces `x2` in Vega).
Basically, whenever `x2` appears, we never make `stacked` truthy.

examples/specs/box_plot.json
arvind: Agreed to only say `axis.title` once for layers. My vote would be for the first layer, such that it feels like it cascades down the textual specification (or layered up in the canvas).

test/compile/stack.test.ts
```diff
@@ -1,3 +1,21 @@
 /* tslint:disable:quotemark */
 
-// TODO: write properly
+import {assert} from 'chai';
+import {parseUnitModel} from '../util';
+
+describe('Stack', function() {
+  describe('compileStackProperties()', function() {
+    it('should return undefined from ranged bars with aggregate', function() {
```
kanitw: For test coverage, you should test a case with `y` and `y2` too.  
(Take 30 seconds more to do, so doesn't hurt to do it.)

PR #1401, opened 2016-06-02
Added Support for Ranged Area Marks

ayush29feb:
- [x] Added Ranged Marks Support
- [x] Added Test Cases
- [x] Refactor for #1403


src/compile/mark/area.ts
```diff
@@ -19,65 +19,134 @@ export namespace area {
...
-    // x2
-    if (orient === 'horizontal') {
-      if (stack && X === stack.fieldChannel) {
+      if (orient === 'horizontal') {
         p.x2 = {
           scale: model.scaleName(X),
           field: model.field(X, { suffix: '_end' })
         };
+      }
+    } else if (isMeasure(xFieldDef) || isMeasure(x2FieldDef)) { // Measure
```
kanitw: why do you need to group `x` and `x2` logic together?  

The old code decouples `x` and `x2`  (same for `y` / `y2`).

test/compile/mark/area.test.ts
```diff
@@ -49,4 +49,38 @@ describe('Mark: Area', function() {
       assert.deepEqual(props.fill, {scale: COLOR, field: 'Miles_per_Gallon'});
     });
   });
+
```
kanitw: Good that you add new tests for new cases.  However, it seems like we don't have test to cover old cases yet and you're changing the logic for old cases in this PR and we should prevent regression.  

Can you checkout the main branch, write unit test for other cases in areas. (use `npm run cover` to see coverage in area.ts), then merge those tests into this branch?  This way we can make sure that we don't introduce new bug in this branch.  :)

kanitw: If you want to punt for after presentation, at least we confirm that you have test with the examples and horizontal/vertical stacked/non-stacked area chart are still doing alright.

src/compile/mark/area.ts
ayush29feb: Mostly the reason was for consistency with `rule` and `bar` where we had a similar logic. In the code before this we were doing `x` `x2` `y` y2` all separately and had independent `if` clauses for each of them. However, `x` & `x2` or `y` & `y2` are actually dependent on each other in some cases where we want to implicitly handle a incorrect `spec`. While reading through this code, I also discovered another problem in #1382 for `rule` and `bar` mark.

test/compile/mark/area.test.ts
ayush29feb: All the old examples in the vega-editor do work. But I didn't know there were no tests for the old code. I can quickly write those tests.

src/compile/mark/area.ts
kanitw: Actually we should make `rule` and `bar` decoupled later too. (If you observe my `point` refactor PR closely.)  

So I would say, if we can keep  `area` decoupled that will save your time later. 

That said, I just asked You Ying to standardize mark.properties methods (except `rule` and `bar`), so maybe you can wait for that before fixing this to avoid merge conflicts.

kanitw: Ok.  I merged that one into master.  You probably want to merge master into your branches.

PR #1405, opened 2016-06-03
Normalization of UnitSpec

ayush29feb:
- [x] Write Tests

### DIFF
--- a/src/spec.ts
+++ b/src/spec.ts
@@ -10,7 +10,7 @@ import {Facet} from './facet';
 import {Mark} from './mark';
 import {Transform} from './transform';
 
-import {COLOR, SHAPE, ROW, COLUMN} from './channel';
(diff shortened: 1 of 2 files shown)


src/spec.ts
```diff
@@ -127,8 +136,27 @@ export function normalize(spec: ExtendedSpec): Spec {
...
+}
 
+export function normalizeUnitSpec(spec: UnitSpec): Spec {
+  if (spec.encoding) {
+    const hasX = has(spec.encoding, X);
+    const hasY = has(spec.encoding, Y);
+    const hasX2 = has(spec.encoding, X2);
+    const hasY2 = has(spec.encoding, Y2);
+
+    let normalizedSpec = duplicate(spec);
```
kanitw: I guess you only need to duplicate if one of the case below is true.

PR #1406, opened 2016-06-03
[Obsolete] Labels

willium:
This is the beginning of the implementation of labeling. Also see https://github.com/willium/vega/tree/will/labels

### DIFF
--- a/src/compile/mark/label.ts
+++ b/src/compile/mark/label.ts
@@ -66,7 +66,7 @@ export namespace label {
     
     let t: any = { 
       type: 'label',
-      anchor: model.has(ANCHOR) ? ANCHOR : 10,
+      anchor: model.has(ANCHOR) ? ANCHOR : 'bottom-center',
       offset: model.has(OFFSET) ? OFFSET : 10,
       color: model.has(COLOR) ? COLOR : 'black',
       opacity: opacity ? opacity : 1,
--- a/src/compile/mark/label.ts
+++ b/src/compile/mark/label.ts
@@ -24,12 +24,12 @@ export namespace label {
   
   export function properties(model: UnitModel, visible: boolean=false) {
     // TODO Use Vega's marks properties interface
-    let p: any = {
+    let p: any = visible ? {
       xc:    {field: 'label_xc'}, // presets so transform can change values
       yc:    {field: 'label_yc'},
       align: {field: 'label_align'},
       fill:  {field: 'label_color'}
-    };
+    } : {};
 
     applyMarkConfig(p, model,
       ['angle', 'baseline', 'dx', 'dy', 'font', 'fontWeight',
@@ -39,7 +39,7 @@ export namespace label {
 
     // size
     if (model.has(SIZE)) {
-      p.fontSize = { field: model.field(SIZE, {datum: true}) };
+      p.fontSize = { field: model.field(SIZE, extend({datum: true}, visible?{prefn: 'datum.'}:{})) };
     } else {
       p.fontSize = { value: sizeValue(model) };
     }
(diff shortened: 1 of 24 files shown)


src/compile/mark/label.ts
```diff
@@ -0,0 +1,95 @@
...
+        scale: model.scaleName(COLOR),
+        field: model.field(COLOR, model.fieldDef(COLOR).type === ORDINAL ? {prefn: 'rank_'} : {})
+      }
+    };
+  }
+
+  export function properties(model: UnitModel) {
+    // TODO Use Vega's marks properties interface
+    let p: any = {
+      xc: {field: 'label_xc'},
```
kanitw: Add comment that these are preset output of the transform?

src/compile/mark/label.ts
```diff
@@ -0,0 +1,95 @@
...
+    } else if (fieldDef.value) {
+      p.text = { value: fieldDef.value };
+    }
+
+    return p;
+  }
+  
+  export function transforms(model: UnitModel) {
+    const opacity        = model.config().mark.opacity;
+    const anchorFieldDef = model.encoding()['anchor'];
```
kanitw: .anchor?  (This way you get type for free)

src/compile/mark/label.ts
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    applyMarkConfig(p, model,
+      ['angle', 'baseline', 'dx', 'dy', 'font', 'fontWeight',
+        'fontStyle', 'radius', 'theta', 'text']);
+        
+    const fieldDef = model.fieldDef(TEXT);
+
+    // size
+    if (model.has(SIZE)) {
```
kanitw: split this into `size()` method and get rid of `sizeValue` method like other files.  (We just refactored this yesterday.)

src/compile/mark/label.ts
```diff
@@ -0,0 +1,95 @@
...
+      p.fontSize = { field: model.field(SIZE, {datum: true}) };
+    } else {
+      p.fontSize = { value: sizeValue(model) };
+    }
+
+    p.opacity = {field: 'label_opacity'};
+
+    // text
+    if (model.has(TEXT)) {
```
kanitw: split into `text()`

src/compile/model.ts
```diff
@@ -231,6 +231,10 @@ export abstract class Model {
   public data() {
     return this._data;
   }
+  
+  public dataRef(suffix?: string, delimiter: string = '_') {
+    return this._data !== undefined && this._data.hasOwnProperty('ref') ? this._data['ref'] + (suffix ? delimiter + suffix : '') : false;
```
kanitw: Do you really need other delimiter as a paramter here?

Also can simplify to
```
 this._data && this._data.ref ? this._data.ref + (suffix ? delimiter + suffix : '') : false;
```

kanitw:
- [ ] I guess `label` requires `data.ref`.  Maybe we should add warning somewhere that if we don't have that with `label` mark it should not work.

src/compile/mark/mark.ts
```diff
@@ -134,33 +137,16 @@ function parseNonPathMark(model: UnitModel) {
         model.has(ORDER) ?
           // if non-stacked, detail field determines the layer order of each mark
           { transform: [{type:'sort', by: sortBy(model)}] } :
+          {},
+        mark === LABEL && referenceMark ?
```
kanitw: If you specify order,  you should be able to order the label, provided that your label transform might still allow some occlusion?  

(Order should come after label - I think.)

Issue #1322, opened 2016-04-25
Maps - Milestone 1: Projections


bobocandys:
- Default projection: Mercator
- Mapping from latitude/longitude to x/y in vega-lite.

kanitw:
- [ ]  Add new type of mark “path” 

- [ ] Add new file `src/compile/data/geo.ts` for geo and geopath transform
  -  [ ] if `mark == 'path'`, then do `geopath` transform else do `geo` transform

Let's revise the TODO more once we are done with input/output

kanitw:
will address as a part of #1423

PR #1408, opened 2016-06-05
Geo projections

bobocandys:
Address issue: #1322 
- [ ] geo
  - [x] points
  - [ ] other marks 
- [ ] geopath
  Q:Support `feature` in `data` in vega-lite?


src/compile/common.ts
```diff
@@ -183,3 +186,58 @@ export function timeFormat(model: Model, channel: Channel): string {
...
+export function hasGeoTransform(model: UnitModel): boolean {
+  const geoPathFieldDef = model.encoding().geopath;
+  if (model.mark() === PATHMARK) {
+    if (geoPathFieldDef && geoPathFieldDef.type === GEOJSON) {
+      return true;
+    }
+  }
+  const xFieldDef = model.encoding().x;
+  const yFieldDef = model.encoding().y;
+  if (xFieldDef &&
```
kanitw: add comment that we only require x or y just in case user want to plot lat/long against another measure.  

(So that reader realize this isn't a mistake.)

kanitw: Also, seems like you can use your extracted `containsLatLong` method.

PR #1415, opened 2016-06-07
Added Error Bar Composite Mark

ayush29feb:
TODO
- [ ] normalize `composite unitSpec` in `LayerSpec`.`layers`
- [x] copy `size` to ticks

### DIFF
--- a/src/spec.ts
+++ b/src/spec.ts
@@ -5,7 +5,7 @@ import {FieldDef} from './fielddef';
 
 import {Config} from './config';
 import {Data} from './data';
-import {Encoding, UnitEncoding, has} from './encoding';
+import {Encoding, UnitEncoding, has, isRanged} from './encoding';
 import {Facet} from './facet';
 import {Mark} from './mark';
 import {Transform} from './transform';
@@ -142,7 +142,10 @@ export function normalizeUnitSpec(spec: UnitSpec): Spec {
   if (contains([ERRORBAR], spec.mark)) {
     return normalizeCompositeUnitSpec(spec);
   }
-  return normalizeRangedUnitSpec(spec);
+  if (isRanged(spec.encoding)) {
+    return normalizeRangedUnitSpec(spec);
+  }
+  return spec;
 }
 
 export function normalizeRangedUnitSpec(spec: UnitSpec): Spec {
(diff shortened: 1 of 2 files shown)


src/spec.ts
```diff
@@ -161,6 +168,42 @@ export function normalizeUnitSpec(spec: UnitSpec): Spec {
   return spec;
 }
 
+export function normalizeCompositeUnitSpec(spec: UnitSpec): Spec {
+  let layerSpec = extend(spec.name ? {name: spec.name} : {},
+    spec.description ? {description: spec.description} : {},
+    spec.data ? {data: spec.data} : {},
+    spec.transform ? {transform: spec.transform} : {},
+    spec.config ? {config: spec.config} : {}, {layers: []});
```
kanitw: break line before the last `);`
and break line before starting the if clause.

src/spec.ts
```diff
@@ -161,6 +168,42 @@ export function normalizeUnitSpec(spec: UnitSpec): Spec {
...
+  if (!spec.encoding) {
+    return normalize(layerSpec);
```
kanitw: I don't really know why you need this clause.  Is it necessary?

src/spec.ts
```diff
@@ -139,6 +139,13 @@ export function normalizeExtendedUnitSpec(spec: ExtendedUnitSpec): Spec {
 }
 
 export function normalizeUnitSpec(spec: UnitSpec): Spec {
+  if (contains([ERRORBAR], spec.mark)) {
+    return normalizeCompositeUnitSpec(spec);
+  }
+  return normalizeRangedUnitSpec(spec);
```
kanitw: If would put the condition to check if this spec is range here -- for consistency.  (so `normalizeRangedUnitSpec` only transform the spec)

kanitw:
- [ ] (later) unit test

kanitw:
- [ ] unit test (later)

src/spec.ts
ayush29feb: Wasn't sure if you can `duplicate` `undefined`

kanitw: Good point.  But why do we need to call `normalize` another time?

kanitw: I think if there is no `encoding`, it shouldn't even enter this method from the beginning.

kanitw: We can make sure `isUnitSpec` check for `spec.encoding`.

ayush29feb: Okay.. that works better.

ayush29feb:
@kanitw @domoritz should be allow composite mark types to be used within Layers? I am guessing yes. Can either of you point me where does each `unitSpec` is being `normalize`'d?

PR #1416, opened 2016-06-07
Also filter NaN values

### LINKED ISSUE #1411: Broken Line Chart for TimeUnit (Maybe due to vega sort transform)
```
{
  "data": {
    "url": "data/movies.json"
  },
  "mark": "line",
  "encoding": {
    "x": {
      "timeUnit": "month",
      "field": "Release_Date",
      "type": "temporal"
    },
    "y": {
      "aggregate": "count",
      "field": "*",
      "type": "quantitative"
    }
  }
}
```

vega_editor

looks like the inline sort transform is somehow incorrect in Vega

```
"transform": [{"type": "sort","by": "-month_Release_Date"}]
```

Maybe `by` in sort transform doesn't work with date object? 

Changing mark type in the Vega output to `symbol` works fine. 

vega_editor

kanitw:
Fix #1411

Question
- Should we only filter this for date field?
- Should we rename `filterNull` to `filterInvalidValues` since it's no longer just null? 

Now temporal line chart sorts correctly!

vega_editor

### DIFF
--- a/src/compile/data/nullfilter.ts
+++ b/src/compile/data/nullfilter.ts
@@ -73,7 +73,8 @@ export namespace nullFilter {
       [{
         type: 'filter',
         test: filteredFields.map(function(fieldName) {
-          return 'datum.' + fieldName + '!==null';
+          return '(datum.' + fieldName + '!==null' +
+            ' && !isNaN(datum.'+ fieldName + '))';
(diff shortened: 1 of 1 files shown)


kanitw:
I'll merge this for now.  We can further discuss in #1426

PR #1433, opened 2016-06-17
Extract `stack` method for external use and add test 

kanitw:
I need this for CompassQL


kanitw:
OK updated.  Should be correct now, with 100% test coverage in `stack.ts`.

kanitw:
Actually let me look at the gallery first!

kanitw:
Ok.  It's good now.  Each commit here are well splitted, so you might want to look at each of them, one at a time.

PR #1437, opened 2016-06-18
ylin/update format type

bobocandys:
- Update the logic for data.format.feature and data.format.mesh. (They should be mutually exclusive: https://github.com/vega/vega/wiki/Data#-topojson) 
- Add tests for data.format.feature and data.format.mesh


kanitw:
@bobocandys   Thanks!  I just did some clean up and also add property.

PR #1445, opened 2016-06-25
Refactor text mark, timeFormat, timeTemplate

bobocandys:
- split methods inside `text` marks
- change `formatMixins` => `numberFormat`

### DIFF
--- a/src/compile/mark/text.ts
+++ b/src/compile/mark/text.ts
@@ -67,7 +67,7 @@ export namespace text {
         };
       }
     }
-
+    // TODO: support x.value, x.datum
     if (textFieldDef && textFieldDef.type === QUANTITATIVE) {
       return { field: { group: 'width' }, offset: -5 };
     } else {
@@ -98,27 +98,9 @@ export namespace text {
           field: field(sizeFieldDef)
         };
       }
-<<<<<<< HEAD
       if (sizeFieldDef.value) {
         return {value: sizeFieldDef.value};
       }
-=======
-    } else {
-      return { value: sizeValue(fieldDefSize, config) };
-    }
-  }
-
-  // when color is apply to background, only set opacity here.
-  function applyColorAndOpacity(p, model: UnitModel) {
-    if (model.config().mark.applyColorToBackground && !model.has(X) && !model.has(Y)) {
-      p.fill = {value: 'black'}; // TODO: add rules for swapping between black and white
-
-      // opacity
-      const opacity = model.config().mark.opacity;
-      if (opacity) { p.opacity = { value: opacity }; };
-    } else {
-      applyColorAndOpacity(p, model);
->>>>>>> Add local applyColorAndOpacity method
     }
     return { value: config.mark.fontSize };
   }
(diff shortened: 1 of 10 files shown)


src/compile/mark/text.ts
```diff
@@ -30,42 +35,69 @@ export namespace text {
...
+    p.x = x(model.encoding().x, model.scaleName(X), config, fieldDef);
+
+    p.y = y(model.encoding().y, model.scaleName(Y), config, fieldDef);
+
+    p.fontSize = size(model.encoding().size, model.scaleName(SIZE), model.scale(SIZE), config);
+
+    p.text = text(model.encoding().text, model.scaleName(TEXT), config);
+
+    // FIXME: need this?
+    applyColorAndOpacity(p, model);
```
bobocandys: I think we don't need this in the properties function because it can be applied in the opacity method.

src/compile/mark/text.ts
```diff
@@ -30,42 +35,69 @@ export namespace text {
...
+          field: field(SIZE)
+        };
+      }
     } else {
-      p.fontSize = { value: sizeValue(model) };
+      return { value: sizeValue(fieldDefSize, config) };
     }
+  }
 
+  function opacity() {
```
bobocandys: Not sure how to implement this function in the same style as others.

src/compile/mark/text.ts
```diff
@@ -75,29 +107,29 @@ export namespace text {
...
-      if (contains([QUANTITATIVE, TEMPORAL], model.fieldDef(TEXT).type)) {
-        const format = model.config().mark.format;
-        extend(p, formatMixins(model, TEXT, format));
-      } else {
-        p.text = { field: model.field(TEXT) };
+    if (fieldDefText) {
+      if (fieldDefText.field) {
+        if (contains([QUANTITATIVE, TEMPORAL], fieldDefText.type)) {
+          const format = config.mark.format;
+          extend(p, formatMixins(model, TEXT, format));
```
bobocandys: This branch needs the model for the `formatMixins` function, how can we get around this?

src/compile/mark/text.ts
kanitw: Hmm.  I think this is a combo that does both color and opacity.  
I guess we can keep it for consistency with other marks?

kanitw: Oh I see what you mean now.  I think the opacity method should become a local `applyColorAndOpacity` which might call the imported `applyColorAndOpacity`.

src/compile/mark/text.ts
```diff
@@ -30,42 +35,69 @@ export namespace text {
...
+        };
+      }
     } else {
-      p.fontSize = { value: sizeValue(model) };
+      return { value: sizeValue(fieldDefSize, config) };
     }
+  }
 
+  function opacity() {
     if (model.config().mark.applyColorToBackground && !model.has(X) && !model.has(Y)) {
```
kanitw: Add comment that 

// when color is apply to background, only set opacity here.

src/compile/mark/text.ts
kanitw: @ayush29feb  has done some refactor for the `formatMixins` method, he will try to make that part mergable soon.

kanitw:
@bobocandys just clean the rest of this PR up for you and merged.  Thanks for helping!

PR #1446, opened 2016-06-25
Un-constrain shape from a fixed enum for custom shapes

mprudhom:
Vega now permits custom shapes: https://github.com/vega/vega/wiki/Custom-Shapes

They work with vega-lite "out of the box":

```json
{
  "description": "A scatterplot with custom star shapes.",
  "mark": "point",
  "encoding": {
    "y": {"field": "Miles_per_Gallon","type": "quantitative"},
    "size": {"field": "Cylinders", "type": "quantitative"}
  },
  "config": {
    "mark": { 
      "shape": "M0,0.2L0.2351,0.3236 0.1902,0.0618 0.3804,-0.1236 0.1175,-0.1618 0,-0.4 -0.1175,-0.1618 -0.3804,-0.1236 -0.1902,0.0618 -0.2351,0.3236 0,0.2Z"
    }
  }
}
```

unknown-1

However, this spec does not pass schema validation, since the SVG path is not one of a fixed set of enums. This PR removes the shape enum constraint so that custom SVG paths will pass schema validation.

### DIFF
--- a/site/docs/config.md
+++ b/site/docs/config.md
@@ -207,7 +207,7 @@ vg.embed('#horizontal_line', {
 
 | Property            | Type                | Description  |
 | :------------------ |:-------------------:| :------------|
-| shape               | Number              | The symbol shape to use. One of `"circle"`, `"square"`, `"cross"`, `"diamond"`, `"triangle-up"`, or `"triangle-down"`<span class="note-line">__Default value:__ `"circle"` </span> |
+| shape               | Number              | The symbol shape to use. One of `"circle"`, `"square"`, `"cross"`, `"diamond"`, `"triangle-up"`, or `"triangle-down"`, or else a custom SVG path string<span class="note-line">__Default value:__ `"circle"` </span> |
 
 
 ### Point Size Config (for Point, Circle, and Square Marks)
--- a/site/docs/config.md
+++ b/site/docs/config.md
@@ -207,7 +207,7 @@ vg.embed('#horizontal_line', {
 
 | Property            | Type                | Description  |
 | :------------------ |:-------------------:| :------------|
-| shape               | Number              | The symbol shape to use. One of `"circle"`, `"square"`, `"cross"`, `"diamond"`, `"triangle-up"`, or `"triangle-down"`, or else a custom SVG path string<span class="note-line">__Default value:__ `"circle"` </span> |
+| shape               | Number              | The symbol shape to use. One of `"circle"`, `"square"`, `"cross"`, `"diamond"`, `"triangle-up"`, or `"triangle-down"`, or else a custom SVG path string.<span class="note-line">__Default value:__ `"circle"` </span> |
 
 
 ### Point Size Config (for Point, Circle, and Square Marks)
--- a/src/compile/legend.ts
+++ b/src/compile/legend.ts
@@ -124,7 +124,8 @@ export namespace properties {
         break;
     }
 
-    const filled = model.config().mark.filled;
+    const cfg = model.config();
+    const filled = cfg.mark.filled;
 
 
     let config = channel === COLOR ?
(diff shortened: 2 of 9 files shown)


src/config.ts
```diff
@@ -293,9 +293,9 @@ export interface MarkConfig {
 
   // ---------- Point ----------
   /**
-   * The symbol shape to use. One of circle (default), square, cross, diamond, triangle-up, or triangle-down.
+   * The symbol shape to use. One of circle (default), square, cross, diamond, triangle-up, or triangle-down, or a custom SVG path.
    */
-  shape?: Shape;
+  shape?: string;
```
kanitw: Please use `Shape | string` so that enum still works and update docs `encoding.md` accordingly.

src/config.ts
```diff
@@ -377,7 +377,7 @@ export interface MarkConfig {
 
 export const defaultMarkConfig: MarkConfig = {
   color: '#4682b4',
-  shape: Shape.CIRCLE,
+  shape: 'circle',
```
kanitw: Please undo this line.

kanitw:
Thanks for the new commit.  The encoding.md is still out of date though.

src/compile/legend.ts
```diff
@@ -162,15 +163,26 @@ export namespace properties {
...
     if (legend.symbolColor !== undefined) {
       symbols.fill = {value: legend.symbolColor};
+    } else if (symbols.fill === undefined) {
+      // fall back to mark config colors for legend fill
+      if (cfg.mark.color !== undefined) {
+        symbols.fill = {value: cfg.mark.color};
+      } else if (cfg.mark.fill !== undefined) {
+        symbols.fill = {value: cfg.mark.fill};
+      } else if (cfg.mark.stroke !== undefined) {
+        symbols.fill = {value: cfg.mark.stroke};
```
kanitw: why isn't this `symbols.stroke` ?

mprudhom: My thinking is that if a mark's "color" and "fill" are unset but "stroke" is set, then the stroke color represents to mark's canonical color, and I think that the legend symbols should always be filled even if the mark itself is not. But I can go either way on that. Let me know if you think legend symbols should be hollow if the mark's config declares a stroke color but no fill color.

domoritz:
Just to make sure. After this pr the legend will also use the custom symbol, right?

mprudhom:
Yes, the legend will use the custom symbol. 

unnamed document

site/docs/config.md
```diff
@@ -207,7 +207,7 @@ vg.embed('#horizontal_line', {
 
 | Property            | Type                | Description  |
 | :------------------ |:-------------------:| :------------|
-| shape               | Number              | The symbol shape to use. One of `"circle"`, `"square"`, `"cross"`, `"diamond"`, `"triangle-up"`, or `"triangle-down"`<span class="note-line">__Default value:__ `"circle"` </span> |
+| shape               | Number              | The symbol shape to use. One of `"circle"`, `"square"`, `"cross"`, `"diamond"`, `"triangle-up"`, or `"triangle-down"`, or else a custom SVG path string<span class="note-line">__Default value:__ `"circle"` </span> |
```
kanitw: Can you add a `.` (dot) at the end of sentence.  (Yeah, we forgot it.)

src/compile/legend.ts
kanitw: Apologies for delayed reply, I think if the marks are hollow, the legend should consistently be hollow too.  Thank you.

kanitw:
- [ ] missing unit tests for the new cases added in `properties.symbols` in `legend.ts`.

kanitw:
@mprudhom We're planning to release 1.2 soon, so if you have time to clean this up -- we can include it as a part of 1.2.  Thanks!

src/compile/legend.ts
mprudhom: OK, I made it so the legend is only filled when the mark is filled.

site/docs/config.md
mprudhom: Done.

kanitw:
Thanks for fixing, only one thing left: unit test.  And we can merge :)

mprudhom:
You had self-assigned that task, so I though you had intended to do it.

I've gone ahead and committed a test case. Let me know if you'd like it to cover anything beyond the legend shape path matching the underlying shape's path.

kanitw:
Thanks.  

> You had self-assigned that task, so I though you had intended to do it.

Not sure when did I indicate that.  If you mean that I assign this PR to myself,  generally assigning a PR means assign a reviewer.   But thanks for fixing anyway!

mprudhom:
I had thought that the "kanitw self-assigned this 11 days ago" note in the issue meant that you had assigned that one "missing unit tests" line-item to yourself. Anyway, thanks for merging!

kanitw:
Haha.  Yeah that reads confusing :p

src/compile/legend.ts
```diff
@@ -162,15 +163,26 @@ export namespace properties {
...
       symbols[filled ? 'fill' : 'stroke'] = symbols[filled ? 'fill' : 'stroke'] ||
-        {value: model.config().mark.color};
+        {value: cfg.mark.color};
     }
 
     if (legend.symbolColor !== undefined) {
       symbols.fill = {value: legend.symbolColor};
+    } else if (symbols.fill === undefined) {
+      // fall back to mark config colors for legend fill
+      if (cfg.mark.color !== undefined) {
```
kanitw: FYI, Looks like this line introduced a bug #1500 -- but I've just fixed in https://github.com/vega/vega-lite/pull/1504

PR #1455, opened 2016-07-02
Extract function to compute orientation

domoritz:
With tests. 

Also contained in #1304


domoritz:
Much simpler now. @kanitw, ready to merge.

kanitw:
I think having "range" takes precedence over being measure.  I'll revise the logic and add more tests to cover the cases I'm thinking about and will ping you.

kanitw:
Instead of having flip logic, I just rewrite the rule so that it should be correct now.   Feel free to take a look.

kanitw:
Merging as we should release things with correct behavior.  

@domoritz  feel free to send follow-up comments.

PR #1461, opened 2016-07-11
Support Filter Object

### LINKED ISSUE #1394: Support Filter Object
Forking from #1376.   We are considering filter object syntax 

```
"filter": [
  {"field": "foo", "equal": "a", "negate": false},   // I want equal separate from in because then we can think about CompassQL's syntax for enumerating the drill down value.  
  {"field": "foo", "in": ["a", "b", "c"], "negate": false},
  {"field": "bar", "min": 5, "max": 10, "negate": false}
];
```

For backward compatability, we *might* want to support Vega Expression `string` or something like this:

```
// full expression version 
"filter": [
  {"test": "datum.foo == 'Kanit'"}
]
```

I don't have a strong opinion whether we should support the `"test": __expr__` version though.

kanitw:
Fix #1394

Setting `data.transform.filter` to
- `{"field": "car_color", "equal": "red"}` checks if the `car_color` field's value is equal to `"red"`.
- `{"field": "car_color", "in": ["red", "yellow"]}` checks if the `car_color` field's value is `"red"` or `"yellow"`.
-  `{"field": "x", "range": [0, 5]}` checks if the `x` field's value is in range `[0,5]` (0 ≤ x ≤ 5).

(`data.transform.filter` can also be an array of filter object of the same form or Vega Expression)

We need this to add filter UI to VLUI + There is no conflict between this and the layer branch.   (Maybe except only tsconfig.json)

Note: 100% Test Coverage for all added lines

@domoritz @arvind @jheer -- any thoughts about syntax before we merge.  (We need this to support filter in polestar for @yhoonkim's study)


kanitw:
@domoritz  If you can't review, please re-assign to @ZeningQu

Issue #1466, opened 2016-07-12
Extend Filter to Support moreThan, lessThan and negation


kanitw:
From discussion on slack, we'll add: 

- [x] `negate`
- [x] `lessThan`, `greaterThan`
- [x] `lessThanOrEqual`, `greaterThanOrEqual`

(`greaterThanOrEqual` is damn long -- not sure if `moreThanOrEqual` or just `gte` is more appropriate.)

PR #1467, opened 2016-07-13
Extends Filter Object to Support gt,lt,gte,lte, and negate. #1466

yhoonkim:
Issue : #1466

As we talked before in slack, I changed lessThan, lessThanOrEqual, greaterThan, greaterThanOrEqual to lt, lte, gt, gte respectively.

And I excluded some cases .
- `range` with `lt`/`lte`/`gt`/`gte`.  
- `lt` with `lte`
- `gt` with `gte`


src/filter.ts
```diff
@@ -32,11 +37,24 @@ export interface RangeFilter {
    * @maxItems 2
    * @minItems 2
    */
-  range: any;
+  range?: any;
+  lt?: number;
```
kanitw: Can you help copy description from the docs to here.  (Altair folks uses description in the JSON schema as part of their iPython autocomplete as well.)

src/filter.ts
```diff
@@ -32,11 +37,24 @@ export interface RangeFilter {
...
+
+  negate?: boolean;
 }
 
 export function isRangeFilter(filter: any): filter is RangeFilter {
-  return filter && !!filter.field && isArray(filter.range) && filter.range.length === 2;
+  if (filter && !!filter.field) {
+    if (isArray(filter.range) && filter.range.length === 2) {
+      return true;
+    } else if (XOR(filter.gt !==undefined, filter.gte!==undefined) || XOR(filter.lt!==undefined, filter.lte!==undefined) ) {
```
kanitw: I don't think we need XOR here.    If users want to stupidly combine  `gt` and `gte`.  That should mean we `&&` the two?  

Pro-tip:    `a!==b` is the way to do XOR if you need in other occasion.

site/docs/transform.md
```diff
@@ -51,25 +51,28 @@ This example use `calculate` to derive a new field, then `filter` data based on
...
+For a filter object, a `field` must be provided with one of the filter operators (`equal`, `in`, `range`, `gt`/`gte`, `lt`/`lte`). Here, `gt`/`gte` and `lt`/`lte` can be combinated. The following table describes each of these properties.
 
 | Property      | Type          | Description    |
 | :------------ |:-------------:| :------------- |
 | field         | String        | Field to be filtered. |
-| equal         | String &#124; Number &#124; Boolean | Value that the `field`'s value should be equal to. |
-| range         | Number[]      | Array of length describing (inclusive) minimum and maximum values for the `field`'s value to be included in the filtered data. |
+| equal         | String | Number | Boolean | Value that the `field`'s value should be equal to. |
+| range          | Number[]      | Array of length describing (inclusive) minimum and maximum values for the `field`'s value to be included in the filtered data. |
+| gt/gte/lt/lte | Number | Value that the `field`'s value should be **g**reater **t**han / **g**reater **t**han or **e**qual / **l**ess **t**han / **l**ess **t**han or **e**qual to. |
```
kanitw: missing `negate`

src/filter.ts
kanitw: Plus, adding this condition will make the logic fail as combination of `gt` and `gte` will make `getFilterExpression` return just the `filter` on [this line](https://github.com/vega/vega-lite/pull/1467/files#diff-74ca0355251fd3e28659aeb89ca5ad34R40).

src/compile/data/filter.ts
```diff
@@ -13,15 +13,36 @@ export namespace filter {
...
+      filterString = field(filter, {datum: true}) + '===' + s(filter.equal);
     } else if (isInFilter(filter)) {
-      return 'indexof(' + s(filter.in) + ', ' + field(filter, {datum: true}) + ') !== -1';
+      filterString = 'indexof(' + s(filter.in) + ', ' + field(filter, {datum: true}) + ') !== -1';
     } else if (isRangeFilter(filter)) {
-      return 'inrange(' + field(filter, {datum: true}) + ', ' + s(filter.range[0]) + ', ' + s(filter.range[1]) + ')';
+      if (!!filter.range) {
+        filterString = 'inrange(' + field(filter, {datum: true}) + ', ' + s(filter.range[0]) + ', ' + s(filter.range[1]) + ')';
+      } else {
+        if (filter.gt !== undefined) {
```
kanitw: As commented below.  You might want to make an array and join in this case :)

kanitw:
I refactor code a bit, but this is awesome!  Gonna merge.  Thanks.

PR #1474, opened 2016-07-17
Support TimeUnit Filter with date time object and support unbounded range filter

### LINKED ISSUE #1463: Extend Filter Object to support `timeUnit` and `bin`
Currently Filter Object as implemented in https://github.com/vega/vega-lite/pull/1461 do not support `timeUnit` but we need to support that. 

Similar for `bin`, but probably far less important as we can always filter binned field by its pre-binned values.

A relevant issue would be supporting `aggregate`, but that's basically #117.

kanitw:
Fix #1463 (I don't think we ever need to support `bin`, so we can close it)

- Add `DateTime` object and `vl.datetime.dateTimeExpr` method
- Refactor timeunitdomain.ts to use `dateTimeExpr` instead of the old weird `refOnly` flag in `fieldExpr`
- add support for unbounded range filter

@yhoonkim since you're the most familiar with filter, probably reasonable to have you review it? 

cc: @ZeningQu  This is probably enough for adding filter support to VLUI


site/docs/transform.md
```diff
@@ -51,25 +51,44 @@ This example use `calculate` to derive a new field, then `filter` data based on
...
 | :------------ |:-------------:| :------------- |
 | field         | String        | Field to be filtered. |
-| equal         | String | Number | Boolean | Value that the `field`'s value should be equal to. |
-| range          | Number[]      | Array of length describing (inclusive) minimum and maximum values for the `field`'s value to be included in the filtered data. |
-| in         | Array         | A set of values that the `field`'s value should be a member of, for a data item included in the filtered data. |
+| equal         | String | Number | DateTime | Boolean | Value that the `field`'s value should be equal to. |
+| range          | Number[] | DateTime[] | Array of length describing (inclusive) minimum and maximum values for the `field`'s value to be included in the filtered data. |
+| in         | String[] | Number[] | DateTime[] | A set of values that the `field`'s value should be a member of, for a data item included in the filtered data. |
+
+{:#datetime}
```
yhoonkim: `{:#datetime}` seems not working. Maybe we need to use just a longer version.

**Suggestion**
Remove `{:#datetime}`
Change `[DateTime definition object](#datetime)` to `[DateTime definition object](#date-time-definition-object)`

kanitw: Free to help fix this because I'm on a road :)

Issue #1476, opened 2016-07-25
vl.spec.fieldDefs() returns empty array for layered line and area charts


ZeningQu:
For the layered line chart and area chart specs:
https://github.com/vega/vega-lite/blob/master/examples/specs/overlay_line_full.json
https://github.com/vega/vega-lite/blob/master/examples/specs/overlay_area_full.json

`vl.spec.fieldDefs()` returns `[]` instead of a complete list of `fieldDefs` that's used for encoding. 

This is blocking https://github.com/vega/vega-tooltip/issues/47

kanitw:
Feel free to send a PR to fix this.  This should be straightforward.

kanitw:
@ZeningQu  I remember you already started a bit on this?  If so, do you mind submitting the PR so someone can take over (or even better finish it).  Thanks

PR #1481, opened 2016-07-28
[WIP] Get fieldDefs for for LayerSpec, FacetSpec, ExtendedFacetSpec

ZeningQu:
- [x] `vl.spec.fieldDefs()` returns all non-duplicate `fieldDefs` in all layers in a LayerSpec (Fix https://github.com/vega/vega-lite/issues/1476) 
- [x] `vl.spec.fieldDefs()` returns all non-duplicate `fieldDefs` from `facet` and `spec` in a FacetSpec
- [x] `vl.spec.fieldDefs()` returns all non-duplicate `fieldDefs` from `facet` and `spec` in ExtendedFacetSpec
- [x] Add unit test

### DIFF
--- a/src/spec.ts
+++ b/src/spec.ts
@@ -340,7 +340,14 @@ export function normalizeOverlay(spec: UnitSpec, overlayWithPoint: boolean, over
 /* Accumulate non-duplicate fieldDefs in a dictionary */
 function accumulate(dict: any, fieldDefs: FieldDef[]): any {
   fieldDefs.forEach(function(fieldDef) {
-    let key = hash(fieldDef);
+    // Consider only pure fieldDef properties (ignoring scale, axis, legend)
+    var pureFieldDef = ['field', 'type', 'value', 'timeUnit', 'bin', 'aggregate'].reduce((f, key) => {
+      if (fieldDef[key] !== undefined) {
+        f[key] = fieldDef[key];
+      }
+      return f;
+    }, {});
+    let key = hash(pureFieldDef);
     dict[key] = dict[key] || fieldDef;
   });
   return dict;
--- a/src/spec.ts
+++ b/src/spec.ts
@@ -10,7 +10,7 @@ import {stack} from './stack';
 import {Transform} from './transform';
 import {ROW, COLUMN, X, Y, X2, Y2} from './channel';
 import * as vlEncoding from './encoding';
-import {contains, duplicate, extend, keys, omit, pick} from './util';
+import {contains, duplicate, extend, hash, keys, omit, pick, vals} from './util';
 
 export interface BaseSpec {
   /**
@@ -119,7 +119,7 @@ export type Spec = UnitSpec | FacetSpec | LayerSpec;
 
 /* Custom type guards */
 
-export function isFacetSpec(spec: ExtendedSpec): spec is FacetSpec {
+export function isSomeFacetSpec(spec: ExtendedSpec | ExtendedFacetSpec): spec is FacetSpec | ExtendedFacetSpec {
   return spec['facet'] !== undefined;
 }
 
@@ -146,7 +146,7 @@ export function isSomeUnitSpec(spec: ExtendedSpec): spec is ExtendedUnitSpec | U
   return spec['mark'] !== undefined;
 }
 
-export function isLayerSpec(spec: ExtendedSpec): spec is LayerSpec {
+export function isLayerSpec(spec: ExtendedSpec | ExtendedFacetSpec): spec is LayerSpec {
   return spec['layers'] !== undefined;
 }
 
@@ -337,9 +337,34 @@ export function normalizeOverlay(spec: UnitSpec, overlayWithPoint: boolean, over
 
 // TODO: add vl.spec.validate & move stuff from vl.validate to here
 
-export function fieldDefs(spec: ExtendedUnitSpec): FieldDef[] {
-  // TODO: refactor this once we have composition
-  return vlEncoding.fieldDefs(spec.encoding);
+/* Accumulate non-duplicate fieldDefs in a dictionary */
+function accumulate(dict: any, fieldDefs: FieldDef[]): any {
+  fieldDefs.forEach(function(fieldDef) {
+    let key = hash(fieldDef);
+    dict[key] = dict[key] || fieldDef;
+  });
+  return dict;
+}
+
+/* Recursively get fieldDefs from a spec, returns a dictionary of fieldDefs */
+function fieldDefIndex(spec: ExtendedSpec | ExtendedFacetSpec, dict: any = {}): any {
+  // TODO: Support repeat and concat
+  if (isLayerSpec(spec)) {
+    spec.layers.forEach(function(layer) {
+      accumulate(dict, vlEncoding.fieldDefs(layer.encoding));
+    });
+  } else if (isSomeFacetSpec(spec)) {
+    accumulate(dict, vlEncoding.fieldDefs(spec.facet));
+    fieldDefIndex(spec.spec, dict);
+  } else { // Unit Spec
+    accumulate(dict, vlEncoding.fieldDefs(spec.encoding));
+  }
+  return dict;
+}
+
+/* Returns all non-duplicate fieldDefs in a spec in a flat array */
+export function fieldDefs(spec: ExtendedSpec | ExtendedFacetSpec): FieldDef[] {
+  return vals(fieldDefIndex(spec));
 };
 
 export function isStacked(spec: ExtendedUnitSpec): boolean {
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -13,11 +13,11 @@ import {LayerModel} from './layer';
 import {Model} from './model';
 import {template as timeUnitTemplate} from '../timeunit';
 import {UnitModel} from './unit';
-import {Spec, isUnitSpec, isFacetSpec, isLayerSpec} from '../spec';
+import {Spec, isUnitSpec, isSomeFacetSpec, isLayerSpec} from '../spec';
 
 
 export function buildModel(spec: Spec, parent: Model, parentGivenName: string): Model {
-  if (isFacetSpec(spec)) {
+  if (isSomeFacetSpec(spec)) {
     return new FacetModel(spec, parent, parentGivenName);
   }
 
--- a/test/spec.test.ts
+++ b/test/spec.test.ts
@@ -287,6 +287,38 @@ describe('fieldDefs()', function() {
     ]);
   });
 
+  it('should get all non-duplicate fieldDefs from all layers in a LayerSpec (merging duplicate fields with different scale types)', function() {
+    const layerSpec: any = {
+      "data": {"url": "data/stocks.csv","format": {"type": "csv"}},
+      "transform": {"filter": "datum.symbol==='GOOG'"},
+      "layers": [
+        {
+          "description": "Google's stock price over time.",
+          "mark": "line",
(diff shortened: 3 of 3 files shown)


src/spec.ts
```diff
@@ -319,9 +319,9 @@ export function alwaysNoOcclusion(spec: ExtendedUnitSpec): boolean {
   return vlEncoding.isAggregate(spec.encoding);
 }
 
-export function fieldDefs(spec: ExtendedUnitSpec): FieldDef[] {
+export function fieldDefs(spec: ExtendedUnitSpec | LayerSpec): FieldDef[] {
   // TODO: refactor this once we have composition
-  return vlEncoding.fieldDefs(spec.encoding);
+  return isLayerSpec(spec) ? vlEncoding.fieldDefs(spec.layers[0].encoding) : vlEncoding.fieldDefs(spec.encoding);
```
kanitw: This is definitely incorrect

kanitw:
> the multiple layers can have different fieldDefs in some cases? In that
case, should fieldDef return a flat array containing fieldDefs from all
layers?

Yes.

> ['x', 'y', 'color']

This is a fieldDefs method, not channels.

src/fielddef.ts
```diff
@@ -60,6 +60,17 @@ export interface FieldDef {
   title?: string;
 }
 
+/* Checks if two FieldDef are identical */
+export function identical(f1: FieldDef, f2: FieldDef): boolean {
```
kanitw: This is not scalable way to implement this.   You can utilize `json-stable-stringify` to compare fieldDef and it will remain correct no matter how we modify fieldDef.

src/spec.ts
```diff
@@ -319,9 +319,31 @@ export function alwaysNoOcclusion(spec: ExtendedUnitSpec): boolean {
...
+function hasFieldDef(fieldDefs: FieldDef[], fieldDef: FieldDef): boolean {
+  for (let i = 0; i < fieldDefs.length; i++) {
+    if (identical(fieldDefs[i], fieldDef)) {
+      return true;
+    }
+  }
+  return false;
+}
+
 export function fieldDefs(spec: ExtendedUnitSpec | LayerSpec): FieldDef[] {
```
kanitw: Missing FacetSpec

src/spec.ts
```diff
@@ -319,9 +319,31 @@ export function alwaysNoOcclusion(spec: ExtendedUnitSpec): boolean {
...
   // TODO: refactor this once we have composition
-  return isLayerSpec(spec) ? vlEncoding.fieldDefs(spec.layers[0].encoding) : vlEncoding.fieldDefs(spec.encoding);
+  if (isLayerSpec(spec)) {
+    let fieldDefsAllLayers: FieldDef[] = [];
```
kanitw: Instead of O(n) check why don't you make a dictionary of FieldDef using stable stringify to produce key?

src/fielddef.ts
kanitw: Hmm. If you want to ignore `scale`,  `axis`, `legend`, `sort` -- this might become more sensible

src/spec.ts
ZeningQu: Should I add `ExtendedFacetSpec` too?

ZeningQu: Out of curiosity, is `FacetSpec` being used in the examples right now? I don't see "facet" in the example specs

kanitw:
Still missing test

src/spec.ts
```diff
@@ -319,9 +319,42 @@ export function alwaysNoOcclusion(spec: ExtendedUnitSpec): boolean {
...
+  return fieldDefs;
+}
+
+export function fieldDefs(spec: ExtendedUnitSpec | LayerSpec | FacetSpec): FieldDef[] {
   // TODO: refactor this once we have composition
-  return vlEncoding.fieldDefs(spec.encoding);
+  if (isLayerSpec(spec)) {
+    return fieldDefsOfAllLayers(spec);
+  } else if (isFacetSpec(spec)) {
+    let facet = spec.spec;
```
kanitw: this chunk for facet looks totally random.   why is `facet = spec.spec`   spec.facet isn't considered at all?

src/spec.ts
kanitw: Are you kidding?

src/spec.ts
```diff
@@ -319,9 +319,42 @@ export function alwaysNoOcclusion(spec: ExtendedUnitSpec): boolean {
...
+function hasFieldDef(fieldDefs: FieldDef[], fieldDef: FieldDef): boolean {
+  for (let i = 0; i < fieldDefs.length; i++) {
+    if (identical(fieldDefs[i], fieldDef)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+function fieldDefsOfAllLayers(spec: LayerSpec): FieldDef[] {
```
kanitw: I don't like this method name and believe that with more efficient implementation, we don't really need this method.

src/fielddef.ts
ZeningQu: Done. I am using `hash` from `util`

src/spec.ts
ZeningQu: Done

src/spec.ts
ZeningQu: Removed

ZeningQu:
It seems to me we should have 
`encoding.fieldDefs(encoding: Encoding): FieldDef[]`
`facet.fieldDefs(facet: Facet): FieldDef[]`

which should be called inside of 
`spec.fieldDefs(spec: ExtendedUnitSpec | LayerSpec | FacetSpec | ExtendedFacetSpec): FieldDef[]`

kanitw:
> It seems to me we should have 
encoding.fieldDefs(encoding: Encoding): FieldDef[]
facet.fieldDefs(facet: Facet): FieldDef[]

Both are the same.  They are both channel-fieldDef mapping.  for facet, you can just call encoding.fieldDefs (or add alias to the same method)

src/spec.ts
```diff
@@ -101,8 +101,28 @@ export type Spec = UnitSpec | FacetSpec | LayerSpec;
 
 /* Custom type guards */
 
-export function isFacetSpec(spec: ExtendedSpec): spec is FacetSpec {
-  return spec['facet'] !== undefined;
+export function isFacetSpec(spec: ExtendedSpec | ExtendedFacetSpec): spec is FacetSpec {
+  if (spec['facet'] !== undefined) {
```
kanitw: I don't understand why you are modifying this method.   It seems irrelevant to the PR and will become incorrect in the future.

src/spec.ts
```diff
@@ -101,8 +101,28 @@ export type Spec = UnitSpec | FacetSpec | LayerSpec;
 
 /* Custom type guards */
 
-export function isFacetSpec(spec: ExtendedSpec): spec is FacetSpec {
-  return spec['facet'] !== undefined;
+export function isFacetSpec(spec: ExtendedSpec | ExtendedFacetSpec): spec is FacetSpec {
```
kanitw: Distinction between facet spec and extended facet spec seems unnecessary.  Your code path for both cases are the same.  Plus, this change will become incorrect in the future.

src/spec.ts
```diff
@@ -128,7 +148,7 @@ export function isSomeUnitSpec(spec: ExtendedSpec): spec is ExtendedUnitSpec | U
   return spec['mark'] !== undefined;
 }
 
-export function isLayerSpec(spec: ExtendedSpec): spec is LayerSpec {
+export function isLayerSpec(spec: ExtendedSpec | ExtendedFacetSpec): spec is LayerSpec {
```
kanitw: Don't add this | here.

src/spec.ts
```diff
@@ -319,9 +339,44 @@ export function alwaysNoOcclusion(spec: ExtendedUnitSpec): boolean {
...
+    let key = hash(fieldDef);
+    if (!dict[key]) {
+      dict[key] = fieldDef;
+    }
+  });
+  return dict;
+}
+
+/* Recursively get fieldDefs from a spec, returns a dictionary of fieldDefs */
+function getFieldDefsRecursively(dict: any, spec: ExtendedSpec | ExtendedFacetSpec): any {
```
kanitw: Please remove  `| ExtendedFacetSpec`

src/spec.ts
```diff
@@ -319,9 +339,44 @@ export function alwaysNoOcclusion(spec: ExtendedUnitSpec): boolean {
   return vlEncoding.isAggregate(spec.encoding);
 }
 
-export function fieldDefs(spec: ExtendedUnitSpec): FieldDef[] {
+/* Accumulate non-duplicate fieldDefs in a dictionary */
```
kanitw: I'm sure there is a way to write these 3 methods in a more compact way as just one method.  
Right now the code seems unnecessarily verbose.  

Hint:

Add  `dict: Dict<FieldDef> = {}` as second parameter to `fieldDefs` and you can just reuse it.

kanitw: Actually you need two.   But make `fieldDefs` one line return that calls `fieldDefIndex` method which basically is the version that returns dictionary (and takes existing `dict: Dict<FieldDef> = {}` as second parameter)

kanitw:
Plus, test is FAILING due to tslint update.  Next time, when you see test failing, please make test pass even if your change does not cause the test to fail.

kanitw:
Plus, test is FAILING due to tslint update.  I just fix it in master, so please rebase to master and it should pass.   Next time, when you see test failing, please make test pass even if your change does not cause the test to fail.

src/spec.ts
ZeningQu: Ok, so I will write something like `isSomeFacetSpec(spec: ExtendedSpec): spec is FacetSpec | ExtendedFacetSpec` ?

src/spec.ts
ZeningQu: I tried to add `ExtendedFacetSpec` to `ExtendedSpec` but that will break the spec `normalize()` function because they are not expecting `ExtendedFacetSpec`...

src/spec.ts
ZeningQu: Done

src/spec.ts
```diff
@@ -376,6 +353,7 @@ export function fieldDefs(spec: ExtendedSpec | ExtendedFacetSpec): FieldDef[] {
   // TODO: refactor this once we have composition
   let fieldDefsDict = {};
   getFieldDefsRecursively(fieldDefsDict, spec);
+  console.log('fieldDefs:', vals(fieldDefsDict).length);
```
kanitw: log?

kanitw:
@kanitw's task
- [ ] Check if this is only required for tooltip?  (Do we use this somewhere else?)

src/spec.ts
```diff
@@ -337,9 +337,34 @@ export function normalizeOverlay(spec: UnitSpec, overlayWithPoint: boolean, over
 
 // TODO: add vl.spec.validate & move stuff from vl.validate to here
 
-export function fieldDefs(spec: ExtendedUnitSpec): FieldDef[] {
-  // TODO: refactor this once we have composition
-  return vlEncoding.fieldDefs(spec.encoding);
+/* Accumulate non-duplicate fieldDefs in a dictionary */
+function accumulate(dict: any, fieldDefs: FieldDef[]): any {
+  fieldDefs.forEach(function(fieldDef) {
+    let key = hash(fieldDef);
```
kanitw: What if we use the same field twice -- once as linear scale and once as log scale.  This seems to produce duplicates?

PR #1484, opened 2016-07-29
Correct scale.domain schema: it can only be a number[] or string[], but not string.

felixcodes:
Correct `scale.domain` schema: it can only be a number[] or string[], but not string.

### DIFF
--- a/src/scale.ts
+++ b/src/scale.ts
@@ -108,9 +108,9 @@ export const defaultFacetScaleConfig: FacetScaleConfig = {
 export interface Scale {
   type?: ScaleType;
   /**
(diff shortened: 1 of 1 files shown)


kanitw:
Test is FAILING due to tslint update.  I have just fixed it for you. 

However, this kind of problem is a **team responsibility**.  So when you see test failing, please try to make test pass even if your change does not cause the test to fail.   (For example, submit a PR to make test pass first.)

kanitw:
@felixcodes  Please see the updated description for example commit message.  Next time try to follow this format.  (Imperative sentence starting with verb saying what this commit does -- ignoring the implied subject "This commit")

PR #1490, opened 2016-08-04
Add timeunit date conversion method

RileyChang:
Adds the `timeunit.convert` method, which takes in a date and timeunit and returns a date that only has the specified timeunits from the provided date. Example:

convert('yearmonth', '2000-12-04 07:58:14') -> '2000-12-00 00:00:00'

### DIFF
--- a/src/timeunit.ts
+++ b/src/timeunit.ts
@@ -57,7 +57,7 @@ export function isSingleTimeUnit(timeUnit: TimeUnit) {
  * i.e. ('yearmonth', '2000-12-04 07:58:14') -> '2000-12-01 00:00:00'
  * Note: the base date is Jan 01 1900 00:00:00
  */
-export function convert(unit: TimeUnit, date: Date, utc: boolean): Date {
+export function convert(unit: TimeUnit, date: Date): Date {
   const result: Date = new Date(0, 0, 1, 0, 0, 0, 0); // start with uniform date
   SINGLE_TIMEUNITS.forEach(function(singleUnit) {
     if (containsTimeUnit(unit, singleUnit)) {
--- a/src/timeunit.ts
+++ b/src/timeunit.ts
@@ -63,9 +63,13 @@ export function convert(unit: TimeUnit, date: Date): Date {
     if (containsTimeUnit(unit, singleUnit)) {
       switch (singleUnit) {
         case TimeUnit.DAY:
-          throw 'Cannot convert to TimeUnits containing \'day\'';
+          throw new Error('Cannot convert to TimeUnits containing \'day\'');
         case TimeUnit.YEAR:
           result.setFullYear(date.getFullYear());
+          break;
+        case TimeUnit.QUARTER:
+          // indicate quarter by setting month to be the first of the quarter i.e. may (4) -> april (3)
+          result.setMonth((Math.floor(date.getMonth() / 3)) * 3);
           break;
         case TimeUnit.MONTH:
           result.setMonth(date.getMonth());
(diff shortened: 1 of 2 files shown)


src/timeunit.ts
```diff
@@ -52,6 +52,54 @@ export function isSingleTimeUnit(timeUnit: TimeUnit) {
...
+    // not sure how to handle this unit. Just put month (since you can infer quarter from month)?
+  } else {
+    SINGLE_TIMEUNITS.forEach(function(singleUnit) {
+      if (containsTimeUnit(unit, singleUnit)) {
+        // can ignore TimeUnit.DAY
+        switch (singleUnit) {
+          case TimeUnit.YEAR:
+            result.setFullYear(date.getFullYear());
+            break;
+          case TimeUnit.MONTH:
```
kanitw: Probably need to handle QUARTER somehow

src/timeunit.ts
```diff
@@ -52,6 +52,54 @@ export function isSingleTimeUnit(timeUnit: TimeUnit) {
...
+ */
+export function convert(unit: TimeUnit, date: Date, utc: boolean): Date {
+  const result: Date = new Date(0, 0, 0, 0, 0, 0, 0); // start with uniform date
+  // handle some units individually
+  if (unit === TimeUnit.QUARTERMONTH) {
+    // not sure how to handle this unit. Just put month (since you can infer quarter from month)?
+  } else {
+    SINGLE_TIMEUNITS.forEach(function(singleUnit) {
+      if (containsTimeUnit(unit, singleUnit)) {
+        // can ignore TimeUnit.DAY
```
kanitw: add a case on top to throw error if day is input?  (We will just not support convert)

RileyChang:
@kanitw Handled quarter unit. Still need to assess if we need the UTC flag.

src/timeunit.ts
```diff
@@ -52,6 +52,55 @@ export function isSingleTimeUnit(timeUnit: TimeUnit) {
...
+        case TimeUnit.MINUTES:
+          result.setMinutes(date.getMinutes());
+          break;
+        case TimeUnit.SECONDS:
+          result.setSeconds(date.getSeconds());
+          break;
+        case TimeUnit.MILLISECONDS:
+          result.setMilliseconds(date.getMilliseconds());
+          break;
+        case TimeUnit.QUARTER:
```
kanitw: move this up above MONTH

src/timeunit.ts
```diff
@@ -52,6 +52,55 @@ export function isSingleTimeUnit(timeUnit: TimeUnit) {
...
+ * i.e. ('yearmonth', '2000-12-04 07:58:14') -> '2000-12-01 00:00:00'
+ * Note: the base date is Jan 01 1900 00:00:00
+ */
+export function convert(unit: TimeUnit, date: Date): Date {
+  const result: Date = new Date(0, 0, 1, 0, 0, 0, 0); // start with uniform date
+  SINGLE_TIMEUNITS.forEach(function(singleUnit) {
+    if (containsTimeUnit(unit, singleUnit)) {
+      switch (singleUnit) {
+        case TimeUnit.DAY:
+          throw 'Cannot convert to TimeUnits containing \'day\'';
```
kanitw: throw new Error("...")

src/timeunit.ts
```diff
@@ -52,6 +52,55 @@ export function isSingleTimeUnit(timeUnit: TimeUnit) {
...
+          result.setMilliseconds(date.getMilliseconds());
+          break;
+        case TimeUnit.QUARTER:
+          result.setMonth((Math.floor(date.getMonth() / 3)) * 3);
+          break;
+      }
+    }
+  });
+
+  // handle weird 'quartermonth' and 'yearquartermonth' cases (should just keep month, ignore quarter)
```
kanitw: Do you really need this hack below here.   MONTH comes after QUARTER in the `SINGLE_TIMEUNITS`, which means it will call `setMonth` with the right value after setting QUARTER anyway?

Issue #1449, opened 2016-06-30
Allow field to be an object?


ZeningQu:
Currently Vega-Lite doesn't allow field to be an object. For example, `"field": "obj.a"` doesn't work.

``` json
{
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"obj": {"a": "A","b": 28}},
      {"obj": {"a": "A","b": 28}},
      {"obj": {"a": "B","b": 55}},
      {"obj": {"a": "C","b": 43}},
      {"obj": {"a": "D","b": 91}},
      {"obj": {"a": "E","b": 81}},
      {"obj": {"a": "F","b": 53}},
      {"obj": {"a": "G","b": 19}},
      {"obj": {"a": "H","b": 87}},
      {"obj": {"a": "I","b": 52}}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "obj.a","type": "ordinal"},
    "y": {"field": "obj.b","type": "quantitative"}
  }
}
```

It would be nice to have fields that are objects (?). 

In Vega field can be an object (e.g., `"field": "unemp.rate"`). Vega also allows definition of multiple data fields (e.g., `"data"` can be an array).

``` json
{
  "width": 960,
  "height": 500,
  "data": [
    {
      "name": "unemp",
      "url": "data/unemployment.tsv",
      "format": {"type": "tsv", "parse": "auto"}
    },
    {
      "name": "counties",
      "url": "data/us-10m.json",
      "format": {"type": "topojson", "feature": "counties"},
      "transform": [
        { "type": "geopath", "projection": "albersUsa" },
        { "type": "lookup", "on": "unemp", "onKey": "id",
          "keys": ["id"], "as": ["unemp"] },
        { "type": "filter",
          "test": "datum.layout_path!=null && datum.unemp!=null" }
      ]
    }
  ],
  "scales": [
    {
      "name": "color",
      "type": "quantize",
      "domain": [0, 0.15],
      "range": ["#f7fbff", "#deebf7", "#c6dbef", "#9ecae1", "#6baed6",
                "#4292c6", "#2171b5", "#08519c", "#08306b"]
    }
  ],
  "marks": [
    {
      "type": "path",
      "from": {"data": "counties"},
      "properties": {
        "enter": { "path": {"field": "layout_path"} },
        "update": { "fill": {"scale": "color", "field": "unemp.rate"} },
        "hover": { "fill": {"value": "red"} }
      }
    }
  ]
[...]

Issue #778, opened 2015-11-23
Make sure we fully support field names with spaces


domoritz:
Right now date fields with names with spaces do not work because we generate code like `data.Foo Bar != null`. Still have to confirm that vega fully supports spaces in names.

domoritz:
Duplicate of https://github.com/vega/vega-lite/issues/358

domoritz:
https://mothereff.in/js-variables

domoritz:
We should just generate array access notation `data['Foo Bar']` to get around any issues. We could of course test whether the name is a valid identifier. As a compromise, we can leave names with only ascii character and a few simple modifiers such as `_` as they are.

domoritz:
Even more relevant https://mothereff.in/js-properties

domoritz:
Ahh, this is really hard. There are so many corner cases.

domoritz:
Some code in https://github.com/vega/vega-lite/pull/779

jakevdp:
It looks like dashes in column names also cause issues (reported in https://github.com/ellisonbg/altair/issues/169)

```
{
  "mark": "bar",
  "data": {
    "values": [
      {"X": "0-12", "Y-10": 17.81},
      {"X": "12-36", "Y-10": 22.35},
      {"X": "36-60", "Y-10": 25.13},
      {"X": ">60", "Y-10": 16.76}
    ]
  },
  "encoding": {
    "y": {"type": "quantitative", "field": "Y-10"},
    "x": {"type": "nominal", "field": "X"}
  }
}
```

willium:
PR that should fix this https://github.com/vega/vega-lite/pull/1492

PR #1492, opened 2016-08-06
Allow spaces, dashes, and dots in field names

willium:
# Fix of #778
This PR should:
- [x] allow for spaces in field names
    - [x] rewrite `field()` to use string interpolation
    - [x] rewrite tests using old `datum.` prefix
    - [x] add an example with fields that contain spaces
    - [x] for better style, add _ separator for prefixes and suffixes in field function instead of options

- [x] Passes Tests and Linter

# Next Steps:
Added new issues: #1493, #1494, #1495 per sub-issues discovered during the fix for #778.

Also see: #1449


src/compile/mark/mark.ts
```diff
@@ -239,8 +239,8 @@ function getStackByFields(model: UnitModel) {
         const fieldDef: FieldDef = channelEncoding;
         const scale = model.scale(channel);
         fields.push(field(fieldDef, {
-          binSuffix: scale && scale.type === ScaleType.ORDINAL ? '_range' : '_start'
-        }));
+          binSuffix: scale && scale.type === ScaleType.ORDINAL ? 'range' : 'start'
+        }) || 'undefined');
```
willium: @kanitw @domoritz without this addition, a test fails because the vega renders to 
`groupby: [null]`
when it used to be
`groupby: ["undefined"]`

So I did this... I'm not really sure why getting rid of the _ here and adding it in field causes the value to become null. So... this passes all tests, but it might not be foolproof because I don't know why this happened. Please check it out before merging.

src/compile/mark/mark.ts
```diff
@@ -239,8 +239,8 @@ function getStackByFields(model: UnitModel) {
         const fieldDef: FieldDef = channelEncoding;
         const scale = model.scale(channel);
         fields.push(field(fieldDef, {
-          binSuffix: scale && scale.type === ScaleType.ORDINAL ? '_range' : '_start'
-        }));
+          binSuffix: scale && scale.type === ScaleType.ORDINAL ? 'range' : 'start'
```
willium: Also -- I don't understand why this check of type is here, because a **very** similar check is done in field() (within fielddef.ts). 

I previous attempted to get rid of it here and get it to be triggered there, but it caused tests to fail. Rather than investigate, I just left it. Factoring this out into FieldDef is probably worth considering.

Furthermore, I wasn't able to find a single example of when _both_ binSuffix and suffix were being used for the same fieldDef, which leads me to believe that we should just have one? I tried implementing this too, and it broke tests because if you pass in suffix it gets set immediately in field(), while binSuffix is only set when `!opt.nofn && fieldDef.bin`. Rather than add a new flag, I left in binSuffix. Wondering why this was done this way initially though.

src/compile/data/stackscale.ts
```diff
@@ -26,7 +26,7 @@ export namespace stackScale {
         transform: [{
           type: 'aggregate',
           // group by channel and other facets
-          groupby: [model.field(groupbyChannel)],
+          groupby: [model.field(groupbyChannel) || 'undefined'],
```
kanitw: why adding `|| undefined` here?

willium: https://github.com/vega/vega-lite/pull/1492#discussion_r73793025

src/compile/mark/mark.ts
kanitw: Seems like both are wrong?    If it's undefined, I guess it should not even produce groupby or may be produce empty array?

src/fielddef.ts
```diff
@@ -95,52 +95,71 @@ export interface OrderChannelDef extends FieldDef {
...
 
-
-
 export interface FieldRefOption {
   /** exclude bin, aggregate, timeUnit */
   nofn?: boolean;
   /** exclude aggregation function */
   noAggregate?: boolean;
-  /** include 'datum.' */
+  /** include 'datum["$1"]' */
```
kanitw: /** Wrap the field inside datum[...] */  ?

src/timeunit.ts
```diff
@@ -294,7 +294,8 @@ export function template(timeUnit: TimeUnit, field: string, shortTimeLabels: boo
    const template = '{{' + field + ' | time:\'' + out.join(' ') + '\'}}';
 
    // FIXME: Remove this RegExp Hack!!!
-   return template.replace(new RegExp('{{' + field + ' \\| time:\'\'}}', 'g'), '');
+   const escapedField = field.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, '\\$1');
```
kanitw: I don't understand this line.  Can you add comment explaining what it does? 

Or even better, get rid of it.  I actually did not approve this RegExp first time I see it but it was already merged in master.  Thus I added this FIXME.

src/compile/mark/mark.ts
willium: Agreed, but felt this was out of scope because I don't really _get_ groupby and felt whomever implemented it would be better to address the inclusion. 

Similarly, I extended a bad pattern in timeunit.ts, with the RegExp. I started to take a crack at fixing it, but didn't think it made sense within this PR. Happy to create news issues for these if you agree.

Those seem like more 'nice to have' and less 'let's get this feature done ASAP' -- up to you!

src/timeunit.ts
willium: Just addressed this in last comment :) we're on the same wavelength. This is necessary because 'field' now has square brackets which need to be escaped in Regex (the dot notation did not need escaping).

I'll open an issue to rewrite this. Didn't feel like it made sense within this PR

willium: Also, I'm having some difficulty following the timeunit method. I sort of see that the way it's being built up sometimes adds these empty templates... but wasn't really sure why it was initially written this way. I can take a look soon :)

kanitw:
I am not sure if we should allow dot in field names, because in that case it's unclear if you want to access nested property or really want field name with dots.

willium:
@kanitw Agreed. How should I address this ambiguity? Happy to address.

OTOH seems like maybe merging this in and then revising the whole field construction per #1449 might be a good path forward?

willium:
Since a field can be prefixed by multiple functions, it makes some sense to have opt.fn be a HashSet and then order these functions as prefixes within fielddef.ts (make sure they're applied in the right order)

usage
`{ fn: [rank, bin] }`

fielddef
```
for [fn in possible_fns]
   if (opt.fn.includes(fn))
     field = `${fn}_${field}`
```

src/compile/mark/mark.ts
willium: @kanitw because its making an array of fields--is there an edge case where one being undefined and another being defined (within the same array) would be the desired result?

kanitw: I think the right fix is assigning `field(fieldDef, {...})` to `const _field` first and then only push if it's defined.   This way it will work correctly with your edge case too.  This one seems important to fix now since we're changing code and we shouldn't add new bug (although it was buggy, this is buggy in a different way ... so worth just fix it here.)

src/compile/mark/mark.ts
```diff
@@ -275,7 +275,7 @@ function stackTransform(model: UnitModel, stackFields: string[]) {
   // add stack transform to mark
   let transform: VgStackTransform = {
     type: 'stack',
-    groupby: [model.field(stack.groupbyChannel, {binSuffix: '_mid'})],
+    groupby: [model.field(stack.groupbyChannel, {binSuffix: 'mid'}) || 'undefined'],
```
kanitw: Similar to here, probably should assign group by only if this field value is not null.  

(We also care out stack transform order that output should come later, so might need to use our nested `extend` pattern.)

willium: Extracted this to #1494

src/compile/mark/mark.ts
kanitw: I think the problem is `field()` doesn't take `scaleType` as input yet and I'm not sure if this condition is always applied when we have `bin`.  But I saw that you file #1493, so let's handle this later.

willium:
Added new issues: #1493, #1494, #1495 per sub-issues discovered during the fix for #778.

Also see: #1449

domoritz:
Good stuff. I'll have a look and see whether sorting and time unit all still works as expected.

Issue #1505, opened 2016-08-16
Barcharts with `scale.zero=false` broken


domoritz:
Reported by @shrjain

```json
{
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"a": "A","b": 28}, {"a": "B","b": 55}, {"a": "C","b": 43},
      {"a": "D","b": 91}, {"a": "E","b": 81}, {"a": "F","b": 53},
      {"a": "G","b": 19}, {"a": "H","b": 87}, {"a": "I","b": 52}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "a", "type": "ordinal"},
    "y": {"field": "b", "type": "quantitative", "scale": {"zero": false}}
  }
}

```

domoritz:
in https://github.com/vega/vega-lite/blob/master/src/compile/mark/bar.ts#L159, we shoulnd't set 0 but instead use the chart height.

kanitw:
Should be addressed in #1507 ... need unit test to prevent regression

PR #1507, opened 2016-08-16
Fix bar/area for log scale and zero = false

kanitw:
Very quick fix for #1505.   @felixcodes will add unit tests later. 

@shrjain For now, can you verify if this works correctly for you?  Thanks!


domoritz:
LGTM

kanitw:
Unit test added.

Issue #1520, opened 2016-08-24
Correct sort line chart when the dimension axis has sort property


kanitw:
For example, see this spec: http://pastebin.com/ZhAtaNw5.

The line chart doesn't suppose to be squiggled.

kanitw:
Fixing this for raw plot is pretty straightforward.  
However, for aggregate plot this can be tricky as they require adding more fields to the aggregate transform to make sure we have the required field.

kanitw:
> for aggregate plot, this can be tricky as they require adding more fields to the aggregate transform to make sure we have the required field.

To be precise, this might even require producing a separate data source.  

@jheer @arvind  Can we read aggregate data for scale sorting in Vega 3? 

(For example, if we sort an ordinal x-scale with "mean of A", we need to sort line value by mean of A too.  If we can't read the aggregate value that the scale already calculate, then we have to manually do it separately to sort line.)

aldanor:
Just stumbled on this -- wonder if this was actually fixed?

Related: https://github.com/altair-viz/altair/issues/961

JoeGermuska:
I believe I've just run into a related problem. When following the [Simple Bar Chart with Labels](https://altair-viz.github.io/gallery/bar_chart_with_labels.html) example in the Altair documentation, I found that adding the labels lost the ordering for the bar chart.

```python
bars = alt.Chart(soy).mark_bar(
             ).encode(y=alt.Y('State', 
                              sort=alt.EncodingSortField(field='Value', 
                                                         op='sum', 
                                                         order='descending')),
                      x = alt.X('Value', 
                                axis=alt.Axis(title='2017 Production in $')))
bars
```
![image](https://user-images.githubusercontent.com/46313/45855378-7df57000-bd15-11e8-8a05-38be1c21eed7.png)

```python
text = bars.mark_text(
    align='left',
    baseline='middle',
    dx=3,
).encode(
    text='Value',
)

bars + text
```
![image](https://user-images.githubusercontent.com/46313/45855409-9d8c9880-bd15-11e8-83ff-081711469d9a.png)

Hopefully this is helpful.

domoritz:
No, this is related to layering and a limitation in Vega. We will address a subset of this issue I'm Vega-Lite 3.

JoeGermuska:
@domoritz Thanks for clarifying.

domoritz:
Do we have a smaller example? I'm not sure whether this is right or wrong. 

[...]

PR #1523, opened 2016-08-25
Correctly sort line chart when the dimension axis has sort property

kanitw:
Part of #1520


kanitw:
vega_editor

Now the spec in #1520 is no longer squiggled.

PR #1528, opened 2016-08-29
Create dictionaries of supported AggregateOps and Channels and export TimeUnit dictionaries

felixcodes:


kanitw:
Actually please use `toMap` from util
and update data lib's toMap signature in .d.ts file to return Dict<any>

felixcodes:
I will instantiate these inside Compass instead so we don't have to release another version of VL.

PR #1533, opened 2016-09-02
Complete sentence in datetime docs

jeffbaumes:
The sentence was not complete before - please correct if I misinterpreted what it should be.


kanitw:
Thanks for catching!

PR #1553, opened 2016-09-22
Upgrade deps

domoritz:
Upgrade all dependencies (including typescript). I made sure the test gallery works.


domoritz:
Apparently tslint has changed significantly.

domoritz:
Merging as I don't think we don't disagree about these changes.

PR #1560, opened 2016-09-29
Move "edit this page" to top of page

### LINKED ISSUE #1359: Should we move "edit this page" link to top of the page?
Putting it at the end is kinda useless.

willium:
Fixes #1359

It'd be great if you could run it and try it out. On second thought, stylistically, I'm not sure I like this.

Reasoning:
- It's not clear where to put it (especially on the homepage, and smaller screens (though I suppose it could be hidden on smaller screens)
- 99% of readers don't/won't/shouldn't edit pages, and it's given a lot of prominence
- it doesn't make good use of space (especially on homepage)
- people generally look for this link in the footer (citation needed)
- Why is putting it at the end useless? Have people had trouble finding it?


domoritz:
I feel similar things. Maybe it makes sense to only do it in the docs (have two links, at the top and bottom). I'd even be okay with a small link right below the title.

willium:
Thoughts @kanitw?

kanitw:
I agree that we shouldn't do this on the homepage and for small screens.   

> people generally look for this link in the footer (citation needed)

> Why is putting it at the end useless? Have people had trouble finding it?

I am pretty sure that 90% of people don't actively look for this link.  
If we don't put the link on the top, the problem is they don't even know that they can edit the docs.  

> 99% of readers don't/won't/shouldn't edit pages, and it's given a lot of prominence

Maybe this will change the number to 98%?   At worst, they will just scroll through it and do nothing.  At least they'll notice that they can contribute.

domoritz:
Okay, then that's decided. Let's put a second edit link at the top of the docs but no other page.

willium:
Should we put it in the footer on other pages?

domoritz:
Just leave it where it is on all pages but add a link on the docs pages.

willium:
Me gusta. will do tonight.

domoritz:
Tonight ten days ago :-P

willium:
err.... yeah I'll get on that 😬

domoritz:
ping

willium:
Should finally be ready... @domoritz @kanitw

site/static/main.css
```diff
@@ -277,10 +278,31 @@ nav a:hover {
   flex: 0 0 auto;
 }
 
+#edit-this-page {
+  display: inline-block;
+  font-size: 0.9em;
+  color: #aaa;
+  padding-top: 30px;
+  cursor: pointer;
```
domoritz: This is weird. It feel like I can click on the whole text but it's not possible.

_layouts/docs.html
```diff
@@ -40,5 +40,5 @@
 
 base: /docs/
 ---
-
+{% include edit.html %}
```
domoritz: I'd prefer a shorter version at the top that does not have the "and submit a pull request!"

domoritz:
LGTM except for the two comments. Will merge after they are done.

willium:
@domoritz ready

_layouts/docs.html
```diff
@@ -40,5 +40,7 @@
 
 base: /docs/
 ---
-
+<span id="edit-this-page">
```
domoritz: Doesn't `edit-this-page` have to be a class now?

willium: Sigh. Yes

domoritz: Sorry for being so pedantic.

willium: No no you're right. I'll fix soon

PR #1564, opened 2016-09-29
Legend can be false. Fixes #1563

### LINKED ISSUE #1563: Schema doesn't allow for legend=false
The vega-lite-schema.json file specifies that encoding.legend must be a Legend object, but I see in the doc (and examples) that legend can be set to false too.

```json
ChannelDefWithLegend: { 
  type: "object", 
  properties: {
    legend: {
      $ref: "#/definitions/Legend"
    },
   ...
```

Axis has the same feature (i.e. can be boolean or Axis), but in this case the schema specifies that axis can be Axis | Boolean. 

Looking at the code, I see:

```typescript
export interface PositionChannelDef extends ChannelDefWithScale {
  axis?: boolean | Axis;
}
export interface ChannelDefWithLegend extends ChannelDefWithScale {
  legend?: Legend;
}
```

Can legend be changed to be ```legend?: boolean | Legend;```? As it currently stands third-parties that are generating code using the json-schema aren't able to support legend=false until this is changed. 

PS. Digging a little I see that the boolean feature of legend is supported by custom parsing logic in UnitModel.

domoritz:


kanitw:
Now that you upgrade to Typescript 2.0, aren't we suppose to eliminate scale/axis/legend = false?

kanitw:
We should finish #1384 and make axis/legend = null as the default way to disable them.

PR #1570, opened 2016-10-01
Add month-date back since we use year 0 

### LINKED ISSUE #1557: Support month date with leap year
From @jakevdp's question [here](https://groups.google.com/forum/#!topic/vega-js/451aFN7_2_k),
thinking more about it -- we can actually support MONTHDATE unit correctly without modifying anything because year 0 that we [automatically augment to the formula](https://github.com/vega/vega-lite/blob/master/src/datetime.ts#L152) is a leap year. 

For example, the output Vega Expression will be `datetime(0, month(datum[\"date\"]), date(datum[\"date\"]), 0, 0, 0, 0)`.

This way, the output time scale will always have February 29 on the time scale no matter the data include a leap year or not.  I think this is the correct behavior since I can't think of any other ways to make the scale correct.  

@arvind @jheer @domoritz - Do I overlook anything?

kanitw:
thus we automatically include February 29th.

I believe this is the correct behavior.

Fix #1557

### DIFF
--- a/src/timeunit.ts
+++ b/src/timeunit.ts
@@ -13,13 +13,13 @@ export enum TimeUnit {
   SECONDS = 'seconds' as any,
   MILLISECONDS = 'milliseconds' as any,
   YEARMONTH = 'yearmonth' as any,
-  // Note: don't add MONTH DATE because it will be incorrect
-  // since days on a leap year will be shifted by one if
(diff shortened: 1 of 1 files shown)


kanitw:
Ping: @domoritz

PR #1575, opened 2016-10-06
Remove browserify-versionify to fix webpack/browserify compatibility

mathisonian:
This pulls the version directly from `package.json` and removes the need for a transform to enable better compatibility with build tools like browserify and webpack. Note webpack users will still need to have the [json-loader](https://github.com/webpack/json-loader) enabled. 

This is related to vega/vega#369 in the overall goal of improving compatibility with the javascript ecosystem.

Please:
- [ ] Provide a test case & update the documentation under `site/docs/`

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
   "devDependencies": {
     "browser-sync": "~2.16.0",
     "browserify": "~13.1.0",
-    "browserify-versionify": "~1.0.6",
     "chai": "~3.5.0",
     "cheerio": "~0.22.0",
     "exorcist": "~0.4.0",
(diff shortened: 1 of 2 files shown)


kanitw:
Thanks @mathisonian

domoritz:
That's a sweet approach. We could do the same in https://github.com/vega/vega, I think.

mathisonian:
Yes! I should have some more PR's like this coming soon. It looks like the work @jheer did on v3 may make it irrelevant https://github.com/uwdata/vega/blob/master/package.json, but I think its still useful for people on v2 in the meantime.

Issue #1436, opened 2016-06-17
Add `data.format.parse` and take them into account for merging data.


kanitw:
- [x] We still don't have `format.parse`
  - [x] need to merge them with default parse properties that we infer from type

- [ ] take them into account for merging data

kanitw:
@domoritz Looks  like we need this and thus I might to do this and merge into master since we can't wait for the layering stuff.

kanitw:
Looks like if we don't have https://github.com/vega/vega/issues/610 fixed by mid-week next week, we'll have to implement this to support custom parsing which will avoid the broken sorting bug.

The question is whether `format.parse` is a natural syntax for Vega-Lite.  
In general, we try to avoid repeating fields in other places.  

Should this become a part of channel's fieldDef  (e.g., `encoding.x.parse` or `encoding.x.primitiveType`)  or it is more natural to be just `format.parse` akin to Vega? 


cc: @arvind @domoritz

Georgeouz:
> Should this become a part of channel's fieldDef (e.g., encoding.x.parse or encoding.x.primitiveType) or it is more natural to be just format.parse akin to Vega?

Wouldn't it be as natural to infer the primitive data type *under the hood* from the encoded data type? 

e.g. if the user sets `type` to `quantitative` for any encoded field, the said field would be parsed as `number`

kanitw:
> Wouldn't it be as natural to infer the primitive data type under the hood from the encoded data type?
e.g. if the user set encoding.x.type to quantitative, the field would be parsed as number

That's what we currently do.  However, that's not always the right solution.  Users may use numbers as an ordinal / nominal field.

Georgeouz:
I guess you're right, plus we're not talking about filtering and sorting data based on fields not encoded in the graphic. In such a context, the parse à la Vega would be the more universal option.

kanitw:
> After all, by default the primitive format would be inferred using the current mechanism and the specific parse option would be only needed to override the defaults.

Right.

One thing that I'm considering is whether we should add `type` as optional parameter for `sort` and `filter` objects, which lessen the need for using this `format.parse`.  
In any case, we still need `format.parse` for some cases.  For example, we need it for `calculate` when we derive a field and the field is only included in the Vega Expression language.

kanitw:
The left over task (making it work for composition is now #1578).

PR #1576, opened 2016-10-11
Better format parse for unit specs

kanitw:
- Basic `format.parse` property support (for overriding inferred parse type)
- Infer format.parse for filtered field

I think this might be still incorrect for composition, but I would rather make it correct for 1.0 first

Part of #1436

(Please don't squash as the commits well splitted)


kanitw:
@domoritz -- note that test is passing, but something else is failing

PR #1580, opened 2016-10-12
Use new ts typings and remove the files section from tsconfig.

### LINKED ISSUE #1450: Replace locally committed declaration file with tsd
https://github.com/DefinitelyTyped/tsd

domoritz:
Closes #1450


package.json
```diff
@@ -54,6 +54,12 @@
   },
   "homepage": "https://github.com/vega/vega-lite",
   "devDependencies": {
+    "@types/assertion-error": "^1.0.30",
+    "@types/chai": "^3.4.34",
+    "@types/d3": "^3.5.36",
```
kanitw: I guess we don't need D3 typing anymore since we don't import D3?

package.json
```diff
@@ -54,6 +54,12 @@
   },
   "homepage": "https://github.com/vega/vega-lite",
   "devDependencies": {
+    "@types/assertion-error": "^1.0.30",
```
kanitw: Do we use this AssertingError thing anywhere?

domoritz: We don't. I removed it.

PR #1584, opened 2016-10-14
[WIP] Start to use namespace instead of odd enums

domoritz:


kanitw:
This is lengthier to declare and looks equally like hack.... What's the true benefit of doing this?

domoritz:
It's less hacky and should be easier to generate correct json for. I thought I could use it to upgrade the ts to json converter to ts 2.0 but there are some more issues with it. I'll punt on this for now.

kanitw:
> It's less hacky

At least, the code looks more complex too me.

PR #1599, opened 2016-10-20
[WIP] Scale Disabling 

kanitw:
- [ ] Blocked by Bar, Rule Refactor


kanitw:
Thinking more about it.  The only case scale should not exist with field is when we use projection. 

To use the raw data as visual values, People should use `identity` scale rather than disabling scale since the identity scale is required to produce correct axis and legend.

PR #1603, opened 2016-10-20
Enable declaration emit

weswigham:
Consuming `vega-lite` as a commonjs package via the typescript compiler was difficult/impossible, as, rather than shipping with generated type definition files, it shipped with the original source typescript. This caused issues where compiler setting mismatches (`noImplicitAny`, `strictNullChecks`, and so on) would expose issues in `vega-lite`'s implementation as part of a consumer's build, which is undesirable.

Turning on declaration emit (and fixing declaration visibility errors), allows the compiler to automatically generate declaration files for the project, which make the project easy to use by a typescript consumer.

### DIFF
--- a/site/static/main.ts
+++ b/site/static/main.ts
@@ -2,6 +2,9 @@
 
 declare const BASEURL, hljs;
 
+// IIFE to prevent function declarations from moving into the global scope
+(() => {
+
 function trim(str: string) {
   return str.replace(/^\s+|\s+$/g, '');
 }
@@ -85,7 +88,7 @@ function renderGallery() {
       // try to retrieve specs for a gallery group from in vl-examples.json
       try {
         galleryGroupSpecs = VL_SPECS[galleryGroupName];
-      } catch (error){
+      } catch (error) {
         console.log(error.message);
         return;
       }
@@ -124,3 +127,5 @@ function renderGallery() {
     }
   });
 }
+
+})();
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -15,7 +15,7 @@ export namespace bin {
     return `format('${format}', ${expr})`;
   }
 
-  export function parseUnit(model: Model): Dict<VgTransform[]> {
+  function parse(model: Model): Dict<VgTransform[]> {
     return model.reduce(function(binComponent, fieldDef: FieldDef, channel: Channel) {
       const bin = model.fieldDef(channel).bin;
       if (bin) {
@@ -65,8 +65,10 @@ export namespace bin {
     }, {});
   }
 
+  export const parseUnit: (model: Model) => Dict<VgTransform[]> = parse;
+
   export function parseFacet(model: FacetModel) {
-    let binComponent = parseUnit(model);
+    let binComponent = parse(model);
 
     const childDataComponent = model.child().component.data;
 
@@ -80,7 +82,7 @@ export namespace bin {
   }
 
   export function parseLayer(model: LayerModel) {
-    let binComponent = parseUnit(model);
+    let binComponent = parse(model);
 
     model.children().forEach((child) => {
       const childDataComponent = child.component.data;
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -15,7 +15,7 @@ export namespace bin {
     return `format('${format}', ${expr})`;
   }
 
-  function parse(model: Model): Dict<VgTransform[]> {
+  export function parseUnit(model: Model): Dict<VgTransform[]> {
     return model.reduce(function(binComponent, fieldDef: FieldDef, channel: Channel) {
       const bin = model.fieldDef(channel).bin;
       if (bin) {
@@ -65,10 +65,8 @@ export namespace bin {
     }, {});
   }
 
-  export const parseUnit = parse;
-
   export function parseFacet(model: FacetModel) {
-    let binComponent = parse(model);
+    let binComponent = parseUnit(model);
 
     const childDataComponent = model.child().component.data;
 
@@ -82,7 +80,7 @@ export namespace bin {
   }
 
   export function parseLayer(model: LayerModel) {
-    let binComponent = parse(model);
+    let binComponent = parseUnit(model);
 
     model.children().forEach((child) => {
       const childDataComponent = child.component.data;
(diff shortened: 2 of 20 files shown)


src/compile/data/bin.ts
```diff
@@ -15,7 +15,7 @@ export namespace bin {
     return `format('${format}', ${expr})`;
   }
 
-  function parse(model: Model): Dict<VgTransform[]> {
+  export function parseUnit(model: Model): Dict<VgTransform[]> {
```
kanitw: If you really need typing for `parseUnit` methods, I would prefer that you keep `parse` method the same but separately introduce function `parseUnit(model: UnitModel)` that call `parse(model: Model)` because semantically `parseFacet/Layer` DO NOT involve `parseUnit(facet/layerModel)` but rather a generic parse method.

src/compile/model.ts
```diff
@@ -64,6 +70,11 @@ class NameMap {
   }
 }
 
+export interface NameMapShape {
```
kanitw: Can you explain why you need to add this interface?  

In any case, `NameMapShape` is quite unconventional IMHO.  Why not `NameMapInterface`?

src/compile/model.ts
```diff
@@ -15,6 +15,12 @@ import {VgData, VgMarkGroup, VgScale, VgAxis, VgLegend} from '../vega.schema';
 import {DataComponent} from './data/data';
 import {LayoutComponent} from './layout';
 import {ScaleComponents} from './scale';
+/* tslint:disable:no-unused-variable */
```
kanitw: Do you really need this set of hacks?  It's weird that tslint isn't friendly with .d.ts export.

src/compile/data/bin.ts
weswigham: The reason for the change is that `parseUnit` was a public const which was set to the value of `parse`. `parse` was a closure variable, so its types were not considered public, resulting in declaration errors. Instead of simply changing parse to `parseUnit` (which seemed right, since they were equivalent), I can revert it to `parse` and write 
```ts
export const parse: (model: Model) => Dict<VgTransform[]> = parseUnit;
```
if that is preferred. I simply thought this would be less verbose.

src/compile/model.ts
weswigham: `tslint` has `no-unused-variables` on, and these variables are, _technically_ unused in the code, _however_, they need to be named so their names can be used in the resulting compiled `.d.ts` (you will get compiler errors from typescript without the references).

src/compile/model.ts
weswigham: I'll happily rename it to `NameMapInterface` if you'd prefer. The issue in this case is as follows:
`NameMap` is internal, but protected (and thus, publicly accessible) members of the exported class below reference the type. To avoid needing to export the `NameMap` implementation, I split the interface and the implementation (hence why `NameMap` implements `NameMapInterface`) so I could just export the interface.

An alternate fix is to just export `NameMap`, if you'd prefer.

weswigham:
@kanitw I'm also not terribly concerned of the compatibility between versions of `vega-lite`; just a given version having a well-defined interface. Additionally, including the original source `ts` files the package was causing webpack with a `ts-loader` to prefer those files to the `js` files it should actually use to create the bundle (while the `.d.ts` files should get used for type checking). So this serves a dual purpose - it gets the `.ts` files out of the way and provides `.d.ts` files to replace them in the build result, to make it easy to consume the library form typescript.

TBH, the _main_ reason I wanted this defined and functional was so I could have an up-to-date and always-correct version of the typescript interface for the `vega-lite` `ExtendedSpec` spec. (I could copy all the type information, but that would be terribly hacky; simply fixing the build to generate declarations is far nicer).

src/compile/data/bin.ts
kanitw: Thx for explanation-- i kinda prefer the verbose version except that It was the other way round. ( export const parseUnit = parse)

(Replying from mobile)

site/static/main.ts
```diff
@@ -2,6 +2,9 @@
 
 declare const BASEURL, hljs;
 
+// IIFE to prevent function declarations from moving into the global scope
```
domoritz: Hmm, this should not be part of the d.ts for vl anyway. Can you remove this?

domoritz: Shouldn't the only declaration file be a single `vl.d.ts` rather than a bunch of declaration files that reference each other?

weswigham: The `.d.ts` structure mimics the file structure of the project (see how the types field in the package mimics the main field). There is no built in tool to bundle `.d.ts` files, nor is there a reason to bundle them, since they only get used in the compilation/editor.

Re: static site ts changes. I honestly would prefer not to change the static site assets, but I can't help it in the current build setup. The site ts is a script, not a module. This is fine, and ts does support mixed compilations. The problem is that with declarations on, that script file has ambient declarations made for it; so everything which leaks into the global scope gets a declaration in the resulting `.d.ts`. Usually, I wouldn't care... Nobody's planning on using that `.d.ts` file, anyway, except it has a declaration visibility error which arises in the compilation. The type of the first argument to `renderExample`  is considered private, which causes a declaration generation error. The solution I chose was to simply remove the global leaks by doing all the work inside an IIFE. IMO, a slightly more complete/better fix would be a separate tsconfig and a separate build for the site and the library (so the site can be compiled with declarations off, rather than on), but this works if that's too complex.

domoritz: Thank you for the explanation. I think a separate `tsconfig` would be the clean solution but to reduce complexity, let's stick with one until it becomes a problem.

test/util.ts
```diff
@@ -2,9 +2,10 @@ import {buildModel} from '../src/compile/common';
 import {UnitModel} from '../src/compile/unit';
 import {ExtendedUnitSpec, normalize} from '../src/spec';
 import {contains} from '../src/util';
+import {Model} from '../src/compile/model';
 
 // TODO: rename to parseModel
```
domoritz: Looks like we don't need the TODO any more :-)

domoritz:
LGTM. I'll merge after `NameMapShape` is renamed.

src/compile/data/bin.ts
weswigham: Reverted and added type annotation.

src/compile/model.ts
weswigham: Renamed to `NameMapInterface`.

test/util.ts
weswigham: Removed. ;)

weswigham:
@kanitw @domoritz I think I've responded to/applied all the changes you've asked for?

domoritz:
Thanks!

PR #1604, opened 2016-10-20
Only return scaleName if the scale exists

kanitw:
Use `model.name()` during scale parsing to name scale
and make `model.scaleName()` only return values if the scale exists.

This is a good preparation for supporting projection.
I don't think I'll do projection now given how much that changes in Vega 3,
but since I have already done this part and it does no harm to the current code. 
Let's merge this part first.  

I have checked that all vega output of example specs remain the same.

### DIFF
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -172,7 +172,7 @@ export class FacetModel extends Model {
         // for each scale, need to rename
         vals(scaleComponent[channel]).forEach(function(scale) {
           const scaleNameWithoutPrefix = scale.name.substr(child.name('').length);
-          const newName = model.name(scaleNameWithoutPrefix);
+          const newName = model.scaleName(scaleNameWithoutPrefix, true);
           child.renameScale(scale.name, newName);
           scale.name = newName;
         });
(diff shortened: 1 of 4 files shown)


domoritz:
Why can't we just always use a single method and the method figures out what code to run? How are `.name` and `.scaleName` different?

kanitw:
> Why can't we just always use a single method and the method figures out what code to run? How are .name and .scaleName different?

The method cannot be smart enough to know whether if it is in the parse phase or not.  

Another alternative would be extend `scaleName` with `parse: boolean` flag and do pretty much the same logic.  There are two options for this approach. 

a) if the parse is not optional, we have to change `scaleName` everywhere, which unnecessarily make the code verbose. 
b) if the parse is optional, then we can change `name` to `scaleName(..., true)`.  This is pretty similar to having two methods.  But maybe it is slightly better?  

Before I do b), let me know what you think.

domoritz:
I prefer b). Having a single method is less error prone because you make the consumer of the method aware of the distinction.

domoritz:
Tests are failing

kanitw:
I'll fix later -- we don't need this for 1.3

PR #1609, opened 2016-10-21
Hyphenate paragraphs

domoritz:


kanitw:
From discussion in https://css-tricks.com/almanac/properties/h/hyphenate/, looks like Chrome suck at this, so I'm against this.    

Our text column is pretty wide, I can't imagine a case where we will really benefit from this.

domoritz:
Chrome 55 supports is now and firefox and safari supported this for a while. It makes text flow much nicer.

domoritz:
Chrome 55 supports it and so do safari and Firefox. I just enabled it on my
website.

willium:
I personally think we should not do this. Hyphen's really only belong in justified text IMO.

kanitw:
I agree with @willium.
So 2:1 @domoritz.  lets close this?

PR #1611, opened 2016-10-22
Split properties inside marks

kanitw:


src/compile/config.ts
```diff
@@ -83,32 +83,23 @@ export function orient(mark: Mark, encoding: Encoding, markConfig: MarkConfig =
...
-        return Orient.VERTICAL;
-      }
-      if (encoding.y) {
-        return Orient.HORIZONTAL;
-      }
-      if (encoding.x) {
-        return Orient.VERTICAL;
-      }
-      // no x/y -- so it's undefined
-      return undefined;
```
domoritz: will this ever happen? If not, use `never` https://basarat.gitbooks.io/typescript/content/docs/types/never.html

src/compile/mark/bar.ts
```diff
@@ -37,17 +37,10 @@ export namespace bar {
       };
     } else if (xIsMeasure) {
       if (orient === Orient.HORIZONTAL) {
-        if (model.has(X)) {
-          p.x = {
-            scale: model.scaleName(X),
-            field: model.field(X)
-          };
-        } else {
-          p.x = {
```
domoritz: the argument is that if we have orient, the model has an x, right?

src/compile/mark/rule.ts
```diff
@@ -11,85 +14,99 @@ export namespace rule {
...
 
-      if (model.has(Y2)) {
-        p.y2 = {
-          scale: model.scaleName(Y),
-          field: model.field(Y2, { binSuffix: 'mid' })
-        };
-      } else {
-        p.y2 = { value: 0 };
-      }
+    // FIXME: this function would overwrite strokeWidth but shouldn't
```
domoritz: can we fix this?

src/compile/mark/rule.ts
```diff
@@ -11,85 +14,99 @@ export namespace rule {
...
-    if (fieldDef && fieldDef.value !== undefined) {
-       return fieldDef.value;
+  function size(fieldDef: FieldDef, scaleName: string, config: Config): VgValueRef {
+    if (fieldDef) {
+      if (fieldDef.field) {
+        return {
+          scale: scaleName,
+          field: field(fieldDef) // TODO: what's the missing suffix
+        };
+      } else if (fieldDef.value) {
```
domoritz: here you use elseif, up you use if + return

src/compile/config.ts
```diff
@@ -83,32 +83,23 @@ export function orient(mark: Mark, encoding: Encoding, markConfig: MarkConfig =
       }
       // y:Q or Ambiguous case, return horizontal
       return Orient.HORIZONTAL;
+
     case RULE:
-      if (xIsRange) {
-        return Orient.HORIZONTAL;
-      }
-      if (yIsRange) {
```
domoritz: Here you use if + return, below you use else if. I prefer to use elseif to indicate exclusiveness.

src/compile/mark/bar.ts
kanitw: Yes.

src/compile/mark/rule.ts
kanitw: Not in this PR.  Irrelevant.

src/compile/mark/rule.ts
kanitw: I'll fix in another pending PR that extract common methods for marks.

src/compile/config.ts
kanitw: as an intermediate step in PoleStar, yes this can happen.

src/compile/config.ts
kanitw: I don't understand why you're bugging about removed lines.

src/compile/mark/rule.ts
kanitw: (This is an old comment that I didn't add. It was already here.)

kanitw:
@domoritz 
-  Test should be passing now 
- Replied to all of your comments.  
-  I squash your lint fix into relevant commits, so this PR should ready for rebase-merge.

src/compile/config.ts
domoritz: Ups...

PR #1613, opened 2016-10-23
[WIP] Getting the right scale in mark

kanitw:
Scale gets merged / etc in composition.  We need some way to get the right scale name and type in marks.  This is one such attempt.   However, I don't like the fact that this approach will couple mark properties method with prior parse phase and make all our unit tests for marks broken. 

There should be away that still make mark standalone.  A few options in mind:
a) Instead of having `component.scale`, replace `model._scale` with the component after parsing.
b) have a common model method for doing so. 

 (I haven't tried hacking any of these, so they might lead to a different set of problems.)

For now, I'll start another PR to refactor marks to use common set of FieldRef methods first

### DIFF
--- a/src/compile/mark/mark-common.ts
+++ b/src/compile/mark/mark-common.ts
@@ -1,0 +1,41 @@
+import {ScaleComponent} from '../scale';
+
+import {X, Y, Channel} from '../../channel';
+import {FieldDef, FieldRefOption, field} from '../../fielddef';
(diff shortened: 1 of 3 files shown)


domoritz:
I find the "Ready for review" label incredibly confusing. Do we still need this now that github has projects?

kanitw:
I don't think this will get merged since we already refactored mark a lot.  Let's just close this!

PR #1614, opened 2016-10-23
Extract common methods for marks

### LINKED ISSUE #1284: Split properties() in mark-*.ts into multiple methods (e.g., x(), y(), shape())
Each of these methods should take `fieldDef` and `scaleName` rather than the `model` as input since one channel might have multiple fieldDefs.

(forked from #1124, so we can assign to other people)

kanitw:
Fix #1284

Common methods are in `src/compile/mark/valueref.ts`.
Mark's code is now cut by roughly half (-711, +376)

More details in each (atomic) commit.

### DIFF
--- a/src/compile/mark/area.ts
+++ b/src/compile/mark/area.ts
@@ -1,14 +1,9 @@
-import {VgValueRef} from '../../vega.schema';
-
 import {X, Y} from '../../channel';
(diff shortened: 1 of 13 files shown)


domoritz:
Sweet.

PR #1618, opened 2016-10-24
Supporting setting gap between binned bar 

### LINKED ISSUE #1337: Draw bar charts with binning without gaps
One could argue that we are binning a continuous domain and there should be no gaps. 

screen shot 2016-05-03 at 12 55 12

We still want the gaps for categorical data

screen shot 2016-05-03 at 12 58 40

What would be a bit annoying is that we have to offset the ticks by .5px so it is between the bars.

kanitw:
Fix #1337

Add `binBarOffset` for setting  gap between binned bar.

### DIFF
--- a/examples/specs/histogram_nospacing.vl.json
+++ b/examples/specs/histogram_nospacing.vl.json
@@ -1,0 +1,21 @@
+{
+  "data": {"url": "data/movies.json"},
+  "mark": "bar",
+  "encoding": {
+    "x": {
+      "bin": {"maxbins": 10},
+      "field": "IMDB_Rating",
+      "type": "quantitative"
+    },
+    "y": {
+      "aggregate": "count",
+      "field": "*",
+      "type": "quantitative"
+    }
+  },
+  "config": {
+    "mark": {
+      "binnedBarSpacing": 0
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


domoritz:
Offset is definitely not the right word. `binBarSpacing`?

kanitw:
renamed to `binnedBarSpacing`

kanitw:
(ready for review)

willium:
Is there a reason we don't let you specify/namespace the mark type you're config-ing?

kanitw:
> Is there a reason we don't let you specify/namespace the mark type you're config-ing?

Are you saying you want `config.mark.bar.binSpacing`?

willium:
More jus curious why this isn't already the structure. Was a little thrown off when working on themes, which seem to intersect weirdly with config

kanitw:
> why this isn't already the structure

What is `this`? 

> Was a little thrown off when working on themes, which seem to intersect weirdly with config

@willium `config` is the theme for Vega-Lite.  
We have opportunity to provide more fine grain control here than in Vega.  
That said, it would be great to know what's missing in VL config that exists in Vega theme.

domoritz:
Manually merged

Issue #1342, opened 2016-05-10
Syntax for Table Heatmap


kanitw:
Currently heatmap can be created using a `bar` mark.  

```
{
  "description": "A bar chart showing the US population distribution of age groups in 2000.",
  "data": { "url": "data/population.json"},
  "mark": "bar",
  "encoding": {
    "y": {
      "field": "age", "type": "ordinal"
    },
    "x": {
      "field": "year", "type": "ordinal"
    },
    "color": {
      "aggregate": "mean", "field": "people", "type": "quantitative" 
    }
  }
}
```

but that might not be the most natural way to do it.  We might want to consider other alternative. 

@willium's thought (copied from #1339)

> I'm also curious to see if the semantics of a heat map could be distanced from the "bar" mark too -- as discussed. It would not be my first guess, if I was trying to implement a heat map in Vega-lite. While troublesome, some sort of table seems more reasonable to me. What about the introduction of a "cell" mark (i.e. table-cell)?

kanitw:
For comparison, Tableau has a weird semantic for its `square` marks for this case. 

Meanwhile, ggplot has a more sensible [`tile` marks](http://docs.ggplot2.org/current/geom_tile.html).

willium:
I like tile as well (though cell might be more flexible if we end up wanting to support tables? It's also more ambiguous.

willium:
https://vega.github.io/vega-editor/?mode=vega-lite&spec=text_table_heatmap is related

kanitw:
Let's go with tile.   We still have to consider how to deal with tile for quantitative field though.

kanitw:
Another alternative to consider is `rect`.  

Basically, if there is only `x` (but not `x2`) we can treat `x` as `xc` and expand the `rect` along the x-axis.  And if both `x` and `x2` exist, then this behaves like rect in Vega.  This would be consistent with what @ayush29feb is doing.

pierre-haessig:
Hi,

I just started exploring vega-lite (through the use of altair), and I quickly fell on the issue of how to draw heatmap (which in my particular use case happens to be just a 2d histogram/density plot). Unless I missed something, the syntax using `bar` mark is not advertised in the doc, rather only present in the issue tracker.

[...]

PR #1621, opened 2016-10-25
Add heatmap example. #1342

domoritz:


aishfenton:
Rather than using `square` would it be better to use `bar`? Since the result it gives looks much better. Although marking it as "bar" does seem confusing, but at least it looks like a heat map :)

domoritz:
@aishfenton Ups. I pushed the wrong version. Thanks for paying attention!

kanitw:
@domoritz  Honestly let's not add this.   It's a hack, not a well principle design. 
Our principled design is to add `rect` mark.

domoritz:
We should add the spec and fix it later. We will change to rect or tile before the 2.0 release.

kanitw:
Have another WIP branch that will handle this correctly.

PR #1626, opened 2016-10-29
Consolidate log + Add Test to check warnings

### LINKED ISSUE #1600: Replace console.warn with internal warning

kanitw:
Fix #1600 

This will allow us to test if a method output appropriate log in the console and flood the command line output less during testing.

(See https://github.com/vega/vega-lite/commit/b1a6d02636ab85ccb02f6cf232c317e26edce6ee)

TODO: 
- [x] Figure out how to import `vega-util` module
- [x] Modify `vl.compile` API to include LocalLogger
- [x] Test all the output warning


kanitw:
@domoritz  This branch is almost ready, the only thing that is HACKY is that I can't directly import logger from vega-util since it's an ES6 module (try first commit). 

So I added the second commit which basically copies code from vega-util as a workaround.  

We should import things property and squash the first two commits, then this branch should be good to go!

.vscode/settings.json
```diff
@@ -3,7 +3,7 @@
   "editor.tabSize": 2,
   "files.exclude": {
     "**/*.js": {"when": "$(basename).ts"},
-    "**/*.d.ts": {"when": "$(basename).ts"},
+    "src/**/*.d.ts": {"when": "$(basename).ts"},
```
domoritz: why only in src?

kanitw: Because we only auto-generate the `*.d.ts` files in `src/` (not in `test/`)  

The old code exclude `typings/*.d.ts` which are not auto-generated.

domoritz: But it only does so there is a .ts file with the same path/name, right?

kanitw: That's a good point, but it somehow still hides my typings files.

domoritz: Hmm. It used to work but not any more. this is annoying.

PR #1640, opened 2016-11-01
Allow Rule to be stacked (for consistency)

### LINKED ISSUE #1629: Consider how stacking works with rule
Also consider if baseOrMax should be for x,y instead of x2, y2

kanitw:
Please merge #1627 first.


domoritz:
Not sure why anyone would ever do this but why not make it consistent.

PR #1649, opened 2016-11-01
Enable noImplicitAny

weswigham:
And then fix all the resulting typescript compiler errors (usually just adding types to places). This identified a handful of typos, which were fixed, and necessitated swapping out the string enum hack with a real, correctly-typed namespace structure.

Without `noImplictAny` on in `vega-lite`, due to current limitations in the typescript compiler, it is almost impossible for a typescript consumer to use vega-lite with `noImplictAny` on, as all the implicit any errors which leak into the generated `.d.ts` API surface in consumers' projects.

And it's just a good idea to have this flag on, in general, as it helps identify many likely bugs. I'd also recommend enabling struct null checks in the future, which will hopefully be less work than this PR was.

### DIFF
--- a/src/aggregate.ts
+++ b/src/aggregate.ts
@@ -1,26 +1,52 @@
 
-export enum AggregateOp {
-    VALUES = 'values' as any,
-    COUNT = 'count' as any,
-    VALID = 'valid' as any,
-    MISSING = 'missing' as any,
-    DISTINCT = 'distinct' as any,
-    SUM = 'sum' as any,
-    MEAN = 'mean' as any,
-    AVERAGE = 'average' as any,
-    VARIANCE = 'variance' as any,
-    VARIANCEP = 'variancep' as any,
-    STDEV = 'stdev' as any,
-    STDEVP = 'stdevp' as any,
-    MEDIAN = 'median' as any,
-    Q1 = 'q1' as any,
-    Q3 = 'q3' as any,
-    MODESKEW = 'modeskew' as any,
-    MIN = 'min' as any,
-    MAX = 'max' as any,
-    ARGMIN = 'argmin' as any,
-    ARGMAX = 'argmax' as any,
+export namespace AggregateOp {
+    export const VALUES: 'values' = 'values';
+    export type VALUES = typeof VALUES;
+    export const COUNT: 'count' = 'count';
+    export type COUNT = typeof COUNT;
+    export const VALID: 'valid' = 'valid';
+    export type VALID = typeof VALID;
+    export const MISSING: 'missing' = 'missing';
+    export type MISSING = typeof MISSING;
+    export const DISTINCT: 'distinct' = 'distinct';
+    export type DISTINCT = typeof DISTINCT;
+    export const SUM: 'sum' = 'sum';
+    export type SUM = typeof SUM;
+    export const MEAN: 'mean' = 'mean';
+    export type MEAN = typeof MEAN;
+    export const AVERAGE: 'average' = 'average';
+    export type AVERAGE = typeof AVERAGE;
+    export const VARIANCE: 'variance' = 'variance';
+    export type VARIANCE = typeof VARIANCE;
+    export const VARIANCEP: 'variancep' = 'variancep';
+    export type VARIANCEP = typeof VARIANCEP;
+    export const STDEV: 'stdev' = 'stdev';
+    export type STDEV = typeof STDEV;
+    export const STDEVP: 'stdevp' = 'stdevp';
+    export type STDEVP = typeof STDEVP;
+    export const MEDIAN: 'median' = 'median';
+    export type MEDIAN = typeof MEDIAN;
+    export const Q1: 'q1' = 'q1';
+    export type Q1 = typeof Q1;
+    export const Q3: 'q3' = 'q3';
+    export type Q3 = typeof Q3;
+    export const MODESKEW: 'modeskew' = 'modeskew';
+    export type MODESKEW = typeof MODESKEW;
+    export const MIN: 'min' = 'min';
+    export type MIN = typeof MIN;
+    export const MAX: 'max' = 'max';
+    export type MAX = typeof MAX;
+    export const ARGMIN: 'argmin' = 'argmin';
+    export type ARGMIN = typeof ARGMIN;
+    export const ARGMAX: 'argmax' = 'argmax';
+    export type ARGMAX = typeof ARGMAX;
 }
+export type AggregateOp = AggregateOp.ARGMAX | AggregateOp.ARGMIN | AggregateOp.AVERAGE
+  | AggregateOp.COUNT | AggregateOp.DISTINCT | AggregateOp.MAX | AggregateOp.MEAN
+  | AggregateOp.MEDIAN | AggregateOp.MIN | AggregateOp.MISSING | AggregateOp.MODESKEW
+  | AggregateOp.Q1 | AggregateOp.Q3 | AggregateOp.STDEV | AggregateOp.STDEVP
+  | AggregateOp.SUM | AggregateOp.VALID | AggregateOp.VALUES | AggregateOp.VARIANCE
+  | AggregateOp.VARIANCEP;
 
 export const AGGREGATE_OPS = [
     AggregateOp.VALUES,
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -8,7 +8,7 @@ import {FieldDef, field, OrderChannelDef} from '../fielddef';
 import {SortOrder} from '../sort';
 import {TimeUnit} from '../timeunit';
 import {QUANTITATIVE, ORDINAL} from '../type';
-import {contains, union} from '../util';
+import {contains, union, Dict} from '../util';
 
 import {FacetModel} from './facet';
 import {LayerModel} from './layer';
@@ -16,7 +16,7 @@ import {Model} from './model';
 import {template as timeUnitTemplate} from '../timeunit';
 import {UnitModel} from './unit';
 import {Spec, isUnitSpec, isSomeFacetSpec, isLayerSpec} from '../spec';
-
+import {VgValueRef} from '../vega.schema';
 
 export function buildModel(spec: Spec, parent: Model, parentGivenName: string): Model {
   if (isSomeFacetSpec(spec)) {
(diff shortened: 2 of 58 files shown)


kanitw:
> necessitated swapping out the string enum hack with a real, correctly-typed namespace structure.

Can u explain more why you need to replace enum with that complex structure?  I really dont like that complex structure and the fact that you need to type 3 times just to declare an enum.
I suspect that there might be a better solution.

kanitw:
@domoritz this kind of PR would be a good reason why we should include json schema in git.  I currently have no clue if the schema has changed.

weswigham:
@kanitw There isn't a better way. There is no sugar for string enums in TS at present (though there are some proposals) - it is functionally equivalent to a "string enum", however. You can also drops large parts of that long-winded declaration if you're willing to forgo some of the enum-likeness (the onyl real required bit is the `export type EnumName = ...(List)...` line, which can have the string literals directly inlined there. The tools should offer completions for string literals nowadays, so you don't _need_ to alias the strings as constants everywhere if you don't want to.

The reason this is needed is that those enums were compatible with numbers rather than strings (since they were enums), meaning that object literals containing fields of those types would never get contextually typed correctly since you had lied to the compiler about what type the field should be (which was causing object literals to need to get cast to the correct spec type all over the place, defeating the type checking completely).

weswigham:
@kanitw I can guarantee the schema changed, actually, in one place, at least - I wanted to ask about that. There was a test which specified using a `'week'` for a timeunit, however there was no such time unit in the type definition. I added it, since the test seems to have been passing.

kanitw:
Okay.  I ll take a closer look at this PR when I get back to computer.

@domoritz Does the schema generator support string literals now?  I wonder if we should just get rid of all enums and use string literals

src/compile/legend.ts
```diff
@@ -162,7 +162,7 @@ export namespace properties {
       delete symbols.opacity;
     }
 
-    let value;
+    let value: {scale: string, field: {prefix?: 'rank'}} | {value: string | number | boolean};;
```
kanitw: VgValueRef

src/compile/mark/valueref.ts
```diff
@@ -112,11 +112,17 @@ defaultRef: VgValueRef | 'base' | 'baseOrMax'): VgValueRef {
 
 export function midX(config: Config): VgValueRef {
   // TODO: For fit-mode, use middle of the width
+  if (typeof config.scale.bandSize === 'string') {
+    throw new Error('midX can not handle string bandSizes');
```
kanitw: Please add 

// TODO: use `band: 0.5` when we migrate to Vega 3

src/compile/mark/valueref.ts
```diff
@@ -112,11 +112,17 @@ defaultRef: VgValueRef | 'base' | 'baseOrMax'): VgValueRef {
...
+  }
   return {value: config.scale.bandSize / 2};
 }
 
 export function midY(config: Config): VgValueRef {
   // TODO: For fit-mode, use middle of the width
+  if (typeof config.scale.bandSize === 'string') {
+    throw new Error('midX can not handle string bandSizes');
```
kanitw: Please add 

// TODO: use `band: 0.5` when we migrate to Vega 3

src/filter.ts
```diff
@@ -105,7 +105,7 @@ export function expression(filter: Filter | string) {
       return fieldExpr + '===' + valueExpr(filter.equal, filter.timeUnit);
     } else if (isOneOfFilter(filter)) {
       // "oneOf" was formerly "in" -- so we need to add backward compatibility
-      const oneOf = filter.oneOf || filter['in'];
+      const oneOf: typeof filter.oneOf = filter.oneOf || filter['in'];
```
kanitw: This line is weird. Why not properly use`OneOfFilter`?  we have an interface for that.

src/transform.ts
```diff
@@ -16,6 +16,12 @@ export interface Transform {
    * Calculate new field(s) using the provided expresssion(s). Calculation are applied before filter.
    */
   calculate?: Formula[];
+
```
kanitw: Please remove this and use `as` where appropriate.  

We intentionally do not add deprecated properties to the Typescript interface so that they won't be generated in the schema.

test/compile/scale.test.ts
```diff
@@ -331,7 +331,7 @@ describe('Scale', function() {
 
     describe('for ordinal', function() {
       it('should return correct domain with the provided sort property', function() {
-        const sortDef = {op: 'min', field:'Acceleration'};
+        const sortDef = {op: 'min' as 'min', field:'Acceleration'};
```
kanitw: Hmm do we really need this `as`?

src/aggregate.ts
```diff
@@ -1,26 +1,52 @@
...
-    MEDIAN = 'median' as any,
-    Q1 = 'q1' as any,
-    Q3 = 'q3' as any,
-    MODESKEW = 'modeskew' as any,
-    MIN = 'min' as any,
-    MAX = 'max' as any,
-    ARGMIN = 'argmin' as any,
-    ARGMAX = 'argmax' as any,
+export namespace AggregateOp {
+    export const VALUES: 'values' = 'values';
```
domoritz: Can we change this and write
```ts
export const COUNT: 'count' = 'count';
...
export type AggregateOp = typeof AggregateOp.ARGMAX | typeof AggregateOp.ARGMIN | ...
```
instead of
```ts
export const COUNT: 'count' = 'count';
export type COUNT = typeof COUNT;
...
export type AggregateOp = AggregateOp.ARGMAX | AggregateOp.ARGMIN | ...
```

weswigham:
Looks like the spec generator handles the new enum-imitating structures correctly in terms of making schema enums out of them, but generates a bad path to the type (namely it includes the filepath in the ref name). Seems fixable?

src/aggregate.ts
weswigham: @domoritz Yes, but the downside is that `AggregateOp.ARGMAX` (and so on) is no longer usable bare in a type position as it was when it was an enum. Is that a fair tradeoff?

kanitw: I think in that case, we should always use `AggregateOp` as the type?

test/compile/scale.test.ts
weswigham: Well. Yes and no. Either the `const sortDef` needs a type annotation of the correct type (which is.... `FieldDef`? Maybe?) to cause `'min'` to be contextually typed as a string literal... or you can just cast it directly to the appropriate string literal type.

src/filter.ts
weswigham: I could! `OneOfFilter` and `typeof filter.oneOf` should be one in the same! Do you prefer `OneOfFilter`?

src/compile/mark/valueref.ts
weswigham: Got it.

src/compile/mark/valueref.ts
weswigham: Got it.

src/compile/common.ts
```diff
@@ -99,7 +99,7 @@ export function applyColorAndOpacity(p, model: UnitModel) {
   }
 }
 
-export function applyConfig(properties, config, propsList: string[]) {
+export function applyConfig(properties: {[index: string]: {value: any}}, config: any, propsList: string[]) {
```
weswigham: Got it!

src/compile/common.ts
```diff
@@ -56,8 +56,8 @@ export function applyColorAndOpacity(p, model: UnitModel) {
     applyMarkConfig(p, model, STROKE_CONFIG);
   }
 
-  let colorValue;
-  let opacityValue;
+  let colorValue: {scale: string, field: {prefix?: 'rank'}} | {value: string | number | true};
+  let opacityValue: {scale: string, field: {prefix?: 'rank'}} | {value: string | number | true};;
```
weswigham: Thanks! I wasn't sure what type these were supposed to be (they were being assigned either object literals or things typed 'any', so it wasn't clear (and I just made a type of the object literals).

src/timeunit.ts
```diff
@@ -3,38 +3,69 @@ import {ScaleType} from './scale';
...
-  YEARQUARTER = 'yearquarter' as any,
-  QUARTERMONTH = 'quartermonth' as any,
-  YEARQUARTERMONTH = 'yearquartermonth' as any,
+export namespace TimeUnit {
+  export const YEAR: 'year' = 'year';
+  export type YEAR = typeof YEAR;
+  export const MONTH: 'month' = 'month';
+  export type MONTH = typeof MONTH;
+  export const WEEK: 'week' =  'week';
+  export type WEEK = typeof WEEK;
```
weswigham: {This line here](https://github.com/vega/vega-lite/blob/master/test/compile/timeunit.test.ts#L44) is the test wrong? It explicitly calls out using `'week'`.

src/filter.ts
kanitw: Yeah `OneOfFilter` is probably better since you can press "Go To Definition" command.  :)

src/transform.ts
kanitw: Actually, since our master now assume work for 2.0 -- we can actually take this change and get rid of all of them.

test/compile/scale.test.ts
kanitw: I see.  
It's not really FieldDef. So let's say keep this for now.  :)

kanitw:
@weswigham  we just include vega-lite-schema.json in the commit in the master.  Maybe you wanna merge that into your branch and re-run `npm run build` and check in the schema diff so we can easily notice the change :)

domoritz:
> Looks like the spec generator handles the new enum-imitating structures correctly in terms of making schema enums out of them, but generates a bad path to the type (namely it includes the filepath in the ref name). Seems fixable?

Should be fixable. Can you create a test for https://github.com/YousefED/typescript-json-schema and maybe submit a pr? I can merge it.

src/timeunit.ts
kanitw: Good catch!  I think we meant "day" there. :)

kanitw:
@domoritz  has created #1659 as a followup to this.

kanitw:
@weswigham  Thanks for you contribution :)

PR #1659, opened 2016-11-03
Dom/no implicit any

domoritz:
New version of #1649

### DIFF
--- a/src/scale.ts
+++ b/src/scale.ts
@@ -29,13 +29,9 @@ export namespace NiceTime {
 export type NiceTime = typeof NiceTime.SECOND | typeof NiceTime.MINUTE | typeof NiceTime.HOUR
   | typeof NiceTime.DAY | typeof NiceTime.WEEK | typeof NiceTime.MONTH | typeof NiceTime.YEAR;
 
-export namespace BandSize {
-  // FIXME: should not be any
-  export const FIT: any = 'fit';
-}
-export type BandSize = typeof BandSize.FIT;
+export const BANDSIZE_FIT: 'fit' = 'fit';
+export type BandSize = typeof BANDSIZE_FIT;
 
-export const BANDSIZE_FIT = BandSize.FIT;
 
 export interface ScaleConfig {
   /**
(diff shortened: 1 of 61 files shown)


src/scale.ts
```diff
@@ -1,30 +1,39 @@
...
+    export const YEAR: 'year' = 'year';
 }
 
-export enum BandSize {
-  FIT = 'fit' as any
+export type NiceTime = typeof NiceTime.SECOND | typeof NiceTime.MINUTE | typeof NiceTime.HOUR
+  | typeof NiceTime.DAY | typeof NiceTime.WEEK | typeof NiceTime.MONTH | typeof NiceTime.YEAR;
+
+export namespace BandSize {
+  // FIXME: should not be any
```
kanitw: Not sure why you add `any` and FIXME here?  Why not just make it right?

vega-lite-schema.json
```diff
@@ -1655,15 +1651,7 @@
...
-                    "minimum": 0,
-                    "anyOf": [
-                        {
-                            "$ref": "#/definitions/BandSize"
-                        },
-                        {
-                            "type": "number"
-                        }
-                    ]
+                    "minimum": 0
```
kanitw: Why does anyOf disappear here?   (Probably related to that FIT is `any` now?)

vega-lite-schema.json
```diff
@@ -2020,7 +2008,7 @@
                 },
                 "titleFontWeight": {
                     "description": "The font weight of the legend title.",
-                    "type": "string"
+                    "type": "number"
```
kanitw: I guess this is either number of string?  (500 or "bold" are both fine?)

vega-lite-schema.json
```diff
@@ -560,11 +560,11 @@
                         false,
                         "year",
                         "month",
-                        "week",
                         "day",
                         "second",
                         "minute",
-                        "hour"
+                        "hour",
+                        "week"
```
domoritz: Why is there still week?

kanitw:
@weswigham  Your contribution just got merged!  Thank you :)

weswigham:
@kanitw Thanks! I may look into enabling strict null checks in the future if I have time - it seems like you guys could really use making null/undefined explicit in the type signatures of your API if you're using it to signal things!

kanitw:
Sure it will be helpful to do so but we just haven't done it because it takes time and not our top priority yet.

PR #1668, opened 2016-11-03
Vg3

domoritz:
Migration to Vega 3.


vega-lite-schema.json
```diff
@@ -2036,43 +2036,13 @@
...
-                            ],
-                            "type": "string"
-                        },
-                        {
-                            "type": "string"
-                        }
-                    ],
-                    "description": "The default symbol shape to use. One of: `\"circle\"` (default), `\"square\"`, `\"cross\"`, `\"diamond\"`, `\"triangle-up\"`, or `\"triangle-down\"`, or a custom SVG path."
+                    "description": "The default symbol shape to use. One of: `\"circle\"` (default), `\"square\"`, `\"cross\"`, `\"diamond\"`, `\"triangle-up\"`, or `\"triangle-down\"`, or a custom SVG path.",
+                    "type": "string"
```
domoritz: @kanitw This is a actually a reasonable optimization but I wanted to call it out to you.

kanitw: I don't know what you mean here.  Can you explain more?

domoritz: Instead of the enum, we now just have the type `string`.

kanitw: Ah make sense now.

PR #1669, opened 2016-11-03
More scale refactor in preparation for Vega 3 migration 

kanitw:


kanitw:
@domoritz  This should be straightforward, so I'm just merging this.  Feel free to send follow up comments

PR #1671, opened 2016-11-03
Scale & Axis Migration MVP

kanitw:
Replace existing ordinal scale with the new 3 discrete scale types
- remove `points` from Scale interface

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "typescript": "^2.0.6",
     "typescript-json-schema": "~0.3.2",
     "uglify-js": "~2.7.4",
-    "vega": "~2.6.3",
+    "vega": "vega/vega",
     "vega-datasets": "vega/vega-datasets#gh-pages",
     "watchify": "~3.7.0",
     "yaml-front-matter": "~3.4.0",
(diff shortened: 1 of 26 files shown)


test/compile/mark/rect.test.ts
```diff
@@ -12,7 +12,7 @@ describe('Mark: Rect', function() {
   describe('simple vertical', function() {
     const model = parseUnitModel({
       "data": {"url": 'data/cars.json'},
-      "mark": "bar",
+      "mark": "rect",
```
domoritz: should be in master, right?

package.json
```diff
@@ -77,7 +77,7 @@
     "typescript": "^2.0.6",
     "typescript-json-schema": "~0.3.2",
     "uglify-js": "~2.7.4",
-    "vega": "~2.6.3",
+    "vega": "vega/vega",
```
domoritz: Can we use the beta version?

kanitw: We should but I don't see it being released yet..

kanitw:
@domoritz  it's not 100% but I put FIXME in all skeptical ones.  I guess we should get something to render first / link editor with Vega 3 so we can debug easier.

kanitw:
So I'll just for now to avoid to many branches

PR #1677, opened 2016-11-04
Make each part of data `assemble` only takes relevant part as input

kanitw:
to allow easier testing.


kanitw:
Turns out we have to modify many parts in the transform's assemble. This simple refactor will allow us to easily write test in the vg3 migration. 

I after we merge this to master I wanna rebase vg3 and its sub-branches

PR #1679, opened 2016-11-05
aggregate & bin transform

kanitw:


jheer:
As an alternative to a formula, you might consider using an inline signal expression to calculate bin midpoints. It will be more memory efficient.

kanitw:
@jheer Yep.  We have an issue to reminder us to do that later, but that require re-structuring many other parts of the code.  For now, we wanna focus on making things render properly first.

PR #1687, opened 2016-11-06
Get rid of formula transform for `bin_mid` and use inline signal instead

### LINKED ISSUE #1681: Get rid of bin_mid formula
Forked from #1665. 

To safe memory, we can use inline signal to calculate this when needed instead.

- [ ] remove `"mid"` from `binSuffix`
  - [ ] For `valueRef.ts`, replace that case with signal that calculate mid point
  - [ ] In other transforms, replace mid with start and end. 
- [ ] Remove mid formula from bin 

## Related Task
- [ ] Revise the condition when we generate `bin_range`

kanitw:
Get rid of formula transform for `bin_mid` and use inline signal instead

Fix #1681


domoritz:
Wanna fix the lint error?

kanitw:
Fixed.

PR #1689, opened 2016-11-06
Scale Update & Refactor

kanitw:
See commits for detail


site/docs/scale.md
```diff
@@ -153,17 +153,12 @@ __Default value:__
...
+### Built-in Color Scheme
```
domoritz: I think palette, range, or scale are the right words, not color scheme.

src/compile/scale.ts
```diff
@@ -41,35 +41,132 @@ export type ScaleComponents = {
   binColorLegend?: ScaleComponent
 }
 
-export function initScale(topLevelSize: number, mark: Mark, channel: Channel, fieldDef: ChannelDefWithScale, scaleConfig: ScaleConfig): Scale {
-  let scale: Scale = duplicate((fieldDef || {}).scale || {});
+export function channelScalePropertyIncompatability(channel: Channel, propName: string): string {
```
domoritz: This function should return `{compatible: bool, message: string}` instead of undefined for valid. See https://basarat.gitbooks.io/typescript/content/docs/tips/null.html

src/compile/scale.ts
```diff
@@ -191,79 +292,114 @@ export function type(specifiedType: ScaleType, fieldDef: FieldDef, channel: Chan
...
+            return ordinalToContinuousType(channel, mark, canHaveBandSize);
+        }
+        return ScaleType.TIME;
+
+      case QUANTITATIVE:
+        return ScaleType.LINEAR;
+    }
+
+    // should never reach this
+    return null;
```
domoritz: Not sure whether you can sue never here. https://basarat.gitbooks.io/typescript/content/docs/types/never.html

src/compile/scale.ts
```diff
@@ -191,79 +292,114 @@ export function type(specifiedType: ScaleType, fieldDef: FieldDef, channel: Chan
...
+      if (
+        // 1) the mark is bar and the scale's bandWidth is 'fit',
+        (mark === BAR && !canHaveBandSize) ||
+        // 2) the mark is rect
+        mark === RECT
+      ) {
+        return ScaleType.BAND;
       }
+    }
+    // Otherwise use ordinal point scale
```
domoritz: I don't like these comments that just explain code. Better say why something is done the way it is done.

src/compile/scale.ts
```diff
@@ -515,20 +666,18 @@ export function rangeMixins(scale: Scale, model: Model, channel: Channel): any {
...
       // else -- ordinal, time, or quantitative
-      // FIXME: this is probably not correct for the new sequential scale
-      return {range: scaleConfig.sequentialColorRange};
+      return {scheme: scaleConfig.sequentialColorScheme};
 
     case OPACITY:
       return {range: scaleConfig.opacity};
   }
-  return {};
+  return undefined;
```
domoritz: Why undefined and not {}?

src/log.ts
```diff
@@ -132,7 +132,16 @@ export namespace message {
   // SCALE
   export const CANNOT_UNION_CUSTOM_DOMAIN_WITH_FIELD_DOMAIN = 'custom domain scale cannot be unioned with default field-based domain';
 
-  export const CANNOT_USE_PADDING_WITH_FACET = 'Cannot use padding with facet\'s scale.  Please use spacing instead.';
+  export const CANNOT_USE_SCHEME_WITH_NON_COLOR = 'Cannot use scheme with non-color channel.';
```
domoritz: I don't like the word scheme here as explained above.

site/docs/scale.md
kanitw: Then you would have to protest at the D3 and Vega level: 

- https://github.com/d3/d3-scale-chromatic.  
- https://github.com/vega/vega/blob/master/spec/heatmap.vg.json#L87

src/log.ts
kanitw: Not my decision.

domoritz: k

src/compile/scale.ts
kanitw: Actually I should throw error instead :) will do that and rebase.

src/compile/scale.ts
kanitw: I'll just throw error instead. 

From both the link you suggested and [TS example](https://www.typescriptlang.org/docs/handbook/basic-types.html), I haven't seen any case where they show that you should call `return never` directly.

src/compile/scale.ts
kanitw: I don't think my code perfectly matches the example in that case.  

There, toInt should return number.  My function returns a string describe incompatibility.  If there is no incompability, then undefined seems fine.  

Of course, you can rewrite this method in a similar fashion to that example, but TBH it seems like an overkill.  I'm not convinced that it's necessary better and worth my effort doing it.

src/compile/scale.ts
domoritz: I don't think you can. But throwing an exception creates a never.

src/compile/scale.ts
kanitw: I'll revise this part in another follow up PR.

kanitw:
@domoritz  This should be ready to merged.  Most comments addressed except two comments.
One of them I disagree and think it's a waste of time to do.  Another will be handle in a follow up PR.

kanitw:
Re: rebasing -- I think we should rebase vg3 to latest master after merging this PR to reduce redundant rebase efforts.

domoritz:
I'm just waiting for travis. Other than that it looks good.

PR #1712, opened 2016-11-24
Correct our reproduction of becker's barley

kanitw:
Now the aggregation function, sorting order, and scale (not starting from zero) are similar to the original plot!

cc: @jakevdp -- I saw that you use this example in Altair example too. You might want to consider updating.  (Btw, Happy Thanksgiving!)

vega_editor_and_hci_stanford_edu_courses_cs448b_papers_becker-trellis-jcgs_pdf

We should cherry pick this to 1.x if we were to update the github pages too.

### DIFF
--- a/examples/specs/trellis_barley.vl.json
+++ b/examples/specs/trellis_barley.vl.json
@@ -3,11 +3,17 @@
   "data": {"url": "data/barley.json"},
   "mark": "point",
   "encoding": {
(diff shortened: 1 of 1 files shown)


jakevdp:
Fortunately the Altair example is auto-generated from the Vega-Lite example, so this should happen automatically once I update the dependency :smile:

PR #1717, opened 2016-11-24
Use faster json schema validator and remove color format.

domoritz:


kanitw:
Why remove color format?

domoritz:
We didn't set it consistently and so I'd rather remove it. Is there a specific use case for it?

kanitw:
I don't know. I think you added it?

domoritz:
I added it when we still used it to generate the properties editor in polestar.

kanitw:
@domoritz oops.  Then it's not good to remove it then!

kanitw:
It's weird that you're suggesting to remove it despite the obvious utility!

kanitw:
I know we don't have property editor right now, but I can't imagine away to know if a field is color in polestar without this!

domoritz:
Not reliable but you could assume that stroke, fill, and color are all colors. Anyway, I'm happy to add it again if we need it but since we didn't use it consistently, I removed it.

kanitw:
Instead of removing it, the right solution is to make it consistent ....

domoritz:
There are two consistent states: use it nowhere and ensure that whenever we use a color, we specify the format.

kanitw:
And the "use it nowhere" is not the ideal one...

Issue #1711, opened 2016-11-23
Default sizing of circles + Legend of Size When Zero is included


domoritz:
```json
{
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"a": "A","b": 10},
      {"a": "B","b": 20},
      {"a": "C","b": 30},
      {"a": "D","b": 40}
    ]
  },
  "mark": "circle",
  "encoding": {
    "x": {"field": "a", "type": "ordinal"},
    "size": {"field": "b", "type": "quantitative"}
  }
}
```
<img width="191" alt="screen shot 2016-11-22 at 21 39 21" src="https://cloud.githubusercontent.com/assets/589034/20552095/260c349a-b0fc-11e6-946f-2b16ab3c2f28.png">

Shouldn't we use a sqrt scale by default to accurately represent the values?

kanitw:
[The size value of Vega already maps to area](https://github.com/vega/vega/wiki/Marks#symbol), so using linear scale is already correct.  

The culprit is that I didn't make the scale include zero by default and set range to start from 0.  

For now, manually overriding the scale with zero=true, range starting from zero produce reasonable output:

```js
{
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"a": "A","b": 10},
      {"a": "B","b": 20},
      {"a": "C","b": 30},
      {"a": "D","b": 40}
    ]
  },
  "mark": "circle",
  "encoding": {
    "x": {"field": "a", "type": "ordinal"},
    "size": {"field": "b", "type": "quantitative", "scale": {"zero": true, "range": [0, 361]}}
  }
}
```

![vega_editor](https://cloud.githubusercontent.com/assets/111269/20552569/3e71cef0-b101-11e6-98d1-594ecd670c7f.png)

I will change the default when I get a chance.  Maybe it's worth doing a hot fix to 1.x as well?

domoritz:
Here is an example where it's easy to see that scale is correct. 

<img width="155" alt="screen shot 2016-11-22 at 22 21 07" src="https://cloud.githubusercontent.com/assets/589034/20552668/03c027b0-b102-11e6-9dd5-b8ad15036803.png">

We should definitely include 0 by default. Let's not backport this to 1.x as it's technically not wrong but just confusing.

kanitw:
The default is wrong...

kanitw:
[...]

PR #1720, opened 2016-11-25
[WIP] Add rule for using quantitative field with size

kanitw:
Address part of #1711 (Still haven't test stuff related to legend, but it's not in the scope of this PR)

- Scale's `zero` should be true by default for `size`
- Break size's range calculation into sizeRangeMin and sizeRangeMax
  - `sizeRangeMin` depends on scale.zero
  - `sizeRangeMax` depends on scale.bandSize (rangeStep)
- Replace scale configs `bar/font/rule/tick/pointSizeRange` with `minBar/Tick/Rule/Point/TextSize` (and `max...`) such as `minBarSize`

Questions:
- I think we should consolidate mark-specific mark config `config.mark.barXXX,tickXXX` with the `config.scale.{min/max}{Bar/Tick/Rule/Point/Text}Size` above?
   - For example, we could have `config.bar.size` (if size is not mapped) and `config.bar.sizeRangeMin/Max` (or shorter `config.bar.min/maxSize`). This will make configs way more concise.  Plus, this way `config.mark.*`and `config.scale.*` can serve as generic mark and scale config, while mark-specific config can be `config.<markname>.*`.  By separating this, we can also have bar use blue color by default while line use grey color by default as well.  
  - If so, I can do that in a follow-up PR

- `minTextSize` or `minFontSize` ?
   - fontSize would be consistent with how people call textSize
   - minTextSize would be consistent with min<Channel>Size format
   - If we agree with the proposal above, this could become just `config.text.size(Range)Min` 

(What do you think?)

- [x] Need to update schema once @domoritz  update merge the schema in vg3 branch

### DIFF
--- a/src/compile/scale.ts
+++ b/src/compile/scale.ts
@@ -228,7 +228,7 @@ function parseMainScale(model: Model, fieldDef: FieldDef, channel: Channel) {
     scaleDef.domain = domain(scale, model, channel);
   }
 
-  // TODO: move range to init
+  // TODO: move range to init, make it come after zero (rangeMixins depends on zero).
   extend(scaleDef, rangeMixins(scale, model, channel));
 
   if (sort && (isSortField(sort) ? sort.order : sort) === SortOrder.DESCENDING) {
@@ -292,7 +292,7 @@ export function type(specifiedType: ScaleType, fieldDef: FieldDef, channel: Chan
   }
 
   if (specifiedType !== undefined) {
-    // Check if explicitly specified scale type is supported
+    // Check if explicitly specified scale type is supported by the channel
     if (supportScaleType(channel, specifiedType)) {
       return specifiedType;
     } else {
@@ -305,7 +305,7 @@ export function type(specifiedType: ScaleType, fieldDef: FieldDef, channel: Chan
   return defaultProperty.type(fieldDef, channel, mark, canHaveBandSize);
 }
 
-namespace defaultProperty {
+export namespace defaultProperty {
   /**
    * Determine appropriate default scale type.
    */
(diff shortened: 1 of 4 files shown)


kanitw:
Note: I know that we don't hit the diff target here, but since I'm considering making size for bar and tick proportional to rangeStep (currently `bandSize`), not sure if it's worth writing test yet.

kanitw:
NVM -- let me fix them first. This should be very easy

src/compile/scale.ts
```diff
@@ -684,20 +677,85 @@ export function rangeMixins(scale: Scale, model: Model, channel: Channel):
...
+    case 'tick':
+      return scaleConfig.minTickSize;
+    case 'rule':
+      return scaleConfig.minRuleSize;
+    case 'text':
+      return scaleConfig.minTextSize;
+    case 'point':
+    case 'square':
+    case 'circle':
+      return scaleConfig.minPointSize;
```
domoritz: This means that the min size is not 0 by default even if the smallest domain value is 0. I'm okay with that but just want to acknowledge it.

src/scale.ts
```diff
@@ -136,20 +136,67 @@ export interface ScaleConfig {
   /** Default range for shape */
   shapeRange?: string[];
 
-  /** Default range for bar size scale */
-  barSizeRange?: number[];
+  /**
+   * Default minimum range for bar size scale with zero=false.
+   * @minimum 1
```
domoritz: Why not allow 0? Doesn't hurt to allow 0 even if zero=false.

kanitw: In this case, users set zero=false thus they don't expect domain to include zero.  
(Otherwise, they wouldn't bother to override the zero value.)

Thus, if you want to set a theme for such cases, I don't think including setting rangeMin in a config is a good idea. 

In any case, users can explicitly set size's scale range to start from zero. 

That said, if you still prefer that we don't include `@minimum 1`, I'm happy to change.

kanitw:
@arvind @domoritz -- let me know what you think about questions above. 

(I think the proposal above is close to what @arvind asks a while ago.)

domoritz:
LGTM

kanitw:
Created https://github.com/vega/vega-lite/issues/1720 for further discussion.

domoritz:
#1727

PR #1726, opened 2016-11-27
Replace all uses of templates with signals.

domoritz:


kanitw:
LGTM except that I think we should use string template (``) to avoid `\'` and `\"`

domoritz:
Done. Let's keep the two commits separate.

domoritz:
Not sure why you squashed the commits but okay.

kanitw:
I look at the diff and it seems natural to have this in one commit? 

The main semantic = template => signal -- the string template seems more like a stylistic change of the main change. So I prefer to lessen our work when we read through commit lists in vg3 to update release note.  :)

PR #1730, opened 2016-11-28
type is called scale

domoritz:


kanitw:
This doesn't look right?  The new Axis doesn't have type.  Scale is an old property that refers to the scale that the axis visualizes.

domoritz:
Okay, then I will rename https://github.com/vega/vega-lite/pull/1730/files#diff-f1371c09748e7113774766ab4a94fda2L81 instead

kanitw:
Not sure what you mean lol

PR #1733, opened 2016-11-29
Move mark-specific size scaleRange to mark specific config

kanitw:
See commits for details

### DIFF
--- a/src/compile/scale.ts
+++ b/src/compile/scale.ts
@@ -6,7 +6,7 @@ import {Config} from '../config';
 import {SOURCE, STACKED_SCALE} from '../data';
 import {DateTime, isDateTime, timestamp} from '../datetime';
 import {ChannelDefWithScale, FieldDef, field} from '../fielddef';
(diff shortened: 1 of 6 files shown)


kanitw:
I'll add more test on the plane.   Let me know what you think about `opacities`

kanitw:
Also let me know what you think about these properties names.  I just try to use full property name with min/max. (Size is usually a short for fontSize/strokeWidth/etc.)

domoritz:
`config.mark.opacities` sounds odd. I think `opacityRange` is more intuitive.

kanitw:
Okie, but we should still move it to config.mark right?

domoritz:
I don't have strong opinions about that.

kanitw:
@domoritz  Ok, I moved all scale range related stuff to mark and squashed that into the first commit.

kanitw:
Note that the second commit is for related but different fix, so ideally shouldn't be squashed -- but not really a big deal.

Issue #1642, opened 2016-11-01
Correctly import subfiles in vega-util


kanitw:
(Currently we import the whole compiled file)

domoritz:
Possible this get's fixed by using a different compiler e.g. rollup.

kanitw:
Or webpack

domoritz:
Why is it better to import only subfiles? I think we almost use all of it at this point.

jheer:
Probably better _not_ to import subfiles -- use the publicly exported API. Vega 3 uses all of it, so in almost all deployments you'll need the whole library anyway.

kanitw:
The real problem we try to tackle here is to minimize file size, esp. redundancy with Vega.  I'll create a new issue for that.

PR #1735, opened 2016-11-29
Remove datalib dependency

### LINKED ISSUE #1569: Replace Datalib Util with Vega Util (and import only necessary methods)

domoritz:
Fix #1569


kanitw:
This is great.  I assume you'll deal with individual import later (#1642)?  If so, feel free to merge!

PR #1739, opened 2016-11-30
Steps to make histogram work

kanitw:
- Add extent transform if `bin.extent` is not specified
- Rename `div` to `divide`
- Add more JSON schema annotation


kanitw:
@domoritz  I still can't get histogram to work due to a mysterious Vega error -- "Error: Operator not defined: -1"

kanitw:
@domoritz This PR should be mergable.  At least all others obvious errors in the histogram spec is gone. 

 https://github.com/vega/vega/issues/668.

PR #1740, opened 2016-11-30
[WIP] Avoid referring to null data component

kanitw:
- [ ] TODO: check other data transforms


kanitw:
Now I wonder why that object can be undefined ...

PR #1741, opened 2016-11-30
Aggregate tranform: `summarize` => `fields` and `ops`

kanitw:


kanitw:
I'm gonna just merge this since all of these lines will be further changed soon, doesn't make sense to write tests now.

PR #1744, opened 2016-12-02
Rename scale's `bandSize` to `rangeStep`

kanitw:
(very likely this will fail codecov test for diff coverage since it touches so many files)


domoritz:
Upping your loc count, I see :-P

kanitw:
> Upping your loc count, I see :-P

Hard to beat adding / removing interface like you lol. 

Honestly, this PR is a pretty boring task. If you had done it for me, I'd greatly appreciate. (Too late!)

src/scale.ts
```diff
@@ -90,13 +90,13 @@ export interface ScaleConfig {
    *  Default band width for `x` ordinal scale when is mark is `text`.
```
domoritz: Are we changing the descriptions as well?

kanitw: Good catch, I'll search for all occurrences for band width and band size

kanitw:
@domoritz Ok force pushed the fix for description

test/compile/mark/bar.test.ts
```diff
@@ -236,7 +236,7 @@ describe('Mark: Bar', function() {
     });
     const props = bar.properties(model);
 
-    it('should use x and with with band size true and 0.5 offset', () => {
+    it('should use y and with band true', () => {
```
domoritz: "should use y with band true"

domoritz:
LGTM after the one comment fix.

PR #1748, opened 2016-12-02
Slightly modify scale.domain() + add explanation

kanitw:
Note that you should not always use bin_mid. 

For example, when you use ordinal scale with x or y, you don't want to add bin_scale for axis to show correct label if not necessary.  If you don't want to do it for axis, I don't see why you want to do it for label unless for really necessary case (e.g., quantitative-based mark properties like color, size, opacity.)


src/compile/scale.ts
```diff
@@ -477,21 +478,23 @@ export function domain(scale: Scale, model: Model, channel:Channel): any {
...
+      if (channel === X || channel === Y) {
+        // X/Y position have to include start and end for non-ordinal scale
+        return {
+          data: model.dataTable(),
+          fields: [
+            model.field(channel, { binSuffix: 'start' }),
+            model.field(channel, { binSuffix: 'end' })
+          ]
+        };
+      } else {
```
domoritz: I think this should be `contains(SCALE_CHANNELS, channel)`

PR #1751, opened 2016-12-03
Add paddingInner/Outer + Correct Layout Formula to match D3

kanitw:


jheer:
FYI your layout formula does not match what I do in the vega/vega-encode Scale operator. I include an additional special case to better handle scales with cardinality <= 1, in which case the inner padding term should be suppressed.

kanitw:
Thank you @jheer.   

I'm curious why the inner padding term should be dropped for cardinality = 1. 
(For zero, it makes sense.)

For example, I tried to vary the number of discrete values for a bar chart in [this spec](http://pastebin.com/R0TAsztX).

The band size of the bars remains constant when cardinality > 1.

vega_editor
vega_editor



However, when cardinality = 1, the size of the bar becomes thicker and no longer maintain the relationship that `bandSize = (1-paddingInner) * rangeStep`

vega_editor

On the other hand, if I use the formula that keep the paddingInner term and set `range` to the calculcate `width` instead of using setting `rangeStep` (thus do not use Vega 3's current formula), I get the same bandSize compared to when cardinality = 3 or 4.  
(See spec [here](http://pastebin.com/YxgiSeNE))

vega_editor

Thus, I wonder if this changing band size is an intended behavior.  If so, could you please explain why?

jheer:
It looks like we're both right and we're both wrong :)

I think the difference is that your test example includes outer padding, while mine didn't. These interact differently with the step size calculation in D3:

```
step = (stop - start) / Math.max(1, n - paddingInner + paddingOuter * 2);
```

When `n == 1`, `paddingInner > 0`, and  `paddingOuter == 0`, the second argument to `Math.max` is less than one, and so the thresholding occurs. In my example, zero outer padding causes bar widths to change under your current layout formula. (I'm not 100% sure why the `max` call is included here, perhaps to deal with the `n == 0` case?)

One solution for now is to change the special casing logic to be slightly more nuanced. I believe the following adjustment works in the presence of outer padding.

```
space = count - inner + outer * 2;
if (space < 1) {
  // should only happen when count <= 1
  space = count + outer * 2;
} 
rangeMax = rangeStep * space;
```

kanitw:
Here is my analysis for a case where the formula could still be wrong:

```
space = count - inner + outer * 2;
if (space < 1) {
  // should only happen when count <= 1
  space = count + outer * 2;
} 
rangeMax = rangeStep * space;
```

I think `Math.max(1,  n - paddingInner + paddingOuter * 2)` was to prevent the calculated step from being larger than the input `rangeSpan` (`stop-start`) when `n - paddingInner + paddingOuter * 2` < 1.  

From first glance, I think this constraint seems to make sense from a perspective that users don't input `step`, so it's weird that `step` can be larger than the input range span.  However, this constraint can be problematic as discussed below. 


Below, I try to analyze all the possible problematic case that we need to consider. 

For this analysis, I define that `rangeSpan = rangeMax - rangeMin = stop -start`. (This is equivalent to `rangeMax` if `rangeMin` = 0)

I also define that `stepCount = rangeSpan / rangeStep`. (This is the same as "space" in your equation, but I found "stepCount" easier for me to think about.)

If I understand correctly, Vega calculates `range` and pass the `range` to the underlying d3-based scale, which do not take `step` as input. (I say d3-based scale because it seems like you modify band scale a bit in `vega-scale`.)

## 1) n - inner + outer * 2 >= 1

In this case, the standard formula `rangeSpan = rangeStep * (n - inner + outer * 2)` works fine, no need to worry. 

## 2) n - inner + outer * 2 < 1

In this case, D3 will use:

```
step = rangeSpan /  Math.max(1, n - paddingInner + paddingOuter * 2) 
     = rangeSpan
```

Let's consider when n==0 and n==1. 

(`n - inner + outer * 2 < 1` is impossible when n>=2. Since `- inner + outer * 2 > -1`, adding n to both sides will make `n - inner + outer * 2 > 1`.)

### 2a) n == 1

Since `n - inner + outer * 2 < 1`, your formula will calculate `rangeSpan = rangeStep * (1 + outer * 2)`.

The underlying d3-based scale will calculate `step = rangeSpan = rangeStep * (1 + paddingOuter * 2)`.
Here the output `step` != input `rangeStep` (unless paddingOuter = 0).

This will result in different bandWidth
```
bandWidth =  step * (1 - paddingInner)
          =  rangeSpan * (1 - paddingInner)                                 ---- (1)
          =  rangeStep * (1 + paddingOuter * 2) * (1 - paddingInner)
```

While for `n>=2` case, `bandWidth = rangeStep * (1-paddingInner)`.  
Thus, I believe that we will still get inconsistent bandWidth/Size problem.  

Example: 

outer = 0.2, inner = 0.6, n = 1, rangeStep = 20

**For n=1**

Vega: rangeSpan = 20 * (1 + 0.2*2) = 28   // ignoring -0.6 term
D3: bandWidth = step * (1 - paddingInner)
              = rangeSpan * (1 - paddingInner)          (from (1))
              = 28 * (1-0.6) = 11.2

**For n=2**
Vega: rangeSpan = 20 * (2 - 0.6 + 0.2*2) = 36
D3: bandWidth = step * (1 - paddingInner)
              = rangeSpan / (n - inner + outer * 2) * (1 - paddingInner)
              = 36 / (2 - 0.6 + 0.2 * 2) * (1 - 0.6)
              = 8

### 2b) n == 0

I think your proposed formula make sense since it will produce `rangeMax = rangeStep * (paddingOuter * 2)`, which is what we would expect from sketching by hand. 

However, the underlying d3-based scale will calculate `step = rangeMax = rangeStep * (paddingOuter * 2)`.
Again, the output `step` != input `rangeStep` (unless `paddingOuter = 0.5`).

That said, I don't really know if we have to worry about this case since it is not a big deal since 
the plot will be empty anyway, this discrepancy will probably be invisible.

Plus, in Vega-Lite, we currently do not show anything if the data source is empty due to a different problem (#1752).

jheer:
As discussed on Slack, I've updated and re-published vega-encode with a modified layout formula that exactly matches D3 and ensures consistent step / bandwidth values. See vega/vega-encode@eb0fda56f423590b550f7431d7bc01b07f716a51.

kanitw:
Thank you @jheer.  I played with the examples and observed that we now get consistent bandWidth values.

However, I have two *minor* notes: 

1) The formula in veda-encode is currently not exact mirror of vega-scale due to the `count ? ` condition in 

```
space = count ? Math.max(1, count - inner + outer * 2) : 0;
```

That said, I guess that you intend to have range = 0 when count = 0, which is reasonable so that we don't draw short line for empty ordinal axis.  So I don't think this is wrong, but just wanna point out that it's not an exact match. 

2) Outer padding can still be inconsistent when moving between n=1 to n=2.

For example, with step = 40, inner = 0.6, outer = 0.2, count = 1,2 

pasted_image_12_3_16__8_49_pm


pasted_image_12_3_16__8_49_pm

Instead, I tried to remove `max(1, ...)` from the formula in https://github.com/vega/vega-encode/pull/1 and https://github.com/vega/vega-scale/pull/1.  Now we also have consistent outer padding.

pasted_image_12_3_16__9_58_pm


pasted_image_12_3_16__8_49_pm

(I also edit the code comment to point out the difference between formula in veda-encode and in  vega-scale.)

src/scale.ts
```diff
@@ -182,15 +191,32 @@ export interface Scale {
 
   /**
    * (For `row` and `column` only) A pixel value for padding between cells in the trellis plots.
-   * @type {number}
+   * @type {integer}
```
domoritz: I fixed this :-P

PR #1759, opened 2016-12-04
Remove `RANGESTEP_FIT` 

### LINKED ISSUE #1743: Remove RANGESTEP_FIT
`null` should actually mean the same thing.

kanitw:
Remove `RANGESTEP_FIT` since we can simply use `null` to specify the same thing without making rangeStep has two types

Fix #1743

(We need to merge #1751 first before reviewing this.)


vega-lite-schema.json
```diff
@@ -2563,19 +2553,9 @@
...
-                            "type": "string"
-                        },
-                        {
-                            "type": "number"
-                        }
-                    ],
                     "description": "Default range step for (1) `y` ordinal scale,\nand (2) `x` ordinal scale when the mark is not `text`.",
-                    "minimum": 0
+                    "minimum": 0,
+                    "type": "number"
```
kanitw: @domoritz  looks like there is a bug in `typescript-json-schema` ?   

This is from `number | null` and thus should include `null` in the schema.

domoritz: Not until we enable strict null checks. you have to add @nullable.

kanitw: Thanks, will add that

PR #1767, opened 2016-12-11
TS 2.1 compatible

domoritz:


src/channel.ts
```diff
@@ -53,12 +53,11 @@ export const ORDER = Channel.ORDER;
...
-export const UNIT_CHANNELS = without(CHANNELS, [ROW, COLUMN]);
-export const UNIT_SCALE_CHANNELS = without(UNIT_CHANNELS, [X2, Y2, PATH, ORDER, DETAIL, TEXT, LABEL, X2, Y2]);
-export const SCALE_CHANNELS = UNIT_SCALE_CHANNELS.concat([ROW, COLUMN]);
-export const NONSPATIAL_CHANNELS = without(UNIT_CHANNELS, [X, Y, X2, Y2]);
-export const NONSPATIAL_SCALE_CHANNELS = without(UNIT_SCALE_CHANNELS, [X, Y]);
+export const UNIT_CHANNELS = [X, Y, X2, Y2, SIZE, SHAPE, COLOR, PATH, ORDER, OPACITY, TEXT, DETAIL, LABEL];  // CHANNELS without COLUMN, ROW
+export const UNIT_SCALE_CHANNELS = [X, Y, SIZE, SHAPE, COLOR, OPACITY]; // UNIT_CHANNELS without X2, Y2, PATH, ORDER, DETAIL, TEXT, LABEL
+export const SCALE_CHANNELS = [X, Y, SIZE, SHAPE, COLOR, OPACITY, ROW, COLUMN];  // UNIT_SCALE_CHANNELS with ROW, COLUMN
+export const NONSPATIAL_CHANNELS = [SIZE, SHAPE, COLOR, PATH, ORDER, OPACITY, TEXT, DETAIL, LABEL]; // UNIT_CHANNELS without X, Y, X2, Y2;
+export const NONSPATIAL_SCALE_CHANNELS = [SIZE, SHAPE, COLOR, OPACITY];  // UNIT_SCALE_CHANNELS without X, Y;
```
domoritz: The advantage of explicitly listing the channels is that the types are correct and so we don't have weird errors in other places.

kanitw:
Ping @domoritz: Need rebase & fixes :)

domoritz:
Is this the way you want it, though? Just wondering since this pr is a month old.

domoritz:
Rebased.

kanitw:
Looking at the diff again, I wonder if  any of these changes are required to make TS 2.1 compatible?

domoritz:
Yes, they are.

kanitw:
I doubt if they are the only way

domoritz:
Probably not. Can you be more specific about which changes you think are problematic.

kanitw:
Already commented ...

domoritz:
> Already commented ...

Emmm, no.

package.json
```diff
@@ -77,7 +77,7 @@
     "source-map-support": "~0.4.6",
     "tsify": "~2.0.2",
     "tslint": "~3.15.1",
-    "typescript": "2.0.10",
+    "typescript": "^2.1.4",
```
domoritz: Working on it. But that's not a problem.

src/channel.ts
```diff
@@ -52,12 +52,11 @@ export const ORDER = Channel.ORDER;
 export const OPACITY = Channel.OPACITY;
 
 export const CHANNELS = [X, Y, X2, Y2, ROW, COLUMN, SIZE, SHAPE, COLOR, ORDER, OPACITY, TEXT, DETAIL, LABEL];
-
-export const UNIT_CHANNELS = without(CHANNELS, [ROW, COLUMN]);
-export const UNIT_SCALE_CHANNELS = without(UNIT_CHANNELS, [X2, Y2, ORDER, DETAIL, TEXT, LABEL, X2, Y2]);
-export const SCALE_CHANNELS = UNIT_SCALE_CHANNELS.concat([ROW, COLUMN]);
-export const NONSPATIAL_CHANNELS = without(UNIT_CHANNELS, [X, Y, X2, Y2]);
-export const NONSPATIAL_SCALE_CHANNELS = without(UNIT_SCALE_CHANNELS, [X, Y]);
+export const UNIT_CHANNELS = [X, Y, X2, Y2, SIZE, SHAPE, COLOR, ORDER, OPACITY, TEXT, DETAIL, LABEL];  // CHANNELS without COLUMN, ROW
```
domoritz: But it provides correct types. See old comment.

src/encoding.ts
```diff
@@ -98,15 +98,15 @@ export interface Encoding extends UnitEncoding {
 export function has(encoding: Encoding, channel: Channel): boolean {
   const channelEncoding = encoding && encoding[channel];
   return channelEncoding && (
-    channelEncoding.field !== undefined ||
+    'field' in channelEncoding ||
```
domoritz: Good point.

src/channel.ts
kanitw: I don't think this is the right way to fix correct type. It will lead to disastrous bugs that are hard to detect later on.  

I guess that you can just add as Channel[] and keep the same code.  (or try call `without<Channel>(...)`) or something.

src/encoding.ts
```diff
@@ -98,15 +98,15 @@ export interface Encoding extends UnitEncoding {
...
     // TODO: check that we have field in the array
     (isArray(channelEncoding) && channelEncoding.length > 0)
   );
 }
 
 export function isAggregate(encoding: Encoding) {
   return some(CHANNELS, (channel) => {
-    if (has(encoding, channel) && encoding[channel].aggregate) {
+    if (has(encoding, channel) && 'aggregate' in encoding[channel]) {
```
domoritz: Error TS2339: Property 'aggregate' does not exist on type 'FieldDef | PositionChannelDef | ChannelDefWithLegend | OrderChannelDef | FieldDef[] | OrderChanne...'.
  Property 'aggregate' does not exist on type 'FieldDef[]'.

src/encoding.ts
domoritz: Error TS2339: Property 'field' does not exist on type 'FieldDef | PositionChannelDef | ChannelDefWithLegend | OrderChannelDef | FieldDef[] | OrderChanne...'.
  Property 'field' does not exist on type 'FieldDef[]'.

src/stack.ts
```diff
@@ -51,7 +51,7 @@ export function stack(mark: Mark, encoding: Encoding, stacked: StackOffset): Sta
 
   // Should have grouping level of detail
   const stackByChannels = STACK_GROUP_CHANNELS.reduce((sc, channel) => {
-    if (has(encoding, channel) && !encoding[channel].aggregate) {
+    if (has(encoding, channel) && !('aggregate' in encoding[channel])) {
```
domoritz: Error TS2339: Property 'aggregate' does not exist on type 'FieldDef | ChannelDefWithLegend | OrderChannelDef | FieldDef[] | OrderChannelDef[]'.
  Property 'aggregate' does not exist on type 'FieldDef[]'.

src/channel.ts
domoritz: No, it doesn't compile. 

```
TypeScript error: src/compile/unit.ts(169,60): Error TS2345: Argument of type 'FieldDef | PositionChannelDef | ChannelDefWithLegend | OrderChannelDef | FieldDef[] | OrderChanne...' is not assignable to parameter of type 'ChannelDefWithScale'.
  Type 'FieldDef[]' is not assignable to type 'ChannelDefWithScale'.
    Types of property 'sort' are incompatible.
      Type '(compareFn?: (a: FieldDef, b: FieldDef) => number) => FieldDef[]' is not assignable to type '"ascending" | "descending" | SortField'.
        Type '(compareFn?: (a: FieldDef, b: FieldDef) => number) => FieldDef[]' is not assignable to type 'SortField'.
          Property 'field' is missing in type '(compareFn?: (a: FieldDef, b: FieldDef) => number) => FieldDef[]'.
```

domoritz: `without` would be great if it worked but it actually doesn't do a type subtraction (because it cannot do it in the general case).

kanitw: Why do we need type subtraction?  Isn't `Channel[]` sufficient?

package.json
kanitw: Why `^2.1.4`?  Not just `2.1.4`?

domoritz: `yarn upgrade typescript` did that. Since we use yarn, it doesn't matter.

src/channel.ts
domoritz: Try it.

kanitw:
Replaced with #1814, which does not require changing channel definitions.

Issue #1597, opened 2016-10-19
Vega 3 Migration MVP


kanitw:
## Marks and Visual Encoding
- [x] The `"properties"` block of visual encodings has been renamed `"encode"`.

- [x] The `"template"` encoding directive has been removed. Instead, use the `"signal"` directive with an expression to perform equivalent text processing. (The `"signal"` encoding directive now accepts anonymous signal expressions in addition to signal names. You can use these to directly write expressions for visual encodings. For example: `{"signal": "datum.field * 2"}`.) – @domoritz 

- ~~The `"band"` property now accepts an interpolation fraction between 0 and 1 for placing elements within a scale band, and can now be used in conjunction with a field-based scale mapping. For example, the following encoding places an item at the midpoint of the appropriate scale band: `{"scale": "xscale", "field": "foo", "band": 0.5}`.~~

- [ ] ~~The `rule` mark type now supports arbitrary line segments, and reliably draws a line from point `"x"`, `"y"` to point `"x2"`, `"y2"`.~~ (later in #1665) 
- [x] [@kanitw] Mark definitions no longer allow embedded data transforms (e.g., `{"type": "rect", "from": {"data": "table, "transform": [...]}}`). Instead, _all_ derived data sources must now be defined within a `"data"` definition block. However, as of Vega 3, `"data"` blocks are no longer constrained to only the top-level of a spec: they can be defined within `group` marks as well! For example, this flexibility allows creation of derived data sets _within_ a faceted group.
- [x] [@kanitw] The `"facet"` transform is no longer supported. Instead, a new `"facet"` directive can be applied within a mark `"from"` block to enable multiple forms of faceting support. For example, the `barley.vg.json` example facets a group mark like so:
  
  ``` js
  "from": {
    "facet": {
      "name": "sites",
      "data": "barley",
      "groupby": "site"
    }
  },
  ```
  
  Child marks can visualize the faceted data by name: `"from": {"data": "sites"}`. The `"facet"` block also accepts the same parameters as an `"aggregate"` transform to apply when generating data elements for `group` mark items. In addition, Vega 3 now supports pre-faceted data, in which a data tuple may contain a nested a set of records. This oft-requested feature allows data to be grouped offline and passed directly to Vega, like so:
  
  ``` js
  "from": {
    "facet": {
      "name":  "facets",
      "data":  "input", // [{children: [...]}, {children: [...]}]
      "field": "children"
    }
  },
  ```
  
[...]

PR #1771, opened 2016-12-11
Use gradient legend for continuous color scale

domoritz:
Part of #1597

### DIFF
--- a/src/compile/legend.ts
+++ b/src/compile/legend.ts
@@ -6,8 +6,8 @@ import {Legend} from '../legend';
 import {title as fieldTitle} from '../fielddef';
 import {AREA, BAR, TICK, TEXT, LINE, POINT, CIRCLE, SQUARE} from '../mark';
 import {hasContinuousDomain} from '../scale';
-import {TEMPORAL} from '../type';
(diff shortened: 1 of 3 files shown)


src/compile/legend.ts
```diff
@@ -106,6 +110,17 @@ export function values(legend: Legend) {
   return vals;
 }
 
+export function type(legend: Legend, fieldDef: FieldDef, channel: Channel) {
+  if (legend.type) {
+    return legend.type;
+  }
+
+  if (channel === COLOR && !fieldDef.bin && !fieldDef.timeUnit && contains([QUANTITATIVE, TEMPORAL], fieldDef.type)) {
```
domoritz: Hmm, you are right. I was thinking that it's similar to binning but I guess we sometimes have quite a large number of bins.

domoritz: I'd say in general that default makes sense. `timeUnit` is similar to binning and so we should treat it as such. Users can override the default. I'm sure we can refine the default once we have vl in voyager. 

screen shot 2016-12-11 at 10 44 14

PR #1774, opened 2016-12-12
Facet clean up and a few more fixes 

kanitw:
Now we can get reasonable facet.

Note that we will to handle the following issues later:
- `paddingOuter` for point doesn't make (may need to just output `padding` if we don't wanna change vega https://github.com/vega/vega/issues/679
- The grid line is too long.
- Need to deal with axis grid layer
- If there is null value, the layout is still incorrect for both facet and non-facet plot. (This should be fixed in Vega https://github.com/vega/vega/issues/678)

vega_editor

cc: @domoritz

### DIFF
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -33,6 +33,11 @@ export interface AxisConfig {
    */
   axisColor?: string;
 
+  /**
+   * Whether to include the axis domain line.
+   */
+  domain?: boolean;
+
   // ---------- Grid ----------
   /**
    * A flag indicate if gridlines should be created in addition to ticks. If `grid` is unspecified, the default value is `true` for ROW and COL. For X and Y, the default value is `true` for quantitative and time fields and `false` otherwise.
@@ -98,6 +103,12 @@ export interface AxisConfig {
    * @minimum 0
    */
   subdivide?: number;
+
+  /**
+   * Whether the axis should include ticks.
+   */
+  tick?: boolean;
+
   /**
    * A desired number of ticks, for axes visualizing quantitative scales. The resulting number may be different so that values are "nice" (multiples of 2, 5, 10) and lie within the underlying scale's range.
    * @minimum 0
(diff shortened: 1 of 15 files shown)


kanitw:
Ah looks like I have to add more test on the plane lol

kanitw:
@arvind  This should be ready for review.  Note that the only failing thing is the diff hit.  (only miss by 2%.) 

I think I added all the tests that are reasonable to test now.  Other stuffs are subject to change, so let's ignore the 2%.  :)

src/compile/facet.ts
```diff
@@ -426,81 +426,65 @@ function parseAxisGroup(model: FacetModel, channel: Channel) {
 }
 
 
-function getXAxesGroup(model: FacetModel): VgMarkGroup {
-  const hasCol = !!model.facet().column;
+export function getSharedAxisGroup(model: FacetModel, channel: 'x' | 'y'): VgMarkGroup {
+  const isX = channel === 'x' ;
+  const facetChannel = isX ? 'column' : 'row';
+  const hasFacet = !!model.facet()[facetChannel];
```
domoritz: `facetChannel in model.facet()`?

src/axis.ts
```diff
@@ -98,6 +103,12 @@ export interface AxisConfig {
    * @minimum 0
    */
   subdivide?: number;
+
+  /**
+   * Whether the axis should include ticks.
```
domoritz: Include => use or show

src/compile/facet.ts
```diff
@@ -21,6 +21,17 @@ import {assembleLayout, parseFacetLayout} from './layout';
 import {Model} from './model';
 import {initScale, parseScaleComponent} from './scale';
 
+/**
+ * Prefix for special data sources for driving column's axis group.
+ */
+
+export const COLUMN_AXES_DATA_PREFIX = 'column-';
```
domoritz: We seem to be a bit inconsistent about whether we use - or _ for scale or data source names. I guess we can use - for data sources and _ for scales given that we use scales in expressions.

kanitw: Good point.

src/compile/facet.ts
kanitw: But sometimes you want the whole object too, I don't want to make the return type inconsistent.

src/axis.ts
kanitw: I think include seems fine here, especially given how many dependent PRs we have.

src/compile/facet.ts
domoritz: I don't get why you would have to change the type.

src/compile/facet.ts
kanitw: This one is a data source, so I guess I don't have to change anything.

src/compile/facet.ts
kanitw: `facet()` return the whole `Facet` object in FacetSpec.  

`facet(channel)` would return `FieldDef`.

PR #1776, opened 2016-12-13
Eliminate Inline Transform

kanitw:
Please review / merge #1774 first. 

You might want look at [branch diff](https://github.com/vega/vega-lite/compare/kw/facet...kw/inline-transform?expand=1) with #1774.


src/compile/mark/mark.ts
```diff
@@ -121,12 +121,11 @@ function parseNonPathMark(model: UnitModel) {
...
 
 /**
- * Returns list of detail fields (for 'color', 'shape', or 'detail' channels)
+ * Returns list of detail (group-by) fields
  * that the model's spec contains.
  */
 function detailFields(model: UnitModel): string[] {
-  // FIXME This is probably not a comprehensive list of detail fields.
-  return [COLOR, DETAIL, OPACITY, SHAPE].reduce(function(details, channel) {
+  return NONSPATIAL_CHANNELS.reduce(function(details, channel) {
```
kanitw: This should become NONSPATIAL_CHANNELS, excluding ORDER

arvind:
As I'm less familiar with the Vega-Lite codebase, a description of proposed changes + rational/justification would really help me review your pull request. For example, it's unclear to me why the `AbstractDataCompiler` interface is needed. The Vega-Lite codebase already has more layers of abstraction than I was expecting, which has presented me a step learning curve. I want to make sure that the complexity of introducing additional layers of abstraction is well reasoned and justified and that we're not overengineering things.

Similarly, based only on the title, I'm not sure how removing the `path` encoding channel is related? This seems like a fairly major proposed change, and there's likely been discussion about it elsewhere but unfortunately I seem to have missed it. With our projects' growing visibility, I think it would be good for us to get in the habit of documenting and discussing these changes in the open.

kanitw:
> As I'm less familiar with the Vega-Lite codebase, a description of proposed changes + rational/justification would really help me review your pull request. 

Sure. I mostly tried to do this in the commit message as (1) they are persisted in the git log and (2) I can write them while I'm writing code even if I don't have internet (e.g., on a flight).  
That said, the coder (e.g., me) often overlooks something obviously to him but not the the reviewer, so I'm glad that you ask for clarification. :) 

> could you explain why the AbstractDataCompiler interface is necessary? I'm a little wary of introducing additional layers of abstraction because the existing ones in the codebase already present me a steeper than expected learning curve. So I want to make sure that the complexity of more abstraction is warranted and cannot be solved by something simpler (e.g., minor code duplication).

Sure.  All of these data compilers (namespaces inside `src/compile/data`) already follow a same pattern from code copying in the past. However, code readers won't know right away without asking or exhaustive code reading.  

Since we use a typed system, I wanted to add this interface to make it obvious that they all follow a same pattern.
I haven't refactored the other data compilers yet, but plan to do it soon.  
As I have to re-implement stack here anyway, I decided to just start with the right code to avoid repeated work.  

> Similarly, based only on the title, I'm not sure how removing the path encoding channel is related? This seems like a fairly major proposed change, and there's likely been discussion about it elsewhere but unfortunately I seem to have missed it. 

It's only related because it changes similar group of code.  That said, thinking more about it I found a corner case where this might not be the best idea yet, so I wanna think about it more. For now, I'm excluding the commit from this PR and force pushed.

kanitw:
We can further discuss the path issue here #1794.

arvind:
Thanks! Your description helped me understand the proposed changes better, and forking `path` into a separate PR has made reviewing this one easier as well :)

PR #1786, opened 2016-12-14
Consolidate rangeMixins to be the only one method that deals with range properties + refactor

kanitw:
(Please merge #1784 and their dependencies first!)

See [branch specific diff](https://github.com/vega/vega-lite/compare/kw/scale-extract...kw/range-mixins) or see each commit for details.


arvind:
I missed this in my code review yesterday, but it looks like some of the test cases are failing.

kanitw:
It should be passing now.  I was busy this morning, so I forgot to comment here that I rebased it to exclude the `path` commit.  That said, after thinking more about it the path commit should be fine, so I have just rebased this PR on top of the path PR (#1794) to avoid changing test cases.  

Thus, please merge https://github.com/vega/vega-lite/pull/1794 before merging this one. :)

PR #1790, opened 2016-12-15
Improve package.json + correct tsconfig, vega schema path

kanitw:
- Add `posttest`
  - include `lint` as a part of `posttest` so we don't forget to run it
  - add `test:examples` and run it as the last step in `posttest` since it can flood the console
    - add `schema` as a part of `npm run test:examples` rather than running up front since there is no need to compile schema if unit tests and lint do not pass. This will make `test` faster to fail!
- Rename `test:only` to `test:noschema`, make it equilvalent to `test` without `schema` and use it with `build` to avoid running `schema` twice.


.travis.yml
```diff
@@ -6,7 +6,7 @@ node_js:
 script:
   - npm run build
   - npm run lint
-  - npm run pretest && npm run test:unit
+  - npm run test:noschema # already run schema in build
```
domoritz: Ran

package.json
```diff
@@ -37,14 +37,15 @@
...
+    "posttest": "npm run lint # && npm run test:examples",
+    "test:examples": "npm run schema && mocha examples",
+    "test:noschema": "npm run pretest && nyc --reporter=html --reporter=text-summary mocha && npm run lint # && mocha examples",
     "codecov": "nyc report --reporter=json && codecov -f coverage/*.json",
     "test:debug": "npm run schema && mocha --debug-brk test examples",
     "watch:build": "watchify src/vl.ts -p tsify -v -d -s vl -o 'exorcist vega-lite.js.map > vega-lite.js'",
-    "watch:test": "nodemon -x 'npm test && npm run lint'",
-    "watch": "nodemon -x 'npm run build && npm run pretest && npm run test:only && npm run lint'",
+    "watch:test": "nodemon -x 'npm test'",
+    "watch": "nodemon -x 'npm run build && npm run test:noschema' # already run schema in build",
```
domoritz: Ran

.travis.yml
kanitw: It's a lazy form of "We've already run", so run should be correct?  

In any case, I'm not gonna fix it here to avoid conflicts.

PR #1794, opened 2016-12-29
Remove `path` channel as we can use `order` channel for sorting line order instead

kanitw:
Note that unlike VL1, now Vega 3 has zindex properties. Thus layer order can be mapped to a scale (haven't implemented yet) while stack order and path order cannot.
Thus, I think it makes sense to consolidate  `order` channel for sorting line order / stack order and have zindex as a separate channel that supports scale. As a result, we can free up `path` for geopath.

(Forked from #1776 -- Please merge #1776 first.)

### DIFF
--- a/examples/specs/scatter_color_zindex.vl.json
+++ b/examples/specs/scatter_color_zindex.vl.json
@@ -5,6 +5,6 @@
     "x": {"field": "Horsepower", "type": "quantitative"},
     "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
     "color": {"field": "Origin", "type": "nominal"},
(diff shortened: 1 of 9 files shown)


kanitw:
I just realized one problem.  

If `order` determines both stacked order and line order, this can be problematic for stacked lines.   
(When we overlay lines on top of stacked areas, the lines have to be stacked too.)  

Basically, we need to distinguish whether users want to order the stacking order or  order the line path (as in connected scatterplot).

kanitw:
Hmm,  actually,  there is not really a problem.    Users won't sort line path for a stacked line.  So this should be okay.  
That said, it's a good idea to separate this from #1776 anyway ....

kanitw:
Rebased to eliminate conflicts

PR #1795, opened 2016-12-29
Unit Selections

arvind:
This pull request adds selection functionality to unit specifications. The syntax for selections has changed subtly from our InfoVis 2016 paper:

* The top-level block is now named `selection` instead of `select` for consistency with `encoding`. 
* `point` and `list` selections have been renamed back to `single` and `multi`.
* The `project` transform is defined over `fields` and `encodings` rather than `fields` and `channels` and is now specified at the top-level of a selection definition, rather than nested under the `project` keyword (i.e., `"type": "single", "fields": ["Horsepower", "Origin"]` or `"type": "interval", "encodings": ["x", "y"]`).
* `"init": {"scales": true}` is now simply `"bind": "scales"`. [See below](https://github.com/vega/vega-lite/pull/1795#issuecomment-269865808) for HTML input binding.

**Task List**

* Base selection types
  * [x] Single
  * [x] Multi
  * [x] Interval
* Selection transforms:
  * [x] project
  * [x] toggle
  * [x] translate
  * [x] zoom
  * [x] nearest
* Bindings
  * [x] Scales (modulo Vega support for pushing signal updates to ancestor scopes).
  * [x] [HTML Input Elements](https://github.com/vega/vega-lite/pull/1795#issuecomment-269865808)
* Spec parameterization:
  * [ ] Production Rules
  * [ ] Scale domain/range
  * [ ] Filter transform
  * [x] Interval selection mark (brush)
  * [ ] Direct reference single selections.
  * [ ] Make legends interactive where appropriate. 
* Miscellaneous
  * [x] Extract defaults to config.
  * [ ] Support `visual` domain (might punt on this for v2.0). 

* Examples
  * [ ] Add all examples to `examples/specs` so that both input specs and output specs will be validated.   Feel free to add different parameterizations of single/multi/interval specs (they will be useful for reference from the docs anyway)


vega-lite-schema.json
```diff
@@ -1025,6 +1025,12 @@
             },
             "type": "object"
         },
+        "Dict<SelectionSpec>": {
```
kanitw: Looks we need other ways to declare this or need @domoritz to modify this?

kanitw: Looks like we need other ways to declare this or need @domoritz to modify the schema generator?

src/selection.ts
```diff
@@ -0,0 +1,62 @@
+export enum SelectionTypes {
```
kanitw: FYI, we are now using string literals for this kind of thing instead.  

You can just do:

a) Simple type declaration 
```
export type SelectionType = 'single' | 'multi' | 'interval';
```

Or b) declare a namespace + type like @domoritz  did (e.g., for [ScaleType](https://github.com/vega/vega-lite/blob/master/src/scale.ts)).  


IMHO, I prefer a) because you can avoid importing types and constants when you use a function. 
(You still need to import the type for function declaration though.)

arvind: Good to know, thanks! I've been mainly using these enums as a way to do both the type checking but also to avoid hardcoded constants. I think I'd need to emulate `ScaleType` in that regard, then. For my own edification, could you explain why we've moved away from the enums? Thanks!

vega-lite-schema.json
arvind: I defer to whatever you two think is the right thing to do! This is what we'd used in the InfoVis prototype, and I just copied it over for expediency's sake since I'm still pretty unfamiliar with TypeScript.

src/selection.ts
kanitw: > avoid hardcoded constants

If you use string literal to declare types, hardcoding constant no longer cause misspelling problem because if you misspell the constants, the build won't pass.   Plus, with string literals, you already got nice autocompletion of possible values with a proper typescript editor (e.g., VSCode).  

Thus, I think that sticking with a) format above makes thing simpler since it reduces number of imports we have to do for each file without causing typo bugs. 

> For my own edification, could you explain why we've moved away from the enums? 

Enum suppose to be numbers, not string in Typescript.  Using as `any` was a hack that works okay in the past.  However, it prevents us from turning on `noImplictyAny` flag as request by external users of our project in #1649.   There might be also other reasons, but I can't remember one now.  Maybe @domoritz can add to this.

arvind: > hardcoding constant no longer cause misspelling problem because if you misspell the constant

Good to know! I guess I was less worried about misspellings and more concerned about how easily constants can be refactored. Perhaps my current usage of enums is incorrect, but I can't seem to get the following code to build and the "renaming symbol" option in VSCode doesn't seem to propagate changes (e.g., `single` -> `point` or `multi` -> `list`). Perhaps I'm doing something wrong?

```
type SelectionType = 'single' | 'multi' | 'interval';
let a:SelectionType = 'single';
if (a === 'multi') {
  console.log(a);
}
```

arvind: I'm also not sure I understand the issue with reducing the "number of imports." In my above snippet, would I not need to still import `SelectionType` in `b.ts` to be able to designate it as the type of `a`? Thanks for your patience as I learn TypeScript, I appreciate it!

kanitw: > "renaming symbol" option in VSCode doesn't seem to propagate changes

AFAIK, I think you still need to rely on find and replace for renaming.  
That said, the typing will ensure that you don't forget to rename them if you rename the value in the type first.  

By this I mean, if you change
```
export type SelectionType = 'single' | 'multi' | 'interval';
```
to
```
export type SelectionType = 'point' | 'list' | 'interval';
```
first.  The build will keep failing until you have replaced all single/multi SelectionTypes.

> would I not need to still import SelectionType in b.ts to be able to designate it as the type of a? 

Yes.  So you still need to import to designate type, but no longer need to import to designate value.  

For example, for declaring variable and function types you still need to import:

```
import {ScaleType} from './scale';
const scaleType: ScaleType = 'linear';  
export function a(scaleType: ScaleType);
```

If you import `a()` to another file, you can just call

```
a('linear');
```

as opposed to 

```
import {ScaleType} from './scale';
a(ScaleType.LINEAR); 
```

That said, if both you and @domoritz prefer the latter (declare a namespace + type like the current ScaleType code) rather than just the type like I mention above, I'm okay with that too.

arvind: Thanks @kanitw, this is super helpful!!

arvind: Made the change as of ff0ae01. You were right on all counts, @kanitw! Thanks!

arvind:
As of 0f55e81, single selections can now be bound to HTML input widgets ala top-level Vega signals. The `bind` property follows that of Vega, with an input widget being generated for every projected field/encoding. For more fine-grained control, bindings can also be nested using keys corresponding to the projections like so:

pasted image at 2016_12_31 06_39 pm

For interval selections (and eventually multi selections as well), `bind` can currently only be set to `scales`. We may want to consider supporting HTML input binding for these selection types as well, but there's enough complexity here that I think it's worth punting for a future release.

arvind:
@kanitw, @domoritz This PR is ready for initial review. All the infrastructure for generating the necessary signals, datasets, and marks for selections + transformations should be in place. My plan is to cut a separate PR for parameterizations to make reviewing easier, because I imagine it will churn the rest of the codebase significantly. 

Since you will not be able to visually determine that a selection is populating correctly, you'll instead need to check what is being stored within a selection's dataset (`name_store`). Interval selections will store tuples in the form `{unit: number, intervals: array<{field:string, extent:array<number>}>}`, multi selections will be `{unit: number, fields: array<string>, values: array<*>}` and single selections the same except values will also be directly accessible via their fields (e.g., `{unit: string, fields: ["Origin"], values: ["Japan"], Origin: "Japan" }`). 

In subsequent comments, I will post some example specs that you can test.

arvind:
A simple single selection:

```json
{
  "mark": "circle",
  "selection": {
    "grid": {
      "type": "single"
    }
  },
  "encoding": {
    "y": {"field": "Miles_per_Gallon","type": "quantitative"},
    "color": {"field": "Origin", "type": "N"},
    "size": {"value": 100}
  }
}
```

Things to try:
* Different values for `on` (e.g., `"on": "click[event.shiftKey]").
* Different projection values (e.g., `"fields": ["Origin", "Cylinders"]`)
* Different [bindings](https://github.com/vega/vega-lite/pull/1795#issuecomment-269865808)
* `"nearest": true`

arvind:
A multi selection:

```json
{
  "mark": "circle",
  "selection": {
    "grid": {
      "type": "multi"
    }
  },
  "encoding": {
    "y": {"field": "Miles_per_Gallon","type": "quantitative"},
    "color": {"field": "Origin", "type": "N"},
    "size": {"value": 100}
  }
}
```

Things to try:
* Different values for `on` (e.g., `"on": "click[event.shiftKey]").
* Different projection values (e.g., `"fields": ["Origin", "Cylinders"]`).
* Different toggle filters (e.g., `"toggle": event.altKey` or even `"toggle": false` in which case points should continuously only be _added_ to the selection).
* `"nearest": true`

arvind:
An interval selection:

```json
{
  "mark": "circle",
  "selection": {
    "grid": {
      "type": "interval"
    }
  },
  "encoding": {
    "y": {"field": "Miles_per_Gallon","type": "quantitative"},
    "color": {"field": "Origin", "type": "N"},
    "size": {"value": 100}
  }
}
```

Things to try:
* Different values for `on` (e.g., `"on": "[mousedown[event.shiftKey], mouseup] > mousemove").
* Different projection values (e.g., `"encodings": ["x"]` or `"encodings": ["y"]`).
* Different translate values (e.g., `"translate": false` or `"translate": "[mousedown[event.shiftKey], mouseup] > mousemove")
* Different zoom values (e.g., `"zoom": false` or `"zoom": "wheel[event.shiftKey]").
* `"bind": "scales"` with the above variations.

arvind:
And finally, you can verify that default selection definitions are being set correctly using the `config.selection.<selection-type>` key. For example,

```json
"config": {
    "selection": {
      "interval": {
        "on": "[mousedown[event.shiftKey], mouseup] > mousemove, [mousedown[event.altKey], mouseup] > mousemove"
      } 
    }
  }
```

src/compile/selection/transforms/translate.ts
```diff
@@ -0,0 +1,103 @@
...
+
+          (y !== null ? 'extent_y: ' + (scales ? domain(model, Y) :
+              'slice(' + name + '_y)') + ', ' : '') + '}'
+      }]
+    }, {
+      name: name + NS.DELTA,
+      value: {},
+      on: [{
+        events: events,
+        update: '{x: x(unit) - ' + anchor + '.x, y: y(unit) - ' + anchor + '.y}'
```
domoritz: This and many other string operations in this PR will be easier to read with [template strings](https://basarat.gitbooks.io/typescript/content/docs/template-strings.html)

vega-lite-schema.json
domoritz: I think if we give this type and alias, we should be fine. The generated schema is not incorrect so I don't think there is anything to fix in the schema generator. I can take care of this.

vega-lite-schema.json
```diff
@@ -2663,6 +2673,11 @@
             ],
             "type": "string"
         },
+        "T": {
```
domoritz: This should definitely not be here.

src/compile/selection/transforms/inputs.ts
```diff
@@ -0,0 +1,57 @@
+import {TransformCompiler} from './';
+import {stringValue} from '../../../util';
+import {warn} from '../../../log';
+
+const inputBindings:TransformCompiler = {
```
domoritz: @kanitw Do we want classes instead of const objects? I think it should be possible to still put the compilers in the `compilers` dict.

src/compile/selection/transforms/zoom.ts
```diff
@@ -0,0 +1,95 @@
...
+import {UnitModel} from './../../unit';
+import {SelectionComponent} from '../../../selection';
+import {X, Y, Channel} from '../../../channel';
+import {stringValue} from '../../../util';
+import {warn} from '../../../log';
+import {TransformCompiler} from './';
+import {default as scalesCompiler, domain} from './scales';
+import {projections as intervalProjections, NS as INTERVAL} from '../types/interval';
+
+const NS = {
```
domoritz: What does NS stand for?

vega-lite-schema.json
```diff
@@ -1025,6 +1025,12 @@
             },
             "type": "object"
         },
+        "Dict<SelectionSpec>": {
+            "additionalProperties": {
+                "$ref": "#/definitions/T"
```
domoritz: This T thing is odd. Hopefully goes away with the right type/ alias.

src/compile/selection/transforms/translate.ts
arvind: Good call! I'd totally forgotten about template strings. Will switch it over.

src/compile/selection/transforms/zoom.ts
arvind: Namespace.

src/compile/selection/transforms/translate.ts
arvind: Done! Made so many of those expressions far more readable. Thanks!

src/compile/selection/transforms/index.ts
```diff
@@ -0,0 +1,33 @@
...
+import toggle from './toggle';
+import translate from './translate';
+import zoom from './zoom';
+import scales from './scales';
+import inputs from './inputs';
+import nearest from './nearest';
+const compilers: Dict<TransformCompiler> = {project, toggle, scales,
+  translate, zoom, inputs, nearest};
+
+export function transforms(sel: SelectionComponent, cb: (tx: TransformCompiler) => void) {
```
kanitw: I wonder if we could rename it to `compileSelectionTransforms` or something

arvind: Could you expand on why you consider this method convoluted? It's a fairly straightforward implementation of a [visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern). I could rename it to `transformsVisitor` but I erred on the side of brevity.

src/compile/compile.ts
```diff
@@ -67,6 +67,7 @@ function assemble(model: Model) {
 }
 
 export function assembleRootGroup(model: Model) {
+  const layout = stringValue(model.name(LAYOUT + ''));
```
arvind: [`stringValue`](https://github.com/vega/vega-util#stringValue) is helpful when including a string value within expression strings. `model.name(LAYOUT) -> 'layout'` whereas `stringValue(model.name(LAYOUT)) -> '"layout"'`. Thus, in the signal expression strings below, I can simply concatenate the `layout` variable without worrying about escaping it, etc.

kanitw: This method is pretty awesome.  @domoritz  We should use it more. :)

src/compile/selection/transforms/inputs.ts
```diff
@@ -0,0 +1,55 @@
+import {TransformCompiler} from './';
+import {stringValue} from '../../../util';
+import {warn} from '../../../log';
```
domoritz: @arvind prefers it as well. It allows you to namespace functions and enables autocomplete.

src/compile/selection/transforms/inputs.ts
domoritz: I don;t have any reasons for one or the other. Consistency is king, though.

kanitw:
FYI, @arvind  I'm appending a TODO for adding example to the description on the top.

arvind:
Yeah, that's how I tested things for now. Another option is to manually add production rules to the output Vega and test things that way. E.g., 

```json
"fill": [
  {"test": "vlPoint('grid_store', parent._id, datum, 'single')", "scale": "color","field": "Origin"},
  {"value": "grey"}
]
```

for `single` and `multi` selections, and switch the expression to use `vlInterval` for `interval` selections.

src/compile/unit.ts
```diff
@@ -25,6 +25,8 @@ import {parseMark} from './mark/mark';
 import initScale from './scale/init';
 import parseScaleComponent from './scale/parse';
 import {stack, StackProperties} from '../stack';
+import {SelectionDef} from '../selection';
+import {parseUnitSelection, assembleUnitSignals, assembleUnitData as assembleSelectionData, assembleUnitMarks as assembleSelectionMarks} from './selection/selection';
```
kanitw: I wonder if we could just rename `assembleUnitMarks` => `assembleUnitSelectionMarks`?  
Basically, aliasing adds another level of indirection, thus our codebase currently avoids it by just using slightly longer, but more descriptive names.

src/compile/selection/interval.ts
```diff
@@ -0,0 +1,164 @@
...
+        events: evt,
+        update: `[${name}[0], ` + invert(`clamp(${coord}, 0, ${size})`) + ']'
+      });
+
+      return on;
+    })
+  };
+}
+
+function events(selCmpt: SelectionComponent, cb: Function) {
```
kanitw: `forEachEvent` for consistency?

vega-lite-schema.json
```diff
@@ -1032,6 +1200,18 @@
             },
             "type": "object"
         },
+        "Dict<SelectionDef>": {
```
kanitw: Actually, let me correct this first!

PR #1812, opened 2017-01-11
Fix incorrect layout for point scales

### LINKED ISSUE #1807: Incorrect Positioning for Vertical Point scales
vega_editor

kanitw:
Fix #1807

The output look good now!

vega_editor

vega_editor

### DIFF
--- a/test/compile/layout.test.ts
+++ b/test/compile/layout.test.ts
@@ -44,7 +44,7 @@ describe('compile/layout', () => {
       });
 
       const sizeExpr = unitSizeExpr(model, X);
-      assert.equal(sizeExpr, 'max(datum["distinct_a"] + 2*0.5, 0) * 21');
+      assert.equal(sizeExpr, 'max(datum["distinct_a"] - 1 + 2*0.5, 0) * 21');
     });
 
     it('should return correct formula for ordinal-band scale with custom padding', () => {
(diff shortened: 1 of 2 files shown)


test/compile/layout.test.ts
```diff
@@ -44,7 +44,7 @@ describe('compile/layout', () => {
       });
 
       const sizeExpr = unitSizeExpr(model, X);
-      assert.equal(sizeExpr, 'max(datum["distinct_a"] + 2*0.5, 0) * 21');
+      assert.equal(sizeExpr, 'max(datum["distinct_a"] - 1 + 2*0.5, 0) * 21');
```
domoritz: -1+1 = 0

kanitw: Sure, but 0.5 can vary based on paddingOuter. So there isn't a better.

kanitw: Note that I intentionally chose 0.5 as the default because - 1 + 2*0.5 = 0

domoritz: I guess I would like to see a comment that explains that 0.5 is the passingOuter.

kanitw: I'm NOT gonna add comments to unit tests unless it's necessary.  
This is a fairly small method, you can easily read the code which is already well commented. 

Plus, I'm not changing the 0.5 part in this PR.

PR #1814, opened 2017-01-11
Typescript 2.1 compatible

kanitw:
Replacing #1767.

### DIFF
--- a/src/compile/unit.ts
+++ b/src/compile/unit.ts
@@ -7,7 +7,7 @@ import {defaultConfig, Config, CellConfig} from '../config';
 import {SOURCE, SUMMARY} from '../data';
 import {Encoding} from '../encoding';
 import * as vlEncoding from '../encoding'; // TODO: remove
-import {FieldDef, FieldRefOption, field} from '../fielddef';
+import {ChannelDefWithScale, ChannelDefWithLegend, FieldDef, FieldRefOption, field} from '../fielddef';
 import {Legend} from '../legend';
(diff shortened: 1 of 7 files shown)


src/compile/unit.ts
```diff
@@ -166,7 +166,8 @@ export class UnitModel extends Model {
           (channel === X && vlEncoding.has(encoding, X2)) ||
           (channel === Y && vlEncoding.has(encoding, Y2))
         ) {
-        const scale = _scale[channel] = initScale(channel, encoding[channel], config, mark,
+        const scale = _scale[channel] = initScale(
+          channel, encoding[channel] as ChannelDefWithScale, config, mark,
```
domoritz: That's a bad hack! Can we do this without the `as`?

src/compile/unit.ts
```diff
@@ -241,7 +242,8 @@ export class UnitModel extends Model {
   private _initLegend(encoding: Encoding, config: Config): Dict<Legend> {
     return NONSPATIAL_SCALE_CHANNELS.reduce(function(_legend, channel) {
       if (vlEncoding.has(encoding, channel)) {
-        const legendSpec = encoding[channel].legend;
+        const channelDef = encoding[channel] as ChannelDefWithLegend;
```
domoritz: Same here.

src/compile/unit.ts
domoritz: Since this is the trade-off we have to make when we don't enumerate the channels, I'll accept this.

PR #1818, opened 2017-01-12
[Refactor] Declare channels just like how @domoritz wants and add uni…

kanitw:
…t tests to prevent bugs for later additions of channels.


domoritz:
Sweet.

Issue #1278, opened 2016-04-04
Add Other Encoding Channels as ValueDef only channel


kanitw:
Many mark properties are currently customizable only via mark config. 
However, they should become new encoding channels. 
For now, let's make them constant values only.  

- [ ] text properties: align, angle, baseline, radius, theta, font, fontSize, fontStyle, fontWeight, ...
- [ ] `interpolate`, `tension`
- [ ] `orient`

- Fills and Strokes (Related, but let do it separately) #1821

## New Properties in Vega 3

- [ ] `cornerRadius`
- [ ] Check Vega 3's schema (See Mark's encodeEntry) to check if there are other properties not listed here 
- [ ] ...

## STEP
- [ ] Make `ValueDef` interface that only contains `value` and make `FieldDef` implements `ValueDef`
- For each one
  - [ ] Add channel to channel.ts
  - [ ] Add its ~~FieldDef~~ ValueDef to encoding.ts. 
  - [ ] ~~For ones that support `field`, add rules for scale type, domain (?), and range~~ 
  - [ ] Add reference in `compile/marks/*.ts`
     - [ ] Instead of calling `applyMarkConfig`, create a similar method in `valueref.ts` to output the value of these properties from `encoding`. 
     - [ ] Remove `applyMarkConfig` and make those references (e.g.,) correctly read from `encoding` instead. 

- [ ] Write Test
- [ ] Manually test specifying each of the property

domoritz:
Opacity?

kanitw:
Any other important ones?

domoritz:
Line width?

kanitw:
> Line width?

`strokeWidth` is an important one.  We couldn't map field to it yet though.   (Vega's line doesn't support variable width yet.)

kanitw:
- [ ] all other properties that are passed through `applyMarkConfig` in `common.ts` – refactor `applyMarkConfig` to become `applyMarkProperty` to read from channel then config.

kanitw:
- [ ] filled

kanitw:
#1655 is one good example why we should support `fill` / `stroke` channel for fine-grain control.

nielsmde:
I was missing an encoding channel for the line-style in vega-lite and came across this issue.
I would look into this and try to implement some of these new channels, if no one else is already on it.

[...]

PR #1825, opened 2017-01-15
Add any `VgEncodeEntry` type and refer to it

kanitw:
Also, use `e` as variable name instead of `p` (from the old name "properties")

(Prepare for @yuhanlu to add VgEncodeEntry as a part of #1278)


kanitw:
@YuhanLu Don't forget to remove branch when you merge :)  (I just did it for you)

PR #1832, opened 2017-01-16
Split Data Interface (Data = UrlData | InlineData)

### LINKED ISSUE #1381: Modify interface for Data
```
export interface Data {
  formatType?: DataFormat;
  url?: string;
  /**
   * Pass array of objects instead of a url to a file.
   */
  values?: any[];
}
```

Practically, we should split this into `type Data = UrlData | InlineData` where `UrlData` contains `url` and `formatType` while `InlineData` contains `values`.

kanitw:
Fix #1381

### DIFF
--- a/src/data.ts
+++ b/src/data.ts
@@ -48,7 +48,9 @@ export namespace DataFormatType {
 
 export type DataFormatType = typeof DataFormatType.JSON | typeof DataFormatType.CSV | typeof DataFormatType.TSV | typeof DataFormatType.TOPOJSON;
 
-export interface Data {
+export type Data = UrlData | InlineData;
+
+export interface UrlData {
   /**
    * An object that specifies the format for the data file or values.
    */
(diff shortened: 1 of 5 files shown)


src/data.ts
```diff
@@ -58,11 +60,22 @@ export interface Data {
...
+
+export interface InlineData {
   /**
    * Pass array of objects instead of a url to a file.
    */
-  values?: any[];
+  values: any[];
+}
+
+export function isUrlData(data: Data): data is UrlData {
```
kanitw: @NeelMohapatra  If you haven't, you might wanna learn about what is a type guard in typescript.  https://www.typescriptlang.org/docs/handbook/advanced-types.html 

For example, try pulling this branch and change this to `boolean` and see what errors you've got. :)

domoritz:
Needs rebasing.

kanitw:
Normally it's branch owner's responsibility (me).  But I'll let @NeelMohapatra do it this time. 
(I wonder if @NeelMohapatra know how to rebase.  If not, this will be a good chance for you to practice.)  :)

domoritz:
@NeelMohapatra Instead of merging vg3 into this branch, can you rebase it?

* To do so, reset the commits you made.
* Then rebase this branch onto vg3
* When you are certain that everything is correct (and you have to because otherwise you may destroy some work we have done), force push only this branch back to github.

kanitw:
FYI, when your branch is out of date .. we prefer rebasing rather than merging unless your branch is very big (e.g., > 10 *atomic* commits) :)

NeelMohapatra:
@kanitw Yeah, I'm on it. Trying to revert to an old commit. :P I clearly made a mistake

kanitw:
@NeelMohapatra  you can use interactive rebase (`git rebase -i`) and drop your "merge commit". :)

kanitw:
and you can use `git push -f` for force pushing (Never do this on master though!)

Issue #1723, opened 2016-11-25
Utilizing JSON schema to populate docs


kanitw:
Currently, we have both docs in the JS interface (which produces "description" in the JSON schema) and our markdown files. 

The benefit of markdown files is that we can add live examples in the docs nicely.  

The problem with this approach is that the (TS) JSDocs isn't always synced with the markdown file. 

We might want to consider other approach for 2.0. 

For example, I wanna look closer at 
https://blog.cloudflare.com/cloudflares-json-powered-documentation-generator/
(maybe it can provide a good solution for us)

kanitw:
Actually, if there is a way for us to generate "properties tables" in our docs from the JSON schema that we can include / import to our docs easily.  (Maybe using template tricks similar to Altair could work.)

cc: @willium

kanitw:
http://typedoc.org/ is another option.

domoritz:
* [ ] Consider using @default properties and populating the docs from that.

kanitw:
For static default values, I think `@default` is a great candidate. 
However, we also need to decide what to do for default rules (when something is conditional).

kanitw:
@YuhanLu you say #2085 fixes part of this -- I guess the only part left is to use this engine with all other pages?  If so, let's have a TODO list

- [ ] aggregate.md
- [x] axis.md
- [ ] bin.md
- [ ] composition.md
- [ ] config.md
- [ ] data.md
- [ ] encoding.md
- [ ] facet.md
- [ ] label.md
- [x] legend.md
- [ ] mark.md
- [ ] overview.md
- [ ] scale.md
- [ ] size.md
- [ ] sort.md
- [ ] spec.md
- [ ] timeunit.md
- [ ] transform.md

YuhanLu:
Those are the pages that we might need to manually update the table:
- [ ] aggregrate

PR #1834, opened 2017-01-16
Add top-level `padding` and make it = 5 by default

### LINKED ISSUE #1822: Try adding explicit padding to the output specs
Based on Jeff's suggestion in https://github.com/vega/vega/issues/713#issuecomment-272627467, we should consider adding explicit padding (e.g., "padding": 5) to output specs to improve the presentation a bit.

kanitw:
(as suggested by Jeff)

Fix #1822

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -3,7 +3,7 @@ import {LegendConfig, defaultLegendConfig} from './legend';
 import {MarkConfig, AreaConfig, BarConfig, LineConfig, PointConfig, TextConfig, TickConfig, RectConfig, RuleConfig} from './mark';
 import * as mark from './mark';
 import {ScaleConfig, defaultScaleConfig} from './scale';
-
+import {Padding} from './spec';
 
 export interface CellConfig {
   width?: number;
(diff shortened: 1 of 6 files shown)


src/config.ts
```diff
@@ -126,6 +126,15 @@ export interface Config {
   background?: string;
 
   /**
+   * The default visualization padding, in pixels, from the edge of the visualization canvas to the data rectangle. This can be a single number or an object with `"top"`, `"left"`, `"right"`, `"bottom"` properties.
+   *
+   * __Default value__: `5`
```
domoritz: Why not use `@default` annotation?

kanitw: I'm undecided about this too. 

For against `@default`:

1. For JSDoc this has different meaning.  

```
/** 
 *@default 
 */
const a = 5;   // This property will automatically 
```

That said, this is TSDoc. 

2. As we discussed that we should populate the markdown docs with text from TSDocs.  This will probably generate better output than `@default: 5`. 

But `@default` is good for JSON Schema.

kanitw: Should we just file an issue and deal with this when we make the docs work?
(It's hard to decide this now.. and we should get the alpha.0 out first!)

domoritz: Sure!

domoritz: https://github.com/vega/vega-lite/issues/1723#issuecomment-272792870

PR #1837, opened 2017-01-16
Make the text table heatmap example work

kanitw:


kanitw:
It renders now!

vega_editor

PR #1839, opened 2017-01-16
Use [] instead of Array

domoritz:
Doesn't matter right now but I'm making some changes and it will matter. Right now we just get consistency.


kanitw:
LGTM but can you explain why it will matter before we merge.   (Feel free to reply & merge.)

domoritz:
I want to enable aliases so we fix some issues that we've had in the past where some properties didn't have comments.

PR #1849, opened 2017-01-18
Distinguish between PointConfig and SymbolConfig.

kanitw:
(Config for `circle` and `square` should not include `shape` and `shapes` properties)

### DIFF
--- a/vega-lite-schema.json
+++ b/vega-lite-schema.json
@@ -835,7 +835,7 @@
                     "description": "Cell Config"
                 },
                 "circle": {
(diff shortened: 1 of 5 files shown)


vega-lite-schema.json
```diff
@@ -908,7 +908,7 @@
                     "description": "Scale Config"
                 },
                 "square": {
-                    "$ref": "#/definitions/PointConfig",
+                    "$ref": "#/definitions/SymbolConfig",
```
domoritz: We can enable alias refs so that these are different.

PR #1851, opened 2017-01-18
[Obsolete] Remove the unnecessary bin_legend scale

kanitw:
This was a hack to make legend work, but it seems we no longer need it.

```
{
  "mark": "point",
  "encoding": {
    "x": {"field": "Acceleration", "type": "quantitative"},
    "color": {"bin": true, "field": "Acceleration", "type": "quantitative"}
  },
  "config": {"numberFormat": "d"}
}
```

pasted_image_1_17_17__9_30_pm

changing to opacity also works 

pasted_image_1_17_17__9_31_pm

### DIFF
--- a/examples/specs/point_1d_bin_color_legend.vl.json
+++ b/examples/specs/point_1d_bin_color_legend.vl.json
@@ -1,0 +1,9 @@
+{
+  "description": "Demonstrate color legend with binned data.",
+  "data": {"url": "data/cars.json"},
+  "mark": "point",
+  "encoding": {
+    "x": {"field": "Acceleration", "type": "quantitative"},
+    "color": {"bin": true, "field": "Acceleration", "type": "quantitative"}
+  }
+}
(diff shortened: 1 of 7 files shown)


domoritz:
Try `scatter_binned_size.vl.json`

screen shot 2017-01-17 at 22 08 07

vs

screen shot 2017-01-17 at 22 08 02

kanitw:
Nice catch, sounds like we should file a bug in Vega!

domoritz:
@kanitw Are you filing an issue for this?

domoritz:
Unfortunately, we cannot merge this because the entries in the size scale overlap. See https://github.com/vega/vega/issues/721

kanitw:
I'm gonna re-open this and keep it as a reminder for me to re-visit why this breaks later.  
(It's not a pressing issue so I'd skip it for now.)

kanitw:
I guess we can close this now given we have #1892

PR #1858, opened 2017-01-19
index scale has discrete domain.

domoritz:
I'm not even sure whether these type annotations are still necessary.


kanitw:
Nice catch.  They are indeed necessary since these are new concepts in Vega 3 and are recently introduced for the migration. 

I'll take a look and see if this break anything.  If not, I'll merge.

domoritz:
Are the type annotations necessary or will TS infer them automatically?

kanitw:
Oh, that part I don't know.  You might wanna test that :)

kanitw:
Just tested.  We still need them!

kanitw:
I also found a bug in `toSet`'s interface.

domoritz:
Merge?

PR #1863, opened 2017-01-21
Make interfaces and schema for `FieldDef`, `ValueDef` more precise

### LINKED ISSUE #1861: Incorrect schema for Detail channel
Detail channel is for additional group-by/measure `field`s and thus does not support `value`.

kanitw:
Fix #1861 

`ChannelDef = FieldDef | ValueDef`

Rename `*ChannelDef` / `ChannelDef*` => `*FieldDef`
- `ChannelDefWithScale` => `ScaleFieldDef`
- `PositionChannelDef` => `PositionFieldDef`
- `ChannelDefWithLegend` => `LegendFieldDef`
- `OrderChannelDef` => `OrderFieldDef`

### DIFF
--- a/src/compile/config.ts
+++ b/src/compile/config.ts
@@ -3,7 +3,7 @@ import * as log from '../log';
 import {X, COLOR, SIZE, DETAIL} from '../channel';
 import {Config} from '../config';
 import {Encoding, isAggregate, channelHasField} from '../encoding';
-import {isMeasure, FieldDef} from '../fielddef';
+import {isMeasure, isFieldDef, FieldDef} from '../fielddef';
 import {MarkConfig, TextConfig, Orient} from '../mark';
 import {BAR, AREA, POINT, LINE, TICK, CIRCLE, SQUARE, RECT, RULE, TEXT, Mark} from '../mark';
 import {Scale, hasDiscreteDomain} from '../scale';
@@ -87,7 +87,7 @@ export function orient(mark: Mark, encoding: Encoding, scale: Dict<Scale>, markC
       if (!hasDiscreteDomain(xScaleType) && (
             !encoding.y ||
             hasDiscreteDomain(yScaleType) ||
-            (encoding.y as FieldDef).bin
+            (isFieldDef(encoding.y) && encoding.y.bin)
         )) {
         return 'vertical';
       }
--- a/src/compile/config.ts
+++ b/src/compile/config.ts
@@ -3,7 +3,7 @@ import * as log from '../log';
 import {X, COLOR, SIZE, DETAIL} from '../channel';
 import {Config} from '../config';
 import {Encoding, isAggregate, channelHasField} from '../encoding';
-import {isMeasure} from '../fielddef';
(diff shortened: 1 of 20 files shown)


src/compile/config.ts
```diff
@@ -87,7 +87,7 @@ export function orient(mark: Mark, encoding: Encoding, scale: Dict<Scale>, markC
       if (!hasDiscreteDomain(xScaleType) && (
             !encoding.y ||
             hasDiscreteDomain(yScaleType) ||
-            encoding.y.bin
+            (encoding.y as FieldDef).bin
```
domoritz: instead you could do `isFieldDef(encoding.y) && encoding.y.bin`

src/encoding.ts
```diff
@@ -48,14 +49,14 @@ export interface UnitEncoding {
...
-  size?: ChannelDefWithLegend;
+  size?: LegendFieldDef | ValueDef<number>;
 
   /**
    * The symbol's shape (only for `point` marks). The supported values are
    * `"circle"` (default), `"square"`, `"cross"`, `"diamond"`, `"triangle-up"`,
    * or `"triangle-down"`, or else a custom SVG path string.
    */
-  shape?: ChannelDefWithLegend; // TODO: maybe distinguish ordinal-only
+  shape?: LegendFieldDef | ValueDef<string>; // TODO: maybe distinguish ordinal-only
```
domoritz: `ValueDef<Shape|string>`? Maybe even better, change shape to be `'circle' | 'square' | 'cross' | 'diamond' | 'triangle-up' | 'triangle-down' | string`.

src/fielddef.ts
```diff
@@ -84,12 +88,22 @@ export interface ChannelDefWithLegend extends ChannelDefWithScale {
...
 }
 
+export type ChannelDef = FieldDef | ValueDef<any>;
+
+export function isFieldDef(channelDef: ChannelDef): channelDef is FieldDef | PositionFieldDef | LegendFieldDef | OrderFieldDef  {
+  return channelDef && !!channelDef['field'];
+}
+
+export function isValueDef(channelDef: ChannelDef): channelDef is ValueDef<any> {
+  return channelDef && !!channelDef['value'];
```
domoritz: what is value is false? Wouldn't `'value' in channelDef` be better?

src/encoding.ts
kanitw: I disagree with this one.  There is no point doing `<string literals> | string`.

src/fielddef.ts
kanitw: Good catch.  I change it to:
```
channelDef && 'value' in channelDef && channelDef['value'] !== undefined;
```

kanitw:
@domoritz  Fixed two comments that I agreed and force pushed.

src/encoding.ts
domoritz: Well, we actually use `Shape | string` in https://github.com/vega/vega-lite/blob/f39a127d70248253c7ec4af93a1290ce9400cfd6/src/mark.ts#L203

PR #1865, opened 2017-01-22
Change `scheme: ...` to be `range: {scheme: ...}`

kanitw:


kanitw:
Actually I start wondering if we should only adjust the output to match Vega while keeping the input flat.

Issue #1828, opened 2017-01-15
Layer Histogram doesn't render


kanitw:
```
{
  "data": {"url": "data/flights-2k.json"},
  "layers": [
    {
      "mark": "bar",
      "encoding": {
        "x": {
          "field": "distance",
          "type": "quantitative",
          "bin": true
        },
        "y": {
          "aggregate": "count",
          "field": "*",
          "type": "quantitative"
        }
      }
    },
    {
      "transform": {"filter": "datum.delay < 5"},
      "mark": "bar",
      "encoding": {
        "x": {
          "field": "distance",
          "type": "quantitative",
          "bin": true
        },
        "y": {
          "aggregate": "count",
          "field": "*",
          "type": "quantitative"
        },
        "color": {"value": "goldenrod"}
      }
    }
  ]
}
```

We got the following error: 

```
Uncaught Error: Undefined data set name: undefined
    at error (http://localhost:1233/vendor/vega.js:806:9)
    at Subscope.prototype$71.getData (http://localhost:1233/vendor/vega.js:29076:5)
    at http://localhost:1233/vendor/vega.js:28209:22
    at Array.map (native)
    at numericMultipleDomain (http://localhost:1233/vendor/vega.js:28208:24)
    at multipleDomain (http://localhost:1233/vendor/vega.js:28145:29)
    at parseScaleDomain (http://localhost:1233/vendor/vega.js:28118:22)
    at parseScale (http://localhost:1233/vendor/vega.js:28074:13)
    at http://localhost:1233/vendor/vega.js:28670:5
    at Array.forEach (native)
```

The output seems to have weird domain for x-scale

```
{
          "name": "x",
          "type": "linear",
          "round": true,
          "nice": true,
          "zero": false,
          "range": [0,200],
          "domain": {
            "fields": [
              "bin_distance_start", // ??
              "bin_distance_end",
              {
                "data": "layer_1_summary",
                "fields": ["bin_distance_start","bin_distance_end"]
              }
            ]
          }
        },
```

So this is probably relevant to layer domain merge.

domoritz:
Works now.

Issue #1602, opened 2016-10-20
Redo the layering revision PR


kanitw:
In #1304, we did a lot of things. 
We should break down different things that we try to achieve in that PR and list them here, so we can discuss do each of them in more atomic fashion. 
## TODO
- [x] re-order transforms
  - [x] rename data sources
  - [x] NEW Consider where post-aggregation calc and filter would be if we were to implement them.  (Avoid major revision in the future) 
- [ ] When unioning scales, show title of the unioned fields in the axis title.
- [x] Make sure composition works with placeholder data (see #1392)

kanitw:
- [ ] re-order transforms

domoritz:
Steps on the way
- [ ] `domain: "preAggregate`
- [ ] Remove the need for separate raw source for pre aggregated domain by including min/max
- [ ] Add `data: {bind: "A"}` (#1392)

kanitw:
@domoritz When you have time, could you please check if you have achieved everything you wanted to do here.  If so, close the issue.

kanitw:
Hmm, we probably haven't union axis title.

kanitw:
union axis title should be a part of #2251

PR #1866, opened 2017-01-22
Pass validation

### LINKED ISSUE #1792: Make all examples render and validates

domoritz:
Fixes #1792 and #1828. 

The proper fix for the scale merging will be done in #1602. See commit comment for explanation.


src/compile/scale/domain.ts
```diff
@@ -79,10 +79,15 @@ export default function domain(scale: Scale, model: Model, channel:Channel): any
       if (channel === 'x' || channel === 'y') {
         // X/Y position have to include start and end for non-ordinal scale
         return {
-          data: model.dataTable(),
+          // FIXME: use unioned domain where the data is defined once and fields is only an array of strings
```
kanitw: I assume this breaks when the original `{data: ..., fields: ...}` format need to get merge with others?  

If so I wonder if we should modify the merge logic to be responsible for converting `{data: D, fields: [A,B]}` into `{fields: [{data: D, field: A},{data: D, field: B}]}` instead of modifying it here (and thus you can remove FIXME).  

(This should be relatively simple, so I guess we don't have to punt on this.)

src/compile/layer.ts
```diff
@@ -113,6 +113,7 @@ export class LayerModel extends Model {
       child.parseScale();
 
       // FIXME: correctly implement independent scale
+      // also need to check whether the scales are actually compatible, e.g. use the same sort
```
kanitw: - Missing test for this case.
- Why don't we just check it now?  It shouldn't be that hard?

If you really prefer not to, then please change the comment to:
```
// FIXME(#1602): check whether the scales are actually compatible, e.g. use the same sort
```

domoritz:
Now with proper fix.

kanitw:
Test still failing

domoritz:
@kanitw No, tests are passing fine. It's just the test coverage :-P

kanitw:
Yep and i'd call that a type of test failure lol

domoritz:
Fair enough. I'll add tests.

src/compile/scale/domain.ts

src/compile/layer.ts
domoritz: I prefer to properly handle non-unioned domains later.

src/compile/layer.ts
```diff
@@ -104,15 +104,43 @@ export class LayerModel extends Model {
     this.component.layout = parseLayerLayout(this);
   }
 
-  public parseScale() {
+  /**
+   * Convert the domain to an array of data refs. Also, thorw away sorting information
+   * since we always sort the domain when we union two domains.
+   */
+  private static normalizeDomain(domain: DataRefUnionedDomain | FieldRefUnionDomain | VgDataRef): VgDataRef[] {
```
kanitw: I wonder if this method should be here since it will be reused by concat as well.  

This seems like a useful utility.  Maybe it can stay in `scale/domain`?

src/vega.schema.ts
```diff
@@ -17,13 +17,15 @@ export type VgParentRef = {
 
 export type VgFieldRef = string | VgParentRef | VgParentRef[];
 
+export type VgOpSort = boolean | {
```
kanitw: VgSortFieldDef?

src/vega.schema.ts
```diff
@@ -41,14 +43,27 @@ export type VgValueRef = {
   band?: boolean | number
 }
 
-export type UnionedDomain = {
-  fields: VgDataRef[]
+export type DataRefUnionedDomain = {
+  fields: VgDataRef[],
```
kanitw: I wonder if we should ping Jeff about the syntax for this case.  It reads pretty confusing.

src/vega.schema.ts
```diff
@@ -67,16 +82,23 @@ export type VgScale = {
   zero?: boolean
 }
 
-export function isUnionedDomain(domain: any[] | UnionedDomain | VgDataRef): domain is UnionedDomain {
+export function isDataRefUnionedDomainDomain(domain: VgDomain): domain is DataRefUnionedDomain {
```
kanitw: isFieldRefUnionDomainDomain

why repeated `DomainDomain`?  It reads pretty confusing.  One `Domain` seems enough. 

Also why `Union`, not `Unioned` like below?

src/vega.schema.ts
```diff
@@ -41,14 +43,27 @@ export type VgValueRef = {
   band?: boolean | number
 }
 
-export type UnionedDomain = {
-  fields: VgDataRef[]
+export type DataRefUnionedDomain = {
```
kanitw: Would `UnionedDataRefDomain` sound better?

src/vega.schema.ts
```diff
@@ -41,14 +43,27 @@ export type VgValueRef = {
...
+  fields: VgDataRef[],
+  sort?: boolean | {
+    op: 'count'
+  }
+}
+
+export type FieldRefUnionDomain = {
```
kanitw: Would `UnionedFieldRefDomain` sound better?

src/vega.schema.ts
```diff
@@ -67,16 +82,23 @@ export type VgScale = {
...
+    return 'fields' in domain && 'data' in domain;
   }
   return false;
 }
 
-export function isDataRefDomain(domain: any[] | UnionedDomain | VgDataRef): domain is VgDataRef {
+export function isDataRefDomain(domain: VgDomain): domain is VgDataRef {
   if (!isArray(domain)) {
-    return 'data' in domain;
+    return !('fields' in domain);
```
kanitw: Wonder if checking `field` in domain is better.

src/compile/layer.ts
```diff
@@ -104,15 +104,43 @@ export class LayerModel extends Model {
     this.component.layout = parseLayerLayout(this);
   }
 
-  public parseScale() {
+  /**
+   * Convert the domain to an array of data refs. Also, thorw away sorting information
+   * since we always sort the domain when we union two domains.
```
kanitw: Why always sort the domain?

domoritz: Well, if we union two domains, how do you put them together? The only sensible thing is sorting, imho because everything else is not predictable.

src/compile/layer.ts
```diff
@@ -107,15 +107,43 @@ export class LayerModel extends Model {
...
     const model = this;
 
     let scaleComponent = this.component.scale = {};
 
     this._children.forEach(function(child) {
       child.parseScale();
 
-      // FIXME: correctly implement independent scale
+      // FIXME(#1602):correctly implement independent scale
+      // Also need to check whether the scales are actually compatible, e.g. use the same sort or throw error
```
kanitw: This "Also" is #1602, the correctly implement independent isn't.

src/compile/layer.ts
domoritz: Not really because this helper does not operate the full type. Arrays are handled differently.

src/compile/layer.ts
kanitw: Ok.   Two concerns here:

1. If the sort of each data ref aren't compatible, we should throw warning that we drop them.

2. If all data refs have identical sort objects, here you drop all of them and use `true` instead, which is wrong.

src/vega.schema.ts
domoritz: We have `VgSort` below

src/vega.schema.ts
domoritz: Really? I find it understandable.

src/vega.schema.ts
domoritz: No, because it's a unioned domain. So maybe `UnionedDomainFieldRef`.

domoritz: Dude, you can't update your comment to match mine later, lol.

src/vega.schema.ts
kanitw: Ok then.

src/vega.schema.ts
kanitw: So?

src/vega.schema.ts
domoritz: Good catch.

src/vega.schema.ts
domoritz: Sure. Fixed.

src/compile/layer.ts
domoritz: Not really. It's all under the umbrella of follow up fixes.

src/compile/layer.ts
kanitw: So they will be only used by layer, but not concat?  

Btw, FYI, we haven't used static method at all -- we mostly extracted then as a separate function.

src/vega.schema.ts
domoritz: Not really.

kanitw: Use UnionDomain then so it's consistent with FieldRefUnionDomain

src/vega.schema.ts
domoritz: I don't think VgSortFieldDef is a better name.

src/vega.schema.ts
domoritz: Fixed to use `DataRefUnionDomain` and `FieldRefUnionDomain`.

src/vega.schema.ts
kanitw: VgSortField? 

The intention for this is to sort by another field, not by the scale's field itself.  
Then you need to add Op to disambiguate which value of the other field you want to use. 

Thus, the matter here is that it's a sort field.  Op is secondary.

domoritz: Okay

src/compile/layer.ts
kanitw: >If all data refs have identical sort objects, here you drop all of them and use true instead, which is wrong.

Actually if I understand correctly this case isn't supported in Vega anyway (if there are more than 1 unique fields)?

domoritz: Yes. The only valid op is `count`.

domoritz:
I believe this branch is failing because of https://github.com/vega/vega-parser/pull/17

src/compile/layer.ts
kanitw: Hmm what about this case?

> If the sort of each data ref aren't compatible, we should throw warning that we drop them.

still haven't been handled.

src/compile/layer.ts
```diff
@@ -107,15 +107,43 @@ export class LayerModel extends Model {
     this.component.layout = parseLayerLayout(this);
   }
 
-  public parseScale() {
+  /**
+   * Convert the domain to an array of data refs. Also, thorw away sorting information
+   * since we always sort the domain when we union two domains.
+   */
+  private static normalizeDomain(domain: DataRefUnionDomain | FieldRefUnionDomain | VgDataRef): VgDataRef[] {
```
kanitw: Somehow this comment was collapsed.  

- FYI, we rarely use static method.  Extracting it as a function on its own would be more consistent + make function call shorter (just `normalizeDomain` rather than `LayerModel.normalizeDomain`.  

- Even if your comment below it true, I suspect that we will still re-use it in concat?

> Not really because this helper does not operate the full type. Arrays are handled differently.

src/compile/mark/common.ts
```diff
@@ -21,8 +21,8 @@ export function applyColorAndOpacity(e: VgEncodeEntry, model: UnitModel) {
     applyMarkConfig(e, model, STROKE_CONFIG);
   }
 
-  let colorRef= ref.midPoint('color', model.encoding().color, model.scaleName('color'), model.scale('color'), undefined);
-  let opacityRef = ref.midPoint('opacity', model.encoding().opacity, model.scaleName('opacity'), model.scale('opacity'), {value: config.mark.opacity});
+  let colorRef = ref.midPoint('color', model.encoding().color, model.scaleName('color'), model.scale('color'), undefined);
+  let opacityRef = ref.midPoint('opacity', model.encoding().opacity, model.scaleName('opacity'), model.scale('opacity'), config.mark.opacity && {value: config.mark.opacity});
```
kanitw: I wonder why using 
```
config.mark.opacity && {value: config.mark.opacity}
```
doesn't fail the TS compiler since `config.mark.opacity && {value: config.mark.opacity}` can be `false` and does not match the type...

src/compile/layer.ts
domoritz: Okay, I extracted a method

src/compile/layer.ts
domoritz: Because we don't allow not unioning the domains.

PR #1867, opened 2017-01-22
[WIP] Support short form of `{range: {scheme: <string>}}` with `{range: string}`

kanitw:
Please merge #1865 first.

Not sure if we should do this yet -- just an idea.  

`{range: {scheme: <string>}}` seems a bit annoyingly nested.

### DIFF
--- a/examples/specs/stacked_area.vl.json
+++ b/examples/specs/stacked_area.vl.json
@@ -11,7 +11,7 @@
     "y": {
       "aggregate": "sum", "field": "count","type": "quantitative"
     },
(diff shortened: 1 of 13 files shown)


kanitw:
Actually this wouldn't work.

kanitw:
Actually it could work, but I don't know if we should introduce alias.
That said, 
`... scale:{range: {scheme: <string>}}` seems a bit annoyingly nested.
`... scale:{range: <string>}` really seems a bit more humane to type.

Issue #1870, opened 2017-01-23
Update output {rangeStep: ...} to be {range: {step: ...}}


kanitw:

kanitw:
Fixed as a part of the pending https://github.com/vega/vega-lite/pull/1876

PR #1876, opened 2017-01-23
Correct Vega schema for rangeStep

kanitw:
Fixes https://github.com/vega/vega-lite/issues/1870

### DIFF
--- a/src/compile/layer.ts
+++ b/src/compile/layer.ts
@@ -1,5 +1,3 @@
-import * as log from '../log';
-
 import {Channel} from '../channel';
 import {defaultConfig, CellConfig, Config} from '../config';
 import {FieldDef} from '../fielddef';
@@ -7,7 +5,7 @@ import {LayerSpec} from '../spec';
 import {StackProperties} from '../stack';
 import {FILL_STROKE_CONFIG} from '../mark';
 import {keys, duplicate, mergeDeep, flatten, unique, isArray, vals, hash} from '../util';
-import {VgData, isDataRefUnionedDomain, isFieldRefUnionDomain, isDataRefDomain, VgDataRef, VgEncodeEntry, DataRefUnionDomain, FieldRefUnionDomain} from '../vega.schema';
+import {VgData, isDataRefUnionedDomain, isFieldRefUnionDomain, isDataRefDomain, VgDomain, VgDataRef, VgEncodeEntry} from '../vega.schema';
 import {isUrlData} from '../data';
 
 import {assembleData, parseLayerData} from './data/data';
@@ -19,11 +17,13 @@ import {UnitModel} from './unit';
 import {ScaleComponents} from './scale/scale';
 
 /**
- * Convert the domain to an array of data refs. Also, throw away sorting information
- * since we always sort the domain when we union two domains.
+ * Convert the domain to an array of data refs or an array of values. Also, throw
+ * away sorting information since we always sort the domain when we union two domains.
  */
-function normalizeDomain(domain: DataRefUnionDomain | FieldRefUnionDomain | VgDataRef): VgDataRef[] {
-  if (isDataRefDomain(domain)) {
+function normalizeDomain(domain: VgDomain): (any[] | VgDataRef)[] {
+  if (isArray(domain)) {
+    return [domain];
+  } if (isDataRefDomain(domain)) {
     delete domain.sort;
     return [domain];
   } else if(isFieldRefUnionDomain(domain)) {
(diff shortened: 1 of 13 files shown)


src/compile/scale/parse.ts
```diff
@@ -80,20 +80,24 @@ function parseMainScale(model: Model, channel: Channel) {
...
+
+  // FIXME(#): Can we always guarantee that the domain that we union is not already unioned?
+  type NonUnionDomain = any[] | VgDataRef;
+
   // If channel is either X or Y then union them with X2 & Y2 if they exist
   if (channel === X && model.channelHasField(X2)) {
     if (model.channelHasField(X)) {
       // FIXME: Verify if this is really correct
-      return { fields: [domain(scale, model, X), domain(scale, model, X2)] };
+      return { fields: [domain(scale, model, X) as NonUnionDomain, domain(scale, model, X2) as NonUnionDomain] };
```
domoritz: I have a much better solution. I extracted a unionDomains method that we can use here and in layer.ts.

domoritz:
@kanitw Done. I cannot check the box in your review.

src/compile/layer.ts
```diff
@@ -158,42 +131,7 @@ export class LayerModel extends Model {
...
-                    fields: fields.map(f => f.field),
-                    sort: true
-                  };
-                }
-              } else {
-                modelScales.main.domain = fields[0];
-              }
-            }
-
+            modelScales.main.domain = unionDomains(modelScales.main.domain, childScales.main.domain);
```
kanitw: Are the two lines of comments above still up to date?

src/compile/scale/domain.ts
```diff
@@ -162,3 +162,67 @@ function _useRawDomain (scale: Scale, model: Model, channel: Channel) {
...
+
+/**
+ * Convert the domain to an array of data refs or an array of values. Also, throw
+ * away sorting information since we always sort the domain when we union two domains.
+ */
+function normalizeDomain(domain: VgDomain): (any[] | VgDataRef)[] {
+  if (util.isArray(domain)) {
+    return [domain];
+  } if (isDataRefDomain(domain)) {
+    delete domain.sort;
```
kanitw: This still doesn't throw warning, I just added #1895 for this so we don't have to deal with it now.

src/compile/scale/domain.ts
```diff
@@ -162,3 +162,67 @@ function _useRawDomain (scale: Scale, model: Model, channel: Channel) {
...
+ */
+export function unionDomains(domain1: VgDomain, domain2: VgDomain): VgDomain {
+  const normalizedDomain1 = normalizeDomain(domain1);
+  const normalizedDomain2 = normalizeDomain(domain2);
+
+  let fields = normalizedDomain1.concat(normalizedDomain2);
+  fields = util.unique(fields, util.hash);
+
+  if (fields.length > 1) {
+    const datas = fields.map(d => {
```
kanitw: `data` is already plural

kanitw: maybe `allData`?

src/compile/scale/domain.ts
```diff
@@ -162,3 +162,67 @@ function _useRawDomain (scale: Scale, model: Model, channel: Channel) {
...
+}
+
+/**
+ * Union two data domains. A unioned domain is always sorted.
+ */
+export function unionDomains(domain1: VgDomain, domain2: VgDomain): VgDomain {
+  const normalizedDomain1 = normalizeDomain(domain1);
+  const normalizedDomain2 = normalizeDomain(domain2);
+
+  let fields = normalizedDomain1.concat(normalizedDomain2);
```
kanitw: I know this is `fields` in the Vega property, but it's not really an array of "field" but rather either an array of `domain array` or `DataRef`.  I wonder if `domains` is better?  (This is partly why I think `fields` is a pretty confusing property name.) 

I wonder if VSCode refactor would do this correct though.  Maybe You'll need to change `{fields, sort: true}` below to `{fields: domains, sort: true}`.  (If so, this could one problem for `{fields}` syntax.)

src/compile/scale/domain.ts
```diff
@@ -191,9 +192,9 @@ function normalizeDomain(domain: VgDomain): (any[] | VgDataRef)[] {
         data: d.data
       };
     });
+  } else {
+    throw 'Should not get here.';
```
domoritz: To be consistent

```ts
/* istanbul ignore next: This should never happen. */
  throw new Error(log.message.SOMETHING);
```

src/compile/scale/domain.ts
domoritz: At least ts catches that issue because it expects a `fields` property.

src/compile/layer.ts
domoritz: Yes

src/compile/scale/domain.ts
kanitw: But that would make one branch untested (hidden else for that case). 

(Yep, in the past I was dumb and didn't do this, but will follow this pattern going forward.)

PR #1887, opened 2017-01-25
Update schema

domoritz:
See individual commits for description.

### DIFF
--- a/scripts/pre-deploy.sh
+++ b/scripts/pre-deploy.sh
@@ -6,16 +6,13 @@ set -e
 RED='\033[0;31m'
 NC='\033[0m' # No Color
 
-# 0.1 Check if jq has been installed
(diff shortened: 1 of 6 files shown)


scripts/pre-deploy.sh
```diff
@@ -6,16 +6,13 @@ set -e
 RED='\033[0;31m'
 NC='\033[0m' # No Color
 
-# 0.1 Check if jq has been installed
-type jq >/dev/null 2>&1 || { echo >&2 "I require jq but it's not installed.  Aborting."; exit 1; }
```
domoritz: I removed the need for jq.

kanitw: Nice!

domoritz: Spent only two hours on getting the sed syntax right...

PR #1890, opened 2017-01-25
Escapes col names containing dots per Vega (fixes #1175)

willium:
Should be a quick fix.

Tested and verified with 
```
{
  "description": "A simple bar chart with a field name containing a dot.",
  "data": {
    "values": [
      {"a.a": "A","b": 28}, {"a.a": "B","b": 55}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "a.a", "type": "ordinal"},
  }
}
```

Aside: we don't seem to have any explicit test coverage of `field()` in `fieldDef.ts`... is this intentional?

### DIFF
--- a/src/fielddef.ts
+++ b/src/fielddef.ts
@@ -156,6 +156,8 @@ export function field(fieldDef: FieldDef, opt: FieldRefOption = {}) {
 
   if (opt.datum) {
     field = `datum["${field}"]`;
+  } else if (field !== undefined && field.indexOf('.') !== -1) { // dot in fieldname
+    field = `[${field}]`;
   }
 
   return field;


kanitw:
> Aside: we don't seem to have any explicit test coverage of field() in fieldDef.ts... is this intentional?

Nope.

src/fielddef.ts
```diff
@@ -156,6 +156,8 @@ export function field(fieldDef: FieldDef, opt: FieldRefOption = {}) {
 
   if (opt.datum) {
     field = `datum["${field}"]`;
+  } else if (field !== undefined && field.indexOf('.') !== -1) { // dot in fieldname
```
kanitw: I wonder if we should just leave as is and tell users to use `[a.a]`
if they want dot in the name and then use `a.a` for nested JSON

domoritz: I'm not a fan of nested JSON and would rather support the more common case of . in field names well.

kanitw: @domoritz  Hmm, but by doing what you suggest would support only one thing if we stick with what Vega already does for us we support both

domoritz: I think asking people to escape with '\' is completely reasonable.

kanitw: > I think asking people to escape with '' is completely reasonable.


Im confused.   So are you saying that we do not need to change anything here, but rather update the document to clarify this?

willium: I think they need to escape with []

kanitw: `[a.a]`, `['a.a']`, or `a\\.a` right?

domoritz: @kanitw \'\\' in markdown becomes '\'

kanitw: @domoritz Can you elaborate on that?  I'm confused.

domoritz: I agree with you.

kanitw: So close this PR and leave the issue open until we add appropriate docs?

jeffbaumes: I'll just mention here that in R, `.` is almost always referring to a dot within the column name, as opposed to nesting. Even the basic example datasets like `iris` include columns such as `sepal.length`. I agree if we always wrap in brackets then nested fields are impossible, but it will mean that potential libraries wrapping Vega-Lite in R may need to always wrap fields in brackets internally to be usable, which is ok but should be documented well.

kanitw: @jeffbaumes Thanks for the heads-up!

PR #1900, opened 2017-01-27
[WIP] removed AxisOrient namespace

### LINKED ISSUE #1675: Get rid of string literal constants?
A lot of properties in Vega-Lite are string enums.  However, enums in TS are numbers, not strings.  Instead, we use string literals to define these string-based enum instead.  However, currently the way we declare types of string literals is quite an overkill:

```
export namespace AxisOrient {
    export const TOP: 'top' = 'top';
    export const RIGHT: 'right' = 'right';
    export const LEFT: 'left' = 'left';
    export const BOTTOM: 'bottom' = 'bottom';
}
export type AxisOrient = typeof AxisOrient.TOP | typeof AxisOrient.RIGHT | typeof AxisOrient.LEFT | typeof AxisOrient.BOTTOM;
```

Except for frequently used items like `Mark`, `Channel`, and `Type` (and maybe `ScaleType`), the namespace seems unnecessary.

Instead, simply do 
```
export type AxisOrient = 'top' | 'right' | 'left' | 'bottom'
```
should be sufficient. 

Thus, for all **top-level files** under `src/`, let's 

- [ ] get rid of the overkill pattern for declaring string literals described above and replace them with simple string literal type declaration (except for `Mark`, `Channel`, `Type`, and `ScaleType`)

- [ ] Update their references in the codebase with the right string values.

nung25649:
Fix #1675 
Test error showed up in `fielddef.test.ts` and `stack.test.ts` if I didn't put the type casting in the string literals that I replace with the removed namespace. It will show that the new replacing string literals are of type `string` instead of `AxisOrient`.

### DIFF
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -1,6 +1,6 @@
 import * as log from '../log';
 
-import {AggregateOp} from '../aggregate';
+// import {AggregateOp} from '../aggregate';
 import {TEXT, Channel} from '../channel';
 import {Config, CellConfig} from '../config';
 import {FieldDef, OrderFieldDef, field} from '../fielddef';
@@ -60,7 +60,7 @@ export function numberFormat(fieldDef: FieldDef, format: string, config: Config,
 
     if (format) {
       return format;
-    } else if (fieldDef.aggregate === AggregateOp.COUNT && channel === TEXT) {
+    } else if (fieldDef.aggregate === 'count' && channel === TEXT) {
       // FIXME: need a more holistic way to deal with this.
       return 'd';
     }
--- a/test/fielddef.test.ts
+++ b/test/fielddef.test.ts
@@ -49,11 +49,14 @@ describe('fieldDef', () => {
     });
 
     it('should return correct title for aggregate', () => {
+      // if "aggregate: 'mean' as 'mean'" is changed to "aggregate: 'mean'", it won't pass the test because
+      // it will show that type of 'mean' is string instead of 'mean'
       const fieldDef = {field: 'f', type: QUANTITATIVE, aggregate: 'mean' as 'mean'};
       assert.equal(title(fieldDef, {}), 'MEAN(f)');
     });
 
     it('should return correct title for count', () => {
+      // The same test error shows up here if 'count' as 'count' is changed to only 'count'
       const fieldDef = {field: '*', type: QUANTITATIVE, aggregate: 'count' as 'count'};
       assert.equal(title(fieldDef, {countTitle: 'baz!'}), 'baz!');
     });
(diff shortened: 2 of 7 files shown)


test/stack.test.ts
```diff
@@ -247,7 +246,9 @@ describe('stack', () => {
 
   it('should always be disabled if the aggregated axis has non-summative aggregate', () => {
     [undefined, StackOffset.CENTER, StackOffset.ZERO, StackOffset.NORMALIZE].forEach((stacked) => {
-      [AggregateOp.AVERAGE, AggregateOp.VARIANCE, AggregateOp.Q3].forEach((aggregate) => {
+      // ['average', 'variance', 'q3'] will be interpreted as array of string instead of
+      // array of AxisOrient type, so I put type casting here.
+      ['average' as 'average', 'variance' as 'variance', 'q3' as 'q3'].forEach((aggregate) => {
```
kanitw: @nung25649  Instead of adding `as` everywhere, you probably want to just do
```
['average', ...].forEach((aggregate: AggregateOp) => ...) 
```

test/fielddef.test.ts
```diff
@@ -50,12 +49,15 @@ describe('fieldDef', () => {
     });
 
     it('should return correct title for aggregate', () => {
-      const fieldDef = {field: 'f', type: QUANTITATIVE, aggregate: AggregateOp.MEAN};
+      // if "aggregate: 'mean' as 'mean'" is changed to "aggregate: 'mean'", it won't pass the test because
+      // it will show that type of 'mean' is string instead of 'mean'
+      const fieldDef = {field: 'f', type: QUANTITATIVE, aggregate: 'mean' as 'mean'};
```
kanitw: @nung25649 Actually this is a valid point and seems annoying. 

However, removing `fieldDef` and just pass the object to the function works. 
```
assert.equal(title({field: 'f', type: QUANTITATIVE, aggregate: 'mean'}, {}), 'MEAN(f)');
```
(I guess before the function know that this object is a FieldDef type.)

Alternatively, we can add type FieldDef to 
```
const fieldDef: FieldDef = ...
```

but the former solution seems shorter.

src/compile/common.ts
```diff
@@ -1,6 +1,6 @@
 import * as log from '../log';
 
-import {AggregateOp} from '../aggregate';
+// import {AggregateOp} from '../aggregate';
```
kanitw: Please remove commented lines

test/stack.test.ts
domoritz: Is `as` really necessary? I'd prefer `typeof` if we cannot just omit it.

kanitw: Just commented a [better solution](https://github.com/vega/vega-lite/pull/1900#discussion_r98308829)

PR #1903, opened 2017-01-29
[WIP] remove AggregateOp namespace

nung25649:
Sorry for creating another pull request, I just created a new branch to refactor the namespace.

### DIFF
--- a/src/aggregate.ts
+++ b/src/aggregate.ts
@@ -1,8 +1,4 @@
 
-export namespace AggregateOp {
-
-}
-
 export type AggregateOp = 'argmax' | 'argmin' | 'average' | 'count'
   | 'distinct' | 'max' | 'mean' | 'median' | 'min' | 'missing' | 'modeskew'
   | 'q1' | 'q3' | 'stdev' | 'stdevp' | 'sum' | 'valid' | 'values' | 'variance'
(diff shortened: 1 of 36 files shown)


src/aggregate.ts
```diff
@@ -1,75 +1,50 @@
...
-    AggregateOp.STDEVP,
-    AggregateOp.MEDIAN,
-    AggregateOp.Q1,
-    AggregateOp.Q3,
-    AggregateOp.MODESKEW,
-    AggregateOp.MIN,
-    AggregateOp.MAX,
-    AggregateOp.ARGMIN,
-    AggregateOp.ARGMAX,
+    'values',
```
kanitw: add  `: AggregateOp` to three of these arrays

src/compile/data/summary.ts
```diff
@@ -1,4 +1,4 @@
-import {AggregateOp} from '../../aggregate';
+// import {AggregateOp} from '../../aggregate';
```
kanitw: Always clean up unused lines, unless you have compelling reasons.

src/aggregate.ts
nung25649: I tried that, but it says AggregateOp it is not assignable to string[]. Should I make that AggregateOp[]?

PR #1906, opened 2017-01-31
[WIP] Labeling

willium:
Beginning re-implementation of labeling

- [ ] Add TextTable Example back

https://github.com/vega/vega/blob/master/spec/wordcloud.vg.json


willium:
For lines
https://github.com/vega/vega-lite/blob/feature/labeling/src/compile/data/formatparse.ts#L65
and
https://github.com/vega/vega-lite/blob/feature/labeling/src/compile/layer.ts#L256

how can I incorporate `isInternalData` @kanitw?

PR #1907, opened 2017-01-31
Add Range Config to schema.

kanitw:
(These should be just passed to Vega)


vega-lite-schema.json
```diff
@@ -946,6 +950,12 @@
             },
             "type": "object"
         },
+        "Dict<VgRangeScheme | (string | number)[] | { step: number; }>": {
+            "additionalProperties": {
+                "$ref": "#/definitions/T"
```
kanitw: @domoritz  Looks like this part is still wrong for your schema generator?

Issue #1773, opened 2016-12-12
Add a Unit Test to test if the vega-lite-schema.json is up to date


kanitw:
## Problem 

Currently, we use `npm run schema` to create a JSON schema and include the output `vega-lite-schema.json` in our commits. This allows us to make sure that we don't accidentally introduce schema changes. However, sometimes we forget to run the command to update the schema when we update our syntax in the code. Meanwhile, our test (and Travis) do not fail if we forget to do so.  

## Solution

We should add a test in Travis that throws an error when `vega-lite-schema.json` is out-of-date. The schema is out of date when `npm run schema` creates a new schema that is different that the committed one.  

To implement this add a test that runs `npm run schema` and checks whether git status says that there have been changes to `vega-lite-schema.json`. Make sure that travis tests fail in that case.

domoritz:
Can you explain this issue again? I don't understand it.

kanitw:
@domoritz Ok updated the description above.   
@YuhanLu I think this one has lower priority compared to other issues assigned to you.

PR #1908, opened 2017-02-01
[WIP] Add a Unit Test to test if the vega-lite-schema.json is up to date #1773

nung25649:
I just started working on this pull request and I am stuck on how to proceed and what is the right way to fix the issue. So far I have added the `npm run schema`, but I'm not sure how I should make the test fail if the schema file is different than the committed one. Do I add another test case inside `schema.test.ts` or write a shell script to execute when the `npm run test` is called to check git status?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ node_js:
 script:
   - npm run build
   - npm run test:noschema # already run schema in build
+  - npm run checkschema
   - npm run codecov
 after_success:
   - bash <(curl -s https://codecov.io/bash)
--- a/scripts/check-schema.sh
+++ b/scripts/check-schema.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+if ! git diff --exit-code HEAD -- ./vega-lite-schema.json
+then
+  echo "vega-lite-schema.json is different from the committed one."
+  exit 1
+else
+  exit 0
+fi
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "build:images": "npm run data && scripts/generate-images.sh",
     "build:toc": "bundle exec jekyll build --incremental -q && scripts/generate-toc",
     "build:test-gallery": "browserify test-gallery/main.ts -p [tsify -p test-gallery] -d > test-gallery/main.js",
+    "checkschema": "./scripts/check-schema.sh",
     "clean": "rm -f vega-lite.* & find -E src test site -regex '.*\\.(js|js.map|d.ts)' -delete & rm -rf examples/_diff examples/_original examples/_output examples/images && rm -rf data",
     "data": "rsync -r node_modules/vega-datasets/data/* data",
     "deploy": "scripts/deploy.sh",


domoritz:
I think we need a shell script or simply another script in `package.json` that checks for git changes and returns an error status when the schema has been changed. I believe we don't want to call this test every single time we run `npm run test`, though because this would be annoying when we develop locally.

kanitw:
@nung25649  is this finished or just a WIP? If WIP, add [WIP] to the title, so I'm not confused.  Thanks!

nung25649:
Yes it is! Sorry I forgot that. 

One more question: How do we make the travis test throw an error if our bash file says that the committed schema doesn't match our current version of schema? Is it okay to just `echo "schema doesn't match` in our bash file so that the output appears when we run test?

domoritz:
@nung25649 I believe if the script returns a [unix exist status != 0](http://bencane.com/2014/09/02/understanding-exit-codes-and-how-to-use-them-in-bash-scripts/), builds will fail on travis.

nung25649:
I just finished writing a script. Please check if I skipped or missed something as I'm pretty new to bash shell script.

.travis.yml
```diff
@@ -5,6 +5,7 @@ node_js:
   - "7"
 script:
   - npm run build
+  - npm run schema
```
kanitw: Not sure what you need to add run schema -- build already run schema

kanitw: Not sure why you need to add run schema -- build already run schema

kanitw:
You haven't run the script you wrote from travis at all, so it's definitely unfinished.  

What you should try is:

1) Write a proper script that travis will call.  Then push, your first commit should pass the test. 
2) Add another commit to modify an interface of Vega-Lite (e.g. add string "aaa" as a new valid string literal to `AggregateOp`) and test if you push the code (without the updated schema) to this PR and travis should fail in this second commit. 

After you successfully do this, we can merge your code but ignore the commit that you add to demonstrate that test will fail in a desired situation.

scripts/check-schema.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+npm run schema
```
kanitw: build already run schema, so you definitely don't need to run twice

test/schema.test.ts
```diff
@@ -22,4 +22,6 @@ describe('Schema', function() {
     }
     assert.equal(valid, true);
   });
+
+  // write another test case here to check that the schema file is different than the committed one
```
kanitw: As @domoritz mentioned, you probably don't have to do this -- unless I misunderstand something.

nung25649:
Okay, I misunderstood the issue. I thought that travis test is run when call `npm run test`, but it is actually the test that runs on github right?

If so, then how can I tell travis to run the script? Do I add that in `.travis.yml`?

kanitw:
> If so, then how can I tell travis to run the script? Do I add that in .travis.yml?
Yes.  For more information, look at Travis documentation online.

.travis.yml
```diff
@@ -6,7 +6,7 @@ node_js:
 script:
   - npm run build
   - npm run schema
-  - npm run test:noschema # already run schema in build
+  - npm run test:noschema && npm run checkschema # already run schema in build
```
domoritz: You can have the commands in separate lines.

nung25649:
Hmm, the scripts work on my local when I changed the schema file and run the script. However, when I test the script on travis by modifying one of the source file and then push, it didn't fail the test as I expected. Is there any specific details about travis that I might not know?

kanitw:
@nung25649 – How about you try to push a commit that suppose to fail and open travis's test log.  That might give you a clue what's wrong.

kanitw:
@nung25649 Can you add a commit that suppose to fail to demonstrate that your script will fail in that case? Otherwise I can't know if it's working correctly.  (If it looks correctly, I'll then merge your stuff ignoring the failure test commit.)

kanitw:
@nung25649  Sweet.  The test looks looks correct now. 

1. Looks like this PR accidentally brings back `namespace AggregateOp` which seems unnecessary -- not sure if you have noticed that. (In general, always review all your changes on github as if you are a reviewer first before asking us to review -- this will help us reduce obvious things to fix and improve review turnaround time. 😄  )

I believe that the root cause is from resolving conflicts incorrectly.  


2.  Since this is on your repo (not the main Vega-Lite repo), it's harder for me to do selective merge myself.  Can you do interactive rebase to drop the "test fail case" and "merge branch ..." commits and force push?  (If you haven't, this is a good time to learn interactive rebase.) 

To summarize, you should do interactive rebase to
1) drop the "test fail case" and "merge branch ..." commits


2. since we are using rebase workflow, in general you should not merge master into your branch.

kanitw:
(A few corrections -- I accidentally hit comment before finish drafting)

@nung25649  Sweet.  The test looks looks correct now.   However:

1. Looks like this PR accidentally brings back `namespace AggregateOp` which seems unnecessary -- not sure if you have noticed that. (In general, always review all your changes on github as if you are a reviewer first before asking us to review -- this will help us reduce obvious things to fix and improve review turnaround time. 😄  )

I believe that the root cause is from forking from older master (older than we merge your previous PR that remove AggregateOp).

2.  Since this is on your repo (not the main Vega-Lite repo), it's harder for me to do selective merge myself.  Can you do interactive rebase to drop the "test fail case" and "merge branch ..." commits and force push?  
(Since we are using rebase workflow, in general you should not merge master into your branch.)

To summarize, you should do interactive rebase to
1) rebase to the latest `origin/master`  (`git fetch` then `git rebase -i origin master`)
2) resolve conflicts correctly (don't bring `namespace AggregateOp` back).
2) drop the "test fail case" and "merge branch ..." commits 

 (If you haven't, this is a good time to learn interactive rebase.)

nung25649:
@kanitw Thanks! I also squashed every commit into one. Please let me know if there's still any changes!

kanitw:
@nung25649  Awesome.  This looks pretty neat now!

PR #1911, opened 2017-02-03
Fix some linting warnings

willium:

### DIFF
--- a/src/compile/mark/mark.ts
+++ b/src/compile/mark/mark.ts
@@ -1,5 +1,5 @@
 import {X, Y, COLOR, NONSPATIAL_CHANNELS, Channel} from '../../channel';
-import {AREA, LINE, TEXT as TEXTMARK} from '../../mark';
+import {AREA, LINE, TEXT as TEXT_MARK} from '../../mark';
 import {contains, without} from '../../util';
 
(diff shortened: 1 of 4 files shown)


domoritz:
Hehe, I like this but @kanitw was not happy about this. https://github.com/vega/vega-lite/commit/7a2d1de43c4df9d5fc60158e9ab60eb86c43f28c

willium:
:( unfortunately it's now deprecated. Is there a middle ground option here? Can we enforce the more strict 
```
"noUnusedLocals": true,
"noUnusedParameters": true,
```

domoritz:
The problem with compiler options is that you cannot compile code until all of them are fixed. This can be annoying while you are debugging. The only option I see is having separate tsconfig for dev and build but that's a nightmare to maintain.

willium:
Ah okay.
We can close/ignore this PR then. I took the other thing I changed in here and attached it to #1912 
https://github.com/vega/vega-lite/pull/1912/commits/e99a6cfabeae33ca9e48593b9accb7a9f6168200

willium:
@domoritz I'll get a PR in one of these days :P

kanitw:
We can just ignore the deprecate warning.  It's totally annoying decision for them to deprecate this.

PR #1917, opened 2017-02-04
Basic refactor

willium:
- Standardize whitespace in objects (remove it everywhere)


kanitw:
@willium This is good.  I wonder how hard it is to do this.  Did you use any automatic tool? 

One thing that I'm worried right now is that it will cause a conflicts for other people.

domoritz:
Let's merge this.

kanitw:
@domoritz  I'm not gonna merge until we get the "improve spec" PR in ...

PR #1920, opened 2017-02-05
[WIP] Make line support size to remove unnecessary special code path

kanitw:
(If size is aggregate value, the whole line would have the same size anyway, so this is actually ok)


kanitw:
This is already done when I rewrite the mark compiler.

PR #1928, opened 2017-02-07
Add stack to x/y encoding channel

kanitw:


src/compile/unit.ts
```diff
@@ -65,7 +65,6 @@ export class UnitModel extends Model {
     // TODO?: ideally we should use config only inside this constructor
     const config = this._config = this._initConfig(spec.config, parent);
 
-    // FIXME move stacked out of config as it's not really a theme.
     // calculate stack properties
     this._stack = stack(mark, encoding, config.mark.stacked);
```
kanitw: @domoritz What bugs me a bit is internally stack is a Spec level operation, not an encoding level operation.

domoritz: I see what you mean but I think conceptually this is where it belongs.

domoritz: Or at least where people might expect it.

domoritz:
LGTM. Why are we not removing stack from anywhere?

kanitw:
@domoritz  IMO, they can still be a config, but shouldn't be the main way for people to change it.
I wonder if it should be `config.mark.stack` or `config.stack` instead of `config.mark.stacked`.

domoritz:
I'd say remove it from config. In the config is like a theme, I don't see why stack would be in there.

kanitw:
>  I don't see why stack would be in there.

Maybe your project want things to be stacked by default or not.  
For example, CompassQL users might want to enumerate stack and make things not stack unless it's explicitly specified.

domoritz:
Okay. In any case, `config.stack` makes more sense now.

kanitw:
Fixed!

domoritz:
But broke the build.

kanitw:
should be good now.

PR #1930, opened 2017-02-08
Add format for text field def

### LINKED ISSUE #1762: Format text without config
I want to do

```json
{
  "data": {"values": [{"m_teps": 859}]},
  "mark": "text",
  "encoding": {
  }
}
```

But have to 

```json
{
  "data": {"values": [{"m_teps": 859}]},
  "encoding": {
    "text": {"field": "m_teps","type": "Q"}
  },
  "mark": "text",
  "config": {
    "numberFormat": "d"
  }
}
```

NeelMohapatra:
Also remove config.text.format, Fixes #1762

### DIFF
--- a/src/fielddef.ts
+++ b/src/fielddef.ts
@@ -93,6 +93,7 @@ export interface OrderFieldDef extends FieldDef {
 }
 
 export interface TextFieldDef extends FieldDef {
+  // FIXME: add more reference to Vega's format pattern or d3's format pattern. 
   /**
    * The formatting pattern for text value. If not defined, this will be determined automatically.
    */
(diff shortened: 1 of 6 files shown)


src/fielddef.ts
```diff
@@ -92,9 +92,13 @@ export interface OrderFieldDef extends FieldDef {
   sort?: SortOrder;
 }
 
+export interface TextFieldDef extends FieldDef {
+  format?: string;
```
kanitw: Missing description -- you can copy this from the removed config.

NeelMohapatra: /**
   * The formatting pattern for text value. If not defined, this will be determined automatically.
   */
Do I just add this on top of it?

PR #1931, opened 2017-02-08
Make UnitSpec generic parameter for `LayerSpec` and `FacetSpec`

kanitw:


src/spec.ts
```diff
@@ -79,10 +79,12 @@ export interface GenericUnitSpec<M, E extends Encoding> extends BaseSpec {
 
 export type UnitSpec = GenericUnitSpec<Mark, Encoding>;
 
+export type LayeredUnitSpec = GenericUnitSpec<AnyMark, Encoding>;
+
 
 export type FacetedUnitSpec = GenericUnitSpec<AnyMark, EncodingWithFacet>;
 
-export interface GenericLayerSpec<M, E> extends BaseSpec {
+export interface GenericLayerSpec<U> extends BaseSpec {
```
domoritz: You can put constraints on generics. See https://www.typescriptlang.org/docs/handbook/generics.html

`GenericLayerSpec<U extends BaseSpec>`?

vega-lite-schema.json
```diff
@@ -1341,7 +1341,7 @@
             ],
             "type": "object"
         },
-        "GenericFacetSpec<AnyMark, EncodingWithFacet>": {
+        "GenericFacetSpec<GenericUnitSpec<AnyMark, EncodingWithFacet>>": {
```
domoritz: That's barely understandable.

kanitw: I guess you can modify your generator to name things based on their alias.

src/spec.ts
kanitw: I just make another commit for `U extends GenericUnitSpec`.  (Let's squash after we are happy with other things.)

vega-lite-schema.json
domoritz: Yeah, it's called --aliasRef :-P

kanitw: Oh and this is something you say you should turn it on at some point?

domoritz: Yes. So we get nice names :-P

domoritz: But we need to make sure we don't break anything.

kanitw: Then I guess we can merge this, assuming that we will turn that nice name flag on soon?

Issue #1882, opened 2017-01-24
Design syntax for customizing mark's interpolate, orientation, and stack


kanitw:
Previously users can always modify directly scale/axis/legend properties, but sometimes need to  use mark config to set properties of a mark.  We want to have a more discoverable place for all mark properties previously "hidden" in `config.mark.*`.

In Vega-Lite 2.0, we will support composition of single view plots to create multi-view and layered plots.  We plan to make `config` only a top-level config for theming and let users directly modify non-config properties to customize things.  Given that we will not make config a part of unit specification (of a single plot) in a composite spec, this "more discoverable place" for mark properties become even more needed.  

## New Encoding Channels

For mark properties that can vary across different marks in a unit spec such as `strokeDash`, they can become new encoding channels with proper field mapping support.    

```TS
{
   mark: 'line',
   encoding: {
       ...
      strokeDash: {field: ..., ... } or {value: ...}
   }
}
```

## Orient / Interpolate / Tension

However, some properties (`orient`, `interpolate`, and `tension`) should be consistent for all marks of the same unit plots and thus should not be an encoding channel.  

For orient / interpolate / tension, we are thinking about an interface like this: 

```TS
type Mark = MarkType | MarkDef;

interface MarkDef {
   mark: MarkType;

   orient: Orient; // one of 'horizontal' | 'veritical'

   // line and area
   interpolate: Interpolate; // Different line interpolator name
   tension: number;
}
```

so a mark can be a string like before, or can be an object.  For example, if users want to customize line's interpolate and tension

```
{
  mark: {
     type: 'line',
     interpolate: 'basis',
     tension: 1
  },
  ...
}
```
## Stack

For `stack`, we are thinking about adding it as a property of x and y encoding channels (for describing different stack offset (`"zero", "center", "normalize", null`).

For example, to make a stacked chart with center baseline (streamgraph), one can use an object form of a mark definition:

```
{
  "mark": "area",
  "encoding": {
    "x": {"timeUnit": "yearmonth", "field": "date", "type": "temporal"},
    "y": {"aggregate": "sum", "field": "count","type": "quantitative", stack: "center"},
    "color": {"field":"series", "type":"nominal"}
  }
}
```

This will be equilvalent to using stack config as in the [VL1 streamgraph example](http://vega.github.io/vega-editor/?mode=vega-lite&spec=stacked_area_stream).

related to #1842

cc: @mattwchun

domoritz:
[...]

PR #1932, opened 2017-02-08
Make mark property can be either mark type or mark definition

kanitw:
merge #1940 first

Fix https://github.com/vega/vega-lite/issues/1882

### DIFF
--- a/src/compile/unit.ts
+++ b/src/compile/unit.ts
@@ -8,7 +8,7 @@ import {Encoding} from '../encoding';
 import * as vlEncoding from '../encoding'; // TODO: remove
 import {ChannelDef, FieldDef, FieldRefOption, field, normalize, isFieldDef, isValueDef} from '../fielddef';
 import {Legend} from '../legend';
-import {Mark, TEXT as TEXT_MARK, FILL_STROKE_CONFIG} from '../mark';
+import {Mark, MarkDef, TEXT as TEXT_MARK, FILL_STROKE_CONFIG, isMarkDef} from '../mark';
 import {Scale, ScaleConfig, hasDiscreteDomain} from '../scale';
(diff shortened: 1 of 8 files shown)


src/spec.ts
```diff
@@ -77,7 +77,7 @@ export interface GenericUnitSpec<M, E extends Encoding> extends BaseSpec {
   encoding: E;
 }
 
-export type UnitSpec = GenericUnitSpec<Mark, Encoding>;
+export type UnitSpec = GenericUnitSpec<Mark | MarkDef, Encoding>;
```
kanitw: @domoritz  Quick question -- should we rename Mark to MarkType.

I'm leaning toward no because we use it in too many places and I don't wanna make it too long.

domoritz: Nah, leave it.

willium:
looking forward to merge conflicts with #1906 😆

kanitw:
@willium  probably not gonna be that bad.  Things should be obvious to fix.

src/compile/unit.ts
```diff
@@ -44,7 +44,7 @@ export class UnitModel extends Model {
    */
   private _height: number;
 
-  private readonly _mark: Mark;
+  private readonly _markDef: MarkDef;
```
domoritz: Not sure why we do this weird underscore business. What fi you want to make a variable public in a subclass. 

Feels like an old hacky thing that we needed when we didn't have a type checker.

kanitw: Maybe, but if we wanna refactor, that sounds like a task for a separate PR.

domoritz: Okay. Lgtm then. https://github.com/vega/vega-lite/issues/1941

Wanna fix merge conflicts?

PR #1940, opened 2017-02-10
Separate CompositeMark from (primitive) Mark

kanitw:
- Provide API for registering new type of CompositeMark
- Remove AnyMark as "any mark" can be any string as long as the string is a primitive mark or a  registered composite mark
- Add isPrimitiveMark() instead of isCompositeMark()

### DIFF
--- a/src/compositemark.ts
+++ b/src/compositemark.ts
@@ -1,0 +1,64 @@
+import {Encoding} from './encoding';
+import {GenericUnitSpec, LayerSpec} from './spec';
+
+export const ERRORBAR: 'error-bar' = 'error-bar';
+export type ERRORBAR = typeof ERRORBAR;
+
+export type UnitNormalizer = (spec: GenericUnitSpec<any, any>)=> LayerSpec;
+
+/**
+ * Registry index for all composite mark's normalizer
+ */
+const normalizerRegistry: {[mark: string]: UnitNormalizer} = {};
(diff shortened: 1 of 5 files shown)


src/compositemark.ts
```diff
@@ -0,0 +1,64 @@
+import {Encoding} from './encoding';
```
domoritz: I'm not a fan or premature abstraction. In this case we have an abstraction for a single instance! Why?

kanitw: No -- it's not single instance.  

This way we can add experimental plugins for box plot / error bar / confidence internal that we need for CompassQL and Voyager without adding them to the main API. 

(In fact, in a future PR, I am leaning toward moving ERRORBAR out of here.)

kanitw: No -- it's not single instance.  This is a preparation for a recurring pattern.

This way we can add experimental plugins for box plot / error bar / confidence internal that we need for CompassQL and Voyager without adding them to the main API until we are happy with the syntax enough.

(In fact, in a future PR, I am leaning toward moving ERRORBAR out of here.)

domoritz: I see. So you plan to add it in compass soon.

kanitw: Yep, @mattwchun will start working on basic boxplot next week.

PR #1942, opened 2017-02-11
Refactor private properties

### LINKED ISSUE #1941: Remove _ prefix for private variables
We have a type checker.

domoritz:
Remove _ for privat properties. Use plural for dictionaries of scales, axes, and legends. Make more properties read only. Remove accessor functions for parent, description, padding, children, child, stack, and data.

Fixes #1941

### DIFF
--- a/src/compile/unit.ts
+++ b/src/compile/unit.ts
@@ -35,18 +35,23 @@ export class UnitModel extends Model {
    * If undefined (e.g., for ordinal scale), the width of the
    * visualization will be calculated dynamically.
    */
-  private _width: number;
+  public readonly width: number;
 
   /**
    * Fixed height for the unit visualization.
    * If undefined (e.g., for ordinal scale), the height of the
    * visualization will be calculated dynamically.
    */
-  private _height: number;
+  public readonly height: number;
 
-  private readonly _markDef: MarkDef;
-  private readonly _encoding: Encoding;
-  private readonly _stack: StackProperties;
+  private readonly markDef: MarkDef;
+  public readonly encoding: Encoding;
+  protected readonly scales: Dict<Scale> = {};
+  protected readonly axes: Dict<Axis> = {};
+  protected readonly legends: Dict<Legend> = {};
+  public readonly config: Config;
+  public readonly stack: StackProperties;
+  public children: Model[] = [];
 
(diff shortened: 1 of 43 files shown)


src/compile/unit.ts
```diff
@@ -59,35 +64,37 @@ export class UnitModel extends Model {
...
     }
 
-    this._axis = this._initAxis(encoding, config);
-    this._legend = this._initLegend(encoding, config);
+    this.axes = this.initAxes(encoding, config);
+    this.legends = this.initLegend(encoding, config);
 
     // width / height
-    this._initSize(mark, this._scale,
+    const {width, height} = this.initSize(mark, this.scales,
```
kanitw: `{width: this.width, height: this.height}`  more concise?

src/compile/unit.ts
```diff
@@ -320,51 +315,43 @@ export class UnitModel extends Model {
     return UNIT_CHANNELS;
   }
 
-  protected mapping() {
-    return this.encoding();
-  }
-
-  public stack(): StackProperties {
-    return this._stack;
+  protected getMapping() {
```
kanitw: I don't like unnecessary `get*` prefix for these function.  It makes code longer everywhere.

domoritz: Okay. I can rename them back. The problem was that I had not renamed scale to scales yet and so the I had a name conflict.

kanitw: Make sense then.

src/compile/unit.ts
domoritz: This creates unused variables and is arguable less predictable.

kanitw: Can you explain more?  I don't understand what you mean.

domoritz: `const {width: this.width, height: this.height}  = ...`

fails linting

domoritz:
Pushed fixes.

kanitw:
cc: @arvind @willium  -- you guys should be aware of this PR by @domoritz, as it will create some conflicts.

domoritz:
It should be as easy as removing `()` from wherever ts tells you that something is broken.

willium:
cool!

PR #1943, opened 2017-02-12
Project config

kanitw:
See commits for details


.npmignore
```diff
@@ -13,22 +13,20 @@
...
-bower.json
-_config.yml
-.travis.yml
-.jekyll-metadata
-.vscode
-.editorconfig
+yarn.lock
 vl.sublime-project
 **/*.ts
 !**/*.d.ts
```
domoritz: what about js files? Aren't they ignored as well?

domoritz:
About tslib. I actually tried this a while ago and the js files got larger, not smaller. Can you make sure that this is not the case?

.npmignore
kanitw: Right now js files are included in release. 

I think in another PR, I want to move all build files (including vega-lite.js|min.js and all compiled js and d.ts to `build/` while keeping the src files in the release).

kanitw:
> About tslib. I actually tried this a while ago and the js files got larger, not smaller. Can you make sure that this is not the case?

Interesting.  I guess it's partly because we rarely use spread operator in the past. 
However, if we use it more (or even replace all extend with it #1953), we will reach a tipping point pretty soon?

.npmignore
kanitw: Just file an issue  #1955 for that. 

The reason I wanna separate this from that PR is that this PR makes the current approach correct.  

Meanwhile, #1955 introduces a different approach.  If we merge this first and have that PR, we can always revert that PR if we find that that approach is actually not better.

PR #1944, opened 2017-02-12
Audit Axis Config

### LINKED ISSUE #1845: Audit Vega-Lite Axis/Legend Config
# Distinguish Between Config and Properties of Axis/Legend

## Axis 

- [x] Axis 

Right now `Axis` extends `AxisConfig`.  However some property such as `axisColor` only make sense as a theme (you won't make one axis red and another black.   Thus, we should distinguish the two and properties such as `axisColor` should be `AxisConfig`.

- [x] If there are shared properties between them, then we want  `Axis` and `AxisConfig` extend common `AxisBase`.

- [x] Modify the axis compile code to only deal with what's in `Axis`  (`_initAxis` in `UnitModel` and `FacetModel`)

- [x] We have a code to copy all config values to actual axis object. However, we should not copy *all* config properties anymore since `Axis` and `AxisConfig` are not the same.
   - [x] We need to think what are the few that we need to copy and how
       - [x] labelMaxLength

- [x] update all docs

- [x] make sure we don't left any property undecided

## Legend

- [x] Do it similar to axis

# 2. Make sure our configs match with Vega

I think Vega-Lite config for each of scale/axis/legend can either be (1) exactly the same as Vega's or (2) has a few more properties.   (It shouldn't have fewer properties.)

- [x] If there are properties missing from VL -- add them.

- [x] If there are properties in VL that are not in VG, check if they are renamed in VG.

- [x] Make new interface `VgAxisConfig` and make `AxisConfig implements VgAxisConfig` (or define `type AxisConfig = VgAxisConfig` if there is no additional properties)

- [x] Update examples in examples/spec accordingly

YuhanLu:
Update Axis Properties and Config.  (Part of #1845)

  - Rename `axisWidth` to `domainWidth`, `axisColor` to `domainColor`, `tickLabelColor` to `labelColor`, `tickLabelFont` to `labelFont`, `tickLabelFontSize` to `labelFontSize`, `tickPadding` to `labelPadding`, `titleOffset` to `titlePadding`
  - Separate Axis from AxisConfig. Make these properties only available as `Axis` properties:  `domain `, `grid `, `labels `, `labelMaxLength `, `labelPadding `, `maxExtent`, `minExtent`, `offset `, `position `, `shortTimeLabels `, `ticks `, `tickCount `, `tickSize`, `titleMaxLength`, `titlePadding`, `zindex`
  - Remove `labelAlign`, `labelBaseline`, `subdivide`, `tickSizeMajor`, `tickSizeMinor`, `tickSizeEnd` from Axis Config
  
  - Remove related code for Vega Axis config as Vega parser will handle these configs.

### DIFF
--- a/examples/specs/stacked_area_stream.vl.json
+++ b/examples/specs/stacked_area_stream.vl.json
@@ -18,5 +18,5 @@
     "cell": {"width": 300, "height": 200},
     "axis": {"domainWidth": 0},
     "stack": "center"
-    }
+  }
 }
--- a/examples/specs/stacked_area_stream.vl.json
+++ b/examples/specs/stacked_area_stream.vl.json
@@ -6,7 +6,7 @@
     "x": {
       "timeUnit": "yearmonth", "field": "date", "type": "temporal",
       "scale": {"nice": "month"},
-      "axis": {"axisWidth": 0, "format": "%Y", "labelAngle": 0, "tickSize": 0}
+      "axis": {"format": "%Y", "labelAngle": 0, "tickSize": 0}
     },
     "y": {
       "aggregate": "sum", "field": "count","type": "quantitative",
@@ -14,5 +14,9 @@
     },
     "color": {"field":"series", "type":"nominal", "scale":{"scheme": "category20b"}}
   },
-  "config": {"cell": {"width": 300, "height": 200}, "stack": "center"}
+  "config": {
+    "cell": {"width": 300, "height": 200},
+    "axis": {"domainWidth": 0},
+    "stack": "center"
+    }
 }
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -4,17 +4,17 @@ import {VgAxisEncode} from './vega.schema';
 export type AxisOrient = 'top' | 'right' | 'left' | 'bottom';
 
 export interface AxisConfig extends AxisBase {
-  // ---------- General ----------
+  // ---------- Axis ----------
   /**
-   * Width of the axis line
+   * Width of the domain line
    */
   domainWidth?: number;
 
-  // ---------- Axis ----------
   /**
-   * Color of axis line.
+   * Color of axis domain line.
    */
   domainColor?: string;
+
   // ---------- Grid ----------
   /**
    * Color of gridlines.
@@ -41,7 +41,6 @@ export interface AxisConfig extends AxisBase {
   gridWidth?: number;
 
   // ---------- Ticks ----------
-
   /**
    * The color of the axis's tick.
    */
@@ -62,6 +61,7 @@ export interface AxisConfig extends AxisBase {
    * @minimum 0
    */
   labelFontSize?: number;
+
   /**
    * The width, in pixels, of ticks.
    * @minimum 0
@@ -89,11 +89,6 @@ export interface AxisConfig extends AxisBase {
    * Weight of the title.
    */
   titleFontWeight?: string | number;
-
-  /**
-   * A title offset value for the axis.
-   */
-  titleOffset?: number;
 }
 
 // TODO: add comment for properties that we rely on Vega's default to produce
@@ -105,10 +100,6 @@ export const defaultAxisConfig: AxisConfig = {
 
 export const defaultFacetAxisConfig: AxisConfig = {
   domainWidth: 0,
-  // TODO: remove these
-  domain: false,
-  grid: false,
-  ticks: false
 };
 
 export interface Axis extends AxisBase {
@@ -116,31 +107,39 @@ export interface Axis extends AxisBase {
    * The padding, in pixels, between axis and text labels.
    */
   labelPadding?: number;
+
   /**
    * The formatting pattern for axis labels.
    */
   format?: string; // default value determined by config.format anyway
+
   /**
    * The orientation of the axis. One of top, bottom, left or right. The orientation can be used to further specialize the axis type (e.g., a y axis oriented for the right edge of the chart).
    */
   orient?: AxisOrient;
+
   /**
    * The offset, in pixels, by which to displace the axis from the edge of the enclosing group or data rectangle.
    */
   offset?: number;
+
   // FIXME: Add Description
   position?: number;
+
   /**
    * A desired number of ticks, for axes visualizing quantitative scales. The resulting number may be different so that values are "nice" (multiples of 2, 5, 10) and lie within the underlying scale's range.
    * @minimum 0
    * @TJS-type integer
    */
   tickCount?: number;
+
   /**
    * A title for the axis. Shows field name and its function by default.
    */
   title?: string;
+
   values?: number[] | DateTime[];
+
   /**
    * A non-positive integer indicating z-index of the axis.
    * If zindex is 0, axes should be drawn behind all chart elements.
@@ -149,69 +148,81 @@ export interface Axis extends AxisBase {
    * @minimum 0
    */
   zindex?: number;
+
   /**
    * Optional mark definitions for custom axis encoding.
    */
   encode?: VgAxisEncode;
 }
 
-export interface AxisBase {
-  /**
-   * Whether to include the axis domain line.
-   */
-  domain?: boolean;
-  /**
-   * A flag indicate if gridlines should be created in addition to ticks. If `grid` is unspecified, the default value is `true` for ROW and COL. For X and Y, the default value is `true` for quantitative and time fields and `false` otherwise.
-   */
-  grid?: boolean;
-  /**
-   * Enable or disable labels.
-   */
-  labels?: boolean;
-  /**
-   * The rotation angle of the axis labels.
-   * @minimum 0
-   * @maximum 360
-   */
-  labelAngle?: number;
+export interface AxisBase extends VgAxisBase, VlAxisBase {}
+
+export interface VlAxisBase {
   /**
    * Truncate labels that are too long.
    * @minimum 1
    * @TJS-type integer
    */
   labelMaxLength?: number;
+}
+
+export interface VgAxisBase {
+  /**
+   * Whether to include the axis domain line.
+   */
+  domain?: boolean;
+
+  /**
+   * A flag indicate if gridlines should be created in addition to ticks. If `grid` is unspecified, the default value is `true` for ROW and COL. For X and Y, the default value is `true` for quantitative and time fields and `false` otherwise.
+   */
+  grid?: boolean;
+
+  /**
+   * Enable or disable labels.
+   */
+  labels?: boolean;
+
+  /**
+   * The rotation angle of the axis labels.
+   * @minimum 0
+   * @maximum 360
+   */
+  labelAngle?: number;  // FIXME: not sure if this should be a theme
+
   /**
    * Whether the axis should include ticks.
    */
   ticks?: boolean;
+
   /**
    * The size, in pixels, of major, minor and end ticks.
    * @minimum 0
    */
   tickSize?: number;
+
   /**
    * Max length for axis title if the title is automatically generated from the field's description. By default, this is automatically based on cell size and characterWidth property.
    * @minimum 0
    * @TJS-type integer
    */
   titleMaxLength?: number;
+
   /**
    * The padding, in pixels, between title and axis.
    */
   titlePadding?: number;
+
   /**
    * Minimum extent, which determines the offset between axis ticks and labels.
    */
   minExtent?: number;
+
   /**
    * Maximum extent, which determines the offset between axis ticks and labels.
    */
   maxExtent?: number;
-  /**
-   * Whether month and day names should be abbreviated.
-   */
-  shortTimeLabels?: boolean;
 }
+
 
 export const AXIS_PROPERTIES:(keyof Axis)[] = [
   // a) properties with special rules (so it has axis[property] methods) -- call rule functions
@@ -219,3 +230,5 @@ export const AXIS_PROPERTIES:(keyof Axis)[] = [
   // b) properties without rules, only produce default values in the schema, or explicit value if specified
     'labelPadding', 'maxExtent', 'minExtent', 'offset', 'position', 'tickSize', 'titlePadding'
 ];
+
+export const AXIS_BASE_PROPERTIES:(keyof VlAxisBase)[] = ['labelMaxLength'];
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -3,42 +3,19 @@ import {VgAxisEncode} from './vega.schema';
 
 export type AxisOrient = 'top' | 'right' | 'left' | 'bottom';
 
-export interface AxisConfig {
+export interface AxisConfig extends AxisBase {
   // ---------- General ----------
   /**
    * Width of the axis line
    */
-  axisWidth?: number;
-  /**
-   * A non-positive integer indicating z-index of the axis.
-   * If zindex is 0, axes should be drawn behind all chart elements.
-   * To put them in front, use zindex = 1.
-   * @TJS-type integer
-   * @minimum 0
-   */
-  zindex?: number;
-  /**
-   * The offset, in pixels, by which to displace the axis from the edge of the enclosing group or data rectangle.
-   */
-  offset?: number;
+  domainWidth?: number;
 
   // ---------- Axis ----------
   /**
    * Color of axis line.
    */
-  axisColor?: string;
-
-  /**
-   * Whether to include the axis domain line.
-   */
-  domain?: boolean;
-
+  domainColor?: string;
   // ---------- Grid ----------
-  /**
-   * A flag indicate if gridlines should be created in addition to ticks. If `grid` is unspecified, the default value is `true` for ROW and COL. For X and Y, the default value is `true` for quantitative and time fields and `false` otherwise.
-   */
-  grid?: boolean;
-
   /**
    * Color of gridlines.
    */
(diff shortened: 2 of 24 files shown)


src/compile/facet.ts
```diff
@@ -108,14 +108,14 @@ export class FacetModel extends Model {
...
-  private _initAxis(facet: Facet, config: Config, child: Model): Dict<Axis> {
+  private _initAxis(facet: Facet, child: Model): Dict<Axis> {
     const model = this;
     return [ROW, COLUMN].reduce(function(_axis, channel) {
       if (facet[channel]) {
         const axisSpec = facet[channel].axis;
         if (axisSpec !== false) {
           const modelAxis = _axis[channel] = extend({},
-            config.facet.axis,
+            defaultFacetAxisConfig,
```
kanitw: Similar to [`_initAxis` for `UnitModel`](https://github.com/vega/vega-lite/pull/1944/files#r100684723), I don't think this is correct. 

However, just simply removing it will make things wrong too.  
Because facet's axis should *always* hide `domain`, `grid` and `ticks` but currently we do so by relying on `defaultFacetAxisConfig`, which is still configurable.

We should remove these 3 properties from `defaultFacetAxisConfig` (Notice [TODO](https://github.com/vega/vega-lite/pull/1944/commits/aca82ca8c8ba034ee4b1d03712e66b524c1eed9a#diff-a8e48aa61f457230707f1f7e707e76e5R230) in their declaration?) and instead add rules for  `domain`, `grid` and `ticks` to always return `false` for facet (`row` and `column` channel)
(For easy review, do this removal + add test in a separate commit)

https://github.com/vega/vega-lite/pull/1944/files#diff-a8e48aa61f457230707f1f7e707e76e5R109

src/axis.ts
```diff
@@ -231,20 +104,18 @@ export const defaultAxisConfig: AxisConfig = {
 };
 
 export const defaultFacetAxisConfig: AxisConfig = {
-  axisWidth: 0,
+  domainWidth: 0,
```
YuhanLu: Could you please explain what does faceted plot mean?

kanitw: Plot with row/column.  For more information, please read VL docs.

examples/specs/stacked_area_stream.vl.json
```diff
@@ -6,13 +6,17 @@
...
       "axis": null
     },
     "color": {"field":"series", "type":"nominal", "scale":{"scheme": "category20b"}}
   },
-  "config": {"cell": {"width": 300, "height": 200}, "stack": "center"}
+  "config": {
+    "cell": {"width": 300, "height": 200},
+    "axis": {"domainWidth": 0},
+    "stack": "center"
+    }
```
YuhanLu: fixed

src/axis.ts
```diff
@@ -231,20 +104,18 @@ export const defaultAxisConfig: AxisConfig = {
...
-export interface Axis extends AxisConfig {
+export interface Axis extends AxisBase {
   /**
-   * The rotation angle of the axis labels.
-   * @minimum 0
-   * @maximum 360
+   * The padding, in pixels, between axis and text labels.
    */
-  labelAngle?: number;
+  labelPadding?: number;
```
YuhanLu: fixed

src/axis.ts
```diff
@@ -254,16 +125,97 @@ export interface Axis extends AxisConfig {
...
+   * @minimum 0
+   */
+  zindex?: number;
+  /**
+   * Optional mark definitions for custom axis encoding.
+   */
+  encode?: VgAxisEncode;
+}
+
+export interface AxisBase {
```
YuhanLu: fixed

src/axis.ts
```diff
@@ -254,16 +125,97 @@ export interface Axis extends AxisConfig {
...
+  /**
+   * Enable or disable labels.
+   */
+  labels?: boolean;
+  /**
+   * The rotation angle of the axis labels.
+   * @minimum 0
+   * @maximum 360
+   */
+  labelAngle?: number;
```
YuhanLu: fixed

src/compile/axis/parse.ts
```diff
@@ -98,7 +98,12 @@ function parseAxis(channel: Channel, model: Model, isGridAxis: boolean): VgAxis
     }
     // TODO(@yuhanlu): instead of calling encode[part], break this line based on part type
     // as different require different parameters.
-    const value = encode[part](model, channel, encodeSpec.labels || {}, vgAxis);
+    let value;
+    if (part === 'labels') {
+        value = encode[part](model, channel, encodeSpec.labels || {}, vgAxis);
```
YuhanLu: fixed

src/compile/unit.ts
```diff
@@ -226,8 +226,8 @@ export class UnitModel extends Model {
         // We no longer support false in the schema, but we keep false here for backward compatability.
         if (axisSpec !== null && axisSpec !== false) {
           _axis[channel] = extend({},
-            config.axis,
-            axisSpec === true ? {} : axisSpec ||  {}
+          defaultAxisConfig,
```
YuhanLu: fixed

src/compile/facet.ts
YuhanLu: Right now I have a rule called `facet()` for both `domain` and `ticks`, and I see there's a little different in `rules.grid()`, do you expect me to refactor these 2 functions into one?

src/compile/axis/rules.ts
```diff
@@ -123,3 +124,10 @@ export function zindex(specifiedAxis: Axis, isGridAxis: boolean) {
   }
   return 1; // otherwise return undefined and use Vega's default.
 };
+
+export function facet(property: keyof VgAxis, specifiedAxis: Axis, isGridAxis: boolean, channel: Channel) {
```
kanitw: `facet` is a pretty confusing name.   This is a common rules for `domain` and `ticks`, which deal with their rules no matter you have facet or not.   

If you want a single method, a better name would be `domainAndTicks`.
For clean API, I would not export `domainAndTicks` and instead 
```
export const domain = domainAndTicks;
export const ticks = domainAndTicks;
```

and just call `domain` and `ticks` from `axis/parse`. 

This way if we wanna customize domain / ticks in the future, we only need to change them inside this file.  

(`axis/parse` shouldn't need to know that rules for these two properties are the same.)

src/compile/facet.ts
kanitw: Nope -- mixing them up only make things more confusing. See my other comment. https://github.com/vega/vega-lite/pull/1944#discussion_r101076930 :)

src/compile/unit.ts
```diff
@@ -234,10 +234,17 @@ export class UnitModel extends Model {
...
-          );
+          let vlAxisBase: VlAxisBase = {};
+          AXIS_BASE_PROPERTIES.forEach(function(property) {
+            if (encoding !== undefined && encoding[property] !== undefined) {
+              vlAxisBase[property] = encoding[property];
+            }
+          });
+          _axis[channel] = {
+            labelMaxLength: config.axis.labelMaxLength,
+            vlAxisBase,
```
kanitw: 1)  It should be `...vlAxisBase`

2) You didn't write unit test for this part.

src/compile/facet.ts
kanitw: In any case, now you have those rules, why don't you fix your code here (and add test).

src/axis.ts
```diff
@@ -1,45 +1,22 @@
...
   /**
    * Color of axis line.
    */
-  axisColor?: string;
-
-  /**
-   * Whether to include the axis domain line.
-   */
-  domain?: boolean;
+  domainColor?: string;
```
kanitw: Color of axis domain line.

src/axis.ts
```diff
@@ -1,45 +1,22 @@
...
-   * To put them in front, use zindex = 1.
-   * @TJS-type integer
-   * @minimum 0
-   */
-  zindex?: number;
-  /**
-   * The offset, in pixels, by which to displace the axis from the edge of the enclosing group or data rectangle.
-   */
-  offset?: number;
+  domainWidth?: number;
```
kanitw: "Width of axis domain line."

src/vega.schema.ts
```diff
@@ -180,3 +180,65 @@ export interface VgImputeTransform {
...
+
+  /**
+   * Maximum extent, which determines the offset between axis ticks and labels.
+   */
+  maxExtent?: number;
+
+  /**
+   * Whether month and day names should be abbreviated.
+   */
+  shortTimeLabels?: boolean;
```
kanitw: 1.  This is definitely not a VG property

2. You remove this in legend PR, but why keep it here? It's not consistent. 

Let's just remove it?

src/axis.ts
YuhanLu: fixed

src/axis.ts
YuhanLu: fixed

src/compile/axis/rules.ts
YuhanLu: fixed

src/compile/facet.ts
YuhanLu: fixed

src/compile/unit.ts
YuhanLu: fixed

src/vega.schema.ts
YuhanLu: fixed and moved `VgAxisBase` to `axis.ts`

src/compile/unit.ts
```diff
@@ -234,10 +234,17 @@ export class UnitModel extends Model {
 
         // We no longer support false in the schema, but we keep false here for backward compatability.
         if (axisSpec !== null && axisSpec !== false) {
-          _axis[channel] = extend({},
-            config.axis,
-            axisSpec === true ? {} : axisSpec ||  {}
-          );
+          let vlAxisBase: VlAxisBase = {};
+          AXIS_BASE_PROPERTIES.forEach(function(property) {
+            if (encoding !== undefined && encoding[property] !== undefined) {
```
kanitw: I doubt if can `encoding` be `undefined` here?

kanitw: Also, why `encoding` not `config`?

src/compile/facet.ts
```diff
@@ -121,16 +121,16 @@ export class FacetModel extends Model {
...
     return [ROW, COLUMN].reduce(function(_axis, channel) {
       if (facet[channel]) {
         const axisSpec = facet[channel].axis;
         if (axisSpec !== false) {
-          const modelAxis = _axis[channel] = extend({},
-            config.facet.axis,
-            axisSpec === true ? {} : axisSpec || {}
-          );
+          const modelAxis = _axis[channel] = {
+            ...defaultFacetAxisConfig,
```
kanitw: This is still wrong.  Just like unit spec (which is [still wrong too](https://github.com/vega/vega-lite/pull/1944/files#r101435512)), you shouldn't copy the whole config.

src/compile/unit.ts
kanitw: You should also add comment explaining why you only copy these properties too.  

```
// Only copy VlAxisBase properties from config.  (VgAxis config are already handled by Vega so we don't have to config)
```

src/compile/facet.ts
kanitw: Of course add similar comment

src/axis.ts
```diff
@@ -63,207 +40,195 @@ export interface AxisConfig {
...
 
   /**
-   * Whether the axis should include ticks.
+   * Weight of the title.
    */
-  ticks?: boolean;
+  titleFontWeight?: string | number;
+}
+
+// TODO: add comment for properties that we rely on Vega's default to produce
```
kanitw: Not sure what you mean here.

src/compile/facet.ts
YuhanLu: fixed

src/axis.ts
YuhanLu: deleted

src/compile/unit.ts
YuhanLu: fixed

src/axis.ts
```diff
@@ -155,9 +66,9 @@ export interface Axis extends AxisBase {
   encode?: VgAxisEncode;
 }
 
-export interface AxisBase extends VgAxisBase, VlAxisBase {}
+export interface AxisBase extends VgAxisBase, VlOnlyAxisConfig {}
```
kanitw: Does this every get used anywhere?  I thought we agree to remove it?

src/compile/axis/encode.ts
```diff
@@ -23,7 +23,7 @@ export function labels(model: Model, channel: Channel, labelsSpec: any, def: VgA
   } else if (fieldDef.type === TEMPORAL) {
     labelsSpec = extend({
       text: {
-        signal: timeFormatExpression('datum.value', fieldDef.timeUnit, axis.format, axis.shortTimeLabels, config)
+        signal: timeFormatExpression('datum.value', fieldDef.timeUnit, axis.format, config.axis.shortTimeLabels, config)
```
kanitw: This line looks wrong.  Why are you passing both `config` and `config.axis.shortTimeLabels`?  

It seems like you can get rid of one parameter from this method.

src/axis.ts
```diff
@@ -166,69 +77,11 @@ export interface VlAxisBase {
...
-
 export const AXIS_PROPERTIES:(keyof Axis)[] = [
   // a) properties with special rules (so it has axis[property] methods) -- call rule functions
   'domain', 'format', 'labels', 'grid', 'orient', 'ticks', 'tickSize', 'tickCount',  'title', 'values', 'zindex',
   // b) properties without rules, only produce default values in the schema, or explicit value if specified
     'labelPadding', 'maxExtent', 'minExtent', 'offset', 'position', 'tickSize', 'titlePadding'
 ];
 
-export const AXIS_BASE_PROPERTIES:(keyof VlAxisBase)[] = ['labelMaxLength'];
+export const VL_AXIS_BASE_PROPERTIES:(keyof VlOnlyAxisConfig)[] = ['labelMaxLength'];
```
kanitw: VL_ONLY_AXIS_PROPERTIES?

We no longer have VlAxisBase types, so you should rename all relevant variable names.

src/compile/facet.ts
```diff
@@ -121,14 +121,21 @@ export class FacetModel extends Model {
...
   }
 
-  private initAxis(facet: Facet, child: Model): Dict<Axis> {
+  private initAxis(facet: Facet, config: Config, child: Model): Dict<Axis> {
     const model = this;
     return [ROW, COLUMN].reduce(function(_axis, channel) {
       if (facet[channel]) {
         const axisSpec = facet[channel].axis;
         if (axisSpec !== false) {
+          let vlAxisBase: VlOnlyAxisConfig = {};
```
kanitw: We no longer have VlAxisBase types, so you should rename all relevant variable names.

src/compile/unit.ts
```diff
@@ -234,14 +234,13 @@ export class UnitModel extends Model {
 
         // We no longer support false in the schema, but we keep false here for backward compatability.
         if (axisSpec !== null && axisSpec !== false) {
-          let vlAxisBase: VlAxisBase = {};
-          AXIS_BASE_PROPERTIES.forEach(function(property) {
-            if (encoding !== undefined && encoding[property] !== undefined) {
-              vlAxisBase[property] = encoding[property];
+          let vlAxisBase: VlOnlyAxisConfig = {};
```
kanitw: We no longer have VlAxisBase types, so you should rename all relevant variable names.

src/vega.schema.ts
```diff
@@ -180,3 +180,154 @@ export interface VgImputeTransform {
   method?: 'value' | 'median' | 'max' | 'min' | 'mean';
   value?: any;
 }
+
+export interface VgAxisBase {
```
kanitw: Add description -- something like:

/**
 * Base object for Vega's Axis and Axis Config.  
 * All of these properties are both properties of Vega's Axis and Axis Config.  
 */

test/compile/facet.test.ts
```diff
@@ -379,7 +379,7 @@ describe('compile/facet', () => {
   });
 
   describe('initAxis', () => {
-    it('should should have defaultFacetAxisConfig that domainWidth = 0', () => {
+    it('should map properties of vl config in config.facet.axis', () => {
```
kanitw: include properties of VlOnlyAxisConfig in config.facet.axis

src/compile/facet.ts
```diff
@@ -127,15 +127,15 @@ export class FacetModel extends Model {
       if (facet[channel]) {
         const axisSpec = facet[channel].axis;
         if (axisSpec !== false) {
-          let vlAxisBase: VlOnlyAxisConfig = {};
-          VL_AXIS_BASE_PROPERTIES.forEach(function(property) {
+          let vlAxisProperties: VlOnlyAxisConfig = {};
```
kanitw: vlOnlyAxisProperties

src/compile/unit.ts
```diff
@@ -234,14 +234,14 @@ export class UnitModel extends Model {
 
         // We no longer support false in the schema, but we keep false here for backward compatability.
         if (axisSpec !== null && axisSpec !== false) {
-          let vlAxisBase: VlOnlyAxisConfig = {};
-          VL_AXIS_BASE_PROPERTIES.forEach(function(property) {
+          let vlAxisProperties: VlOnlyAxisConfig = {};
```
kanitw: vlOnlyAxisProperties

PR #1947, opened 2017-02-14
Legend Properties/Config refactor

### LINKED ISSUE #1845: Audit Vega-Lite Axis/Legend Config

YuhanLu:
Update Legend Properties and Config. (Part of #1845)
- Add `symbolType`
- Remove `margin`, `symbolShape` from Legend Config
- Separate Legend from LegendConfig. Make these properties only available as Legend properties: `entryPadding`, `orient`, `offset`, `padding`, `tickCount`, `zindex`
- Remove related code for Vega Legend configs

### DIFF
--- a/src/compile/legend/encode.ts
+++ b/src/compile/legend/encode.ts
@@ -86,10 +86,7 @@ export function symbols(fieldDef: FieldDef, symbolsSpec: any, model: UnitModel, 
   const shapeDef = model.encoding.shape;
   if (channel !== SHAPE) {
     if (isValueDef(shapeDef)) {
-      value = {value: shapeDef.value};
-    }
-    if (value !== undefined) {
-      symbols.shape = value;
+      symbols.shape = {value: shapeDef.value};
     }
   }
 
--- a/src/compile/legend/encode.ts
+++ b/src/compile/legend/encode.ts
@@ -14,7 +14,6 @@ import {UnitModel} from '../unit';
 export function symbols(fieldDef: FieldDef, symbolsSpec: any, model: UnitModel, channel: Channel) {
   let symbols:any = {};
   const mark = model.mark();
-  const legend = model.legend(channel);
 
   switch (mark) {
     case BAR:
@@ -75,9 +74,7 @@ export function symbols(fieldDef: FieldDef, symbolsSpec: any, model: UnitModel, 
       {value: cfg.mark.color};
   }
 
-  if (legend.symbolColor !== undefined) {
-    symbols.fill = {value: legend.symbolColor};
-  } else if (symbols.fill === undefined) {
+  if (symbols.fill === undefined) {
     // fall back to mark config colors for legend fill
     if (cfg.mark.fill !== undefined) {
       symbols.fill = {value: cfg.mark.fill};
@@ -87,16 +84,8 @@ export function symbols(fieldDef: FieldDef, symbolsSpec: any, model: UnitModel, 
   }
 
   if (channel !== SHAPE) {
-    if (legend.symbolShape !== undefined) {
-      symbols.shape = {value: legend.symbolShape};
-    } else if (cfg.point.shape !== undefined) {
+    if (cfg.point.shape !== undefined) {
       symbols.shape = {value: cfg.point.shape};
-    }
-  }
-
-  if (channel !== SIZE) {
-    if (legend.symbolSize !== undefined) {
-      symbols.size = {value: legend.symbolSize};
     }
   }
 
@@ -118,10 +107,6 @@ export function symbols(fieldDef: FieldDef, symbolsSpec: any, model: UnitModel, 
       default:
         throw Error(`Legend for channel ${channel} not implemented`);
     }
-  }
-
-  if (legend.symbolStrokeWidth !== undefined) {
-    symbols.strokeWidth = {value: legend.symbolStrokeWidth};
   }
 
   symbols = extend(symbols, symbolsSpec || {});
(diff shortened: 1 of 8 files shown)


src/compile/legend/encode.ts
```diff
@@ -75,9 +74,7 @@ export function symbols(fieldDef: FieldDef, symbolsSpec: any, model: UnitModel,
       {value: cfg.mark.color};
   }
 
-  if (legend.symbolColor !== undefined) {
-    symbols.fill = {value: legend.symbolColor};
-  } else if (symbols.fill === undefined) {
+  if (symbols.fill === undefined) {
```
kanitw: 1) Do we even need to set symbols.fill to `cfg.mark.fill`?  I wonder if Vega is doing this for us too, please test.

2) I believe the missing case here is when users set `encoding.color.value` -- the code here should be consistent with `applyColor(AndOpacity)` method

src/compile/legend/encode.ts
```diff
@@ -87,19 +84,11 @@ export function symbols(fieldDef: FieldDef, symbolsSpec: any, model: UnitModel,
   }
 
   if (channel !== SHAPE) {
-    if (legend.symbolShape !== undefined) {
-      symbols.shape = {value: legend.symbolShape};
-    } else if (cfg.point.shape !== undefined) {
+    if (cfg.point.shape !== undefined) {
```
kanitw: 1) Do we need to set symbols.shape to `config.point.shape`?  (I guess we do, but please test.)

2) I think we have to handle the case when users specify `encoding.shape.value`.

src/compile/legend/encode.ts
```diff
@@ -146,58 +131,13 @@ export function labels(fieldDef: FieldDef, labelsSpec: any, model: UnitModel, ch
   } else if (fieldDef.type === TEMPORAL) {
     labelsSpec = extend({
       text: {
-        signal: timeFormatExpression('datum.value', fieldDef.timeUnit, legend.format, legend.shortTimeLabels, config)
+        // FIXME: shortTimeLabels set as true by default
```
kanitw: Why set to true by default?  Why not just always read from config?  (Modify `timeFormatExpression`?)

src/compile/unit.ts
```diff
@@ -244,16 +244,20 @@ export class UnitModel extends Model {
...
   }
 
-  private initLegend(encoding: Encoding, config: Config): Dict<Legend> {
+  private initLegend(encoding: Encoding): Dict<Legend> {
     return NONSPATIAL_SCALE_CHANNELS.reduce(function(_legend, channel) {
       const channelDef = encoding[channel];
       if (isFieldDef(channelDef)) {
         const legendSpec = channelDef.legend;
         // We no longer support false in the schema, but we keep false here for backward compatability.
+        let vlLegendBase: VlLegendBase = {};
```
kanitw: 1) this logic is definitely wrong.  Why copy `encoding.<legend-base-property>`, which doesn't exists?

2) Bad code tab alignment.

src/legend.ts
```diff
@@ -159,10 +147,30 @@ export interface Legend extends LegendConfig {
...
+export interface LegendBase extends VgLegendBase, VlLegendBase {}
+
 export const defaultLegendConfig: LegendConfig = {
   orient: undefined, // implicitly "right"
 };
 
-export const LEGEND_PROPERTIES:(keyof Legend)[] = ['entryPadding', 'format', 'offset', 'orient', 'tickCount', 'title', 'titlePadding', 'type', 'values' ,'zindex'];
+export const LEGEND_PROPERTIES:(keyof Legend)[] = ['entryPadding', 'format', 'offset', 'orient', 'tickCount', 'title', 'type', 'values' ,'zindex'];
+
+export const LEGEND_BASE_PROPERTIES:(keyof VlLegendBase)[] = ['shortTimeLabels'];
```
kanitw: `VL_LEGEND_BASE_PROPERTIES`

src/vega.schema.ts
```diff
@@ -180,3 +180,13 @@ export interface VgImputeTransform {
   method?: 'value' | 'median' | 'max' | 'min' | 'mean';
   value?: any;
 }
+
+export interface VgLegendBase {
+  entryPadding?: number;
```
kanitw: 1) Move this to legend.ts?  (The concept of LegendBase doesn't exist in Vega.)

2) Missing description comment for all properties.  (When you review your PR, you should inspect the schema too. If you did that, you would have notice missing "description" from the schema file.)

src/legend.ts
```diff
@@ -1,115 +1,88 @@
...
    * The color of the legend symbol,
    */
   symbolColor?: string;
+
   /**
    * The shape of the legend symbol, can be the 'circle', 'square', 'cross', 'diamond',
    * 'triangle-up', 'triangle-down', or else a custom SVG path string.
    */
-  symbolShape?: string;
+  symbolType?: string;
```
kanitw: `symbolShape` isn't renamed to `symbolType`.  `symbolType` is simply a new property.  Please

- [ ] correct description
- [ ] update the PR description / commit message  (rename => remove + add)

kanitw:
@YuhanLu  -- since I will squash and merge at the end anyway.  Maybe you wanna make your fixes (to address to my comment new commit(s) so that it's easier for me to notice changes.

src/legend.ts
kanitw: (outdated -- VL_ONLY_LEGEND .... something something)

src/compile/legend/encode.ts
YuhanLu: I tested with and without `symbols.update.fill` been specified, and it does make a difference.

src/compile/legend/encode.ts
YuhanLu: 1. No, vega will use circle by default if we don't specify it.

kanitw: I guess you mean "yes we do?"

YuhanLu: I mean it's not necessary to set it lol

src/compile/legend/encode.ts
YuhanLu: fixed

src/compile/unit.ts
YuhanLu: fixed

src/legend.ts
YuhanLu: fixed

src/legend.ts
YuhanLu: fixed

src/vega.schema.ts
YuhanLu: fixed

kanitw:
@YuhanLu Please add tests for missing lines. (Click details for codecov/patch and see diff tab)

PR #1948, opened 2017-02-14
split function applyColorAndOpacity()

### LINKED ISSUE #1869: Make `applyColorAndOpacity` just `applyColor()`.
`applyColorAndOpacity()` in src/compile/mark/common shouldn't couple opacity in the same method.

Instead, for opacity we can just call `e.opacity = ref.midPoint(...)` for all marks.

YuhanLu:
Fix #1869 
Slpit function applyColorAndOpacity() to applyColor() and one line to set opacity.


kanitw:
## Problems

Please be more careful:

You miss `e.opacity = `  in many lines -- they are not gonna work if you do not assign them.  (Obviously, you didn't test this -- maybe add more assertion to an existing test case for checking opacity?) 

Even if you don't forget `e.opacity`, this line also have two problems:

```
e.opacity = ref.midPoint('opacity', model.encoding.opacity, model.scaleName('opacity'), model.scale('opacity'), config.mark.opacity && {value: config.mark.opacity});
```

- `e.opacity` can become `undefined` but if it's undefined, it should not be assigned.
- `config.mark.opacity` is already handled by mark so we shouldn't pass it to midpoint.

- [ ] Thus do this instead:
```
const opacity = ref.midPoint('opacity', model.encoding.opacity, model.scaleName('opacity'), model.scale('opacity'));
if (opacity !== undefined) {
  e.opacity = opacity;
}
```
(Yes, we have to do the same for other mark properties, but let's deal with opacity for now.)

- [ ] Instead of only removing test `applyColorAndOpacity()`, please also add test for applyColor()`

## Style Comment

- [ ]  Make `e.opacity = ...` come before `applyColor` for consistency.

kanitw:
@YuhanLu I'm gonna squash and edit PR description a bit. 
> Fix #1869
> Slpit function applyColorAndOpacity() to applyColor() and one line to set opacity.

Please use this as an example to improve future description.

```
Refactor applyColorAndOpacity() to applyColor() 

- Decouple opacity from applyColorAndOpacity() and use ref.midPoint to set instead. 
- Add tests for `applyColor`

Fix #1869
```

PR #1956, opened 2017-02-17
Revert "Remove unnecessary Mark namespace"

kanitw:
This reverts commit f895054a0dd611e3ffed347a96dddcaf74fa1c32.

cc: @willium turns out we actually use this everywhere in CompassQL... removing it will be a disaster --- sorry for merging conflicts

### DIFF
--- a/src/mark.ts
+++ b/src/mark.ts
@@ -1,22 +1,35 @@
 import {extend, toSet} from './util';
 
-export const AREA: 'area' = 'area';
-export const BAR: 'bar' = 'bar';
-export const LINE: 'line' = 'line';
-export const POINT: 'point' = 'point';
-export const RECT: 'rect' = 'rect';
-export const RULE: 'rule' = 'rule';
-export const TEXT: 'text' = 'text';
-export const TICK: 'tick' = 'tick';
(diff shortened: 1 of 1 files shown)


src/mark.ts
```diff
@@ -1,22 +1,35 @@
...
 
 /**
  * All types of primitive marks.
  */
-export type Mark = 'area' | 'bar'  | 'line' | 'point' |
-  'rect' | 'rule' | 'text' | 'tick' | 'circle' | 'square';
+export type Mark = typeof Mark.AREA | typeof Mark.BAR | typeof Mark.LINE | typeof Mark.POINT | typeof Mark.TEXT | typeof Mark.TICK | typeof Mark.RECT | typeof Mark.RULE | typeof Mark.CIRCLE | typeof Mark.SQUARE;
+
+
+export const AREA = Mark.AREA;
```
domoritz: `export * from Mark`?

kanitw: doesn't work ><"

kanitw:
Note: I agree that I don't like this too but it's not the top priority to invent a better pattern right now.

PR #1957, opened 2017-02-17
Redirect compiled output to `build/`

kanitw:
See https://github.com/vega/compassql/releases/tag/v0.9.2 for rationale


kanitw:
@willium @domoritz Build and test are working file. 

However, the website seems to be broken even in master, so I'm gonna defer to you guys to fix it.

domoritz:
Not sure what you mean by broken... the website works for me.

site/static/main.js
```diff
@@ -1,36 +1,34 @@
 /* tslint:disable:no-unused-variable */
```
domoritz: Let's leave this as ts and instead invoke tsc just for this one file.

domoritz:
Let's get https://github.com/vega/vega-lite/pull/1984 in first so we know the site works before these changes.

site/static/main.js
kanitw: why?  I think it's pretty annoying and didn't like this to be TS from the beginning.  This is definitely an overkill.

domoritz: Because we can check that the code is compatible. Mainly, I like that we don't have any js in the repo. Makes it easier.

kanitw: But you will have to run tsc locally in this folder and etc. (or import from `build/site/static`). IMHO, this complexity already outweighs the minor benefits you like, so I'm *not* gonna do that.  If you prefer that you can submit another PR later.

domoritz:
Image builds work again

domoritz:
Ping me when you want me to look at it/ merge.

kanitw:
@domoritz  I think this works with the website. 

One question -- should the schema go in `build/`?  
In a way, we are building it, but we also check it in.

kanitw:
Hmm, not sure what's failing, but I can't open the test log in travis.  Maybe travis is broken today.

site/static/main.js
domoritz: Let's do this in a separate pr. If we want to compile a bundle, ts will be useful.

domoritz:
The website does not work with this pr.

Github doesn't let me upload a screenshot right now but looks like bower is not set up correctly.

kanitw:
@domoritz 

> Github doesn't let me upload a screenshot right now but looks like bower is not set up correctly.
 
You need to run npm install and npm run build in Vega because Vega doesn't include `build/` in bower

domoritz:
@kanitw I think you forgot to add `bower install` to `npm run site`.

site/static/main.js
domoritz: The test gallery still uses ts. Why don't we just use the same approach that use their for this file as well?

kanitw: Actually does the test gallery works?

kanitw: Oh man.... that's another point in case of an overkill..

That file can easily be a JS and we can remove one more ugly npm script.

"build:test-gallery": "browserify test-gallery/main.ts -p [tsify -p test-gallery] -d > test-gallery/main.js",

domoritz:
Corresponding cdnjs issue https://github.com/cdnjs/cdnjs/pull/10760

domoritz:
No idea why this fails.

kanitw:
@domoritz That seems pretty messy, which makes me wonder if we're going with the right approach.

package.json
```diff
@@ -20,13 +20,12 @@
     "test": "test"
   },
   "scripts": {
-    "prebuild": "mkdir -p build",
-    "build": "browserify src/vl.ts -p tsify -g [ babelify ] -d -s vl | exorcist build/vega-lite.js.map > build/vega-lite.js && cp package.json build",
-    "postbuild": "uglifyjs build/vega-lite.js -cm --in-source-map build/vega-lite.js.map --source-map build/vega-lite.min.js.map > build/vega-lite.min.js && tsc && npm run schema",
-    "build:all": "npm run clean && npm run build && npm run test:noschema && npm run build:images",
+    "prebuild": "mkdir -p build/site build/examples/images",
+    "build": "tsc && cp package.json build && browserify build/src/vl.js -g [ babelify ] -d -s vl | exorcist build/vega-lite.js.map > build/vega-lite.js",
```
kanitw: Removing `tsify` make sourcemap no longer works for browserify bundle (e.g., in editor)

kanitw:
@domoritz Also, removing tsify breaks sourcemap in browserify bundle.  Sourcemap no longer works in the editor because of that.  Source map still works fine in the editor for master. (But broken for mocha.)

domoritz:
And with tsify that does not happen?

domoritz:
https://stackoverflow.com/questions/32486196/preserve-original-sourcemap-with-browserify

PR #1961, opened 2017-02-17
Restructure Mark to be very lean with mixins

kanitw:
This will make production rules implementation way easier. 

Please merge #1968 first!


src/compile/mark/area.ts
```diff
@@ -11,33 +12,37 @@ import * as ref from './valueref';
...
+    // TODO(zoe): this should become just a line below that call mixins.color(...)
+    let colorMixins = {};
+    applyColor(colorMixins, model);
+
+    return {
+      x: ref.stackable(X, encoding.x, model.scaleName(X), model.scale(X), stack, 'base'),
+      y: ref.stackable(Y, encoding.y, model.scaleName(Y), model.scale(Y), stack, 'base'),
+      ...x2y2(model, orient),
+      ...colorMixins,
+      ...mixins.midPoint('opacity', model.encoding.opacity, model.scaleName('opacity'), model.scale('opacity'), undefined),
```
kanitw: @YuhanLu Notice that I no longer include `config.mark.opacity && {value: config.mark.opacity}`

kanitw: Plus, the 
```
if (opacity !== undefined) {		 +    };
   e.opacity = opacity;		
}
```

are consolidated into a one-liner.

src/compile/mark/area.ts
```diff
@@ -12,32 +12,36 @@ export const area: MarkCompiler = {
...
-    }
-
-    applyColor(e, model);
-    applyMarkConfig(e, model, ['interpolate', 'tension']);
-    return e;
+    // TODO(zoe): this should become just a line below that call mixins.color(...)
+    let colorMixins = {};
+    applyColor(colorMixins, model);
+
+    return {
```
kanitw: @YuhanLu Let's order properties from  x, y, x2,y2, width, height, color, opacity, other channels

domoritz:
The partial stuff from ts 2.1 might be useful here https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html.

PR #1962, opened 2017-02-17
Rename VlOnlyAxisConfig => VlOnlyAxisBase

kanitw:
(It should be *Base as labelMaxLength is an Axis property too.)

Also rename one local variable for consistency

cc: @Yuhanlu

(Please use the same format for `VlOnlyLegendBase`)

I think we should also ask Jeff if he wanna add `labelMaxLength` and `titleMaxLength` to Axis/Legend and their config in Vega.  (Maybe we can get rid of `VlOnlyAxis/LegendBase` and complicated code cause by it.)

### DIFF
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -3,7 +3,7 @@ import {VgAxisEncode, VgAxisBase, VgAxisConfig} from './vega.schema';
 
 export type AxisOrient = 'top' | 'right' | 'left' | 'bottom';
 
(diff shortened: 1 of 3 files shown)


domoritz:
The reason for renaming it is that Base is not a complete config?

kanitw:
> The reason for renaming it is that Base is not a complete config?

Yes.  Add we have `VgAxisBase` for exactly the same thing (but not VL only)

PR #1967, opened 2017-02-18
Update Mark Config

kanitw:


src/config.ts
```diff
@@ -161,7 +161,7 @@ export interface Config {
 
   // MARK-SPECIFIC CONFIGS
   /** Area-Specific Config */
-  area?: AreaConfig;
+  area?: MarkConfig;
```
domoritz: So area can have a shape now?

kanitw: No but we dont want to define so many subversion of it.  If users really wanna specify it, good luck!

domoritz: Makes sense. This will also make it easier to switch from one mark to another without having to delete pros just to make the spec valid.

PR #1971, opened 2017-02-21
Support conditional encoding

kanitw:
(Merge #1970 first)

@arvind I just added scaffolding code for generating a Vega production rule from our conditional encoding syntax.  

I'll let you take over and add code for how to generate `"test"` property for selection. 

Since we no longer have to deal with how it interacts with transform, I think we're pretty close to have selection really working!


arvind:
Sweet, thank you for this! Gonna make my life *much* easier.

PR #1972, opened 2017-02-22
Replace scale.useRawDomain with scale.domain = ‘unaggregated’

domoritz:
Part of #1602

### DIFF
--- a/src/compile/scale/domain.ts
+++ b/src/compile/scale/domain.ts
@@ -15,7 +15,7 @@ import {Model} from '../model';
 export default function domain(scale: Scale, model: Model, channel:Channel): any[] | VgDataRef | FieldRefUnionDomain {
   const fieldDef = model.fieldDef(channel);
 
-  if (scale.domain) { // explicit value
+  if (scale.domain && scale.domain !== 'unaggregated') { // explicit value
     if (isDateTime(scale.domain[0])) {
       return (scale.domain as DateTime[]).map((dt) => {
         return timestamp(dt, true);
@@ -51,12 +51,12 @@ export default function domain(scale: Scale, model: Model, channel:Channel): any
     };
   }
 
-  // FIXME refactor _useRawDomain's signature
-  const useRawDomain = _useRawDomain(scale, model, channel);
+  // FIXME refactor signature
+  const noAggDomain = useUnaggregatedDomain(scale, model, channel);
 
   const sort = domainSort(model, channel, scale.type);
 
-  if (useRawDomain) { // useRawDomain - only Q/T
+  if (noAggDomain) {
     return {
       data: SOURCE,
       field: model.field(channel, {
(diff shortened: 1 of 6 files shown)


src/compile/scale/domain.ts
```diff
@@ -143,7 +143,7 @@ export function domainSort(model: Model, channel: Channel, scaleType: ScaleType)
 function _useRawDomain (scale: Scale, model: Model, channel: Channel) {
   const fieldDef = model.fieldDef(channel);
 
-  return scale.useRawDomain && //  if useRawDomain is enabled
+  return scale.domain === 'pre-aggregation' && //  if useRawDomain is enabled
```
kanitw: // if use pre-aggregate domain.

src/compile/scale/domain.ts
```diff
@@ -15,7 +15,7 @@ import {Model} from '../model';
 export default function domain(scale: Scale, model: Model, channel:Channel): any[] | VgDataRef | FieldRefUnionDomain {
   const fieldDef = model.fieldDef(channel);
 
-  if (scale.domain) { // explicit value
+  if (scale.domain && scale.domain !== 'pre-aggregation') { // explicit value
```
kanitw: is `"pre-aggregate"` better?  (It's at least shorter.)

domoritz: Sure. Will replace.

src/scale.ts
```diff
@@ -170,8 +163,11 @@ export interface Scale {
   type?: ScaleType;
   /**
    * The domain of the scale, representing the set of data values. For quantitative data, this can take the form of a two-element array with minimum and maximum values. For ordinal/categorical data, this may be an array of valid input values.
+   *
+   * If the domain is 'pre-aggregation', we use the source data range as scale domain instead of aggregated data for aggregate axis.
+   * This property only works with aggregate functions that produce values within the raw data domain (`"mean"`, `"average"`, `"stdev"`, `"stdevp"`, `"median"`, `"q1"`, `"q3"`, `"min"`, `"max"`). For other aggregations that produce values outside of the raw data domain (e.g. `"count"`, `"sum"`), this property is ignored.
```
kanitw: I think stdev and stdevp can be out of min/max range.

test/compile/scale/domain.test.ts
```diff
@@ -145,14 +144,14 @@ describe('compile/scale', () => {
           assert.deepEqual(_domain.data, SOURCE);
         });
 
-      it('should return the raw domain if useRawDomain is true for year T',
+      it('should return the pre-aggregationd domain if requested for year T',
```
kanitw: typo: extra "d"

src/compile/scale/domain.ts
domoritz: I like `pre-aggregation` better because it doesn't sound like a verb.

src/scale.ts
domoritz: Yes.

test/compile/scale/domain.test.ts
domoritz: global replace issue

src/scale.ts
kanitw: Yep, and we have removed two of these from `SHARED_DOMAIN_OPS` already, so we just need to update comment here :)

src/compile/scale/domain.ts
domoritz: Fixed

src/scale.ts
domoritz: Done

src/compile/scale/domain.ts
kanitw: "pre-aggregate"

domoritz:
@kanitw Note that I removed the config property. If you need it for compass, shout.

kanitw:
Yes, I do.

src/compile/scale/domain.ts
domoritz: could still be a verb

domoritz:
Done.

src/compile/scale/domain.ts
arvind: "unaggregated"?

src/compile/scale/domain.ts
```diff
@@ -51,12 +51,12 @@ export default function domain(scale: Scale, model: Model, channel:Channel): any
     };
   }
 
-  // FIXME refactor _useRawDomain's signature
-  const useRawDomain = _useRawDomain(scale, model, channel);
+  // FIXME refactor signature
+  const noAggDomain = _useUnaggregatedDomain(scale, model, channel);
```
kanitw: remove _ for the method?

domoritz: Done. Merge when you feel like it.

PR #1973, opened 2017-02-22
remove need for source data set when using unaggregated domain

domoritz:


src/compile/data/summary.ts
```diff
@@ -45,6 +45,13 @@ export namespace summary {
         } else {
           meas[fieldDef.field] = meas[fieldDef.field] || {};
           meas[fieldDef.field][fieldDef.aggregate] = true;
+
+          // add min/max so we can use their union as unaggregated domain
+          const scale = model.scale(channel);
+          if (scale && useUnaggregatedDomain(scale, model, channel)) {
```
kanitw: Just an observation, we currently `useUnaggregatedDomain` in multiple places, but maybe we should just refactor this method so that it becomes a rule for `domain` during initialization?  

If you agree, we don't have to do it now -- we can file an issue and keep this PR lean.

kanitw:
@domoritz feel free to merge after reading my comment.

PR #1974, opened 2017-02-22
Drop domain: "unaggregated" in init if it's not applicable

kanitw:


src/log.ts
```diff
@@ -146,10 +146,22 @@ export namespace message {
   // SCALE
   export const CANNOT_UNION_CUSTOM_DOMAIN_WITH_FIELD_DOMAIN = 'custom domain scale cannot be unioned with default field-based domain';
 
-  export function CANNOT_USE_SCALE_PROPERTY_WITH_NON_COLOR(prop: string) {
+  export function cannotUseScalePropertyWithNonColor(prop: string) {
```
domoritz: Use `Cannot use ${prop} with non-color channel.` instead.

src/compile/data/summary.ts
```diff
@@ -48,14 +47,14 @@ export namespace summary {
 
           // add min/max so we can use their union as unaggregated domain
           const scale = model.scale(channel);
-          if (scale && useUnaggregatedDomain(scale, model, channel)) {
+          if (scale && scale.domain === 'unaggregated') {
```
domoritz: Data is parsed before scales are parsed. However, we initialize scales before we parse them so we will have this property set correctly. This is the main idea of this PR, but I just want to make sure I understand it correctly.

src/log.ts
kanitw: Feel free to fix it.   (I didn't write this method ... and don't feel like fix it now.)  I'm still in the middle of sth else.)

src/compile/data/summary.ts
kanitw: Yep.  And that's why we have to initialize many scale properties in init phase rather than just doing everything in the parse phrase.  In this PR, I'm just making `domain` consistent with others.  :)

PR #1980, opened 2017-02-24
Support `field` in condition

kanitw:
Follow up from https://github.com/vega/vega-lite/pull/1971. 

It's probably more natural to do the following way in many cases:

```
x: {
  condition: { selection: <name>,  field: ..., type: ...},
  value: ...
}
```

Currently this PR only contains interface changes, but haven't make it work.  

This will require a major refactor since we can no longer just refer to
`encoding.<channel>.field` for accessing field, so I'm gonna punt on it for now.

- [x] initScale
- [x] initLegend
- [x] fieldDef
  - [x] channelHasField
  - [x] getFieldDef
- [x] aggregate
- [x] mixins of marks 
- [x] schema rename script

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1156,7 +1156,7 @@
       ],
       "type": "object"
     },
-    "ConditionLegendDefAlias<NumberValueDef>": {
+    "ConditionLegendDef<NumberValueDef>": {
       "anyOf": [
         {
           "$ref": "#/definitions/ConditionLegendFieldDef<NumberValueDef>"
@@ -1170,7 +1170,7 @@
       ],
       "description": "Generic type for conditional channelDef.\nF defines the underlying FieldDef type while V defines the underlying ValueDef type."
     },
-    "ConditionLegendDefAlias<StringValueDef>": {
+    "ConditionLegendDef<StringValueDef>": {
       "anyOf": [
         {
           "$ref": "#/definitions/ConditionLegendFieldDef<StringValueDef>"
@@ -1184,7 +1184,7 @@
       ],
       "description": "Generic type for conditional channelDef.\nF defines the underlying FieldDef type while V defines the underlying ValueDef type."
     },
-    "ConditionTextDefAlias": {
+    "ConditionTextDef": {
       "anyOf": [
         {
           "$ref": "#/definitions/ConditionTextFieldDef"
@@ -1390,15 +1390,6 @@
         "type"
       ],
       "type": "object"
-    },
-    "ConditionalLegendDef<number>": {
-      "$ref": "#/definitions/ConditionLegendDefAlias<NumberValueDef>"
-    },
-    "ConditionalLegendDef<string>": {
-      "$ref": "#/definitions/ConditionLegendDefAlias<StringValueDef>"
-    },
-    "ConditionalTextDef": {
-      "$ref": "#/definitions/ConditionTextDefAlias"
     },
     "ConditionLegendValueDef<NumberValueDef>": {
       "additionalProperties": false,
@@ -1792,7 +1783,7 @@
       "additionalProperties": false,
       "properties": {
         "color": {
-          "$ref": "#/definitions/ConditionalLegendDef<string>",
+          "$ref": "#/definitions/ConditionLegendDef<StringValueDef>",
           "description": "Color of the marks – either fill or stroke color based on mark type.\n(By default, fill color for `area`, `bar`, `tick`, `text`, `circle`, and `square` /\nstroke color for `line` and `point`.)"
         },
         "detail": {
@@ -1810,7 +1801,7 @@
           "description": "Additional levels of detail for grouping data in aggregate views and\nin line and area marks without mapping data to a specific visual channel."
         },
         "opacity": {
-          "$ref": "#/definitions/ConditionalLegendDef<number>",
+          "$ref": "#/definitions/ConditionLegendDef<NumberValueDef>",
           "description": "Opacity of the marks – either can be a value or a range."
         },
         "order": {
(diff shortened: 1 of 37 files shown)


domoritz:
Do we still need this?

src/compile/unit.ts
```diff
@@ -156,10 +156,13 @@ export class UnitModel extends ModelWithField {
       if (isFieldDef(channelDef)) {
         fieldDef = channelDef;
         specifiedScale = channelDef.scale;
+      } else if (isConditionalDef(channelDef) && isFieldDef(channelDef.condition)) {
+        fieldDef = channelDef.condition;
+        specifiedScale = channelDef.condition.scale;
```
arvind: Could these first two branches be consolidated using the new `getFieldDef` method?

src/compile/unit.ts
```diff
@@ -249,12 +252,15 @@ export class UnitModel extends ModelWithField {
   private initLegend(encoding: Encoding<string>): Dict<Legend> {
     return NONSPATIAL_SCALE_CHANNELS.reduce(function(_legend, channel) {
       const channelDef = encoding[channel];
-      if (isFieldDef(channelDef)) {
-        const legendSpec = channelDef.legend;
-        if (legendSpec !== null && legendSpec !== false) {
-          _legend[channel] = {...legendSpec};
+      if (channelDef) {
+        const legend = isFieldDef(channelDef) ? channelDef.legend :
+          (channelDef.condition && isFieldDef(channelDef.condition)) ? channelDef.condition.legend : null;
```
arvind: Similarly, here, rather than multiple ternaries depending on where the `fieldDef` might be, could this be simplified with a single `getFieldDef` call?

src/channel.ts
```diff
@@ -55,6 +55,10 @@ export const TOOLTIP = Channel.TOOLTIP;
 export const CHANNELS = [X, Y, X2, Y2, ROW, COLUMN, SIZE, SHAPE, COLOR, ORDER, OPACITY, TEXT, DETAIL, TOOLTIP];
 const CHANNEL_INDEX = toSet(CHANNELS);
 
+export const SINGLE_DEF_CHANNELS = [X, Y, X2, Y2, ROW, COLUMN, SIZE, SHAPE, COLOR, OPACITY, TEXT, TOOLTIP];
```
arvind: What is a "single def" channel? Perhaps a comment to explain what this class of channels is and why it's needed?

src/compile/axis/rules.ts
```diff
@@ -29,12 +29,7 @@ export function gridShow(model: UnitModel, channel: Channel) {
   return !model.hasDiscreteDomain(channel) && !model.fieldDef(channel).bin;
 }
 
-export function grid(model: UnitModel, channel: Channel, isGridAxis: boolean) {
-  if (channel === ROW || channel === COLUMN) {
-    // never apply grid for ROW and COLUMN since we manually create rule-group for them
-    return false;
-  }
-
+export function grid(model: UnitModel, channel: SpatialScaleChannel, isGridAxis: boolean) {
```
arvind: For my edification, why do we no longer skip grid for `ROW` and `CHANNEL`?

src/compile/selection/multi.ts
```diff
@@ -15,8 +15,10 @@ const multi:SelectionCompiler = {
     const fields = proj.map((p) => stringValue(p.field)).join(', ');
     const values = proj.map((p) => {
       const channel = p.encoding;
+      const fieldDef = model.fieldDef(channel);
       // Binned fields should capture extents, for a range test against the raw field.
-      return model.fieldDef(channel).bin ? (bins[p.field] = 1,
+      // FIXME: Arvind -- please log proper warning when the specified encoding channel has no field
```
arvind: Good catch!

src/compile/selection/transforms/project.ts
```diff
@@ -11,12 +12,16 @@ const project:TransformCompiler = {
...
-        fields[model.field(channel)] = channel;
+      if (fieldDef) {
+        if (fieldDef.timeUnit) {
+          fields[model.field(channel)] = channel;
+        } else {
+          fields[fieldDef.field] = channel;
+        }
       } else {
-        fields[fieldDef.field] = channel;
+        log.warn(log.message.cannotProjectOnChannelWithoutField(channel));
```
arvind: Thanks!

src/fielddef.ts
```diff
@@ -173,7 +173,18 @@ export interface TextFieldDef<F> extends FieldDef<F> {
...
+
+/**
+ * Return if a channelDef is a ConditionalValueDef with ConditionFieldDef
+ */
+export function hasConditionFieldDef<F>(channelDef: ChannelDef<F>): channelDef is (ValueDef<any> & {condition: FieldDef<F>}) {
+  return !!channelDef && !!channelDef.condition && isFieldDef(channelDef.condition);
+}
 
 export function isFieldDef<F>(channelDef: ChannelDef<F>): channelDef is FieldDef<F> | PositionFieldDef<F> | LegendFieldDef<F, any> | OrderFieldDef<F> | TextFieldDef<F> {
   return !!channelDef && (!!channelDef['field'] || channelDef['aggregate'] === 'count');
```
kanitw: Yeah,  say a channelDef could be 

1) FieldDef (with no condition)
2) FieldDef with ConditionValueDef

(1 & 2 are `ConditionalFieldDef` as I make `condition?` so I don't have to declare too many types)

3) ValueDef (with no condition)
4) ValueDef with ConditionFieldDef
5) ValueDef with ConditionValueDef

(3-5 are `ConditionalValueDef` as I make `condition?` so I don't have to declare too many types)

6) ConditionFieldDef only  (akin to if without else)
7) ConditionValueDef only 

(6-7 are `ConditionOnlyDef`)

`isFieldDef` is more scoped to one object and can check if the object is a fieldDef.  
The input can be **either**:
- a)  the parent object of a channelDef (`encoding.<channel>`), which will return `true` for 1 or 2, 
- b) the condition object of a channelDef  (`encoding.<channel>.condition`), which will return true for case 4 or 6.  

On the other hand, `getFieldDef` is for getting the `fieldDef` from a channelDef. 
Given a channelDef (from `encoding.<channel>`), this will return either the parent channelDef object (for case 1 or 2) or its condition object (for case 4 or 6).  

In a way, getFieldDef is more like `getFieldDefFromChannelDef` but it is kinda verbose, so I just chose the brief getFieldDef. That said, I agree that it can be confusing. 

If you have better name suggestions, let me know :)

src/compile/unit.ts
kanitw: Yeah I tried this earlier too. However, TypeScript won't know that `channelDef.condition` is not only a FieldDef, but also a `ScaleFieldDef`.  So we either have to do type casting or have to do this and I generally prefer to avoid casting.

src/compile/unit.ts
kanitw: Similar to https://github.com/vega/vega-lite/pull/1980#discussion_r121235425, TS won't know that this is a `LegendFieldDef`, not just `FieldDef`.

src/compile/axis/rules.ts
kanitw: With the new layout operator, we no longer use axes for row/column but use layout header instead.   Here I basically forgot to remove the code.

src/channel.ts
kanitw: Good catch.  

I'll add
```
/**
 * Channels cannot have an array of channelDef.
 * model.fieldDef, getFieldDef only work for these channels.
 *
 * (The only two channels that can have an array of channelDefs are "detail" and "order".
 * Since there can be multiple fieldDefs for detail and order, getFieldDef/model.fieldDef
 * are not applicable for them.  Similarly, selection projecttion won't work with "detail" and "order".)
 */
```

build/vega-lite-schema.json
```diff
@@ -1032,32 +1032,47 @@
...
         }
       },
       "required": [
-        "selection",
-        "value"
+        "selection"
+      ],
+      "type": "object"
+    },
+    "Condition<(LegendFieldDef|ValueDef<string>)>": {
```
kanitw: @domoritz  looks like we have a bug in the schema generator?

Some properties are missing here

kanitw: (None of the properties from `LegendFieldDef` or `ValueDef` are here at all.)

kanitw:
@domoritz 

>Upgrade to https://github.com/vega/typescript-to-json-schema/releases/tag/v0.4.0


I got `Error: Cannot find module '../dist/typescript-to-json-schema'` after upgrading

kanitw:
@arvind -- I added a few more commits to simplify the schema.  
Now we longer have generics in the schema.  I also re-declare the part that triggers a bug in the schema generator with a better way to define it and also avoid the bug.

arvind:
Travis is detecting differences with the schema. Is that due to the issue you mention above?

kanitw:
@arvind I don't think so.  I just force push an update.  looks like I somehow forgot to commit the diff.

kanitw:
Note that we can update the schema and remove https://github.com/vega/vega-lite/pull/1980/commits/b82b8abfac5bd8b5be2f80db4bf2004097f71934 in a separate PR. 
So if this passes and you like everything else, we can merge.

Issue #1684, opened 2016-11-05
Use `zindex` for layer `order` channel instead 


kanitw:
Currently we use sort transform to determine layer order for `order` channel. 
In Vega 3, we can consider using `zindex` property instead.

kanitw:
Maybe we just add `zindex` or `layer` as its own channel?  

This might allow us to make `order` means either path order or stack order and free up `path` channel for Geo stuff.

jheer:
In case this effects your reasoning (it may not, but just to be safe): `zindex` is used only for sorting **sibling nodes** in the scenegraph tree. If `zindex` is set on individual items, it will have no effect on the ordering relationship between items that are children of different marks.

kanitw:
Since zindex expects non-negative integer, looks like we have to apply rank transform to determine range of zindex scale.

kanitw:
# TODOs

- [ ]  Try modifying Vega manually
  - [ ] add scale for `zindex` based on origin. 
  - [ ] How can we automatically determine range for `zindex`'s scale 
  - [ ] Add `zindex` to mark's `encode` using the scale above. 

- [ ] Modify the code
  - [ ] deal with the range for zindex scale
  - [ ] Add zindex following the typical steps for adding channel in #1278

kanitw:
> How can we automatically determine range for zindex's scale

If we really need to do this, we might want to add this as a part of the layout data source rather than adding a new data source.

kanitw:
Since this is (1) complicated and (2) doesn't seem that useful anyway.  Let's not include `zindex` in VL2.  Whenever users need to re-order zindex their marks, it's a sign of overplotting anyway.

arvind:
`zindex` support would be helpful for interactive use-cases to pull selected points to the foreground:

z-index-less: partially occluded points can be selected, but then are only partially visible.
![untitled](https://cloud.githubusercontent.com/assets/42262/23317361/ec4e4c18-fa82-11e6-86a9-60ec0faa16a2.gif)

with a simple z-index definition (in Vega)
![untitled2](https://cloud.githubusercontent.com/assets/42262/23317362/ed81383e-fa82-11e6-9b2d-438f6abc93be.gif)

kanitw:
Actually, one quick question --- do we always want this behavior for selected point?  
If so, we could add zindex to the output Vega without adding new encoding channel in Vega too.

(Or even if we don't, we make this a config.)

arvind:
[...]

PR #1982, opened 2017-02-25
Initial interactive examples.

arvind:
Four interactive examples to kick us off:

1. Rectangular brush:
brush

2. Paintbrush
paintbrush

3. Pan & Zoom a Scatterplot
panzoom_scatter

4. Dynamic query widgets -- this one would greatly benefit from `zindex` support (#1684).
query_widgets

### DIFF
--- a/examples/specs/brush.vl.json
+++ b/examples/specs/brush.vl.json
@@ -1,0 +1,22 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "Drag out a rectangular brush to highlight points.",
+  "data": {"url": "data/cars.json"},
+  "selection": {
+    "brush": {
+      "type": "interval"
+    }
+  },
+  "mark": "point",
+  "encoding": {
+    "x": {"field": "Horsepower", "type": "quantitative"},
+    "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
+    "color": {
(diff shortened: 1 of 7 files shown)


arvind:
I'm having trouble understanding why the schema test is failing on the `bind` definitions for `query_widgets.vl.json`. The TypeScript [interface definition](https://github.com/vega/vega-lite/blob/master/src/compile/selection/selection.ts#L24) looks correct to me. Any ideas @domoritz, @kanitw?

kanitw:
These look awesome!

examples/specs/brush.vl.json
```diff
@@ -0,0 +1,22 @@
...
+      "type": "interval"
+    }
+  },
+  "mark": "point",
+  "encoding": {
+    "x": {"field": "Horsepower", "type": "quantitative"},
+    "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
+    "color": {
+      "field": "Cylinders", "type": "ordinal",
+      "condition": {
```
kanitw: @arvind  A follow up question

Stylistically, I'm wondering if we should put condition before other properties so it feels more like if-else
```
condition: { value for condition }
...value for else...
```

arvind: Sure, will make the change.

arvind: Done with 0a222b3.

PR #1983, opened 2017-02-25
[WIP] Sw/fix multi bin

### LINKED ISSUE #1747: Cannot bin the same field twice
`Uncaught Error: Duplicate signal name: Rotten_Tomatoes_Rating_extent`

NeelMohapatra:
Fixes #1747

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -1,4 +1,4 @@
-import {Channel, COLUMN, ROW, SHAPE, SIZE} from './channel';
+import {isBoolean} from './util';
 
 export interface BinBase {
   /**
@@ -46,16 +46,9 @@ export interface Bin extends BinBase {
   extent?: number[];  // VgBinTransform uses a different extent so we need to pull this out.
 }
 
-export function autoMaxBins(channel: Channel): number {
-  switch (channel) {
-    case ROW:
-    case COLUMN:
-    case SIZE:
-      // Facets and Size shouldn't have too many bins
-      // We choose 6 like shape to simplify the rule
-    case SHAPE:
-      return 6; // Vega's "shape" has 6 distinct values
-    default:
-      return 10;
+export function binToString(bin: Bin | boolean) {
+  if (isBoolean(bin)) {
+    return 'bin';
   }
+  return 'bin' + Object.keys(bin).map(p => `_${p}_${bin[p]}`).join('');
 }
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -65,5 +65,5 @@ export function autoMaxBins(channel: Channel): number {
       return 6; // Vega's "shape" has 6 distinct values
     default:
       return 10;
-   }
+  }
 }
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -1,3 +1,4 @@
+import {Channel, COLUMN, ROW, SHAPE, SIZE} from './channel';
 import {isBoolean} from './util';
 
 export interface BinBase {
@@ -51,4 +52,18 @@ export function binToString(bin: Bin | boolean) {
     return 'bin';
   }
   return 'bin' + Object.keys(bin).map(p => `_${p}_${bin[p]}`).join('');
+}
+
+export function autoMaxBins(channel: Channel): number {
+  switch (channel) {
+    case ROW:
+    case COLUMN:
+    case SIZE:
+      // Facets and Size shouldn't have too many bins
+      // We choose 6 like shape to simplify the rule
+    case SHAPE:
+      return 6; // Vega's "shape" has 6 distinct values
+    default:
+      return 10;
+   }
 }
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -4,8 +4,8 @@ import {Bin, binToString} from '../../bin';
 import {Channel} from '../../channel';
 import {field, FieldDef} from '../../fielddef';
 import {hasDiscreteDomain} from '../../scale';
-import {Dict, extend, flatten, isBoolean, vals} from '../../util';
-import {VgTransform} from '../../vega.schema';
+import {Dict, extend, flatten, isBoolean, vals, varName} from '../../util';
+import {VgBinTransform, VgTransform} from '../../vega.schema';
 
 import {FacetModel} from './../facet';
 import {LayerModel} from './../layer';
@@ -47,23 +47,31 @@ function parse(model: Model): Dict<VgTransform[]> {
       if (!transform) {
         binComponent[key] = transform = [];
         const extentSignal = model.getName(key + '_extent');
-        const binTrans: VgTransform = {
-          ...{
+        // const binTrans: VgBinTransform = {
+        //   ...{
+        //     type: 'bin',
+        //     field: fieldDef.field,
+        //     as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})],
+        //     signal: varName(model.getName(key + '_bins'))
+        //   },
+        //   ...bin,
+        //   // add extent if it's not specified
+        //   ...(!bin.extent ? {extent: {signal: extentSignal}} : {}),
+        // };
+        let binTrans: VgBinTransform = {
             type: 'bin',
             field: fieldDef.field,
             as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})],
-            signal: model.getName(key + '_bins')
-          },
-          ...bin,
-          // add extent if it's not specified
-          ...(!bin.extent ? {extent: {signal: extentSignal}} : {}),
+            signal: varName(model.getName(key + '_bins'))
         };
+        extend(binTrans, bin);
         if (!bin.extent) {
           transform.push({
             type: 'extent',
             field: fieldDef.field,
             signal: extentSignal
           });
+          binTrans.extent = {signal: extentSignal};
         }
         transform.push(binTrans);
       }
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -1,11 +1,11 @@
 import {DataComponentCompiler} from './base';
 
-import {autoMaxBins} from '../../bin';
+import {Bin, binToString} from '../../bin';
 import {Channel} from '../../channel';
 import {field, FieldDef} from '../../fielddef';
 import {hasDiscreteDomain} from '../../scale';
-import {Dict, extend, flatten, hash, isBoolean, vals, varName} from '../../util';
-import {VgBinTransform, VgTransform} from '../../vega.schema';
+import {Dict, extend, flatten, isBoolean, vals} from '../../util';
+import {VgTransform} from '../../vega.schema';
 
 import {FacetModel} from './../facet';
 import {LayerModel} from './../layer';
@@ -16,60 +16,61 @@ function numberFormatExpr(expr: string, format: string) {
   return `format(${expr}, '${format}')`;
 }
 
+function addRangeFormula(model: Model, transform: VgTransform[], fieldDef: FieldDef, channel: Channel) {
+    const discreteDomain = hasDiscreteDomain(model.scale(channel).type);
+    if  (discreteDomain) {
+      // read format from axis or legend, if there is no format then use config.numberFormat
+      const format = (model.axis(channel) || model.legend(channel) || {}).format ||
+        model.config.numberFormat;
+
+      const startField = field(fieldDef, {datum: true, binSuffix: 'start'});
+      const endField = field(fieldDef, {datum: true, binSuffix: 'end'});
+
+      transform.push({
+        type: 'formula',
+        as: field(fieldDef, {binSuffix: 'range'}),
+        expr: `${numberFormatExpr(startField, format)} + ' - ' + ${numberFormatExpr(endField, format)}`
+      });
+    }
+}
+
 function parse(model: Model): Dict<VgTransform[]> {
   return model.reduceFieldDef(function(binComponent: Dict<VgTransform[]>, fieldDef: FieldDef, channel: Channel) {
-    const bin = model.fieldDef(channel).bin;
-    if (bin) {
-
-       let binTrans: VgBinTransform = {
-        type: 'bin',
-        field: fieldDef.field,
-        as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})],
-        signal: varName(model.getName(fieldDef.field + '_bins'))
-      };
-
-      if (!isBoolean(bin)) {
-         binTrans = extend(binTrans, bin);
+    const fieldDefBin = model.fieldDef(channel).bin;
+    if (fieldDefBin) {
+      const bin: Bin = isBoolean(fieldDefBin) ? {} : fieldDefBin;
+      if (!bin.maxbins && !bin.steps) {
+        bin.maxbins = model.config.maxbins;
       }
-
-      const transform: VgTransform[] = [];
-
-      if (!binTrans.extent) {
-        const extentSignal = varName(model.getName(fieldDef.field + '_extent'));
-        transform.push({
-          type: 'extent',
-          field: fieldDef.field,
-          signal: extentSignal
-        });
-
-        binTrans.extent = {signal: extentSignal};
+      const key = `${binToString(fieldDef.bin)}_${fieldDef.field}`;
+      let transform: VgTransform[] = binComponent[key];
+      if (!transform) {
+        binComponent[key] = transform = [];
+        const extentSignal = model.getName(key + '_extent');
+        const binTrans: VgTransform = {
+          ...{
+            type: 'bin',
+            field: fieldDef.field,
+            as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})],
+            signal: model.getName(key + '_bins')
+          },
+          ...bin,
+          // add extent if it's not specified
+          ...(!bin.extent ? {extent: {signal: extentSignal}} : {}),
+        };
+        if (!bin.extent) {
+          transform.push({
+            type: 'extent',
+            field: fieldDef.field,
+            signal: extentSignal
+          });
+        }
+        transform.push(binTrans);
       }
-
-      if (!binTrans.maxbins && !binTrans.step) {
-        // if both maxbins and step are not specified, need to automatically determine bin
-        binTrans.maxbins = autoMaxBins(channel);
+      // if formula doesn't exist already
+      if (transform.length > 0 && transform[transform.length - 1].type !== 'formula') {
+        addRangeFormula(model, binComponent[key], fieldDef, channel);
       }
-
-      transform.push(binTrans);
-
-      const discreteDomain = hasDiscreteDomain(model.scale(channel).type);
-      if (discreteDomain) {
-        // read format from axis or legend, if there is no format then use config.numberFormat
-        const format = (model.axis(channel) || model.legend(channel) || {}).format ||
-          model.config.numberFormat;
-
-        const startField = field(fieldDef, {datum: true, binSuffix: 'start'});
-        const endField = field(fieldDef, {datum: true, binSuffix: 'end'});
-
-        transform.push({
-          type: 'formula',
-          as: field(fieldDef, {binSuffix: 'range'}),
-          expr: `${numberFormatExpr(startField, format)} + ' - ' + ${numberFormatExpr(endField, format)}`
-        });
-      }
-      // FIXME: current merging logic can produce redundant transforms when a field is binned for color and for non-color
-      const key = hash(bin) + '_' + fieldDef.field + 'oc:' + discreteDomain;
-      binComponent[key] = transform;
     }
     return binComponent;
   }, {});
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -45,7 +45,7 @@ function parse(model: Model): Dict<VgTransform[]> {
         binComponent[key] = transform = [];
         const extentSignal = model.getName(key + '_extent');
 
-        let binTrans: VgBinTransform = {
+        const binTrans: VgBinTransform = {
             type: 'bin',
             field: fieldDef.field,
             as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})],
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -49,9 +49,9 @@ function parse(model: Model): Dict<VgTransform[]> {
             type: 'bin',
             field: fieldDef.field,
             as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})],
-            signal: varName(model.getName(key + '_bins'))
+            signal: varName(model.getName(key + '_bins')),
+            ...bin
         };
-        extend(binTrans, bin); // Has side effect. TODO: Try to remove them
         if (!bin.extent) {
           transform.push({
             type: 'extent',
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -39,9 +39,6 @@ function parse(model: Model): Dict<VgTransform[]> {
     const fieldDefBin = model.fieldDef(channel).bin;
     if (fieldDefBin) {
       const bin: Bin = isBoolean(fieldDefBin) ? {} : fieldDefBin;
-      if (!bin.maxbins && !bin.steps) {
-        bin.maxbins = model.config.maxbins;
-      }
       const key = `${binToString(fieldDef.bin)}_${fieldDef.field}`;
       let transform: VgTransform[] = binComponent[key];
       if (!transform) {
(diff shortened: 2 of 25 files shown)


src/compile/data/bin.ts
```diff
@@ -19,36 +19,42 @@ function numberFormatExpr(expr: string, format: string) {
...
-
-      const transform: VgTransform[] = [];
-      if (!binTrans.extent) {
-        const extentSignal = model.getName(fieldDef.field + '_extent');
-        transform.push({
-          type: 'extent',
+      const key = hash(bin) + fieldDef.field;
+      // FIXME: Check
+      if (!binComponent[key]) {
+        let binTrans: VgTransform = {
```
domoritz: why not const?

src/compile/data/bin.ts
```diff
@@ -19,36 +19,42 @@ function numberFormatExpr(expr: string, format: string) {
 function parse(model: Model): Dict<VgTransform[]> {
   return model.reduceFieldDef(function(binComponent: Dict<VgTransform[]>, fieldDef: FieldDef, channel: Channel) {
     const bin = model.fieldDef(channel).bin;
+    // Make this bin variable always an object
+    // and extend it with maxbins if needed here.
+    let transform: VgTransform[] = [];
```
domoritz: const?

src/compile/data/bin.ts
```diff
@@ -19,36 +19,42 @@ function numberFormatExpr(expr: string, format: string) {
...
-        typeof bin === 'boolean' ? {} : bin
-      );
-
-      const transform: VgTransform[] = [];
-      if (!binTrans.extent) {
-        const extentSignal = model.getName(fieldDef.field + '_extent');
-        transform.push({
-          type: 'extent',
+      const key = hash(bin) + fieldDef.field;
+      // FIXME: Check
```
domoritz: check what?

src/compile/data/bin.ts
```diff
@@ -19,36 +19,42 @@ function numberFormatExpr(expr: string, format: string) {
...
+          transform.push({
+            type: 'extent',
+            field: fieldDef.field,
+            signal: extentSignal
+          });
+
+          binTrans.extent = {signal: extentSignal};
+        }
+
+        //
```
domoritz: what do you want to say?

src/compile/data/bin.ts
```diff
@@ -19,36 +19,42 @@ function numberFormatExpr(expr: string, format: string) {
...
+        if (!binTrans.maxbins && !binTrans.step) {
+          // if both maxbins and step are not specified, need to automatically determine bin
+          binTrans.maxbins = autoMaxBins(channel);
+        }
+
+        transform.push(binTrans);
+      } else {
+        transform = binComponent[key].slice();
```
domoritz: what are you slicing here?

typings/vega-util.d.ts
```diff
@@ -22,7 +22,7 @@ declare module 'vega-util' {
   export function truncate(a: string, length: number): string;
 
   export function isArray(a: any | any[]): a is any[];
-  export function isObject(a: any): a is any;
+  export function isObject(a: any): a is Object;
```
domoritz: I think you mean `object`? Why change this? Also, this is a ts 2.2 feature.

NeelMohapatra: I think @kanitw suggested that

kanitw: Actually I did try that with you and it *did not* work yet. Instead we used isBoolean, so we don't need this.  

That said, we should change this later when we upgrade to TS2.2 as the original `is any` is also wrong.

src/compile/data/bin.ts
```diff
@@ -83,7 +88,7 @@ export const bin: DataComponentCompiler<Dict<VgTransform[]>> = {
     // If child doesn't have its own data source, then merge
     if (!childDataComponent.source) {
       // FIXME: current merging logic can produce redundant transforms when a field is binned for color and for non-color
-      extend(binComponent, childDataComponent.bin);
+      binComponent = {...childDataComponent.bin};
```
kanitw: This does not have the same logic as the old one.

src/compile/data/bin.ts
```diff
@@ -97,7 +102,7 @@ export const bin: DataComponentCompiler<Dict<VgTransform[]>> = {
 
       // If child doesn't have its own data source, then merge
       if (!childDataComponent.source) {
-        extend(binComponent, childDataComponent.bin);
+        binComponent = {...childDataComponent.bin};
```
kanitw: This does not have the same logic as the old one.

src/compile/data/bin.ts
kanitw: The equivalent code to the old code would be

```
binComponent = {
  ...binComponent,
  ...childDataComponent.bin
}
```

That said, I think when we want to cause side effect the old `extend` method seems better.  
(I take back my comment from slack this morning.)

src/compile/data/bin.ts
```diff
@@ -65,7 +71,6 @@ function parse(model: Model): Dict<VgTransform[]> {
         });
       }
       // FIXME: current merging logic can produce redundant transforms when a field is binned for color and for non-color
```
kanitw: Outdated FIXME -- You are fixing this issue, so you should remove FIXME from the code.

src/compile/data/bin.ts
kanitw: Currently this is  extended with `extent` below.

Also, this spread code is poorly formatted. 

This is way better:

```
let binTrans: VgTransform = {
   type: bin, 
   field: fieldDef.field,
   as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})], 
   ...(isBoolean(bin) ? {} : bin}
}
```

You can also make it const 

```
let binTrans: VgTransform = {
   type: bin, 
   field: fieldDef.field,
   as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})], 
   ...(isBoolean(bin) ? {} : bin},
   // add extent if it's not specified
   ...(isBoolean(bin) || bin.extent === undefined ? {} : {extent: {signal: extentSignal}})
}
```

This would require minor code reordering and revision (hint: revise `if (!binTrans.extent) {` below), but this should be very simple to do.  I'm gonna leave that for you to figure out as it should be a good exercise for you.

src/compile/data/bin.ts
```diff
@@ -19,36 +19,42 @@ function numberFormatExpr(expr: string, format: string) {
 function parse(model: Model): Dict<VgTransform[]> {
   return model.reduceFieldDef(function(binComponent: Dict<VgTransform[]>, fieldDef: FieldDef, channel: Channel) {
     const bin = model.fieldDef(channel).bin;
+    // Make this bin variable always an object
```
kanitw: The comment here is obviously left from our prior discussion. 

Your code is still incorrect when you apply bin to two channels, and each of them gets different auto `maxbins` parameters.  In that case, your code should produce two separate bin transform since the parameter is different. 

You should add a new test case to test this case.

kanitw:
Also, look at the codecov diff to make sure that you add test for all the cases.

src/compile/data/bin.ts
```diff
@@ -18,37 +18,41 @@ function numberFormatExpr(expr: string, format: string) {
 
 function parse(model: Model): Dict<VgTransform[]> {
   return model.reduceFieldDef(function(binComponent: Dict<VgTransform[]>, fieldDef: FieldDef, channel: Channel) {
-    const bin = model.fieldDef(channel).bin;
+    const fieldDefBin = model.fieldDef(channel).bin;
+    const bin: Bin = isBoolean(fieldDefBin) ? {} : fieldDefBin;
+    if (bin && !(<VgTransform>bin).maxbins && !(<VgTransform>bin).steps) {
```
kanitw: 1) You have to check if bin is defined or not first.  If not there is no point processing the rest. 

2) These casting `<VgTransform>` seem totally out of place.

src/compile/data/bin.ts
```diff
@@ -18,37 +18,41 @@ function numberFormatExpr(expr: string, format: string) {
...
+      (<VgTransform>bin).maxbins = autoMaxBins(channel);
+    }
+    // Make this bin variable always an object
```
kanitw: out of place comment

src/compile/data/bin.ts
```diff
@@ -18,37 +18,41 @@ function numberFormatExpr(expr: string, format: string) {
...
-          field: fieldDef.field,
-          signal: extentSignal
-        });
-
-        binTrans.extent = {signal: extentSignal};
+      const key = hash(bin) + '_' + fieldDef.field;
+      if (binComponent[key]) {
+        transform.push.apply(transform, binComponent[key]); // concat in place
+      } else {
+        const checkExtentAbsent: boolean = !bin.extent;
```
kanitw: This variable makes it harder to read the code, just simply remove it and use `!bin.extent`

src/compile/data/bin.ts
```diff
@@ -18,37 +18,41 @@ function numberFormatExpr(expr: string, format: string) {
...
-        transform.push({
-          type: 'extent',
-          field: fieldDef.field,
-          signal: extentSignal
-        });
-
-        binTrans.extent = {signal: extentSignal};
+      const key = hash(bin) + '_' + fieldDef.field;
+      if (binComponent[key]) {
+        transform.push.apply(transform, binComponent[key]); // concat in place
```
kanitw: this line seems unnecessary convoluted

kanitw: If binComponent already has that key, I don't think you should do anything to transform here.   In fact, I would put `transform` inside the else case and no longer do `binComponent[key] = transform;`  at the end.

src/compile/data/bin.ts
```diff
@@ -16,57 +16,62 @@ function numberFormatExpr(expr: string, format: string) {
...
-        const extentSignal = model.getName(fieldDef.field + '_extent');
-        transform.push({
-          type: 'extent',
-          field: fieldDef.field,
-          signal: extentSignal
-        });
-
-        binTrans.extent = {signal: extentSignal};
-      }
+function checkDiscreteDomainOrLegend(model: Model, transform: VgTransform[], fieldDef: FieldDef, channel: Channel) {
```
kanitw: `addRangeFormulaForDiscreteDomainOrLegend` or `addRangeFormula` would be a more descriptive name for this method.

src/compile/data/bin.ts
```diff
@@ -16,57 +16,62 @@ function numberFormatExpr(expr: string, format: string) {
...
-        transform.push({
-          type: 'formula',
-          as: field(fieldDef, {binSuffix: 'range'}),
-          expr: `${numberFormatExpr(startField, format)} + ' - ' + ${numberFormatExpr(endField, format)}`
-        });
+function parse(model: Model): Dict<VgTransform[]> {
+  return model.reduceFieldDef(function(binComponent: Dict<VgTransform[]>, fieldDef: FieldDef, channel: Channel) {
+    const fieldDefBin = model.fieldDef(channel).bin;
+    if (fieldDefBin) {
+      // Make this bin variable always an object
```
kanitw: Remove these two lines of comments as it was meant to be a TODO for you.

src/compile/data/bin.ts
```diff
@@ -16,57 +16,62 @@ function numberFormatExpr(expr: string, format: string) {
...
+        if (!bin.extent) {
+          transform.push({
+            type: 'extent',
+            field: fieldDef.field,
+            signal: extentSignal
+          });
+        }
+        transform.push(binTrans);
+
+        checkDiscreteDomainOrLegend(model, transform, fieldDef, channel);
```
kanitw: You can refactor code such that you only need to call this method once at the end.

src/compile/data/bin.ts
```diff
@@ -16,57 +16,62 @@ function numberFormatExpr(expr: string, format: string) {
...
-          type: 'extent',
-          field: fieldDef.field,
-          signal: extentSignal
-        });
-
-        binTrans.extent = {signal: extentSignal};
-      }
+function checkDiscreteDomainOrLegend(model: Model, transform: VgTransform[], fieldDef: FieldDef, channel: Channel) {
+  const hasDiscreteDomainOrHasLegend = hasDiscreteDomain(model.scale(channel).type) || model.legend(channel);
+  if (hasDiscreteDomainOrHasLegend) {
```
kanitw: You haven't checked if the range formula has already been added.  

Hint: if it's added, it's always the last item in the array.

test/compile/data/bin.test.ts
```diff
@@ -49,14 +49,14 @@ describe('compile/data/bin', function() {
         assert.deepEqual(transform[0], {
           type: 'extent',
           field: 'Acceleration',
-          signal: 'Acceleration_extent'
+          signal: '{"maxbins":10}_Acceleration_extent'
```
domoritz: is this a valid signal name?

kanitw: Nope.  Valid signal name should be a valid javascript variable name.

kanitw:
@NeelMohapatra Please address the signal name issue and add tests for others and commit all of them at once.

kanitw:
- [ ] There is actually another problem when the same field has multiple bin parameters.  The output data still uses the same `bin` field.   This is definitely important to make this work, but I'm also okay if you wanna separate this issue into another PR.

kanitw:
Before tomorrow, focus on the fixing the signal and make test pass.  

Then think about how would you solve the problem I've just mentioned: 

> There is actually another problem when the same field has multiple bin parameters. The output data still uses the same bin field. This is definitely important to make this work, but I'm also okay if you wanna separate this issue into another PR.

We can discuss about how you should fix this part tomorrow.

src/fielddef.ts
```diff
@@ -147,7 +147,8 @@ export function field(fieldDef: FieldDef, opt: FieldRefOption = {}) {
 
     if (!opt.nofn) {
       if (fieldDef.bin) {
-        fn = 'bin';
+        const bin: Bin = isBoolean(fieldDef.bin) ? {} : {...fieldDef.bin};
```
kanitw: You forgot to handle the case that we automatically add autoCount to bin.  (Extract that part of that to binToString too)

src/bin.ts
```diff
@@ -39,6 +39,15 @@ export interface Bin {
    */
   maxbins?: number;
 }
+export function binToString(bin: Bin) {
```
kanitw: rename to `binParamsString` (more descriptive)?

src/fielddef.ts
kanitw: Actually, you need a separate method for that to make it correct inside transform.  Let's make a `getBinDef(bin: boolean | Bin): Bin`.

NeelMohapatra: Yeah, you're right. I'm taking notes from now on

domoritz:
stacked_area_binned

Master
screen shot 2017-04-06 at 22 34 09

Your branch
screen shot 2017-04-06 at 22 34 30

NeelMohapatra:
Ouch. Is it from the example vega spec?

domoritz:
No, this is an example Vega-Lite spec. Looks like something with the order got messed up.

src/fielddef.ts
```diff
@@ -250,6 +252,11 @@ export function defaultType(fieldDef: FieldDef, channel: Channel): Type {
 export function normalize(fieldDef: ChannelDef, channel: Channel) {
   // If a fieldDef contains a field, we need type.
   if (isFieldDef(fieldDef)) { // TODO: or datum
+    // TODO: swojit - you can normalize bin here
```
kanitw: When you finish some TODO, remove the TODO

src/fielddef.ts
```diff
@@ -250,6 +252,11 @@ export function defaultType(fieldDef: FieldDef, channel: Channel): Type {
...
+    if (!!fieldDef.bin) {
+      fieldDef.bin = isBoolean(fieldDef.bin) ? {maxbins: 10} : fieldDef.bin;
```
kanitw: This is causing side effect. Below, I already modified this method to cause no side effect to provide an example for you.  Please follow that.

kanitw: (Actually, there is one case I forgot to change it to prevent side effect, please fix that too.)

src/fielddef.ts
```diff
@@ -250,6 +252,14 @@ export function defaultType(fieldDef: FieldDef, channel: Channel): Type {
 export function normalize(fieldDef: ChannelDef, channel: Channel) {
   // If a fieldDef contains a field, we need type.
   if (isFieldDef(fieldDef)) { // TODO: or datum
+    // TODO: swojit - you can normalize bin here
+    if (fieldDef.bin) {
+      fieldDef.bin = isBoolean(fieldDef.bin) ? {maxbins: 10} : fieldDef.bin;
```
kanitw: This is still causing side effect to the existing bin object.  As I show below for type you should do something like:
```
fieldDef = {
   ...fieldDef,
   bin: {
    maxbins: 10,
    ...(isBoolean(fieldDef.bin) ? {} : fieldDef.bin)
   }
}
```

src/compile/data/bin.ts
```diff
@@ -16,60 +16,59 @@ function numberFormatExpr(expr: string, format: string) {
...
+        binComponent[key] = transform = [];
+        const extentSignal = model.getName(key + '_extent');
+
+        let binTrans: VgBinTransform = {
+            type: 'bin',
+            field: fieldDef.field,
+            as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})],
+            signal: varName(model.getName(key + '_bins'))
+        };
+        extend(binTrans, bin); // Has side effect. TODO: Try to remove them
```
kanitw: As mentioned on slack please just avoid side effect
```
let binTrans: ... = extend({ 
  ...
}, bin);
```

src/compile/data/bin.ts
```diff
@@ -16,60 +16,59 @@ function numberFormatExpr(expr: string, format: string) {
...
-        as: [field(fieldDef, {binSuffix: 'start'}), field(fieldDef, {binSuffix: 'end'})],
-        signal: varName(model.getName(fieldDef.field + '_bins'))
-      };
-
-      if (!isBoolean(bin)) {
-         binTrans = extend(binTrans, bin);
+    const fieldDefBin = model.fieldDef(channel).bin;
+    if (fieldDefBin) {
+      const bin: Bin = isBoolean(fieldDefBin) ? {} : fieldDefBin;
+      if (!bin.maxbins && !bin.steps) {
```
kanitw: If we already initialize this in unit, why do we have to augment the value again here?

src/fielddef.ts
```diff
@@ -250,6 +252,13 @@ export function defaultType(fieldDef: FieldDef, channel: Channel): Type {
 export function normalize(fieldDef: ChannelDef, channel: Channel) {
   // If a fieldDef contains a field, we need type.
   if (isFieldDef(fieldDef)) { // TODO: or datum
+    if (fieldDef.bin) {
+      fieldDef.bin = isBoolean(fieldDef.bin) ? {maxbins: 10} : fieldDef.bin;
```
kanitw: 1) As I have commented twice ([1](https://github.com/vega/vega-lite/pull/1983#discussion_r110508590) and [2](https://github.com/vega/vega-lite/pull/1983#discussion_r110524611)), please don't cause side effect!

2) Why don't you bring back `autoMaxBins` and call it here?  Now we got channel in the context and thus we don't really have to remove it. 

3) If you have a convincing reason why you should not bring  `autoMaxBins` back, at least use the default `maxbins` config here.  (If you use the old `autoMaxBins`, just get rid of the `maxbins` config for now.

src/fielddef.ts
```diff
@@ -250,6 +252,13 @@ export function defaultType(fieldDef: FieldDef, channel: Channel): Type {
...
+      if (!fieldDef.bin.maxbins) {
```
kanitw: Incorrect logic.  If there is `step`, no need to insert maxbins.

test/compile/data/bin.test.ts
```diff
@@ -50,18 +50,60 @@ describe('compile/data/bin', function() {
...
+          as: ["bin_maxbins_10_Acceleration_start", "bin_maxbins_10_Acceleration_end"],
           maxbins: 10,
-          signal: "Acceleration_bins",
-          extent: {signal: 'Acceleration_extent'}
+          signal: "bin_maxbins_10_Acceleration_bins",
+          extent: {signal: 'bin_maxbins_10_Acceleration_extent'}
         });
       });
     });
+    describe('should bin the same field', () => {
```
kanitw: why using `describe` here?  This definitely looks like an `it`.

Also please update description to be clearer (like `it('should ... when two channels encode the same binned field', `)

test/compile/data/bin.test.ts
```diff
@@ -50,18 +50,60 @@ describe('compile/data/bin', function() {
...
+            type: "quantitative"
+          },
+          size: {
+            aggregate: "count",
+            field: "*",
+            type: "q"
+          }
+        }
+      });
+      const transform = vals(bin.parseUnit(model))[0];
```
kanitw: You test here doesn't even check the field that is used twice!  

Make sure to (1) check that and (2) check that the output bin component only has two items (not three!)

NeelMohapatra: Can it be done only by checking the transform? `Rotten Tomato Rating` doesn't appear anywhere in the transform array

kanitw: Why do you wrap `bin.parseUnit(model)` with `vals(...)[0]` (especially the `[0]` part)?

NeelMohapatra: That was by mistake

test/compile/data/bin.test.ts
```diff
@@ -12,53 +12,142 @@ describe('compile/data/bin', function() {
...
+          as: ['bin_maxbins_10_Acceleration_start', 'bin_maxbins_10_Acceleration_end'],
           maxbins: 10,
-          signal: "Acceleration_bins",
-          extent: {signal: 'Acceleration_extent'}
+          signal: 'bin_maxbins_10_Acceleration_bins',
+          extent: {signal: 'bin_maxbins_10_Acceleration_extent'}
+        });
+      });
+    });
+    describe('same field encoded twice', () => {
```
kanitw: As I just commented [earlier today](https://github.com/vega/vega-lite/pull/1983#discussion_r110553917), this should be `it`. 

I know then you wanna complain why are there similarly weird `describe` above.  We actually discussed this in a meeting earlier and I told you that yeah these should be `it` too -- please help change them so you have good example tests to follow. 

To provide an example for you, this one could be
```js
it('should apply bin transform only once for a binned field encoded in multiple channels', 
```

test/compile/data/bin.test.ts
```diff
@@ -12,53 +12,142 @@ describe('compile/data/bin', function() {
...
+            type: 'q'
+          }
+        }
+      });
+      const transformArray = bin.parseUnit(model);
+      const transformArrayVals = vals(transformArray);
+      const imdbTransform = transformArrayVals[0];
+      const rotten10Transform = transformArrayVals[1];
+      const rotten20Transform = transformArrayVals[2];
+      it('should be binned twice', function() {
```
kanitw: This test and its description are a bit confusing.

1) The binned field that is encoded twice should be bin *once*, not twice.
2) The description is confusing partly because the test spec is a bit too complicated. Why do you need to include `x: IMDB_Rating` and `size: count`? I think better use just a minimal spec that demonstrates the thing you want to test.

test/compile/data/bin.test.ts
NeelMohapatra: Ah, I thought we needed both describe and it. My bad. I am on it

kanitw: To remind, `describe` should be for grouping test suites (for namespaces and methods). 
`it` = test cases.

src/fielddef.ts
```diff
@@ -251,6 +251,7 @@ export function defaultType(fieldDef: FieldDef, channel: Channel): Type {
  */
 export function normalize(fieldDef: ChannelDef, channel: Channel) {
   // If a fieldDef contains a field, we need type.
+  fieldDef = JSON.parse(JSON.stringify(fieldDef)); // Clone
```
NeelMohapatra: @kanitw I couldn't figure out a good way to remove side effects without cloning it. If you think there's one, I'll update it

test/compile/data/bin.test.ts
```diff
@@ -45,110 +43,59 @@ describe('compile/data/bin', function() {
...
-          signal: 'bin_maxbins_20_Rotten_Tomatoes_Rating_bins',
-          maxbins: 20,
-          extent: {signal: 'bin_maxbins_20_Rotten_Tomatoes_Rating_extent'}
-        });
-        assert.deepEqual(rotten20Transform[2], {
-          type: 'formula',
-          as: 'bin_maxbins_20_Rotten_Tomatoes_Rating_range',
-          expr: 'format(datum["bin_maxbins_20_Rotten_Tomatoes_Rating_start"], \'s\') + \' - \' + format(datum["bin_maxbins_20_Rotten_Tomatoes_Rating_end"], \'s\')'
-        });
+      const transform = vals(bin.parseUnit(model))[0];
```
kanitw: The test right now doesn't really check if it apply the bin only once.  Make sure to assert the length...

domoritz:
🎉

PR #1984, opened 2017-02-26
Fix site

domoritz:
Use cdnjs for site as we suggest (may change to compile into one file later). Also make sure our examples compile again with latest vega embed and vega. Also fix version file.


domoritz:
Yep, separate issue.

package.json
```diff
@@ -38,7 +38,7 @@
     "start": "nodemon -x 'npm run build:test-gallery' & browser-sync start --server --files 'test-gallery/main.js' --index 'test-gallery/index.html'",
     "poststart": "rm examples/all-examples.json",
     "schema": "typescript-json-schema --required true --noExtraProps true src/spec.ts ExtendedSpec > vega-lite-schema.json",
-    "presite": "tsc && npm run build && bower install && npm run data && npm run build:toc && npm run build:versions",
+    "presite": "tsc && npm run build && npm run data && npm run build:toc && npm run build:versions",
```
kanitw: so now we only rely on NPM?  If so, should we remove all `devDependencies` from bower?

tsconfig.json
```diff
@@ -21,7 +21,8 @@
     "test/**/*.ts",
     "examples/**/*.ts",
     "typings/*.d.ts",
-    "node_modules/@types/**/*.d.ts"
+    "node_modules/@types/**/*.d.ts",
+    "site/static/main.ts"
```
kanitw: FYI, I'm gonna remove site from tsconfig is a subsequent PR because it's totally an overkill that site's JS file has to be inside `build` and require TS.

kanitw:
> Yep, separate issue.

Not really a separate issue.   We could merge this PR first, but without that, site still doesn't work correctly if you don't have image precompiled previously.  (I don't!)

tsconfig.json
domoritz: Okay. And then you would just call tsc for that one file? Makes sense.

package.json
domoritz: We do

domoritz:
Okay, it looks like we are compiling vega into vega-lite. The .js is currently 1MB!

domoritz:
screen shot 2017-02-26 at 21 11 59

domoritz:
Okay, the issue is vega-parser. See https://github.com/vega/vega-lite/issues/1987. Let's ignore that for now and merge this.

_layouts/base.html
```diff
@@ -8,20 +8,20 @@
     <link rel="shortcut icon" href="https://vega.github.io/favicon.ico"/>
     <meta name="description" content="{{ site.description }}">
 
-    <script src="{{site.baseurl}}/bower_components/d3/d3.min.js"></script>
-    <script src="{{site.baseurl}}/bower_components/vega/vega.min.js"></script>
+    <script src="https://d3js.org/d3.v3.min.js"></script>
```
kanitw: @domoritz  this is a bad idea as the website will no longer work offline.  

If you strongly prefer to use cdn for all of these, you should have dev and production configs.

package.json
kanitw: we do what?

_layouts/base.html
domoritz: I just want to get rid of bower. The right way would be to compile a bundle imho but this seemed already an improvement.

PR #1986, opened 2017-02-26
[WIP] Labeling

willium:
Beginning re-implementation of labeling 

 - [ ] Add TextTable Example back

Commit history:
```
begin reactive geometry data sourcing

make label use MarkCompiler instead of deprecated TextCompiler

export textRef for use in label

updated schema to include internalData

rename source to ref for internal data

make name function also a getter

update 'source' for internalData handling

point to reference mark

refactor test

progress

revert channel to orig

add vgLayoutTransform and Label interfaces

add Layout compiler

use text mark to encode, implement new compiler / transform

fix errors

quickfix on vgmark

example + label

schema update

revert to const outside (called once)

inherit more from text mark

add label to channels it supports

revert to not using text's functions

add number to encoding

update schema

fix merge issues

change name to public

fix merge issues

remove space
  
```


domoritz:
Closing as it's not going to get merged.

PR #1989, opened 2017-02-27
Only import event selector parser from vega-parser

### LINKED ISSUE #1987: Including vega-parser imports almost all of vega
Vega-Lite is not almost 1MB large because we include a lot of Vega. Also, we cannot build images because of this. 

https://github.com/vega/vega-lite/pull/1984#issuecomment-282629274

arvind:
Fixes #1987. Uses [Babel](https://babeljs.io/) to transpile the vega-parser files. Some lessons learned after two hours, so documenting here in case this is useful for others:

1. Babelify (the Babel browserify plugin) must be invoked as a global transform (`-g`) in order to work on files in `node_modules`.
2. A Babel "helper" file is necessary for mocha, h/t to [this SO answer](http://stackoverflow.com/questions/35040978/babel-unexpected-token-import-when-running-mocha-tests/37076911#37076911).


test/babel.ts
```diff
@@ -0,0 +1,3 @@
+require('babel-core/register')({
```
domoritz: Can you add a short comment why this is necessary?

arvind: Done.

domoritz:
I'm surprised if this doesn't fail because the `yarn.lock` is not up to date. If it works, yarn is doing odd things.

src/compile/selection/transforms/translate.ts
```diff
@@ -1,4 +1,4 @@
-import {selector as parseSelector} from 'vega-parser';
+import {default as parseSelector} from 'vega-parser/src/parsers/event-selector';
```
kanitw: Btw, I think you can do this for the same import :) 
```
import parseSelector from 'vega-parser/src/parsers/event-selector';
```

domoritz: 9cc6c980

PR #1991, opened 2017-02-28
[WIP] Use npm instead of bower for site

kanitw:
Somehow I can't find a way to override jekyll's `exclude` by including only a few files we need.


_config.yml
```diff
@@ -17,7 +17,9 @@ baseurl: "/vega-lite" # the subpath of your site, e.g. /blog/
 url: "http://vega.github.io" # the base hostname & protocol for your site
 exclude: [bin, coverage, examples/_original, examples/_output, examples/_diff, Gemfile, Gemfile.lock, gulp,
   lib/schema.json, node_modules, scripts, test, typings, package.json, bower.json, src, tsconfig.json, tslint.json,
-  nodemon.json, CONTRIBUTING.md, README.md, LICENSE, bower_components/vega-lite, bower_components/vega/node_modules, bower_components/vega-embed/node_modules]
+  nodemon.json, CONTRIBUTING.md, README.md, LICENSE, bower_components]
+
+include: [node_modules/vega/build/*,node_modules/d3/build/d3.min.js, node_modules/vega-embed/vega-embed.js, node_modules/highlightjs/highlight.pack.js, /node_modules/highlightjs/styles/github.css, /node_modules/octicons/build/font/octicons.css]
```
kanitw: This doesn't work .. basically

domoritz:
This would be a step forward. What are your thoughts on compiling a bundle?

kanitw:
> What are your thoughts on compiling a bundle?

I don't think we need a bundle.  
IMHO, what we need is a dev/production config.  Dev = use local, production = use CDN?  

But if you have good argument for bundling, I'm happy to listen :)

domoritz:
> IMHO, what we need is a dev/production config. Dev = use local, production = use CDN?

How critical is it that we can edit the website when we are offline? I'd be okay with dropping support for that.

kanitw:
> How critical is it that we can edit the website when we are offline? I'd be okay with dropping support for that.

I'm not okay with dropping that. I think it's higher priority that bundle / CDN for me.

PR #2000, opened 2017-03-02
remove labelMaxLength

YuhanLu:
- Remove `labelMaxLength` from VlOnlyAxisBase
- Add `labelLimit`, `titleLimit` to AxisConfig and LegendConfig
- Edit and delete related code

### DIFF
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -3,7 +3,7 @@ import {VgAxisBase, VgAxisConfig, VgAxisEncode} from './vega.schema';
 
 export type AxisOrient = 'top' | 'right' | 'left' | 'bottom';
 
-export interface AxisConfig extends VgAxisConfig, VlOnlyAxisBase {
+export interface AxisConfig extends VgAxisConfig {
   /**
    * Whether month names and weekday names should be abbreviated.
    */
@@ -11,10 +11,10 @@ export interface AxisConfig extends VgAxisConfig, VlOnlyAxisBase {
 }
 
 export const defaultAxisConfig: AxisConfig = {
-  labelMaxLength: 25,
+  labelLimit: 25,
 };
 
-export interface Axis extends VgAxisBase, VlOnlyAxisBase {
+export interface Axis extends VgAxisBase {
   /**
    * The padding, in pixels, between axis and text labels.
    */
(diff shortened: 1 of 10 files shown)


src/vega.schema.ts
```diff
@@ -455,6 +455,11 @@ export interface VgLegendConfig extends VgLegendBase {
   labelFontSize?: number;
 
   /**
+   * Maximum allowed pixel width of axis tick labels.
+   */
+  labelLimit?: number;
```
kanitw: What about title/labelDir and title/labelEllipsis?

src/axis.ts
```diff
@@ -3,18 +3,18 @@ import {VgAxisEncode, VgAxisBase, VgAxisConfig} from './vega.schema';
...
+export interface AxisConfig extends VgAxisConfig {
   /**
    * Whether month names and weekday names should be abbreviated.
    */
   shortTimeLabels?: boolean;
 }
 
 export const defaultAxisConfig: AxisConfig = {
-  labelMaxLength: 25,
+  labelLimit: 25,
```
kanitw: Let's remove this default limit as we are not handling it ourselves any more

src/compile/axis/encode.ts
```diff
@@ -12,11 +12,11 @@ export function labels(model: Model, channel: Channel, labelsSpec: any, def: VgA
   const config = model.config;
 
   // Text
-  if (contains([NOMINAL, ORDINAL], fieldDef.type) && axis.labelMaxLength) {
+  if (contains([NOMINAL, ORDINAL], fieldDef.type) && config.axis.labelLimit) {
```
kanitw: We should no longer handle this ourselves - let Vega do it.

test/compile/unit.test.ts
```diff
@@ -178,7 +164,7 @@ describe('UnitModel', function() {
           x: {field: 'a', type: 'ordinal', axis: {offset: 345}},
           y: {field: 'b', type: 'ordinal'}
         },
-        config: {axis: {labelMaxLength: 123}}
+        config: {axis: {tickWidth: 123}}
```
kanitw: Why does this become tickWidth?

src/compile/facet.ts
```diff
@@ -127,15 +127,7 @@ export class FacetModel extends Model {
...
         const axisSpec = facet[channel].axis;
         if (axisSpec !== false) {
-          let vlOnlyAxisProperties: VlOnlyAxisBase = {};
-          VL_ONLY_AXIS_PROPERTIES.forEach(function(property) {
-            if (config.facet.axis[property] !== undefined) {
-              vlOnlyAxisProperties[property] = config.facet.axis[property];
-            }
-          });
-
           const modelAxis = _axis[channel] = {
```
kanitw: I think we still need to copy config.facet.axis because it does not exist in Vega, but we should handle that in another PR.  Please create an issue for this.

kanitw: Actually let me create one now

src/vega.schema.ts
YuhanLu: I test those config and they don't work. I have asked Jeff on slack and still waiting for his response.

test/compile/unit.test.ts
YuhanLu: it's redundant and I will remove it.

src/compile/axis/encode.ts
YuhanLu: fixed

src/axis.ts
YuhanLu: fixed

src/compile/facet.ts
YuhanLu: Currently `VL_ONLY_AXIS_PROPERTIES` does not have `labelMaxLength`(which was the only property), why do we need to have it back? Do you want to add `VL_ONLY_AXIS_PROPERTIES` back as an empty object or add it back later if we have anything else in the future?

kanitw: We don't need `VL_ONLY_AXIS_PROPERTIES`, but still need to copy `config.facet.axis` as Vega has no facet config.

YuhanLu: Ok, I can do it in another PR

PR #2001, opened 2017-03-02
Errorbar Upper Tick Encoding Fix

mattwchun:
Error bar should only provide x2 or y2 not both to the encoding of the upper tick for the errorbar in compositemark.


domoritz:
@mattwchun In tests you can use `"`. See other tests files for how to disable the linter.

PR #2004, opened 2017-03-02
Update Codecov command to match latest tutorial

kanitw:
https://github.com/istanbuljs/nyc#integrating-with-codecov


kanitw:
Hmm ... it still doesn't run codecov.

kanitw:
@arvind Actually it does not.  I'm just trying randomly to see what causes codecov to break...

arvind:
I feel you!

kanitw:
Doesn't work yet ...

PR #2007, opened 2017-03-03
Allow empty named data source. Fixes #1392

### LINKED ISSUE #1392: Add PlaceholderData interface
(For binding data at runtime) 

To allow sharing across multiple units, maybe allow `name` property?

domoritz:
Does not fully work yet because vega fails to compute layout when there is no data.  I consider that beyond the scope of this pr.

### DIFF
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -3,7 +3,7 @@ import * as log from '../log';
 import {Axis} from '../axis';
 import {Channel, X, COLUMN} from '../channel';
 import {Config, CellConfig} from '../config';
-import {Data, DataSourceType} from '../data';
+import {Data, DataSourceType, isNamedData, SOURCE} from '../data';
 import {reduce, forEach} from '../encoding';
 import {FieldDef, FieldRefOption, field, isFieldDef, ChannelDef} from '../fielddef';
 import {Legend} from '../legend';
@@ -271,6 +271,9 @@ export abstract class Model {
   public abstract channelHasField(channel: Channel): boolean;
 
   public getName(text: string, delimiter: string = '_') {
+    if (this.data && text === SOURCE && isNamedData(this.data)) {
+      return this.data.name;
+    }
     return (this.name ? this.name + delimiter : '') + text;
   }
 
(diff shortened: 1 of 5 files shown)


src/compile/model.ts
```diff
@@ -271,6 +271,9 @@ export abstract class Model {
   public abstract channelHasField(channel: Channel): boolean;
 
   public getName(text: string, delimiter: string = '_') {
+    if (this.data && text === String(SOURCE) && isNamedData(this.data)) {
```
kanitw: Do we really need `String(SOURCE)`?  Can't it just be `SOURCE`?

kanitw:
> Does not fully work yet because vega fails to compute layout when there is no data. I consider that beyond the scope of this pr.

Hmm.  What happens after we bind data?

src/compile/model.ts
domoritz: Fixed

domoritz:
> Hmm. What happens after we bind data?
🤷

kanitw:
Actually, does the spec looks correct for layered plot too? If so, LGTM.

kanitw:
@domoritz  FYI, I rebase and codecov is all good right now awesome!!!!

domoritz:
I tried

Does work:
```js
{
  "data": {
    "name": "foo"
  },
  "mark": "bar",
  "encoding": {
  }
}
```

Does not work (yet)
```js
{
  "data": {
    "name": "foo"
  },
  "layer": [
    {
      "mark": "bar",
      "encoding": {
        "x": {"field": "a","type": "ordinal"},
        "y": {"field": "b","type": "quantitative"}
      }
    },
    {
      "mark": "line",
      "encoding": {
        "x": {"field": "a","type": "ordinal"},
        "y": {"field": "b","type": "quantitative"},
        "color": {"value": "red"}
      }
    }
  ]
}
```

kanitw:
I guess we can merge but leave #1392 open for composition?

kanitw:
Or if you wanna fix it now, fine.  up to you

domoritz:
I added a note in https://github.com/vega/vega-lite/issues/1602

PR #2009, opened 2017-03-03
Use bin scales

### LINKED ISSUE #1892: Use the new bin scales in vega 3
https://github.com/vega/vega/releases/tag/v3.0.0-beta.17 adds new scale types that automatically give us nice axes and legends. We should use them in Vega-Lite. 

screen shot 2017-01-25 at 16 13 14

See examples spec in https://github.com/vega/vega/pull/727

domoritz:
Fixes #1892

```json
{
  "data": {
    "values": [
      {"v": 28}, {"v": 55}, {"v": 43}, {"v": 91}, {"v": 34},
        {"v": 81}, {"v": 53}, {"v": 19}, {"v": 87}, {"v": 38},
        {"v": 52}, {"v": 48}, {"v": 24}, {"v": 49},
        {"v": 87}, {"v": 66}, {"v": 17}, {"v": 27},
        {"v": 68}, {"v": 12}, {"v": 49}, {"v":  4}
    ]
  },
  "transform": {
    "calculate": [{"as": "u", "expr": "datum.v"}, {"as": "w", "expr": "datum.v"}]
  },
  "mark": "bar",
  "encoding": {
    "y": {"field": "*", "aggregate": "count","type": "Q"},
    "x": {"field": "v", "bin": {"maxbins": 5}, "type": "Q"},
    "color": {"field": "u", "bin": {"maxbins": 5}, "type": "Q"},
    "size": {"field": "w", "bin": {"maxbins": 5}, "type": "Q"}
  }
}
```

screen shot 2017-03-05 at 18 21 06

```json
{
  "description": "Demonstrate color legend with binned data.",
  "mark": "point",
  "encoding": {
    "x": {"field": "Acceleration", "type": "quantitative"},
    "color": {"bin": {"maxbins": 5}, "field": "Acceleration", "type": "quantitative"}
  }
}
```

screen shot 2017-03-05 at 18 52 26

```json
{
  "data": {
    "values": [
      {"v": 28}, {"v": 55}, {"v": 43}, {"v": 91}, {"v": 34},
        {"v": 81}, {"v": 53}, {"v": 19}, {"v": 87}, {"v": 38},
        {"v": 52}, {"v": 48}, {"v": 24}, {"v": 49},
        {"v": 87}, {"v": 66}, {"v": 17}, {"v": 27},
        {"v": 68}, {"v": 12}, {"v": 49}, {"v":  4}
    ]
  },
  "transform": {
    "calculate": [{"as": "u", "expr": "datum.v"}, {"as": "w", "expr": "datum.v"}]
  },
  "mark": "bar",
  "encoding": {
    "y": {"field": "*", "aggregate": "count","type": "Q"},
    "x": {"field": "v", "bin": {"maxbins": 5}, "type": "O"}
  }
}
```

screen shot 2017-03-05 at 18 01 54

Make sure work

* [x] Binned colors
* [x] Binned x, y
* [x] Binned size
* [x] Binned opacity
* [x] Legends/Axes for all of the above


src/compile/mark/valueref.ts
```diff
@@ -80,10 +80,15 @@ export function midPoint(channel: Channel, channelDef: ChannelDef, scaleName: st
...
   // TODO: datum support
 
+  // FIXME: walk through the logic here
+
   if (channelDef) {
     /* istanbul ignore else */
     if (isFieldDef(channelDef)) {
       if (hasDiscreteDomain(scale.type)) {
+        if (channelDef.bin) {
+          return fieldRef(channelDef, scaleName, {binSuffix: 'start'});
```
kanitw: Interesting that this is bin-start, but seems like Jeff's example uses start too.

kanitw: FYI, line 98 checks for bin too, so definitely needs some revision.  

I wonder if we should add new scale.hasBinnedDomain as a third case since bin is a mixed between the two.

domoritz:
Good idea

src/compile/mark/valueref.ts
kanitw: Actually, if there should be a third case, maybe we should remove hasContinuousDomain / hasDiscreteDomain and replace them with `domainType` that returns one of the 3 string literals?  That will probably avoid misusing it.

kanitw: Actually looking at relevant code base both in Vega-Lite and CompassQL, I think bin should use be a case of `hasDiscreteDomain`.  After all, binning discretizes the continuous domain anyway.

domoritz: Yep, hence the comment at the top of this function. This only made my example work but is certainly wrong :-P

kanitw:
@NeelMohapatra You should be aware of changes in this PR as you'll have to handle merge conflicts.

src/scale.ts
```diff
@@ -53,12 +54,18 @@ const CONTINUOUS_DOMAIN_INDEX = toSet(CONTINUOUS_DOMAIN_SCALES);
...
 
+const BIN_SCALES_INDEX = toSet(['bin-linear', 'bin-ordinal']);
+
 export const TIME_SCALE_TYPES: ScaleType[] = ['time', 'utc'];
 
 export function hasDiscreteDomain(type: ScaleType): type is 'ordinal' | 'point' | 'band' {
   return type in DISCRETE_DOMAIN_INDEX;
 }
 
+export function isBinScale(type: ScaleType): type is 'bin-linear' | 'bin-ordinal' {
```
kanitw: As mentioned on slack, it's ok to add `isBinScale`.  

However, you should also make `hasContinuousDomain` opposite of `hasDiscreteDomain()` because our logic are written with this assumption.  Otherwise, we will definitely get some other code wrong. 

bin-ordinal = discrete
bin-linear = continuous

kanitw:
- [ ] Please revise `scaleTypeSupportProperty`

src/compile/scale/type.ts
```diff
@@ -16,56 +16,56 @@ export type RangeType = 'continuous' | 'discrete' | 'flexible' | undefined;
  */
 // NOTE: CompassQL uses this method.
 export default function type(
-  specifiedType: ScaleType, type: Type, channel: Channel, timeUnit: TimeUnit, mark: Mark,
+  specifiedType: ScaleType, type: Type, channel: Channel, timeUnit: TimeUnit, binned: boolean, mark: Mark,
```
kanitw: Now we have so many parameters, making it hard to read.

Following https://basarat.gitbooks.io/typescript/docs/tips/functionParameters.html, I wonder if we should group parts from fieldDef like this `fieldDefParts: {bin?: boolean, timeUnit?: TimeUnit  type: Type}`.  This will make the method slightly easier to read plus eliminate many `undefined` in the test.

src/compile/mark/valueref.ts
```diff
@@ -83,18 +75,25 @@ export function midPoint(channel: Channel, channelDef: ChannelDef, scaleName: st
   if (channelDef) {
     /* istanbul ignore else */
     if (isFieldDef(channelDef)) {
+      if (isBinScale(scale.type)) {
+        if (contains(['x', 'y', 'x2', 'y2', 'size'], channel)) {
```
kanitw: Based on [this revision](https://github.com/vega/vega-lite/pull/2009/files#r104353272), we might not want to call `isBinScale` here, but instead add special case inside two branches (discrete/continuous) below.

src/compile/layer.ts
```diff
@@ -212,6 +208,13 @@ export class LayerModel extends Model {
     return [];
   }
 
+  public assembleScales(): VgScale[] {
```
domoritz: It was not necessary because we always merged scales.

src/log.ts
```diff
@@ -176,8 +176,12 @@ export namespace message {
       channel === 'x' ? 'width' : 'height'} is provided.`;
   }
 
-  export function scaleTypeNotWorkWithChannel(channel: Channel, scaleType: ScaleType, newScaleType: ScaleType) {
-    return `Channel ${channel} does not work with ${scaleType} scale. We are using ${newScaleType} scale instead.`;
+  export function cannotOverrideBinScaleType(channel: Channel) {
+    return `Cannot override scale type for binned channel ${channel}.`;
+  }
+
+  export function scaleTypeNotWorkWithChannel(channel: Channel, scaleType: ScaleType) {
```
domoritz: Because we had to compute the right scale twice. You can just look at the vega output.

src/compile/scale/type.ts
```diff
@@ -16,56 +16,56 @@ export type RangeType = 'continuous' | 'discrete' | 'flexible' | undefined;
...
 
   if (!hasScale(channel)) {
     // There is no scale for these channels
     return null;
   }
   if (specifiedType !== undefined) {
-    // Check if explicitly specified scale type is supported by the channel
-    if (supportScaleType(channel, specifiedType)) {
-      return specifiedType;
+    // for binned fields we don't allow overriding the default scale
```
domoritz: Done

src/compile/scale/type.ts
domoritz: Why not just require a filedDef?

src/compile/scale/parse.ts
```diff
@@ -30,18 +29,9 @@ export default function parseScaleComponent(model: Model): Dict<ScaleComponents>
  */
 export function parseScale(model: Model, channel: Channel) {
    if (model.scale(channel)) {
-    const fieldDef = model.fieldDef(channel);
-    const scales: ScaleComponents = {
+    return {
```
domoritz: I was waiting because you said we might now want that. But I agree.

src/compile/mark/valueref.ts
```diff
@@ -83,18 +75,25 @@ export function midPoint(channel: Channel, channelDef: ChannelDef, scaleName: st
...
+          // return the middle point for size and length channels
+          return {
```
domoritz: Method does not exist any more.

src/compile/model.ts
```diff
@@ -356,7 +350,8 @@ export abstract class Model {
    * (DO NOT USE THIS METHOD DURING SCALE PARSING, use model.name() instead)
    */
   public scaleName(this: Model, originalScaleName: Channel|string, parse?: boolean): string {
-    const channel = originalScaleName.replace(BIN_LEGEND_SUFFIX, '').replace(BIN_LEGEND_LABEL_SUFFIX, '');
+    // FIXME?
```
domoritz: Fixed

src/compile/mark/valueref.ts
domoritz: Done

src/compile/layer.ts
```diff
@@ -117,19 +117,14 @@ export class LayerModel extends Model {
       if (true) { // if shared/union scale
         keys(child.component.scales).forEach(function(channel) {
           let childScales: ScaleComponents = child.component.scales[channel];
-          if (!childScales) {
-            // the child does not have any scales so we have nothing to merge
+          const modelScales: ScaleComponents = scaleComponent[channel];
+
+          if (!childScales || isSignalRefDomain(childScales.main.domain) || modelScales && isSignalRefDomain(modelScales.main.domain)) {
```
domoritz: Yes, but they won't be in the current logic. This needs more work so I just made it work for now.

src/scale.ts
domoritz: Done

src/log.ts
kanitw: >  we had to compute the right scale twice

That's not true.  We compute it *once* (but have two separate lines that call it, which is totally okay.)  

>  it requires some expertise to change the scale type anyway.

I disagree.  It lets users know what we automatically do for them, which is important (making the system transparent.)

> You can just look at the vega output.

This option is not always available for users (e.g., in Altair).

src/compile/mark/valueref.ts
```diff
@@ -83,18 +75,22 @@ export function midPoint(channel: Channel, channelDef: ChannelDef, scaleName: st
   if (channelDef) {
     /* istanbul ignore else */
     if (isFieldDef(channelDef)) {
+      if (scale.type === 'bin-linear') {
+        return {
+          signal: `(scale("${scaleName}", ${field(channelDef, {binSuffix: 'start', datum: true})}) + scale("${scaleName}", ${field(channelDef, {binSuffix: 'end', datum: true})}))/2`
```
kanitw: I still feel like the old binMidSignal make it easy to read without the need for comment -- not sure why you wanna remove it.

domoritz:
I guess blue is the new green.

screen shot 2017-03-06 at 11 35 02

src/compile/mark/valueref.ts
domoritz: The problem is that it only works for certain channels and making it a separate function hides the implementation.

src/log.ts
domoritz: I don't care enough. Will revert it.

domoritz:
Added some more tests and addresses your concerns.

src/compile/mark/valueref.ts
kanitw: > it only works for certain channels

Current way doesn't make it better in this aspect.  

>  making it a separate function hides the implementation.

That's better that showing complex code that's hard to read.  Plus, keeping the method provides consistent level of abstraction in this method.

domoritz: My point is that there is no consistent level of abstraction for this method. Anyway, there are bigger fish to fry :-P

kanitw: > bigger fish to fry

Yes, but this takes 1 min to fix and I didn't know why you spend a minute earlier to make this part worse :p

src/compile/scale/type.ts
```diff
@@ -16,56 +16,58 @@ export type RangeType = 'continuous' | 'discrete' | 'flexible' | undefined;
  */
 // NOTE: CompassQL uses this method.
 export default function type(
-  specifiedType: ScaleType, type: Type, channel: Channel, timeUnit: TimeUnit, mark: Mark,
+  specifiedType: ScaleType, channel: Channel, fieldDef: FieldDef, mark: Mark,
   hasTopLevelSize: boolean, specifiedRangeStep: number, scaleConfig: ScaleConfig): ScaleType {
 
+  const defaultScaleType = defaultType(channel, fieldDef, mark, hasTopLevelSize, specifiedRangeStep, scaleConfig);
```
kanitw: It's not really a big deal, but I think it's worth mentioning that my original code only runs defaultType if necessary.  Currently code always run it, so it's a little more wasteful, but I'm okay with this too, since it's a very cheap method to run.

domoritz: Yeah, code readability over perf :-P I'm aware of it.

Issue #1772, opened 2016-12-12
Reconsider rules for using gradient legend with temporal fields with timeUnit


kanitw:
Follow-up from #1771, I think timeUnit like Year, YearMonth, YearMonthDay should still use gradient.

PR #2017, opened 2017-03-06
Legends for color scales should use gradient.

domoritz:


kanitw:
LGTM.  Any chance you get to revise  https://github.com/vega/vega-lite/issues/1772 too?
(Probably worth looking at now with less context switch in the future? -- I'm fine if you wanna ignore this comment though.)

domoritz:
Done.

domoritz:
Resolved conflicts.

src/compile/legend/rules.ts
```diff
@@ -30,8 +31,7 @@ export function type(legend: Legend, fieldDef: FieldDef, channel: Channel) {
   if (legend.type) {
     return legend.type;
   }
-
-  if (channel === COLOR && !fieldDef.bin && !fieldDef.timeUnit && contains([QUANTITATIVE, TEMPORAL], fieldDef.type)) {
+  if (channel === COLOR && !fieldDef.bin && (!fieldDef.timeUnit || containsTimeUnit(fieldDef.timeUnit, 'year')) && contains([QUANTITATIVE, TEMPORAL], fieldDef.type)) {
```
kanitw: Should we use scale type here (scale.type === `time` or `utc`)?  (Instead of checking for time prefix directly)

domoritz:
Done.

src/compile/legend/rules.ts
domoritz: Done and with more tests as well.

src/compile/legend/rules.ts
```diff
@@ -26,12 +27,11 @@ export function values(legend: Legend) {
...
 }
 
-export function type(legend: Legend, fieldDef: FieldDef, channel: Channel) {
+export function type(legend: Legend, timeUnit: TimeUnit, channel: Channel, scaleType: ScaleType) {
   if (legend.type) {
     return legend.type;
   }
-
-  if (channel === COLOR && !fieldDef.bin && !fieldDef.timeUnit && contains([QUANTITATIVE, TEMPORAL], fieldDef.type)) {
+  if (channel === COLOR && (!timeUnit || containsTimeUnit(timeUnit, 'year')) && contains<ScaleType>(['time', 'utc', 'sequential'], scaleType)) {
```
kanitw: I meant replace  `containsTimeUnit(timeUnit, 'year')` with scale type check just for temporal case.  

I think the new logic seems wrong.   You can sensibly use linear, log, pow with quantitative field too.  I think the old `contains([QUANTITATIVE, TEMPORAL], fieldDef.type)` was sensible.

domoritz: Ahh, I see.

PR #2019, opened 2017-03-07
Channel-Bin Compatibility fix

kanitw:


src/fielddef.ts
```diff
@@ -295,7 +295,7 @@ export function channelCompatibility(fieldDef: FieldDef, channel: Channel): {com
     case 'size':
     case 'x2':
     case 'y2':
-      if (isDiscrete(fieldDef)) {
+      if (isDiscrete(fieldDef) && !fieldDef.bin) {
```
domoritz: I guess this could use the scale information but this makes sense as well.

PR #2021, opened 2017-03-07
[WIP] Warns if you format ordinal guides

### LINKED ISSUE #1763: Cannot format row/col
```json
{
  "data": {"values": [{"a": 1.01,"b": 1.01,"m_teps": 859}]},
  "encoding": {
    "column": {"axis": {"format": "d"},"field": "b","type": "O"},
    "row": {"axis": {"format": "d"},"field": "a","type": "O"},
  },
  "mark": "text"
}
```

screen shot 2016-12-05 at 20 40 35

NeelMohapatra:
Fixes #1763

Just throws a console warning if you try to format an ordinal field

### DIFF
--- a/src/fielddef.ts
+++ b/src/fielddef.ts
@@ -144,6 +144,11 @@ export interface TextFieldDef<F> extends FieldDef<F> {
    * The formatting pattern for text value. If not defined, this will be determined automatically.
    */
   format?: string;
+
+  /**
+   * Specifies how a text field def should be formatted
+   */
+  formatType?: 'number' | 'time' | 'utc';
 }
 
 export type ChannelDef<F> = FieldDef<F> | ValueDef<any>;
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,5 +6,6 @@
     "coverage/**/*": true,
     "_site/**/*": true
   },
-  "files.trimTrailingWhitespace": true
+  "files.trimTrailingWhitespace": true,
+  "vsicons.presets.angular": false
 }
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -68,6 +68,11 @@ export interface Axis extends VgAxisBase, Guide {
    * Optional mark definitions for custom axis encoding.
    */
   encode?: VgAxisEncode;
+
+  /**
+   * Specifies how format wshould be formatted for axis
+   */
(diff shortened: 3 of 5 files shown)


src/fielddef.ts
```diff
@@ -265,6 +265,11 @@ export function normalize(fieldDef: ChannelDef, channel: Channel) {
     if (!compatible) {
       log.warn(warning);
     }
+    if (fieldDef.type === 'ordinal') {
+      if ((<PositionFieldDef>fieldDef).axis.format) {
```
domoritz: The `<PositionFieldDef>` should not be necessary. Why does it not work otherwise?

domoritz:
Travis is failing.

src/fielddef.ts
NeelMohapatra: fieldDef is of type FieldDef. FieldDef doesn't have axis property

domoritz: No, it is not. You check `fieldDef.type === 'ordinal'` in the line above.

kanitw: Yeah in general avoid casting because its usually wrong.

This seems like the wrong location to add this warning.  Maybe initAxis in unit.ts and facet.ts.

You should also add test that consumes the warning using log.wrap

NeelMohapatra: Do you mean initAxes?

kanitw: yes

domoritz:
@NeelMohapatra can you finish this?

NeelMohapatra:
Isn't the issue blocked?

domoritz:
By what?

NeelMohapatra:
Ham, the issue is labeled blocked now

kanitw:
@NeelMohapatra Since you're responsible for this issue, I'd expect that you explain why it is blocked (rather than pinging me).  In any case, It is blocked because we think maybe Vega should support this. However, since that is not happening soon, I'd say finish it and we can remove the warning once this becomes feasible.

NeelMohapatra:
My bad, I read "by whom"

kanitw:
Let's support it rather than showing warning as discussed in https://github.com/vega/vega-lite/issues/1763#issuecomment-299562995

kanitw:
@NeelMohapatra  I think you should start a new branch to add the support properly so I'm gonna close this.

PR #2023, opened 2017-03-07
Bin-linear does not support zero, only x and y should use bin middle …

### LINKED ISSUE #2022: Bin-linear should not output zero: true

domoritz:
…signal.

Fixes #2022

### DIFF
--- a/src/compile/mark/valueref.ts
+++ b/src/compile/mark/valueref.ts
@@ -89,7 +89,7 @@ export function midPoint(channel: Channel, channelDef: ChannelDef, scaleName: st
     /* istanbul ignore else */
     if (isFieldDef(channelDef)) {
       if (isBinScale(scale.type)) {
-        // Use middle only for x an y to place marks in the center between start and end of the bin range. 
+        // Use middle only for x an y to place marks in the center between start and end of the bin range.
         if (contains(['x', 'y'], channel)) {
           return binMidSignal(channelDef, scaleName);
         }
--- a/src/compile/mark/valueref.ts
+++ b/src/compile/mark/valueref.ts
@@ -89,6 +89,7 @@ export function midPoint(channel: Channel, channelDef: ChannelDef, scaleName: st
     /* istanbul ignore else */
     if (isFieldDef(channelDef)) {
       if (isBinScale(scale.type)) {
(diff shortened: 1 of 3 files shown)


src/compile/mark/valueref.ts
```diff
@@ -85,9 +88,10 @@ export function midPoint(channel: Channel, channelDef: ChannelDef, scaleName: st
   if (channelDef) {
     /* istanbul ignore else */
     if (isFieldDef(channelDef)) {
-      if (scale.type === 'bin-linear') {
-        return binMidSignal(channelDef, scaleName);
-      } else if (scale.type === 'bin-ordinal') {
+      if (isBinScale(scale.type)) {
+        if (contains(['x', 'y'], channel)) {
```
kanitw: Probably nice to add comment why only x and y?  (I have no idea why.)

domoritz: Done

src/compile/mark/valueref.ts
```diff
@@ -85,9 +88,11 @@ export function midPoint(channel: Channel, channelDef: ChannelDef, scaleName: st
   if (channelDef) {
     /* istanbul ignore else */
     if (isFieldDef(channelDef)) {
-      if (scale.type === 'bin-linear') {
-        return binMidSignal(channelDef, scaleName);
-      } else if (scale.type === 'bin-ordinal') {
+      if (isBinScale(scale.type)) {
+        // Use middle only for x an y to place marks in the center between start and end of the bin range.
```
kanitw: I guess for x and y, it makes sense to use mid.  What is unclear to me is why we don't use mid for others.  For example, using non-mid for size leads to slightly misleading encoding.

domoritz: As we saw today, it is wrong for size to use the mid point because the correspondence to the legends gets lost. Yes, bin does a "floor", not a "round" but if we want to change that, we should do so in Vega.

kanitw: Wanna add that to comment with a bit better description than "as we saw today"? :p

domoritz: Done. Feel free to squash commits.

PR #2030, opened 2017-03-11
Clean up vega-embed and babel

domoritz:


domoritz:
This may be much easier with https://github.com/vega/vega-lite/issues/2036

domoritz:
Removed babel. cc @arvind

PR #2031, opened 2017-03-11
Add varName method and apply it to bin's signal name

### LINKED ISSUE #2029: binning fields with special characters break
For example

```json
{
  "data": { "url": "data/cars.csv" },
  "mark": "bar",
  "encoding": {
    "x": {"field": "city-mpg", "type": "quantitative", "bin": true, "title": "City MPG"},
    "y": {"field": "*", "type": "quantitative", "aggregate": "count"}
  }
}
```

kanitw:
Fix #2029

cc: @NeelMohapatra -- this is a small PR that will slightly affect your PR.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -185,3 +185,14 @@ export function duplicate<T>(obj: T): T {
 export function isBoolean(b: any): b is boolean {
   return b === true || b === false;
 }
+
+/**
+ * Convert a string into a valid variable name
+ */
+export function varName(s: string) {
+  // Replace non-alphanumeric characters (anything besides a-zA-Z0-9_) with _
+  const alphanumericS = s.replace(/\W/g, '_');
+
+  // Add _ if the string has leading numbers.
+  return (s.match(/^\d+/) ? '_' : '') + alphanumericS;
+}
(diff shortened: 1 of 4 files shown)


src/util.ts
```diff
@@ -185,3 +185,11 @@ export function duplicate<T>(obj: T): T {
 export function isBoolean(b: any): b is boolean {
   return b === true || b === false;
 }
+
+/**
+ * Convert a string into a valid variable name
+ */
+export function varName(a: string) {
+  return a.replace(/^\d+/, '') // remove leading numbers
+          .replace(/\W/g, ''); // remove non-alphanumeric characters (anything besides a-zA-Z0-9_)
```
domoritz: I'd say instead of removing, we should replace with _.

src/util.ts
```diff
@@ -185,3 +185,11 @@ export function duplicate<T>(obj: T): T {
 export function isBoolean(b: any): b is boolean {
   return b === true || b === false;
 }
+
+/**
+ * Convert a string into a valid variable name
+ */
+export function varName(a: string) {
+  return a.replace(/^\d+/, '') // remove leading numbers
```
domoritz: why not just add something in front of a leading number?

kanitw:
Done

PR #2033, opened 2017-03-13
Dom/aggregate

domoritz:
Rename summary to aggregate and make it return a transform instead of a data source


src/compile/data/aggregate.ts
```diff
@@ -144,34 +143,23 @@ export namespace summary {
...
-            t.ops.push(op);
-          }
-          return t;
-        }, {
-          type: 'aggregate',
-          groupby: groupby,
-          fields: [],
-          ops: []
+  export function assemble(components: SummaryComponent[]): VgAggregateTransform [] {
+    return components.filter(comp => keys(comp.measures).length > 0).map(summaryComponent => {
```
kanitw: Not sure why you like filter+map / forEach better than reduce.  
I think reduce is more concise and eliminates the need for let/side effect.

In any case, please add line break before `.map` so this line isn't too long.

src/vega.schema.ts
```diff
@@ -148,6 +148,15 @@ export interface VgFormulaTransform {
   expr: string;
 }
 
+export interface VgAggregateTransform {
+  type: 'aggregate';
+  groupby: VgFieldRef[];
+  fields: VgFieldRef[];
+  ops: string[];
+  as?: string[];
+  drop?: boolean;
```
kanitw: Side Note for future:

This property is interesting.  We should think about a way to specify this.
For `summarize` transform in `transform`, we can just add `drop` property.

For channelDef's inline transform, this is trickier though.
(We could add this to config, but should also consider non-config approach.

src/compile/data/aggregate.ts
domoritz: It's a reduce over objects and we only used it here. I'll break it down.

PR #2037, opened 2017-03-17
[WIP] Boxplot 

mattwchun:
Refers to #319 compassql issue.

Will create normalizer to make compositemark boxplots. Includes two versions of boxplot where the upper and lower ticks are:
1) max and min
2) IQR*1.5 + Q3 and Q1 - IQR*1.5


kanitw:
Reminder:

Since the BoxPlot with IQR isn't possible until we refactor the transform, you should just include the spec for boxplot with min/max whisker and write normalizer to produce boxplot with min/max whisker.  

The IQR version should be done later -- otherwise you won't be able to make the test pass.

src/compositemark.ts
```diff
@@ -4,6 +4,11 @@ import {GenericUnitSpec, LayerSpec} from './spec';
 
 export const ERRORBAR: 'error-bar' = 'error-bar';
 export type ERRORBAR = typeof ERRORBAR;
+// TODO: export BOXPLOT (IQR ticks)
+// export const BOXPLOT: 'boxplot' = 'boxplot';
+// export type BOXPLOT = typeof BOXPLOT;
+export const BOXPLOT_MINMAX: 'boxplot-minmax' = 'boxplot-minmax';
```
kanitw: Just call this `BOX = 'box'`

src/compositemark.ts
```diff
@@ -72,3 +77,68 @@ add(ERRORBAR, (spec: GenericUnitSpec<ERRORBAR, Encoding>): LayerSpec => {
     ]
   };
 });
+
+add(BOX, (spec: GenericUnitSpec<ERRORBAR, Encoding>): LayerSpec => {
```
kanitw: Also, why `ERRORBAR` in `GenericUnitSpec<ERRORBAR, Encoding>` ?  :)

examples/specs/box_plot_minmax.vl.json
```diff
@@ -1,6 +1,6 @@
 {
```
kanitw: I think you would want to rename this fix to

examples/specs/box_plot_minmax_full.vl.json

and create a new 

examples/specs/box_plot_minmax_short.vl.json

using your new syntax :)

src/compositemark.ts
```diff
@@ -72,3 +79,175 @@ add(ERRORBAR, (spec: GenericUnitSpec<ERRORBAR, Encoding>): LayerSpec => {
...
+
+add(BOX, (spec: GenericUnitSpec<BOX, Encoding>): LayerSpec => {
+  const {mark: _m, encoding: encoding, ...outerSpec} = spec;
+  const {x: _x, y: _y, ...nonPositionEncoding} = encoding;
+
+  let fullEncoding: any = {};
+
+  let isVertical: boolean;
+
+  if (isDiscrete(encoding.x) && isContinuous(encoding.y)) {
```
kanitw: The code is repeated twice for vertical and horizontal.  

(1) This is hard to read and review.  
(2) This might cause problems in the future when we want to extend box plot syntax.  (We will be forced to write every single logic twice, which makes it very prone for error.) 

I'm sure you can write this with half of the code plus a bit more logic checking.

test/compositemark.test.ts
```diff
@@ -98,4 +98,349 @@ describe("normalizeErrorBar", () => {
...
+            "axis": {"title": "age"}
+          },
+          "size": {"value": 5},
+          "color": {"value" : "white"}
+        }
+      });
+    }, Error, 'Need one continuous and one discrete axis');
+  });
+
+  it("should produce correct layered specs for vertical boxplot with min and max lower and upper ticks", () => {
```
kanitw: Please add tests for 1D boxplot (with only y and only x)

src/compositemark.ts
```diff
@@ -72,3 +79,175 @@ add(ERRORBAR, (spec: GenericUnitSpec<ERRORBAR, Encoding>): LayerSpec => {
...
+    // vertical
+    isVertical = true;
+    const fieldDefEncodingY: PositionFieldDef = encoding.y;
+    const encodingYWithoutAxis = {
```
kanitw: `yFieldDef`

src/compositemark.ts
```diff
@@ -72,3 +79,175 @@ add(ERRORBAR, (spec: GenericUnitSpec<ERRORBAR, Encoding>): LayerSpec => {
...
+  const {x: _x, y: _y, ...nonPositionEncoding} = encoding;
+
+  let fullEncoding: any = {};
+
+  let isVertical: boolean;
+
+  if (isDiscrete(encoding.x) && isContinuous(encoding.y)) {
+    // vertical
+    isVertical = true;
+    const fieldDefEncodingY: PositionFieldDef = encoding.y;
```
kanitw: `yChannelDef`

src/compositemark.ts
kanitw: Tip: you can use the [computed properties syntax](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#computed-properties) do something like 

```ts
return {
  ...outerSpec,
  layer: [{
    mark: 'rule',
    encoding: {
       [discreteAxisChannel]: <...some code...>,
       [continuousAxisChannel]: <...some code..>,
    },
    <...more code...>
  ]
}

```

src/compositemark.ts
```diff
@@ -72,3 +79,175 @@ add(ERRORBAR, (spec: GenericUnitSpec<ERRORBAR, Encoding>): LayerSpec => {
     ]
   };
 });
+
+add(BOX, (spec: GenericUnitSpec<BOX, Encoding>): LayerSpec => {
+  const {mark: _m, encoding: encoding, ...outerSpec} = spec;
+  const {x: _x, y: _y, ...nonPositionEncoding} = encoding;
+
+  let fullEncoding: any = {};
```
kanitw: This name is kind of weird since it's not really the `fullEncoding`. It's rather a key-value list of different fieldDefs you want to reuse.  
I'd say get rid of it and instead declare a variable for each encoding mixins (not fieldDef) that you *need* to re-use.  

In fact, with [this suggestion](https://github.com/vega/vega-lite/pull/2037/files#r107746245), you won't really need that many variables pre-declared.  

For the ones that you still need, avoid using `let` if you can reasonably use `const`.  
(`let` is okay sometimes, but whenever `const` seems reasonable -- never use `let`.)

src/compositemark.ts
```diff
@@ -72,3 +79,175 @@ add(ERRORBAR, (spec: GenericUnitSpec<ERRORBAR, Encoding>): LayerSpec => {
...
+  if (isDiscrete(encoding.x) && isContinuous(encoding.y)) {
+    // vertical
+    isVertical = true;
+    const fieldDefEncodingY: PositionFieldDef = encoding.y;
+    const encodingYWithoutAxis = {
+      field: fieldDefEncodingY.field,
+      type: fieldDefEncodingY.type
+    };
+
+    fullEncoding.noAggregate = encoding.x;
```
kanitw: fullEncoding.noAggregate => discreteAxisFieldDef

test/compositemark.test.ts
mattwchun: Is what you're looking for:
`it("should produce correct layered specs for horizontal with no nonpositional encoding properties boxplot with min and max lower and upper ticks",...)`
which is the last test in compositemark.test.ts?

kanitw: That one has both x and y.  I'm looking for two more: one with only x, one with only y.

kanitw: Like the top one in this page: http://www.physics.csbsju.edu/stats/box2.html.  There is no categories :)

src/compositemark.ts
```diff
@@ -81,36 +81,44 @@ add(BOX, (spec: GenericUnitSpec<BOX, Encoding>): LayerSpec => {
...
+    throw new Error('Need a continuous axis for 1D boxplots');
+  }
 
-  } else if (isDiscrete(encoding.y) && isContinuous(encoding.x)) {
-    // horizontal
-    discreteAxis = 'y';
-    continuousAxis = 'x';
-    continuousAxisChannelDef = encoding.x;
-    continuousFieldType = {
+  const continuousFieldType = {
```
kanitw: baseContinuousFieldDef ?  This is a fieldDef, after all.

src/compositemark.ts
```diff
@@ -137,43 +145,45 @@ add(BOX, (spec: GenericUnitSpec<BOX, Encoding>): LayerSpec => {
     ...continuousFieldType
   };
 
+  const discreteAxisChannelDef = discreteAxisFieldDef !== undefined ? {[discreteAxis]: discreteAxisFieldDef} : {};
```
kanitw: A channelDef is a super class of a fieldDef.
Rename this to `discreteAxisEncodingMixin`?

examples/specs/box_plot_minmax_full.vl.json
```diff
@@ -0,0 +1,88 @@
...
+              "type": "quantitative",
+              "axis": {"title": "population"}
+            },
+            "y2": {
+              "aggregate": "max",
+              "field": "people",
+              "type": "quantitative"
+            },
+            "size": {"value": 5},
+            "color": {"value" : "white"}
```
kanitw: Why is everything white?

examples/specs/box_plot_minmax_short.vl.json
```diff
@@ -0,0 +1,16 @@
...
+  "mark": "box",
+  "encoding": {
+    "x": {"field": "age","type": "ordinal"},
+    "y": {
+      "field": "people",
+      "type": "quantitative",
+      "axis": {"title": "population"}
+    },
+    "size": {"value": 5},
+    "color": {"value" : "white"}
```
kanitw: why white?

examples/specs/box_plot_minmax_full.vl.json
mattwchun: Just to make sure, only the bar marks should have the color property in their encoding?

kanitw: It depends, but more importantly --  why do you make plot white?  
(Basically, if things are white, they are not visible.)

We might want to introduce a mark definition syntax for parts of box marks. 

```
mark: {
  type: 'box',
  upperWhisker: {
     ... // MarkConfig 
  },
  lowerWhisker/whiskers/lowerBox/upperBox/boxes/middleTick/upperTick/lowerTick: ...
} 
```

kanitw: Maybe it's easier to discuss this a bit in person

kanitw:
Actually, talking about MarkDef reminds me that you should also output `role` in the output normalized specs (as I cc you in https://github.com/vega/vega-lite/pull/2039)

```
mark: {
  type: 'bar',
  role: 'upperBox'
}
```

You can use the following names: upper/lowerWhisker, lower/upperBox, upper/lowerTick.
You might also want to add optional `middleTick` (Guess why? .. this is related to the question I have above.)

examples/specs/box_plot_minmax_full.vl.json
kanitw: Actually, maybe it's not that hard to explain. 

Basically, with this code: 

```
mark: {
  type: 'box',
  upperBox: {
     color: "skyblue"
  }
},
encoding: {
   ...
   color: {value: 'blue'}
} 
```
The output normalized spec should color everything `"blue"` except the upperBox that has `"skyBlue"`.   

But you will need to loop over all possible mark config properties. 
(Feel free to create a `MARK_CONFIG_PROPERTIES: {keyof MarkConfig)[]` array in `mark.ts`)

kanitw:
(I'm actually wondering if 

`boxUpper/Lower,  boxWhiskerUpper/Lower, boxTickUpper/TickLower` are better names for both roles and 
boxUpper/Lower, whiskerUpper/Lower, tickUpper/Lower/Middle for mark def properties -- the motivation is that all of box related roles will have `box` prefix.

src/compositemark.ts
```diff
@@ -72,3 +76,120 @@ add(ERRORBAR, (spec: GenericUnitSpec<ERRORBAR, Encoding>): LayerSpec => {
...
+      {
+        mark: 'rule',
+        encoding: {
+          ...discreteAxisChannelDef,
+          [continuousAxis]: minWithAxisFieldDef,
+          [continuousAxis + '2']: maxFieldDef,
+          ...nonPositionEncoding
+        }
+      },{ // Lower tick
+        mark: 'tick',
```
kanitw:
```
mark: {
   type: 'tick',
   role: 'boxTickUpper'
}
```

so users can customize their theme using `config.boxTickUpper.*`

src/compositemark.ts
```diff
@@ -72,3 +76,120 @@ add(ERRORBAR, (spec: GenericUnitSpec<ERRORBAR, Encoding>): LayerSpec => {
     ]
   };
 });
+
+add(BOX, (spec: GenericUnitSpec<BOX, Encoding>): LayerSpec => {
```
kanitw: `GenericUnitSpec<BOX | BoxDef, Encoding>`

```
interface BoxDef {
   type: BOX;
   upperWhisker: MarkConfig;
   ...
}
```

kanitw: valueIfDefined might be helpful.

kanitw:
@mattwchun Please be aware of https://github.com/vega/vega-lite/pull/2104/commits/58a0ff01dd534569ba75999c3521638fe393a23a in https://github.com/vega/vega-lite/pull/2104 since we no longer gonna wait for this PR.

src/compositemark/boxplot.ts
```diff
@@ -0,0 +1,136 @@
...
+  /** Size of the box and mid tick of a box plot */
+  size?: number;
+}
+
+export function normalizeBoxPlot(spec: GenericUnitSpec<BOXPLOT, Encoding<Field>>, config: Config): LayerSpec {
+  const {mark: _m, encoding: encoding, ...outerSpec} = spec;
+  const {x: _x, y: _y, ...nonPositionEncoding} = encoding;
+  const {size: size, ...nonPositionEncodingWithoutSize} = nonPositionEncoding;
+  const {color: _color, ...nonPositionEncodingWithoutColorSize} = nonPositionEncodingWithoutSize;
+  const midTickSizeChannelDef = size ? {size: size} : {};
```
kanitw: use the same variable

src/compositemark/boxplot.ts
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  const discreteAxisEncodingMixin = discreteAxisFieldDef !== undefined ? {[discreteAxis]: discreteAxisFieldDef} : {};
+
+  return {
+    ...outerSpec,
+    layer: [
+      { // lower whisker
+        mark: 'rule',
+        encoding: {
```
kanitw: role: 'boxWhisker'

PR #2039, opened 2017-03-20
Allow users to customize role

kanitw:
This will be useful for composite marks.  For example, box plot may use bar to implement upper and lower boxes, but config.bar.* should not be applied to these boxes as they represent semantically different things.

Instead, users should able to use role config such as `config.box, config.upperBox, config.lowerBox, config.whisker, config.upper/lowerWhisker` for customizing parts of composite marks.

cc: @mattwchun

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -218,6 +218,13 @@ export interface Config {
 
   /** Selection Config */
   selection?: SelectionConfig;
+
+  // Support arbitrary key for role config
+  // Note: Technically, the type for role config should be `MarkConfig`.
+  // However, Typescript requires that the index type must be compatible with all other properties.
+  // Basically, it will complain that `legend: LegendConfig` is not an instance of `MarkConfig`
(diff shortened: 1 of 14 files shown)


src/config.ts
```diff
@@ -218,6 +218,12 @@ export interface Config {
 
   /** Selection Config */
   selection?: SelectionConfig;
+
+  // Support arbitrary key for role config
+  // Note: Technically, the type for role config should be MarkConfig.
+  // However, Typescript requires that the index type must be compatible with all other properties.
+  // Thus, we have to use any here.
```
domoritz: What does "Typescript requires that the index type must be compatible with all other properties" mean?

kanitw: just pushed with a bit more description.

domoritz: Why not use a type union?

kanitw: not sure how that makes things better?.

Issue #1820, opened 2017-01-12
Remove unnecessary "as"


kanitw:
Looks like with Typescript 2.1, we should remove unnecessary "as" from our codebase. 

- [ ] `{} as Dict <T>`
- [ ] `as any`
- [ ] Also remove explicit `: any` if possible.

kanitw:
@dbautistav – If I remember correctly, I think I have already fixed this but still have some occurrences of these left behind as we can't remove some of them. 

That said, feel free to check if there are some other places that we can still remove this.  Thanks!

dbautistav:
Ok. Thank you!

Issue #1937, opened 2017-02-10
Turn on "ordered-imports" option in tslint


kanitw:
(and fix errors)

PR #2040, opened 2017-03-21
Turn on "ordered-imports" option in tslint - #1937

dbautistav:
Please:
- [x] ~Provide a test case & update the documentation under `site/docs/`~

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "site": "bundle exec jekyll serve",
 
     "lint": "tslint -c tslint.json 'src/**/*.ts' 'test/**/*.ts'",
-    "lint:fix": "tslint -c tslint.json --fix 'src/**/*.ts' 'test/**/*.ts'",
 
     "test": "tsc && npm run schema && npm run data && npm run test:nocompile",
     "test:nocompile": "npm run test:only && npm run lint && npm run mocha:examples",
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "site": "bundle exec jekyll serve",
 
     "lint": "tslint -c tslint.json 'src/**/*.ts' 'test/**/*.ts'",
+    "lint:fix": "tslint -c tslint.json --fix 'src/**/*.ts' 'test/**/*.ts'",
 
     "test": "tsc && npm run schema && npm run data && npm run test:nocompile",
     "test:nocompile": "npm run test:only && npm run lint && npm run mocha:examples",
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -1,5 +1,5 @@
 import {DateTime} from './datetime';
-import {VgAxisEncode, VgAxisBase, VgAxisConfig} from './vega.schema';
+import {VgAxisBase, VgAxisConfig, VgAxisEncode} from './vega.schema';
 
 export type AxisOrient = 'top' | 'right' | 'left' | 'bottom';
 
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -1,4 +1,4 @@
-import {Channel, ROW, COLUMN, SHAPE, SIZE} from './channel';
+import {Channel, COLUMN, ROW, SHAPE, SIZE} from './channel';
 
 /**
  * Binning properties or boolean flag for determining whether to bin data or not.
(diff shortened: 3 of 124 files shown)


kanitw:
@dbautistav – Thanks! I like this tslint flag, and was thinking that it would be great if someone could do this for us!  Look forward to a finished PR :)

dbautistav:
I've changed the related configuration at `tslint.json` ([lines 68-71](https://github.com/dbautistav/vega-lite/blob/0e82ccb0d64aa8358af35dac96a31adafa0a0cef/tslint.json#L68-L71)) as follows:
```json
    "ordered-imports": [true, {
      "import-sources-order": "case-insensitive",
      "named-imports-order": "case-insensitive"
    }]
```
since it's what makes sense to me. This rule was created mainly to mitigate git conflicts when merging branches but I found the default configuration confusing, so _I prefer_ (and use in current projects) these proposed settings.

- [tslint ordered-imports rule definition](https://palantir.github.io/tslint/rules/ordered-imports/)

Is this the configuration that you were expecting or do you prefer the default?

Please let me know what you prefer (and preferably also why) and I'll refactor the source code and tests to pass the lint test.

dbautistav:
Also I want to know if I can fix #1820 here too since I'm going to traverse almost all files so I can do that in one pass instead of two. Do you agree?

kanitw:
From a first glance, I slightly prefer `"lowercase-last"` but since `"case-insensitive"` is the default, I think let's go with the default.  They are not that different anyway. 

I believe that you can use the tslint's fixer to fix all of them without fixing them manually too. 

One question:

From the documentation here: https://palantir.github.io/tslint/rules/ordered-imports/,
it seems like `"case-insensitive"` is the default.  Thus, I wonder if `"ordered-imports": true` is actually equivalent to the proposed: 

```
    "ordered-imports": [true, {
      "import-sources-order": "case-insensitive",
      "named-imports-order": "case-insensitive"
    }]
```

> Also I want to know if I can fix #1820 here too since I'm going to traverse almost all files so I can do that in one pass instead of two. Do you agree?

To make it easy for us to review, let's fix one thing at the time (esp. if you can use tslint fixer to fix it, this should be easy!)

dbautistav:
- My bad. You are right! The "case-insensitive" option is the default now.

- I'll check the tslint fixer. I was thinking to use my IDE support for this.

- So atomic PR will be.

Thank you @kanitw!

package.json
```diff
@@ -47,6 +47,7 @@
     "site": "bundle exec jekyll serve",
 
     "lint": "tslint -c tslint.json 'src/**/*.ts' 'test/**/*.ts'",
+    "lint:fix": "tslint -c tslint.json --fix 'src/**/*.ts' 'test/**/*.ts'",
```
domoritz: I think you can just run `npm run lint -- --fix` instead of adding a new command.

dbautistav: Yes, you are right! I'll remove the added task.
Thank you @domoritz.

dbautistav:
I run the tests locally and everything seems alright. Also the 'start' task is working as expected (displaying the charts into the browser as before the changes).
And Travis give me the green check too.

dbautistav:
Please:
- [x] ~Provide a test case & update the documentation under `site/docs/`~

package.json
kanitw: Hmm, since our Travis test rely on `lint` script.  

I wonder if this will actually make lint pass on travis even though the source code fails the test?

kanitw:
Thanks -- I had [a comment](https://github.com/vega/vega-lite/pull/2040#discussion_r107068930) about the `--fix`.  Other than that, we should be good to go.

package.json
dbautistav: No because Travis just executes something like `npm run lint` and @domoritz suggestion was to pass the flag _locally_ into the CLI to fix the raised issues when the "ordered-imports" was turned on.

This flag enables tslint to fix (and change) the files and those changes were what I commit (cc0f4cc) and pushed. So as this was manually (and locally) executed, Travis will continue working as expected.

This makes sense to you?

kanitw: Ah I misread it :)

kanitw:
Thanks so much @dbautistav!

dbautistav:
No, thank you guys!

PR #2043, opened 2017-03-23
Differentiate top-level spec from normal spec.

kanitw:
This PR differentiates top-level specs from normal specs.
Basically, normal (internal) specs should not have top-level properties such as `padding`,`$schema`, `config`. etc.

- Add `toplevel.ts` that `spec.ts` and `config.ts` can import (avoid circular dependencies)
- Add new `TopLevelExtendedSpec = TopLevel<ExtendedSpec>` that we will build schema from

- Rewrite how we assemble top-level properties in `compile.ts`
- Remove the old config cascading logic (Config is now always at the top-level and thus there is no cascading anymore -- we can now initialize config once in compile.ts and in our test utility)

- for `unit.ts`, avoid passing around config for the class method as they can just use `this.config`

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -2,13 +2,334 @@
     "$schema": "http://json-schema.org/draft-04/schema#",
     "anyOf": [
         {
-            "$ref": "#/definitions/GenericUnitSpec<string | MarkDef, EncodingWithFacet>"
+            "additionalProperties": false,
+            "properties": {
+                "$schema": {
+                    "description": "URL to JSON schema for this Vega-Lite specification.",
+                    "format": "uri",
+                    "type": "string"
+                },
+                "background": {
+                    "description": "CSS color property to use as background of visualization. Default is `\"transparent\"`.",
+                    "type": "string"
+                },
+                "config": {
+                    "$ref": "#/definitions/Config",
+                    "description": "Configuration object"
+                },
+                "data": {
+                    "anyOf": [
+                        {
+                            "$ref": "#/definitions/UrlData"
+                        },
+                        {
+                            "$ref": "#/definitions/InlineData"
+                        },
+                        {
+                            "$ref": "#/definitions/NamedData"
+                        }
+                    ],
+                    "description": "An object describing the data source"
+                },
+                "description": {
+                    "description": "An optional description of this mark for commenting purpose.\nThis property has no effect on the output visualization.",
+                    "type": "string"
+                },
+                "encoding": {
+                    "$ref": "#/definitions/EncodingWithFacet",
+                    "description": "A key-value mapping between encoding channels and definition of fields."
+                },
+                "height": {
+                    "type": "number"
+                },
+                "mark": {
+                    "anyOf": [
+                        {
+                            "$ref": "#/definitions/MarkDef"
+                        },
+                        {
+                            "type": "string"
+                        }
+                    ],
+                    "description": "The mark type.\nOne of `\"bar\"`, `\"circle\"`, `\"square\"`, `\"tick\"`, `\"line\"`,\n`\"area\"`, `\"point\"`, `\"rule\"`, and `\"text\"`."
+                },
+                "name": {
+                    "description": "Name of the visualization for later reference.",
+                    "type": "string"
+                },
+                "padding": {
+                    "anyOf": [
+                        {
+                            "additionalProperties": false,
+                            "properties": {
+                                "bottom": {
+                                    "type": "number"
+                                },
+                                "left": {
+                                    "type": "number"
+                                },
+                                "right": {
+                                    "type": "number"
+                                },
+                                "top": {
+                                    "type": "number"
+                                }
+                            },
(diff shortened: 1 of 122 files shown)


build/vega-lite-schema.json
```diff
@@ -2,13 +2,79 @@
...
-            "$ref": "#/definitions/GenericUnitSpec<string | MarkDef, EncodingWithFacet>"
+            "allOf": [
+                {
+                    "$ref": "#/definitions/GenericUnitSpec<string | MarkDef, EncodingWithFacet>"
+                },
+                {
+                    "$ref": "#/definitions/TopLevelProperties"
+                },
+                {
+                    "additionalProperties": false,
```
kanitw: @domoritz 

My declaration [in the code here](https://github.com/vega/vega-lite/pull/2043/files#diff-f0bc433010488ea7f9df55b35517b91bR20) seems reasonable. 

However, the output JSON schema actually cause the validation to stupidly fail because each part of the `allOf` has `additionalProperties: false`.  (See error https://github.com/vega/vega-lite/pull/2043/files#diff-f0bc433010488ea7f9df55b35517b91bR20.) 

There are two solutions here:
a) Fix it in the JSON schema generator.  Basically, just like interface `extends` we could just generate a merged object instead of using `allOf`.

b) Don't do `TopLevel<S extends BaseSpec>`.  Instead, for every occurrence of `TopLevel<XXX>`, declare a custom TopLevelXXX that `extends XXX, TopLevelProperties`.

I think a) is a cleaner solution if it's not too hard to do.  b) is obviously a hack and will lead to many tedious and repetitive declarations. 

Let me know if it is possible to do a)

kanitw: (updated comment)

@domoritz 

My declaration [in the code here](https://github.com/vega/vega-lite/pull/2043/files#diff-f0bc433010488ea7f9df55b35517b91bR20) seems reasonable. 

However, the output JSON schema actually cause the validation to stupidly fail because each part of the `allOf` has `additionalProperties: false`.  (See error https://github.com/vega/vega-lite/pull/2043/files#diff-f0bc433010488ea7f9df55b35517b91bR20.) 

There are two solutions here:
a) Fix it in the JSON schema generator.  Basically, just like interface `extends` we could just generate a merged object instead of using `allOf`.

b) Don't do `TopLevel<S extends BaseSpec>`.  Instead, for every occurrence of `TopLevel<XXX>`, declare a custom TopLevelXXX that `extends XXX, TopLevelProperties`.

I think a) is a cleaner solution if it's not too hard to do.  (Currently, `&` will always break the output schema if `additionalProperties` is globally set `false` anyway. 
On the other hand, b) is will lead to many tedious and repetitive declarations. 
I think a) is a cleaner solution if it's not too hard to do.  b) is obviously a hack and will lead to  repetitive declarations (That will add 5-10 more repeated declarations in `spec.ts`). 


I could do b) but I want to check with you if you want to look into a) first.  Let me know!

domoritz:
If you give me a small example where the schema generator is wrong, I can
fix it. It's better to do it properly.

kanitw:
Sure!

`{a:1, b: 2}`  would fail to validate type `AB` where

```
type AB = A & B;
interface A { a: number }
interface B { b: number }
```

given `--noExtraProps` flag.

domoritz:
Fixed in master.

kanitw:
@domoritz @YuhanLu  there are a few specs that somehow fail the validation, wonder if any of you could help trying to see what's broken so we can get this PR in and avoid a lot of conflicts?  I might not have much time this week to clean this up.

examples/vg-specs/overlay_line_short.vg.json
```diff
@@ -146,9 +146,12 @@
                                 "scale": "y",
                                 "field": "price"
                             },
-                            "fill": {
+                            "stroke": {
```
kanitw: Hmm this seems wrong

examples/vg-specs/overlay_line_full.vg.json
```diff
@@ -122,7 +122,7 @@
                                 "scale": "y",
                                 "field": "price"
                             },
-                            "stroke": {
+                            "fill": {
```
kanitw: This too

examples/vg-specs/overlay_area_short.vg.json
```diff
@@ -185,9 +185,12 @@
                                 "scale": "y",
                                 "field": "price"
                             },
-                            "fill": {
+                            "stroke": {
```
kanitw: this too

examples/vg-specs/scatter_connected.vg.json
```diff
@@ -146,9 +146,12 @@
                                 "scale": "y",
                                 "field": "gas"
                             },
-                            "fill": {
+                            "stroke": {
```
kanitw: Need to check

kanitw:
@domoritz, @YuhanLu  I'm gonna merge despite broken overlay specs because those spec should be revised anyway.

PR #2052, opened 2017-03-27
Update package.json

domoritz:

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "test": "test"
   },
   "scripts": {
-    "tsc": "rm -rf build/*/** && tsc",
+    "tsc": "rm -rf build && tsc",
     "prebuild": "mkdir -p build/site build/examples/images build/test-gallery",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "test": "test"
   },
   "scripts": {
-    "tsc": "rm -rf build/*/** && tsc",
+    "tsc": "rm -rf build && tsc",
```
kanitw: I actually don't want to remove files directly under `build/`; I only wanna remove files in `build/`'s subfolders.  

Two reasons:
1) when I run `test`, I shouldn't have to wipe out the bundled files and the schema (i don't rebuild them anyway.) 
2) These files at the top-level of `build/` should never get outdated unless we do major folder restructure (unlike ones in src/ and test/) so it's okay just to remove subfolder like this.

domoritz: This will cause inconsistencies and lead to errors, I'm sure.

kanitw: How?  Tsc generate those subfolders but not top level build so its fair to clean only what is regenerated.

Your proposed change will break linked editor and voyager every time you run test.

domoritz: I see.

Issue #446, opened 2015-05-22
Support UTC Time


kanitw:
- [ ] UTC TimeUnit
- [ ] UTC Scale
  - Add `utc` to the schema and the docs
- [ ] UTC DateTime Object
  - Simply add `utc: boolean` property 
- [ ] Add an example in `scale.md` that uses utc scale.

kanitw:
@domoritz  -- not sure if this is on your radar, but it seems relevant to the stuff that you're doing.

domoritz:
We should not use UTC for the output because dates are parsed using the current time zone. When we add a way to parse dates as UTC, then we should also output them as UTC.

domoritz:
Right now, nothing is parsed or shown in UTC. 

We should add an option to parse and view in UTC but it's not urgent so I'll leave this.

kanitw:
If I understand correctly, you fixed the make the local time stuff working correctly, so the only stuff left is to support UTC as well?

domoritz:
Yes. Modulo some weird js behavior. See https://github.com/vega/datalib/issues/63

kanitw:
Renamed to "Support UTC Time and updated the description above"

peller:
Seems like a major usability issue. Could this be as simple as offering the user a switch and using the `.utc` time scales in d3?

kanitw:
Actually I just checked, you can use `utc` as a scale type too, we just haven't updated the docs.

domoritz:
> Actually I just checked, you can use utc as a scale type too, we just haven't updated the docs.

It's more than that. We also have to make sure that all the `timeUnit` stuff works with utc.

peller:
> Actually I just checked, you can use utc as a scale type too, we just haven't updated the docs.

[Like this](https://jsfiddle.net/p93coymL/)?  I'm getting strange results for the X axis.  Not sure if this is what @domoritz meant.

domoritz:
Yeah, like that. We simply haven't tested it properly. UTC is a new feature that is currently not supported but implementing it should only require a few adjustments.

kanitw:
Yeah, that's mainly why we haven't added it in http://vega.github.io/vega-lite/docs/scale.html#type yet

kanitw:
> We should add an option to parse and view in UTC but it's not urgent so I'll leave this.

Is there really an easy way to parse things in UTC? AFAIK the parse behavior depends on the input data and the specified parse format.  

[...]

PR #2057, opened 2017-03-28
Include output vega examples in the repo. 

### LINKED ISSUE #1598: Better mechanism to notice changes in Vega output
In the past, we skim the example gallery to make sure that we do not break the output.

However, given that the number of example will only keep growing, this approach is no longer sustainable. 

As I mentioned earlier, we might consider adding compiled Vega output to the repo so we can explicitly see diffs when we make changes (and run test that pass only if the compiled examples match the stored output.) 

Other suggestions are welcomed!

kanitw:
I'm gonna create another PR (https://github.com/vega/vega-lite/pull/2058) to demonstrate error.

Fix #1598.


domoritz:
Nice.

kanitw:
@domoritz looks like we found a bit related to timezone as well! (That's why the test is broken right now.)

domoritz:
So another thing to fix in https://github.com/vega/vega-lite/issues/446

Issue #2005, opened 2017-03-03
Transforms as an array


domoritz:
[Dependencies of transforms](https://github.com/vega/vega-lite/wiki/Dependencies) are not obvious. Instead of implicitly encoding dependencies for the first explicit `transform`s (`filter`, `calculate`, `filterInvalid`), we want to make them an array. 

```js
transform: [{
  "filter": {"field": "a", "equal": 5}
}, {
  "filter": {"field": "a", "range": [0,5]}
}, {
  "filter": "datum.b > 10"
},{ // can have multiple filter"
  "filter": {"selection": "<selection-name>"}
}, {
  "calculate": "<formula>",
  "as": "<field_name>"
}]
```

In addition, we plan to add support for transforms that run before encoding. Note that you only need to do this if you want to do for example binning before a filter transform. The syntax for these transforms that are usually in encoding would be something like

```json
{
  "bin": true, "field": "acceleration", "as": "binned-acceleration"
}
```

```json
{
  "summarize": [
    {"aggregate": "sum", "field": "acceleration", "as": "sum_acceleration"},
    {"aggregate": "mean", "field": "horsepower", "as": "mean_horsepower"}
  ],
  "groupby": ["origin", "cylinders"]
}
```

this will be critical for statistical graphics that require post-aggregate transformations.

We can then express the crossfilter example from the paper without the weird "filterWith".

```js
{
  "data": {"url": "data/flights-2k.json"},
  "transform": [
    {"calculate": "hours(datum.date)", "as": "hour"}
    // {"timeUnit": "hours", "field": "date", "as": "hour"}
  ],
  "repeat": {"column": ["hour", "delay", "distance"]},
  "spec": {
    "transform": [
      // This binning process happens before any filtering
      {"bin": {"maxbins": 30}, "field": {"repeat": "column"}, "as": "binned-measure"}
    ],
    "layer": [
      {
        "selection": {
          "region": {
            "type": "interval",
            "project": {"channels": ["x"]},
            "resolve": "intersect"
          },
          "bins": {
            "type": "single",
            "project": {"channels": ["x"]},
            "on": "mousemove",
            "resolve": "union"
          }
        },
        "mark": "bar",
        "encoding": {
          "x": {"field": "binned-measure", "type": "Q"},
          "y": {"aggregate": "count", "field": "*", "type": "quantitative"},
          "color": {"value": "steelblue"}
        }
      },{
        "transform": [
[...]

PR #2064, opened 2017-03-29
Filter and calculate transforms as an array. Part of #2005

domoritz:


src/compile/data/nullfilter.ts
```diff
@@ -60,7 +57,7 @@ export const nullFilter: DataComponentCompiler<Dict<FieldDef>> = {
       const childDataComponent = child.component.data;
       if (model.compatibleSource(child) && !differ<FieldDef>(childDataComponent.nullFilter, nullFilterComponent)) {
         extend(nullFilterComponent, childDataComponent.nullFilter);
-        delete childDataComponent.nullFilter;
+        childDataComponent.nullFilter = {};
```
kanitw: Can you explain why you need to change this line?

src/compile/data/nullfilter.ts
```diff
@@ -17,12 +17,9 @@ const DEFAULT_NULL_FILTERS = {
...
 /** Return Hashset of fields for null filtering (key=field, value = true). */
 function parse(model: Model): Dict<FieldDef> {
-  const filterInvalid = model.filterInvalid();
-
   return model.reduceFieldDef(function(aggregator: Dict<FieldDef>, fieldDef: FieldDef) {
-    if (fieldDef.field !== '*') { // Ignore * for count(*) fields.
-      if (filterInvalid ||
-        (filterInvalid === undefined && fieldDef.field && DEFAULT_NULL_FILTERS[fieldDef.type])) {
+    if (fieldDef.aggregate !== 'count') { // Ignore * for count(*) fields.
+      if (fieldDef.field && DEFAULT_NULL_FILTERS[fieldDef.type]) {
```
kanitw: It seems like we always filterInvalid for now.  If you wanna keep the functionality, at least add it to config to allow disabling?

src/compile/data/source.ts
```diff
@@ -38,16 +37,16 @@ export namespace source {
...
-            {type: formatType ? formatType : defaultExtension},
-            dataFormat.property ? {property: dataFormat.property} : {},
-            // Feature and mesh are two mutually exclusive properties
-            dataFormat.feature ?
-              {feature : dataFormat.feature} :
-            dataFormat.mesh ?
-              {mesh : dataFormat.mesh} :
-              {}
-          );
+        extend(
```
kanitw: the new tabbing is weird. 

How about just using the ... syntax.

const {property, feature, mesh} = dataFormat;

sourceData.format = {
  type: formatType ? formatType : defaultExtension,
  ...(property ? {property} : {}),
  ...(feature ? {feature} : {}),
  ...(mesh ? {mesh} : {}),
}

src/transform.ts
```diff
@@ -1,33 +1,29 @@
...
-  filter?: string | Filter | (string | Filter)[];
-  /**
-   * Whether to filter invalid values (`null` and `NaN`) from the data. By default (`undefined`), only quantitative and temporal fields are filtered. If set to `true`, all data items with null values are filtered. If `false`, all data items are included.
-   */
-  filterInvalid?: boolean;
-  /**
-   * Calculate new field(s) using the provided expresssion(s). Calculation are applied before filter.
-   */
-  calculate?: Formula[];
+  filter: Filter | Filter[];
```
kanitw: I thought you wanna get rid of `Filter[]`.  (This could become a separate PR though.)

src/transform.ts
```diff
@@ -1,33 +1,29 @@
...
+export interface CalculateTransform {
   /**
-   * The field in which to store the computed formula value.
+   * A string containing an expression for the formula. Use the variable `datum` to to refer to the current data object.
    */
-  as: string;
+  calculate: string;
   /**
-   * A string containing an expression for the formula. Use the variable `datum` to to refer to the current data object.
+   * The field in which to store the computed formula value.
```
kanitw: The field for storing the computed formula value.

src/transform.ts
```diff
@@ -1,33 +1,29 @@
...
- */
-export interface Formula {
+export function isFilter(t: Transform): t is FilterTransform {
+  return t['filter'] !== undefined;
+}
+
+export interface CalculateTransform {
   /**
-   * The field in which to store the computed formula value.
+   * A string containing an expression for the formula. Use the variable `datum` to to refer to the current data object.
```
kanitw: containing a Vega Expression formula

src/compile/data/nullfilter.ts
kanitw: Another possibly idea is that we leave them in the array.  
If they use it twice, then they are just stupid.

src/transform.ts
kanitw: I think for Voyager, the `Filter[]` is a little handy though.

src/compile/data/nullfilter.ts
domoritz: We only filter Q and T, which is a sensible default.

src/compile/data/nullfilter.ts
domoritz: I was expecting nullFilter to be always defined. If I don't do this, `Object.keys` breaks.

src/transform.ts
domoritz: It is simple so I left it.

src/compile/data/nullfilter.ts
kanitw: "sensible default" means users should be able to override?

domoritz: Will fix it.

domoritz:
Done.

src/compile/data/nullfilter.ts
domoritz: > "sensible default" means users should be able to override?

Actually, no. You cannot disable the filtering. We always filter nulls when we aggregate.

kanitw: But it's terrible if there is no way to show null values even for raw plots.

domoritz: I see. I thought we didn't support it in the past but we did. Fixed it.

PR #2067, opened 2017-03-31
[WIP] Enable parallel for build:example

### LINKED ISSUE #2060: Parallelize the build:examples script
We now include the output Vega specs in the repo, using the `build:examples` (introduced in #2057).

However, `build:examples` is now single-thread.  We can make it faster using `parallel` as we use in `generate-images.sh`.  However, previously we assume that users of  `generate-images.sh` will manually install the right dependencies. 

Two things we have to guarantee:
- [x] Make sure `npm install` include the right dependency.
- [ ] Make sure travis includes the right dependency.

YuhanLu:
Part of #2060

### DIFF
--- a/scripts/build-examples.sh
+++ b/scripts/build-examples.sh
@@ -16,5 +16,6 @@ then
     bin/vl2vg -p $file > $dir/$name.vg.json
   done
 else
+  echo "yeah we use parallel!"
   ls examples/specs/*.vl.json | parallel  --plus --halt 1 "bin/vl2vg -p {} > examples/vg-specs/{/..}.vg.json"
 fi
--- a/scripts/build-examples.sh
+++ b/scripts/build-examples.sh
@@ -8,8 +8,12 @@ echo "compiling examples to $dir"
 mkdir -p $dir
 rm -f $dir/*
 
-for file in examples/specs/*.vl.json; do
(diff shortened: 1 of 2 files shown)


YuhanLu:
I still need to test if travis use parallel.

YuhanLu:
I don't know why travis doesn't use parallel :(

domoritz:
Maybe we need sudo. I think there is something in the travis docs.

scripts/build-examples.sh
```diff
@@ -5,11 +5,17 @@ dir=${dir-"examples/vg-specs"}
...
+
+if type parallel >/dev/null 2>&1
+then
+  for file in examples/specs/*.vl.json; do
+    filename=$(basename "$file")
+    name="${filename%.vl.json}"
+    bin/vl2vg -p $file > $dir/$name.vg.json
+  done
+else
+  echo "yeah we use parallel!"
```
kanitw: echo "Using parallel to generate vega specs from examples in parallel"

For the other case, let's also say 


echo "Parallel not found! Sequentially generate vega specs from examples."

kanitw:
- [ ] Probably should add a script to postinstall to make sure `parallel` is install for local development too?  (It's probably okay to make the script only work for mac and does not thing for Linux / window and add note to CONTRIBUTING.md.)

PR #2072, opened 2017-04-03
Fix the bug labelAngle cannot be applied when labelAngle == 0

### LINKED ISSUE #1664: Label rotation does not work for row when labelAngle = 0
Please:
- [X] Check for duplicate issues
- [X] Describe how to reproduce the bug /  the goal of the new feature request
Can be reproduced using the online editor:
```json
{
    "data": {"values": [
      {"a": "A","b": 28, "c": "A"}, 
      {"a": "B","b": 55, "c": "A"}, 
      {"a": "C","b": 43, "c": "A"}
    ]},
    "mark": "bar",
    "encoding": {
        "x": {
            "field": "b", "type": "quantitative", "axis": false
        },
        "y": {
            "field": "a", "type": "nominal", "axis": false
        },
        "row": {
            "field": "c", "type": "nominal",
            "axis": {
                "title": "", "labelAngle": 0
            }
        }
    }
}
```
- [X] Provide an example spec and/or screenshots (if applicable) 
When labelAngle is not set, we have:
vega 1
However, when labelAngle = 0, there is no difference from default:
vega 1
Further, when labelAngle = 360, it works:
vega 2

It seems labelAngle = 0 is treated the same as default. How can I force it without using 360?

YuhanLu:
Fix #1664 
```
{
    "data": {"values": [
      {"a": "A","b": 28, "c": "A"}, 
      {"a": "B","b": 55, "c": "A"}, 
      {"a": "C","b": 43, "c": "A"}
    ]},
    "mark": "bar",
    "encoding": {
        "x": {
            "field": "b", "type": "quantitative", "axis": false
        },
        "y": {
            "field": "a", "type": "nominal", "axis": false
        },
        "row": {
            "field": "c", "type": "nominal",
            "axis": {
                "title": "", "labelAngle": 0
            }
        }
    }
}
```

screen shot 2017-04-03 at 10 08 39 am

### DIFF
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -121,10 +121,10 @@ export class FacetModel extends Model {
 
           if (channel === ROW) {
             const yAxis: any = child.axis(Y);
-            if (yAxis && yAxis.orient !== 'right' && !modelAxis.orient) {
+            if (yAxis && yAxis.orient !== 'right' && modelAxis.orient === undefined) {
               modelAxis.orient = 'right';
             }
-            if (model.hasDescendantWithFieldOnChannel(X) && !modelAxis.labelAngle) {
+            if (model.hasDescendantWithFieldOnChannel(X) && modelAxis.labelAngle === undefined) {
               modelAxis.labelAngle = modelAxis.orient === 'right' ? 90 : 270;
             }
           }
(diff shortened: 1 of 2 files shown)


domoritz:
Awesome. Can you create a test case where the label angle is 0?

PR #2073, opened 2017-04-03
correct `interpolate` to support more scale type

### LINKED ISSUE #1922: Correct "interpolate" support
Currently interpolate only supports linear, but it should be supported by others too? 
(Need to check with Vega team first.)

https://github.com/vega/vega-lite/pull/1888#discussion_r99174470

YuhanLu:
Fix #1922 
Now interpolate support `linear`, `bin-linear`, `log`, `pow`, `sqrt` scale type.

### DIFF
--- a/src/scale.ts
+++ b/src/scale.ts
@@ -368,7 +368,7 @@ export function scaleTypeSupportProperty(scaleType: ScaleType, propName: keyof S
     case 'scheme':
       return true;
     case 'interpolate':
-      return scaleType === 'linear' || scaleType === 'bin-linear';
+      return contains(['linear', 'bin-linear', 'pow', 'log', 'sqrt', 'utc', 'time'], scaleType);
     case 'round':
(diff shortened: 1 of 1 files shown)


domoritz:
What about time and utc?

YuhanLu:
Based on Jeff: 

> Jeffrey Heer [1:10 PM] 
Interpolate _should_ work for continuous color scales (linear, log, sqrt, pow). It definitely will not work on sequential, which in general tend to have pre-defined color schemes in which the interpolation has already been decided. We could in theory make it work for "derived" sequential scales that take an array of colors as the range input, but that would require changes to Vega's scale generation code.

interpolate should only support linear, log, sqrt, pow for now

domoritz:
Well, it also works for `bin-linear`. I just looked at https://vega.github.io/vega/docs/scales/ and was wondering.

domoritz:
@YuhanLu Did you check with the vega schema?

YuhanLu:
I have tested them and they work for all those types

PR #2076, opened 2017-04-03
Add proper types for all vega transforms that we use.

domoritz:

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -1,16 +1,6 @@
 import {Channel, COLUMN, ROW, SHAPE, SIZE} from './channel';
 
-/**
- * Binning properties or boolean flag for determining whether to bin data or not.
- */
-export interface Bin {
-  /**
-   * A two-element (`[min, max]`) array indicating the range of desired bin values.
-   * @minItems 2
-   * @maxItems 2
-   */
-  extent?: number[];
-
+export interface BinBase {
   /**
    * The number base to use for automatic bin determination (default is base 10).
    */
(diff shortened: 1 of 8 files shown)


src/bin.ts
```diff
@@ -1,16 +1,6 @@
...
- */
-export interface Bin {
-  /**
-   * A two-element (`[min, max]`) array indicating the range of desired bin values.
-   * @minItems 2
-   * @maxItems 2
-   */
-  extent?: number[];
-
+export interface AbstractBin {
```
kanitw: Why AbstractBin shouldn't have `extent`?  (Please add comment to explain)

Also, we use `Base` prefix rather than `Abstract` for the rest of the project.

src/compile/layout.ts
```diff
@@ -39,7 +39,12 @@ export function assembleLayout(model: Model, layoutData: VgData[]): VgData[] {
   if (true) { // if both are shared scale, we can simply merge data source for width and for height
     const distinctFields = keys(extend(layoutComponent.width.distinct, layoutComponent.height.distinct));
     const formula = layoutComponent.width.formula.concat(layoutComponent.height.formula)
-      .map(f => extend({type: 'formula'}, f));
+      .map(f => {
+        return {
+          type: 'formula',
+          ...f
+        } as VgFormulaTransform;
```
kanitw: Instead of using `as` to cast, I wonder if it's possible to do

```
const formula: VgTransform[]
```

src/bin.ts
domoritz: commented

src/compile/layout.ts
domoritz: Nope

```
[ts]
Type '{ as: string; expr: string; type: string; }[]' is not assignable to type 'VgFormulaTransform[]'.
  Type '{ as: string; expr: string; type: string; }' is not assignable to type 'VgFormulaTransform'.
    Types of property 'type' are incompatible.
      Type 'string' is not assignable to type '"formula"'.
```

src/bin.ts
kanitw: ? I don't see the comment

Also

> Also, we use `Base` prefix rather than `Abstract` for the rest of the project.

domoritz: Ups, didn't push. Sorry.

src/bin.ts
```diff
@@ -1,16 +1,6 @@
...
- */
-export interface Bin {
-  /**
-   * A two-element (`[min, max]`) array indicating the range of desired bin values.
-   * @minItems 2
-   * @maxItems 2
-   */
-  extent?: number[];
-
+export interface BasicBin {
```
kanitw: BinBase?  for consistency with LegendBase, AxisBase, etc.

domoritz: Done

PR #2079, opened 2017-04-05
Replace `\"` in examples with `'`

kanitw:


src/compile/facet.ts
```diff
@@ -312,11 +312,16 @@ export class FacetModel extends Model {
   }
 
   public assembleMarks(): any[] {
+
```
domoritz: why is this in here?

kanitw:
Fixed

PR #2080, opened 2017-04-05
Enable checking on scale type if is supported by data type

### LINKED ISSUE #1885: Check if a scale type is supported by the data type
In [`compile/scale/type.ts`](https://github.com/vega/vega-lite/blob/master/src/compile/scale/type.ts#L28), we checked if a scale type is supported by an encoding channel and throw warning.  

In addition, we should check if a scale type is in compatible with a data type (in addition to channel).

For rules, we can extract `dataTypeAndFunctionMatchScaleType` constraint from CQL (See https://github.com/vega/compassql/blob/master/src/constraint/encoding.ts).

- [ ] After finishing this, we can make `dataTypeAndFunctionMatchScaleType` in CQL calls our new method in VL

YuhanLu:
Fix #1885 
Enable checking on scale type if is supported by data type

### DIFF
--- a/src/compile/scale/type.ts
+++ b/src/compile/scale/type.ts
@@ -6,7 +6,10 @@ import {ScaleConfig, ScaleType} from '../../scale';
 import {isDiscreteByDefault} from '../../timeunit';
 
 import {FieldDef} from '../../fielddef';
+import {hasDiscreteDomain} from '../../scale';
+import {Type} from '../../type';
 import * as util from '../../util';
+import {contains} from '../../util';
 
 export type RangeType = 'continuous' | 'discrete' | 'flexible' | undefined;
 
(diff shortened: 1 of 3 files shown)


src/compile/scale/type.ts
```diff
@@ -33,8 +36,8 @@ export default function type(
       return defaultScaleType;
     }
 
-    // Check if explicitly specified scale type is supported by the channel
-    if (supportScaleType(channel, specifiedType)) {
+    // Check if explicitly specified scale type is supported by the channel and data type
+    if (supportScaleType(channel, specifiedType) && dataTypeMatchScaleType(specifiedType, fieldDef)) {
```
kanitw: The error below outputs error for scaleTypeNotWorkWithChannel, which can be incorrect.

Better make it throws the right warnings for not work with channel / not work with data type.

PR #2083, opened 2017-04-05
Support `filled` in mark definition

kanitw:


src/mark.ts
```diff
@@ -51,6 +51,12 @@ export interface MarkDef {
   role?: string;
 
   /**
+   * Whether the mark\'s color should be used as fill color instead of stroke color.
```
domoritz: why do you need to escape the `'` ?

kanitw: IDK, I was wondering why the old comment keep it too lol -- let's remove it.

PR #2085, opened 2017-04-07
Auto-generate table in Documentation

YuhanLu:
Fix part of #1723 

Also apply this auto-generation for Axis and Legend documentation.


YuhanLu:
I tried to make the doc the same as before but there's still 2 places I could not fix:
1. the link on type of value (DateTime[])
2. the size of the table for Axis Config General is not fixed

I guess the second issue should be fixed in css?

domoritz:
Unfortunately, Jekyll does not allow symlinks. What are your thoughts on using jq to extract only the relevant bits from the schema? I don't like that we have a full copy of the schema. Another option is to copy but not include it in git and only commit it when we release to gh-pages. Which one do you prefer?

domoritz:
This is awesome btw.

kanitw:
> Another option is to copy but not include it in git and only commit it when we release to gh-pages. Which one do you prefer?

I prefer this option (at least for now).  Using jq sounds like an unnecessary overhead + work.

site/docs/axis.md
```diff
@@ -38,64 +38,44 @@ Axis properties can be customized by setting `axis` to an axis property object.
...
-| titleFontSize	| Number 		| The font size of title, in pixels. <span class="note-line">__Default value:__ 10. </span> |
-| titleOffset   | Number        | The offset (in pixels) from the axis at which to place the title.|
-| titleMaxLength  | Integer     | Max length for axis title when the title is automatically generated from the field\'s description. <span class="note-line">__Default value:__  automatically determined based on the cell size (`config.cell.width`, `config.cell.height`) and `characterWidth` property.</span> |
-| characterWidth  | Integer       | Character width for automatically determining the value of `titleMaxLength`. <span class="note-line">__Default value:__  derived from [axis config](config.html#axis-config)'s `characterWidth` (`6` by default). </span> |
+### Axis Properties
+{% assign schema = site.data.vega-lite-schema.definitions %}
+
+#### General
+
+{% include table.html props="domain,orient,offset,position,zindex" source= schema.Axis.properties %}
```
kanitw: Is there any other pattern besides `schema.XXX.properties`?

If not, make your script only take `XXX` (e.g., `Axis` in this case)?

site/docs/axis.md
```diff
@@ -38,64 +38,44 @@ Axis properties can be customized by setting `axis` to an axis property object.
...
-| titleColor	| String 		| The color of the title. |
-| titleFont		| String 		| The font of title (e.g., `Helvetica Neue`). |
-| titleFontWeight	| Number 	 	|  The font weight of title (e.g., `bold`). |
-| titleFontSize	| Number 		| The font size of title, in pixels. <span class="note-line">__Default value:__ 10. </span> |
-| titleOffset   | Number        | The offset (in pixels) from the axis at which to place the title.|
-| titleMaxLength  | Integer     | Max length for axis title when the title is automatically generated from the field\'s description. <span class="note-line">__Default value:__  automatically determined based on the cell size (`config.cell.width`, `config.cell.height`) and `characterWidth` property.</span> |
-| characterWidth  | Integer       | Character width for automatically determining the value of `titleMaxLength`. <span class="note-line">__Default value:__  derived from [axis config](config.html#axis-config)'s `characterWidth` (`6` by default). </span> |
+### Axis Properties
```
kanitw: The intro paragraph above is outdated.   Now people can customize axis either via Axis Properties (which is applied to a specific config.) or via Axis Config (which get applies to all Axis).

Please update the intro paragraph accordingly.  You might want to add an introduce paragraph for "Axis Properties" and "Axis Config" as well.

site/docs/axis.md
```diff
@@ -38,64 +38,44 @@ Axis properties can be customized by setting `axis` to an axis property object.
...
+{% assign schema = site.data.vega-lite-schema.definitions %}
```
kanitw: Why do we need to assign this here?  Isn't this the same for every file?  If so, just assign this in table.html?

src/vega.schema.ts
```diff
@@ -254,17 +254,23 @@ export interface VgAxisBase {
   domain?: boolean;
 
   /**
-   * A flag indicate if gridlines should be created in addition to ticks. If `grid` is unspecified, the default value is `true` for ROW and COL. For X and Y, the default value is `true` for quantitative and time fields and `false` otherwise.
+   * A flag indicate if gridlines should be created in addition to ticks. If `grid` is unspecified, the default value is `"true"` "for `ROW` and `COL`. For `X` and `Y`, the default value is `"true"` for quantitative and time fields and `"false"` otherwise.
```
kanitw: For `x` and `y`,   

(They should be small caps.)

src/vega.schema.ts
```diff
@@ -254,17 +254,23 @@ export interface VgAxisBase {
...
    * A boolean flag indicating if labels should be included as part of the axis (default true).
+   *
+   * __Default value:__  derived from [axis config](config.html#axis-config)'s `labels` (`"true"` by default).
    */
   labels?: boolean;
 
   /**
    * The rotation angle of the axis labels.
+   *
+   * __Default value:__ `"-45"` for time or ordinal axis and `"0"` otherwise.
```
kanitw: Why are "-45" and "0" strings?

src/vega.schema.ts
```diff
@@ -277,12 +283,18 @@ export interface VgAxisBase {
 
   /**
    * The size, in pixels, of major, minor and end ticks.
+   *
+   * __Default value:__  derived from [axis config](config.html#axis-config)'s `tickSize` (`"6"` by default).
```
kanitw: 6 is a number!

_includes/table.html
```diff
@@ -0,0 +1,51 @@
...
+          {% capture type %}
+            {% assign para = include.source[prop].anyOf %}
+            {% include type.html types=para %}
+          {% endcapture %}
+        {% endif %}
+      {% elsif type == "array" %}
+         <!--special case for type is arrays of objects -->
+        {% assign type = include.source[prop].items.type | append: "[]" | capitalize %}
+      {% elsif type.first != nil %}
+        <!--special case for type is multiple objects-->
```
kanitw: Do you mean if "type in the schema is an array of multiple types"?

If so, maybe update the comment a bit?

_includes/table.html
```diff
@@ -0,0 +1,51 @@
...
+        {% assign res = ""%}
```
kanitw: what is res?   result?

_includes/type.html
```diff
@@ -0,0 +1,19 @@
+{% assign res = "" %}
```
kanitw: This file seems to be the `type` for only the "anyOf" case?   If so, rename this to `type-anyof.html` or make this file contain all logic for type?  (The former would be easier.)

domoritz:
LGTM besides the changes @kanitw asked for.

_includes/docs_toc.md
```diff
@@ -15,7 +15,6 @@
     - [Text]({{site.baseurl}}/docs/mark.html#text)
     - [Tick]({{site.baseurl}}/docs/mark.html#tick)
     - [Bar]({{site.baseurl}}/docs/mark.html#bar)
-    - [Rect]({{site.baseurl}}/docs/mark.html#rect)
```
domoritz: where did the rect doc go? cc @kanitw

domoritz:
@YuhanLu I pushed some style fixes.

_includes/docs_toc.md
kanitw: Weird.  Why is this changing?

domoritz: It isn't. I reran the script and it added it back.

domoritz: https://github.com/vega/vega-lite/pull/2085/commits/2c87747ab01779748bed71d06b08f8517030e4ec

domoritz:
@kanitw I think this is ready to be merged.

PR #2086, opened 2017-04-07
Dataflow graph

domoritz:
Implement a new dataflow that replaces data components. 

```
     +--------+
     | Source |
     +---+----+
         |
         v
       Parse
         |
         v
     Transforms
(Filter, Compute, ...)
         |
         v
     Null Filter
         |
         v
      Binning
         |
         v
     Timeunit
         |
         v
      +--+--+
      | Raw |
      +-----+
         |
         v
     Aggregate
         |
         v
       Stack
         |
         v
      >0 Filter
         |
         v
     Path Order
         |
         v
   +----------+
   |   Main   +----> Layout
   +----------+
         |
         v
     +-------+
     | Facet |----> Child data...
     +-------+
```


kanitw:
File Reviewed (Will keep adding to this list)

- src/compile/data/dataflow.ts
- src/compile/data/parse.ts

src/compile/data/stack.ts
```diff
@@ -1,28 +1,31 @@
...
+      binSuffix: scale && hasDiscreteDomain(scale.type) ? 'range' : 'start'
+    });
+    if (_field) {
+      fields.push(_field);
+    }
+    return fields;
+  }, [] as string[]);
+}
 
+export interface Stack {
```
kanitw: (Minor) Why don't these become just properties of `StackNode`?
I guess just for convenience?

src/compile/data/parse.ts
```diff
@@ -0,0 +1,193 @@
...
+  if (model instanceof UnitModel) {
+    const order = new OrderNode(model);
+    if (order.hasSort()) {
+      order.parent = head;
+      head = order;
+    }
+  }
+
+  // add filter for non-positive data if needed
+  const nonPosFilter = new NonPositiveFilterNode(model);
```
kanitw: Why does NonPositiveFilter come after order by default?

src/compile/data/parse.ts
```diff
@@ -0,0 +1,193 @@
...
+import {NullFilterNode} from './nullfilter';
+import {OrderNode} from './pathorder';
+import {SourceNode} from './source';
+import {StackNode} from './stack';
+import {TimeUnitNode} from './timeunit';
+import {parseTransformArray} from './transforms';
+
+function parseRoot(model: Model, sources: Dict<SourceNode>): DataFlowNode {
+  if (model.data) {
+    // If we have data, try to insert it into existing sources.
```
kanitw: Do you really insert anything here? I think you just try to "reuse existing sources"

src/compile/data/parse.ts
```diff
@@ -0,0 +1,193 @@
...
+export function parseData(model: Model): DataComponent {
+  const root = parseRoot(model, model.component.data.sources);
+
+  const outputNodes = model.component.data.outputNodes;
+
+  // the current head of the tree that we are appending to
+  let head = root;
+
+  // add format parse
+  const parse = new ParseNode(model);
```
domoritz: Yeah. We cannot guarantee that the parse always goes up to the source.

src/compile/data/dataflow.ts
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * A node in the dataflow graph.
+ */
+export class DataFlowNode {
+  private _children: DataFlowNode[];
+
+  private _parent: DataFlowNode;
+
+  constructor(public readonly debugName?: string) {
+    this._children = [];
```
domoritz: Ts requires the super() call so you cannot forget it. Your code won't compile without it.

domoritz: But what you say makes sense nonetheless.

src/compile/data/bin.ts
```diff
@@ -1,110 +1,141 @@
...
+interface Component {
+  bin: Bin;
+  field: string;
+  extentSignal: string;
+  signal: string;
+  as: string[];
+  /**
+   * Range formula.
+   */
+  formula?: string;
```
domoritz: But won't we need the model?

kanitw: I think I miss one of the other parameter required, but let's not worry about this for now as it is minor.

src/compile/data/assemble.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  roots.forEach(append);
+  return leaves;
+}
+
+/**
+ * Print debug information for dataflow tree.
+ */
+function debug(node: DataFlowNode) {
```
kanitw: This seems useful, but currently is never called.  I wonder if we should always call it, but check if log.level is DEBUG?

In any case, replace console.log with log.debug

src/compile/data/assemble.ts
```diff
@@ -0,0 +1,238 @@
...
+          walkTree(child, newData);
+        });
+        break;
+    }
+  }
+
+  return walkTree;
+}
+
+export function assembleFacetData(root: FacetNode): VgData[] {
```
kanitw: If i understand correctly, this is for 
```
/** assembling data sources that are derived from faceted data. */
````

If so, add comment?

File: src/compile/data/optimizers.ts [hunk 9]
```diff
@@ -0,0 +1,150 @@
...
+import {StackNode} from './stack';
+import {TimeUnitNode} from './timeunit';
+import {CalculateNode, FilterNode} from './transforms';
+
+/**
+ * Start optimization path at the leaves. Useful for merging up things.
+ */
+export function optimizeFromLeaves(f: (node: DataFlowNode) => void) {
+  function optimizeNextFromLeaves(node: DataFlowNode) {
+    if (!node || !node.parent || node.parent instanceof SourceNode) {
```
kanitw: I’m wondering:
1) Can nodes that are not SourceNode have no parent?
2) Can `SourceNode`s have parent?

src/compile/data/parse.ts
```diff
@@ -0,0 +1,193 @@
...
+    if (hash in sources) {
+      // use a reference if we already have a source
+      return sources[hash];
+    } else {
+      // otherwise add a new one
+      sources[hash] = source;
+      return source;
+    }
+  } else {
+    // If we don't have a source defined, use the parent's facet root or main.
```
domoritz: Not just facet. Any parent.

src/compile/data/parse.ts
domoritz: I moved it right before main.

src/compile/data/optimizers.ts
```diff
@@ -0,0 +1,150 @@
...
+  return optimizeNextFromLeaves;
+}
+
+
+export function parse(node: DataFlowNode) {
+  const parent = node.parent;
+
+  // move parse up by merging or swapping
+  if (node instanceof ParseNode) {
+    if (parent instanceof ParseNode) {
```
domoritz: Done

src/compile/data/nullfilter.ts
```diff
@@ -15,59 +12,44 @@ const DEFAULT_NULL_FILTERS = {
...
-        // define this so we know that we don't filter nulls for this field
-        // this makes it easier to merge into parents
-        aggregator[fieldDef.field] = null;
-      }
-    }
-    return aggregator;
-  }, {});
-}
+export class NullFilterNode extends DataFlowNode {
+  private _aggregator: Dict<FieldDef>;
```
domoritz: Done

src/compile/data/stack.ts
domoritz: Yeah. And testing.

src/compile/data/parse.ts
domoritz: Fixed

src/compile/data/assemble.ts
```diff
@@ -0,0 +1,242 @@
...
+      if (node.parent instanceof SourceNode && dataSource.format) {
+        dataSource.format.parse = node.assemble();
+      } else {
+        throw new Error('Can only instantiate parse next to source.');
+      }
+    }
+
+    if (node instanceof FacetNode) {
+      if (!dataSource.name) {
+        dataSource.name = `data_${datasetIndex++}`;
```
kanitw: I assume that this is a faceted data source that faceted marks will refer to?  If so, name it `faceted_`?

src/compile/data/assemble.ts
domoritz: `log.debug` doesn't print anything. I need to see it during development.

src/compile/data/assemble.ts
domoritz: Done

src/compile/data/optimizers.ts
domoritz: 1) no
2) no

domoritz: Fixed

domoritz:
Node 4 sorts differently. Drop it or what? https://travis-ci.org/vega/vega-lite/jobs/221835692

src/compile/data/aggregate.ts
```diff
@@ -49,9 +48,21 @@ export class AggregateNode extends DataFlowNode implements NewFieldNode, Depende
   /** dictionary mapping field name => dict set of aggregation functions */
   private measures: Dict<StringSet>;
 
+  public clone(): this {
+    const cloneObj = new (<any>this.constructor);
```
kanitw: This looks like a hack to get around the fact that `AggregateNode()` constructor requires `model` as input. 

I wonder if we should make AggregateNode takes `dimensions` and `measures` as input and make a factory that takes a model as input?  

I believe making aggregateNode taking `dimensions` and `measures` as input might make writing test easier as well (not sure about this latter comment though).

kanitw: (<any>this.constructor) looks like a hack to get around the fact that AggregateNode() constructor requires model as input.

I think we should make AggregateNode takes dimensions and measures as input and make a factory that takes a model as input.

I believe making aggregateNode taking dimensions and measures as input might make writing test easier as well (not sure about this latter comment though).

kanitw:
-[ ] `new (<any>this.constructor)` looks like a hack to get around the fact that AggregateNode() constructor requires model as input.  I think we should make AggregateNode takes dimensions and measures as input and make a factory that takes a model as input.
  - I believe making aggregateNode taking dimensions and measures as input might make writing test easier as well (not sure about this latter comment though).

src/compile/data/dataflow.ts
```diff
@@ -127,3 +135,23 @@ export interface DependentNode {
...
+   */
+  clone: () => this;
+}
+
+
+export function isNewFieldNode(node: any): node is NewFieldNode {
+  return 'produces' in node && isFunction(node.produces);
+}
+
+export function isClonable(node: any): node is ClonableNode {
```
kanitw: Instead of using `ClonableNode` mixins and `isClonable` function like this.

I think we could just add two methods `isClonable()` and `clone()` to all DataFlowNode. 
`clone()` by default (if not overridden) can throw `new Error("Not clonable")`.

src/compile/data/dataflow.ts
```diff
@@ -127,3 +135,23 @@ export interface DependentNode {
...
+ */
+export interface ClonableNode {
+  /**
+   * Clone this node with a deep copy.
+   */
+  clone: () => this;
+}
+
+
+export function isNewFieldNode(node: any): node is NewFieldNode {
```
kanitw: Can't we just make produces return `null` for nodes that don't produce new field?

src/compile/data/dataflow.ts
```diff
@@ -127,3 +135,23 @@ export interface DependentNode {
    */
   dependentFields: () => StringSet;
```
kanitw: Can't we just make produces return `null` for nodes that don't depend on others?

src/compile/data/assemble.ts
```diff
@@ -36,6 +42,79 @@ function removeUnnecessaryNodes(node: DataFlowNode) {
 
   node.children.forEach(removeUnnecessaryNodes);
 }
+
+/**
+ * Clones the subtree and ignores output nodes except for the leafs, which are renamed.
```
kanitw: It seems like you also ignore unclonable node, but I'm not sure why.  
Can you explain more?

src/compile/data/assemble.ts
```diff
@@ -36,6 +42,79 @@ function removeUnnecessaryNodes(node: DataFlowNode) {
...
+      const copy = node.clone();
+
+      if (copy instanceof OutputNode) {
+        const newName = FACET_SCALE_PREFIX + facet.model.getName(copy.source);
+        copy.source = newName;
+
+        facet.model.component.data.outputNodes[newName] = copy;
+
+        flatten(node.children.map(clone)).forEach((n: DataFlowNode) => n.parent = copy);
+      } else if (copy instanceof AggregateNode) {
```
kanitw: Just like `AggregateNode`, you have to augment `groupby` of `StackNode` as well.

src/compile/data/assemble.ts
```diff
@@ -36,6 +42,79 @@ function removeUnnecessaryNodes(node: DataFlowNode) {
...
+  return clone;
+}
+
+/**
+ * Move facet nodes down to the next fork or output node. Also pull the main output with the facet node.
+ * After moving down the facet node, make a copy of the subtree and make it a child of the main output.
+ */
+function moveFacetDown(node: DataFlowNode) {
+  if (node instanceof FacetNode) {
+    if (node.numChildren() === 1 && !(node.children[0] instanceof OutputNode)) {
```
kanitw: I think if there is no fork but there is a unit plot with some shared scales and some independent scales, we have to perform a similar subtree clone procedure.  

If you think this is true, we can file an issue and handle this later though.

src/compile/model.ts
```diff
@@ -247,7 +247,7 @@ export abstract class Model {
 
   public abstract channelHasField(channel: Channel): boolean;
 
-  public getName(text: string, delimiter: string = '_') {
+  public getName = (text: string, delimiter: string = '_') => {
```
kanitw: why is this better?

src/compile/data/bin.ts
```diff
@@ -47,9 +45,20 @@ interface BinComponent {
...
+  public clone(): this {
+    const cloneObj = new (<any>this.constructor);
+    cloneObj.bins = duplicate(this.bins);
+    return cloneObj;
+  }
+
   constructor(model: Model) {
     super();
 
+    if (!model) {
```
kanitw: See [this comment](https://github.com/vega/vega-lite/pull/2086#pullrequestreview-32722315) -- we should get rid of this hack!

src/compile/model.ts
domoritz: Because this is available inside if you pass the function somewhere. But I'm not using it any more so I'll clean it up.

src/compile/data/assemble.ts
kanitw: We should allow Vega-editor to change the log level.

src/compile/data/assemble.ts
domoritz: I can neither confirm nor deny.

src/compile/data/assemble.ts
domoritz: Good catch!

src/compile/data/assemble.ts
domoritz: All nodes that can happen are clonable.

src/compile/data/bin.ts
domoritz: How do I remove this hack?

src/compile/data/dataflow.ts
domoritz: Done.

src/compile/data/dataflow.ts
domoritz: Better return []?

src/compile/data/dataflow.ts
domoritz: Better return {}?

PR #2090, opened 2017-04-08
Follow up for preventing side effect in fieldDef 

kanitw:
(forgot to fix it for one condition)

cc: @NeelMohapatra -- you probably wanna rebase after we merge this too.


domoritz:
Looks like we want to use readonly to enforce some of these things, right?

kanitw:
Good idea.

PR #2092, opened 2017-04-08
Get rid of config.cell and config.stack in the examples

kanitw:


examples/specs/stacked_area_normalize.vl.json
```diff
@@ -1,22 +1,20 @@
...
   "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
   "data": {"url": "data/unemployment-across-industries.json"},
+  "width": 300, "height": 200,
   "mark": "area",
   "encoding": {
     "x": {
       "timeUnit": "yearmonth", "field": "date", "type": "temporal",
       "scale": {"nice": "month"},
-      "axis": {"format": "%Y", "labelAngle": 0}
+      "axis": {"domain": false, "format": "%Y", "labelAngle": 0}
```
domoritz: what's up with this?

kanitw: This is the right way to do what used to be `config.axis.domainWidth = 0`

Issue #2102, opened 2017-04-11
tsc is broken by the latest Chai typings


JoshuaKGoldberg:
Cross-posting here what was reported by @domoritz in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/15768. I'm averse to changing the typings to not being as strict by default because I think there's benefit in enforcing type safety on these supposed-to-be-type-safe assertions. But, since I inconvenienced you with this, I'm willing to go through your test files and add the types.

domoritz:
That would be awesome! Since the corresponding ts issue has been open for a while, it seems worth it to explicitly specify the types. However, I'm also okay with sticking with `3.4.35` (don't want to create unnecessary manual labor for you).

JoshuaKGoldberg:
Too late :)

domoritz:
Many thanks from the other side of the lake!

PR #2103, opened 2017-04-11
Manually specified some typings for assertions

JoshuaKGoldberg:
The other errors I don't know how to deal with. Are they caught type problems or...?

Starts on #2102.


domoritz:
Fixed in #2106

JoshuaKGoldberg:
:+1:

PR #2104, opened 2017-04-11
Simplify Spec.ts

kanitw:
Update spec.ts to
- remove unnecessary declaration
- make `TopLevelExtendedSpec` more correct.  (Facet spec and layer spec should not include row, column inside.)

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -48,6 +48,19 @@
                             "$ref": "#/definitions/MarkDef"
                         },
                         {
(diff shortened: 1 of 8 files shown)


build/vega-lite-schema.json
```diff
@@ -48,6 +48,19 @@
                             "$ref": "#/definitions/MarkDef"
                         },
                         {
+                            "enum": [
```
domoritz: yaay

Issue #1976, opened 2017-02-23
Doc improvements


domoritz:
The Vega docs have a few improvements over the Vega-Lite docs. We should port them over here. The docs will be published in the vega repo shortly.

- [x] Link to types
- [x] Make the images in the example gallery scroll on hover
- [x] Change aspect ratio of images to match Vega docs
- [x] Purple header

domoritz:
@kanitw anything else?

domoritz:
@YuhanLu I think you did the first one, right?

domoritz:
@NeelMohapatra this is the issue we talked about. feel free to add more things to the list above.

PR #2117, opened 2017-04-12
Updated colors of Header

NeelMohapatra:
Addresses 4th issue of #1976

### DIFF
--- a/site/index.md
+++ b/site/index.md
@@ -1,41 +1,21 @@
 ---
-layout: page
+layout: home
 permalink: /
+title: A High-Level Visualization Grammar
 
 images:
  - spec: stacked_area_stream
-   bgposition: 44% 22%
  - spec: circle
-   bgposition: 75% 30%
  - spec: bar_layered_transparent
-   bgposition: 25% 0
-   bgsize: 250%
  - spec: tick_strip
-   bgposition: 75% 0%
-   bgsize: 150%
  - spec: line_color
-   bgposition: 41% 26%
-   bgsize: 170%
  - spec: scatter_colored_with_shape
-   bgposition: 50% 30%
  - spec: trellis_barley
-   bgsize: 226%
-   bgposition: 40% 5%
  - spec: bar_grouped
-   bgposition: 75% 59%
  - spec: stacked_bar_weather
-   bgsize: 170%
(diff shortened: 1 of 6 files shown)


_layouts/plain.html
```diff
@@ -5,7 +5,6 @@
 <div class="fill">
   <header>
     <nav class="page-centered">
-      <a class="vl-title" href="{{site.baseurl}}/">{{site.title}}</a>
```
domoritz: Vega didn't remove this and we wouldn't have a way to get back to the home page.

site/index.md
```diff
@@ -35,6 +35,11 @@ images:
 {:.hidden}
 # Vega-Lite: a high-level visualization grammar
```
domoritz: This hidden link was there for seo, I think. Can you use it instead of the new one below. 

```
{:.page-centered}
```

applies a class as you can see.

site/static/main.css
```diff
@@ -260,6 +261,10 @@ nav a:hover {
   text-decoration: none;
 }
 
+.title h1 {
+  margin-top: 0.3em;
```
domoritz: Not sure why this is needed. Why is this different than any other page on our website?

_layouts/plain.html
NeelMohapatra: Ah. I thought Vega removed it from their main documentation. It's not there on their main page. I will fix it

site/static/main.css
NeelMohapatra: I didn't like the gap in the title. That's why I added it there. Even Vega does it

domoritz: Well, I don't like special casing a single page without a reason. But this makes sense. Let's make it cleaner though by referring to the title through a class for the homepage.

_layouts/plain.html
domoritz: https://github.com/vega/vega/pull/773 :-)

site/static/main.css
NeelMohapatra: On it

domoritz:
Can you resolve the conflicts?

domoritz:
I made a bunch more fixes and improvements.

PR #2122, opened 2017-04-12
Fast iteration testing.

domoritz:


CONTRIBUTING.md
```diff
@@ -146,6 +146,10 @@ If you only want subset of these actions, you can use:
 
 - `npm run watch:build` to start a watcher task that **re-compiles Vega-Lite** when `.ts` files related to VL change.
 
+#### Fast iteration testing
+
+To quickly run tests without long compile times, run `yarn run tsc -- -w` in a separate terminal session. Then run `npm run mocha:test` to quickly run tests (or `npm run mocha:test -- --inspect --debug-brk` to inspect tests). Please note that this does not update the schema and other things required for testing. Use this trick with care.
```
kanitw: why only `yarn run` in the first one but `npm run` later?

CONTRIBUTING.md
```diff
@@ -146,6 +146,10 @@ If you only want subset of these actions, you can use:
 
 - `npm run watch:build` to start a watcher task that **re-compiles Vega-Lite** when `.ts` files related to VL change.
 
+#### Fast iteration testing
```
kanitw: Fast Unit Testing

What you suggests below only runs unit test, but not example validation test. 
("other things required for testing" is vague.)  

Since this won't make test really pass on Travis, I wonder if we should really include this in CONTRIBUTING.md.

domoritz: Well, I'm using it and it's sooooo fast. I can revise the text to say this is only for dev but you should run full tests before pushing.

CONTRIBUTING.md
domoritz: Fixed

domoritz:
Done

PR #2126, opened 2017-04-12
[WIP] Refactor

kanitw:


domoritz:
Not sure yet. I might want to rename the optimizers.ts to something different. The other one is actually needed because it removes nodes.

PR #2127, opened 2017-04-12
Revert "Revert "Enable auto-generation for bin.md""

YuhanLu:
Reverts vega/vega-lite#2120


domoritz:
Really? Revert the revert?

we-need-to-go-deeper

PR #2132, opened 2017-04-13
Add support for tooltips

doug:
Please:
- [x] Provide a test case & update the documentation under `site/docs/`

### DIFF
--- a/src/compile/mark/mixins.ts
+++ b/src/compile/mark/mixins.ts
@@ -84,9 +84,10 @@ function selectionTest(model: UnitModel, selectionName: string) {
     predicate(selection.name, selection.type, selection.resolve);
 }
 
-export function text(model: UnitModel) {
-  const channelDef = model.encoding.text;
-  return wrapCondition(model, channelDef && channelDef.condition, 'text', ref.text(channelDef, model.config));
+export function text(model: UnitModel, vgChannel: 'text' | 'tooltip' = 'text') {
+  const channelDef = model.encoding[vgChannel];
+  const valueRef = (vgChannel === 'tooltip' && !channelDef) ? undefined : ref.text(channelDef, model.config);
+  return wrapCondition(model, channelDef && channelDef.condition, vgChannel, valueRef);
 }
 
 export function bandPosition(channel: 'x'|'y', model: UnitModel) {
--- a/src/encoding.ts
+++ b/src/encoding.ts
@@ -91,6 +91,11 @@ export interface Encoding<F> {
   text?: TextFieldDef<F> | ConditionalValueDef<string|number|boolean>;
 
   /**
+   * The tooltip text to show upon mouse hover.
+   */
+  tooltip?: TextFieldDef<F> | ConditionalValueDef<string>;
+
+  /**
    * stack order for stacked marks or order of data points in line marks.
    */
   order?: OrderFieldDef<F> | OrderFieldDef<F>[];
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -2155,6 +2155,17 @@
                     ],
                     "description": "Text of the `text` mark."
                 },
+                "tooltip": {
+                    "anyOf": [
+                        {
+                            "$ref": "#/definitions/ConditionalValueDef<string>"
+                        },
+                        {
+                            "$ref": "#/definitions/TextFieldDef"
+                        }
+                    ],
+                    "description": "The tooltip text to show upon mouse hover."
+                },
                 "x": {
                     "anyOf": [
                         {
@@ -2295,6 +2306,17 @@
                         }
                     ],
                     "description": "Text of the `text` mark."
+                },
+                "tooltip": {
+                    "anyOf": [
+                        {
+                            "$ref": "#/definitions/ConditionalValueDef<string>"
+                        },
+                        {
+                            "$ref": "#/definitions/TextFieldDef"
+                        }
+                    ],
+                    "description": "The tooltip text to show upon mouse hover."
                 },
                 "x": {
                     "anyOf": [
--- a/site/docs/encoding.md
+++ b/site/docs/encoding.md
@@ -22,6 +22,7 @@ An integral part of the data visualization process is encoding data with visual 
     "size": ...,
     "shape": ...,
     "text": ...,
+    "tooltip": ...,
     "detail": ...
   },
   ...
@@ -47,7 +48,7 @@ Position channels determine the position of the marks.
 
 Mark properties channels map data fields directly to visual properties of the marks. Unlike other channel types, they can be mapped to [constant values](#value) as well. Here are the supported mark properties:
 
-{% include table.html props="color,opacity,shape,size,text" source="Encoding" %}
+{% include table.html props="color,opacity,shape,size,text,tooltip" source="Encoding" %}
 
 ### Additional Level of Detail Channel
 
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -31,6 +31,7 @@ export namespace Channel {
   export const TEXT: 'text' = 'text';
   export const ORDER: 'order' = 'order';
   export const DETAIL: 'detail' = 'detail';
+  export const TOOLTIP: 'tooltip' = 'tooltip';
 }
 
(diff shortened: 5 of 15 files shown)


domoritz:
Thank your for the pull request @doug! Before we review this, note that we are working on a separate library for flexible tooltips https://github.com/vega/vega-tooltip. However, a separate tooltip channel could be great to define what should be shown.

doug:
Makes sense. A more flexible tooltip would be great. This is clearly limited, as it is only using the tooltip string property from the Vega spec and passing it through. Still useful as I calculate a tooltip string and use that field for the tooltip. Let me know if there is anything I can do to help. I clearly want this to happen :) Will the vega-tooltip library still be declarative and how will it relate to the spec? I still want to be able to serialize the entire representation including tooltips without having to call javascript.

kanitw:
@doug Thanks so much for your PR.  

I agree that this is definitely useful but we're a bit busy this week so there might be some delay for reviewing.  I hope to review this PR this weekend or next week.  

The tooltip library will probably require an additional call to JavaScript, but can produce a more powerful output than what standard `tooltip` channel in purely Vega could do.

doug:
No worries, understand being busy. And no rush on this. Honestly, I wasn't quite sure what the best place was in the code to extend, so let me know if this is useful at all, and feel free to ignore. :)

I understand tooltips in javascript are very powerful (a lot more control with less verbosity), but I would like to push back a little bit and express how much I would prefer to be able to express them declaratively in the vega lite spec.

Clearly nice tooltips can be made in Vega declaratively (https://github.com/vega/vega/blob/master/test/specs-valid/budget-forecasts.vg.json) using groups. 

For me I also care about expressing and generating the spec from languages other than javascript (particularly python at the moment) so having to call additional javascript is something I want to avoid.

So even without compiling into some complex group just using the very basic tooltip string field in Vega does most of what I need.

This patch basically allow for:
```json
{
  "data": {
    "values": [
    ]
  },
  "transform": [
    {"calculate": "datum.a + ', ' + datum.b", "as": "tooltip"},
  ],
  "mark": "bar",
  "encoding": {
    "tooltip": { "field": "tooltip" },
    },
  }
}
```

If there was a desire for more complex declarative tooltips, maybe building something similar to the new "layer" call could allow you to construct a layered encoding inside a tooltip, which would then compile in Vega to a group with multiple marks positioned relative to the mark they are attached to nearest the mouse.

And of course if you want something complex and dynamic dropping to javascript and using the vega-tooltip seems like the way to go.

craigcitro:
👍 from the peanut gallery: it would be great to have a lightweight way to do basic tooltips in vega-lite.

domoritz:
CC @sirahd

domoritz:
I'm trying to understand how this actually shows any tooltips. Can you explain this a bit more @doug?

kanitw:
@domoritz  He is attemping to use Vega's new tooltip channel, which output `title` tag in svg.  (Can't remember what Jeff uses for Canvas though.)

domoritz:
Hmm, I tried it with Vega 30 and didn't see anything in the SVG output. If it works, we should get this in as a basic tooltip support. vega-tooltip could then read that property. cc @sirahd

kanitw:
@domoritz Vega-tooltip could read that property if exists, but it doesn't the default tooltip that vega-tooltip provides is actually way more powerful.

doug:
Yes. Just passing the tooltip encoding string property to vega.

A major constraint being that it must be a string and you get no further
control on how it is displayed. Does work with both canvas and svg when I
tested it.

https://vega.github.io/vega/docs/marks/shape/
Tooltip at the bottom of general mark properties.

domoritz:
screen shot 2017-04-29 at 09 24 39

domoritz:
This is a great addition @doug! I'm happy to merge after you rebase the commit on the latest master and add a few unit tests. Now that we have more time, we will review your changes promptly!

src/compile/mark/valueref.ts
```diff
@@ -135,6 +135,16 @@ export function midPoint(channel: Channel, channelDef: ChannelDef, scaleName: st
   return defaultRef;
 }
 
+export function tooltip(def: FieldDef | ValueDef<string>, config: Config): VgValueRef {
```
kanitw: This new method doesn't deal with `timeUnit`, `bin`, and `format` correctly yet. 

Instead of creating new helper method here, I believe you can modify `text()` in `mixins.ts`  to accept a second optional parameter 

```
export function text(model: UnitModel, vgChannel: 'text' | 'tooltip' = 'text') {
   ...
}
```

You can then replace `...mixins.tooltip(model)` with `...mixins.text(model, 'text')` in all mark compilers.

With these minor modifications, I believe it should work correctly without a need for new methods in both `valueref.ts` and `mixins.ts`.  (Of course, please try to test some of these different cases.)

src/encoding.ts
```diff
@@ -75,6 +75,11 @@ export interface Encoding {
    * stack order for stacked marks or order of data points in line marks.
    */
   order?: OrderFieldDef | OrderFieldDef[];
+
+  /**
+   * Tooltip adds basic string tooltip support.
+   */
+  tooltip?: FieldDef | ValueDef<string>;
```
kanitw: This could be `TextFieldDef | ConditionalValueDef<string|number|boolean>` just like text.  

(Since tooltip will be pretty similar to `text`, you might wanna move it to above `order`, right after `text`.)

src/compile/mark/mixins.ts
```diff
@@ -87,6 +87,12 @@ export function text(model: UnitModel) {
   return wrapCondition(model, channelDef && channelDef.condition, 'text', ref.text(channelDef, model.config));
 }
 
+export function tooltip(model: UnitModel) {
```
kanitw: See my comment in `valueref.ts`.

src/encoding.ts
```diff
@@ -75,6 +75,11 @@ export interface Encoding {
    * stack order for stacked marks or order of data points in line marks.
    */
   order?: OrderFieldDef | OrderFieldDef[];
+
+  /**
+   * Tooltip adds basic string tooltip support.
```
kanitw: "The tooltip text to show upon mouse hover."  would be a bit better.  

(FYI, we use this JSDocs comment in the JSON schema, which is in turn used to populate the docs.)

src/compile/mark/valueref.ts
doug: Sounds good. Done.

src/encoding.ts
doug: Done, thanks.

src/encoding.ts
doug: I think this can only be TextFieldDef<F> | ConditionalValueDef<string>
also moved above order.

src/encoding.ts
```diff
@@ -114,6 +119,12 @@ export function normalizeEncoding(encoding: Encoding, mark: Mark): Encoding {
       return normalizedEncoding;
     }
 
+    if (channel === 'tooltip') {
```
doug: Removed. I had this in here because normalizeEncoding is called and the normalization of the tooltip tries to add a type, but a type is irrelevant for a tooltip since it has to be a string. This is no longer logging an error with the improved valueref code, so removed.

src/compile/mark/mixins.ts
doug: Done.

doug:
@kanitw @domoritz it helps to work with a well thought out code base, thanks for all the hard work on the project! I updated the pull request with the modifications (thank you for the suggestions and review @kanitw), all tests are passing, and I added some tests for tooltip as well as updated the docs. Should be good to go. :smile:

domoritz:
Awesome! Can you rerun schema to make travis pass?

doug:
Schema updated, thanks.

kanitw:
@doug Thanks so much.  Your change is now merged into master!

Are you still at Google in Seattle?  If so, we should have lunch/dinner sometimes.

domoritz:
Thank you so much for your contribution @doug!

g3o2:
Unfortunately, the spec shown here above does not work with the current implementation (vega-lite 2.0.0-beta.2 and vega 3.0.0-beta.30).

PR #2134, opened 2017-04-13
Enable auto-generation for scale.md and refactor doc generator

YuhanLu:
Fix part of #1723

### DIFF
--- a/_includes/description.html
+++ b/_includes/description.html
@@ -1,0 +1,17 @@
+{% if include.source.description != nil %}
+  <!--for normal case description is for all specified types-->
+  {{include.source.description | markdownify}}
+
+{% else %}
+  <!-- HACK: for nullable description-->
+  <!--when we have @nullable, the format of a property schema will become:
+  {rangeStep: {anyof:{type: null}, {type: string, description: ...}}}
+  which is not correct since description only applies to one data type.
+  So we need to add this hack to read the description from different data
+  type and concatenate them-->
+  {% for type in include.source["anyOf"] %}
+    {% if type.description != nil %}
+      {{type.description | markdownify}}
+    {% endif %}
+  {% endfor %}
+{% endif %}
--- a/_includes/table.html
+++ b/_includes/table.html
@@ -11,50 +11,32 @@
 
     {% assign props = include.props | split: ","  %}
     {% for prop in props %}
-      {% assign type = schema[prop].type %}
-      {% assign description =  schema[prop].description" %}
-      {% if type == nil %}
-        <!--special case for $ref-->
-        {% if schema[prop]["$ref"] %}
-          {% assign array = schema[prop]["$ref"] | split: "/" %}
-          {% assign last_index = array | size | minus: 1 %}
-          {% assign type = array[last_index] %}
-        {% elsif schema[prop].anyOf %}
-          <!--special case for anyOf-->
-          {% capture type %}
-            {% assign para = schema[prop].anyOf %}
-            {% include type-anyof.html types=para %}
-          {% endcapture %}
-        {% endif %}
-      {% elsif type == "array" %}
-         <!--special case for type is arrays of objects -->
-        {% assign type = schema[prop].items.type | append: "[]" | capitalize %}
-      {% elsif type.first != nil %}
-        <!-- special case for type in the schema is an array of multiple types-->
-        {% assign newType = ""%}
-        {% for element in type %}
-          {% assign elem = element | capitalize %}
-          {% assign newType = newType | append: " | " | append: elem %}
-        {% endfor %}
-        {% assign type =  newType | remove_first: " | " %}
-      {% else %}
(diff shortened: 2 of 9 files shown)


src/scale.ts
```diff
@@ -299,19 +302,29 @@ export interface Scale {
...
+   *
    * @TJS-type integer
    */
   spacing?: number;
 
   /**
    * Applies spacing among ordinal elements in the scale range. The actual effect depends on how the scale is configured. If the __points__ parameter is `true`, the padding value is interpreted as a multiple of the spacing between points. A reasonable value is 1.0, such that the first and last point will be offset from the minimum and maximum value by half the distance between points. Otherwise, padding is typically in the range [0, 1] and corresponds to the fraction of space in the range interval to allocate to padding. A value of 0.5 means that the band size will be equal to the padding width. For more, see the [D3 ordinal scale documentation](https://github.com/mbostock/d3/wiki/Ordinal-Scales).
    * A convenience property for setting the inner and outer padding to the same value.
+   *
+   * The default from `x` and `y` channels are derived from [scale config](config.html#scale-config)'s `pointPadding` for `point` scale and `bandPadding` for `band` scale.  Other channels has `0` padding by default.
```
kanitw: Use `__Default value:__`

src/scale.ts
```diff
@@ -299,19 +302,29 @@ export interface Scale {
...
+   * For `row` and `column`, `padding` is ignored. Please use `spacing` instead!
```
kanitw: Please remove this line about row and column as it will be removed very soon.

src/scale.ts
```diff
@@ -333,21 +346,35 @@ export interface Scale {
...
+   *
+   * __Supported types:__ only `linear`, `pow`, `sqrt`, and `log` (Not applicable for `quantile`, `quantize`, and `threshold` scales as they output discrete ranges.)
+   *
    */
   clamp?: boolean;
   /**
-   * If specified, modifies the scale domain to use a more human-friendly value range. If specified as a true boolean, modifies the scale domain to use a more human-friendly number range (e.g., 7 instead of 6.96). If specified as a string, modifies the scale domain to use a more human-friendly value range. For time and utc scale types only, the nice value should be a string indicating the desired time interval.
+   * As quantitative scale property, if specified, modifies the scale domain to use a more human-friendly value range. If specified as a `true` boolean, modifies the scale domain to use a more human-friendly number range (e.g., 7 instead of 6.96). If specified as a string, modifies the scale domain to use a more human-friendly value range. For time and utc scale types only, the nice value should be a string indicating the desired time interval.
+   *
+   * __Default value:__ `true` only for quantitative x and y scales and `false` otherwise.
```
kanitw: Default should be at the end?

src/scale.ts
```diff
@@ -264,11 +264,14 @@ export function isExtendedScheme(scheme: string | ExtendedScheme): scheme is Ext
 }
 
 export interface Scale {
+  /**
+   * The type of scale. <br/> •  For a _quantitative_ field, supported quantitative scale types  are `"linear"` (default), `"log"`, `"pow"`, `"sqrt"`, `"quantile"`, `"quantize"`, and `"threshold"`. <br/> • For a _temporal_ field without `timeUnit`, the scale type should be `"time"` (default) or `"ordinal"`. <br/>  • For _ordinal_ and _nominal_ fields, the type is always `"ordinal"`. <br/>Unsupported values will be ignored.
```
kanitw: Since this gets markdownify, I think we should get rid of `<br/>` and manual bullet listing and instead make something like

```
The type of scale.

 - For a _quantitative_ field, supported quantitative scale types  are `"linear"` (default), `"log"`, `"pow"`, `"sqrt"`, `"quantile"`, `"quantize"`, and `"threshold"
- For a _temporal_ field without `timeUnit`, the scale type should be `"time"` (default) or `"ordinal"`.
...
```

and check if the output still looks reasonable.

_includes/table.html
```diff
@@ -11,50 +11,32 @@
...
+      {% endcapture %}
+
+      <!--extract description-->
+      {% assign src=schema[prop] %}
+      {% capture description %}
+        {% include description.html source=src %}
+      {% endcapture %}
 
-      <!--check if prop is a link or not.-->
+      <!--check if prop is a link.-->
```
kanitw: Why is channel=Axis below here?

_includes/description.html
```diff
@@ -0,0 +1,12 @@
+{% if include.source.description != nil %}
+  <!--for normal case description is for all specified types-->
+  {{include.source.description | markdownify}}
+
+{% else %}
+  <!-- HACK: for nullable description-->
```
kanitw: As mentioned in person, need more description about why you are doing this.

_includes/table.html
YuhanLu: fixed

PR #2138, opened 2017-04-14
Auto generate tables for transform

YuhanLu:
Fix part of #1723 

Might still need help from ham to get the docs done:)


domoritz:
I made some changes (please look at them). Thanks for the great work!

PR #2139, opened 2017-04-14
Dom/clone

domoritz:


src/compile/data/aggregate.ts
```diff
@@ -42,40 +42,32 @@ function mergeMeasures(parentMeasures: Dict<Dict<boolean>>, childMeasures: Dict<
...
-  /** dictionary mapping field name => dict set of aggregation functions */
-  private measures: Dict<StringSet>;
-
-  public clone(): this {
-    const cloneObj = new (<any>this.constructor);
-    cloneObj.dimensions = duplicate(this.dimensions);
-    cloneObj.measures = duplicate(this.measures);
-    return cloneObj;
+  public clone() {
+    return new AggregateNode(extend({}, this.dimensions), duplicate(this.measures));
```
kanitw: why one extend, one duplicate?

domoritz: Because one is a dict of primitives, the other is not.

PR #2141, opened 2017-04-14
Example with layer overlay (layer of layer). Fix layout and don’t ins…

domoritz:
…tantiate unused data sources.

screen shot 2017-04-14 at 09 17 36

### DIFF
--- a/src/compile/data/assemble.ts
+++ b/src/compile/data/assemble.ts
@@ -157,8 +157,11 @@ function makeWalkTree(data: VgData[]) {
    */
   function walkTree(node: DataFlowNode, dataSource: VgData) {
     if (node instanceof ParseNode) {
-      if (node.parent instanceof SourceNode && dataSource.format) {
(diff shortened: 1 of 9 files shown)


src/compile/data/assemble.ts
```diff
@@ -157,8 +157,11 @@ function makeWalkTree(data: VgData[]) {
    */
   function walkTree(node: DataFlowNode, dataSource: VgData) {
     if (node instanceof ParseNode) {
-      if (node.parent instanceof SourceNode && dataSource.format) {
-        dataSource.format.parse = node.assemble();
+      if (node.parent instanceof SourceNode) {
+        dataSource.format = {
+          ...dataSource.format || {},
```
kanitw: I wonder if format can already contain `parse` -- hopefully not.

domoritz: No, it cannot.

PR #2154, opened 2017-04-14
Correct incompatibleChannel log message

kanitw:


src/log.ts
```diff
@@ -122,7 +122,7 @@ export namespace message {
 
   export function incompatibleChannel(channel: Channel, markOrFacet: Mark | 'facet', when?: string) {
     return `${channel} dropped as it is incompatible with ${markOrFacet}` +
-      when ? `when ${when}` : '';
+      (when ? `when ${when}` : '')  + '.';
```
domoritz: I think this misses a space. I will fix it differently.

domoritz:
Check out my improvements and merge if you like them.

src/log.ts
```diff
@@ -121,8 +121,7 @@ export namespace message {
   }
 
   export function incompatibleChannel(channel: Channel, markOrFacet: Mark | 'facet', when?: string) {
-    return `${channel} dropped as it is incompatible with ${markOrFacet}` +
-      (when ? `when ${when}` : '')  + '.';
+    return `${channel} dropped as it is incompatible with ${markOrFacet}${when ? ` when ${when}` : ''}.`;
```
kanitw: I don't like nested string literals

kanitw: But it's not a big deal. I'll just merge!

Issue #2144, opened 2017-04-14
Adding size encoding to rule crashes


jheer:
Expected behavior:
- Rule mark line width is sized by value
- Or, size encoding is ignored (perhaps with console warning)

Actual behavior:
- Crash!

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A bar chart showing the US population distribution of age groups in 2000.",
  "data": {
    "url": "data/population.json"
  },
  "transform": [
    {
      "filter": "datum.year == 2000"
    }
  ],
  "layer": [
    {
      "mark": "bar",
      "encoding": {
        "y": {
          "field": "age",
          "type": "ordinal",
          "scale": { "rangeStep": 17 }
        },
        "x": {
          "aggregate": "sum",
          "field": "people",
          "type": "quantitative",
          "axis": { "title": "population" }
        },
        "color": { "value": "steelblue" }
      }
    },
    {
      "mark": "rule",
      "encoding": {
        "x": {
          "aggregate": "average",
          "field": "people",
          "type": "quantitative"
        },
        "size": {"value": 1000},
        "color": { "value": "red" }
      }
    }
  ]
}
```

domoritz:
Instead of setting the rule width to 1000, set it to something smaller. 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A bar chart showing the US population distribution of age groups in 2000.",
  "data": {
    "url": "data/population.json"
  },
  "transform": [
    {
      "filter": "datum.year == 2000"
    }
  ],
  "layer": [
    {
      "mark": "bar",
      "encoding": {
        "y": {
          "field": "age",
          "type": "ordinal",
          "scale": { "rangeStep": 17 }
        },
        "x": {
          "aggregate": "sum",
          "field": "people",
          "type": "quantitative",
          "axis": { "title": "population" }
        },
        "color": { "value": "steelblue" }
      }
    },
    {
      "mark": "rule",
      "encoding": {
        "x": {
          "aggregate": "average",
          "field": "people",
          "type": "quantitative"
        },
        "size": {"value": 5},
        "color": { "value": "red" }
      }
    }
  ]
}
```

![screen shot 2017-04-14 at 14 06 45](https://cloud.githubusercontent.com/assets/589034/25056116/a8081cd6-211b-11e7-8a4d-549e02a2fcd3.png)

kanitw:
[...]

PR #2156, opened 2017-04-14
Rule should not support size

kanitw:


kanitw:
Fix https://github.com/vega/vega-lite/issues/2144

PR #2157, opened 2017-04-14
Fixes #2149

### LINKED ISSUE #2149: Generated Vega for binned plot references non-existent properties
After performing binning, the Vega generated for the Vega-Lite spec below includes `..._range` fields that are not generated by the `bin` transform. This may be vestigial from VL1? (Also, the aggregate can be made more efficient in this case by using the `key` parameter to optimize groupby cell key generation.)

```json
{
  "data": {
    "url": "data/movies.json"
  },
  "mark": "circle",
  "encoding": {
    "x": {
      "bin": {
        "maxbins": 10
      },
      "field": "IMDB_Rating",
      "type": "quantitative"
    },
    "y": {
      "bin": {
        "maxbins": 10
      },
      "type": "quantitative"
    },
    "size": {
      "aggregate": "count",
      "type": "quantitative"
    }
  }
}
```

domoritz:


src/compile/data/aggregate.ts
```diff
@@ -11,11 +12,9 @@ function addDimension(dims: {[field: string]: boolean}, fieldDef: FieldDef) {
     dims[field(fieldDef, {binSuffix: 'start'})] = true;
     dims[field(fieldDef, {binSuffix: 'end'})] = true;
 
-    // const scale = model.scale(channel);
-    // if (scaleType(scale, fieldDef, channel, model.mark()) === ScaleType.ORDINAL) {
-    // also produce bin_range if the binned field use ordinal scale
-    dims[field(fieldDef, {binSuffix: 'range'})] = true;
-    // }
+    if (fieldDef.type === ORDINAL) {
```
kanitw: why only fieldDef.type === ORDINAL?

domoritz: We need the range only when the user explicitly forces a binned field to be ordinal (range used in axis and legend labels). We could check whether the axis or legend exists but that seems overkill.

domoritz: We check for `hasDiscreteDomain(scale.type)` in other places and I believe that is equivalent to an ordinal type, right?

kanitw: They are not equivalent, this can be binned ordinal or temporal that uses an ordinal scale.  
Given you ask this question, I suspect that this fix is incorrect.  
Let me take a look.

kanitw: Ok -- this fix is correct because it's in bin so it can only be `bin-ordinal`. 

However, `hasDiscreteDomain` does not always imply `type === ORDINAL`.

domoritz: But we are in the context of binned data. I still think it's correct but thanks for looking over it.

PR #2158, opened 2017-04-14
Rule should not support size

kanitw:
Prevent #2144


domoritz:
I don't like this. Rule should support size (and actually does in the one example that changes on travis.

domoritz:
It actually works https://github.com/vega/vega-lite/issues/2144.

PR #2159, opened 2017-04-14
Clean up model properties.

domoritz:

### DIFF
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -36,15 +36,6 @@ export class FacetModel extends Model {
   public readonly child: Model;
 
   public readonly children: Model[];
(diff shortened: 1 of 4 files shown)


kanitw:
I'm confused, why are you moving these into Model? Should we just remove it from LayerModel?

domoritz:
I'm not moving anything. They were defined in model before. We can remove them when only unit needs them.

kanitw:
After we move facet to use the new layout, we'll basically have to revert half of this -- so let's say I'll take care of this later as this doesn't make my life easier.

domoritz:
No, that's not true. You just cut and paste from model to unit. The previous state was worse because we had tons of copies.

PR #2160, opened 2017-04-14
Support selections with layered specs

### LINKED ISSUE #2146: Selection unresponsive in layer plot
Expected behavior: x-projected brush interval selection
Actual behavior: no brush

```json
{
  "description": "A bar chart showing the US population distribution of age groups in 2000.",
  "data": {
    "url": "data/population.json"
  },
  "transform": [
    {
      "filter": "datum.year == 2000"
    }
  ],
  "layer": [{
  "mark": "bar",
  "selection": {
    "brush": {
      "type": "interval",
      "encodings": [
        "x"
      ]
    }
  },
  "encoding": {
    "y": {
      "field": "age",
      "type": "ordinal",
      "scale": { "rangeStep": 17 }
    },
    "x": {
      "aggregate": "sum",
      "field": "people",
      "type": "quantitative",
      "axis": { "title": "population" }
    },
    "color": { "value": "steelblue" }
  }
  },{
    "mark": "rule",
    "encoding": {
      "x": {
        "aggregate": "average",
        "field": "people",
        "type": "quantitative"
      },
      "color": { "value": "red" }
    }
  }]
}
```

arvind:
Fixes #2146. The following spec demonstrates selections defined in both layers and used across them. 

```json
{
  "layer": [{
    "selection": {
      "grid": {
        "type": "interval", "bind": "scales",
        "translate": "[mousedown[!event.shiftKey], mouseup] > mousemove"
      }
    },
    "mark": "point",
    "encoding": {
      "color": {
        "condition": {"selection": "brush", "value": "grey"},
        "field": "Cylinders", "type": "ordinal"
      },
      "size": {"value": 100}
    }
  }, {
    "mark": "square",
      "selection": {
      "brush": {
        "type": "interval",
        "on": "[mousedown[event.shiftKey], mouseup] > mousemove",
        "translate": "[mousedown[event.shiftKey], mouseup] > mousemove"
      }
    },
    "encoding": {
      "color": {
        "condition": {"selection": "!brush", "value": "grey"},
        "field": "Cylinders", "type": "ordinal"
      }
    }
  }]
}
```

### DIFF
--- a/src/compile/layer.ts
+++ b/src/compile/layer.ts
@@ -71,8 +71,13 @@ export class LayerModel extends Model {
   }
 
   public parseSelection() {
-    // TODO: @arvind can write this
-    // We might need to split this into compileSelectionData and compileSelectionSignals?
+    this.component.selection = {};
+    this.children.forEach(child => {
+      child.parseSelection();
+      keys(child.component.selection).forEach((key) => {
+        this.component.selection[key] = child.component.selection[key];
+      });
+    });
   }
 
   public parseLayoutData() {
@@ -174,12 +179,13 @@ export class LayerModel extends Model {
     return applyConfig({}, cellConfig, FILL_STROKE_CONFIG.concat(['clip']));
   }
 
+  // TODO: Support same named selections across children.
   public assembleSignals(signals: any[]): any[] {
-    return [];
+    return this.children.reduce((signals, child) => child.assembleSignals(signals), []);
   }
 
   public assembleSelectionData(data: VgData[]): VgData[] {
-    return [];
+    return this.children.reduce((data, child) => child.assembleSelectionData(data), []);
   }
 
   public assembleData(): VgData[] {
--- a/src/compile/layer.ts
+++ b/src/compile/layer.ts
@@ -17,7 +17,7 @@ import {assembleLayout, parseLayerLayout} from './layout';
 import {Model} from './model';
 import {unionDomains} from './scale/domain';
 import {UnitModel} from './unit';
-
+import {assembleLayerMarks} from './selection/selection';
 
 export class LayerModel extends Model {
   public readonly children: UnitModel[];
@@ -213,9 +213,9 @@ export class LayerModel extends Model {
 
   public assembleMarks(): any[] {
     // only children have marks
-    return flatten(this.children.map((child) => {
+    return assembleLayerMarks(this, flatten(this.children.map((child) => {
       return child.assembleMarks();
-    }));
+    })));
   }
 
   public channels(): Channel[] {
--- a/src/compile/layer.ts
+++ b/src/compile/layer.ts
@@ -16,7 +16,7 @@ import {parseData} from './data/parse';
 import {assembleLayout, parseLayerLayout} from './layout';
 import {Model} from './model';
 import {unionDomains} from './scale/domain';
-import {assembleLayerMarks} from './selection/selection';
+import {assembleLayerMarks as assembleLayeredSelectionMarks} from './selection/selection';
 import {UnitModel} from './unit';
 
 export class LayerModel extends Model {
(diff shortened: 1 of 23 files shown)


arvind:
Travis is failing based on diffs with a number of the output Vega specs. I assume that's leftover from the multiview work rather than something particular to this PR?

src/compile/layer.ts
```diff
@@ -184,12 +189,13 @@ export class LayerModel extends Model {
     return applyConfig({}, cellConfig, FILL_STROKE_CONFIG.concat(['clip']));
   }
 
+  // TODO: Support same named selections across children.
```
domoritz: Can you add the name of the model to the key to make this work? Call `model.getName('selection name')`.

src/compile/selection/selection.ts
```diff
@@ -150,42 +151,41 @@ export function assembleTopLevelSignals(model: Model) {
 }
 
 export function assembleUnitData(model: UnitModel, data: VgData[]): VgData[] {
-  return data
-    .concat(Object.keys(model.component.selection)
-      .map(function(k: string) {
-        return {name: k + STORE};
-      }));
+  forEachSelection(model, function(selCmpt, _) {
+    data.push.apply(data, [{name: selCmpt.name + STORE}]);
```
domoritz: Is this the same as `data.push({name: selCmpt.name + STORE})`.

domoritz: If not, revert my change ;-)

src/compile/layer.ts
arvind: No, that's not what I meant by this comment. Rather, if I have a selection named the same across multiple concatenated views, it should populate the same top-level state in Vega. I haven't yet reasoned through it all in my head, so I left the TODO in here to remind me once I've resolved the higher priority items.

domoritz: Thanks for the clarification.

src/compile/selection/selection.ts
```diff
@@ -150,42 +151,41 @@ export function assembleTopLevelSignals(model: Model) {
...
-          width: {field: {group: 'width'}},
-          height: {field: {group: 'height'}},
-          fill: {value: 'transparent'},
-          clip: {value: true}
-        }
-      },
-      marks: selMarks.map(model.correctDataNames)
-    }];
+  if (model.parent && model.parent instanceof LayerModel) {
+    return [selMarks, clippedGroup];
```
kanitw: I see that you later check `unit[1]` in `assembleLayerMarks` but not sure why.

Can you add a comment (either here or in `assembleLayerMarks` or both)  to explain why if the parent is a `LayerModel`, we have to set this up differently like this?

src/compile/layer.ts
```diff
@@ -71,8 +71,13 @@ export class LayerModel extends Model {
   }
 
   public parseSelection() {
-    // TODO: @arvind can write this
-    // We might need to split this into compileSelectionData and compileSelectionSignals?
+    this.component.selection = {};
+    this.children.forEach(child => {
+      child.parseSelection();
+      keys(child.component.selection).forEach((key) => {
```
kanitw: If I understand correctly, we always merge selection to the parent because selections exist across the entire hierarchy?  If so, maybe add a comment to explain.

I'm also curious why you don't need to remove the `selection` from `child.component`? Wouldn't this cause problem if you resolve and merge the selection later in the upper layers?  (I might have overlooked something, but probably a good idea to ask to clarify.)

src/compile/layer.ts
```diff
@@ -207,9 +213,9 @@ export class LayerModel extends Model {
 
   public assembleMarks(): any[] {
     // only children have marks
-    return flatten(this.children.map((child) => {
+    return assembleLayerMarks(this, flatten(this.children.map((child) => {
```
kanitw: Can we just rename this `assembleLayerSelectionMarks`?   A little more verbose, but clearer.

I think the old comment `// only children have marks` might not make sense any more too.

src/compile/unit.ts
```diff
@@ -259,7 +260,9 @@ export class UnitModel extends Model {
 
   public assembleMarks() {
     let marks = this.component.mark || [];
-    marks = assembleSelectionMarks(this, marks);
+    if (!this.parent || !(this.parent instanceof LayerModel)) {
```
kanitw: Need a comment why we don't have to do this for `LayerModel`

domoritz:
Wohoo 🎉

PR #2163, opened 2017-04-15
Support selections with faceted specs

arvind:
Follows #2160. Here's a test spec:

```json
{
  "description": "Anscombe's Quartet",
  "data": {"url": "data/anscombe.json"},
  "mark": "circle",
  "selection": {
    "brush": {
      "type": "interval", "resolve": "global",
      "encodings": ["x"]
    }
  },
  "encoding": {
    "column": {"field": "Series","type": "nominal"},
    "x": {
      "field": "X",
      "type": "quantitative",
      "scale": {"zero": false}
    },
    "y": {
      "field": "Y",
      "type": "quantitative",
      "scale": {"zero": false}
    },
    "size": {
      "value": 100,
      "condition": {"selection": "brush","value": 250}
    }
  },
  "config": {"mark": {"opacity": 1}}
}
```

Things to vary:
* Resolution scheme: `global`, `independent`, `union`, `union_others`, `intersect`, `intersect_others`.
* Binding should only work with `global` resolutions.

### DIFF
--- a/src/compile/selection/interval.ts
+++ b/src/compile/selection/interval.ts
@@ -1,8 +1,8 @@
 import {Channel, X, Y} from '../../channel';
 import {warn} from '../../log';
-import {extend, stringValue, keys} from '../../util';
+import {extend, keys, stringValue} from '../../util';
 import {UnitModel} from '../unit';
-import {channelSignalName, invert as invertFn, SelectionCompiler, SelectionComponent, TUPLE, STORE} from './selection';
+import {channelSignalName, invert as invertFn, SelectionCompiler, SelectionComponent, STORE, TUPLE} from './selection';
 import scales from './transforms/scales';
 
 export const BRUSH = '_brush',
@@ -105,7 +105,7 @@ const interval:SelectionCompiler = {
         update[key] = [{
           test: `${store}.length && ${tpl} && ${tpl}.unit === ${store}[0].unit`,
           ...update[key]
-        }, {value: 0}]
+        }, {value: 0}];
       });
     }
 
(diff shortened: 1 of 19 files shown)


arvind:
[This line](https://travis-ci.org/vega/vega-lite/jobs/222256827#L351) of the failing test cases doesn't look right, but I'm not sure how exactly to fix `spec.ts`. Not sure why selection would be an array of `undefined`... Any advice would be appreciated!

domoritz:
> This line of the failing test cases doesn't look right, but I'm not sure how exactly to fix spec.ts. Not sure why selection would be an array of undefined... Any advice would be appreciated!

Fixed in https://github.com/vega/vega-lite/pull/2163/commits/07833d81951822afbbc1c648fc349963b9cf8b9b

src/compile/selection/interval.ts
```diff
@@ -94,9 +97,18 @@ const interval:SelectionCompiler = {
 
       y2: extend({}, y !== null ?
         {scale: model.scaleName(Y), signal: `${name}[${y}].extent[1]`} :
-        {field: {group: 'height'}}),
+        {field: {group: 'height'}})
     };
 
+    if (selCmpt.resolve === 'global') {
+      keys(update).forEach(function(key) {
+        update[key] = [{
```
kanitw: Maybe add some comments here?
A few things I don't understand:
- I assume `.unit` mean something in Vega Expression that's useful for selection resolution?  It would be great if we could know what it means.
- Why `${store}`needs to refer to `[0]` ?

PR #2164, opened 2017-04-15
Use instanceof instead of is*

domoritz:


kanitw:
Haha, why didn't we use `instanceof` from the beginning.

domoritz:
Yeah, right :-)

Issue #1544, opened 2016-09-07
Merge axis groups when facet child is not unit


kanitw:
(There is no overlayed points)

kanitw:
For example:

```
{
  "description": "Google's stock price over time.",
  "data": {"url": "data/stocks.csv", "format": {"type": "csv"}},
  
  "mark": "line",
  "encoding": {
    "row": {"field": "symbol", "type": "nominal"},
    "x": {"field": "date", "type": "temporal"},
    "y": {"field": "price", "type": "quantitative"}
  },
  "config": {
    "overlay": {
      "line": true
    }
  }
}

```

kanitw:
The normalizer logic should is corrected as of #2084 (I wrote a test for it).  
However, a bug in layer seems to prevent it from working.  

We now got the following error:
```
Uncaught Error: Undefined data set name: "child_source"
```

Since this should be fixed with the new transform merging, I'm assigning this to @domoritz to double check this spec once the transform refactor is done.

domoritz:
Kinda works with #2086

![download 1](https://cloud.githubusercontent.com/assets/589034/25018983/d6dabde0-203d-11e7-9f24-82675c72b0f6.png)

kanitw:
not yet lol

domoritz:
The problem is https://github.com/vega/vega-lite/blob/463299bc94920860ad4f81240b599d661ac36d22/src/compile/facet.ts#L354

domoritz:
More examples

```json
{
  "data": {"url": "data/seattle-weather.csv","formatType": "csv"},
  "facet": {"column": {"field": "weather","type": "nominal"}},
  "spec": {
    "layer": [
      {
        "mark": "line",
        "encoding": {
          "x": {
            "field": "date",
            "type": "temporal",
            "timeUnit": "month"
          },
          "y": {
            "field": "temp_max",
            "type": "quantitative",
            "aggregate": "mean"
          }
        }
      },
      {
        "mark": "circle",
        "encoding": {
          "x": {
            "field": "date",
            "type": "temporal",
            "timeUnit": "month"
          },
          "y": {
            "field": "temp_max",
            "type": "quantitative",
            "aggregate": "mean"
          }
        }
      }
    ]
  }
}
```


```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
[...]

PR #2168, opened 2017-04-15
[WIP] Re-build Facet with the new Layout Operator

kanitw:
See commit message for details

## To Follow-up
- [ ] ~~Make FacetFieldDef support `sort`~~ #2176
- [x] Rebuild Labels and Title (now that the axis has gone)
  - [x] labels group
  - [x] title 
  - [ ] Re-build using `row-title` and `column-title`
- [ ] Correct write `parseAxis` logic
- [ ] Deal with non-unit child in `parseAxisGroups`
- [x] correctly set `column` of `layout`

- [ ] allow customization
  - [ ] `padding: {header: number, row: number, column: number}`
  - [ ] Axis Title, Axis Labels

### DIFF
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -386,6 +386,9 @@ export function getLabelGroup(model: FacetModel, channel: 'row' | 'column') {
     name: model.getName(`${channel}-labels`),
     from: {data: model.getName(channel)},
     groupEncode: childSizeEncodeEntryMixins(model, sizeChannel),
+    // TODO: support customizing + remove if Vega will provide padding for this one.
+    offset: -10,
+    // TODO: support customizing row title orientation (horizontal / vertical (using textOrient)
     textRole: `${channel}-labels`,
     textRef: {field: {parent: model.field(channel)}},
     positionRef: {field: {group: sizeChannel}, mult: 0.5}
@@ -398,6 +401,9 @@ export function getTitleGroup(model: FacetModel, channel: 'row' | 'column') {
   return getTextHeader({
     channel,
     name: model.getName(`${channel}-title`),
+
+    // TODO: support customizing row title orientation (horizontal / vertical)
+    textOrient: (channel === 'row' ? 'vertical' : undefined),
     textRole: `${channel}-labels`,
     textRef: {value: fieldDefTitle(fieldDef, model.config)},
     positionRef: {signal: `0.5 * ${sizeChannel}`}
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -1,7 +1,7 @@
 import {Axis} from '../axis';
 import {Channel, COLUMN, ROW, X, Y} from '../channel';
 import {Config} from '../config';
-import {MAIN} from '../data';
+import {LAYOUT, MAIN} from '../data';
 import {reduce} from '../encoding';
 import {Facet} from '../facet';
 import {FieldDef, normalize} from '../fielddef';
@@ -203,6 +203,15 @@ export class FacetModel extends ModelWithField {
   }
 
   public assembleSignals(signals: any): any[] {
+    if (this.channelHasField('column')) {
+      const columnDistinct = this.field('column',  {prefix: 'distinct'});
+      return [
+        {
+          name: this.getName('column'),
+          update: `data('${this.getName(LAYOUT)}')[0].${columnDistinct}`
+        }
+      ];
+    }
     return [];
   }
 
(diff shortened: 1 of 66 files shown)


src/compile/data/optimizers.ts
```diff
@@ -15,10 +15,10 @@ import {CalculateNode, FilterNode} from './transforms';
  */
 export function optimizeFromLeaves(f: (node: DataFlowNode) => void) {
   function optimizeNextFromLeaves(node: DataFlowNode) {
-    if (node.parent instanceof SourceNode || node instanceof SourceNode) {
+    if (node.parent instanceof SourceNode) {
       return;
     } else if (!node || !node.parent) {
-      throw new Error('A source node cannot have parents and roots haev to be source nodes.');
+      throw new Error('Bad state: a source node cannot have parents and roots have to be source nodes.');
```
kanitw: I still think that this error is very confusing. 

In this branch you can only conclude that
1. Node.parent is not a SourceNode.
2. Either `!node` or `!node.parent`.
For the former one (!node), it's very weird why would that case happen at all.
For the latter (`!node.parent`), yes this node is a root.  While you can infer that root should be a source node (comment missing), you cannot say that this source node has parents.

domoritz:
https://github.com/vega/vega-lite/pull/2174 has better logic for resolving the problem with parse.

src/compile/data/assemble.ts
```diff
@@ -156,6 +156,19 @@ function makeWalkTree(data: VgData[]) {
    * Recursively walk down the tree.
    */
   function walkTree(node: DataFlowNode, dataSource: VgData) {
+    if (node instanceof ParseNode) {
+      if (node.parent instanceof SourceNode && !node.parent.parent)  {
+        // If its parent is a root source node, use normal format parse
+        dataSource.format = {
```
kanitw: I rebased #2168 to #2174 and found that even though a parse node's parent is a source, it does not mean that the `dataSource` we are assembling here is the sourceNode's data source.  In #2168, for box_plot example and several others, `dataSource` here are new data sources that depends on the sourceNode's data source.

domoritz: Yeah, that's the bug I was talking about yesterday. Which is exactly why I added the parse expression stuff. Basically, when there si a fork, you have to use parse as expressions.

domoritz:
bar_grouped looks funky now

screen shot 2017-04-16 at 12 26 05

domoritz:
Why do we still need to set the group dimensions?

```json
"encode": {
        "update": {
          "width": {"field": "width"},
          "height": {"field": "height"}
        }
      },
```

domoritz:
variety is bold but site is not

screen shot 2017-04-16 at 12 32 48

domoritz:
Second example from https://github.com/vega/vega-lite/issues/1544#issuecomment-294271110 now works.

domoritz:
```json
{
  "data": {"url": "data/seattle-weather.csv","formatType": "csv"},
  "facet": {"row": {"field": "weather","type": "nominal"}},
  "spec": {
    "layer": [
      {
        "mark": "line",
        "encoding": {
          "x": {
            "field": "date",
            "type": "temporal",
            "timeUnit": "month"
          },
          "y": {
            "field": "temp_max",
            "type": "quantitative",
            "aggregate": "mean"
          }
        }
      },
      {
        "mark": "circle",
        "encoding": {
          "x": {
            "field": "date",
            "type": "temporal",
            "timeUnit": "month"
          },
          "y": {
            "field": "temp_max",
            "type": "quantitative",
            "aggregate": "mean"
          }
        }
      }
    ]
  }
}
```

produces

screen shot 2017-04-16 at 12 38 35

src/compile/model.ts
```diff
@@ -387,3 +325,42 @@ export abstract class Model {
     return this.component[type][name] || this.parent.getComponent(type, name);
   }
 }
+
+export abstract class ModelWithField extends Model {
```
domoritz: Can you add a comment that says what this class is used for?

src/facet.ts
```diff
@@ -1,4 +1,7 @@
-import {FacetFieldDef} from './fielddef';
+import {FieldDef} from './fielddef';
+
+// TODO: add more facet properties
+export type FacetFieldDef = FieldDef;
```
domoritz: The type has to be ordinal, nominal, or quantitative with bin, or time with timeunit. Should we enforce that?

kanitw: I fear that it won't be incompatible with other `FieldDef`s for the rest of the codebase.

src/compile/facet.ts
```diff
@@ -297,9 +288,6 @@ export class FacetModel extends Model {
     return this.facet;
   }
 
-  public spacing(channel: Channel) {
-    return this._spacing[channel];
-  }
 
   public isFacet() {
```
domoritz: delete this

PR #2169, opened 2017-04-15
Use new Vega `layout`

kanitw:
- Rename `assembleLayout()` to `assembleLayoutData()` and add assembleLayout() for really assembling the new `layout` object.

- We can re-use this infrastructure for all of `facet`, `concat`, and `repeat`.

- We can make this branch the new merge target until we have a new Vega release.  (As after we add layout, Travis won't pass until then.)

## TODO

- [ ] Bring `main-group` back for necessary cases
- [ ] Test if interactive specs are still working
- [x] Wait for Vega release


src/compile/facet.ts
```diff
@@ -256,9 +257,14 @@ export class FacetModel extends Model {
...
+  public assembleLayout(): VgLayout {
+    // TODO: add layout
+    return null;
+  }
+
+  public assembleLayoutData(layoutData: VgData[]): VgData[] {
     // Postfix traversal – layout is assembled bottom-up
-    this.child.assembleLayout(layoutData);
+    this.child.assembleLayoutData(layoutData);
     return assembleLayout(this, layoutData);
```
domoritz: rename?

domoritz:
LGTM

domoritz:
So now we are only blocked by a vega beta 29 release, wohoo.

kanitw:
And the main-group issue that i will take a stab tmr

PR #2174, opened 2017-04-16
Implement format parse as expressions.

domoritz:
Why: Avoids problems with merging up parse and allows us to have parse anywhere is the dataflow tree (not just next to source). 

Doesn't work until we have https://github.com/vega/vega-expression/issues/13

### DIFF
--- a/examples/vg-specs/bar_grouped.vg.json
+++ b/examples/vg-specs/bar_grouped.vg.json
@@ -28,7 +28,9 @@
             "url": "data/population.json",
             "format": {
                 "type": "json",
-                "parse": {}
+                "parse": {
+                    "people": "number"
+                }
             },
             "transform": [
                 {
@@ -39,11 +41,6 @@
                     "type": "formula",
                     "expr": "datum.sex == 2 ? 'Female' : 'Male'",
                     "as": "gender"
-                },
-                {
-                    "type": "formula",
-                    "expr": "toNumber(datum[\"people\"])",
-                    "as": "people"
                 },
                 {
                     "type": "filter",
--- a/examples/vg-specs/bar_grouped.vg.json
+++ b/examples/vg-specs/bar_grouped.vg.json
@@ -28,9 +28,7 @@
             "url": "data/population.json",
             "format": {
                 "type": "json",
-                "parse": {
-                    "people": "number"
-                }
+                "parse": {}
             },
             "transform": [
                 {
@@ -41,6 +39,11 @@
                     "type": "formula",
                     "expr": "datum.sex == 2 ? 'Female' : 'Male'",
                     "as": "gender"
+                },
+                {
+                    "type": "formula",
+                    "expr": "toNumber(datum[\"people\"])",
+                    "as": "people"
                 },
                 {
                     "type": "filter",
--- a/src/compile/data/assemble.ts
+++ b/src/compile/data/assemble.ts
@@ -158,7 +158,7 @@ function makeWalkTree(data: VgData[]) {
   function walkTree(node: DataFlowNode, dataSource: VgData) {
     if (node instanceof ParseNode) {
       if (node.parent instanceof SourceNode && !dataSource.source)  {
-        // If node's parent is a root source and the data source does not referencence another data source, use normal format parse
+        // If node's parent is a root source and the data source does not refer to another data source, use normal format parse
         dataSource.format = {
           ...dataSource.format || {},
           parse: node.assembleFormatParse()
--- a/src/compile/data/assemble.ts
+++ b/src/compile/data/assemble.ts
@@ -10,7 +10,7 @@ import {FacetNode} from './facet';
 import {ParseNode} from './formatparse';
 import {NonPositiveFilterNode} from './nonpositivefilter';
 import {NullFilterNode} from './nullfilter';
-import {optimizeFromLeaves} from './optimizers';
+import {iterateFromLeaves} from './optimizers';
 import * as optimizers from './optimizers';
 import {OrderNode} from './pathorder';
 import {SourceNode} from './source';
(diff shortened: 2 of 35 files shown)


examples/vg-specs/bar_grouped.vg.json
```diff
@@ -28,9 +28,7 @@
             "url": "data/population.json",
             "format": {
                 "type": "json",
-                "parse": {
-                    "people": "number"
-                }
+                "parse": {}
```
kanitw: @domoritz  not sure if we better output nothing or `{}`.  

(If I understand correctly,  `{}` means don't apply auto-parsing?)

src/compile/data/optimizers.ts
```diff
@@ -11,14 +11,12 @@ import {TimeUnitNode} from './timeunit';
 import {CalculateNode, FilterNode} from './transforms';
 
 /**
- * Start optimization path at the leaves. Useful for merging up things.
+ * Start optimization path at the leaves. Useful for merging up or removing things.
  */
-export function optimizeFromLeaves(f: (node: DataFlowNode) => void) {
+export function iterateFromLeaves(f: (node: DataFlowNode) => void) {
```
kanitw: @domoritz  I wonder why you have to change the logic here since this function seems unused. 

This cause some conflicts with https://github.com/vega/vega-lite/pull/2168/commits/29ebae677239d2048dab7a99c8010ad516704570 in #2168. 
(I also wonder if I still need that `removeUnusedSubtrees` logic?)

domoritz: We don't need the remove subtree logic per se but it's correct.

examples/vg-specs/bar_grouped.vg.json
domoritz: I forgot this case. Let me check.

src/compile/data/assemble.ts
```diff
@@ -156,6 +156,19 @@ function makeWalkTree(data: VgData[]) {
    * Recursively walk down the tree.
    */
   function walkTree(node: DataFlowNode, dataSource: VgData) {
+    if (node instanceof ParseNode) {
+      if (node.parent instanceof SourceNode && !node.parent.parent)  {
```
domoritz: drop `&& !node.parent.parent` as a source node has to be a root. I will add code to enforce that in source nodes.

examples/vg-specs/bar_grouped.vg.json
domoritz: Looks like it is https://vega.github.io/vega/docs/data/#format. But I can't tell for sure. 

Whatever the logic is, we do not want auto parse.

src/compile/data/assemble.ts
```diff
@@ -157,8 +157,8 @@ function makeWalkTree(data: VgData[]) {
    */
   function walkTree(node: DataFlowNode, dataSource: VgData) {
     if (node instanceof ParseNode) {
-      if (node.parent instanceof SourceNode && !node.parent.parent)  {
-        // If its parent is a root source node, use normal format parse
+      if (node.parent instanceof SourceNode && !dataSource.source)  {
+        // If node's parent is a root source and the data source does not referencence another data source, use normal format parse
```
kanitw: "referencence" => refer to   :p

domoritz: Fixed

examples/vg-specs/bar_grouped.vg.json
kanitw: I think this is correct then.

src/compile/data/formatparse.ts
```diff
@@ -1,14 +1,34 @@
...
+    return `toNumber(${f})`;
+  } else if (parse === 'boolean') {
+    return `toBoolean(${f})`;
+  } else if (parse === 'string') {
+    return `toString(${f})`;
+  } else if (parse === 'date') {
+    return `toDate(${f})`;
+  } else if (parse.indexOf('date:') === 0) {
+    const specifier = parse.slice(6, parse.length - 1);  // specifier is in "" or ''
+    return `timeParse(${f},"${specifier}")`;
```
kanitw: why not using ' anymore, I saw that you use ' earlier.  The output will be cleaner.

domoritz:
This may pass Travis but still needs a new version of vega for expressions to work correctly.

src/compile/data/optimizers.ts
```diff
@@ -11,34 +11,63 @@ import {TimeUnitNode} from './timeunit';
...
   const parent = node.parent;
 
   // move parse up by merging or swapping
   if (node instanceof ParseNode) {
+    if (parent instanceof SourceNode) {
+      return false;
+    }
+
+    if (parent.numChildren() > 1) {
+      return true;
```
kanitw: I guess this return true because it will continue move other parse upstream up?  If so, maybe add comment?

src/compile/data/optimizers.ts
```diff
@@ -11,34 +11,63 @@ import {TimeUnitNode} from './timeunit';
...
+
+/**
+ * Repeatedly remove leaf nodes that are not output nodes.
+ * The reason is that we don't need subtrees that don't have any output nodes.
+ */
+export function removeUnusedSubtrees(node: DataFlowNode) {
+  const parent = node.parent;
+
+  if (node instanceof OutputNode || node.numChildren() > 0) {
+    // no need to continue with parent because it is output node or will have children (there was a fork)
```
kanitw: I'm confused.   Shouldn't fork mean `numChildren() > 1`?

kanitw:
> This may pass Travis but still needs a new version of vega for expressions to work correctly.


Ok the question is if this passes.  Do we merge or just close this PR since it's already a part of `layout` branch?

src/compile/data/formatparse.ts
domoritz: Then we should sue it everywhere (whenever we generate `datum[...]`). Moved to https://github.com/vega/vega-lite/issues/2179

src/compile/data/optimizers.ts
domoritz: Yep. I think that's clear because returning true means continuing. I don't want to add too many comments.

src/compile/data/optimizers.ts
domoritz: There was a fork before we removed the node in the previous iteration.

domoritz:
The question is whether @jheer will release the next version of vega with layout support or only the new parse expression functions.

src/compile/data/optimizers.ts
kanitw: Then make the comment clearer.

src/compile/data/optimizers.ts
kanitw: I think what's unclear is why continuing.  
(We need to continue to move other ParseNode upstream.)

src/compile/data/formatparse.ts
kanitw: But this one is not datum.

jheer:
Next version of Vega will include layout and expression updates. Earliest release date is Monday 4/17.

kanitw:
Closing as this is already a part of `layout` branch in #2169.

PR #2178, opened 2017-04-16
Remove All Inline Examples in Docs

### LINKED ISSUE #2098: Do not use inline example in the documentation
For all markdown files inside `site`, let's no longer use inline examples.

For example, `bin.md` has this inline example.
```
<div class="vl-example">
{
  "mark": "point",
  "encoding": {
    "x": {
      "field": "IMDB_Rating",
      "type": "quantitative"
    }
  }
}
</div>
```

We should just convert all of these into an example in `examples/specs` and refer to it using `data-name` like this:

```
<div class="vl-example" data-name="point_1d_IMDB_Rating"></div>
```

If you need suggestion for naming any new file (from inline examples), ask us :)

NeelMohapatra:
Fixes #2098

- [x] Remove all inline examples
- [x] Remove support for them


site/docs/bin.md
```diff
@@ -35,59 +35,17 @@ If `bin` is `true`, default binning parameters are used. To customize binning pa
...
-  "encoding": {
-    "x": {
-      "field": "IMDB_Rating",
-      "type": "quantitative"
-    }
-  }
-}
-</div>
-
+<div class='vl-example' data-name='point_1d'></div>
```
domoritz: Please use `"` instead of `'`

site/tutorials/explore.md
```diff
@@ -18,173 +18,42 @@ To load the CSV file with Vega-Lite, we need to provide a URL and set the format
...
-"transform": {
-  "calculate": [
-    {
-      "field": "temp_range",
-      "expr": "datum.temp_max - datum.temp_min"
-    }
-  ]
-}
+"transform": [
+  {"calculate": "datum.temp_max - datum.temp_min","as": "temp_range"}
```
domoritz: Make it multiline for better readability

PR #2190, opened 2017-04-16
encoding docs

kanitw:
@yuhanlu  I think we should use table generator for this page too.  Could you please take over this PR?

I split it into a bit more sections, which would fit better with your table generator.

Thanks

### DIFF
--- a/_includes/table.html
+++ b/_includes/table.html
@@ -29,7 +29,7 @@
       endcapture %}{%
       capture typeWithLink %}{%
         include checkLink.html check="type" channel=include.source param=type %}{%
-      endcapture%}
+      endcapture %}
       <tr>
         <td style="text-align: left">{{propWithLink}}</td>
         <td style="text-align: center">{{typeWithLink}}</td>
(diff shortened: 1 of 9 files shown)


YuhanLu:
Done

_includes/table.html
```diff
@@ -29,8 +29,8 @@
       endcapture %}{%
       capture typeWithLink %}{%
         include checkLink.html check="type" channel=include.source param=type %}{%
-      endcapture
-    %}<tr>
+      endcapture%}
```
domoritz: add space before %

src/encoding.ts
```diff
@@ -22,13 +22,15 @@ export interface Encoding {
   y?: PositionFieldDef | ValueDef<number>;
 
   /**
-   * X2 coordinates for ranged `bar`, `rule`, `area`
+   * X2 coordinates for ranged `bar`, `rule`, `area`.
    */
+  // TODO: Ham need to add default behavior
```
domoritz: TODO

src/fielddef.ts
```diff
@@ -35,6 +35,9 @@ export interface ConditionalValueDef<T> extends ValueDef<T> {
 export interface FieldDef {
   /**
    * Name of the field from which to pull a data value.
+   *
+   *  __Default value:__ `false`
```
kanitw: false?????

Since this is required, there is no default value. 

Add __Required.__ before "Name of ..." and remove default value.

PR #2195, opened 2017-04-17
[WIP] Remove layout data

kanitw:
This PR get rid of LayoutData and use signal to compute layout of unit spec and faceted spec instead.

Note that this does not work with independent scale *under facet* yet.  (Independent scale under facet is more complicated than others as the scale all share the same name and is with a special scope.)  With the mix of shared and independent scale under facet, we probably have to do some parsing and merging logic. 

However, this is a good intermediate step to move forward, so let's merge this first. 

----

- [ ] unit
- [ ] facet of unit
- [ ] layer of unit
- [ ] facet of layer of unit

## Follow up Later

- [ ] Independent scale variation

### DIFF
--- a/src/compile/data/facet.ts
+++ b/src/compile/data/facet.ts
@@ -62,6 +62,17 @@ export class FacetNode extends DataFlowNode {
           groupby: [this.columnField]
         }]
       });
+
+      // Column needs another data source to calculate cardinality as input to layout
+      data.push({
+        name: this.columnName + '-layout',
+        source: this.columnName,
+        transform: [{
+          type: 'aggregate',
+          ops: ['distinct'],
+          fields: [this.columnField]
+        }]
+      });
     }
 
     if (this.rowName) {
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -1,7 +1,7 @@
 import {Axis} from '../axis';
 import {Channel, COLUMN, ROW, X, Y} from '../channel';
 import {Config} from '../config';
-import {LAYOUT, MAIN} from '../data';
+import {MAIN} from '../data';
 import {reduce} from '../encoding';
 import {Facet} from '../facet';
 import {FieldDef, normalize, title as fieldDefTitle} from '../fielddef';
@@ -11,7 +11,7 @@ import {Scale} from '../scale';
 import {FacetSpec} from '../spec';
 import {StackProperties} from '../stack';
 import {contains, Dict, extend, flatten, keys, vals} from '../util';
-import {FontWeight} from '../vega.schema';
+import {FontWeight, VgSignal} from '../vega.schema';
 import {
   isDataRefDomain,
   isDataRefUnionedDomain,
@@ -26,7 +26,7 @@ import {gridShow} from './axis/rules';
 import {buildModel} from './common';
 import {assembleData, assembleFacetData, FACET_SCALE_PREFIX} from './data/assemble';
 import {parseData} from './data/parse';
-import {assembleLayoutData, parseFacetLayout} from './layout';
+import {parseFacetLayout} from './layout';
 import {getTextHeader} from './layout/header';
 import {Model, ModelWithField} from './model';
 import initScale from './scale/init';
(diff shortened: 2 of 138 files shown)


domoritz:
Repeat always uses independent scale, though. So I'd need support for that.

domoritz:
- [ ] Change the order in the JSON. Output signals before data to reduce the number of changes.

src/compile/data/facet.ts
```diff
@@ -62,6 +62,17 @@ export class FacetNode extends DataFlowNode {
           groupby: [this.columnField]
         }]
       });
+
+      // Column needs another data source to calculate cardinality as input to layout
+      data.push({
+        name: this.columnName + '-layout',
```
domoritz: we use _ for dataset names in other places. Let's keep things consistent.

src/compile/facet.ts
```diff
@@ -229,10 +221,9 @@ export class FacetModel extends ModelWithField {
     };
   }
 
-  public assembleLayoutData(layoutData: VgData[]): VgData[] {
-    // Postfix traversal – layout is assembled bottom-up
-    this.child.assembleLayoutData(layoutData);
-    return assembleLayoutData(this, layoutData);
+  public assembleLayoutSignals(): VgSignal[] {
+    // FIXME correct this for column
```
domoritz: Will you do this or is there an issue for follow up?

src/compile/facet.ts
```diff
@@ -242,6 +233,13 @@ export class FacetModel extends ModelWithField {
     );
   }
 
+  private columnDistinctSignal() {
+    // In facetNode.assemble(), the name is always this.getName('column') + '-layout'.
```
domoritz: I wonder whether we should use output nodes instead so that we can merge if two subtrees are equivalent. I guess it's not relevant for now since you didn't add an aggregate node but maybe for the future.

src/compile/facet.ts
```diff
@@ -242,6 +233,13 @@ export class FacetModel extends ModelWithField {
...
+    const facetLayoutDataName = this.getName('column') + '-layout';
```
domoritz: Will need to fix this if you rename the data source. I'm also happy to use - everywhere instead.

src/compile/facet.ts
```diff
@@ -421,6 +410,7 @@ export function getTitleGroup(model: FacetModel, channel: 'row' | 'column') {
     textRef: {value: fieldDefTitle(fieldDef, model.config)},
 
     // TODO: customize alignment
+    // FIXME: this is not working.  Need to wait for Jeff's layout update that include title role
```
domoritz: Follow up, right?

src/compile/layer.ts
```diff
@@ -169,8 +170,14 @@ export class LayerModel extends Model {
   }
 
   // TODO: Support same named selections across children.
-  public assembleSignals(signals: any[]): any[] {
-    return this.children.reduce((sg, child) => child.assembleSignals(sg), []);
+  public assembleSignals(): VgSignal[] {
+    return this.children.reduce((signals, child) => {
+      return [].concat(
+        child.assembleLayoutSignals(),
+        child.assembleSignals(),
```
domoritz: maybe later rename this to asssembleSelectionSignals

src/compile/layout/index.ts
```diff
@@ -87,12 +51,56 @@ function parseUnitSizeLayout(model: UnitModel, channel: Channel): SizeComponent
...
+
+      return `bandspace(${cardinality}, ${paddingInner}, ${paddingOuter}) * ${scale.rangeStep}`;
+    }
+  }
+  return `${model[sizeType]}`;
+}
+
+
+
+export function oldUnitSizeExpr(model: UnitModel, channel: Channel): string {
```
domoritz: why do we still need this if it's old?

test/compile/compile.test.ts
```diff
@@ -43,7 +43,7 @@ describe('Compile', function() {
         }
       ]);
 
-      assert.equal(spec.data.length, 2); // just source and layout
+      assert.equal(spec.data.length, 1); // just source
```
domoritz: these tests have become meaningless but whatever.

src/compile/facet.ts
kanitw: Yes.

src/compile/layout/index.ts
kanitw: Because I plan to remove a lot of other things (now) but want you to get a reasonable codebase to rebase to first.

src/compile/facet.ts
kanitw: I guess you can merge two subtree that have the same facet in the future.

PR #2208, opened 2017-04-18
Only add format parse when necessary. I also think we don’t need the …

domoritz:
…`parse: {}` because Vega seems to not parse anything unless we specify it.


examples/vg-specs/layer_overlay.vg.json
```diff
@@ -56,14 +62,20 @@
...
+            "source": "data_2",
+            "transform": [
                 {
                     "type": "aggregate",
                     "groupby": [
                         "Cylinders"
                     ],
                     "ops": [
-                        "max"
+                        "min"
```
kanitw: @domoritz  do you know why max and min got swap in all these specs?

domoritz: It's a different source. It's all good.

PR #2210, opened 2017-04-18
Union non-spatial scales in repeat.

domoritz:
screen shot 2017-04-18 at 09 34 54

### DIFF
--- a/src/compile/repeat.ts
+++ b/src/compile/repeat.ts
@@ -1,4 +1,5 @@
 import {isArray} from 'vega-util';
+import {NONSPATIAL_SCALE_CHANNELS} from '../channel';
 import {CellConfig, Config} from '../config';
 import {Encoding} from '../encoding';
 import {Facet} from '../facet';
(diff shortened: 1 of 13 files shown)


src/compile/repeat.ts
```diff
@@ -144,8 +177,21 @@ export class RepeatModel extends Model {
   }
 
   public parseLegend() {
+    const legendComponent = this.component.legends = {};
+
     for (const child of this.children) {
       child.parseLegend();
+
+      // TODO: correctly implement independent axes
```
kanitw: independent legend*

src/compile/repeat.ts
```diff
@@ -144,8 +177,21 @@ export class RepeatModel extends Model {
...
+      if (true) { // if shared/union scale
+        keys(child.component.legends).forEach(function(channel) {
```
kanitw: // TODO: need to decide if the legend are not compatible. Do we still merge?

src/compile/repeat.ts
```diff
@@ -120,10 +122,41 @@ export class RepeatModel extends Model {
...
 
-    this.component.scales = {};
+    const scaleComponent: Dict<VgScale> = this.component.scales = {};
 
     this.children.forEach(function(child) {
       child.parseScale();
+
+      // FIXME(#1602): correctly implement independent scale
+      // Also need to check whether the scales are actually compatible, e.g. use the same sort or throw error
+      if (true) { // if shared/union scale
```
kanitw: // TODO: need to decide if the scales are not compatible. Do we still merge?

src/compile/repeat.ts
domoritz: Done

PR #2211, opened 2017-04-18
[WIP] Reimplement Facet to use the latest layout and headers

kanitw:


src/compile/facet.ts
```diff
@@ -153,23 +154,70 @@ export class FacetModel extends ModelWithField {
     }];
   }
 
-  public parseAxis() {
-    this.child.parseAxis();
-    // FIXME: Correct write parseAxis logic
-    this.component.axes = {};
+  // TODO: rename to parseAxisAndHeaders
```
domoritz: done

src/compile/layout/header.ts
```diff
@@ -1,65 +1,138 @@
...
+  if (header) {
+    let title = null;
+    if (layoutHeader.field && header.labels) {
+      title = {
+        text: {signal: `parent['${layoutHeader.field}']`},
+        offset: 10,
+        orient: channel === 'row' ? 'left' : 'top',
+        encode: {
+          update: {
+            fontWeight: {value: 'normal'},
```
domoritz: so this is the hack we talked about today with jeff?

domoritz:
LGTM

src/compile/layout/header.ts
kanitw: Yes.

PR #2212, opened 2017-04-18
Implement concat operator.

domoritz:
I implemented only vconcat for now because I think we have to have a conversation about what general layouts we want to support.  

screen shot 2017-04-18 at 11 35 07

### DIFF
--- a/src/compile/concat.ts
+++ b/src/compile/concat.ts
@@ -1,0 +1,133 @@
+import {CellConfig, Config} from '../config';
+import {Repeat} from '../repeat';
+import {ConcatSpec, RepeatSpec} from '../spec';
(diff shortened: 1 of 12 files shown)


src/compile/concat.ts
```diff
@@ -0,0 +1,133 @@
...
+    // combine with scales from children
+    return this.children.reduce((scales, c) => {
+      return scales.concat(c.assembleScales());
+    }, super.assembleScales());
+  }
+
+  public assembleLayout(): VgLayout {
+    // TODO: allow customization
+    return {
+      padding: {row: 10, column: 10, header: 10},
```
kanitw: padding: {row: 10, column: 10},
offset: 10,

PR #2238, opened 2017-04-20
Temporarily bring back the main group and Revert incorrect selection predicate fix

kanitw:
@arvind 

This makes `query_widgets` works  and allow us to revert  a2e88d2

But `faceted_selection` still seems funky.  I think the problem for  `faceted_selection` seems not relate to main-group, so I'll leave it up to you to figure out since you better understand how it works.  

(Please merge https://github.com/vega/vega-lite/pull/2237 first.)

### DIFF
--- a/examples/vg-specs/area.vg.json
+++ b/examples/vg-specs/area.vg.json
@@ -67,33 +67,49 @@
     ],
     "marks": [
         {
-            "name": "marks",
-            "type": "area",
-            "from": {
-                "data": "source_0"
-            },
+            "name": "nested-main-group",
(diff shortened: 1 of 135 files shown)


domoritz:
> allow us to revert a559552.

What do you mean? Why does it allow it?

kanitw:
> allow us to revert a559552.

a559552 was an incorrect fix for the problem that predicate selection refers to non-existent parent property for unit-plot after we remove the main-group. 

Now that we temporarily bring the `main-group` back, we can revert that incorrect fix.

Issue #2232, opened 2017-04-19
Add document for stack property in encoding.md


kanitw:
- [ ] add to the table

- [ ] add a few examples to demo

YuhanLu:
Do you mean `stack` in `PositionFieldDef`? Should I put the example right under the `PositionFieldDef` table?

kanitw:
> Do you mean stack in PositionFieldDef?

Yes

> Should I put the example right under the PositionFieldDef table?

IDK.  It depends.  

Another related task would be read mark.md and see if it still correctly describes stack.

YuhanLu:
I think this one is merged

PR #2241, opened 2017-04-20
add stack to encoding.md and make changes to new named interface

### LINKED ISSUE #2240: Update table with new interface name
After a lot of changes, some names of interfaces changed and those tables need to be updated

YuhanLu:
Fix #2240 and #2232


kanitw:
@YuhanLu rebase this?  (This seems conflicting with some changes you made earlier today)

You might wanna extract the rename script to be its own file with multiple lines and comments so that we can easily see the scripts diff

PR #2249, opened 2017-04-21
Modify sort that if aggregation is count, field will not be required.

### LINKED ISSUE #2095: Make `sort` no longer require `field: "*"` if `op` is `"count"`
Since a `fieldDef` no longer requires `field` if `aggregate` is `"count"`, `sort` should have a consistent behavior.

YuhanLu:
Fix #2095
screen shot 2017-04-20 at 6 29 26 pm

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -4903,7 +4903,6 @@
                 }
             },
             "required": [
-                "field",
                 "op"
             ],
             "type": "object"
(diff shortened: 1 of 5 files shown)


kanitw:
Thanks!   Can you 

- [ ] Add `?` to `field` in `SortField`
- [ ] Add your example spec the `examples/specs` (`bar_sort_by_count`)
- [ ] Make the bar really sorted by count.  (The screenshot doesn't show the bar correctly sorted.)

YuhanLu:
I tried to dig into this vega output, i still couldn't find out why it's not sorted by count...

kanitw:
If its a bug, file a bug in Vega

PR #2250, opened 2017-04-21
Removed unnecessary imports from bin.test.ts

NeelMohapatra:
I thought tslint takes care of this


kanitw:
Thanks!

domoritz:
Thanks but don't spent time on this for other files. Once the linter works again it will do it for us.

kanitw:
There was a bug in tslint so we had to hide this flag.

NeelMohapatra:
Ah. I only checked the imports in this file because I wrote it myself.

PR #2255, opened 2017-04-22
Selections cleanup

arvind:


src/compile/selection/selection.ts
```diff
@@ -239,17 +239,6 @@ export function channelSignalName(selCmpt: SelectionComponent, channel: Channel)
...
-        width: model.getSizeSignalRef('width'),
-        height: model.getSizeSignalRef('height'),
-        fill: {value: 'transparent'},
-        clip: {value: true}
-      }
-    },
-    marks: marks.map(model.correctDataNames)
-  }];
+function clipMarks(marks: any[]): any[] {
+  return marks.map((m) => (m.clip = true, m));
```
kanitw: Oh interesting that we don't need to clip anymore. 
(I wonder why?) 

Also, do we I need to add `fill: {value: 'transparent'}` to the `nested-main-group`

arvind: We still need clipping, but with beta 29, we can do it directly on marks rather than nesting within a group. Good catch on the transparent fill. Force pushed an update.

PR #2274, opened 2017-04-23
Add selection filter support.

arvind:
This commit also removes the Filter[] support.

crossfiltering


domoritz:
This PR also removes support for arrays of filters. You just have to create multiple filters.

PR #2291, opened 2017-04-26
Update docs for Marks

kanitw:
@domoritz  -- I wonder if this branch breaks all the examples in marks.html for you or it's just weird in my computer.


domoritz:
It breaks travis at least.

PR #2297, opened 2017-04-27
Dom/skip

domoritz:


kanitw:
Why is ^ better?

domoritz:
> Why is ^ better?

because it matches the latest major version

PR #2304, opened 2017-04-28
Make layer adopts parent's width/height.

### LINKED ISSUE #2300: `width` doesn't work with layer of layer
```json
{
  "width": 600,
  "data": {
    "values": [
      {"year": 1875,"population": 1309},
      ....
      {"year": 2014,"population": 41777}
    ]
  },
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "x": {"field": "year", "type": "temporal"},
        "y": {"field": "population", "type": "quantitative"}
      }
    }
  ]
}
```
produces 
screen shot 2017-04-27 at 4 54 12 pm
but add `overlay` and
```json
{
  "width": 600,
  "data": {
    "values": [
      {"year": 1875,"population": 1309},
      ....
      {"year": 2014,"population": 41777}
    ]
  },
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "x": {"field": "year", "type": "temporal"},
        "y": {"field": "population", "type": "quantitative"}
      }
    }
  ],
  "config": {"overlay": {"line": true}}
}
```
produces
screen shot 2017-04-27 at 4 55 49 pm

kanitw:
Fix #2300

vega_editor_

### DIFF
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -4,7 +4,7 @@ import {Channel, TEXT} from '../channel';
 import {CellConfig, Config} from '../config';
 import {field, FieldDef, OrderFieldDef} from '../fielddef';
 import {Mark, MarkConfig, TextConfig} from '../mark';
-import {isConcatSpec, isFacetSpec, isLayerSpec, isRepeatSpec, isUnitSpec, Spec} from '../spec';
+import {isConcatSpec, isFacetSpec, isLayerSpec, isRepeatSpec, isUnitSpec, Spec, UnitSize} from '../spec';
 import {TimeUnit} from '../timeunit';
 import {formatExpression} from '../timeunit';
 import {QUANTITATIVE} from '../type';
@@ -18,17 +18,18 @@ import {RepeaterValue, RepeatModel} from './repeat';
 import {UnitModel} from './unit';
 
 
-export function buildModel(spec: Spec, parent: Model, parentGivenName: string, repeater: RepeaterValue, config: Config): Model {
+export function buildModel(spec: Spec, parent: Model, parentGivenName: string,
+  unitSize: UnitSize, repeater: RepeaterValue, config: Config): Model {
   if (isFacetSpec(spec)) {
     return new FacetModel(spec, parent, parentGivenName, repeater, config);
   }
 
   if (isLayerSpec(spec)) {
-    return new LayerModel(spec, parent, parentGivenName, repeater, config);
+    return new LayerModel(spec, parent, parentGivenName, unitSize, repeater, config);
   }
 
   if (isUnitSpec(spec)) {
-    return new UnitModel(spec, parent, parentGivenName, repeater, config);
+    return new UnitModel(spec, parent, parentGivenName, unitSize, repeater, config);
   }
 
   if (isRepeatSpec(spec)) {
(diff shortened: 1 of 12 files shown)


src/compile/layer.ts
```diff
@@ -42,8 +42,13 @@ export class LayerModel extends Model {
 
     super(spec, parent, parentGivenName, config);
 
-    this.width = spec.width;
-    this.height = spec.height;
+    this.width = spec.width !== undefined ? spec.width :
```
domoritz: I don't think this is sufficient. What about layer of layer of layer? I think you need to walk up the tree until you find a size definition similar to what @arvind did for selections. Can we use the same logic?

domoritz:
Proposed algorithm in https://github.com/vega/vega-lite/issues/2308

kanitw:
@domoritz  I'm not gonna implement the algorithm in #2308 now -- so you might wanna merge this and make it less buggy and fix #2308 later.

kanitw:
Good catch though!

src/compile/layer.ts
kanitw: @domoritz Can you be more specific about which part of selections?

kanitw:
Wait wait -- passing things down seems like a better idea than lookup

domoritz:
Check out https://github.com/vega/vega-lite/blob/master/src/compile/model.ts#L359 `getComponent`. I don't like this pr if we implement another way for the same problem. We should use the same logic for the selections/scale lookup and the size lookup. 

Whether you want to pass things down (as we already do for config, repeat, and name maps) to look up, doesn't matter to me.

domoritz:
> @domoritz I'm not gonna implement the algorithm in #2308 now -- so you might wanna merge this and make it less buggy and fix #2308 later.

Fair enough. But I think using the same infrastructure for scales, selections, and size lookup makes sense. Probably want to add axes as well.

kanitw:
I disagree.  For selection, you look up components, not primitive structure of the spec.  

Trying to look up for primitive structure in the spec is an overkill.

domoritz:
> Trying to look up for primitive structure in the spec is an overkill.

Can you explain why.

kanitw:
During initialization, we only cascade things down. 

For selection or scale that are merged , we have to look up to the parents because items got merged and move up to ancestors. Those calls happens later in the pipeliine (likely during assemble or the end of parse.)

src/compile/common.ts
```diff
@@ -18,17 +18,18 @@ import {RepeaterValue, RepeatModel} from './repeat';
 import {UnitModel} from './unit';
 
 
-export function buildModel(spec: Spec, parent: Model, parentGivenName: string, repeater: RepeaterValue, config: Config): Model {
+export function buildModel(spec: Spec, parent: Model, parentGivenName: string,
```
domoritz: I wonder whether we should just put the size in the config. No need to add another parameter everywhere.

kanitw: Nah.   People are clamoring for width/height to be not hidden in config.

domoritz: Well, for users this would be invisible. it's just less code and in the unit you can directly read from the config.

kanitw: We can't do that.    For ordinal, if you explicitly set `"width"`, then you get flexible band scale.
If you don't set `"width"`, config.cell.width will be ignored. 

If you want to change this rule, propose better idea in another issue.  That's beyond the scope of this PR.

domoritz: Oh, I see. I forgot about the case where `width` overrides `rangeStep`

PR #2307, opened 2017-04-28
Correct types for layer's `children`

kanitw:
`LayerModel`'s children can be `LayerModel` as well (not just `UnitModel`).

By correcting the type, I notice that `assembleLayerSelectionMarks` previously didn't handle when LayerModel's child is not a UnitModel.  

@arvind could you please verify if my change below is right. If not, please correct it.


arvind:
If a `LayerModel`'s children can only be `UnitModel` or another `LayerModel` then the change in 90a1258 LGTM. Feel free to rebase + merge once the Travis build passes (concerning error about exceeding the max call stack).

PR #2318, opened 2017-04-29
Allow concat of repeat

domoritz:
Also fix size signals.

### DIFF
--- a/examples/vg-specs/bar_grouped.vg.json
+++ b/examples/vg-specs/bar_grouped.vg.json
@@ -264,20 +264,11 @@
                     },
                     "encode": {
                         "update": {
-                            "width": {
-                                "signal": "child_width"
-                            },
-                            "height": {
-                                "signal": "child_height"
-                            },
                             "stroke": {
                                 "value": "#ccc"
                             },
                             "strokeWidth": {
                                 "value": 0
-                            },
-                            "fill": {
-                                "value": "transparent"
                             }
                         }
                     },
--- a/examples/vg-specs/bar_grouped.vg.json
+++ b/examples/vg-specs/bar_grouped.vg.json
@@ -264,6 +264,12 @@
                     },
                     "encode": {
                         "update": {
+                            "width": {
+                                "signal": "child_width"
+                            },
+                            "height": {
+                                "signal": "child_height"
+                            },
                             "stroke": {
                                 "value": "#ccc"
                             },
--- a/examples/vg-specs/bar_grouped.vg.json
+++ b/examples/vg-specs/bar_grouped.vg.json
@@ -275,6 +275,9 @@
                             },
                             "strokeWidth": {
                                 "value": 0
+                            },
+                            "fill": {
+                                "value": "transparent"
                             }
                         }
                     },
--- a/src/compile/compile.ts
+++ b/src/compile/compile.ts
@@ -105,7 +105,7 @@ export function assembleNestedMainGroup(model: Model) {
       type: 'group',
       layout,
       signals,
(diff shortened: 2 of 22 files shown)


domoritz:
Tests are passing.

src/compile/compile.ts
```diff
@@ -96,15 +96,7 @@ export function assembleNestedMainGroup(model: Model) {
...
-  const hasLayout = !!model.assembleLayout();
-  const parentEncodeEntry = {
-    ...(!hasLayout ? {
-      width: {signal: 'width'},
-      height: {signal: 'height'},
-    } : {}),
-    ...model.assembleParentGroupProperties()
-  };
-
+  const parentEncodeEntry = model.assembleParentGroupProperties();
```
kanitw: I think this is can be incorrect, if the root model has `name`, you won't correctly output `width` and `height` signals for this nested main group.

domoritz: Hmm, can you give an example?

domoritz: Okay, looks like it's already incorrect on master.

domoritz: https://github.com/vega/vega-lite/issues/2326

examples/vg-specs/bar_grouped.vg.json
```diff
@@ -264,20 +264,11 @@
...
-                            "height": {
-                                "signal": "child_height"
-                            },
                             "stroke": {
                                 "value": "#ccc"
                             },
                             "strokeWidth": {
                                 "value": 0
-                            },
-                            "fill": {
```
kanitw: @domoritz Why removing fill in facet group?  I guess this might break brushing.

domoritz: Any idea why this is happening? I don't see where I made a change that could cause this.

domoritz: I tried it and it works.

kanitw: What works?

src/compile/facet.ts
```diff
@@ -332,6 +332,9 @@ function getFacetGroupProperties(model: FacetModel) {
   const encodeEntry = model.assembleParentGroupProperties();
 
   return {
+    width: model.child.getSizeSignalRef('width'),
```
kanitw: I wonder if this is the right fix.  Shouldn't `model.assembleParentGroupProperties` already produce what we need?  

I wonder if there is something actually wrong with `hasSubPlotWithXy`? 
(As commented, I'm not sure if we really need it anyway.)

examples/vg-specs/bar_grouped.vg.json
domoritz: All interactive specs still work.

src/compile/facet.ts
domoritz: I wonder the same but I wanted to fix other things first.

examples/vg-specs/bar_grouped.vg.json
arvind: My vote would be to keep the transparent fills, unless there's a compelling reason to get rid of them. Having an undefined fill is a frequent source of bugs with capturing interaction events.

domoritz: Okay, I wish I understood which line changed that... looking into it.

domoritz: Fixed.

arvind:
Thanks!
-- 
-Arvind
-----
Arvind Satyanarayan <http://arvindsatya.com>
Sent from my iPhone

Issue #2251, opened 2017-04-21
Correctly merge scale


kanitw:
(Currently we just always use the first axis/legend)

- [x] Consider precedence of explicitly specified properties
- [x] Consider if there are conflicting properties

- [x] when merging band vs point scale, favor band scale as it produces better selection 

- [ ] For example, layer_histogram's x-scales somehow have not been merged.

- [x] `parseDomain` must be called separately after `parseScale` since `domain` depends on `type` and will persist after merging.  (Other properties that depends on type can get overridden.)   
- [x] `initSize` should become a part of parse and come after `parseScale` as size (`width`/`height`) depends on `rangeStep`
- [x] `initMarkDef` should not depend on scale (as it can have incorrect scale type) -- maybe `initMarkDef` should simply become a part of `parseMark`
  - orient depends on scale.type (whether the type is discrete or not)   
  - parsePathOrder depends on orient, but is a part of `parseData`, which happens before `parseMark`

- [x] get rid of initResolve()

- [x] rename
  - [x] layout/index => layout/assemble
  - [x] scale.rules => properties

- [x] Don't read `width` / `height` directly

domoritz:
- [ ] Be smarter about merging the grid for dual axis charts.

domoritz:
We could be smart about things and use a single scale and axis if the layers use the same field and transformations.

kanitw:
> We could be smart about things and use a single scale and axis if the layers use the same field and transformations.

Isn't our default is union scale already?

domoritz:
> Isn't our default is union scale already?
Not for concat.

kanitw:
Yeah but your comment above discusses layer

>  if the __layers__ use the same field and transformations.

kanitw:
Looks like a lot of things depends on scale init, but to make sure that we treat scale merging correctly, we will have to remove a lot of these dependencies.  (I'm still investigating -- will update more.)

kanitw:
- [ ] when merging band vs point scale, favor band scale as it produces better selection

kanitw:
- [ ] `parseDomain` must be called separately after `parseScale` since `domain` depends on `type`.

kanitw:
- [ ] initMarkDef should not depend on scale (as it can have incorrect scale type)

kanitw:
- [ ] Consolidate initScale with initSize (previously https://github.com/vega/vega-lite/issues/1802)

kanitw:
[...]

PR #2327, opened 2017-04-30
Resolve

domoritz:
Depends on https://github.com/vega/vega-lite/pull/2318

- Add resolve types
- Add resolve to layer
- Implement dual axis chart support

screen shot 2017-04-29 at 18 26 14

Example spec

```json
{
  "width": 460,
  "height": 240,
  "layer": [
    {
      "data": {"url": "data/co2.csv"},
      "mark": "line",
      "encoding": {
        "x": {
          "field": "Year",
          "type": "temporal",
          "timeUnit": "year"
        },
        "y": {
          "field": "Mean",
          "type": "quantitative",
          "scale": {"zero": false},
          "axis": {"title": "CO2 in PPM"}
        }
      }
    },
    {
      "layer": [
        {
          "data": {"values": [{"rule": 0}]},
          "mark": "rule",
          "encoding": {
            "y": {
              "field": "rule",
              "type": "quantitative",
              "axis": {"title": "Mean Global Temperature"}
            },
            "opacity": {"value": 0.4},
            "color": {"value": "black"}
          }
        },
        {
          "data": {"url": "data/temp.csv"},
          "mark": "line",
          "encoding": {
            "x": {
              "field": "Year",
              "type": "temporal",
              "timeUnit": "year"
            },
            "y": {"field": "Value","type": "quantitative"},
            "color": {"value": "firebrick"}
          }
        }
      ]
    }
  ],
  "resolve": {"y": {"scale": "independent"}}
}
```

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -214,6 +214,16 @@
                     "description": "The default visualization padding, in pixels, from the edge of the visualization canvas to the data rectangle. This can be a single number or an object with `\"top\"`, `\"left\"`, `\"right\"`, `\"bottom\"` properties.\n\n__Default value__: `5`",
                     "minimum": 0
                 },
+                "resolve": {
+                    "anyOf": [
+                        {
+                            "$ref": "#/definitions/{ color?: NonspatialResolve; opacity?: NonspatialResolve; size?: NonspatialResolve; shape?: Nonsp..."
+                        },
+                        {
+                            "$ref": "#/definitions/{ x?: SpatialResolve; y?: SpatialResolve; }"
+                        }
+                    ]
+                },
                 "transform": {
                     "description": "An array of data transformations such as filter and new field calculation.",
                     "items": {
@@ -2692,6 +2702,16 @@
                     "description": "Name of the visualization for later reference.",
                     "type": "string"
                 },
+                "resolve": {
+                    "anyOf": [
+                        {
+                            "$ref": "#/definitions/{ color?: NonspatialResolve; opacity?: NonspatialResolve; size?: NonspatialResolve; shape?: Nonsp..."
+                        },
+                        {
+                            "$ref": "#/definitions/{ x?: SpatialResolve; y?: SpatialResolve; }"
+                        }
+                    ]
+                },
                 "transform": {
                     "description": "An array of data transformations such as filter and new field calculation.",
                     "items": {
@@ -5010,6 +5030,16 @@
                 },
                 "title": {
                 }
+            },
+            "type": "object"
+        },
+        "{ color?: NonspatialResolve; opacity?: NonspatialResolve; size?: NonspatialResolve; shape?: Nonsp...": {
+            "properties": {
+            },
+            "type": "object"
+        },
+        "{ x?: SpatialResolve; y?: SpatialResolve; }": {
+            "properties": {
             },
             "type": "object"
         }
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -215,14 +215,28 @@
                     "minimum": 0
                 },
                 "resolve": {
-                    "anyOf": [
-                        {
-                            "$ref": "#/definitions/{ color?: NonspatialResolve; opacity?: NonspatialResolve; size?: NonspatialResolve; shape?: Nonsp..."
+                    "additionalProperties": false,
+                    "properties": {
+                        "color": {
+                            "$ref": "#/definitions/NonspatialResolve"
                         },
-                        {
-                            "$ref": "#/definitions/{ x?: SpatialResolve; y?: SpatialResolve; }"
+                        "opacity": {
+                            "$ref": "#/definitions/NonspatialResolve"
+                        },
+                        "shape": {
+                            "$ref": "#/definitions/NonspatialResolve"
+                        },
+                        "size": {
+                            "$ref": "#/definitions/NonspatialResolve"
+                        },
+                        "x": {
+                            "$ref": "#/definitions/SpatialResolve"
+                        },
+                        "y": {
+                            "$ref": "#/definitions/SpatialResolve"
                         }
-                    ]
+                    },
+                    "type": "object"
                 },
                 "transform": {
                     "description": "An array of data transformations such as filter and new field calculation.",
@@ -2703,14 +2717,28 @@
                     "type": "string"
                 },
                 "resolve": {
-                    "anyOf": [
-                        {
-                            "$ref": "#/definitions/{ color?: NonspatialResolve; opacity?: NonspatialResolve; size?: NonspatialResolve; shape?: Nonsp..."
+                    "additionalProperties": false,
+                    "properties": {
+                        "color": {
+                            "$ref": "#/definitions/NonspatialResolve"
                         },
-                        {
-                            "$ref": "#/definitions/{ x?: SpatialResolve; y?: SpatialResolve; }"
+                        "opacity": {
+                            "$ref": "#/definitions/NonspatialResolve"
+                        },
+                        "shape": {
+                            "$ref": "#/definitions/NonspatialResolve"
+                        },
+                        "size": {
+                            "$ref": "#/definitions/NonspatialResolve"
+                        },
+                        "x": {
+                            "$ref": "#/definitions/SpatialResolve"
+                        },
+                        "y": {
+                            "$ref": "#/definitions/SpatialResolve"
                         }
-                    ]
+                    },
+                    "type": "object"
                 },
                 "transform": {
                     "description": "An array of data transformations such as filter and new field calculation.",
@@ -3572,6 +3600,18 @@
             ],
             "type": "object"
         },
+        "NonspatialResolve": {
+            "additionalProperties": false,
+            "properties": {
+                "legend": {
+                    "$ref": "#/definitions/ResolveMode"
+                },
+                "scale": {
+                    "$ref": "#/definitions/ResolveMode"
+                }
+            },
+            "type": "object"
+        },
         "OneOfFilter": {
             "additionalProperties": false,
             "properties": {
@@ -3884,6 +3924,13 @@
                 }
             },
             "type": "object"
+        },
+        "ResolveMode": {
+            "enum": [
+                "independent",
+                "shared"
+            ],
+            "type": "string"
         },
         "Scale": {
             "additionalProperties": false,
@@ -4535,6 +4582,18 @@
             ],
             "type": "string"
         },
+        "SpatialResolve": {
+            "additionalProperties": false,
+            "properties": {
+                "axis": {
+                    "$ref": "#/definitions/ResolveMode"
+                },
+                "scale": {
+                    "$ref": "#/definitions/ResolveMode"
+                }
+            },
+            "type": "object"
+        },
         "StackOffset": {
             "enum": [
                 "center",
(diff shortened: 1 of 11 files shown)


src/compile/layer.ts
```diff
@@ -108,27 +107,37 @@ export class LayerModel extends Model {
...
           // Just use the first axes definition for each channel
           // TODO: what if the axes from different children are not compatible
           if (!axisComponent[channel]) {
             axisComponent[channel] = child.component.axes[channel];
-            delete child.component.axes[channel];
           }
         } else {
-          // Otherwise do nothing for independent axes
+          // If axes are independent
+          if (!axisComponent[channel]) {
```
kanitw: Add `// TODO(https://github.com/vega/vega-lite/issues/2251): correctly merge axis`

src/compile/layer.ts
```diff
@@ -108,27 +107,37 @@ export class LayerModel extends Model {
...
+            // copy the first axis
+            axisComponent[channel] = child.component.axes[channel];
+          } else if (axisComponent[channel].axes.length === 1) {
+            // and put the second one on the right/top
+            axisComponent[channel].axes.push({
+              ...child.component.axes[channel].axes[0],
+              orient: channel === 'y' ? 'right' : 'top'
+            });
+            axisComponent[channel].gridAxes.push({
```
kanitw: (Future) Do we want to automatically deal with grid for dual axis?  
For dual axes with independent grid, it seems like the best idea is to hide grid for both sides.  However, if users explicitly ask for it, they should still get them.  (Maybe we can add additional TODO to #2251).

src/compile/layer.ts
```diff
@@ -108,27 +107,37 @@ export class LayerModel extends Model {
...
+              ...child.component.axes[channel].gridAxes[0]
+            });
+          }
+          // Ignore all other axes.
```
kanitw: Why ignore all other axes?  This seems wrong.  At least showing them with occlusion and let users set offset manually is better than not showing. 

A better logic is probably alternating between left/right (or top/bottom) add incrementally add offset so the axes do not overlap.  However, I don't care that much about multiple axes anyway. Simply showing them as overlapping seems sufficient.

src/resolve.ts
```diff
@@ -0,0 +1,47 @@
+import {NonspatialScaleChannel, SCALE_CHANNELS, SPATIAL_SCALE_CHANNELS, SpatialScaleChannel, UNIT_SCALE_CHANNELS} from './channel';
+import * as log from './log';
+import {contains} from './util';
+
+export type ResolveMode = 'independent' | 'shared';
+
+export interface SpatialResolve {
+  scale: ResolveMode;
```
kanitw: Why is scale required?    

Suppose a scale is shared and I want to override its axes to be independent, I don't need to specify scale resolve.

examples/specs/layer_bar_dual_axis.vl.json
```diff
@@ -0,0 +1,41 @@
...
+          "timeUnit": "month",
+          "field": "date",
+          "type": "temporal"
+        },
+        "y": {
+          "aggregate": "mean",
+          "field": "precipitation",
+          "type": "quantitative",
+          "axis": {
+            "grid": false
```
kanitw: Why hiding grid for just one of the layers, not both?

src/compile/layer.ts
```diff
@@ -108,27 +107,37 @@ export class LayerModel extends Model {
...
-          // If shared/union axis (only possible if the scale is shared in the first place)
+      keys(child.component.axes).forEach((channel: SpatialScaleChannel) => {
+        if (this.resolve[channel].axis === 'shared') {
+          // If shared/union axis
 
           // Just use the first axes definition for each channel
           // TODO: what if the axes from different children are not compatible
           if (!axisComponent[channel]) {
             axisComponent[channel] = child.component.axes[channel];
-            delete child.component.axes[channel];
```
kanitw: Why no longer deleting from child?

src/resolve.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function isSpatialResolve(resolve: Resolve): resolve is SpatialResolve {
+  return 'axis' in resolve;
+}
+
+export function isNonspatialResolve(resolve: Resolve): resolve is NonspatialResolve {
+  return 'legend' in resolve;
+}
+
+export type ResolveMapping = {[P in NonspatialScaleChannel]?: NonspatialResolve} | {[P in SpatialScaleChannel]?: SpatialResolve};
```
kanitw: `&`, not `|`.

src/resolve.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export interface NonspatialResolve {
+  scale: ResolveMode;
+  legend?: ResolveMode;
+}
+
+export type Resolve = SpatialResolve | NonspatialResolve;
+
+export function isSpatialResolve(resolve: Resolve): resolve is SpatialResolve {
```
kanitw: `axis` is optional, so it might not really be in the `resolve` you provide.  Thus, this type guard is not always correct.  (I know it might work okay for the logic below, but still can lead to confusion.)

I believe you remove these two type guard and instead check whether the channel is `NonspatialScaleChannel` or `SpatialScaleChannel`.

examples/specs/layer_bar_dual_axis.vl.json
domoritz: I want some grid. Also, an interesting test case. We can refine this later.

src/resolve.ts
domoritz: Yeah, I do that in the code. I just need this to be able to access the axis or legend propertly.

src/resolve.ts
domoritz: Good catch. It actually isn't required.

src/compile/layer.ts
domoritz: Well, in your implementation we ignore all axes. I think if you have more than 2 axes, you should not expect us to do a good job. Hence this version for now.

src/compile/layer.ts
domoritz: Moved down. I have no idea why we didn't do this earlier.

src/compile/layer.ts
kanitw: This one is really easy -- just do it so we don't have to come back and fix it.

> At least showing them with occlusion and let users set offset manually is better than not showing.

Of course, don't waste time on this one

> A better logic is probably alternating between left/right (or top/bottom) add incrementally add offset so the axes do not overlap.

src/resolve.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function isNonspatialResolve(resolve: Resolve): resolve is NonspatialResolve {
+  return 'legend' in resolve;
+}
+
+export type ResolveMapping = {[P in NonspatialScaleChannel]?: NonspatialResolve} | {[P in SpatialScaleChannel]?: SpatialResolve};
+
+export function initLayerResolve(resolve: ResolveMapping): ResolveMapping {
+  UNIT_SCALE_CHANNELS.forEach(channel => {
```
kanitw: Initializing resolve for all channel no matter they exist or not is weird.  
I think it's better to do this kind of thing in parse when we know if children contains a particular scale or not.  

Also, it seems like this method can cause side effect to the resolve object in the input spec, which is not good.

src/resolve.ts
domoritz: Deleted

src/resolve.ts
domoritz: I changed the method so that it doesn't cause side effects. Having all resolves makes things much easier downstream but I agree that it is a bit wasteful.

kanitw: > Having all resolves makes things much easier downstream

I doubt if it is really easier.

src/compile/layer.ts
domoritz: Okay, I alternate between the two sides now.

build/vega-lite-schema.json
```diff
@@ -4990,6 +5010,16 @@
                 }
             },
             "type": "object"
+        },
+        "{ color?: NonspatialResolve; opacity?: NonspatialResolve; size?: NonspatialResolve; shape?: Nonsp...": {
```
kanitw: I think we have to add additional `sed` to properly rename this.

src/resolve.ts
kanitw: Thinking more about it I really don't like it this way, but we have a lot more other important fish to catch, we can fix this later.

domoritz: I agree that there are better ways. However, the nice thing about this method is that it does not need to see the model.

build/vega-lite-schema.json
domoritz: Yeah, I looked into fixing this properly with the ts -> json converter. I think we can do it but I need some more time. 

@YuhanLu can you add the missing sed calls?

src/compile/layer.ts
domoritz: Done

src/resolve.ts
domoritz: Done

src/compile/layer.ts
domoritz: Done

src/compile/layer.ts
```diff
@@ -108,27 +107,37 @@ export class LayerModel extends Model {
...
           }
         } else {
-          // Otherwise do nothing for independent axes
+          // If axes are independent
+          // TODO(#2251): correctly merge axis
+          if (!axisComponent[channel]) {
+            // copy the first axis
+            axisComponent[channel] = child.component.axes[channel];
+          } else {
+            // put every add numbered axis on the right/top
```
kanitw: add => odd?

domoritz: Fixed

kanitw:
Interestingly, I check out this branch and try to hide grid for both layers and nothing is rendered.

```
{
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "x": {
          "timeUnit": "month",
          "field": "date",
          "type": "temporal"
        },
        "y": {
          "aggregate": "mean",
          "field": "precipitation",
          "type": "quantitative",
          "axis": {
            "grid": false
          }
        }
      }
    },
    {
      "mark": "line",
      "encoding": {
        "x": {
          "timeUnit": "month",
          "field": "date",
          "type": "temporal"
        },
        "y": {
          "aggregate": "mean",
          "field": "temp_max",
          "type": "quantitative",
          "axis": {
            "grid": false
          }
        },
        "color": {"value": "firebrick"}
      }
    }
  ],
  "resolve": {"y": {"scale": "independent"}}
}
```

domoritz:
Good catch. Let me see.

domoritz:
The problem was that I always pushed the grid axis regardless of whether it existed or not. Fixed.

src/resolve.ts
```diff
@@ -0,0 +1,38 @@
...
+  scale?: ResolveMode;
+  axis?: ResolveMode;
+}
+
+export interface NonspatialResolve {
+  scale?: ResolveMode;
+  legend?: ResolveMode;
+}
+
+export type Resolve = SpatialResolve | NonspatialResolve;
```
kanitw: This is somehow still `|`, although you said you fixed it.

src/resolve.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export type ResolveMapping = {[P in NonspatialScaleChannel]?: NonspatialResolve} | {[P in SpatialScaleChannel]?: SpatialResolve};
```
kanitw: This is somehow still `|`, although you said you fixed it.

examples/specs/layer_bar_dual_axis.vl.json
kanitw: I'm curious why you want grid for dual axis though. (It will be a grid for only one side.)

src/resolve.ts
kanitw: (After changing to `&`, it would be great to check if we still need to do `sed` -- maybe not anymore.)

domoritz: Well, I changed it to & but it was wrong so I changed it back. We definitely want a union, not the intersection.

PR #2336, opened 2017-05-01
Use bar for precipitation and no zero for temperature and add another…

domoritz:
… example with layer of layer.

screen shot 2017-05-01 at 13 52 24

screen shot 2017-05-01 at 13 53 27

### DIFF
--- a/examples/specs/layer_bar_dual_axis_minmax.vl.json
+++ b/examples/specs/layer_bar_dual_axis_minmax.vl.json
@@ -1,0 +1,64 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "Barchart for precipitation overlayed with linecharts to show temperature range.",
+  "data": {"url": "data/seattle-weather.csv"},
+  "layer": [
+    {
+      "mark": "bar",
+      "encoding": {
+        "x": {
+          "timeUnit": "month",
+          "field": "date",
+          "type": "temporal"
+        },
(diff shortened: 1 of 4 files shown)


examples/specs/layer_bar_dual_axis_minmax.vl.json
```diff
@@ -0,0 +1,64 @@
+{
```
kanitw: We already have layer_box_plot_circle as a good example for layer of layer, so I prefer not to include this.

domoritz: This one is nice though because it is a layer of unit and layer. And it's actually meaningful.

kanitw: So what's your goal here?  

If you want it as an example in the gallery, I don't think it's a good example to include in the gallery.  Even if you are okay with dual axes, a better design would be to use `area` band rather than two lines anyway. 

If you want a test case for layer of layer, then layer_box_plot_circle already serves the purpose and is less controversial example.

domoritz: What's neat about this example is that it uses shared and independent scales in the same layered chart (with layer of layer) for the same channel.

kanitw: Ok.  Make sense.  Let's only keep it as a testing example, not showcase though.

PR #2353, opened 2017-05-06
[WIP] Add Lookup Transform

willium:


src/compile/model.ts
```diff
@@ -361,6 +361,20 @@ export abstract class Model {
...
+  public getLookups(predicate?: (lookup: LookupTransform) => boolean): LookupTransform[] {
+    const lookups: LookupTransform[] = this.transforms.filter((t: Transform): t is LookupTransform => isLookup(t) && (!predicate || predicate(t)));
+    return lookups;
+  }
+}
+
+// above works if we add this, though it should be already added from typescript!!!
+// See this PR https://github.com/Microsoft/TypeScript/issues/7657 which was merged in ~ TS 2.1
+// TODO: remove this before merge. If we can't figure out TS, just have getLookups return Transform[]
+declare global {
```
willium: this is DEFINITELY not staying. it's weird that its even needed since the same line is in Typescript's lib.d.ts and therefore the above filter should throw an error like it does if we remove this global bit. So weird.

domoritz: We should be using ts 2.3. I don't really understand the problem or have a solution.

src/compile/data/source.ts
```diff
@@ -65,13 +66,13 @@ export class SourceNode extends DataFlowNode {
     this._name = name;
   }
 
-  set parent(parent: DataFlowNode) {
-    throw new Error('Source nodes have to be roots.');
-  }
+  // set parent(parent: DataFlowNode) {
```
domoritz: Any reason why you removed this?

willium: yeah. because of ordering I was trying to properly set the parent. I think this was a mistake, but it was what I sort of understood to this point. In my head lookup Source Nodes != root Source Nodes.... which I guess is wrong? Was having difficult reasoning through that.

domoritz: I think source nodes will always be root nodes. They should never have parents.

willium: make sense. :) I was trying to reason my way around the ordering. This was definitely something I wasn't sure about!

kanitw:
Is this one superseded by https://github.com/vega/vega-lite/pull/2375?

willium:
Yes.

PR #2345, opened 2017-05-03
Add UTC Support for Scale and TimeUnit

YuhanLu:
Fix #446, support UTC input and output

## INPUT
To parse data in local time or UTC time, there are two cases: 
1) By default, user will parse data in local time.
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "Google's stock price over time.",
  "data": {
    "values": [
      {"date": "10 Oct 2011 22:48:00"},
      {"date": "11 Oct 2022 23:00:00"}
    ],
    "format": {"type": "json"}
  },
  "mark": "point",
  "encoding": {
    "x": {"timeUnit": "date","field": "date","type": "temporal"},
    "y": {"timeUnit": "hours","field": "date","type": "temporal"}
  }
}
```

screen shot 2017-05-02 at 1 47 20 pm

2) To parse it in utc time, we need to have custom `format` with `utc`
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "Google's stock price over time.",
  "data": {
    "values": [
      {"date": "10 Oct 2011 22:48:00"},
      {"date": "11 Oct 2022 23:00:00"}
    ],
    "format": {"type": "json", "parse": {"date": "utc:'%d %b %Y %H:%M:%S'"}}
  },
  "mark": "point",
  "encoding": {
    "x": {"timeUnit": "date","field": "date","type": "temporal"},
    "y": {"timeUnit": "hours","field": "date","type": "temporal"}
  }
}
```
screen shot 2017-05-02 at 1 50 32 pm

See [vega](https://vega.github.io/vega/docs/data/) for `format`

## Output:
1) User will output local time by default.
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "Google's stock price over time.",
  "data": {
    "values": [
        {
          "date": "Sun, 01 Jan 2012 23:00:00",
          "price": 150
        },
        {
          "date": "Sun, 02 Jan 2012 00:00:00",
          "price": 100
        },
        {
          "date": "Sun, 02 Jan 2012 01:00:00",
          "price": 170
        },
        {
          "date": "Sun, 02 Jan 2012 02:00:00",
          "price": 165
        },
        {
          "date": "Sun, 02 Jan 2012 03:00:00",
          "price": 200
        }
      ]
  },
  "mark": "line",
  "encoding": {
[...]

PR #2354, opened 2017-05-08
Bring back the code to let utc timeunit and utc scale not be consistent

### LINKED ISSUE #2048: Off-by-one year in temporal-year axis
The following spec:

```json
{
  "description": "Line chart of Horsepower over time",
  "mark": "bar",
  "encoding": {
    "x": {"field": "Year", "type": "temporal", "timeUnit": "year"},
    "y": {"field": "Horsepower", "type": "quantitative", "aggregate": "mean"}
  }
}
```

results in the following plot:

Note that the domain of the data is 1970 - 1982 (*not* 1969 - 1981).

Interestingly, the following spec:
```json
{
  "data": {"url": "data/seattle-weather.csv","format": {"type": "csv"}},
  "mark": "bar",
  "encoding": {
    "x": { "field": "date", "type": "temporal", "timeUnit": "year" },
    "y": { "field": "precipitation", "type": "quantitative", "aggregate": "sum" }
  }
}
```

results in the following *correct* plot:

(VL: v2.0.0-alpha.6; VG: v3.0.0-beta.27)

YuhanLu:
After the discussion with @kanitw, we agree on:

### Input
read input data as utc behavior should remain mostly the same as https://github.com/vega/vega-lite/pull/2345 Only difference is for input time data in ISO format, we will parse the data in utc.

### Output
To have user output in utc, we need to let user specify either one of:
1) utc timeunit when you have input data in local time, or 
2) utc scale type when you have input data in utc time
to have the correct output in utc time. But not **both** of utc timeunit and utc scale type since that will cause a trouble by shifting the output twice in vega.

Fix #2048 
The reason for why the #2048 output is wrong is that the input data is in ISO format and we will parse it in utc time automatically. So we need to be consistent to have data output in utc or in local time as well. So the right spec should be either:
1) utc scale type to let user output in utc

```
{
  "description": "Line chart of Horsepower over time",
  "mark": "bar",
  "encoding": {
    "x": {"field": "Year", "type": "temporal", "scale": {"type": "utc"}},
    "y": {"field": "Horsepower", "type": "quantitative", "aggregate": "mean"}
  }
}

```

screen shot 2017-05-08 at 12 22 57 pm

or 

2) utc timeunit to let user output in localtime

```
{
  "description": "Line chart of Horsepower over time",
  "mark": "bar",
  "encoding": {
    "x": {"field": "Year", "type": "temporal", "timeUnit": "utcyear"},
    "y": {"field": "Horsepower", "type": "quantitative", "aggregate": "mean"}
  }
}
```

screen shot 2017-05-08 at 12 24 31 pm

### DIFF
--- a/src/compile/scale/type.ts
+++ b/src/compile/scale/type.ts
@@ -157,7 +157,7 @@ export function fieldDefMatchScaleType(specifiedType: ScaleType, fieldDef: Field
     if (!fieldDef.timeUnit) {
       return contains([ScaleType.TIME, ScaleType.UTC, undefined], specifiedType);
     } else {
-      return contains([ScaleType.UTC, ScaleType.TIME, undefined], specifiedType) || hasDiscreteDomain(specifiedType);
+      return contains([ScaleType.TIME, ScaleType.UTC, undefined], specifiedType) || hasDiscreteDomain(specifiedType);
     }
   } else if (type === Type.QUANTITATIVE) {
     if (fieldDef.bin) {
--- a/src/compile/scale/type.ts
+++ b/src/compile/scale/type.ts
@@ -3,7 +3,7 @@ import * as log from '../../log';
 import {Channel, hasScale, rangeType, supportScaleType} from '../../channel';
 import {Mark} from '../../mark';
 import {ScaleConfig, ScaleType} from '../../scale';
-import {isDiscreteByDefault, isUTCTimeUnit} from '../../timeunit';
+import {isDiscreteByDefault} from '../../timeunit';
 
 import {field, FieldDef} from '../../fielddef';
 import {hasDiscreteDomain} from '../../scale';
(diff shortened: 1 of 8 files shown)


src/compile/scale/type.ts
```diff
@@ -160,11 +157,7 @@ export function fieldDefMatchScaleType(specifiedType: ScaleType, fieldDef: Field
     if (!fieldDef.timeUnit) {
       return contains([ScaleType.TIME, ScaleType.UTC, undefined], specifiedType);
     } else {
-      if (isUTCTimeUnit(fieldDef.timeUnit)) {
-        return contains([ScaleType.UTC, undefined], specifiedType) || hasDiscreteDomain(specifiedType);
-      } else {
-        return contains([ScaleType.TIME, undefined], specifiedType) || hasDiscreteDomain(specifiedType);
-      }
+      return contains([ScaleType.UTC, ScaleType.TIME, undefined], specifiedType) || hasDiscreteDomain(specifiedType);
```
kanitw: `[ScaleType.TIME, ScaleType.UTC, ...` would be nicer

kanitw:
> read input data as utc behavior should remain mostly the same as #2345 Only difference is for input time data in ISO format, we will parse the data in utc.

Does the code here change any behavior at all?  I guess  this is simply correction?

YuhanLu:
Yes, it is a correction.

PR #2356, opened 2017-05-09
set default text alignment to left and enable customized alignment

### LINKED ISSUE #2349: text mark is center aligned and not customizable

YuhanLu:
Fix #2349

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -4782,6 +4782,10 @@
                     ],
                     "type": "string"
                 },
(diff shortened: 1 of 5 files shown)


kanitw:
@YuhanLu  -- test failing

kanitw:
As mentioned in https://github.com/vega/vega-lite/issues/2349, I think you should use another method.

kanitw:
Note that I don't like adding align to text's fieldDef as practically it can be an encoding channel too.

PR #2360, opened 2017-05-12
add BinTransform, TimeUnitTransform, and AggregateTransform

### LINKED ISSUE #2344: Add bin, aggregate, timeUnit to the transform array
The `transform` array should support the following transforms:

- `BinTransform`

```json
{
  "bin": true, "field": "acceleration", "as": "binned_acceleration"
}
```

- `TimeUnitTransform`

```json
{
  "timeUnit": "year", "field": "date", "as": "year_date"
}
```

- `AggregateTransform`

```json
{
  "summarize": [
    {"aggregate": "sum", "field": "acceleration", "as": "sum_acceleration"},
    {"aggregate": "mean", "field": "horsepower", "as": "mean_horsepower"}
  ],
  "groupby": ["origin", "cylinders"]
}
```

## TODO 

- [ ] Add these interfaces to `transform.ts`
- [ ] Modify `parseTransformArray` in `src/compile/data/transforms.ts`
All of these transforms should be able to reuse existing `Bin/TimeUnit/AggregateNode` in the transform pipeline. 

- [ ] Add a test example where we calculate / filter based on aggregate value
  - [ ] Create a "full" example of boxplot with 1.5 IQR whiskers

YuhanLu:
Fix #2344


src/compile/data/transforms.ts
```diff
@@ -65,6 +66,9 @@ export function parseTransformArray(model: Model) {
       node = new CalculateNode(t);
     } else if (isFilter(t)) {
       node = new FilterNode(model, t.filter);
+    } else if (isBin(t)) {
+      const modelWithField = model as ModelWithField;
+      node = BinNode.makeTransform(modelWithField, t);
```
kanitw: 1) As mentioned a few times earlier, in general, you should never cast anything in the main code without a good reason.

2) In this case, I don't see why the model here should be a `ModelWithField` -- it should just be `Model`.

src/transform.ts
```diff
@@ -23,8 +27,74 @@ export interface CalculateTransform {
...
+   */
+  field: string;
+
+  /**
+   * The field for storing the computed formula value.
+   */
+  as: string;
+}
+
+export interface AggregateTransform {
```
kanitw: `SummarizeTransform`

src/transform.ts
```diff
@@ -23,8 +27,74 @@ export interface CalculateTransform {
...
+   */
+  summarize: Aggregrate[];
+
+  /**
+   * Array of fields we will be useing for group by
+   */
+  groupby: VgFieldRef[];
+}
+
+export interface Aggregrate {
```
kanitw: `Summarize`

test/compile/data/bin.test.ts
```diff
@@ -4,14 +4,15 @@ import {assert} from 'chai';
...
+import {BinTransform} from '../../../src/transform';
 import {VgTransform} from '../../../src/vega.schema';
 import {parseUnitModel} from '../../util';
 
 function assemble(model: ModelWithField) {
   return BinNode.make(model).assemble();
 }
 
-describe('compile/data/bin', function() {
+describe.only('compile/data/bin', function() {
```
kanitw: Remove "only"

test/compile/data/bin.test.ts
```diff
@@ -102,4 +103,64 @@ describe('compile/data/bin', function() {
       expr: 'format(datum["bin_maxbins_10_Rotten_Tomatoes_Rating_start"], \'s\') + \' - \' + format(datum["bin_maxbins_10_Rotten_Tomatoes_Rating_end"], \'s\')'
     });
   });
+
+  it ('should add bin transform and correctly apply bin for binned field without custom extent', () => {
```
kanitw: Are you just adding this to improve test coverage? 
(It seems irrelevant to what you are adding in this PR)

test/compile/data/bin.test.ts
```diff
@@ -102,4 +103,64 @@ describe('compile/data/bin', function() {
...
+    });
+  });
+
+  it('should apply the bin transform if specified BinTransform', () => {
+    const t: BinTransform = {bin: {"maxbins" : 10}, field: "Rotten_Tomatoes_Rating", as: "something"};
+    const model = parseUnitModel({
+      data: {url: "data/movies.json"},
+      mark: "circle",
+      transform: [t],
+      encoding: {
```
kanitw: Not sure why you need to add x and color here -- in any case this is a funny spec because you don't really use the output field from `BinTransform`.

test/compile/data/bin.test.ts
```diff
@@ -102,4 +103,64 @@ describe('compile/data/bin', function() {
...
+      field: 'Acceleration',
+      as: ['bin_maxbins_10_Acceleration_start', 'bin_maxbins_10_Acceleration_end'],
+      maxbins: 10,
+      signal: 'bin_maxbins_10_Acceleration_bins',
+      extent: {signal: 'bin_maxbins_10_Acceleration_extent'}
+    });
+  });
+
+  it('should apply the bin transform if specified BinTransform', () => {
+    const t: BinTransform = {bin: {"maxbins" : 10}, field: "Rotten_Tomatoes_Rating", as: "something"};
```
kanitw: Can you name it a bit better than something

src/compile/data/bin.ts
```diff
@@ -89,6 +90,38 @@ export class BinNode extends DataFlowNode {
     return new BinNode(bins);
   }
 
+  public static makeTransform(model: ModelWithField, t: BinTransform) {
```
kanitw: As mentioned in person, you can name this `makeBinFromTransform` and rename the other make to `makeBinFromEncoding`.

src/compile/data/bin.ts
```diff
@@ -89,6 +90,38 @@ export class BinNode extends DataFlowNode {
...
+    const bins: Dict<BinComponent> = {};
+    model.forEachFieldDef((fieldDef, channel) => {
```
kanitw: Why is this calling `forEachFieldDef`, which basically enumerates `encoding`?  

This bin in the transform basically should be orthogonal to the encoding.

src/compile/data/bin.ts
```diff
@@ -89,6 +90,27 @@ export class BinNode extends DataFlowNode {
     return new BinNode(bins);
   }
 
+  public static makeBinFromTransform(model: Model, t: BinTransform) {
+    const bins: Dict<BinComponent> = {};
+
+    if (t.bin) {
+      const bin: Bin = isBoolean(t.bin) ? {} : t.bin;
+      const key = `${binToString(t.bin)}_${t.field}`;
```
kanitw: We need to augment `bin` with `maxbins` if both `maxbins` and step` are` not provided like in `normalize` in fielddef.ts

kanitw: (You might wanna extract `normalizeBin` method from `normalize` in `fielddef.ts` and put that in `bin.ts` so that we can reuse them in both places)

src/compile/data/bin.ts
```diff
@@ -89,6 +90,27 @@ export class BinNode extends DataFlowNode {
...
+      bins[key] = {
+        bin: bin,
+        field: t.field,
+        as: [binToString(t.bin) + '_' + t.field + '_start', binToString(t.bin) + '_' + t.field + '_end'],
```
kanitw: Why is this different from the logic from binFromEncoding?

src/compile/data/bin.ts
```diff
@@ -89,6 +90,27 @@ export class BinNode extends DataFlowNode {
     return new BinNode(bins);
   }
 
+  public static makeBinFromTransform(model: Model, t: BinTransform) {
+    const bins: Dict<BinComponent> = {};
+
+    if (t.bin) {
```
kanitw: Can `t.bin` really be falsy here?

src/compile/data/bin.ts
```diff
@@ -89,6 +90,27 @@ export class BinNode extends DataFlowNode {
...
+      const key = `${binToString(t.bin)}_${t.field}`;
+      bins[key] = {
+        bin: bin,
+        field: t.field,
+        as: [binToString(t.bin) + '_' + t.field + '_start', binToString(t.bin) + '_' + t.field + '_end'],
+        signal: model.getName(`${key}_bins`),
+        extentSignal: model.getName(key + '_extent')
+      };
+    }
+    if (Object.keys(bins).length === 0) {
```
kanitw: Can the length really be 0 here?

src/compile/data/bin.ts
```diff
@@ -89,6 +90,27 @@ export class BinNode extends DataFlowNode {
     return new BinNode(bins);
   }
 
+  public static makeBinFromTransform(model: Model, t: BinTransform) {
+    const bins: Dict<BinComponent> = {};
+
+    if (t.bin) {
+      const bin: Bin = isBoolean(t.bin) ? {} : t.bin;
+      const key = `${binToString(t.bin)}_${t.field}`;
+      bins[key] = {
```
kanitw: You might want to extract a method so that we use common method to create the object here.

src/compile/data/timeunit.ts
```diff
@@ -44,6 +45,24 @@ export class TimeUnitNode extends DataFlowNode {
     return new TimeUnitNode(formula);
   }
 
+  public static makeFromTransfrom(model: Model, t: TimeUnitTransform) {
+    const timeUnitComponent: Dict<TimeUnitComponent> = {};
+      if (t.timeUnit) {
```
kanitw: 1) Why tabbed?
2) Can TimeUnit really be falsy here?

src/compile/data/timeunit.ts
```diff
@@ -44,6 +45,24 @@ export class TimeUnitNode extends DataFlowNode {
...
+        timeUnitComponent[t.field] = {
+          as: t.as,
+          timeUnit: t.timeUnit,
+          field: t.field
+        };
+      }
+
+
+    if (Object.keys(timeUnitComponent).length === 0) {
+      return null;
```
kanitw: Can the object really be empty here?

src/compile/data/transforms.ts
```diff
@@ -65,6 +68,12 @@ export function parseTransformArray(model: Model) {
       node = new CalculateNode(t);
     } else if (isFilter(t)) {
       node = new FilterNode(model, t.filter);
+    } else if (isBin(t)) {
```
kanitw: Create a test that include all types of transform to test this method?

test/compile/data/aggregate.test.ts
```diff
@@ -114,5 +115,29 @@ describe('compile/data/summary', function () {
         fields: ['*']
       });
     });
+
+    it('should produce the correct summary component from transform array', function() {
+      const t: SummarizeTransform = {summarize: [
```
kanitw: tab better

test/compile/data/bin.test.ts
```diff
@@ -102,4 +107,33 @@ describe('compile/data/bin', function() {
       expr: 'format(datum["bin_maxbins_10_Rotten_Tomatoes_Rating_start"], \'s\') + \' - \' + format(datum["bin_maxbins_10_Rotten_Tomatoes_Rating_end"], \'s\')'
     });
   });
+
+  it('should add bin transform from transform arrat and correctly apply bin with custom extent', function() {
+    const t: BinTransform = {bin: {extent: [0, 100]},
+                  field: 'Acceleration',
+                  as: 'Bin_Transform'};
+     const model = parseUnitModel({
```
kanitw: not align

test/compile/data/bin.test.ts
```diff
@@ -102,4 +107,33 @@ describe('compile/data/bin', function() {
       expr: 'format(datum["bin_maxbins_10_Rotten_Tomatoes_Rating_start"], \'s\') + \' - \' + format(datum["bin_maxbins_10_Rotten_Tomatoes_Rating_end"], \'s\')'
     });
   });
+
+  it('should add bin transform from transform arrat and correctly apply bin with custom extent', function() {
+    const t: BinTransform = {bin: {extent: [0, 100]},
```
kanitw: bad tab

src/compile/data/aggregate.ts
```diff
@@ -99,7 +101,36 @@ export class AggregateNode extends DataFlowNode {
...
+    for(const s of t.summarize) {
+      if (s.aggregate) {
+        if (s.aggregate === 'count') {
+          meas['*'] = meas['*'] || {};
+          meas['*']['count'] = true;
+          as['*'] = s.as;
+        } else {
+          meas[s.field] = meas[s.field] || {};
+          meas[s.field][s.aggregate] = true;
+          as[s.field] = s.as;
```
kanitw: Or even just `Dict<Dict<string>>` and always output name.

src/compile/data/bin.ts
YuhanLu: I manually extracted the correct logic from what we have in the other method without encoding channel

src/compile/data/bin.ts
```diff
@@ -93,20 +97,7 @@ export class BinNode extends DataFlowNode {
...
-        field: t.field,
-        as: [binToString(t.bin) + '_' + t.field + '_start', binToString(t.bin) + '_' + t.field + '_end'],
-        signal: model.getName(`${key}_bins`),
-        extentSignal: model.getName(key + '_extent')
-      };
-    }
-    if (Object.keys(bins).length === 0) {
-      return null;
-    }
+    createBin(bins, t, model, undefined);
```
kanitw:
```
const key = binKey(...);

return new BinNode({
  [key]: createBinComponent(...)
});
```

src/compile/data/timeunit.ts
```diff
@@ -47,18 +47,11 @@ export class TimeUnitNode extends DataFlowNode {
...
-          timeUnit: t.timeUnit,
-          field: t.field
-        };
-      }
-
-
-    if (Object.keys(timeUnitComponent).length === 0) {
-      return null;
-    }
+    timeUnitComponent[t.field] = {
```
kanitw: Just make this one line
```
return new TimeUnitNode({
  [t.field]: ...
}
```

src/fielddef.ts
```diff
@@ -285,22 +286,9 @@ export function normalize(channelDef: ChannelDef<string>, channel: Channel) {
...
-      } else if (!bin.maxbins && !bin.step) {
-        fieldDef = {
-          ...fieldDef,
-          bin: {
-            ...bin,
-            maxbins: autoMaxBins(channel)
-          }
-        };
-      }
+    const bin = normalizeBin(fieldDef.bin, channel);
```
kanitw: 1) Why should you call this no matter fieldDef.bin is truthy or not?  The predicate for if should be `fieldDef.bin` and `normalizeBin` should be called inside.  
2) Setting `fieldDef.bin` cause side effect.  Why don't you follow the pattern we had earlier?

src/fielddef.ts
```diff
@@ -332,6 +320,18 @@ export function normalize(channelDef: ChannelDef<string>, channel: Channel) {
   return channelDef;
 }
 
+export function normalizeBin(bin: Bin|boolean, channel: Channel) {
+  if (bin) {
```
kanitw: why `if (bin)` here?

test/compile/data/aggregate.test.ts
kanitw: @YuhanLu  This bad tab is still here...

test/compile/data/transform.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      const result = parseTransformArray(model);
+      assert.isTrue(result.first instanceof CalculateNode);
+      assert.isTrue(result.last instanceof FilterNode);
+    });
+
+    it('should return a BinNode node and a TimeUnitNode', () => {
+      const model = parseUnitModel({
+        data: {values: []},
+        mark: 'point',
+        transform: [{bin: true, field: 'field', as: 'as'}, {timeUnit: 'month', field: 'field', as: 'as'}],
```
kanitw: This is a bad example -- why name two different things with the same name (`'as'`) ?

test/compile/data/transform.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      const result = parseTransformArray(model);
+      assert.isTrue(result.first instanceof BinNode);
+      assert.isTrue(result.last instanceof TimeUnitNode);
+    });
+
+    it('should return a BinNode and a AggregateNode', () => {
+      const model = parseUnitModel({
+        data: {values: []},
+        mark: 'point',
+        transform: [{bin: true, field: 'field', as: 'as'}, {summarize: [{aggregate: 'count', field: 'f', as: 'a'}, {aggregate: 'sum', field: 'f', as: 'a'}], groupby: ['field']}],
```
kanitw: This is a bad example -- why name two different things with the same name ('a') ?

test/compile/data/transform.test.ts
```diff
@@ -0,0 +1,55 @@
+import {assert} from 'chai';
+import {AggregateNode} from '../../../src/compile/data/aggregate';
+import {BinNode} from '../../../src/compile/data/bin';
+import {TimeUnitNode} from '../../../src/compile/data/timeunit';
+import {CalculateNode, FilterNode, parseTransformArray} from '../../../src/compile/data/transforms';
+import {TimeUnit} from '../../../src/timeunit';
+import {parseUnitModel} from '../../util';
+describe('compile/data/transform', () => {
```
kanitw: add a line break between import and describe.

src/compile/data/aggregate.ts
```diff
@@ -117,7 +117,8 @@ export class AggregateNode extends DataFlowNode {
         } else {
           meas[s.field] = meas[s.field] || {};
           meas[s.field][s.aggregate] = true;
-          as[s.field] = s.as;
+          as[s.field] = as[s.field] || {};
```
kanitw: Remove `as` -- make `meas` a `Dict<Dict<string>>` and always include the correct as. 
(For encoding's aggregate, call `field(...)` with the right params.

kanitw: For summarize transform, automatically add `as` if `as` is not defined (using `||`)

src/fielddef.ts
YuhanLu: Why setting fieldDef.bin will cause side effect? I see what we had before is actually overwriting fieldDef.bin, isn't it?

src/compile/data/transforms.ts
YuhanLu: Done

src/compile/data/aggregate.ts
```diff
@@ -26,18 +26,18 @@ function addDimension(dims: {[field: string]: boolean}, fieldDef: FieldDef<strin
...
       // when we merge a measure, we either have to add an aggregation operator or even a new field
-      const ops = childMeasures[field];
+      const ops = childMeasures[f];
       for (const op in ops) {
         if (ops.hasOwnProperty(op)) {
-          if (field in parentMeasures) {
+          if (f in parentMeasures) {
             // add operator to existing measure field
-            parentMeasures[field][op] = true;
+            parentMeasures[f][op] = field({field: f});
```
kanitw: Don't call `field`

src/compile/data/aggregate.ts
```diff
@@ -26,18 +26,18 @@ function addDimension(dims: {[field: string]: boolean}, fieldDef: FieldDef<strin
...
           } else {
-            parentMeasures[field] = {op: true};
+            parentMeasures[f] = {op: field({field: f})};
```
kanitw: Don't call `field`

src/fielddef.ts
```diff
@@ -175,7 +176,7 @@ export interface FieldRefOption {
   aggregate?: AggregateOp;
 }
 
-export function field(fieldDef: FieldDef<string>, opt: FieldRefOption = {}): string {
+export function field(fieldDef: FieldDef<string>|BinTransform, opt: FieldRefOption = {}): string {
```
kanitw: I already remove `|BinTransform` for you -- why you bring it back?

src/compile/data/aggregate.ts
```diff
@@ -139,18 +165,25 @@ export class AggregateNode extends DataFlowNode {
...
 
-    return {
+    let result: VgAggregateTransform = {
       type: 'aggregate',
       groupby: keys(this.dimensions),
       ops,
       fields
     };
+
+    if (as.length !== 0) {
```
kanitw: I don't think as.length can really be 0

src/compile/data/aggregate.ts
kanitw: I guess `ops[op]` is the right value here.

One potential issue here is that if `parentMeasures[f][op] ` already exist, and `ops[op]` is different, the merge will be incorrect.  The right solution is to make measure `Dict<Dict<string>>` -- but that's hairy -- feel free to file a bug  with very descriptive description.

src/compile/data/aggregate.ts
kanitw: See https://github.com/vega/vega-lite/pull/2360#discussion_r116889012

kanitw:
One last thing -- create example.

kanitw:
One last thing -- add example :)

examples/vg-specs/area.vg.json
```diff
@@ -33,6 +33,9 @@
                     ],
                     "fields": [
                         "count"
+                    ],
+                    "as": [
```
domoritz: Why do we add this all of a sudden? Is it necessary? I think this should only be generated if we don't use the default.

kanitw: 1. This makes the code way more maintainable
2. This makes the output code explicit about field names, which is not necessary a bad thing.

domoritz: I was just wondering why this is on this PR. Seems a bit unrelated but whatever.

kanitw: (It's way more maintainable because AggregateNode from encoding become way easier to merge with AggregateNode from transform)

kanitw:
@mattwchun FYI this is merged.

PR #2361, opened 2017-05-12
Fix json-e file caused by sed

### LINKED ISSUE #2316: What is the json-e file?
This is in my build folder:

screen shot 2017-04-28 at 23 19 19

YuhanLu:
Fix #2316

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,8 +45,8 @@
     "poststart": "rm examples/all-examples.json",
 
     "preschema": "npm run prebuild",
-    "schema": "typescript-json-schema --required true --noExtraProps true src/spec.ts TopLevelExtendedSpec > build/vega-lite-schema.json && rm -f _data/vega-lite-schema.json && npm run replaceschema && cp build/vega-lite-schema.json _data/",
-    "replaceschema": "scripts/rename-schema.sh",
+    "schema": "typescript-json-schema --required true --noExtraProps true src/spec.ts TopLevelExtendedSpec > build/vega-lite-schema.json && rm -f _data/vega-lite-schema.json && npm run renameschema && cp build/vega-lite-schema.json _data/",
+    "renameschema": "scripts/rename-schema.sh",
     "presite": "npm run build && npm run data && npm run build:site && npm run build:toc && npm run build:versions",
     "site": "bundle exec jekyll serve",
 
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
 
     "preschema": "npm run prebuild",
     "schema": "typescript-json-schema --required true --noExtraProps true src/spec.ts TopLevelExtendedSpec > build/vega-lite-schema.json && rm -f _data/vega-lite-schema.json && npm run replaceschema && cp build/vega-lite-schema.json _data/",
-    "replaceschema": "sed -i -e 's/GenericLayerSpec<.*..\\./LayerSpec/g' build/vega-lite-schema.json && sed -i -e 's/GenericRepeatSpec<.*..\\./RepeatSpec/g' build/vega-lite-schema.json && sed -i -e 's/GenericUnitSpec<Encoding<.*..\\./UnitSpec/g' build/vega-lite-schema.json && sed -i -e 's/GenericUnitSpec<EncodingWithFacet<.*..\\./FacetedUnitSpec/g' build/vega-lite-schema.json && sed -i -e 's/<Field>//g' build/vega-lite-schema.json",
+    "replaceschema": "scripts/rename-schema.sh",
```
kanitw: Pick one name and make the script file and the script name match.

kanitw:
1. Can you explain more why doing this solves the problem?
2. The test is still failing.

YuhanLu:
There was a extra file with json-e suffix is because I was using the wrong command `sed -i -e`, whose second parameter should be the name of the backup of the original file. So if i use `sed -i `` ` instead, I will skip the backup and do a inline replace instaed.

YuhanLu:
It is actually the problem of sed working differently under mac os and linux. Mac interpret `sed -i -e` as a -e suffix. And linux understand `sed -i` perfectly but mac required another parameter so `sed -i '' ` is for mac...

YuhanLu:
http://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux
Based on the solution above, I will use perl here since it works for both mac os and linux.

domoritz:
Thank you @YuhanLu!

Issue #1539, opened 2016-09-06
Better default axis number format


kanitw:
SI seems confuses users a lot!

domoritz:
Maybe we shouldn't bother for the default but for voyager/polestar use scientific notation to avoid super long labels.

kanitw:
- [ ] Also make sure that bin range are not weird as described in #1757

kanitw:
- [ ] Make sure this case works https://github.com/vega/vega-lite/issues/1459

jheer:
OK, I just did some basic testing with Vega 3's default settings for linear, log and sqrt/pow scales. All have pretty reasonable defaults as-is (with no special formatting directive applied), but could be further refined in how "extreme" values are handled.

- `linear`, `sqrt`, `pow` - These all use the same default setting, which generates a linear set of ticks with precision automatically configured by the scale domain. Problems arise when the numbers get very big or very small: the labels grow exceeding long with leading or trailing zeros. These labels switch to exponential notation once values are greater than or equal to `1e+21`.
- `log` - This uses a special formatter that determines precision in a non-linear fashion. It automatically switches to using exponential labels once the digit precision exceeds a certain threshold: `1e-7` for small numbers, `1e+12` for large numbers.

Questions:
- For `linear`, etc should we institute similar policies as we use for `log` (or vice versa) and switch to exponential (scientific) notation for large/small numbers at the same threshold values?
- What cutoffs should we use for switching notation? For example, the log formatter currently uses [a default precision of 12 digits](https://github.com/vega/vega-encode/blob/master/src/ticks.js#L55).
- One other issue is that the formatters currently interleave formats: for example both `1,000,000` and `1e+12` might be visible simultaneously. Would we want the formatters to switch such that all labels use the same style? I could see arguments in both directions (stability vs. consistency), so would like to get other folks' opinions here.

cc @kanitw @domoritz @arvind

jheer:
Also, I should note that the formatting for `linear`, _etc_ is currently done using d3's number formatting. So these defaults are inherited. Changes/fixes of a general nature might require edits / pull requests to the d3-format repo, or additional wrapper code in Vega (as we currently do for log scales).

kanitw:
Here are my quick opinions from thinking *without* experimenting:

> What cutoffs should we use for switching notation? For example, the log formatter currently uses a default precision of 12 digits.

[...]

PR #2369, opened 2017-05-15
 Support Formatting In Nominal And Ordinal Axes

### LINKED ISSUE #1763: Cannot format row/col

NeelMohapatra:
Fixes #1763

### DIFF
--- a/src/compile/axis/encode.ts
+++ b/src/compile/axis/encode.ts
@@ -5,7 +5,7 @@ import {ScaleType} from '../../scale';
 import {NOMINAL, ORDINAL, TEMPORAL} from '../../type';
 import {contains, keys} from '../../util';
 import {AxisOrient, VgAxis} from '../../vega.schema';
-import {timeFormatExpression} from '../common';
+import {guideLabels, numberFormat, timeFormatExpression} from '../common';
 import {Split} from '../split';
 import {UnitModel} from '../unit';
 
@@ -22,14 +22,10 @@ export function labels(model: UnitModel, channel: SpatialScaleChannel, specified
   let labelsSpec: any = {};
 
   // Text
-  if (fieldDef.type === TEMPORAL) {
-    const isUTCScale = model.getScaleComponent(channel).get('type') === ScaleType.UTC;
-
-    labelsSpec.text =  {
-      signal: timeFormatExpression('datum.value', fieldDef.timeUnit, axis.format, config.axis.shortTimeLabels, config.timeFormat, isUTCScale)
-    };
+  const text = guideLabels(fieldDef, config, model, channel, axis);
+  if (text) {
+    labelsSpec.text = text;
   }
-
   // Label Angle
   const angle = labelAngle(axis, channel, fieldDef);
   if (angle) {
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -1,13 +1,19 @@
-
+import {Channel, COLUMN, NonspatialScaleChannel, ROW, SpatialScaleChannel, TEXT, TOOLTIP} from '../channel';
 import {CellConfig, Config} from '../config';
 import {field, FieldDef, FieldRefOption, isScaleFieldDef, OrderFieldDef} from '../fielddef';
 import {MarkConfig, MarkDef, TextConfig} from '../mark';
 import {ScaleType} from '../scale';
 import {TimeUnit} from '../timeunit';
 import {formatExpression} from '../timeunit';
-import {QUANTITATIVE} from '../type';
-import {isArray} from '../util';
+import {FormatType, NOMINAL, ORDINAL, QUANTITATIVE, TEMPORAL} from '../type';
+import {contains, duplicate, isArray} from '../util';
 import {VgEncodeEntry, VgMarkConfig, VgSort} from '../vega.schema';
+import {ConcatModel} from './concat';
+import {FacetModel} from './facet';
+import {LayerModel} from './layer';
+import {Model} from './model';
+import {RepeatModel} from './repeat';
+import {RepeaterValue} from './repeater';
 import {Explicit} from './split';
 import {UnitModel} from './unit';
 
@@ -70,24 +76,25 @@ export function getMarkConfig<P extends keyof MarkConfig>(prop: P, mark: MarkDef
   return value;
 }
 
-export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: string, expr: 'datum' | 'parent', config: Config, useBinRange?: boolean) {
-  const format = numberFormat(fieldDef, specifiedFormat, config);
-  if (fieldDef.bin) {
-    if (useBinRange) {
-      // For bin range, no need to apply format as the formula that creates range already include format
-      return {signal: field(fieldDef, {expr, binSuffix: 'range'})};
-    } else {
-      const startField = field(fieldDef, {expr});
-      const endField = field(fieldDef, {expr, binSuffix: 'end'});
-      return {
-        signal: binFormatExpression(startField, endField, format, config)
-      };
+export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: string, expr: 'datum' | 'parent', config: Config, formatType: FormatType | undefined, useBinRange?: boolean) {
+  if (fieldDef.type === 'quantitative' || formatType === 'number') {
+    const format = numberFormat(fieldDef, specifiedFormat, config);
+    if (fieldDef.bin) {
+      if (useBinRange) {
+        // For bin range, no need to apply format as the formula that creates range already include format
+        return {signal: field(fieldDef, {expr, binSuffix: 'range'})};
+      } else {
+        const startField = field(fieldDef, {expr});
+        const endField = field(fieldDef, {expr, binSuffix: 'end'});
+        return {
+          signal: binFormatExpression(startField, endField, format, config)
+        };
+      }
     }
-  } else if (fieldDef.type === 'quantitative') {
     return {
       signal: `${formatExpr(field(fieldDef, {expr}), format)}`
     };
-  } else if (fieldDef.type === 'temporal') {
+  } else if (fieldDef.type === 'temporal' || contains(['utc', 'time'], formatType)) {
     const isUTCScale = isScaleFieldDef(fieldDef) && fieldDef['scale'] && fieldDef['scale'].type === ScaleType.UTC;
     return {
       signal: timeFormatExpression(field(fieldDef, {expr}), fieldDef.timeUnit, specifiedFormat, config.text.shortTimeLabels, config.timeFormat, isUTCScale)
@@ -112,18 +119,13 @@ export function getSpecifiedOrDefaultValue<T>(specifiedValue: T, defaultValue: T
  * @param format explicitly specified format
  */
 export function numberFormat(fieldDef: FieldDef<string>, specifiedFormat: string, config: Config) {
-  if (fieldDef.type === QUANTITATIVE) {
-    // add number format for quantitative type only
-
-    // Specified format in axis/legend has higher precedence than fieldDef.format
-    if (specifiedFormat) {
-      return specifiedFormat;
-    }
-
-    // TODO: need to make this work correctly for numeric ordinal / nominal type
-    return config.numberFormat;
+  // Specified format in axis/legend has higher precedence than fieldDef.format
+  if (specifiedFormat) {
+    return specifiedFormat;
   }
-  return undefined;
+
+  // TODO: need to make this work correctly for numeric ordinal / nominal type
+  return config.numberFormat;
 }
 
 function formatExpr(field: string, format: string) {
@@ -146,12 +148,8 @@ export function binFormatExpression(startField: string, endField: string, format
 export function timeFormatExpression(field: string, timeUnit: TimeUnit, format: string, shortTimeLabels: boolean, timeFormatConfig: string, isUTCScale: boolean): string {
   if (!timeUnit || format) {
     // If there is not time unit, or if user explicitly specify format for axis/legend/text.
-    const _format = format || timeFormatConfig; // only use config.timeFormat if there is no timeUnit.
-    if (isUTCScale) {
-      return `utcFormat(${field}, '${_format}')`;
-    } else {
-      return `timeFormat(${field}, '${_format}')`;
-    }
+    // only use config.timeFormat if there is no timeUnit.
+    return `${isUTCScale ? 'utc' : 'time'}Format(${field}, '${format || timeFormatConfig}')`;
   } else {
     return formatExpression(timeUnit, field, shortTimeLabels, isUTCScale);
   }
@@ -175,4 +173,24 @@ export function titleMerger(v1: Explicit<string>, v2: Explicit<string>) {
       v1.value : // if title is the same just use one of them
       v1.value + ', ' + v2.value // join title with comma if different
   };
+}
+
+export function guideLabels(fieldDef: FieldDef<string>, config: Config, model: Model, channel: SpatialScaleChannel | NonspatialScaleChannel, label: any) {
+    if (fieldDef.type === TEMPORAL) {
+    const isUTCScale = model.getScaleComponent(channel).get('type') === ScaleType.UTC;
+    return {
+      signal: timeFormatExpression('datum.value', fieldDef.timeUnit, label.format, config.legend.shortTimeLabels, config.timeFormat, isUTCScale)
+    };
+  } else if ((fieldDef.type === NOMINAL || fieldDef.type === ORDINAL) && label.format) {
+    if (label.formatType === 'number') {
+      return {
+        signal: `format(${fieldDef.field}, '${numberFormat(fieldDef, label.format, config)}')`
+      };
+    } else if (label.formatType) {
+      return {
+        signal: timeFormatExpression('datum.value', fieldDef.timeUnit, label.format, config.legend.shortTimeLabels, config.timeFormat, label.formatType === 'utc')
+      };
+    }
+  }
+  return undefined;
 }
--- a/src/compile/layout/header.ts
+++ b/src/compile/layout/header.ts
@@ -91,7 +91,7 @@ export function getHeaderGroup(model: Model, channel: HeaderChannel, headerType:
       const format = facetFieldDef.header ? facetFieldDef.header.format : undefined;
 
       title = {
-        text: formatSignalRef(facetFieldDef, format, 'parent', model.config, true),
+        text: formatSignalRef(facetFieldDef, format, 'parent', model.config, undefined, true), // undefined passed since formatType not supported in headers
         offset: 10,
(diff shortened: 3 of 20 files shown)


src/compile/common.ts
```diff
@@ -99,10 +99,19 @@ export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: str
   } else if (fieldDef.type === 'temporal') {
     const isUTCScale = isScaleFieldDef(fieldDef) && fieldDef['scale'] && fieldDef['scale'].type === ScaleType.UTC;
     return {
-      signal: timeFormatExpression(field(fieldDef, {expr}), fieldDef.timeUnit, specifiedFormat, config.text.shortTimeLabels, config.timeFormat, isUTCScale)
+      signal: timeFormatExpression(field(fieldDef, {expr}), fieldDef.timeUnit,
+        specifiedFormat, config.text.shortTimeLabels, config.timeFormat, isUTCScale)
     };
   } else {
-    return {signal: field(fieldDef, {expr})};
+    if (!specifiedFormat || specifiedFormat === 'number') {
```
kanitw: The predicate logic here seems pretty weird.    From a quick skim, there are multiple problems:

- `format` and `formatType` are two different things. 
- There are still many cases where you shouldn't add `format` (e.g., normal string)
- Why is else case adding timeFormatExpression

Revise your logic!

NeelMohapatra: It's WIP. I haven't even finished the PR. Did I accidentally assign you to review it? I am sorry if I did :|

kanitw: ok no worries then!

kanitw:
@NeelMohapatra I saw you pushed a few things, but tests are still failing so I assume this is not ready.  Let us know when it is then!

domoritz:
@NeelMohapatra Can you rebase and finish this PR?

examples/vg-specs/area_vertical.vg.json
```diff
@@ -166,6 +166,7 @@
         },
         {
             "scale": "y",
+            "format": "s",
```
domoritz: `s` is a bad format for year.

NeelMohapatra: Hmm, have to figure out numberFormat. Will talk today in meeting

NeelMohapatra: I'm not quite sure what the restrictions should be on this function

src/compile/common.ts
```diff
@@ -114,9 +139,6 @@ export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: str
 export function numberFormat(fieldDef: FieldDef<string>, specifiedFormat: string, config: Config, channel: Channel) {
   // Specified format in axis/legend has higher precedence than fieldDef.format
   const format = specifiedFormat;
-  if (fieldDef.type === QUANTITATIVE) {
```
NeelMohapatra: @kanitw  I need advice about what I should do here? When should I return a number format apart from when fieldDef.type is quantitative? I think my visualizations are fine but it's still bad style and it might be breaking something else somewhere. What if the type is O, and the user wants to get integers?

domoritz: I think the type can be O and we might still want to return a format. Look through some specs and see what makes the most sense. Also, we will soon change the default away from `s`. I don't think it should make a difference to you, though.

domoritz: https://github.com/vega/vega-lite/issues/1539

NeelMohapatra: Okay and what about Nominal types?

domoritz: I think even a nominal type can be a number, right? My gut says no.

NeelMohapatra: Lol, my bad

kanitw:
@NeelMohapatra  -- we merged the format issue -- could you please finish this so we can merge :)

src/compile/axis/encode.ts
```diff
@@ -27,8 +27,17 @@ export function labels(model: UnitModel, channel: SpatialScaleChannel, specified
     labelsSpec.text =  {
       signal: timeFormatExpression('datum.value', fieldDef.timeUnit, axis.format, config.axis.shortTimeLabels, config.timeFormat, isUTCScale)
     };
+  } else if ((fieldDef.type === NOMINAL || fieldDef.type === ORDINAL) && axis.format) {
+    if (fieldDef['formatType'] === 'number') {
+      labelsSpec.text = {
+        signal: `format(${fieldDef.field}, '${numberFormat(fieldDef, axis.format, config)}')`
+      };
+    } else if (fieldDef['formatType']) {
```
kanitw: This seems like timeFormat is the default, which is incorrect.  Number should be the default.

src/compile/axis/encode.ts
```diff
@@ -27,8 +27,17 @@ export function labels(model: UnitModel, channel: SpatialScaleChannel, specified
     labelsSpec.text =  {
       signal: timeFormatExpression('datum.value', fieldDef.timeUnit, axis.format, config.axis.shortTimeLabels, config.timeFormat, isUTCScale)
     };
+  } else if ((fieldDef.type === NOMINAL || fieldDef.type === ORDINAL) && axis.format) {
+    if (fieldDef['formatType'] === 'number') {
```
kanitw: Why `fieldDef['formatType']` ?   First of all, trying to access value with string is a hack, which can be risky.  In this case, you add `formatType` to axis but access it from fieldDef, so this is definitely wrong.  Maybe you should have access this from `axis.formatType` instead?

src/compile/axis/encode.ts
```diff
@@ -27,8 +27,17 @@ export function labels(model: UnitModel, channel: SpatialScaleChannel, specified
...
+      labelsSpec.text = {
+        signal: `format(${fieldDef.field}, '${numberFormat(fieldDef, axis.format, config)}')`
+      };
+    } else if (fieldDef['formatType']) {
+      labelsSpec.text = {
+        signal: timeFormatExpression('datum.value', fieldDef.timeUnit, axis.format, config.legend.shortTimeLabels, config.timeFormat, fieldDef['formatType'] === 'utc')
```
kanitw: Same problem with formatType here

src/compile/common.ts
```diff
@@ -110,7 +136,7 @@ export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: str
  * @param format explicitly specified format
  */
 export function numberFormat(fieldDef: FieldDef<string>, specifiedFormat: string, config: Config) {
-  if (fieldDef.type === QUANTITATIVE) {
+  if (fieldDef.type === QUANTITATIVE || fieldDef.type === ORDINAL) {
```
kanitw: Not sure why only ORDINAL is added here?

src/compile/facet.ts
```diff
@@ -155,7 +155,7 @@ export class FacetModel extends ModelWithField {
 
       this.component.layoutHeaders[channel] = {
         title,
-        fieldRef: formatSignalRef(fieldDef, header.format, 'parent', this.config, true),
+        fieldRef: formatSignalRef(fieldDef, header.format, 'parent', this.config, fieldDef['formatType'] || 'number', channel, true),
```
kanitw: Same problem with formatType here.

src/compile/mark/valueref.ts
```diff
@@ -149,7 +149,7 @@ export function text(textDef: Conditional<TextFieldDef<string>, ValueDef<any>>,
   // text
   if (textDef) {
     if (isFieldDef(textDef)) {
-      return formatSignalRef(textDef, textDef.format, 'datum', config);
+      return formatSignalRef(textDef, textDef.format, 'datum', config, textDef.formatType || 'number', 'text');
```
kanitw: I don't think it's always the text channel here.  (It could be a `tooltip` when ref.text is called via `mixins.text`)

src/fielddef.ts
```diff
@@ -169,6 +169,8 @@ export interface TextFieldDef<F> extends FieldDef<F> {
    * The formatting pattern for text value. If not defined, this will be determined automatically.
    */
   format?: string;
+
+  formatType?: 'number' | 'time' | 'utc';
```
kanitw: Declare 
```
export type FormatType =  'number' | 'time' | 'utc'
```
and consistently use it across all declarations

test/compile/common.test.ts
```diff
@@ -1,72 +1,92 @@
...
+import {formatSignalRef, numberFormat, timeFormatExpression} from '../../src/compile/common';
 import {defaultConfig} from '../../src/config';
 import {field, FieldDef} from '../../src/fielddef';
 import {TimeUnit} from '../../src/timeunit';
 import {NOMINAL, ORDINAL, QUANTITATIVE, TEMPORAL} from '../../src/type';
 
-describe('Common', function() {
-  describe('timeFormat()', function() {
-    it('should get the right time expression for month with shortTimeLabels=true', function() {
+describe('Common', () => {
```
kanitw: (Minor Tip) In general, you would want to separate refactoring from feature implementation so it's easier to review.

src/compile/common.ts
```diff
@@ -77,7 +77,34 @@ export function getMarkConfig<P extends keyof MarkConfig>(prop: P, mark: Mark, c
...
+                signal: `format(${field(fieldDef, {expr, binSuffix: 'start'})}, '${format}')+'-'+format(${field(fieldDef, {expr, binSuffix: 'end'})}, '${format}')`
+              };
+            }
+          } else {
+            return {
+              signal: `format(${field(fieldDef, {expr})}, '${format}')`
+            };
+          }
+        }
+      } else {
```
kanitw: Again, this makes timeFormat the default.

src/compile/common.ts
```diff
@@ -77,7 +77,34 @@ export function getMarkConfig<P extends keyof MarkConfig>(prop: P, mark: Mark, c
   return config.mark[prop];
 }
 
-export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: string, expr: 'datum' | 'parent', config: Config, useBinRange?: boolean) {
+export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: string, expr: 'datum' | 'parent', config: Config, formatType: 'number' | 'time' | 'utc', channel: Channel, useBinRange?: boolean) {
+  if (channel === TEXT || channel === ROW || channel === COLUMN || channel === TOOLTIP) {
```
kanitw: Can you explain why we need to add so many channels check up here? 

The logic of this chunk is very redundant with the original group of code. 

I guess the difference is whether a channel has a scale or not, but copying the same chunk of code and run it twice is definitely no a good practice better only create conditions for necessary lines.

src/compile/common.ts
kanitw: it seems like you're treating formatType is timeForamt by default.

src/compile/legend/encode.ts
```diff
@@ -82,6 +83,20 @@ export function labels(fieldDef: FieldDef<string>, labelsSpec: any, model: UnitM
         signal: timeFormatExpression('datum.value', fieldDef.timeUnit, legend.format, config.legend.shortTimeLabels, config.timeFormat, isUTCScale)
       }
     }, labelsSpec || {});
+  } else if ((fieldDef.type === NOMINAL || fieldDef.type === ORDINAL) && legend.format) {
```
kanitw: This chunk of code seems identical to axis and very similar to `formatSignalRef`.  

I'd say try to extract a common method for both axis and legend  (name it `guideLabels` text, which return `{signal: ...}`.   And if possible, see if there is a way to reuse `formatSignalRef`.  Note that the latter is trickier since we only set `labels.text` when `format` property of axis/legend does not natively work. So I'm fine if you only do the former first (for now).

src/compile/common.ts
NeelMohapatra: Hmm, what else would be a good idea to add here? Temporal? Nominal doesn't make sense.

kanitw:
Why does ordinal make sense but nominal not?

(Both rarely make sense but when users really want it, they are not that
different.)

src/compile/axis/encode.ts
NeelMohapatra: formatType can only be time, number or utc, right? I'm checking for number above. So that leaves time or utc. timeFormat isn't the default

src/compile/common.ts
domoritz: What happens if we just say 

```
if (specifiedFormat)
  return specifiedFormat
return config.numbverFormat
```

?

NeelMohapatra: That was what I'd done once. But it added `format: 's'` to a lot of generated specs

domoritz: Well, since we removed that a few weeks ago, it may not be an issue anymore.

NeelMohapatra: Point. Yeah, I believe this would work. I can't think of any other issue

domoritz: I can think of one. What if the config overrides the number format? Please try it but otherwise apply this format in all cases.

kanitw:
Rebase?

src/compile/common.ts
```diff
@@ -86,7 +86,29 @@ export function getMarkConfig<P extends keyof MarkConfig>(prop: P, mark: MarkDef
   return config.mark[prop];
 }
 
-export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: string, expr: 'datum' | 'parent', config: Config, useBinRange?: boolean) {
+export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: string, expr: 'datum' | 'parent', config: Config, formatType: FormatType | undefined, useBinRange?: boolean) {
+  // TODO: Refactor
```
kanitw: Can you just refactor rather than leaving it like this?

This seems to generate a very redundant chunk of code ...

src/compile/common.ts
```diff
@@ -177,3 +189,23 @@ export function titleMerger(v1: Explicit<string>, v2: Explicit<string>) {
...
+    const isUTCScale = model.getScaleComponent(channel).get('type') === ScaleType.UTC;
+    return {
+      signal: timeFormatExpression('datum.value', fieldDef.timeUnit, label.format, config.legend.shortTimeLabels, config.timeFormat, isUTCScale)
+    };
+  } else if ((fieldDef.type === NOMINAL || fieldDef.type === ORDINAL) && label.format) {
+    if (label.formatType === 'number') {
+      return {
+        signal: `format(${fieldDef.field}, '${numberFormat(fieldDef, label.format, config)}')`
+      };
+    } else if (label.formatType) {
```
kanitw: if (label.formatType) ??

Why don't you check if this is really for time

src/compile/common.ts
```diff
@@ -177,3 +189,23 @@ export function titleMerger(v1: Explicit<string>, v2: Explicit<string>) {
...
 }
+
+export function guideLabels(fieldDef: FieldDef<string>, config: Config, model: Model, channel: SpatialScaleChannel | NonspatialScaleChannel, label: any) {
+    if (fieldDef.type === TEMPORAL) {
+    const isUTCScale = model.getScaleComponent(channel).get('type') === ScaleType.UTC;
+    return {
+      signal: timeFormatExpression('datum.value', fieldDef.timeUnit, label.format, config.legend.shortTimeLabels, config.timeFormat, isUTCScale)
+    };
+  } else if ((fieldDef.type === NOMINAL || fieldDef.type === ORDINAL) && label.format) {
+    if (label.formatType === 'number') {
```
kanitw: If `formatType` is not specified, it should mean number by default so the logic here is wrong.

src/compile/common.ts
```diff
@@ -177,3 +189,23 @@ export function titleMerger(v1: Explicit<string>, v2: Explicit<string>) {
...
+      return {
+        signal: `format(${fieldDef.field}, '${numberFormat(fieldDef, label.format, config)}')`
+      };
+    } else if (label.formatType) {
+      return {
+        signal: timeFormatExpression('datum.value', fieldDef.timeUnit, label.format, config.legend.shortTimeLabels, config.timeFormat, label.formatType === 'utc')
```
kanitw: guideLabels, but why read `config.legend`?

Just passing in `shortTimeLabels` would be less risky.

src/compile/common.ts
```diff
@@ -177,3 +189,23 @@ export function titleMerger(v1: Explicit<string>, v2: Explicit<string>) {
       v1.value + ', ' + v2.value // join title with comma if different
   };
 }
+
+export function guideLabels(fieldDef: FieldDef<string>, config: Config, model: Model, channel: SpatialScaleChannel | NonspatialScaleChannel, label: any) {
+    if (fieldDef.type === TEMPORAL) {
+    const isUTCScale = model.getScaleComponent(channel).get('type') === ScaleType.UTC;
+    return {
+      signal: timeFormatExpression('datum.value', fieldDef.timeUnit, label.format, config.legend.shortTimeLabels, config.timeFormat, isUTCScale)
```
kanitw: guideLabels, but why read `config.legend`?

Just passing in `shortTimeLabels` would be less risky.

src/compile/layout/header.ts
```diff
@@ -97,7 +97,7 @@ export function getHeaderGroup(model: Model, channel: HeaderChannel, headerType:
       const format = facetFieldDef.header ? facetFieldDef.header.format : undefined;
 
       title = {
-        text: formatSignalRef(facetFieldDef, format, 'parent', model.config, true),
+        text: formatSignalRef(facetFieldDef, format, 'parent', model.config, undefined, true), // undefined passed since formatType not supported in headers
```
kanitw: Why is `formatType` not supported in headers?

src/compile/common.ts
```diff
@@ -95,7 +117,7 @@ export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: str
         return {signal: field(fieldDef, {expr, binSuffix: 'range'})};
       } else {
         return {
-          signal: `${formatExpr(field(fieldDef, {expr, binSuffix: 'start'}), format)} + '-' + ${formatExpr(field(fieldDef, {expr, binSuffix: 'end'}), format)}`
+          signal: `${formatExpr(field(fieldDef, {expr, binSuffix: 'start'}), format)}+'-'+${formatExpr(field(fieldDef, {expr, binSuffix: 'end'}), format)}`
```
kanitw: Why removing the nice space that we had?

domoritz:
We're still waiting for updates here, right @kanitw @NeelMohapatra?

kanitw:
We might not need this, depending on how we implement timeUnit with ordinal type.

kanitw:
The original use case was formatting row/column, which now no longer uses axis / scale so it is not an issue anymore.

PR #2313, opened 2017-04-29
[WIP] Boxplot Orient

mattwchun:
Allow the use of aggregate and or orient to determine box plot orientation.

Refers to issue: #2303

Issue #2303, opened 2017-04-28
Add `orient` to boxplot


kanitw:
Earlier you convinced me that having `orient` in boxplot does not make sense.
However, thinking more about it.   The following spec seems totally sensible:

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A horizontal box plot showing median, min, and max in the US population distribution of age groups in 2000.",
  "data": {"url": "data/population.json"},
  "mark": "box-plot",
  "encoding": {
    "x": {"field": "age","type": "quantitative"},
    "y": {
      "field": "people",
      "type": "quantitative",
      "axis": {"title": "population"}
    },
    "size": {"value": 5}
  }
}
```

The default should be `"vertical"` if there is no aggregate to hint.

It should also take aggregate into account though:

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A horizontal box plot showing median, min, and max in the US population distribution of age groups in 2000.",
  "data": {"url": "data/population.json"},
  "mark": "box-plot",
  "encoding": {
    "y": {"field": "age","type": "quantitative"},
    "x": {
      "aggregate": "box-plot",  // this should make orient be horizontal by default
      "field": "people",
      "type": "quantitative",
      "axis": {"title": "population"}
    },
    "size": {"value": 5}
  }
}
```

PR #2370, opened 2017-05-16
Boxplot Orient

mattwchun:
Allow the use of aggregate and or orient to determine box plot orientation.

Refers to issue: #2303 

This is the updated version of the boxplot pr in a new branch and pr because of weird merging in the #2313 pr.

### DIFF
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -30,15 +30,6 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT 
 
   if (encoding.x && encoding.y) {
     // 2D
-    if (isContinuous(encoding.x) && isContinuous(encoding.y)) {
-      const xEncChannel = encoding.x as FieldDef<Field>;
-      const yEncChannel = encoding.y as FieldDef<Field>;
-      if (xEncChannel.aggregate === BOXPLOT && yEncChannel.aggregate === BOXPLOT) {
-        throw new Error('Both x and y cannot have aggregate');
-      }
-    } else if (isDiscrete(encoding.x) && isDiscrete(encoding.y)) {
-      throw new Error('Both x and y cannot be discrete');
-    }
 
     const orient: Orient = box2DOrient(spec);
     const params = box2DParams(spec, orient);
@@ -154,6 +145,16 @@ export function box2DOrient(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | Box
   const xEncChannel = encoding.x as FieldDef<Field>;
   const yEncChannel = encoding.y as FieldDef<Field>;
   let resultOrient: Orient;
+
+  if (isContinuous(encoding.x) && isContinuous(encoding.y)) {
+    const xEncChannel = encoding.x as FieldDef<Field>;
+    const yEncChannel = encoding.y as FieldDef<Field>;
+    if (xEncChannel.aggregate === BOXPLOT && yEncChannel.aggregate === BOXPLOT) {
+      throw new Error('Both x and y cannot have aggregate');
+    }
+  } else if (isDiscrete(encoding.x) && isDiscrete(encoding.y)) {
+    throw new Error('Both x and y cannot be discrete');
+  }
 
   if (isDiscrete(encoding.x) && isContinuous(encoding.y)) {
     resultOrient = 'vertical';
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -13,6 +13,10 @@ export interface BoxPlotDef {
   orient: Orient;
 }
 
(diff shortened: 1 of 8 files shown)


src/compositemark/boxplot.ts
```diff
@@ -26,23 +30,23 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT>
...
+      const yEncChannel = encoding.y as FieldDef<Field>;
+      if (xEncChannel.aggregate === BOXPLOT && yEncChannel.aggregate === BOXPLOT) {
+        throw new Error('Both x and y cannot have aggregate');
+      }
+    } else if (isDiscrete(encoding.x) && isDiscrete(encoding.y)) {
+      throw new Error('Both x and y cannot be discrete');
     }
+
+    const orient: Orient = box2DOrient(spec);
+    const params = box2DParams(spec, orient);
```
kanitw: Why don't you use 
```
const {discreteAxisFieldDef, continuousAxisChannelDef, discreteAxis, continuousAxis} = box2DParams(spec, orient);
```

like suggested?

src/compositemark/boxplot.ts
```diff
@@ -26,23 +30,23 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT>
...
-    } else if (isDiscrete(encoding.y) && isContinuous(encoding.x)) {
-      // horizontal
-      discreteAxis = 'y';
-      continuousAxis = 'x';
-      continuousAxisChannelDef = encoding.x;
-
-      discreteAxisFieldDef = encoding.y;
-    } else {
-      throw new Error('Need one continuous and one discrete axis for 2D boxplots');
+    if (isContinuous(encoding.x) && isContinuous(encoding.y)) {
```
kanitw: Wondering why aren't these error won't be moved to the new box2DOrient method?

src/compositemark/boxplot.ts
```diff
@@ -155,6 +146,16 @@ export function box2DOrient(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | Box
   const yEncChannel = encoding.y as FieldDef<Field>;
   let resultOrient: Orient;
 
+  if (isContinuous(encoding.x) && isContinuous(encoding.y)) {
+    const xEncChannel = encoding.x as FieldDef<Field>;
+    const yEncChannel = encoding.y as FieldDef<Field>;
+    if (xEncChannel.aggregate === BOXPLOT && yEncChannel.aggregate === BOXPLOT) {
+      throw new Error('Both x and y cannot have aggregate');
```
kanitw: Why are these separated up here?  These two cases should replace the return undefined case down there.

src/compositemark/boxplot.ts
```diff
@@ -145,3 +140,81 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT>
...
+    discreteAxis = 'y';
+    continuousAxis = 'x';
+    continuousAxisChannelDef = encoding.x;
+    discreteAxisFieldDef = encoding.y;
+  }
+
+  if (continuousAxisChannelDef && continuousAxisChannelDef.aggregate) {
+    const {aggregate: aggregate, ...continuousAxisWithoutAggregate} = continuousAxisChannelDef;
+    if (aggregate !== BOXPLOT) {
+      throw new Error('Continuous axis should not be aggregated');
```
kanitw: `Continuous axis should not have customized aggregation function ${aggregate}`

src/compositemark/boxplot.ts
mattwchun: I will need to add redundant logic into box2DOrient that's already done in box2DParams to find the continuous axis if I want to throw the error in box2DOrient

src/compositemark/boxplot.ts
```diff
@@ -145,3 +140,81 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT>
...
+  } else {
+    if (isContinuous(encoding.x) && isContinuous(encoding.y)) {
+      if (xEncChannel.aggregate === undefined && yEncChannel.aggregate === BOXPLOT) {
+        resultOrient = 'vertical';
+      } else if (yEncChannel.aggregate === undefined && xEncChannel.aggregate === BOXPLOT) {
+        resultOrient = 'horizontal';
+      } else if (xEncChannel.aggregate === BOXPLOT && yEncChannel.aggregate === BOXPLOT) {
+        return undefined; // invalid spec
+      } else {
+        const markChannel = mark as BoxPlotDef;
```
kanitw: Why casting rather than checking type guard?

kanitw: Use type guard so you don't have to declare a new name (`markChannel` is a weird name anyway.)

PR #2375, opened 2017-05-16
Implement lookup transform. Fixes #2298

### LINKED ISSUE #2298: Lookup transform
```js
{
    "lookup": "key in source",
    "from": {"data": Data, key: "key in from", "fields" : ["a", "b"]},
    "as": ["foo", "bar"]
}
```

* `as` is optional
* If `fields` is not specified, `as` has to be a string and we put the whole object into the data. To enforce this, use a union type (`|` in TypeScript)

The trickiest bit is figuring out how to create the right dataflow. You need to add a separate source node. Unfortunately, we have to break the assumption that every dataflow node has a single parent for this operator. To make it less painful, I think we can just use a reference to the source in the lookup node (not as a parent) and add lookup as a child of the source it references. I think the rest should just work (TM).

domoritz:


kanitw:
@willium Since this is relevant to your stuff, you should be the primary reviewer.

That said, this seems like a very critical change -- don't merge until I approve this.  

@domoritz  -- one high level comment -- is there a way to make the data source tree generation and naming more stable?  Given that here you try to add lookup, but the output diff shows a lot of re-reordering of data source names, making it hard to reason about the main changes in this PR.

domoritz:
@willium if you have time, could you add a commit with tests for the lookup?
@kanitw, yep, looking into it.

domoritz:
@kanitw I fixed it. Now we only move sources without transforms to the beginning.

domoritz:
I added some tests. Should be good to go now. @willium @kanitw

src/compile/data/assemble.ts
```diff
@@ -326,5 +330,17 @@ export function assembleData(roots: SourceNode[]): VgData[] {
     }
   });
 
+  // move sources without transforms (the onces that are potentially need for lookups) to the beginning
```
kanitw: "the ones"

test/compile/data/transforms.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        from: 'lookup_0',
+        key: 'name',
+        fields: ['person'],
+        as: ['foo']
+      });
+    });
+
+    it('should warn if fields are not specified and as is missing', function () {
+      log.runLocalLogger((localLogger) => {
+        (new LookupNode({
```
kanitw: (Minor) This seems a bit ugly -- declare const ?

examples/vg-specs/lookup.vg.json
```diff
@@ -0,0 +1,191 @@
+{
+    "$schema": "http://vega.github.io/schema/vega/v3.0.json",
+    "autosize": "pad",
+    "padding": 5,
+    "data": [
```
kanitw: I don't see any "lookup_people.csv" and "lookup_0"  here.

domoritz: Good catch. I broke this when refactored the code.

src/compile/data/assemble.ts
domoritz: Fixed

test/compile/data/transforms.test.ts
domoritz: Fixed

domoritz:
Pushed the fixes. Ready for merge.

willium:
probably fine to merge, but haven't had a chance to really test it. Your tests look sufficient though!

domoritz:
@kanitw please merge

kanitw:
@domoritz  I look at the change in the output.  It seems reasonable now.  
I don't know what else you change though -- I assume it's minimal just to fix this part. 
If so, LGTM Feel free to rebase and merge.

PR #2372, opened 2017-05-16
Config strip

kanitw:
Fix https://github.com/vega/vega-lite/issues/1995 and https://github.com/vega/vega-embed/issues/27

This requires follow-up changes in vega-embed and vega-editor.

PR #2376, opened 2017-05-16
Miscellaneous selection cleanups

arvind:
Four commits to address four minor issues with selections.

### DIFF
--- a/src/compile/selection/interval.ts
+++ b/src/compile/selection/interval.ts
@@ -114,6 +114,9 @@ const interval:SelectionCompiler = {
       });
     }
 
+    // Two brush marks ensure that fill colors and other aesthetic choices do
+    // not interefere with the core marks, but that the brushed region can still
+    // be interacted with (e.g., dragging it around).
     return [{
       name: undefined,
       type: 'rect',
--- a/src/compile/selection/interval.ts
+++ b/src/compile/selection/interval.ts
@@ -118,14 +118,22 @@ const interval:SelectionCompiler = {
       name: undefined,
       type: 'rect',
       encode: {
-        enter: {fill: {value: '#eee'}},
+        enter: {
+          fill: {value: '#333'},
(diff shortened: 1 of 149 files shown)


src/compile/selection/interval.ts
```diff
@@ -118,14 +118,22 @@ const interval:SelectionCompiler = {
       name: undefined,
       type: 'rect',
       encode: {
-        enter: {fill: {value: '#eee'}},
+        enter: {
```
kanitw: I guess we should make brush style configurable (in a later PR) ?  

We can just output some role name and use the role config. -- Given that we will output config in the Vega output soon in https://github.com/vega/vega-lite/pull/2372 -- This should be easily to implement.  

File an issue?

src/compile/selection/interval.ts
```diff
@@ -118,14 +118,22 @@ const interval:SelectionCompiler = {
...
+          stroke: undefined
+        },
         update: update
       }
     }].concat(marks, {
       name: name + BRUSH,
       type: 'rect',
       encode: {
-        enter: {fill: {value: 'transparent'}},
+        enter: {
```
kanitw: This maybe a dumb question, but I wonder why we need two separate rects here for fill and stroke. (Maybe add a comment?)

src/compile/selection/interval.ts
kanitw: For the implementation, after we merge #2372, I think we can make these brush have some role names and add role configs to the `DEFAULT_CONFIG` in `config.ts` for these roles  and it should work.

arvind: Done #2378.

src/compile/selection/interval.ts
arvind: To ensure that the fill color of the brush doesn't interfere with the marks/grid lines/etc. but still allow users to interact with the brush. Will add a comment.

Issue #2315, opened 2017-04-29
Inversion of ordinal brush looks funky


domoritz:
![brush](https://cloud.githubusercontent.com/assets/589034/25553429/a052b3aa-2c67-11e7-8c6b-5b306c17a0d7.gif)

Spec:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {"url": "data/europe_world_bank.csv"},
  "hconcat": [
    {
      "mark": "bar",
      "height": 700,
       "selection": {
        "brush": {"type": "interval", "encodings": ["y"]}
      },
      "encoding": {
        "y": {
          "field": "Country",
          "type": "ordinal",
          "sort": {
            "op": "min",
            "field": "Population_ages_15_64_of_total",
            "order": "descending"
          }
        },
        "x": {
          "field": "Population_ages_15_64_of_total",
          "type": "quantitative"
        },
        "color": {
          "condition": {"selection": "!brush", "value": "steelblue"},
          "value": "goldenrod"
        }
      }
    },
    {
      "mark": "bar",
      "height": 700,
      "selection": {
        "brush": {"type": "interval", "encodings": ["y"]}
      },
      "encoding": {
        "y": {
          "field": "Country",
          "type": "ordinal",
          "sort": {
            "op": "min",
            "field": "Population_ages_65_and_above_of_total",
            "order": "descending"
          }
        },
        "x": {
          "field": "Population_ages_65_and_above_of_total",
          "type": "quantitative"
        },
        "color": {
          "condition": {"selection": "!brush", "value": "steelblue"},
          "value": "goldenrod"
        }
      }
    },
    {
      "mark": "point",
      "width": 260,
      "height": 260,
      "selection": {"brush": {"type": "interval"}},
      "encoding": {
        "y": {
          "field": "Population_ages_15_64_of_total",
          "type": "quantitative",
          "scale": {"zero": false}
        },
        "x": {
          "field": "Population_ages_65_and_above_of_total",
          "type": "quantitative",
          "scale": {"zero": false}
        },
        "color": {
          "condition": {"selection": "!brush", "value": "steelblue"},
          "value": "goldenrod"
        }
      }
    }
  ]
}
```

Data: https://gist.github.com/domoritz/c52e1ff798f325533d04f1c6d7b8fd83

arvind:
Duplicate of #2148.

domoritz:
Hmm, not sure whether this is a duplicate. 

[...]

Issue #2288, opened 2017-04-26
Brush interactions exhibit buggy behavior


jheer:
- https://vega.github.io/new-editor/?mode=vega-lite&spec=interactive_splom
Translating the brush (with shift-drag) can result in distortion / non-uniform sizing of the visual brush extents, for example after panning/zooming a display. I expect brush extent to stay constant when translating the interval selection.

- https://vega.github.io/new-editor/?mode=vega-lite&spec=overview_detail
Dragging out a brush works fine, but when I attempt to translate or scale the brush it disappears entirely. Possibly due to an issue with handling the temporal data type?

kanitw:
Fixed in #2377

Issue #2284, opened 2017-04-26
Interval brush over binned domain is incorrect


jheer:
As noted in this example: https://vega.github.io/new-editor/?mode=vega-lite&spec=layered_crossfilter

The interval brush snaps to the bins, but the actual filtering range appears to be set in a continuous fashion. I would expect both a continuous brush and filter range by default. On a related note, for ordinal domains (such as band scales), I would expect the brush itself to interact in a continuous fashion, even though the resulting selections are of a discrete set of elements.

For discrete selections, perhaps a multi selection is sufficient? In cases where data is of a known order, it might support shift-click selections that do not merely toggle, but select discrete intervals much like a file system browser. And/or, perhaps there can be an option to indicate a discrete interval selection interaction behavior?

kanitw:
> The interval brush snaps to the bins, but the actual filtering range appears to be set in a continuous fashion. I would expect both a continuous brush and filter range by default. On a related note, for ordinal domains (such as band scales), I would expect the brush itself to interact in a continuous fashion, even though the resulting selections are of a discrete set of elements.


I have just tested -- the interval brush in #2377 seems correct now (See the first screenshot in #2377).  

I guess the discrete selection part is not addressed, so I will leave this open.  @arvind feels free to close it if I'm wrong.

arvind:
I've edited this issue to better reflect what remains: a feature request for alternate interactions for specifying discrete intervals (e.g., shift-click to select a range a la a file browser). I've cleared out the milestone because I think such interactions are best considered as part of a future minor release.

kanitw:
Sounds good.  I'm moving this to  2.x.x Features & Patches so we can preserve "issues with no milestone" for ones that are not categorized.

kanitw:
Besides binning, intervals make come from ranged marks (see https://github.com/vega/vega-lite/issues/3678).

Issue #2148, opened 2017-04-14
Interval selection fails over ordinal scale


jheer:
Expected behavior: A continuous brush that selects the discrete values it contains
Actual behavior: Error messages in console

(Note: `single`/`multi` selections projected over "y" are working.)

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A bar chart showing the US population distribution of age groups in 2000.",
  "data": {
    "url": "data/population.json"
  },
  "transform": [
    {
      "filter": "datum.year == 2000"
    }
  ],
  "mark": "bar",
  "selection": {
    "brush": {
      "type": "interval",
      "encodings": [
        "y"
      ]
    }
  },
  "encoding": {
    "y": {
      "field": "age",
      "type": "ordinal",
      "scale": { "rangeStep": 17 }
    },
    "x": {
      "aggregate": "sum",
      "field": "people",
      "type": "quantitative",
      "axis": { "title": "population" }
    },
    "color": { "value": "steelblue" }
  }
}
```

arvind:
The error seems to be in inverting coordinates with the `point` y-scale generated. 

```json
{
  "name": "y",
  "type": "point",
  "domain": {
    "data": "source_0",
    "field": "age",
    "sort": true
  },
  "range": {
    "step": 17
  },
  "round": true,
  "padding": 0.5
}
```

If I switch it to the following `band` scale, the interval selection works again. @jheer, would you like me to file an issue in vega-scale?

```json
{
  "name": "y",
  "type": "band",
  "domain": {
    "data": "source_0",
    "field": "age",
    "sort": true
  },
  "range": [0, 400]
}
```

jheer:
I don't think so. AFAICT the scale inversions are performing as designed. The console errors do go away when I switch the generated Vega to a band scale, but the brush still does not behave correctly (or, at minimum, as I expect it should).

I believe the Vega code as generated will _never_ work with a point scale. (1) On mousedown, the brush interval is initialized through `invert` by passing the mouse position as the brush extent (e.g., `[x, x]`). For a point scale, that will typically resolve to nothing (i.e., `undefined`), as the scale bandwidth is zero and so a single coordinate will only resolve to a range if it _exactly_ matches a value. (2) On mousemove, only a single coordinate (not a range) is run through `invert`. Again, for a point scale this will fail unless an _exact_ match occurs.

[...]

PR #2377, opened 2017-05-17
Break apart pixel and data signals for interval selections

arvind:
~~Please merge after #2376.~~  Done

This PR should address the issues with interval selections identified in ~~#2148~~, ~~#2284~~, ~~#2288~~, and ~~#2315~~ (once vega/vega-parser#32 is merged). Interval selections now generate separate signals in pixel and data space. 

crossfiltering

overviewdetail

One todo remains before this PR can be reviewed: unbound intervals (i.e., brushes) do not yet update when bound intervals (i.e., panning & zooming) are modified. This results in the behaviour shown here whereby points do not recolour when panning/zooming the grid: 

interactivesplom

### DIFF
--- a/src/compile/selection/interval.ts
+++ b/src/compile/selection/interval.ts
@@ -1,5 +1,6 @@
 import {Channel, X, Y} from '../../channel';
 import {warn} from '../../log';
+import {hasContinuousDomain, isBinScale} from '../../scale';
 import {extend, keys, stringValue} from '../../util';
 import {UnitModel} from '../unit';
 import {channelSignalName, ProjectComponent, SelectionCompiler, SelectionComponent, STORE, TUPLE} from './selection';
@@ -16,10 +17,12 @@ const interval:SelectionCompiler = {
         name = selCmpt.name;
 
     if (selCmpt.translate && !(scales.has(selCmpt))) {
+      const filterExpr = `!event.item || event.item.mark.name !== ${stringValue(name + BRUSH)}`;
       events(selCmpt, function(_: any[], evt: any) {
         const filters = evt.between[0].filter || (evt.between[0].filter = []);
-        filters.push('!event.item || (event.item && ' +
-          `event.item.mark.name !== ${stringValue(name + BRUSH)})`);
+        if (filters.indexOf(filterExpr) < 0) {
+          filters.push(filterExpr);
+        }
       });
     }
 
@@ -29,18 +32,14 @@ const interval:SelectionCompiler = {
         return;
       }
 
-      const cs = channelSignals(model, selCmpt, p.encoding);
+      const cs = channelSignals(model, selCmpt, p.encoding),
+        csName = channelSignalName(selCmpt, p.encoding, 'data');
       signals.push.apply(signals, cs);
       intervals.push(`{encoding: ${stringValue(p.encoding)}, ` +
-      `field: ${stringValue(p.field)}, extent: ${cs[1].name}}`);
+      `field: ${stringValue(p.field)}, extent: ${csName}}`);
     });
 
-    signals.push({
-      name: name,
-      update: `[${intervals.join(', ')}]`
-    });
-
-    return signals;
+    return signals.concat({name: name, update: `[${intervals.join(', ')}]`});
   },
 
   tupleExpr: function(model, selCmpt) {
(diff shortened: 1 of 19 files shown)


domoritz:
About the last gif, why doesn't the brush move with the zooming and panning?

jheer:
It seems there are two options here:
1. Upon pan/zoom, the brush selection should not change relative to the data space. It should pan/zoom in pixel space along with the chart. The actual selection query would be unchanged.
2. The brush stays put in pixel space but the selection query changes as the chart pans/zooms.

Option 1 seems much more natural to me, as it keeps the selection state stable. There is the potential problem of the brush disappearing off-screen, but that seems reasonable enough -- especially if in the most cases clicking and dragging with reset the brush state, removing the off-screen brush.

Do we agree on that point? If so, my understanding is that the main issue here is how to correctly compute the new pixel end points of the brush after the scale has updated. Correct?

domoritz:
Option 1 also seems more natural to me. @arvind Can you briefly explain how separate signals in data and pixel space address the issues you mentioned?

domoritz:
What's the status of this PR? It's marked WIP.

arvind:
After trying some fancier strategies involving calculating normalized pan/zoom coords, a simpler approach works more reliably. For continuous, unbinned scales we simply rescale brushes in response to scale changes. All other scale types cannot be panned or zoomed, and it can be difficult to predict the effect of other scale domain changes (e.g., filtering). So, for such scale types, the brush is simply cleared.

workingsplom

This approach requires `vega-parser@1.0.0-beta.55` and the resultant changes to the scheme. Thus, Travis will fail until a new Vega 3 beta is issued.

src/compile/selection/interval.ts
```diff
@@ -84,21 +64,10 @@ const interval:SelectionCompiler = {
...
-        {field: {group: 'width'}}),
-
-      y: extend({}, yi !== null ?
-        {scale: model.scaleName(Y), signal: `${name}[${yi}].extent[0]`} :
-        {value: 0}),
-
-      y2: extend({}, yi !== null ?
-        {scale: model.scaleName(Y), signal: `${name}[${yi}].extent[1]`} :
-        {field: {group: 'height'}})
+      x: extend({}, xi !== null ? {signal: `${name}_x[0]`} : {value: 0}),
```
kanitw: Why do we need `extend` for these lines? 

It seems like `xi !== null ? {signal: `${name}_x[0]`} : {value: 0}` would be sufficient for the first line and same for others?

src/compile/selection/interval.ts
```diff
@@ -156,29 +125,40 @@ export function projections(selCmpt: SelectionComponent) {
...
-        events: evt.between[0],
-        update: invert(`[${coord}, ${coord}]`)
-      });
+      coord = `${channel}(unit)`;
 
-      on.push({
-        events: evt,
-        update: `[${name}[0], ` + invert(`clamp(${coord}, 0, ${size})`) + ']'
-      });
+  const on = events(selCmpt, function(def: any[], evt: any) {
```
kanitw: Why do `evt` here and `events` in `SelectionComponent` have the type `any`?

src/compile/selection/transforms/translate.ts
```diff
@@ -32,14 +31,12 @@ const translate:TransformCompiler = {
       value: {},
       on: [{
         events: events.map((e) => e.between[0]),
-        update: '{x: x(unit), y: y(unit), ' +
-          `width: ${size}.width, height: ${size}.height, ` +
+        update: '{x: x(unit), y: y(unit)' +
+          (x !== null ? ', extent_x: ' + (hasScales ? domain(model, X) :
```
kanitw: can you explain why width and height are now removed?

src/compile/selection/transforms/translate.ts
```diff
@@ -65,30 +62,29 @@ const translate:TransformCompiler = {
...
+    return channel === Y ? '+' : '-';
   }
-  return s;
+  return '+';
 }
 
 function onDelta(model: UnitModel, selCmpt: SelectionComponent, channel: Channel, size: 'width' | 'height', signals: any[]) {
   const name = selCmpt.name,
-      signal:any = signals.filter((s:any) => s.name === channelSignalName(selCmpt, channel))[0],
+      hasScales = scalesCompiler.has(selCmpt),
```
kanitw: Don't worry about changing, but just in case you write new code after this - we normally declare one `const` per variable like

```
const a=1;
const b=2;
```

so that the diff for removing one variable only involve one line.

src/compile/selection/interval.ts
```diff
@@ -156,29 +145,36 @@ export function projections(selCmpt: SelectionComponent) {
   return {x, xi, y, yi};
 }
 
-function channelSignal(model: UnitModel, selCmpt: SelectionComponent, channel: Channel): any {
-  const name  = channelSignalName(selCmpt, channel),
+function channelSignals(model: UnitModel, selCmpt: SelectionComponent, channel: Channel): any {
```
kanitw: /**
 * Return visual and data signals for interval selection. 
 */

src/compile/selection/transforms/translate.ts
```diff
@@ -64,31 +64,17 @@ const translate:TransformCompiler = {
...
-      lo = `invert(${scale}` + (channel === X ? ', 0' : `, unit.${size}`) + ')',
-      hi = `invert(${scale}` + (channel === X ? `, unit.${size}` : ', 0') + ')';
+      extent = `${anchor}.extent_${channel}`,
+      range = `[${extent}[0] + ${delta}.${channel}, ` +
+        `${extent}[1] + ${delta}.${channel}]`;
 
   signal.on.push({
     events: {signal: delta},
-    update: scalesCompiler.has(selCmpt) ? range : `clampRange(${range}, ${lo}, ${hi})`
+    update: scalesCompiler.has(selCmpt) ? range : `clampRange(${range}, 0, unit.${size})`
```
kanitw: This might be a dumb question but I'm curious why we no longer have to swap `0` and `unit.${size}` ?

src/compile/selection/transforms/zoom.ts
```diff
@@ -49,17 +54,6 @@ const zoom:TransformCompiler = {
       onDelta(model, selCmpt, 'y', 'height', signals);
     }
 
-    const size = signals.filter((s:any) => s.name === name + INTERVAL_SIZE);
```
kanitw: Why do we no longer need this part?

kanitw:
I was playing with all the changed examples and they look reasonable.

@domoritz Feel free to double check just in case I overlook anything.

src/compile/selection/interval.ts
arvind: Because they are [event stream objects](https://vega.github.io/vega/docs/event-streams/#object).

src/compile/selection/transforms/translate.ts
arvind: We no longer need to determine them from the `unit` signal because I now call `model.getSizeSignalRef`.

src/compile/selection/transforms/zoom.ts
arvind: The signal has been removed as part of breaking visual/pixel and data signals apart.

arvind:
Thanks @kanitw. Requested changes made!

src/compile/selection/interval.ts
domoritz: Can you add a type? Even if just `type VgEventStream = any`. Then this question will never come up again.

src/compile/selection/transforms/translate.ts
arvind: Sorry about that. Will follow your style henceforth.

domoritz: Yes, strong +1 for separating declarations.

src/compile/selection/interval.ts
arvind: Done!

domoritz:
Looks great. I just noticed some incredible lag in the editor. Do you see the same?

editor

screen shot 2017-06-05 at 22 03 17

src/compile/selection/interval.ts
```diff
@@ -156,33 +126,47 @@ export function projections(selCmpt: SelectionComponent) {
   return {x, xi, y, yi};
 }
 
-function channelSignal(model: UnitModel, selCmpt: SelectionComponent, channel: Channel): any {
-  const name  = channelSignalName(selCmpt, channel),
+/**
+ * Returns the visual and data signals for an interval selection.
+ */
+function channelSignals(model: UnitModel, selCmpt: SelectionComponent, channel: Channel): any {
```
domoritz: instead of `any`, return `VgSignal`

arvind:
I see some lag with the Canvas renderer but not with the SVG. Things should also be much smoother now that #2452 has been merged, which removes all the redundant scale updates.

src/compile/selection/interval.ts
arvind: The `VgSignal` type isn't fully descriptive yet. So rather than deal with compiler errors, I figured `any` was good enough for now. `any` is used in place of `VgSignal` throughout this part of the codebase, I think, so I would prefer to handle that separately from this PR (and perhaps as a lower-priority item for release) if you're amenable.

src/compile/selection/transforms/translate.ts
kanitw: @arvind  Don't worry it's very minor.  :)

PR #2380, opened 2017-05-17
Use sequential color scale by default for ordinal color

### LINKED ISSUE #2373: Custom color range error for ordinal scales
From the second example in the documentation of [custom color range](https://vega.github.io/vega-lite/docs/scale.html#example-custom-color-range):

> For ordinal, quantitative, and time fields, range can be a two-element array describing the two colors for interpolation.

However the example just cycles through the two colors.

kanitw:
Fix #2373

### DIFF
--- a/examples/vg-specs/brush.vg.json
+++ b/examples/vg-specs/brush.vg.json
@@ -489,13 +489,14 @@
         },
         {
             "name": "color",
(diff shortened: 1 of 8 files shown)


domoritz:
Test fail

kanitw:
@domoritz pass now

PR #2384, opened 2017-05-18
Small refactor from forEach to for of

domoritz:


NeelMohapatra:
@domoritz Just out of curiosity, why did you do this?

domoritz:
It's cleaner. forEach is not a language construct.

PR #2398, opened 2017-05-22
Fix missing output of raw data source. Fixes #2390

### LINKED ISSUE #2390: Sorting no longer draw from raw data
```
{
  "mark": "bar",
  "encoding": {
    "x": {"field": "Cylinders","type": "nominal", "sort": {"op": "count", "field": "*", "order": "descending"}},
    "y": {"aggregate": "count","type": "quantitative"}
  }
}
```

vega_editor

domoritz:


src/compile/scale/domain.ts
```diff
@@ -134,8 +134,7 @@ function parseSingleChannelDomain(scale: Scale, model: UnitModel, channel:Channe
         };
       }
     }
-  } else if (sort) { // have sort -- only for ordinal
-
+  } else if (sort) { // have sort -- only for ordinal and nominal
```
kanitw: Only for discrete scale (ordinal/point/band)

domoritz:
> You said you are going to add a test to prevent regression.

We have a sample spec. Writing a test for this one case requires new test infra, which is beyond the scope of this pr.

src/compile/scale/domain.ts
domoritz: I deleted the comment to avoid confusion.

kanitw:
Please file an issue and merge this

domoritz:
Follow up issue about testing: https://github.com/vega/vega-lite/issues/2399

Issue #2402, opened 2017-05-22
groupby in SummarizeTransform is not camelCase (groupBy)


mattwchun:

domoritz:
It isn't in Vega either: https://vega.github.io/vega/docs/transforms/aggregate/. However, `timeUnit` is camel case in vega-lite. But `maxbin` is not...

We should be consistent and break it now rather than after 2.0.

jheer:
For the record, and with little rational reason, I'm not a fan of camel case. I'm even less a fan of breaking things others have become accustomed to.

kanitw:
> For the record, and with little rational reason, I'm not a fan of camel case. I'm even less a fan of breaking things others have become accustomed to.

Yeah, but `timeUnit` and many scale, axis, legend, and config properties all use camel case.
For the new properties (e.g., `groupBy` in the new `SummarizeTransform` for `transform` array),
 we definitely should be consistent.

(I'm not asking to breaking these in Vega as the number of users who use `groupby` in Vega would 
be way larger.) 

For `maxbins` in Vega-Lite, I'm also slightly leaning toward breaking it now for consistency too -- although I agree it's a bit annoying to introduce breaking changes.  (If we gonna break this in Vega-Lite, now is the time -- Another option is to break it in the schema but offers backward compatibility in the compiler.)

domoritz:
I vote for breaking it now and consistently use **no** camel case. Do we have other examples besides `timeUnit`?

kanitw:
> many scale, axis, legend, and config properties

There are too many things to break.
I would say keep the camelCase.

Funny enough, most of these properties have a camel case equivalent in Vega.

domoritz:
> Note that most of these properties also have camel case in Vega.

Oh okay. I just randomly looked around and saw a lot of things without camel case.

kanitw:
> I just randomly looked around and saw a lot of things without camel case.

What are the things without camel case besides (`maxbins`, `zindex`, and `groupby`)?

kanitw:
`vconcat`, `hconcat`.

kanitw:
(`vconcat` and `hconcat` are new things though.

domoritz:
> What are the things without camel case
`modeskew`, `minstep` maybe

kanitw:
Yeah, `minstep` is a `bin` parameter like `maxbins`.  

However, `modeskew` is value of a property (`aggregate`), not a key so it does not count.  

That said, for many property values, we actually use dash.
[...]

PR #2404, opened 2017-05-23
hconcat -> hConcat and vconcat -> vConcat

domoritz:
Ref #2402

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -608,7 +608,7 @@
                     },
                     "type": "array"
                 },
-                "vconcat": {
+                "vConcat": {
                     "items": {
                         "anyOf": [
                             {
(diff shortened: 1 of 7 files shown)


domoritz:
@kanitw Do you know why this is not passing? It works on my machine and I have no idea what I'm missing.

kanitw:
IDK -- but using this logic here: https://github.com/vega/vega-lite/issues/2402#issuecomment-303457414 we shouldn't make it camelcase anyway.

domoritz:
@kanitw Did you see https://github.com/vega/vega-lite/pull/2404/commits/6e70f955eae53f7b929d5e558be806222c4fc25a?

kanitw:
Nope.  Seeing it now I wonder why we need to use ES2016 syntax for a node script anyway.

domoritz:
Well, the point was that we used require to load JSON and had an unused import. I cherry picked it.

PR #2406, opened 2017-05-23
groupby => groupBy

kanitw:
For groupBy case, I strongly think that this should have camel case for consistency with *most* of the rest of the language.  

(If I had notice that it's not camel case in the original PR, I wouldn't have merged it.)


jheer:
I do not support this change. Please see my comments here: https://github.com/vega/vega-lite/issues/2402#issuecomment-303443272

kanitw:
Ok.  Responses here: https://github.com/vega/vega-lite/issues/2402#issuecomment-303457414

PR #2407, opened 2017-05-24
Automatically resize charts on updates.

domoritz:
cc @yifanwu

### DIFF
--- a/src/compile/compile.ts
+++ b/src/compile/compile.ts
@@ -59,7 +59,10 @@ function assemble(model: Model, topLevelProperties: TopLevelProperties) {
   const output = {
     $schema: 'http://vega.github.io/schema/vega/v3.0.json',
     ...(model.description ? {description: model.description} : {}),
-    autosize: 'pad', // By using Vega layout, we don't support custom autosize
+    autosize: {
+      type: 'pad', // By using Vega layout, we don't support custom autosize
+      resize: true
+    },
     ...topLevelProperties,
     data: [].concat(
       model.assembleSelectionData([]),
--- a/src/compile/compile.ts
+++ b/src/compile/compile.ts
@@ -56,14 +56,15 @@ function assemble(model: Model, topLevelProperties: TopLevelProperties) {
   // Config with Vega-Lite only config removed.
   const vgConfig = model.config ? stripConfig(model.config) : undefined;
 
+  // autoResize has to be put under autosize
+  const {autoResize, ...topLevelProps} = topLevelProperties;
+
   const output = {
     $schema: 'http://vega.github.io/schema/vega/v3.0.json',
     ...(model.description ? {description: model.description} : {}),
-    autosize: {
-      type: 'pad', // By using Vega layout, we don't support custom autosize
-      resize: true
-    },
-    ...topLevelProperties,
+    // By using Vega layout, we don't support custom autosize
+    autosize: topLevelProperties.autoResize ? {type: 'pad', resize: true} : 'pad',
+    ...topLevelProps,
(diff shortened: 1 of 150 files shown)


jheer:
A word of warning (just to be safe): setting `resize: true` can cause layout to re-run on *every view update*, not just streaming data updates. That means if an interactive selection changes the axis labels and that changes the chart size, everything might "jump" between frames. That is one reason why resize is not true by default (by default a layout "resize" is only run when the chart width and/or height signals are modified). The other reason is for performance considerations, as "fit" and "pad" auto-sizing require running *two* pulse propagation cycles. 

If instead you'd like a way to trigger re-layout from the View API, that would not be hard to add. Just file a feature request issue on the vega-view repo.

kanitw:
@domoritz Sounds like we shouldn't merge this then.

domoritz:
The problem is that the legend in the streaming example is cut off. At least, we need an option to enable automatic resizing.

kanitw:
>  in the streaming example
which one?

kanitw:
Also, for streaming use case, seems like re-layout trigger API that Jeff suggests is more appropriate?

>  an option to enable automatic resizing.
sounds more reasonable than enable by default.

domoritz:
The streaming example is forthcoming. I filed an issue at https://github.com/vega/vega-view/issues/6. 

I still think we need an option to enable automatic resizing. I will add a top level option `resize`. 

Another option I thought about last night is to automatically detect whether a spec has named datasources. However, this is again only a heuristic.

domoritz:
@jheer Can you confirm that Vega-Lite has to use `autosize: "pad"` because we use `layout`? I'm asking because I want to add a top level property `autosize` if we do support other types and only `resize` if not.

jheer:
Yes, unless you somehow track all your space usage manually and incorporate that into width/height/padding signal calculations, you'll definitely need to use "pad".

domoritz:
@kanitw I pushed another commit with a new top level option `autoResize`.

src/compile/compile.ts
```diff
@@ -56,11 +56,18 @@ function assemble(model: Model, topLevelProperties: TopLevelProperties) {
...
+  // remove autoResize as it has to be put under autosize
+  const topLevelProps = extend({}, topLevelProperties);
+  delete topLevelProps.autoResize;
+
   const output = {
     $schema: 'http://vega.github.io/schema/vega/v3.0.json',
     ...(model.description ? {description: model.description} : {}),
-    autosize: 'pad', // By using Vega layout, we don't support custom autosize
-    ...topLevelProperties,
+    autosize: {
```
kanitw: Wonder if you think
```
autosize: topLevelProperties.autoResize ? {type: 'pad', resize: true} : 'pad'
```
is a bit nicer, and won't require changing all existing specs.

domoritz: Done

domoritz:
Done.

src/compile/compile.ts
```diff
@@ -56,11 +56,16 @@ function assemble(model: Model, topLevelProperties: TopLevelProperties) {
   // Config with Vega-Lite only config removed.
   const vgConfig = model.config ? stripConfig(model.config) : undefined;
 
+  // remove autoResize as it has to be put under autosize
+  const topLevelProps = extend({}, topLevelProperties);
```
kanitw: `const {_autoResize, ...topLevelProps} = topLevelProperties;`  ?

domoritz: does the _ work?

domoritz:
Thanks!

Issue #2176, opened 2017-04-16
Make facet support sort


kanitw:
We will need to add additional nodes to perform collect transform.

domoritz:
- [ ] Make sure that we sort cells correctly. #2412

kanitw:
Sorting when there is only `row` or `column` should be straightforward.  
However, sorting both  `row` and `column` can be tricky as we want to sort by the global value of each row and each column, not each cell.  

Thus, group mark's `sort` is not sufficient if `sort` is specified for both row and column.  

cc: @jheer

jheer:
Before the new layout features, I believe this would be handled by sorting the scale domain. In that case, sorting works by generating an aggregate transform under the hood and sorting by that. An analogous approach could be taken here. Either (1) one could create an explicit aggregate and sort by the results using the group mark `sort` (possibly requiring a lookup along the way), or (2) one could generate a scale (with the same parameterization as before) whose sole purpose is to generate range values (e.g., over `[0,1]`) that one can the sort by. I think option (1) is cleaner, less convoluted, and possibly more performant.

kanitw:
> Before the new layout features, I believe this would be handled by sorting scale domains.

Yep. 

>  Either (1) one could create an explicit aggregate and sort by the results using the group mark sort (possibly requiring a lookup along the way)

I was thinking that we need to do this too, but that will add some more complexity to the dataflow as we will have to fork two datasets to calculate aggregate for row and for column and then "merge" both of them using lookup (twice).  Thus, I wanna run it by you first if this is the only way. 
(This is quite complicated, so I also wonder if we better support this post 2.0)
  
Meanwhile, I was thinking for a while whether Vega should support a transform that adds a new column with group summary values (like in [dplyr](https://stackoverflow.com/a/34157744/866989)).  This would be equivalent to for a data source, aggregate and join it back, but can be done within one transform.   This could be useful in many cases.  For example, calculating group residual (value - group mean) would require such operation.  Obviously, this transform would make sorting operation here way simpler too.  

In any case, I think we would start by supporting sorting when there is only `row`  or only `column` first as doing so do not require forking new data sources and lookup.  And we can see if we have enough time to support sorting when there are both row and column.

jheer:
Sure, I could imagine adding an operator the performs an aggregate and automatically joins the results back onto the input stream. That could be useful for a number of calculations. The internal implementation could reuse the existing `aggregate` and I assume the actual parameters would be more or less identical.

[...]

PR #2412, opened 2017-05-24
Fix facet with multiple dims

domoritz:
screen shot 2017-05-23 at 20 07 55

### DIFF
--- a/examples/specs/normalized/trellis_row_column_normalized.vl.json
+++ b/examples/specs/normalized/trellis_row_column_normalized.vl.json
@@ -23,6 +23,14 @@
       "y": {
         "field": "Miles_per_Gallon",
         "type": "quantitative"
+      },
+      "color": {
+        "field": "Cylinders",
+        "type": "nominal"
+      },
+      "shape": {
+        "field": "Origin",
+        "type": "nominal"
       }
     }
   }
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -136,6 +136,9 @@ export class FacetModel extends ModelWithField {
   public parseMark() {
     this.child.parseMark();
 
+    // if we facet by two dimensions, we need to add a cross operator to the aggregation
+    // so that we create all groups
+    const isCrossedFacet = this.channelHasField(ROW) && this.channelHasField(COLUMN);
     this.component.mark = [{
       name: this.getName('cell'),
       type: 'group',
@@ -146,9 +149,16 @@ export class FacetModel extends ModelWithField {
           groupby: [].concat(
             this.channelHasField(ROW) ? [this.field(ROW)] : [],
             this.channelHasField(COLUMN) ? [this.field(COLUMN)] : []
-          )
+          ),
+          ...(isCrossedFacet ? {aggregate: {
+            cross: true
+          }}: {})
         }
       },
+      ...(isCrossedFacet ? {sort: {
+        field: [this.field(ROW, {expr: 'datum'}), this.field(COLUMN, {expr: 'datum'})],
+        order: ['ascending', 'ascending']
+      }} : {}),
       encode: {
         update: getFacetGroupProperties(this)
       }
--- a/tslint.json
+++ b/tslint.json
@@ -73,5 +73,6 @@
     "valid-typeof": true,
     "variable-name": [true, "check-format", "allow-leading-underscore", "ban-keywords"],
     "ordered-imports": [true]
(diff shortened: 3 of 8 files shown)


examples/specs/normalized/trellis_row_column_normalized.vl.json
```diff
@@ -23,6 +23,14 @@
       "y": {
         "field": "Miles_per_Gallon",
         "type": "quantitative"
+      },
+      "color": {
+        "field": "Cylinders",
+        "type": "nominal"
+      },
+      "shape": {
```
kanitw: Why over-encoding both shape and color?  (Over-encoding is generally ineffective.)

src/compile/facet.ts
```diff
@@ -136,6 +136,7 @@ export class FacetModel extends ModelWithField {
   public parseMark() {
     this.child.parseMark();
 
+    const facetByTwo = this.channelHasField(ROW) && this.channelHasField(COLUMN);
```
kanitw: crossFacet?

tslint.json
```diff
@@ -73,5 +73,6 @@
     "valid-typeof": true,
     "variable-name": [true, "check-format", "allow-leading-underscore", "ban-keywords"],
     "ordered-imports": [true]
+    // "object-literal-key-quotes": [true, "as-needed"]
```
kanitw: why disabling this?

kanitw:
I wonder we should have grids for those empty cell too.
(Feel free to file an issue if you think the grid should be in a separate PR.)

examples/specs/normalized/trellis_row_column_normalized.vl.json
domoritz: Look in the example above. It makes it clear what is happening. This is not a great visualization but a great test.

tslint.json
domoritz: Becuase all our tests fail this. I didn't want to pollute this pr.

src/compile/facet.ts
domoritz: That's not what it is semantically.

domoritz:
I disagree with your comments. Feel free to disagree or merge.

examples/specs/normalized/trellis_row_column_normalized.vl.json
kanitw: I still don't understand how adding these two channel help with testing. 

Also, we will probably need to add one example of trellis with both row and column to the doc and adding these two channels make this example unnecessarily complicated.

src/compile/facet.ts
kanitw: Yeah -- but in any case both `facetByTwo` and  `crossFacet`are terrible name -- better figure a better one.

tslint.json
kanitw: Why is it failing now but not in the past?

src/compile/facet.ts
kanitw: isCrossedFacet?

tslint.json
domoritz: We never had that option. Note how I added it. I know I shouldn't have put this in this PR.

src/compile/facet.ts
domoritz: Done

examples/specs/normalized/trellis_row_column_normalized.vl.json
domoritz: I added it because I wanted to be sure that I got the order right. And I know once we implement https://github.com/vega/vega-lite/issues/2176, we will have to revisit this again. 

Yes, we need a different example for the docs but I'd say this one is not great anyway (too many cells).

kanitw: I see what you mean now.  I guess we should add some suffix to this file to suggest that it's a test example, not demo example.  (maybe `_test`) so we can preserve the original name for the docs.

tslint.json
kanitw: Removing it seems easy?

domoritz: It's a separate commit. So we can just rebase.

examples/specs/normalized/trellis_row_column_normalized.vl.json
domoritz: Let's cross that bridge when we get to it.

domoritz:
Ready

PR #2413, opened 2017-05-24
adding streaming tutorial

yifanwu:
tutorial for streaming data in vega created with the help of @domoritz 

not sure why site/static/explore.md was included as my change --- didn't make changes to existing tutorial.


site/tutorials/streaming.md
```diff
@@ -0,0 +1,89 @@
...
+  "category": number
+}
+```
+
+As for removing data points, you could either supply the actual data points to be removed similar to the insert above (which requires your storing the data), or you could filter existing rendered data by some attributes and pass the predicate to the `remove()` function, such as in this case `function (t) { return t.x < minimumX; }`.  If you are familiar with D3, this `insert` and `remove` pattern is similar to the [enter exit](https://bost.ocks.org/mike/circles/#entering) pattern.
+
+Putting it together, we have
+```
+  var changeSet = vega.changeset().insert(valueGenerator()).remove(function (t) { return t.x < minimumX; });
+  res.view.change('table', changeSet).run();
```
domoritz: Maybe better just write `view.change` and say that we will cover where the `view` comes from later.

site/tutorials/streaming.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+For simplicity, we have generated some data to simulate real time data updates, but the code could easily be swapped with a server data request.  In the example, the "real time" is being simualted by `window.setInterval` that triggers `view.change` periodically.
+
+The last piece in this example is the Vega [embed](https://github.com/vega/vega-embed) method, which help load Vega/Vega-Lite specs on your webpages. The callback argument to `embed` is invoked when embed is complete, with parameters: error, then an object containing the `view`, which is how we could access the `view.change` mentioned earlier, and `spec`, which is the compiled Vega spec from the original Vega-Lite spec given.
```
domoritz: `error`

site/tutorials/streaming.md
```diff
@@ -0,0 +1,89 @@
...
+
+Below is the javascript code to run this example.  Make sure your html contains a div with id `'chart'`.
```
domoritz: I think we say "JavaScript" in other places.

site/tutorials/streaming.md
```diff
@@ -0,0 +1,89 @@
...
+
+```
+function runStreamingExample(eleId) {
```
domoritz: For the code in the tutorial, I'd say remove this wrapper.

site/tutorials/streaming.md
```diff
@@ -0,0 +1,89 @@
...
+title: Streaming Data
+permalink: /tutorials/streaming.html
+---
+
+
+In this tutorial, you'll learn how to stream new data in a Vega-Lite visualization. If you are not familiar with Vega-Lite, please read the [getting started tutorial]({{site.baseurl}}/tutorials/getting_started.html) first.
+
+We will be using the `View` API from the Vega, where we update data via the [`change`](https://vega.github.io/vega/docs/api/view/#data) method.  Data in Vega and Vega-Lite is updated either via interactions, or by specifying the data changes directly.  In both cases, the data is pushed through the Vega runtime for optimized performance.  This tutorial focuses on the latter for streaming data, where no user interaction is involved.
+
+<div id="chart" class="vl-example"></div>
```
domoritz: why do you need `class="vl-example"`?

yifanwu: i wanted to inline the visualization for better formatting (like in explore data tutorial) but didn't locate the css that did it; and float left looked weird. I'll remove the vl-example (forgot to afterwards)

domoritz:
Since I helped with this, I will ask @kanitw for an additional review.

kanitw:
From skimming quickly, I think this seems helpful.   LGTM

PR #2414, opened 2017-05-24
Set role for path marks as well. cc @sorenmacbeth

domoritz:
cc @sorenmacbeth


kanitw:
LGTM. Feel free to merge when test pass.

Issue #2331, opened 2017-04-30
Do not init all resolve for all channels.


kanitw:
Follow up for this thread: https://github.com/vega/vega-lite/pull/2327#discussion_r114066873.

domoritz:
wasteful -> confusing (I don't care about perf here)

kanitw:
By initing resolve, we cannot choose not to merge scales if their types are conflicting.  
This can come after 2.0 though.

kanitw:
>By initing resolve, we cannot choose not to merge scales if their types are conflicting.
This can come after 2.0 though.

We actually can, but the problem is that their corresponding axis/legend would not know from the resolve value.

kanitw:
We might be able to work around this by check if scale component exist.

kanitw:
Fixed in #2642

PR #2416, opened 2017-05-24
Resolve for repeat

domoritz:
cc @willium

### DIFF
--- a/src/compile/repeat.ts
+++ b/src/compile/repeat.ts
@@ -1,13 +1,14 @@
 import {isArray} from 'vega-util';
-import {NONSPATIAL_SCALE_CHANNELS} from '../channel';
-import {CellConfig, Config} from '../config';
+import {ScaleChannel} from '../channel';
+import {Config} from '../config';
 import {Encoding} from '../encoding';
 import {Facet} from '../facet';
 import {Field, FieldDef, isRepeatRef} from '../fielddef';
 import * as log from '../log';
 import {Repeat} from '../repeat';
+import {initRepeatResolve, ResolveMapping} from '../resolve';
 import {RepeatSpec} from '../spec';
-import {contains, Dict, keys, vals} from '../util';
+import {Dict, keys} from '../util';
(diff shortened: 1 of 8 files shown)


src/compile/repeat.ts
```diff
@@ -127,44 +132,39 @@ export class RepeatModel extends Model {
...
-
-            // remove merged scales from children
-            delete child.component.scales[channel];
+      // Check whether the scales are actually compatible, e.g. use the same sort or throw error
+      keys(child.component.scales).forEach((channel: ScaleChannel) => {
+        if (this.resolve[channel].scale === 'shared') {
+          const childScale = child.component.scales[channel];
+          const modelScale = scaleComponent[channel];
+
+          if (!childScale || isSignalRefDomain(childScale.domain) || (modelScale && isSignalRefDomain(modelScale.domain))) {
```
kanitw: FYI, I don't stand the logic for this line.  I know you are not really changing anything here, but this TODO things seems worth noting as an issue that we should deal with before 2.0?

src/resolve.ts
```diff
@@ -46,3 +46,11 @@ export function initLayerResolve(resolve: ResolveMapping): ResolveMapping {
 
   return out;
 }
+
+export function initLayerResolve(resolve: ResolveMapping): ResolveMapping {
+  return initResolve(resolve, channel => 'shared');
+}
+
+export function initRepeatResolve(resolve: ResolveMapping): ResolveMapping {
```
kanitw: I thought in #2331 we agree not to do initResolve?

domoritz: Yeah, but that can come later. Basically, I will add an argument to pass in the channels.

src/compile/repeat.ts
domoritz: Ok

src/resolve.ts
kanitw: Ok

kanitw:
Feel free to merging after filing an issue.

PR #2419, opened 2017-05-24
Implement resolve for facet and concat

domoritz:
screen shot 2017-05-24 at 16 38 31

screen shot 2017-05-24 at 19 25 19

### DIFF
--- a/src/compile/concat.ts
+++ b/src/compile/concat.ts
@@ -1,5 +1,7 @@
+import {ScaleChannel} from '../channel';
 import {CellConfig, Config} from '../config';
 import {Repeat} from '../repeat';
+import {initConcatResolve, ResolveMapping} from '../resolve';
 import {ConcatSpec, isVConcatSpec, RepeatSpec} from '../spec';
 import {Dict, keys, vals} from '../util';
 import {VgData, VgLayout, VgScale, VgSignal} from '../vega.schema';
@@ -8,6 +10,7 @@ import {assembleData} from './data/assemble';
 import {parseData} from './data/parse';
 import {Model} from './model';
 import {RepeaterValue} from './repeat';
+import {moveSharedScaleUp} from './scale/parse';
 
 
 export class ConcatModel extends Model {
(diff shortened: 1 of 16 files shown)


domoritz:
Done and ready.

src/compile/scale/domain.ts
```diff
@@ -235,15 +228,7 @@ function normalizeDomain(domain: UnionableDomain): (any[] | VgDataRef)[] {
...
-    return domain.fields.map(d => {
-      if (util.isArray(d)) {
-        return d;
-      }
-      return {
-        data: d.data,
-        field: d.field
-      };
-    });
+    return domain.fields;
```
kanitw: Can you explain why is this part changing?

src/compile/scale/parse.ts
```diff
@@ -65,3 +66,27 @@ export function parseScale(model: UnitModel, channel: Channel) {
...
 }
+
+export function moveSharedScaleUp(model: Model, scaleComponent: Dict<ScaleComponent>, child: Model, channel: Channel) {
+  // Check whether the scales are actually compatible, e.g. use the same sort or throw error
+
+  const childScale = child.component.scales[channel];
+  let modelScale = scaleComponent[channel];
+
+  if (modelScale) {
+    modelScale.domain = unionDomains(modelScale.domain, childScale.domain);
```
kanitw: Add `// TODO: merge properties if they are not conflicting`

(Beyond the scope of this PR.)

src/compile/scale/domain.ts
```diff
@@ -275,10 +253,12 @@ export function unionDomains(domain1: VgDomain, domain2: VgDomain): VgDomain {
     });
 
     if (util.unique(allData, x => x).length === 1 && allData[0] !== null) {
-      return {
+      // create a union domain of different fields with a single data source
+      const domain: FieldRefUnionDomain = {
```
kanitw: not sure why returning right away doesn't work anymore..

src/compile/concat.ts
```diff
@@ -77,6 +92,16 @@ export class ConcatModel extends Model {
 
     for (const child of this.children) {
       child.parseLegend();
+
+      keys(child.component.legends).forEach((channel: ScaleChannel) => {
+        if (this.resolve[channel].legend === 'shared') {
+          // just use the first legend definition for each channel
```
kanitw: Add `// TODO: merge properties if they are not conflicting`

(This is still wrong, but the fix is beyond the scope of this PR.)

kanitw: Isn't this logic shared by other models?  Extract?

domoritz: Good catch. Done.

src/compile/scale/parse.ts
domoritz: done

PR #2420, opened 2017-05-25
Lookup fixes: move lookup up, stop parse at lookup

willium:
Doesn't work yet.

### DIFF
--- a/src/compile/data/optimizers.ts
+++ b/src/compile/data/optimizers.ts
@@ -8,7 +8,7 @@ import {NullFilterNode} from './nullfilter';
 import {SourceNode} from './source';
 import {StackNode} from './stack';
 import {TimeUnitNode} from './timeunit';
-import {CalculateNode, FilterNode} from './transforms';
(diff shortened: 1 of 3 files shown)


src/compile/data/optimizers.ts
```diff
@@ -38,7 +38,7 @@ export function moveParseUp(node: DataFlowNode) {
 
   // move parse up by merging or swapping
   if (node instanceof ParseNode) {
-    if (parent instanceof SourceNode) {
+    if (parent instanceof SourceNode || parent instanceof LookupNode) {
```
domoritz: I think we want to return true for Lookup Lookup nodes as parents.

src/compile/data/parse.ts
```diff
@@ -100,6 +104,16 @@ export function parseData(model: Model): DataComponent {
   // the current head of the tree that we are appending to
   let head = root;
 
+  const lookups = model.transforms.filter(isLookup);
```
domoritz: Hmm, this is bad. We want to preserve the order. I see that this is a problem...

PR #2424, opened 2017-05-26
Create parse nodes together with the filter nodes and do not move the…

domoritz:
… parse nodes through lookup nodes if the lookup node creates fields that the format parse node needs.

This is necessary because we don't know whether a parse should go into the primary or secondary source of a lookup. This PR resolves this issue by moving parse after transforms but parse gets pushed through transforms if possible. 

Supercedes https://github.com/vega/vega-lite/pull/2420

### DIFF
--- a/src/compile/data/optimizers.ts
+++ b/src/compile/data/optimizers.ts
@@ -8,7 +8,7 @@ import {NullFilterNode} from './nullfilter';
 import {SourceNode} from './source';
 import {StackNode} from './stack';
 import {TimeUnitNode} from './timeunit';
-import {CalculateNode, FilterNode} from './transforms';
+import {CalculateNode, FilterNode, LookupNode} from './transforms';
 
 /**
  * Start optimization path at the leaves. Useful for merging up or removing things.
@@ -43,12 +43,18 @@ export function moveParseUp(node: DataFlowNode) {
     }
 
     if (parent.numChildren() > 1) {
+      // don't move parse further up but continue with parent.
       return true;
     }
 
     if (parent instanceof ParseNode) {
       parent.merge(node);
     } else {
+      // don't swap with nodes that produce something that the parse node depends on (e.g. lookup)
+      if (hasIntersection(parent.producedFields(), node.dependentFields())) {
+        return true;
+      }
+
       node.swapWithParent();
     }
   }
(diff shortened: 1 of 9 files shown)


willium:
Great work! Works with geo! 
image

src/compile/data/optimizers.ts
```diff
@@ -43,12 +43,18 @@ export function moveParseUp(node: DataFlowNode) {
...
 
     if (parent.numChildren() > 1) {
+      // don't move parse further up but continue with parent.
       return true;
     }
 
     if (parent instanceof ParseNode) {
       parent.merge(node);
     } else {
+      // don't swap with nodes that produce something that the parse node depends on (e.g. lookup)
```
kanitw: The comment seems a bit unclear.   Don't swap "what"

src/compile/data/parse.ts
```diff
@@ -125,6 +119,12 @@ export function parseData(model: Model): DataComponent {
     head = last;
   }
 
+  const parse = ParseNode.make(model);
```
kanitw: Why should parse from `encoding` be moved down?

src/compile/data/optimizers.ts
domoritz: the current node we are moving up

src/compile/data/parse.ts
domoritz: See main PR comment. The issue is that the field may come from the secondary source of a lookup.

src/compile/data/optimizers.ts
kanitw: Why "that the parse node depends on (e.g. lookup)"?  

There is no indication that this is a parse node?

src/compile/data/parse.ts
kanitw: Now I wonder why it only changes one file.

domoritz: Becuase we move parse up unless there is a conflict.

src/compile/data/optimizers.ts
domoritz: moveParseUp

kanitw: Ok I'm fooled in github's UI lol

PR #2428, opened 2017-05-29
Ignore scale.domain when binned

### LINKED ISSUE #2422: Bin is not working with specifying domain
When I specified `scale.domain`, `bin` didn't work.
```json
{
  "data": {
    "values": [
      {"x": 0.3349},
      {"x": 0.1216},
      {"x": 0.8341},
      {"x": 0.3341}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "x",
      "type": "quantitative",
      "scale": {"domain": [0,1]},
      "bin": {"maxbins": 20}
    },
    "y": {
      "field": "*",
      "type": "quantitative",
      "aggregate": "count"
    }
  }
}


```
Rendered:
image

What I expected (Vega-Lite 1 returns):
vega 69

yhoonkim:
Fix #2422.

### DIFF
--- a/src/compile/scale/domain.ts
+++ b/src/compile/scale/domain.ts
@@ -64,13 +64,17 @@ export function parseDomain(model: UnitModel, channel: Channel): VgDomain {
 function parseSingleChannelDomain(scale: Scale, model: UnitModel, channel:Channel): VgDomain {
   const fieldDef = model.fieldDef(channel);
 
-  if (scale.domain && scale.domain !== 'unaggregated' && !isSelectionDomain(scale.domain) && !fieldDef.bin) { // explicit value
-    if (isDateTime(scale.domain[0])) {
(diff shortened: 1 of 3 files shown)


domoritz:
Beautiful PR. Thanks!

kanitw:
@domoritz @yhoonkim  I wonder why it should be ignored?  

Currently the system only tells users that bin will ignore `scale.domain` but does not provide a rationale.  Shouldn't this be allowed and instead `extent` is also automatically set to be the same as domain?   (It should then be allow only when both `domain` and `extent` are specified.)

domoritz:
@kanitw Agreed that there may be a smarter behavior but I couldn't thing through all the cases and a warning is better than nothing. I'll send a pr to refine the message.

PR #2429, opened 2017-05-30
Revert "Ignore scale.domain when binned. Fixes #2422"

### LINKED ISSUE #2422: Bin is not working with specifying domain

kanitw:
@domoritz @yhoonkim 
As discussed [here](https://github.com/vega/vega-lite/pull/2428#issuecomment-304756108), 
I wonder if we should revert #2428.

Shouldn't `scale.domain` be allowed for binned field and instead `extent` is also automatically set to be the same as `domain`?

### DIFF
--- a/src/compile/scale/domain.ts
+++ b/src/compile/scale/domain.ts
@@ -65,16 +65,12 @@ function parseSingleChannelDomain(scale: Scale, model: UnitModel, channel:Channe
   const fieldDef = model.fieldDef(channel);
 
   if (scale.domain && scale.domain !== 'unaggregated' && !isSelectionDomain(scale.domain)) { // explicit value
-    if (fieldDef.bin) {
-      log.warn(log.message.conflictedDomain(channel));
-    } else {
-      if (isDateTime(scale.domain[0])) {
-        return (scale.domain as DateTime[]).map((dt) => {
-          return {signal: dateTimeExpr(dt, true)};
-        });
-      }
-      return scale.domain;
+    if (isDateTime(scale.domain[0])) {
+      return (scale.domain as DateTime[]).map((dt) => {
+        return {signal: dateTimeExpr(dt, true)};
+      });
     }
+    return scale.domain;
   }
 
   const stack = model.stack;
--- a/src/log.ts
+++ b/src/log.ts
@@ -212,10 +212,6 @@ export namespace message {
     return `Setting the scale to be independent for ${channel} means we also have to set the guide (axis or legend) to be independent.`;
   }
 
-  export function conflictedDomain(channel: Channel) {
-    return `${channel}-scale's "domain" is ignored as it is binned.`;
-  }
-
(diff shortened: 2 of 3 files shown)


kanitw:
(Or maybe we shouldn't revert, but rather submit another PR to correct the behavior.)

kanitw:
Or at very least, provide more comment that please use "extent" instead.

yhoonkim:
I agree with Ham's point about providing the rationale of ignoring `scale.domain`. (If I got that message, it would have saved my time.) 

I thought `bin.extent` exists for a case that an user sets `scale.domain` as a subset of `bin.extent` like zooming in a binned axis. (Which seems rare.)

I'm agree with hiding `bin.extent` from document and using `scale.domain` as the domain for binning.

kanitw:
> I thought `bin.extent` exists for a case that an user sets `scale.domain` as a subset of `bin.extent` like zooming in the binned axis, which seems rare.

Hmm.  Good question.  
Maybe there is a use case for interactive plots.  


> I agree with hiding bin.extent from document and using scale.domain as the domain for binning.

I didn't suggest hiding `bin.extent` from the document.

yhoonkim:
If you think the case will not be rare, 
My suggestion is

If `scale.domain` is specified, and the channel is binned without `bin.extent`, 
->  Ignore `scale.domain` and warn to use `bin.extent` instead.
(Encouraging to use `bin.extent`)

If `scale.domain` is specified, the channel is binned with `bin.extent`, and `scale.domain` is not a subset of `bin.exten`,
-> Ignore `scale.domain` and warn to use `bin.extent` instead.

If `scale.domain` is specified, the channel is binned with `bin.extent`, and `scale.domain` is a subset of `bin.extent`,
-> Use `scale.domain` as the domain of the channel and use `bin.extent` for binning.


If it will be rare, I just want to hide `bin.extent` and encourage to use `scale.domain` which seems more instinct for me.

kanitw:
I haven't got a time to think about it deeply yet, but Iet me cc: @arvind to make sure we don't overlook anything about zooming and panning.

domoritz:
Updated message in https://github.com/vega/vega-lite/pull/2430

kanitw:
Closing as we still have #2422

PR #2433, opened 2017-05-31
Don't output undefined properties to fix brush

domoritz:
None of the examples that used a brush worked in the example gallery because we passed the generated Vega spec with undefined properties directly to Vega. It caused issues like

screen shot 2017-05-30 at 15 30 52

This PR fixes that.


arvind:
Awesome, I had added the `undefined`s because the TypeScript compiler complained. Didn't think to cast it as `any`. Thanks for the fix!

domoritz:
More appropriately we would cast to "VgMark" but we don't have a type for it yet.

PR #2447, opened 2017-06-04
Add link to vega title config docs

domoritz:


kanitw:
Let's add "intro" paragraph for all sections when we make a pass.  As mentioned on slack, the link is not very useful.

PR #2451, opened 2017-06-05
Fixes for selections over binned/timeUnit fields

### LINKED ISSUE #2320: Incorrect single selection predicate for repeated binned views
I'd expect to be able to just replace the interval selection with a single selection and clock on bars to select them. 

However, the spec below throws an error: `Uncaught Error: Duplicate signal name: "sel"`

```json
{
  "data": {
    "url": "data/flights-2k.json",
    "format": {"parse": {"date": "date"}}
  },
  "transform": [{"calculate": "hours(datum.date)", "as": "time"}],
  "repeat": {"column": ["distance", "delay", "time"]},
  "spec": {
    "layer": [{
      "selection": {
        "sel": {"type": "single", "encodings": ["x"]}
      },
      "mark": "bar",
      "encoding": {
        "x": {
          "bin": {"maxbins": 20},
          "type": "quantitative"
        },
        "y": {"aggregate": "count", "type": "quantitative"}
      }
    }, {
      "transform": [{"filter": {"selection": "sel"}}],
      "mark": "bar",
      "encoding": {
        "x": {
          "bin": {"maxbins": 20},
          "type": "quantitative"
        },
        "y": {"aggregate": "count", "type": "quantitative"},
      }
    }]
  }
}

```

### LINKED ISSUE #2448: Push timeUnit calculations to source dataset?
Surfacing a Slack conversation for us to dig into further. 

Currently, point (single/multi) selections projected over binned and timeUnit fields do not work across multiview displays (#2320). @domoritz and I came up with the following strategy to deal with binned fields:
* Capture the `bin_start` and `bin_end` during the interaction.
* The `vlPoint` expression function should treat those as a range over the raw field.

A similar strategy could make sense with `timeUnit` (where `vlPoint` recalculates a `timeUnit` during the test) but has a couple of drawbacks: (1) duplicating (or at a minimum referencing) `timeUnit` logic in vega-parser; (2) performance of these recalculations.

We didn't want to push bin operations up the transform pipeline because filters could affect the binning strategy. However, I don't think such concerns affect `timeUnit`. So, would it be possible to push `timeUnit` calculations to the source dataset? I believe this would make them accessible globally, with the only danger being a user manually overriding them (something that could potentially be error-checked).

arvind:
Please merge after #2441. 

This PR includes two fixes when using selections across multiview displays:
* For binned fields, point selections store the bin extents which is tested by the corresponding predicate function (vega/vega-parser#35). Fixes #2320.
* For timeUnit fields, the selection is constructed using the derived (not raw field). This assumes we're able to resolve #2448 accordingly.

### DIFF
--- a/src/compile/selection/multi.ts
+++ b/src/compile/selection/multi.ts
@@ -1,4 +1,4 @@
-import {stringValue} from '../../util';
+import {keys, stringValue} from '../../util';
 import {SelectionCompiler, TUPLE} from './selection';
 
 const multi:SelectionCompiler = {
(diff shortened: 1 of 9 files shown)


src/compile/selection/multi.ts
```diff
@@ -8,21 +8,32 @@ const multi:SelectionCompiler = {
     const proj = selCmpt.project,
         datum  = '(item().isVoronoi ? datum.datum : datum)',
         encodings = proj.map((p) => stringValue(p.encoding)).join(', '),
+        bins = {},
         fields = proj.map((p) => stringValue(p.field)).join(', '),
-        values = proj.map((p) => `${datum}[${stringValue(p.field)}]`).join(', ');
+        values = proj.map((p) => {
+          const c = p.encoding, f = p.field;
+          // Binned fields should capture extents, for a range test against the raw field.
+          return model.fieldDef(c).bin ? (bins[f] = 1,
```
domoritz: Here you use f but not in ll 19. I think in this case it's better not to create an alias.

src/compile/selection/transforms/project.ts
```diff
@@ -11,7 +11,14 @@ const project:TransformCompiler = {
     let fields = {};
     // TODO: find a possible channel mapping for these fields.
     (selDef.fields || []).forEach((f) => fields[f] = null);
-    (selDef.encodings || []).forEach((c: Channel) => fields[model.fieldDef(c).field] = c);
+    (selDef.encodings || []).forEach((c: Channel) => {
```
domoritz: Can you rename

f -> field
c -> channel

arvind:
Requested changes have been made, @domoritz!

PR #2452, opened 2017-06-06
Don't generate redundant scales

### LINKED ISSUE #2438: Redundant scales with repeat
With a repeated spec (e.g., the [SPLOM](https://vega.github.io/new-editor/?mode=vega-lite&spec=interactive_splom)), each child contains an x and y scale and then these x and y scales are also collected at the top-level. These additional scales incur a performance penalty if we're panning/zooming them, so we should try to reduce this redundancy if possible.

domoritz:
Fixes #2438 

Note that the fix was to delete code 😎 .


kanitw:
@domoritz  Hmm.  Why is the fix removing assembleScale entirely instead of delete scale component from the child when merge? 

Would this still remain correct for independent scales in the children?

domoritz:
Copying response from slack. 

If you look at the specs, we used to output the exact same scale multiple times. All repeat and concat specs already use independent scales and they work.

PR #2463, opened 2017-06-07
Remove main group

### LINKED ISSUE #2203: Query Widget contains signals with duplicate name
Previously the output for the `query_widget` example contains two `CylYr` signals: one at the top level and another inside the `"main-group"`.

However, by removing layout data in https://github.com/vega/vega-lite/pull/2195, we no longer need the `"main-group"` from importing `layout` data and thus can flatten the output Vega spec.  

However, now the two `CylYr` signals reside at the same level and thus cause Vega error

```
Duplicate signal name: "CylYr"
```

See the spec diff here
https://github.com/vega/vega-lite/pull/2195/commits/1a151d96efcb5bb7380143ced8a0e6fde9d554ed#diff-d50e1bd9ff23ce19ef96cce2d08d4e98

Since these two signal probably serve different purpose anyway, I think we should just rename them.  

To avoid unintended consequences, I think I better leave this to @arvind.

arvind:
This PR removes the "nested_main_group" again, and makes the necessary changes to the selection infrastructure. 

Fixes #2203.

### DIFF
--- a/examples/vg-specs/area_vertical.vg.json
+++ b/examples/vg-specs/area_vertical.vg.json
@@ -3,6 +3,13 @@
     "description": "Area chart showing weight of cars over time (vertical).",
     "autosize": "pad",
     "padding": 5,
+    "encode": {
+        "update": {
+            "fill": {
+                "value": "transparent"
+            }
+        }
+    },
     "data": [
         {
             "name": "source_0",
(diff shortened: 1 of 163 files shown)


src/compile/compile.ts
```diff
@@ -85,14 +92,6 @@ function assemble(model: Model, topLevelProperties: TopLevelProperties) {
         model.assembleSelectionTopLevelSignals([])
       )
     ),
-
-    // FIXME: get rid of the top-level `nested-main-group`
-    // HACK: this is a hack to temporarily make selections works as
-    // 1) Currently, some selection's signals rely on the main group's scope to shadow duplicate names.
-    // 2) Selection predicate depends on parent reference which may not exist for top-level mark.
-    ...assembleNestedMainGroup(model),
```
kanitw: Let's also remove the unused `assembleNestedMainGroup`?

arvind: D'oh, think I temporarily undid it's removal during some churn and forgot to redo it. Fixed.

examples/vg-specs/area_vertical.vg.json
```diff
@@ -3,6 +3,13 @@
     "description": "Area chart showing weight of cars over time (vertical).",
     "autosize": "pad",
     "padding": 5,
+    "encode": {
+        "update": {
+            "fill": {
+                "value": "transparent"
```
arvind: Yes. Try `brush.vg.json` but without this block. You'll find that events don't fire if you're interacting in the "whitespace."

src/compile/compile.ts
```diff
@@ -59,17 +59,26 @@ function assemble(model: Model, topLevelProperties: TopLevelProperties) {
   // autoResize has to be put under autosize
   const {autoResize, ...topLevelProps} = topLevelProperties;
 
+  const encode = model.assembleParentGroupProperties();
+  let width;
```
arvind: They're not except to extract them from `encode` to be able to pass the remaining properties to the subsequent block. I could have used `delete` as well I guess...

domoritz: Hmm, I think you can use a destuctor by name 

```ts
const foo = {a: 1, b:2};
const {b} = foo;
```

examples/vg-specs/area_vertical.vg.json
domoritz: I guess later we might only add this if there is a selection in the spec. Thanks for the clarification.

src/compile/compile.ts
arvind: Cleaned up with f9dbb9e. I didn't want to specifically pull out just `fill` because there may be other encoding properties specified in the config?

kanitw: Looking at this again, maybe it's worth commenting why we don't need width/height here, but need it when we call `assembleParentGroupProperties` in other places?

PR #2465, opened 2017-06-07
Allow data parse after scale parse. 

domoritz:


src/compile/facet.ts
```diff
@@ -304,7 +304,7 @@ export class FacetModel extends ModelWithField {
       ...this.child.assembleGroup()
     }];
 
-    return marks.map(this.correctDataNames);
```
domoritz: @kanitw I don't think we need this.

src/compile/data/dataflow.ts
```diff
@@ -91,44 +91,48 @@ export class DataFlowNode {
...
   }
 
-  constructor(source: string, public readonly type: DataSourceType) {
+  constructor(source: string, public readonly type: DataSourceType, private readonly refCounts: Dict<number>) {
     super(source);
 
-    this._source = source;
+    this._source = this._name = source;
+
+    if (this.refCounts && !(this._name in this.refCounts)) {
```
kanitw: I assume you have one `refCounts` per one spec compiled?
If so, add comment somewhere to make it clear.

Also, I wonder what should we do when we have name collision?

src/compile/model.ts
```diff
@@ -269,7 +269,11 @@ export abstract class Model {
     // Increase ref count. This is critical because otherwise we won't create a data source.
     // We also increase the ref counts on OutputNode.getSource() calls.
     const refCounts = this.component.data.outputNodeRefCounts;
-    refCounts[fullName]++;
+    if (!(fullName in refCounts)) {
```
kanitw: If you wanna do one line, you can do

refCounts[fullName] = (refCounts[fullName] || 0) + 1;

src/compile/data/dataflow.ts
domoritz: We should not have name collisions because we use `model.getName(type)`

kanitw: It is probably still possible due to a user's mistake?  (If they specify two sub-specs with the same name, possibly because they copy one to another?) 

We could handle this in a separate issue though.

Issue #2326, opened 2017-04-29
Layout signals do not work if we set a name


domoritz:
Error: `Uncaught Error: Unrecognized signal name: "foo_width"`

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "name": "foo",
  "description": "The Trellis display by Becker et al. helped establish small multiples as a “powerful mechanism for understanding interactions in studies of how a response depends on explanatory variables”. Here we reproduce a trellis of Barley yields from the 1930s, complete with main-effects ordering to facilitate comparison.",
  "data": {"url": "data/barley.json"},
  "mark": "point",
  "encoding": {
    "row": {
      "field": "site", "type": "ordinal"
    },
    "x": {
      "aggregate": "median", "field": "yield", "type": "quantitative",
      "scale": {"zero": false}
    },
    "y": {
      "field": "variety", "type": "ordinal",
      "sort": {"field": "yield","op": "median", "order": "descending"},
      "scale": {"rangeStep": 12}
    },
    "color": {"field": "year", "type": "nominal"}
  }
}
```

kanitw:
It seems to work if we don't have `"row"` so the bug is in facet.

kanitw:
The problem is that our output produces unnecessary "width" and "height" signals for FacetModel

PR #2466, opened 2017-06-07
Do not generate unnecessary top-level width/height signals for layout-based models

kanitw:
Fix https://github.com/vega/vega-lite/issues/2326

vega_editor

### DIFF
--- a/examples/specs/normalized/trellis_barley_normalized.vl.json
+++ b/examples/specs/normalized/trellis_barley_normalized.vl.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "name": "trellis_barley",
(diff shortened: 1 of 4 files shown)


domoritz:
Can you add an example spec that triggers this error?

kanitw:
Do you mean add to the gallery?   (An example is in the screenshot)

domoritz:
> Do you mean add to the gallery? (An example is in the screenshot)

More like the test specs.

kanitw:
Done

domoritz:
hmm, I don't understand why none of the top-level size signals disappear.

kanitw:
@domoritz  -- splitting this into two commits should make it obvious now.

domoritz:
Now I see it. Lgtm.

PR #2469, opened 2017-06-07
Use `model.getScaleComponent()` instead of `model.scale()`

kanitw:
This way we read the right scale properties (including type, etc.) post merging. 

A small step toward #2251


examples/vg-specs/box-plot_minmax_2D_horizontal.vg.json
```diff
@@ -266,8 +266,7 @@
                     },
                     "yc": {
                         "scale": "y",
-                        "field": "age",
-                        "band": 0.5
```
kanitw: For context, it turns out that from scale merging, these plots are using point scale but the mark compiler previously read scale type from the original scale prior to merging.

PR #2471, opened 2017-06-07
Support structured logical composition of selections

### LINKED ISSUE #2322: Union two different selections in conditionals
In our infovis prototype, we allowed two selections in `filterWith`. Can we use something similar in conditionals?

More specifically, I'd like to be able to `"condition": {"selection": "!brush || !bar", "value": "steelblue"},` where `brush` is an interval selection and `bar` is a single or multi selection. 

From infovis prototype

```js
"transform": {
          "calculate": [{"field": "hour", "expr": "hours(datum.date)"}],
          "filterWith": ["bins", "region"]
        },
```

arvind:
This PR resolves #2322 by allowing structured logical composition of selections in both the `condition` and selection `filter`. A general purpose `logicalComposition` utility (and associated type definitions) has been introduced for future use with the filter transform as well.

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1036,7 +1036,7 @@
       "additionalProperties": false,
       "properties": {
         "selection": {
-          "type": "string"
+          "$ref": "#/definitions/LogicalOperand"
         },
         "value": {
           "type": [
@@ -1055,7 +1055,7 @@
       "additionalProperties": false,
       "properties": {
         "selection": {
-          "type": "string"
+          "$ref": "#/definitions/LogicalOperand"
         },
         "value": {
           "type": [
@@ -1075,7 +1075,7 @@
       "additionalProperties": false,
       "properties": {
         "selection": {
-          "type": "string"
+          "$ref": "#/definitions/LogicalOperand"
         },
         "value": {
           "type": "number"
@@ -1091,7 +1091,7 @@
       "additionalProperties": false,
       "properties": {
         "selection": {
-          "type": "string"
+          "$ref": "#/definitions/LogicalOperand"
         },
         "value": {
           "type": "string"
@@ -2816,6 +2816,64 @@
       },
       "type": "object"
     },
+    "LogicalAnd": {
+      "additionalProperties": false,
+      "properties": {
+        "and": {
+          "items": {
+            "$ref": "#/definitions/LogicalOperand"
+          },
+          "type": "array"
+        }
+      },
+      "required": [
+        "and"
+      ],
+      "type": "object"
+    },
+    "LogicalNot": {
+      "additionalProperties": false,
+      "properties": {
+        "not": {
+          "$ref": "#/definitions/LogicalOperand"
+        }
+      },
+      "required": [
+        "not"
+      ],
+      "type": "object"
+    },
+    "LogicalOperand": {
+      "anyOf": [
+        {
+          "$ref": "#/definitions/LogicalNot"
+        },
+        {
+          "$ref": "#/definitions/LogicalAnd"
+        },
+        {
+          "$ref": "#/definitions/LogicalOr"
+        },
+        {
+          "type": "string"
+        }
+      ]
+    },
+    "LogicalOr": {
+      "additionalProperties": false,
+      "properties": {
+        "or": {
+          "items": {
+            "$ref": "#/definitions/LogicalOperand"
+          },
+          "type": "array"
+        }
+      },
+      "required": [
+        "or"
+      ],
+      "type": "object"
+    },
     "LookupData": {
       "additionalProperties": false,
       "properties": {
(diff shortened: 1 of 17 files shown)


build/vega-lite-schema.json
```diff
@@ -2816,6 +2826,64 @@
       },
       "type": "object"
     },
+    "LogicalAnd<string>": {
```
kanitw: `<string>` wouldn't be friendly for generating commands in Altair. 

Maybe add rename commands for `LogicalAnd/Or/Not/Operand<string>` to `scripts/rename-schema.sh`?

src/logical.ts
```diff
@@ -0,0 +1,25 @@
+export type LogicalOperand<T> = LogicalNot<T> | LogicalAnd<T> | LogicalOr<T> | T;
```
kanitw: I think this could be applied to `FilterTransform` too? 

```diff
-filter: Filter;
+filter: LogicalOperand<Filter>;
```

src/util.ts
```diff
@@ -209,3 +210,15 @@ export function varName(s: string): string {
   // Add _ if the string has leading numbers.
   return (s.match(/^\d+/) ? '_' : '') + alphanumericS;
 }
+
+export function logicalExpr<T>(op: LogicalOperand<T>, cb: Function): string {
+  if (isLogicalNot(op)) {
+    return '!' + logicalExpr(op.not, cb);
```
kanitw: These would be safer with more parentheses `!(${logicalExpr(op.not, cb)})`.

For example, just in case someone apply this to `filter :{not: "A && B"}` then it should become `!(A && B)` rather than `!A && B` -- obviously this is a strawman example but better be safe.
(Note that I imagine logical expression can be applied for `transform.filter` and thus A and B are predicates, not  selection names.)

Similarly, I think we could do `.join(') && (')` instead of `.join(' && ')` (and same for `||`)

src/filter.ts
```diff
@@ -110,8 +110,7 @@ export function expression(model: Model, filter: Filter): string {
   if (isString(filter)) {
     return filter;
   } else if (isSelectionFilter(filter)) {
-    const selection = model.getSelectionComponent(filter.selection);
-    return predicate(model, filter.selection, selection.type, selection.resolve, null);
+    return predicate(model, filter.selection);
```
kanitw: FYI, `selection` in SelectionFilter is still a string, not a `LogicalOperand` yet

build/vega-lite-schema.json
arvind: Aha, I was wondering how to handle that. Thanks!

src/logical.ts
arvind: Good catch. Fixed.

src/util.ts
arvind: I'm not sure additional parentheses are needed. [I added](https://github.com/vega/vega-lite/blob/7c3d1e42e4976be023035182a407bb2fa042690a/test/compile/selection/predicate.test.ts#L49) your `!(A && B)` example to the tests and it produces what I would expect. I think we only need parentheses around the array operators to group all those items together.

src/filter.ts
arvind: Thanks. [Added tests](https://github.com/vega/vega-lite/blob/7c3d1e42e4976be023035182a407bb2fa042690a/test/compile/selection/predicate.test.ts#L83) to ensure that kind of oversight can't happen in the future!

arvind:
Thanks for the comments, @kanitw. They should be addressed now, and the branch rebased with builds passing.

src/util.ts
kanitw: Maybe I wasn't very clear (sorry). 

To clarify, it is not needed for combining selection predicates, but can be safer for using with normal (non-selection) filter expression.  

For example,  `filter :{not: "datum.a > 10 && datum.a<0"}` should become `!(datum.a > 10 && datum.a<0)`, not `!datum.a > 10 && datum.a<0`.    

Note that `filter :{not: "datum.a > 10 && datum.a<0"}` is literally the spec and it needs extra parentheses.  We don't need extra parentheses for `filter :{not: {and: ["datum.a > 10", "datum.a<0"]}}`

arvind: Oh I see. In such cases, I wonder whether the corresponding callback function should be responsible for wrapping the string in parentheses? It's seems specific to the case that `op` is a logical expression. It also results in slightly more convoluted strings for selection predicates:

```diff
{"or": ["one", {"and": ["two", {"not": "three"}]}]}
-      '(vlPoint("one_store", "", datum, "union", "all") || ' +
-      '(vlPoint("two_store", "", datum, "union", "all") && ' +
-      '!vlInterval("three_store", "", datum, "intersect", "others")))');
+      '(vlPoint("one_store", "", datum, "union", "all")) || ' +
+      '((vlPoint("two_store", "", datum, "union", "all")) && ' +
+      '(!(vlInterval("three_store", "", datum, "intersect", "others"))))');
```
Anyways, fairly minor point so I've pushed the requested change in 11fa46e.

PR #2472, opened 2017-06-07
Cleaner approach for SelectionDomain

### LINKED ISSUE #2470: Why SelectionDomain doesn't require selection?
From the type guard in `src/scale.ts`
```
export function isSelectionDomain(domain: Domain): domain is SelectionDomain {
  return domain && domain['selection'];
}
```
SelectionDomain should have `selection`. 

However, its declaration does not require `selection`. 

```
export type SelectionDomain = {signal?: string, selection?: string, field?: string, encoding?: string};
```

Any reason for this?

arvind:
In 77ff32c, we introduced a hack to allow scale domains to be parameterized by selections. The motivation for the hack still persists: scales are parsed before selections, but we need selections to be parsed in order to find their component in the model tree. 

However, rather than polluting our type/interface definitions, this PR resolves #2470 by piggybacking off `scale.domainRaw` using a special key (`SELECTION_DOMAIN`). If this key is found during scale assembly, it is replaced with the correct expression function. 

(The circular dependency issue we experienced in 77ff32c seems to have resolved itself, phew!). 

Because this is still somewhat hacky, I've also added tests to ensure that the correct things are happening at both scale parsing and assembly.

### DIFF
--- a/src/compile/scale/assemble.ts
+++ b/src/compile/scale/assemble.ts
@@ -8,12 +8,17 @@ import {isRawSelectionDomain, scaleDomain} from '../selection/selection';
 
 export function assembleScale(model: Model) {
     return vals(model.component.scales).map(scale => {
-      // Correct any raw selection domains.
+      // As scale parsing occurs before selection parsing, a temporary signal
+      // is used for domainRaw. Here, we detect if this temporary signal
+      // is set, and replace it with the correct domainRaw signal.
+      // For more information, see isRawSelectionDomain in selection.ts.
       if (scale.domainRaw && isRawSelectionDomain(scale.domainRaw)) {
         scale.domainRaw = scaleDomain(model, scale.domainRaw);
       }
 
-      // Correct references to data
+      // Correct references to data as the original domain's data was determined
+      // in parseScale, which happens before parseData. Thus the original data
+      // reference can be incorrect.
       const domain = scale.domain;
       if (isDataRefDomain(domain) || isFieldRefUnionDomain(domain)) {
         domain.data = model.lookupDataSource(domain.data);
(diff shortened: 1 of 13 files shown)


src/compile/scale/parse.ts
```diff
@@ -53,8 +55,11 @@ export function parseScale(model: UnitModel, channel: Channel) {
     range: parseRange(scale)
   };
 
+  // See comment in selection.ts for why such a hack is necessary.
```
kanitw: This comment won't be clear for reader what is the hack here. 

Maybe add a comment right above the signal line to say that the signal there is not a proper signal name but r rather a temporary name that will be replaced later (and see comment for `isRawSelectionDomain` in selection.ts for more details)?

src/compile/scale/assemble.ts
```diff
@@ -4,37 +4,16 @@ import {isSelectionDomain} from '../../scale';
...
-          log.warn('Use "bind": "scales" to setup a binding for scales and selections within the same view.');
-        } else {
-          selCmpt = model.getSelectionComponent(selName);
-          scale.domainRaw = {
-            signal: (selCmpt.type === 'interval' ? 'vlIntervalDomain' : 'vlPointDomain') +
-            `(${stringValue(selCmpt.name + '_store')}, ${stringValue(raw.encoding)}, ${stringValue(raw.field)}, ` +
-            `${stringValue(SELECTION_OPS[selCmpt.resolve])})`
-          };
-        }
+      // Correct any raw selection domains.
```
kanitw: // Correct any raw selection domains since the original `domainRaw` value was determined in `parseScale`, which happens before `parseSelection`. Thus the original `domainRaw` value can be incorrect.

src/compile/scale/assemble.ts
```diff
@@ -4,37 +4,16 @@ import {isSelectionDomain} from '../../scale';
...
+      if (scale.domainRaw && isRawSelectionDomain(scale.domainRaw)) {
+        scale.domainRaw = scaleDomain(model, scale.domainRaw);
       }
 
-      // correct references to data
+      // Correct references to data
```
kanitw: // Correct references to data as the original `domain`'s `data` was determined in `parseScale`, which happens before `parseData`. Thus the original data reference can be incorrect.

arvind:
Requested changes made with e019daf.

PR #2474, opened 2017-06-07
Support structured composition of Filters

arvind:
#2471 introduced structured composition of logical operators but only applied it to composition of selection predicates. This PR extends it to support all `Filter` operators. 

~Note: the previously requested changes to parenthesis (11fa46e) were buggy. This PR fixes this bug as well -- parentheses should wrap the base operand as well as group operands that are and'd/or'd together.~

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1977,7 +1977,7 @@
       "additionalProperties": false,
       "properties": {
         "filter": {
-          "$ref": "#/definitions/Filter",
+          "$ref": "#/definitions/FilterOperand",
           "description": "A string containing the filter Vega expression. Use `datum` to refer to the current data object."
         }
       },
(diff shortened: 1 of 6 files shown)


kanitw:
> parentheses should wrap the base operand as well as group operands that are and'd/or'd together.

Could you clarify why it is necessary to group operands that are and'd/or'd together?  
(Maybe give an example where it is necessary -- I was thinking about it earlier and thought that it was not necessary, but I could be wrong.)

arvind:
Sure, consider `{"and": ["A", {"or": ["B", "C"]}]}`. By "grouping operands", I mean that we want `(B || C)` to be and'd with `A` i.e., `A && (B || C)`. This is essentially what the fix produces (with additional parens for security): `((A) && ((B) || (C))`. 

The prior suggested fix correctly identified the need to wrap the base operand in parents but applied it in incorrect places (along the not, and, or branches). With the above example, it produced the following (notice the mismatched parens): `(A) && (B) || (C))`.

arvind:
I have no idea what bugs in parentheses I thought I'd observed earlier -- I can't seem to reproduce it whatsoever! Think I was just going cross-eyed matching nested parentheses! I've undone the change which produces expressions with fewer parens, which is a pro!

PR #2480, opened 2017-06-08
Make `value` and `type` required for `ValueDef` and `FieldDef` respectively

kanitw:
Make `value` and `type` required for `ValueDef` and `FieldDef` respectively.

(I was trying to implement condition with fieldDef inside the condition as discover that value and type are somehow not required.) 

This actually helps me discover a few bugs and refactor the codebase.

- Refactor method signature for mark mixins
- Make sure to call `isFieldDef` in many places
- Add type in many places
- Correct logic for array of fieldDefs for `detail` and `order` channel

cc: @mattwchun  -- please look at reference changes in `boxplot.ts`

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1118,6 +1118,9 @@
           ]
         }
       },
+      "required": [
+        "value"
+      ],
       "type": "object"
     },
     "ConditionalValueDef<number>": {
@@ -1131,6 +1134,9 @@
           "type": "number"
         }
       },
+      "required": [
+        "value"
+      ],
       "type": "object"
     },
     "ConditionalValueDef<string>": {
@@ -1144,6 +1150,9 @@
           "type": "string"
         }
       },
+      "required": [
+        "value"
+      ],
       "type": "object"
     },
     "Config": {
@@ -1897,6 +1906,9 @@
           "description": "The encoded field's type of measurement. This can be either a full type\nname (`\"quantitative\"`, `\"temporal\"`, `\"ordinal\"`,  and `\"nominal\"`)\nor an initial character of the type name (`\"Q\"`, `\"T\"`, `\"O\"`, `\"N\"`).\nThis property is case-insensitive."
         }
       },
+      "required": [
+        "type"
+      ],
       "type": "object"
     },
     "FacetedUnitSpec": {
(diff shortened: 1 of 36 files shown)


build/vega-lite-schema.json
```diff
@@ -1952,6 +1964,45 @@
           "description": "The encoded field's type of measurement. This can be either a full type\nname (`\"quantitative\"`, `\"temporal\"`, `\"ordinal\"`,  and `\"nominal\"`)\nor an initial character of the type name (`\"Q\"`, `\"T\"`, `\"O\"`, `\"N\"`).\nThis property is case-insensitive."
         }
       },
+      "required": [
+        "type"
+      ],
+      "type": "object"
+    },
+    "FieldDefBase": {
```
kanitw: This is not really referenced, but the new generator somehow already refers to every extended class.  (This is a known bug, but does not harm besides making the schema a bit unnecessary big.)

src/encoding.ts
```diff
@@ -176,12 +193,19 @@ export function isRanged(encoding: EncodingWithFacet<any>) {
 
 export function fieldDefs(encoding: EncodingWithFacet<Field>): FieldDef<Field>[] {
   const arr: FieldDef<Field>[] = [];
-  CHANNELS.forEach(function(channel) {
-    if (channelHasField(encoding, channel)) {
+  CHANNELS.forEach(function(channel: Channel) {
```
kanitw: TODO: Check why we have to change this part at all

src/fielddef.ts
```diff
@@ -274,51 +277,53 @@ export function defaultType(fieldDef: FieldDef<Field>, channel: Channel): Type {
...
-      const {aggregate, ...fieldDefWithoutAggregate} = fieldDef;
-      log.warn(log.message.invalidAggregate(fieldDef.aggregate));
-      fieldDef = fieldDefWithoutAggregate;
-    }
+  if (isFieldDef(channelDef)) {
+    return normalizeFieldDef(channelDef, channel);
+  }
+  return channelDef;
+}
+export function normalizeFieldDef(fieldDef: FieldDef<string>, channel: Channel) {
```
kanitw: Do we really need to change this part?

test/compile/selection/scales.test.ts
```diff
@@ -36,11 +37,11 @@ describe('Selection + Scales', function() {
             },
             color: {
               field: "symbol", type: "nominal",
-              scale: {domain: {selection: "brush2"}}
+              scale: {domain: {selection: "brush2"} as Domain}
```
kanitw: @arvind    Currently 
```
export type SelectionDomain = {selection: string, field: string} | {selection: string, encoding: string};
```

so `{selection: "brush2"}` actually does not satisfy any case.  Not sure if the test spec here is intended.  (Same for the `{selection: "brush3"}` below.)

arvind: How bizarre -- I don't understand why my VSCode didn't show this as an error, nor why it wasn't caught by Travis. The cast looks fine to me -- I had wanted to test a warning in case the spec was invalid. `brush3` can include an encoding or field, that's testing another warning.

kanitw: I think it was previously not detected by Typescript because we didn't require `value` in ValueDef and `type` in FieldDef -- thus preventing it from infer if a spec is a unit spec and so on.  

In this ambiguous state, somehow Typescript just ignore type checking totally.  I guess this behavior makes sense for a codebase that migrates from JS and thus partially have type defined.  

Anyway, I'm glad I notice this and have fixed a lot of things in this PR :)

PR #2481, opened 2017-06-08
Fix line grouping when the line is grouped by binned field	

kanitw:
Please merge #2480 first


kanitw:
@arvind Thanks for adding the screenshot that I forgot to add. :)

PR #2485, opened 2017-06-09
Boxplot IQR

mattwchun:
This pr adds IQR support in the boxplot compositemark.

Fixed issue in groupby in box plot so it groupby’s the correct fields. Also added tests for these cases. Changed the name of transform aggregate in summarize for the median of the box to midBox.

### DIFF
--- a/src/compile/data/assemble.ts
+++ b/src/compile/data/assemble.ts
@@ -211,6 +211,12 @@ function makeWalkTree(data: VgData[]) {
       dataSource.transform = dataSource.transform.concat(node.assemble());
     }
 
+    if (node instanceof AggregateNode) {
+      if (!dataSource.name) {
+        dataSource.name = `data_${datasetIndex++}`;
+      }
+    }
+
     if (node instanceof OutputNode) {
       if (dataSource.source && dataSource.transform.length === 0) {
         node.setSource(dataSource.source);
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -157,7 +157,6 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT 
 export function boxOrient(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | BoxPlotDef>): Orient {
   const {mark: mark, encoding: encoding, ...outerSpec} = spec;
 
-  const is2D = encoding.x && encoding.y;
   const continuousXAxis = isFieldDef(encoding.x) && isContinuous(encoding.x);
   const continuousYAxis = isFieldDef(encoding.y) && isContinuous(encoding.y);
 
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -13,7 +13,8 @@ export type BoxPlotRole = 'boxWhisker' | 'box' | 'boxMid';
 
 export interface BoxPlotDef {
   type: BOXPLOT;
-  orient: Orient;
+  orient?: Orient;
+  extent?: 'min-max' | number;
 }
 
 export function isBoxPlotDef(mark: BOXPLOT | BoxPlotDef): mark is BoxPlotDef {
@@ -49,13 +50,25 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT 
   const {color: _color, ...nonPositionEncodingWithoutColorSize} = nonPositionEncodingWithoutSize;
   const midTickAndBarSizeChannelDef = size ? {size: size} : {size: {value: config.box.size}};
 
-  let discreteAxisFieldDef;
-  let continuousAxisChannelDef: PositionFieldDef<Field>;
-  let discreteAxis;
-  let continuousAxis;
+  let kIQRScalar: number = undefined;
+  if (isBoxPlotDef(mark)) {
+    if (mark && mark.extent) {
+      if(typeof(mark.extent) === 'number') {
+        kIQRScalar = mark.extent;
+      }
+    }
+  }
+  const isMinMax = kIQRScalar === undefined;
 
-  if (isFieldDef(encoding.x) && isFieldDef(encoding.y)) {
+  let discreteAxisFieldDef: PositionFieldDef<Field>, continuousAxisChannelDef: PositionFieldDef<Field>;
+  let discreteAxis, continuousAxis;
+
+  let is1D = true;
+
+  if (encoding.x && encoding.y) {
     // 2D
+
+    is1D = false;
 
     const orient: Orient = box2DOrient(spec);
     const params = box2DParams(spec, orient);
@@ -80,40 +93,13 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT 
     throw new Error(`Continuous axis should not have customized aggregation function ${continuousAxisChannelDef.aggregate}`);
   }
 
-  const baseContinuousFieldDef = {
-      field: continuousAxisChannelDef.field,
-      type: continuousAxisChannelDef.type
-  };
-
-  const minFieldDef = {
-    aggregate: 'min',
-    ...baseContinuousFieldDef
-  };
-  const minWithAxisFieldDef = {
-    axis: continuousAxisChannelDef.axis,
-    ...minFieldDef
-  };
-  const q1FieldDef = {
-    aggregate: 'q1',
-    ...baseContinuousFieldDef
-  };
-  const medianFieldDef = {
-    aggregate: 'median',
-    ...baseContinuousFieldDef
-  };
-  const q3FieldDef = {
-    aggregate: 'q3',
-    ...baseContinuousFieldDef
-  };
-  const maxFieldDef = {
-    aggregate: 'max',
-    ...baseContinuousFieldDef
-  };
+  const transformDef = boxTransform(encoding, discreteAxisFieldDef, continuousAxisChannelDef, kIQRScalar, is1D);
 
   const discreteAxisEncodingMixin = discreteAxisFieldDef !== undefined ? {[discreteAxis]: discreteAxisFieldDef} : {};
 
   return {
     ...outerSpec,
+    transform: transformDef,
     layer: [
       { // lower whisker
         mark: {
@@ -122,8 +108,15 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT 
         },
         encoding: {
           ...discreteAxisEncodingMixin,
-          [continuousAxis]: minWithAxisFieldDef,
-          [continuousAxis + '2']: q1FieldDef,
+          [continuousAxis]: {
+            axis: continuousAxisChannelDef.axis,
+            field: 'lowerWhisker',
+            type: continuousAxisChannelDef.type
+          },
+          [continuousAxis + '2']: {
+            field: 'lowerBox',
+            type: continuousAxisChannelDef.type
+          },
           ...nonPositionEncodingWithoutColorSize
         }
       }, { // upper whisker
@@ -133,8 +126,14 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT 
         },
         encoding: {
           ...discreteAxisEncodingMixin,
-          [continuousAxis]: q3FieldDef,
-          [continuousAxis + '2']: maxFieldDef,
+          [continuousAxis]: {
+            field: 'upperBox',
+            type: continuousAxisChannelDef.type
+          },
+          [continuousAxis + '2']: {
+            field: 'upperWhisker',
+            type: continuousAxisChannelDef.type
+          },
           ...nonPositionEncodingWithoutColorSize
         }
       }, { // box (q1 to q3)
@@ -144,8 +143,14 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT 
         },
         encoding: {
           ...discreteAxisEncodingMixin,
-          [continuousAxis]: q1FieldDef,
-          [continuousAxis + '2']: q3FieldDef,
+          [continuousAxis]: {
+            field: 'lowerBox',
+            type: continuousAxisChannelDef.type
+          },
+          [continuousAxis + '2']: {
+            field: 'upperBox',
+            type: continuousAxisChannelDef.type
+          },
           ...nonPositionEncodingWithoutSize,
           ...midTickAndBarSizeChannelDef
         }
@@ -156,7 +161,10 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT 
         },
         encoding: {
           ...discreteAxisEncodingMixin,
-          [continuousAxis]: medianFieldDef,
+          [continuousAxis]: {
+            field: 'midBox',
+            type: continuousAxisChannelDef.type
+          },
           ...nonPositionEncoding,
           ...midTickAndBarSizeChannelDef,
           color: {value : 'white'}
@@ -248,4 +256,86 @@ export function box2DParams(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | Box
     discreteAxis: discreteAxis,
     continuousAxis: continuousAxis
   };
+}
+
+export function boxTransform(encoding: Encoding<Field>, discreteAxisFieldDef: PositionFieldDef<Field>, continuousAxisChannelDef: PositionFieldDef<Field>, kIQRScalar: 'min-max' | number, is1D: boolean) {
+  const isMinMax = kIQRScalar === undefined;
+  const groupbyDef = boxTransformGroupby(discreteAxisFieldDef, encoding);
+
+  let transformDef:any = [
+      {
+        summarize: [
+          {
+            aggregate: 'q1',
+            field: continuousAxisChannelDef.field,
+            as: 'lowerBox'
+          },
+          {
+            aggregate: 'q3',
+            field: continuousAxisChannelDef.field,
+            as: 'upperBox'
+          },
+          {
+            aggregate: 'median',
+            field: continuousAxisChannelDef.field,
+            as: 'midBox'
+          }
+        ]
+      }
+  ];
+
+  if (isMinMax) {
+    transformDef[0].summarize = transformDef[0].summarize.concat([
+      {
+        aggregate: 'min',
+        field: continuousAxisChannelDef.field,
+        as: 'lowerWhisker'
+      },
+      {
+        aggregate: 'max',
+        field: continuousAxisChannelDef.field,
+        as: 'upperWhisker'
+      }
+    ]);
+  } else {
+    transformDef = transformDef.concat([
+      {
+        calculate: 'datum.upperBox - datum.lowerBox',
+        as: 'IQR'
+      },
+      {
+        calculate: 'datum.lowerBox - datum.IQR * ' + kIQRScalar,
+        as: 'lowerWhisker'
+      },
+      {
+        calculate: 'datum.upperBox + datum.IQR * ' + kIQRScalar,
+        as: 'lowerWhisker'
+      }
+    ]);
+  }
+
+  if (groupbyDef !== undefined) {
+    transformDef[0].groupby = groupbyDef;
+  }
+
+  return transformDef;
+}
+
+export function boxTransformGroupby(discreteAxisFieldDef: PositionFieldDef<Field>, encoding: Encoding<Field>): Array<Field | string> {
+  const result: Array<Field | string> = [];
+  if (discreteAxisFieldDef !== undefined) {
+    result.push(discreteAxisFieldDef.field);
+  }
+  const supportedEncChannels = ['size', 'color', 'opacity'];
+
+  for (const fieldName in encoding) {
+    if (supportedEncChannels.indexOf(fieldName) > -1) {
+      const fieldDef = encoding[fieldName];
+      if (fieldDef.field && fieldDef.aggregate !== BOXPLOT) {
+        result.push(fieldName);
+      }
+    }
+  }
+
+  return result;
 }
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -63,31 +63,17 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT 
   let discreteAxisFieldDef: PositionFieldDef<Field>, continuousAxisChannelDef: PositionFieldDef<Field>;
   let discreteAxis, continuousAxis;
 
-  let is1D = true;
-
-  if (encoding.x && encoding.y) {
-    // 2D
-
-    is1D = false;
-
-    const orient: Orient = box2DOrient(spec);
-    const params = box2DParams(spec, orient);
-    discreteAxisFieldDef = params.discreteAxisFieldDef;
-    continuousAxisChannelDef = params.continuousAxisChannelDef;
-    discreteAxis = params.discreteAxis;
-    continuousAxis = params.continuousAxis;
-
-  } else if (isFieldDef(encoding.x) && isContinuous(encoding.x) && encoding.y === undefined) {
-    // 1D horizontal
-    continuousAxis = 'x';
-    continuousAxisChannelDef = encoding.x;
-  } else if (encoding.x === undefined && isFieldDef(encoding.y) && isContinuous(encoding.y)) {
-    // 1D vertical
-    continuousAxis = 'y';
-    continuousAxisChannelDef = encoding.y;
-  } else {
-    throw new Error('Need a continuous axis for 1D boxplots');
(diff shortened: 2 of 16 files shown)


src/compile/data/assemble.ts
```diff
@@ -200,6 +200,12 @@ function makeWalkTree(data: VgData[]) {
       dataSource.transform = dataSource.transform.concat(node.assemble());
     }
 
+    if (node instanceof AggregateNode) {
```
kanitw: Can you explain why we need this chunk of changes?

src/compositemark/boxplot.ts
```diff
@@ -48,71 +48,44 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
   const {color: _color, ...nonPositionEncodingWithoutColorSize} = nonPositionEncodingWithoutSize;
   const midTickAndBarSizeChannelDef = size ? {size: size} : {size: {value: config.box.size}};
 
-  let discreteAxisFieldDef;
+  let kIQRScalar: number = undefined;
+  if (isBoxPlotDef(mark)) {
+    if (mark && mark.extent) {
+      if(typeof(mark.extent) === 'number') {
```
kanitw: We have `isNumber` in util -- let's consistenly use it :)

src/compositemark/boxplot.ts
```diff
@@ -48,71 +48,44 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
   const {color: _color, ...nonPositionEncodingWithoutColorSize} = nonPositionEncodingWithoutSize;
   const midTickAndBarSizeChannelDef = size ? {size: size} : {size: {value: config.box.size}};
 
-  let discreteAxisFieldDef;
+  let kIQRScalar: number = undefined;
+  if (isBoxPlotDef(mark)) {
+    if (mark && mark.extent) {
```
kanitw: why `mark &&`?   (can really `isBoxPlotDef(mark)` be true if `mark` is falsy?)

src/compositemark/boxplot.ts
```diff
@@ -48,71 +48,44 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
-    continuousAxis = 'y';
-    continuousAxisChannelDef = encoding.y;
-  } else {
-    throw new Error('Need a continuous axis for 1D boxplots');
+  if (encoding.x === undefined && encoding.y === undefined) {
+    throw new Error('Need at least one axis');
   }
 
+  const orient: Orient = boxOrient(spec);
+  const params = boxParams(spec, orient);
```
kanitw: use `const {a,b,c,d} = boxParams(...)` syntax?

kanitw: (Right now you declare these four params above, but they don't really get reassign anymore.)

src/compositemark/boxplot.ts
```diff
@@ -48,71 +48,44 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
-  };
-  const q3FieldDef = {
-    aggregate: 'q3',
-    ...baseContinuousFieldDef
-  };
-  const maxFieldDef = {
-    aggregate: 'max',
-    ...baseContinuousFieldDef
-  };
+  const transformDef = boxTransform(encoding, discreteAxisFieldDef, continuousAxisChannelDef, kIQRScalar, is1D);
```
kanitw: transformDef => transform?

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
-export function box2DOrient(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | BoxPlotDef>): Orient {
+export function boxOrient(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | BoxPlotDef>): Orient {
   const {mark: mark, encoding: encoding, ...outerSpec} = spec;
 
-  // FIXME: refactor code such that we don't have to do this casting
-  // We can cast here as we already check from outside that both x and y are FieldDef
-  const xDef = encoding.x as FieldDef<Field>;
-  const yDef = encoding.y as FieldDef<Field>;
-  let resultOrient: Orient;
+  const is2D = encoding.x && encoding.y;
```
kanitw: I think the boxOrient method here is a bit overly complicated. 

In a way, there are 3 main cases. 

1) One axis is continuous and another is not (either discrete or empty).  
2) Both axes are continuous
3) None of the axes are continuous

kanitw: Instead of branching all sub-cases, maybe set up logic in a way that we can easier group things?  This would be a good exercise for you anyway.

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
-            // default orientation = vertical
+        if (isContinuous(encoding.x) && isContinuous(encoding.y)) {
+          if (encoding.x.aggregate === undefined && encoding.y.aggregate === BOXPLOT) {
             resultOrient = 'vertical';
+          } else if (encoding.y.aggregate === undefined && encoding.x.aggregate === BOXPLOT) {
+            resultOrient = 'horizontal';
+          } else if (encoding.x.aggregate === BOXPLOT && encoding.y.aggregate === BOXPLOT) {
+            throw new Error('Both x and y cannot have aggregate');
+          } else {
+            if (isBoxPlotDef(mark)) {
```
kanitw: two redundant "else = vertical" here

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
   let continuousAxisChannelDef: PositionFieldDef<Field>;
   let discreteAxis;
   let continuousAxis;
 
-  // FIXME: refactor code such that we don't have to do this casting
-  // We can cast here as we already check from outside that both x and y are FieldDef
-  const xDef = encoding.x as FieldDef<Field>;
-  const yDef = encoding.y as FieldDef<Field>;
-
+  const is2D = encoding.x && encoding.y;
```
kanitw: if x or y is a ValueDef, it doesn't count as true 2D.

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
 
-  if (orient === 'vertical') {
-    discreteAxis = 'x';
-    continuousAxis = 'y';
-    continuousAxisChannelDef = yDef;
-    discreteAxisFieldDef = xDef;
+  if (!is2D) {
```
kanitw: I bet you can have shorter code by checking orientation first and check if your discrete axis is a fieldDef or not.

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
+    if (orient === 'horizontal') {
+      continuousAxis = 'x';
+      if (isFieldDef(encoding.x)) {
```
kanitw: Can this really be false?   If it's not a field def, it shouldn't even be a continousAxis from the beginning.

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
+        continuousAxisChannelDef = encoding.x;
+      }
+    } else {
+      continuousAxis = 'y';
+      if (isFieldDef(encoding.y)) {
```
kanitw: Can this really be false?   If it's not a field def, it shouldn't even be a continousAxis from the beginning.

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
     discreteAxis: discreteAxis,
-    continuousAxis: continuousAxis
+    continuousAxis: continuousAxis,
+    is1D: !is2D
   };
 }
+
+export function boxTransform(encoding: Encoding<Field>, discreteAxisFieldDef: PositionFieldDef<Field>, continuousAxisChannelDef: PositionFieldDef<Field>, kIQRScalar: 'min-max' | number, is1D: boolean) {
+  const isMinMax = kIQRScalar === undefined;
+  const groupbyDef = boxTransformGroupby(discreteAxisFieldDef, encoding);
```
kanitw: Since `groupbyDef` is never used in the chunk below, maybe moving it to right above where you assign it would make the code easier to read. 

Also, rename to `groupBy`.

src/compositemark/boxplot.ts
```diff
@@ -48,71 +48,44 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
-  } else {
-    throw new Error('Need a continuous axis for 1D boxplots');
+  if (encoding.x === undefined && encoding.y === undefined) {
+    throw new Error('Need at least one axis');
   }
 
+  const orient: Orient = boxOrient(spec);
+  const params = boxParams(spec, orient);
+  discreteAxisFieldDef = params.discreteAxisFieldDef;
+  continuousAxisChannelDef = params.continuousAxisChannelDef;
```
kanitw: Why is one of them `...ChannelDef`, but another is `...FieldDef`?  
Let's use "ChannelDef" for both?

src/compositemark/boxplot.ts
kanitw: Also looks like this method totally ignore discreteAxisChannelDef if it is a `ValueDef`.  After simplifying your logic tree, you should also handle this (which should be way easier after you simplify the logic tree.)

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
+
+  return transformDef;
+}
+
+export function boxTransformGroupby(discreteAxisFieldDef: PositionFieldDef<Field>, encoding: Encoding<Field>): Array<Field | string> {
+  const result: Array<Field | string> = [];
+  if (discreteAxisFieldDef !== undefined) {
+    result.push(discreteAxisFieldDef.field);
+  }
+  const supportedEncChannels = ['size', 'color', 'opacity'];
```
kanitw: Why only these 3 channels?   `detail` is definitely missing.

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
+
+  if (groupbyDef !== undefined) {
+    transformDef[0].groupby = groupbyDef;
+  }
+
+  return transformDef;
+}
+
+export function boxTransformGroupby(discreteAxisFieldDef: PositionFieldDef<Field>, encoding: Encoding<Field>): Array<Field | string> {
+  const result: Array<Field | string> = [];
```
kanitw: `result` => `groupBy`

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
+  if (discreteAxisFieldDef !== undefined) {
+    result.push(discreteAxisFieldDef.field);
+  }
+  const supportedEncChannels = ['size', 'color', 'opacity'];
+
+  for (const fieldName in encoding) {
+    if (supportedEncChannels.indexOf(fieldName) > -1) {
```
kanitw: I think this is the wrong place to excluding unsupported field.  

You exclude fields from your aggregate calculation if you still output these fields to the `encoding`.

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
+      const fieldDef = encoding[fieldName];
+      if (fieldDef.field && fieldDef.aggregate !== BOXPLOT) {
```
kanitw: Why `fieldDef.aggregate !== BOXPLOT`?  What if color shows mean value? -- you shouldn't add it to a groupby anyway.

src/compositemark/boxplot.ts
```diff
@@ -165,86 +160,193 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
+  }
+
+  if (groupbyDef !== undefined) {
+    transformDef[0].groupby = groupbyDef;
+  }
+
+  return transformDef;
+}
+
+export function boxTransformGroupby(discreteAxisFieldDef: PositionFieldDef<Field>, encoding: Encoding<Field>): Array<Field | string> {
```
kanitw: This method only handle groupBy but doesn't add additional nodes that need more aggregation.  (For example, if color/size encodes aggregate value, you still need to add them to `summarize`.)

src/compositemark/boxplot.ts
kanitw: (We mostly use `Def` to field/channelDef)

kanitw: It's also nice to generally keep things short anyway.

src/compile/data/assemble.ts
mattwchun: Yes I can. We need it because if we do not then the vega spec has a null name for one of the data objects but the vega-schema.json only allows for string names.

src/compositemark/boxplot.ts
```diff
@@ -165,72 +153,71 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
+  } else if (continuousYAxis
+      && (encoding.x === undefined || (isFieldDef(encoding.x) && isDiscrete(encoding.x)))) {
+    return 'vertical';
+  }
+
+  if (!continuousXAxis && !continuousYAxis) {
+    throw new Error('Need a valid continuous axis for boxplots');
+  }
+
+  if (isFieldDef(encoding.x) && isFieldDef(encoding.y) && isContinuous(encoding.x) && isContinuous(encoding.y)) {
```
mattwchun: @kanitw Only reason I didn't use continuousXAxis and continuousYAxis is because of that typeguard issue I asked you about. encoding.x and encoding.y aren't typed as FieldDef's and I need to access their aggregate property inside these if blocks

kanitw:
I think you should always look at your overall diff after changing --all examples are now being removed and somehow test are still passing.

src/compositemark/boxplot.ts
kanitw: Thanks for the note.

One possible minor trick is to declare a function (maybe locally)

```
isContinuousFieldDef(channelDef): channelDef is FieldDef
```

:)   This way your code would still be shorter.

src/compositemark/boxplot.ts
```diff
@@ -165,72 +153,71 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
-  const xDef = encoding.x as FieldDef<Field>;
-  const yDef = encoding.y as FieldDef<Field>;
-
-
   if (orient === 'vertical') {
-    discreteAxis = 'x';
-    continuousAxis = 'y';
-    continuousAxisChannelDef = yDef;
-    discreteAxisFieldDef = xDef;
+    if (isFieldDef(encoding.y)) {
```
kanitw: if it's vertical, it seems impossible that y is not a fieldDef?

src/compositemark/boxplot.ts
```diff
@@ -165,72 +153,71 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT
...
+    if (isFieldDef(encoding.x)) {
+      discreteAxis = 'x';
+      discreteAxisChannelDef = encoding.x;
+    }
   } else {
-    discreteAxis = 'y';
-    continuousAxis = 'x';
-    continuousAxisChannelDef = xDef;
-    discreteAxisFieldDef = yDef;
+    if (isFieldDef(encoding.x)) {
```
kanitw: if it's horizontal, it seems impossible that x is not a fieldDef?

src/compositemark/boxplot.ts
```diff
@@ -242,9 +229,101 @@ export function box2DParams(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | Box
...
+  }
+
+  for (const fieldName in encoding) {
+    if (supportedEncChannels.indexOf(fieldName) > -1) {
+      const fieldDef = encoding[fieldName];
+      if (fieldDef.field && fieldDef.aggregate !== BOXPLOT) {
+        transformDef[0].summarize = transformDef[0].summarize.push({
+          aggregate: fieldDef.aggregate,
+          field: fieldDef.field,
+          as: 'fieldName_' + fieldDef.field
```
kanitw: why `'fieldName_'+ fieldDef.field`? 

Just in case you don't know the right method -- it should be `field(fieldDef)`.

In addition, you will have to modify the encoding to remove `aggregate` and instead use `field: field(fieldDef)`.
(Your data is already aggregated, thus channelDef should not have `aggregate` anymore.)

Also, make sure that
1) You do not cause side effect to the original input spec
2) [derivatives of `encoding`](https://github.com/vega/vega-lite/pull/2485/files#diff-d57f23a9ad71355a4febffed6b676f84R48) get updated encoding without aggregate.

I would suggest that you create an example with color encoding mean and make sure it renders correct in the vega-editor and include it in the `examples/`

src/compositemark/boxplot.ts
```diff
@@ -242,9 +229,101 @@ export function box2DParams(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | Box
...
+}
+
+export function boxTransformGroupby(discreteAxisFieldDef: PositionFieldDef<Field>, encoding: Encoding<Field>): Array<Field | string> {
+  const groupby: Array<Field | string> = [];
+  if (discreteAxisFieldDef !== undefined) {
+    groupby.push(discreteAxisFieldDef.field);
+  }
+
+  for (const fieldName in encoding) {
+    if (supportedEncChannels.indexOf(fieldName) > -1) {
```
kanitw: As mentioned earlier, you cannot filter encoding channels that you do not want to support only within the transform. You have to do that to `encoding` before you input them here.

src/compositemark/boxplot.ts
```diff
@@ -242,9 +229,101 @@ export function box2DParams(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | Box
...
+      },
+      {
+        calculate: 'datum.upperBox + datum.IQR * ' + kIQRScalar,
+        as: 'lowerWhisker'
+      }
+    ]);
+  }
+
+  for (const fieldName in encoding) {
+    if (supportedEncChannels.indexOf(fieldName) > -1) {
```
kanitw: As mentioned earlier, you cannot filter encoding channels that you do not want to support only within the transform. You have to do that to `encoding` before you input them here.

src/compositemark/boxplot.ts
```diff
@@ -242,9 +229,101 @@ export function box2DParams(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | Box
...
+  return transformDef;
+}
+
+export function boxTransformGroupby(discreteAxisFieldDef: PositionFieldDef<Field>, encoding: Encoding<Field>): Array<Field | string> {
+  const groupby: Array<Field | string> = [];
+  if (discreteAxisFieldDef !== undefined) {
+    groupby.push(discreteAxisFieldDef.field);
+  }
+
+  for (const fieldName in encoding) {
```
kanitw: This loop is redundant with the one that add to summarize why don't you combine them and just get rid of this `boxTransformGroupby` method.

src/compositemark/boxplot.ts
mattwchun: @kanitw yes I have that only because I need to typeguard encoding.y to ensure it is of type FieldDef<Field> which I assign to continuousAxisChannelDef

kanitw:
@mattwchun -- FYI I just rebase, drop the example change PR and force pushed for you.

src/compositemark/boxplot.ts
```diff
@@ -42,78 +45,60 @@ export const VL_ONLY_BOXPLOT_CONFIG_PROPERTY_INDEX: {
...
-    ...baseContinuousFieldDef
-  };
-  const maxFieldDef = {
-    aggregate: 'max',
-    ...baseContinuousFieldDef
-  };
-
-  const discreteAxisEncodingMixin = discreteAxisFieldDef !== undefined ? {[discreteAxis]: discreteAxisFieldDef} : {};
+  const transform = boxTransform(encoding, discreteAxisChannelDef, continuousAxisChannelDef, kIQRScalar, is1D);
+  const {x: _xTemp, y: _yTemp, ...nonPositionEncoding2} = encoding;
```
mattwchun: @kanitw This and the next line are a "hack" because when I generate the summarize array I need to edit the encoding to use the new name of the field. I implemented another function to edit the encoding with the changes from the summarize array however the logic looked essentially the same as the end of boxTransform() so I didn't want redundant code.

src/compositemark/boxplot.ts
```diff
@@ -44,7 +45,25 @@ export const VL_ONLY_BOXPLOT_CONFIG_PROPERTY_INDEX: {
...
+    if (supportedEncChannels.indexOf(c) > -1) {
+      newNonPositionEncoding[c] = f;
+    } else {
+      log.warn(log.message.incompatibleChannel(c, BOXPLOT));
+    }
+  });
+  newNonPositionEncoding['x'] = x;
+  newNonPositionEncoding['y'] = y;
+
+  spec.encoding = newNonPositionEncoding;
```
kanitw: 1) This can cause side effect to the original `spec`.  Use `...` and return a new spec object.  
2) `newNonPositionEncoding`  is not non-position so this is a misleading variable name.

src/compositemark/boxplot.ts
kanitw: I don't think I understand this but this sounds super complicated and I don't like the fact that you claim that you have a "hack" here. 

1) why `_xTemp` -- is it temporary in any way -- it also does not seem used.  

2) What is `nonPositionEncoding2`?  Can you name it in a more descriptive way?  This will definitely confuse people.

kanitw: I think `nonPositionEncoding2` comes from the fact that you declare a lot of variables at the top but don't really use them. In fact, you should move logics related to encoding to after finishing with the `boxTransform`. 
(Get the transform, then modify the encoding to use transformed field, then extract required parts.)

You can also rename `boxTransform` and make it return both `encoding` and `transform` if you wanna reuse the logic. 

`{encoding, transform} = boxTransformAndEncoding(...)`

src/compositemark/boxplot.ts
kanitw: Thinking more about it. This may be a place where we can do type casting 

(encoding.y as FieldDef<string>) and explain why it is safe to do so.

kanitw:
I think you also need to re-run `build:examples` as other code might change your specs after rebasing :)

src/compositemark/boxplot.ts
```diff
@@ -243,9 +249,99 @@ export function box2DParams(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | Box
...
+  }
+
+  const groupby: Array<Field | string> = [];
+  if (discreteAxisFieldDef !== undefined) {
+    groupby.push(discreteAxisFieldDef.field);
+  }
+
+  const {x: _x, y: _y, ...nonPositionEncoding} = encoding;
+
+  for (const fieldName in nonPositionEncoding) {
```
kanitw: `fieldName` => `channel`

src/compositemark/boxplot.ts
```diff
@@ -243,9 +249,99 @@ export function box2DParams(spec: GenericUnitSpec<Encoding<Field>, BOXPLOT | Box
...
+    if (nonPositionEncoding.hasOwnProperty(fieldName)) {
+      const fieldDef = nonPositionEncoding[fieldName];
```
kanitw: how do you know if this is a fieldDef, not valueDef?

src/compile/data/assemble.ts
domoritz: Hmm, I just came across this code and still don't really get it. Why is this the right fix?

mattwchun: Still don't remember why I added this chunk 1+ years ago but ran tests with and without the chunk and all pass.

domoritz: Thank you so much for looking into this. If we don't need this, let's remove it. Can you send a PR?

mattwchun: #4178

PR #2487, opened 2017-06-11
Fix #2486 Rename LogicalOperand/And/Or/Not to Selection*

### LINKED ISSUE #2486: Rename LogicalOperand/And/Or/Not to Selection*
Similar to FilterOperand/And/Or/Not, we should make the name more specific.

kanitw:
(Since this is only used for selection, this should be a better name for Altair)

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1075,7 +1075,7 @@
           "$ref": "#/definitions/Scale"
         },
         "selection": {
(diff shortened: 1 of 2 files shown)


kanitw:
>  So while we currently only use them for selections, they seem more broadly applicable for future use cases that may arise.

I agree.  We can later rename if such use case arise.  Given we probably won't have one in the near future, I think Selection* is better for now.  :)

Issue #2198, opened 2017-04-17
Logic for merging size signal


kanitw:
Currently we always generate redundant size signal (e.g.,  for line that have point overlay, we will generate similar signals for both layers and the size of layer simply calculate max of all signals.)

This obviously could be improved. 

Cases:

- [x] Merge redundant signal inside 
  - [x] layer
  - [x] repeat
  - [ ] concat
- [ ] Merge the top-level's width/height signal with the top-level model

This should depends on correctly merging scale

kanitw:
The tricky part is  1D plot (when there is no x scale or no y scale). 
We may need size component that can be merged.

PR #2491, opened 2017-06-12
Add Split<Scale/ScaleComponent> and uses them as return types of `initScale, parseScale`	

kanitw:
This is a first step for scale refactor #2251.

With more details about explicitly specified properties and one that are from our rules, we can improve scale merging logic.  

cc: @willium

### DIFF
--- a/examples/vg-specs/layer_precipitation_mean.vg.json
+++ b/examples/vg-specs/layer_precipitation_mean.vg.json
@@ -105,7 +105,7 @@
         },
         {
             "name": "layer_1_width",
-            "update": "21"
+            "update": "bandspace(domain('undefined').length, 0.1, 0.05) * 21"
         },
         {
             "name": "layer_1_height",
--- a/src/compile/scale/parse.ts
+++ b/src/compile/scale/parse.ts
@@ -47,25 +47,28 @@ export function parseScale(model: UnitModel, channel: ScaleChannel) {
   const scale = model.scale(channel);
   const sort = model.sort(channel);
 
+  // FIXME parseScale should return something similar to Split<VgScale>
   const scaleComponent: VgScale = {
     name: model.scaleName(channel + '', true),
-    type: scale.type,
+    type: scale.get('type'),
     domain: parseDomain(model, channel),
     range: parseRange(scale)
   };
 
-  if (isSelectionDomain(scale.domain)) {
+  if (isSelectionDomain(scale.get('domain'))) {
     // As scale parsing occurs before selection parsing, we use a temporary
     // signal here and append the scale.domain definition. This is replaced
     // with the correct domainRaw signal during scale assembly.
     // For more information, see isRawSelectionDomain in selection.ts.
+
+    // FIXME: replace this with an explicit property in the scaleComponent
     scaleComponent.domainRaw = {
-      signal: SELECTION_DOMAIN + JSON.stringify(scale.domain)
+      signal: SELECTION_DOMAIN + JSON.stringify(scale.get('domain'))
     };
   }
 
   NON_TYPE_DOMAIN_RANGE_VEGA_SCALE_PROPERTIES.forEach((property) => {
-    scaleComponent[property] = scale[property];
+    scaleComponent[property] = scale.get(property);
   });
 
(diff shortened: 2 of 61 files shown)


examples/vg-specs/layer_precipitation_mean.vg.json
```diff
@@ -105,7 +105,7 @@
         },
         {
             "name": "layer_1_width",
-            "update": "21"
+            "update": "bandspace(domain('undefined').length, 0.1, 0.05) * 21"
```
kanitw: Note that this one actually is not perfectly correct.  
The correct behavior is that the scale should be merged (and should be handled as a part of #2198).  

In any case, the visual output looks fine for now away.

kanitw:
Note that a number of layered examples now have consistent width/height signals

src/compile/split.ts
```diff
@@ -0,0 +1,71 @@
...
+    const base = keys.reduce((b, key) => {
+      const value = this.get(key);
+      if (value) {
+        b[key] = value;
+      }
+      return b;
+    }, {} as Partial<T>);
+
+    // FIXME remove "as any".
+    // Add "as any" to avoid an error "Spread types may only be created from object types".
```
domoritz: Can you do `T extends object`?

src/compile/scale/parse.ts
```diff
@@ -47,25 +48,37 @@ export function parseScale(model: UnitModel, channel: ScaleChannel) {
   const scale = model.scale(channel);
   const sort = model.sort(channel);
 
-  const scaleComponent: VgScale = {
-    name: model.scaleName(channel + '', true),
-    type: scale.type,
-    domain: parseDomain(model, channel),
-    range: parseRange(scale)
-  };
+  // FIXME parseScale should return something similar to Split<VgScale>
```
domoritz: Can you explain what this means? Is there an issue for this?

src/compile/scale/parse.ts
```diff
@@ -47,25 +48,37 @@ export function parseScale(model: UnitModel, channel: ScaleChannel) {
...
+  const scaleComponent: ScaleComponent = new Split<Partial<VgScale>>(
+    {},
+    // Implicit
+    {
+      name: model.scaleName(channel + '', true)
+    }
+  );
+  scaleComponent.copyKeyFrom('type', scale);
+
+  // FIXME: this condition to be implicit/explicit is incorrect
```
domoritz: Why not fix it?

src/compile/scale/parse.ts
```diff
@@ -47,25 +48,37 @@ export function parseScale(model: UnitModel, channel: ScaleChannel) {
...
+  if (isSelectionDomain(scale.get('domain'))) {
     // As scale parsing occurs before selection parsing, we use a temporary
     // signal here and append the scale.domain definition. This is replaced
     // with the correct domainRaw signal during scale assembly.
     // For more information, see isRawSelectionDomain in selection.ts.
-    scaleComponent.domainRaw = {
-      signal: SELECTION_DOMAIN + JSON.stringify(scale.domain)
-    };
+
+    // FIXME: replace this with a special property in the scaleComponent
```
domoritz: Why not do it now?

kanitw: So many things to do.  This is less important and could be done post 2.0.

src/compile/split.ts
kanitw: Unfortunately, I tried and it threw a different error.

src/compile/scale/parse.ts
kanitw: Good catch.  This is actually done  in the last two commits of this PR.  I'll remove the comment.

src/compile/scale/parse.ts
kanitw: Because it requires refactoring parseDomain, which would become hairy ... I wanna work on more important parts first.  Filed an issue here https://github.com/vega/vega-lite/issues/2497

src/compile/split.ts
domoritz: What was it?

domoritz:
looks good to be merged.

PR #2498, opened 2017-06-13
Dom/falkensee

domoritz:


src/compile/data/index.ts
```diff
@@ -29,4 +29,9 @@ export interface DataComponent {
    * For facets, we store the reference to the root node.
    */
   facetRoot?: FacetNode;
+
+  /**
+   * Parse properties passed down from ancestors.
+   */
+  modelParse: Dict<string>;
```
kanitw: modelParse is a pretty confusing name -- it gives me no clue what it means when I first read the code.   maybe `ancestorParse`?

kanitw:
Also, I think writing test for this would only take your extra 5-10 mins so please do it.

src/compile/data/index.ts
domoritz: Done

domoritz:
Wrote a test

kanitw:
Cool.  Feel free to merge when test finishes running.

Issue #2493, opened 2017-06-13
Check occurences of isBinScale


kanitw:
Since the rationale for https://github.com/vega/vega-lite/pull/2492 is that bin scale does not work with selection, we should revise `isBinScale` calls inside `src/compiler/selections`.  (There are still some references.)

domoritz:
I looked over them and it seemed like they were there to prevent selections over bin scales. So it all seems correct.

However, I wonder whether those checks are necessary since we will never have bin scales for x and y.

arvind:
Yes! Thanks to #2492, interval selections now work when x/y is binned. However, I would prefer to leave the `isBinScale` checks in place because the conditions they test for still hold -- our strategy for panning, zooming, and reacting to scale changes will not work with binned scales. Keeping the checks in place seems more future proof to me -- e.g., if we enable brushing a binned color legend in the future.

kanitw:
Sounds good.  (I only filed this issue to make sure they are still correct :))

arvind:
Agreed, it was good to re-check those assumptions. And as a result, I found two bugs: I inadvertently deleted a ternary branch for the `isBinScale` condition and, more problematically, #2492 broke the cross filter example revealing that it is currently possible to specify an infinite loop with our selections. Need to think on how to resolve that.

kanitw:
Ok i'll re-open this for now then.

PR #2492, opened 2017-06-13
Use linear for binned x and y.

domoritz:
Fixes #2461

PR #2501, opened 2017-06-14
Proxy scale reactions to prevent infinite loop w/interval selection filters

arvind:
As described in #2493, the switch to a linear scale in #2492 revealed an infinite loop when interval selections filter datasets that back scales in its view (e.g., cross filtering). The specific flow is interaction events fire `modify` calls to the selection's backing store. These updates trigger reevaluations of the selection predicate applied to the source/raw dataset. This in turn marks scale functions as modified, which would loop back to keep brushes in sync with scale changes. 

This PR instead proxies the final part of that cycle. Instead of having the brush signals react directly to scale changes, a new `scale_trigger` signal is introduced. This new signal reacts to scale changes and determines if the brush coordinates should change. If they should, propagation continues and the brush coordinates update (as they now listen to the `scale_trigger` signal rather than the actual scale). If brush coordinates remain the same, propagation is stopped by returning an unchanged `scale_trigger` value.

### DIFF
--- a/src/compile/selection/interval.ts
+++ b/src/compile/selection/interval.ts
@@ -9,17 +9,20 @@ import scales from './transforms/scales';
 
 
 export const BRUSH = '_brush';
+export const SCALE_TRIGGER = '_scale_trigger';
 
 const interval:SelectionCompiler = {
   predicate: 'vlInterval',
   scaleDomain: 'vlIntervalDomain',
 
   signals: function(model, selCmpt) {
+    const name = selCmpt.name;
+    const hasScales = scales.has(selCmpt);
     const signals: any[] = [];
     const intervals:any[] = [];
-    const name = selCmpt.name;
+    const scaleTriggers: any[] = [];
 
-    if (selCmpt.translate && !(scales.has(selCmpt))) {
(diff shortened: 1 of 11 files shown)


arvind:
Travis should pass once the schema change in vega/vega-parser#40 is merged.

src/compile/selection/interval.ts
```diff
@@ -29,18 +32,43 @@ const interval:SelectionCompiler = {
...
     });
 
+    // Proxy scale reactions to ensure that an infinite loop doesn't occur
+    // when an interval selection filter touches the scale.
+    if (!hasScales) {
+      signals.push({
+        name: name + SCALE_TRIGGER,
+        value: {},
+        on: [{
+          events: scaleTriggers.map((t) => { return {scale: t.scaleName}; }),
```
kanitw: No need to change at all but FYI you can do 

```
scaleTriggers.map(t => ({scale: t.scaleName}))
```

(The trick for returning object is to use a parenthesis to wrap it.)

arvind: Aha! This always frustrated me. Thanks!

Issue #2479, opened 2017-06-08
Expose more axis label formatting options


nyurik:
Customizing label rotation is not very useful unless it either smartly adjusts the alignment/dx/..., or unless it exposes all the core Vega label placement options.  The later is probably easier.

![image](https://user-images.githubusercontent.com/1641515/26911802-ee2ca424-4bdd-11e7-9709-acf7260168bf.png)

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "values": [
      {"a": "2001-01-01", "b": 28, "c": "P"},
      {"a": "2001-01-02", "b": 95, "c": "Q"},
      {"a": "2001-01-03", "b": 43, "c": "R"},
      {"a": "2001-01-04", "b": 91, "c": "Q"},
      {"a": "2001-01-05", "b": 81, "c": "P"},
      {"a": "2001-01-06", "b": 53, "c": "P"},
      {"a": "2001-01-07", "b": 19, "c": "R"},
      {"a": "2001-01-08", "b": 87, "c": "Q"},
      {"a": "2001-01-09", "b": 52, "c": "P"},
      {"a": "2001-01-10", "b": 81, "c": "Q"},
      {"a": "2001-01-11", "b": 53, "c": "R"},
      {"a": "2001-01-12", "b": 19, "c": "P"},
      {"a": "2001-01-13", "b": 87, "c": "Q"},
      {"a": "2001-01-14", "b": 52, "c": "R"}
    ]
  },
  "mark": {"type": "line", "interpolate": "natural"},
  "encoding": {
    "x": {
      "field": "a",
      "type": "temporal",
      "axis": {
        "domain": false,
        "format": "%b %d",
        "labelAngle": -90,
        "labelPadding": 20,
        "tickExtra": true
      }
    },
    "y": {"field": "b", "type": "quantitative"},
    "color": {"field": "c", "type": "nominal", "scale": {"scheme": "reds"}}
  }
}
```

nyurik:
It seems that defining `labelAngle=-90` causes a strange label shift, even though the orientation of the label is the same as by default.

kanitw:
Thanks for reporting.  Previously we only automatically adjust  text's `baseline` when the angle is `270`.   https://github.com/vega/vega-lite/pull/2509 will make it works for -90 as well.  

For exposing other properties, see https://github.com/vega/vega-lite/pull/2508

Issue #2173, opened 2017-04-15
Revise Axis Labels rules (in encode.ts)


kanitw:
- [x] align
- [x] baseline

PR #2509, opened 2017-06-15
Revise Axis Label Rules

kanitw:
Fixes https://github.com/vega/vega-lite/issues/2479, #2173


kanitw:
@domoritz -- okay I have checked this in the editor and it is correct. It is also not conflicting with the format PR.

PR #2510, opened 2017-06-15
Dom/format

domoritz:
Fix https://github.com/vega/vega-lite/issues/1539

### DIFF
--- a/src/compile/axis/parse.ts
+++ b/src/compile/axis/parse.ts
@@ -118,6 +118,8 @@ function getSpecifiedOrDefaultValue(property: keyof VgAxis, specifiedAxis: Axis,
   switch (property) {
     case 'labels':
       return isGridAxis ? false : specifiedAxis[property];
+    case 'labelOverlap':
+      return rules.labelOverlap(fieldDef, specifiedAxis, channel, isGridAxis);  // TODO: scaleType
     case 'domain':
       return rules.domain(property, specifiedAxis, isGridAxis, channel);
     case 'ticks':
--- a/src/compile/axis/parse.ts
+++ b/src/compile/axis/parse.ts
@@ -123,7 +123,7 @@ function getSpecifiedOrDefaultValue(property: keyof VgAxis, specifiedAxis: Axis,
     case 'ticks':
       return rules.ticks(property, specifiedAxis, isGridAxis, channel);
     case 'format':
-      return rules.format(specifiedAxis, channel, fieldDef, model.config);
+      return rules.format(specifiedAxis, fieldDef, model.config);
     case 'grid':
       return rules.grid(model, channel, isGridAxis); // FIXME: refactor this
     case 'gridScale':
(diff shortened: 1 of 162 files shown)


src/compile/common.ts
```diff
@@ -111,24 +110,29 @@ export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: str
...
     }
+
     // TODO: need to make this work correctly for numeric ordinal / nominal type
     return config.numberFormat;
   }
   return undefined;
 }
 
+function formatExpr(field: string, format: string) {
+  return `format(${field}, ${format ? `'${format}'` : null})`;
```
kanitw: So format `''`  is different from `null`?

src/compile/common.ts
```diff
@@ -111,24 +110,29 @@ export function formatSignalRef(fieldDef: FieldDef<string>, specifiedFormat: str
...
+}
+
+export function numberFormatExpr(field: string, specifiedFormat: string, config: Config) {
+  return formatExpr(field, specifiedFormat || config.numberFormat);
```
kanitw: '' || x = x, so this could be wrong?

src/compile/axis/parse.ts
```diff
@@ -120,6 +120,8 @@ function getSpecifiedOrDefaultValue(property: keyof VgAxis, specifiedAxis: Axis,
   switch (property) {
     case 'labels':
       return isGridAxis ? false : specifiedAxis[property];
+    case 'labelOverlap':
+      return rules.labelOverlap(fieldDef, specifiedAxis, channel, isGridAxis);  // TODO: scaleType
```
kanitw: This doesn't read the  specified `labelOverlap`

src/vega.schema.ts
```diff
@@ -335,6 +335,15 @@ export interface VgAxisBase {
   labelAngle?: number;
 
   /**
+   * Automatic overlap removal.
+   *
+   * `true` or `"parity"`: Reduce overlap by attempting to remove every other label, repeating this process until overlap is removed or only two labels remain. If only two labels remain, the first and last label will be shown. This setting is appropriate for linear, sqrt, and pow settings.
+   * `"greedy"`: Reduce overlap by sequentially scanning the labels, removing any items whose bounds overlap the last visible label. This strategy works better than "parity" for log scales, though is not "ideal" in that it does not actually reason about which ticks are the most useful for interpreting a log scale.
+   * __Default value:__ `true` for x-axes, otherwise `false`
```
kanitw: The default condition specified does not perfectly match what you implemented

src/compile/axis/parse.ts
kanitw: Also, you can easily getScaleComponent and read the type here to handle overlapping for discrete and log scales.

src/compile/common.ts
domoritz: idk, but https://github.com/vega/vega-lite/issues/1539#issuecomment-303504194 suggests that we should use null. It's a bit more explicit.

src/compile/common.ts
domoritz: Fixed

src/compile/common.ts
kanitw: ok

src/compile/axis/parse.ts
domoritz: Done

src/compile/common.ts
domoritz: Okay, I revise. I get an error now. So we need to set the format to ''.

domoritz:
I have no idea what's going wrong with the schema. Will look at it tomorrow.

domoritz:
Needs https://github.com/vega/vega-parser/pull/41

PR #2508, opened 2017-06-15
Define axis/legend.encoding for using Vega's axis/legend.encode

kanitw:
Part of #1874 

~~This doesn't work yet as the json schema generator doesn't support [Mapped Types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html) for the following lines yet:~~

```
+export type AxisEncoding = {
+  [k in keyof VgAxisEncode]?: GuideEncodingEntry;
+};
```

PR #2512, opened 2017-06-15
Upgrade schema gen

domoritz:
@kanitw Can you generate the yarn.lock file for me?


kanitw:
Fixed.  I also ensured we run yarn based on the old lock file.

domoritz:
I just pushed another update. We now support mapped types for https://github.com/vega/vega-lite/pull/2508

PR #2513, opened 2017-06-16
Correct getScaleComponent's signature

kanitw:
and update type accordingly

- Note that I change  `ProjectComponent`'s `encoding` to be only `ScaleChannel` as the rest of the code only support it for scale channel anyway.

### DIFF
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -363,14 +363,15 @@ export abstract class Model {
   }
 
   /**
-   * Traverse a model's hierarchy to get the specified component.
-   * @param type Scales or Selection
-   * @param name Name of the component
+   * Traverse a model's hierarchy to get the scale component for a particular channel.
    */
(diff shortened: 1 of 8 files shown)


src/compile/model.ts
```diff
@@ -363,14 +363,15 @@ export abstract class Model {
...
   /**
-   * Traverse a model's hierarchy to get the specified component.
-   * @param type Scales or Selection
-   * @param name Name of the component
+   * Traverse a model's hierarchy to get the scale component for a particular channel.
    */
-  public getScaleComponent(name: string): ScaleComponent {
-    return this.component.scales[name] || (this.parent ? this.parent.getScaleComponent(name) : undefined);
+  public getScaleComponent(channel: ScaleChannel): ScaleComponent {
+    return this.component.scales[channel] || (this.parent ? this.parent.getScaleComponent(channel) : undefined);
```
domoritz: 👍

PR #2514, opened 2017-06-16
Fix Travis Not Running Parallel

### LINKED ISSUE #2060: Parallelize the build:examples script

NeelMohapatra:
Fixes #2060

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,15 @@ language: node_js
 node_js:
   - "6"
   - "7"
+before_install:
+  - (wget -O - pi.dk/3 || curl pi.dk/3/ || fetch -o - http://pi.dk/3) | bash # To install latest parallel. apt-get installs outdated
 script:
-  - npm run build
-  - npm run build:examples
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,6 +2,8 @@ language: node_js
 node_js:
   - "6"
   - "7"
+before_install:
+  - brew upgrade
```
domoritz: brew is only available on mac os, travis is ubuntu

NeelMohapatra: Ah

kanitw:
Yay!  Thanks!

PR #2519, opened 2017-06-16
Test node 8, switch back to npm run from yarn run.

domoritz:


kanitw:
Can you explain why we need to switch back?

PR #2520, opened 2017-06-16
Make imports consistent. No newline between imports. Two lines after …

domoritz:
…the import block.

### DIFF
--- a/src/aggregate.ts
+++ b/src/aggregate.ts
@@ -1,4 +1,6 @@
 import {toSet} from './util';
+
+
 export type AggregateOp = 'argmax' | 'argmin' | 'average' | 'count'
   | 'distinct' | 'max' | 'mean' | 'median' | 'min' | 'missing' | 'modeskew'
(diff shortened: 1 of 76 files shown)


kanitw:
If there is no linter / auto formatter applied, this won't be preserved anyway?

domoritz:
No, but at least we have all imports in one block so we have fewer merge issues. We might get a linter but I'm waiting to hear back from a github issue. 

Can you merge it? Doesn't hurt to be more consistent.

kanitw:
Did u automatically generate this somehow?

This will cause merge conflict disaster for my giant pr - so i dont know why u choose to do it now

domoritz:
If you just do the same, it will be easy to do. Just remove all empty lines
between imports and you won't have any conflicts.

kanitw:
> If you just do the same, it will be easy to do. Just remove all empty lines
between imports and you won't have any conflicts.

That's not true because I will modify the same line as you -- anyway I'll merge it despite its bad timing

PR #2525, opened 2017-06-17
Rewrite parseScale

kanitw:
- Rewrite parseScale()
  - Instead of parsing the whole scale together, we divide `parseScale` by different properties of the scale.  This allows scale properties that depends on other scale properties (esp. scale type) to determine its value after the depended value have been merged. 
  - Nove all rules for scale properties to parse (except for `scale.type`, which should be moved too, but require a bit more refactor)
  - `parseScale` is now centralized in `compile/scale/parse.ts` and is called directed from the abstract `Model`
  - When merging scale, we no longer delete scale component from the child, but instead mark them as `merged`.  This allows us to know if we have to merge a scale property. 

- add `layoutSizeComponent` for storing size, which depend on scale
- add `parseLayoutSize` logic similar to `parseScale`

Fix https://github.com/vega/vega-lite/issues/2251

You might want to merge locally and push so we can easily merge the follow up PR.


examples/vg-specs/layer_box_plot_circle.vg.json
```diff
@@ -421,7 +426,7 @@
         },
         {
             "name": "y",
-            "type": "point",
+            "type": "band",
```
kanitw: FYI, when merging band and point, now I give higher precedence to band.

domoritz:
> This allows us to know if we have to merge a scale property.

How do we use this? Would it be enough to put a pointer to the merged scale or a tombstone (e.g. null)?

kanitw:
> > How do we use this? Would it be enough to put a pointer to the merged scale or a tombstone (e.g. null)?
> How do we use this? Would it be enough to put a pointer to the merged scale or a tombstone (e.g. null)?

It's also a placeholder for merging scale property up the tree.  So in a way, we need it.

src/compile/split.ts
```diff
@@ -29,7 +29,7 @@ export class Split<T extends Object> {
     return this.explicit[key] !== undefined ? this.explicit[key] : this.implicit[key];
   }
 
-  public getWithType<K extends keyof T>(key: K): {explicit: boolean, value: T[K]} {
+  public getWithExplicit<K extends keyof T>(key: K): Explicit<T[K]> {
```
domoritz: Why not just `getExplicit`?

src/compile/split.ts
```diff
@@ -74,3 +78,47 @@ export interface Explicit<T> {
...
+  } else {
+    const score1 = scoringFn ? scoringFn(v1.value) : 0;
+    const score2 = scoringFn ? scoringFn(v2.value) : 0;
+
+    if (score2 > score1) {
+      return v2;
+    } else if (score1 > score2) {
+      return v1;
+    } else {
+      // FIXME more warning
```
domoritz: This seems easy to fix. Please add a warning.

src/scale.ts
```diff
@@ -46,6 +46,62 @@ export const SCALE_TYPES: ScaleType[] = [
...
+  'bin-linear': 0,
+  sequential: 0,
+  ordinal: 0,
+  'bin-ordinal': 0,
+};
+
+/**
+ * Return scale categories -- only scale of the same categories can be merged together.
+ */
+export function scaleTypePrecedence(scaleType: ScaleType): number {
```
domoritz: This function seems excessive. Can we just export the index?

test/util.ts
```diff
@@ -31,6 +31,13 @@ export function parseUnitModelWithScale(spec: TopLevel<UnitSpec>) {
   return model;
 }
 
+export function parseUnitModelWithScaleAndLayoutSize(spec: TopLevel<UnitSpec>) {
```
domoritz: I'm not a huge fan of all of these test util function. I think tests should be as explicit as possible. But it's beyond the scope of this pr.

src/compile/split.ts
kanitw: get the particular value of the given key with its `explicit` metadata so I prefer `getWithExplicit`.

src/scale.ts
kanitw: I like this pattern better since you can't modify the index.

test/util.ts
kanitw: Another method would be making all parse support chaining -- however I am not sure if we can do that yet since we are not done with all merging stuff.  

In any case, these helpers method make it way easier to find and replace test methods so I won't waste my precious time refactoring them more.

kanitw:
cc: @willium  -- FYI I rewrote the parse scale to take into account of explicitness of each property.  This pattern is probably useful for your projection implementation.  (I'll apply the same trick to axis and legend in a later PR.)

willium:
Thanks for letting me know @kanitw. Hopefully I can get that in soon. Worried about future merge conflicts.

PR #2528, opened 2017-06-18
parseScale follow-up refactor

kanitw:
Follow up for https://github.com/vega/vega-lite/pull/2525, please merge https://github.com/vega/vega-lite/pull/2525 first!

PR #2530, opened 2017-06-18
Add Text Heatmap example

### LINKED ISSUE #1840: Reproduce spec for the old `applyColorToBackground`
It's a really hacky way to support text table heatmap!

Instead, text table heatmap should be a faceted layered plot with one rect layer and one text layer.

Note: the tricky part is to correctly merge the scale and layout.

kanitw:
Fix #1840

Merge #2528 first!


kanitw:
vega_editor

PR #2535, opened 2017-06-19
Add warning for scale merged.

kanitw:


examples/vg-specs/layer_text_heatmap.vg.json
```diff
@@ -225,6 +225,9 @@
                         "angle": {
                             "value": 270
                         },
+                        "align": {
+                            "value": "right"
+                        },
```
arvind: I'm a little confused about where this addition came from as there's no associated edit to a `vl.json` spec?

kanitw: It's artifact from a former commit that got rebase merged.

src/compile/scale/properties.ts
```diff
@@ -102,6 +102,7 @@ export function parseNonUnitScaleProperty(model: Model, property: keyof (Scale |
         // TODO: precedence rule
         valueWithExplicit = mergeValuesWithExplicit(
           valueWithExplicit, childValueWithExplicit,
+          property,
```
domoritz: How come you are adding a new parameter here?

kanitw: Well, you need to know what's the property to make the warning message include the property name.

Issue #2511, opened 2017-06-15
`domain` appears broken in `layer` context.


yhoonkim:
Here is the spec where I wanted to change the domain of x-axis to give a space. But it didn't work.
![image](https://user-images.githubusercontent.com/6269456/27193033-41b879a4-51b2-11e7-9b53-0e2ec4e14524.png)

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "width": 320,
  "description":"A candlestick chart inspired from Protovis(http://mbostock.github.io/protovis/ex/candlestick.html)",
  "data": {
    "values": [
      {
        "date": "01-Jun-2009",
        "open": 28.7,
        "high": 30.05,
        "low": 28.45,
        "close": 30.04,
        "signal": "short",
        "ret": -4.89396411092985
      },
      {
        "date": "02-Jun-2009",
        "open": 30.04,
        "high": 30.13,
        "low": 28.3,
        "close": 29.63,
        "signal": "short",
        "ret": -0.322580645161295
      },
      {
        "date": "03-Jun-2009",
        "open": 29.62,
        "high": 31.79,
        "low": 29.62,
        "close": 31.02,
        "signal": "short",
        "ret": 3.68663594470045
      },
      {
        "date": "04-Jun-2009",
        "open": 31.02,
        "high": 31.02,
        "low": 29.92,
        "close": 30.18,
        "signal": "short",
        "ret": 4.51010886469673
      },
      {
        "date": "05-Jun-2009",
        "open": 29.39,
        "high": 30.81,
        "low": 28.85,
        "close": 29.62,
        "signal": "short",
        "ret": 6.08424336973478
      },
      {
        "date": "08-Jun-2009",
        "open": 30.84,
        "high": 31.82,
        "low": 26.41,
        "close": 29.77,
        "signal": "short",
        "ret": 1.2539184952978
      },
      {
        "date": "09-Jun-2009",
        "open": 29.77,
        "high": 29.77,
        "low": 27.79,
        "close": 28.27,
        "signal": "short",
        "ret": -5.02431118314424
      },
      {
        "date": "10-Jun-2009",
        "open": 26.9,
        "high": 29.74,
        "low": 26.9,
        "close": 28.46,
        "signal": "short",
        "ret": -5.46623794212217
      },
      {
        "date": "11-Jun-2009",
        "open": 27.36,
        "high": 28.11,
        "low": 26.81,
        "close": 28.11,
        "signal": "short",
[...]

PR #2505, opened 2017-06-15
Added a candlestick example

yhoonkim:
Fix #2432.

PR #2536, opened 2017-06-19
Initial code for making `domain` => `domains` in scaleComponent

kanitw:


domoritz:
Thanks! Can you make sure that the type of `domains` is only `any[] | VgDataRef | VgSignalRef`?

kanitw:
> Thanks! Can you make sure that the type of domains is only any[] | VgDataRef | VgSignalRef?

I tried but I don't think you can.  Try to put the second commit and see.  Maybe it's not the whole `VgDomain`, but a bit more than just `any[] | VgDataRef | VgSignalRef`

domoritz:
Makes sense. Basically, I don't want any unioned domains in there because that makes it much harder to merge them.

domoritz:
Can you rebase this and fix the types of the array so I can take this over?

domoritz:
Getting there. Todo is:

* [x] remove unionDomain
* [x] simplify code by requiring domain to be array until assemble
* [x] add tests for domain merge method
* [x] make things compile again
* [ ] fix things

domoritz:
This may fix https://github.com/vega/vega-lite/issues/2511

kanitw:
I remove the text "may fix" as github will accidentally close the issue.

examples/specs/normalized/box-plot_minmax_1D_horizontal_normalized.vl.json
```diff
@@ -44,14 +12,16 @@
       },
       "encoding": {
         "x": {
-          "field": "lowerWhisker",
-          "type": "quantitative",
           "axis": {
             "title": "population"
-          }
+          },
```
kanitw: Why does this revert boxplot?

domoritz: Hmm, weird. Does it actually revert it or does the aggregation not happen in the encoding? What line chauses this change anyway?

domoritz: Whatever is happening, the Vega output is the same so it's fine.

domoritz:
@kanitw this is ready to be merged.

domoritz:
This fixes https://github.com/vega/vega-lite/pull/2505

src/compile/scale/assemble.ts
```diff
@@ -5,17 +5,21 @@ import {stringValue, vals} from '../../util';
...
       if (scaleComponent.merged) {
         // Skipped merged scales
         return scales;
       }
 
       // We need to cast here as combine returns Partial<VgScale> by default.
-      const scale = scaleComponent.combine(['name', 'type', 'domain', 'domainRaw', 'range']) as VgScale;
+      const scaleComProps = scaleComponent.combine(['name', 'type', 'domainRaw', 'range']);
+      // FIXME: why do I need to delete this property?
+      delete scaleComProps.domains;
```
kanitw: I thought we agree to just make `domains` a part of the scale component directly? 

In any case, having a `// FIXME: why do I need to delete this property?` comment isn't good.

src/compile/scale/assemble.ts
```diff
@@ -5,17 +5,21 @@ import {stringValue, vals} from '../../util';
...
-    return vals(model.component.scales).reduce((scales: VgScale[], scaleComponent) => {
+    return vals(model.component.scales).reduce((scales, scaleComponent) => {
       if (scaleComponent.merged) {
         // Skipped merged scales
         return scales;
       }
 
       // We need to cast here as combine returns Partial<VgScale> by default.
-      const scale = scaleComponent.combine(['name', 'type', 'domain', 'domainRaw', 'range']) as VgScale;
+      const scaleComProps = scaleComponent.combine(['name', 'type', 'domainRaw', 'range']);
```
kanitw: Instead of passing `['name', 'type', 'domainRaw', 'range']` to combine.  

Maybe do this below to preserve the nice scale order:
```
const {name, type, domainRaw, range, ...otherScaleProps} = scale;
scales.push({
  name,
  type,
  ...(domain ? {domain} : {}),
  ...(domainRaw ? {domainRaw} : {}),
  ...(range ? {range} : {}),
})
```

I think we can also remove the optional parameter from the `combine` method as I believe it is only used here.

src/compile/scale/domain.ts
```diff
@@ -166,85 +160,89 @@ function parseSingleChannelDomain(scaleType: ScaleType, domain: Domain, model: U
...
-    return {
-      data: model.requestDataName(MAIN),
-      fields: [
-        model.field(channel, {suffix: 'start'}),
-        model.field(channel, {suffix: 'end'})
-      ]
-    };
+
+    const data = model.requestDataName(MAIN);
+    return [{
```
kanitw: Do we really need to split these? 

I know you don't want `data` with `fields` this also makes this method a bit unnecessarily complicated with the wrapped array.

If you really prefer this, I think this method should be renamed to `parseSingleChannelDomains`

src/compile/scale/domain.ts
```diff
@@ -314,60 +312,97 @@ export function canUseUnaggregatedDomain(fieldDef: FieldDef<string>, scaleType:
...
+      let sort = sorts[0];
+      if (sorts.length > 1) {
+        log.warn(log.message.MORE_THAN_ONE_SORT);
+        sort = true;
+      }
       return {
-        data: domain.data,
-        field: d
+        ...domain,
+        sort: sort
```
kanitw: just `sort` for consistency?

src/compile/scale/domain.ts
```diff
@@ -314,60 +312,97 @@ export function canUseUnaggregatedDomain(fieldDef: FieldDef<string>, scaleType:
...
-  throw new Error(log.message.INVAID_DOMAIN);
-}
 
-/**
- * Union two data domains. A unioned domain is always sorted.
- */
-export function unionDomains(domain1: VgDomain, domain2: VgDomain): VgDomain {
-  const normalizedDomain1 = normalizeDomain(domain1);
-  const normalizedDomain2 = normalizeDomain(domain2);
+  let sort: VgUnionSortField = true;
```
kanitw: (minor) move `sort` here to above the if below for readability?

src/compile/scale/domain.ts
```diff
@@ -314,60 +312,97 @@ export function canUseUnaggregatedDomain(fieldDef: FieldDef<string>, scaleType:
...
+    const domain: FieldRefUnionDomain = {
+      data: allData[0],
+      fields: uniqueDomains.map(d => (d as VgDataRef).field),
+      sort
+    };
 
-    return {fields: domains, sort: true};
-  } else {
-    return domains[0];
+    return domain;
```
kanitw: why not just return directly?

src/compile/scale/parse.ts
```diff
@@ -8,13 +8,12 @@ import {Model} from '../model';
 import {defaultScaleResolve} from '../resolve';
 import {Explicit, mergeValuesWithExplicit, tieBreakByComparing} from '../split';
 import {UnitModel} from '../unit';
-import {ScaleComponent, ScaleComponentIndex} from './component';
+import {ScaleComponent, ScaleComponentIndex, ScaleComponentProps} from './component';
```
kanitw: New import but not used?

src/compile/scale/domain.ts
domoritz: No, if there is no sort, we don't add it. See code below.

src/compile/scale/domain.ts

src/compile/scale/domain.ts
domoritz: let's me check the type: `FieldRefUnionDomain`

src/compile/scale/parse.ts
domoritz: done

src/compile/scale/domain.ts
```diff
@@ -317,60 +312,97 @@ export function canUseUnaggregatedDomain(fieldDef: FieldDef<string>, scaleType:
...
-  } else if (isDataRefDomain(domain)) {
-    delete domain.sort;
-    return [domain];
-  } else if(isFieldRefUnionDomain(domain)) {
-    return domain.fields.map(d => {
+export function mergeDomains(domains: VgNonUnionDomain[]): VgDomain {
+  const uniqueDomains = util.unique(domains.map(d => {
+    // ignore sort property when computing the unique domains
+    if (isDataRefDomain(d)) {
+      const {sort: _s, ...dom} = d;
```
domoritz: Cleaned up differently.

src/compile/scale/domain.ts
domoritz: I don't get this comment. I renamed the method.

domoritz: I don't get this comment. The domain is still a single domain.

src/compile/scale/assemble.ts
domoritz: I thought you would do this, lol. Anyway, done.

src/compile/scale/assemble.ts
domoritz: Done

domoritz:
Done

src/compile/scale/domain.ts
kanitw: huh?    `{sort: sort}` is equivalent to `{sort}`.    So I'm not sure what you're talking about.

src/compile/scale/domain.ts
kanitw: It now becomes an array of domains, rather than a single Vega domain object. 

In fact, I don't think you really need to do this (but rather modify the merge method to handle this case.)  That said, if you're lazy, at least rename the method.

domoritz: I disagree with the second part. I want the domain to be an array as until assemble because that makes merging very easy. During parse, we only concat and during assembly, we don't have to deal with unioned domains. 

I also don't want to rename the method because the method is parsing the domain (but happens to return an array of domain objects, which represent the domain.

src/compile/scale/domain.ts
domoritz: `domain`  here doesn't have a sort anymore. It comes from `uniqueDomains`.

kanitw: It still does -- otherwise github would have hide it...

src/compile/scale/domain.ts
kanitw: > I disagree with the second part. I want the domain to be an array as until assemble because that makes merging very easy. During parse, we only concat and during assembly, we don't have to deal with unioned domains.

But my proposal still makes domain component an array that will be merge later.  Only that `parseSingleChannelDomains` still return a single domain object. 
You don't really have to deal with really union domain from multiple data.  You only have to deal with domain with two fields (but same data), which is not really complicated and do not need to make this method weird.

src/compile/scale/assemble.ts
```diff
@@ -5,47 +5,51 @@ import {stringValue, vals} from '../../util';
...
+      // domains is an array that has to be merged into a single vega domain
+      const domain = mergeDomains(domains);
+
+
+      scales.push({
+        name,
+        type,
+        domain: domain,
+        range: range,
+        ...(domainRaw ? {domainRaw} : {}),
```
kanitw: Actually, domainRaw should come before range.  (My previous comment might be wrong too.)

src/compile/scale/domain.ts
domoritz: What?

src/compile/scale/domain.ts
domoritz: But I still want `parseDomainForChannel` to return an array of domains (without any of the individual items being a union.

kanitw: Why?

kanitw: The hard part is not FieldRefUnionDomain, but rather the other one?

domoritz:
Rebased

PR #2538, opened 2017-06-19
Facet of Facet

domoritz:


domoritz:
This is ready. The only issue is that a width signal is missing. 

```
Unrecognized signal name: "child_width"
```

kanitw:
Can't reproduce your problem yet ... I got this weird warnings 

developer_tools_-_http___localhost_3000__and_vega_editor_and_1__vy2-redux__zsh_

domoritz:
Those warnings come from react, right?

kanitw:
Yes.

domoritz:
Then I suppose they are unrelated.

src/compile/data/assemble.ts
```diff
@@ -47,7 +47,7 @@ function removeUnnecessaryNodes(node: DataFlowNode) {
  */
 function cloneSubtree(facet: FacetNode) {
   function clone(node: DataFlowNode): DataFlowNode[] {
-    if (!(node instanceof OrderNode)) {  // we can ignore order nodes beacuse they don't change the scale domain
+    if (!(node instanceof OrderNode || node instanceof FacetNode)) {  // we can ignore order nodes beacuse they don't change the scale domain
```
kanitw: > // we can ignore order nodes beacuse they don't change the scale domain

1.  Typo "beacuse" => "because"
2. This is not always true if the scale domain does not have additional sorting.

src/compile/scale/domain.ts
```diff
@@ -90,12 +90,12 @@ function parseNonUnitScaleDomain(model: Model) {
     if (model instanceof FacetModel) {
       // Replace the scale domain with data output from a cloned subtree after the facet.
       if (isDataRefDomain(domain) || isFieldRefUnionDomain(domain)) {
-        domain.data = FACET_SCALE_PREFIX + model.getName(domain.data);
+        domain.data = FACET_SCALE_PREFIX + domain.data.replace(FACET_SCALE_PREFIX, '');
```
kanitw: Need comment why you are replacing here.

src/compile/scale/domain.ts
```diff
@@ -90,12 +90,12 @@ function parseNonUnitScaleDomain(model: Model) {
...
       } else if (isDataRefUnionedDomain(domain)) {
         domain.fields = domain.fields.map((f: VgDataRef) => {
           return {
             ...f,
-            data: FACET_SCALE_PREFIX + model.getName(f.data)
+            data: FACET_SCALE_PREFIX + f.data.replace(FACET_SCALE_PREFIX, '')
```
kanitw: Need comment why you are replacing here.

kanitw:
I also add a hack that it always output compiling spec but may not always correctly center headers.  (Need additional input from @jheer as asked on slack to properly make things correct.)

kanitw:
If you fix other things, feel free to merge the PR with my hack.  We can deal with layout in a separate PR.

src/compile/data/assemble.ts
domoritz: Good point. Fixed.

src/compile/scale/domain.ts

domoritz:
Ups, I accidentally force pushed a rabase. Can you push your commit with the hack again? Sorry about that.

package.json
```diff
@@ -49,7 +49,7 @@
     "poststart": "rm examples/all-examples.json",
 
     "preschema": "npm run prebuild",
-    "schema": "typescript-to-json-schema --path tsconfig.json --type TopLevelExtendedSpec > build/vega-lite-schema.json && npm run renameschema && cp build/vega-lite-schema.json _data/",
+    "schema": "node --stack-size=10000 node_modules/.bin/typescript-to-json-schema --path tsconfig.json --type TopLevelExtendedSpec > build/vega-lite-schema.json && npm run renameschema && cp build/vega-lite-schema.json _data/",
```
domoritz: Not super happy about this but don't have a better fix right now.

kanitw:
Done

domoritz:
Awesome. This should be good to go then.

domoritz:
Rebased

domoritz:
Rebased and fixed. Just need to figure out why schema gen is slow.

src/compile/scale/domain.ts

domoritz:
Fixed the slow runtime.

kanitw:
Yay!  -- improve the codecov?

domoritz:
@kanitw, the majority of the path coverage is dataflow stuff and needs some more work before I can write tests. So no for now :-P

PR #2539, opened 2017-06-19
Apply nullFilter only to fields with continuous domain scales

### LINKED ISSUE #2066: filterInvalid
Based on the current transform refactor https://github.com/vega/vega-lite/pull/2064 by @domoritz, 
`filterInvalid` is temporarily a part of `config`.  However, if our goal is to make config only for theming, we need alternative way for specifying `filterInvalid`.

kanitw:
- To make results consistent among multiple subplots, we should apply filter invalid values  only when necessary (fields with continuous domain scales)
- Fix ordinal scale type for color (discovered bug while implementing this). 
- Change config's `filterInvalid: boolean` to `invalidValues: "filter"` to support future extension for other ways to handle invalid values. For example, we could have a config where we make invalid values grey (to help distinguish them from zero values) instead of filtering
-- See https://github.com/vega/vega-lite/issues/2544.  (Note that it was `transform.filterInvalid` in VL1.0 anyway so even if we don't change, we don't really prevent breaking changes.) 
  
Fix #2066.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -147,11 +147,11 @@ export interface VLOnlyConfig {
   countTitle?: string;
 
   /**
-   * Whether to filter invalid values (`null` and `NaN`) from the data.
-   * - If set to `true` (default), all data items with null values are filtered.
-   * - If `false`, all data items are included. In this case, invalid values will be interpreted as zeroes.
+   * Defines how Vega-Lite should handle invalid values (`null` and `NaN`).
+   * - If set to `"filter"` (default), all data items with null values are filtered.
+   * - If `null`, all data items are included. In this case, invalid values will be interpreted as zeroes.
    */
-  filterInvalid?: boolean;
+  invalidValues?: 'filter';
 
   /**
    * D3 Number format for axis labels and text tables. For example "s" for SI units.(in the form of [D3 number format pattern](https://github.com/mbostock/d3/wiki/Formatting)).
@@ -222,7 +222,7 @@ export const defaultConfig: Config = {
   timeFormat: '%b %d, %Y',
   countTitle: 'Number of Records',
 
-  filterInvalid: true,
+  invalidValues: 'filter',
 
   cell: defaultCellConfig,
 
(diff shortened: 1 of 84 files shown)


kanitw:
@domoritz  I don't understand why the data pipeline can create a data source with one transform  = null for `layer_overlay.vl.json`.

```
  1) Examples layer_overlay.vl.json should produce valid vega:
     TypeError: Cannot read property 'type' of null
      at Object.assembleData (src/compile/data/assemble.ts:349:12)
      at LayerModel.assembleData (src/compile/layer.ts:172:14)
      at assemble (src/compile/compile.ts:80:13)
      at Object.compile (src/compile/compile.ts:39:12)
      at validateVega (examples/examples.test.ts:36:20)
      at Context.<anonymous> (examples/examples.test.ts:59:9)
```

src/config.ts
```diff
@@ -147,12 +147,11 @@ export interface VLOnlyConfig {
...
-   * Whether to filter invalid values (`null` and `NaN`) from the data.
-   * - By default (`undefined`), only quantitative and temporal fields are filtered.
-   * - If set to `true`, all data items with null values are filtered.
-   * - If `false`, all data items are included. In this case, null values will be interpreted as zeroes.
+   * Defines how Vega-Lite should handle invalid values (`null` and `NaN`).
+   * - If set to `"filter"` (default), all data items with null values are filtered.
+   * - If `null`, all data items are included. In this case, invalid values will be interpreted as zeroes.
    */
-  filterInvalid?: boolean;
+  invalidValues?: 'filter';
```
domoritz: Why not true and false? `filter` and `null` are weird.

kanitw: > Change config's filterInvalid: boolean to invalidValues: "filter" to support future extension for other ways to handle invalid values. For example, we could have a config where we make invalid values grey (to help distinguish them from zero values) instead of filtering.

kanitw: This will prevent future breaking changes as mentioned.  

Plus, `filterInvalid` was weird too.

domoritz:
You know how I feel about adding thinks for a distant future but I don't
want to argue about this one.

kanitw:
> You know how I feel about adding thinks for a distant future but I don't
want to argue about this one.

In general, it is good to think forward how the spec would be like so we don't have to add breaking changes.
Plus, (1) that "distant" future may not be too far from now (possibly in a few weeks likely in a few months)  (2) the current state of `filterInvalid` is bad anyway.

domoritz:
> that "distant" future may not be too far from now (possibly in a few weeks likely in a few months) 

The main issue I see is how we do those highlights. We have similar issues with default interactions where we don't know how to highlight in general since the user may override the opacity channel themselves for example.

kanitw:
>  The main issue I see is how we do those highlights. We have similar issues with default interactions where we don't know how to highlight in general since the user may override the opacity channel themselves for example.

I don't plan to override opacity channel but instead use production rules for fill/stroke, which will still preserve existing color encoding. (A bit more similar to Jeff's null bin example)

PR #2542, opened 2017-06-20
Ensure selection and signal names are legal JavaScript identifiers.

arvind:
Fixes vega/vega#869.


kanitw:
@domoritz  -- Did we implement similar things somewhere?  I kinda remember but couldn't find it.  I think we should migrate that part of code to use `varName` too.

Issue #1804, opened 2017-01-10
Redesign overlay config


kanitw:
Forked from #1727 

We currently have overlay config hidden as the design may not be optimal yet.

We might want to
- [ ] remove `config.overlay.line` and add `config.line.point` (or `.pointOverlay`) where `point` is either null or a mark config.  (This will allow users to customize the theme of the overlayed points as well.) 

- [ ] remove `config.overlay.area` and add `config.area.line` (or `.lineOverlay`) and `config.area.pointOverlay` (or just `.pointOverlay`) similar to line.

kanitw:
We might be able to fit this as a composite mark as in https://github.com/vega/vega-lite/pull/1940

kanitw:
- [ ] Drop shape from line/area in normalizeOverlay (#1646)

kanitw:
- [ ] also revise `selection`

kanitw:
This will be a part of #3604.

PR #2546, opened 2017-06-21
Exclude overlay config from the schema, …

kanitw:
so it's a secret feature only for Voyager and CompassQL for now.

(This will allow us to punt on https://github.com/vega/vega-lite/issues/1804 without making breaking changes in the future.)

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -45,14 +45,6 @@
         }
       ]
     },
-    "AreaOverlay": {
-      "enum": [
-        "line",
-        "linepoint",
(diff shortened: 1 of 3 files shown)


arvind:
What does `overlay` do? I didn't even realize this was something we supported! :)

kanitw:
It allows users of single spec to overlay line with points / area with lines.  

vega_editor

As discussed in #1804, I don't like the current syntax but this is used in Voyager.  
Instead of redesigning it now (or migrate the code elsewhere), we can make it a secret feature and work on either of them later.

PR #2548, opened 2017-06-21
Update CONTRIBUTING.md

domoritz:


CONTRIBUTING.md
```diff
@@ -178,10 +169,10 @@ for instructions).
 ## Developing Vega-Lite and Vega-Util
 
 Vega-Lite depends on [vega-util](https://github.com/vega/vega-util).
-If you plan to make changes to datalib and test Vega-Lite without publishing / copying compiled vega-util all the time, use [`npm link`](http://justjs.com/posts/npm-link-developing-your-own-npm-modules-without-tears) command.
+If you plan to make changes to the uils and test Vega-Lite without publishing / copying compiled vega-util all the time, use [`npm link`](http://justjs.com/posts/npm-link-developing-your-own-npm-modules-without-tears) command.
```
ayush29feb: is it suppose to be utils?

domoritz:
What else? It's just an example. Maybe we should just delete the whole
block because it's not that relevant anymore.b

CONTRIBUTING.md
domoritz: What else? It's just an example. Maybe we should just delete the whole
block because it's not that relevant anymore.

ayush29feb: Oh I mean it looks like a typo, you added uils

domoritz: Good catch. Fixed in master.

PR #2549, opened 2017-06-21
Implement method to add node into dataflow

domoritz:


test/compile/data/dataflow.test.ts
```diff
@@ -4,7 +4,7 @@ import {assert} from 'chai';
 import {DataFlowNode} from '../../../src/compile/data/dataflow';
 
 describe('compile/data/dataflow', function() {
-  describe('DataFlowNode', function() {
+  describe.only('DataFlowNode', function() {
```
kanitw: describe.only **

src/compile/data/dataflow.ts
```diff
@@ -69,6 +69,16 @@ export class DataFlowNode {
     this._parent.removeChild(this);
   }
 
+  /**
+   * Insert another node as a parent of this node.
+   */
+  public insertAsParentOf(other: DataFlowNode) {
```
kanitw: other => child?

src/compile/data/dataflow.ts
```diff
@@ -69,6 +69,16 @@ export class DataFlowNode {
...
+    const parent = other.parent;
+    parent.removeChild(this);
```
kanitw: removeChild(other) ?

test/compile/data/dataflow.test.ts
domoritz: Fixed

src/compile/data/dataflow.ts
domoritz: Done

src/compile/data/dataflow.ts
domoritz: Done and force pushed. Also added test.

kanitw:
Not sure why the other test doesn't pass.

domoritz:
Needs to be restarted. Same issue that I reported earlier.

domoritz:
Feel free to merge. The issues are just in the build system.

PR #2550, opened 2017-06-21
Use mapped type to declare `ResolveMapping`

kanitw:


src/resolve.ts
```diff
@@ -17,16 +17,13 @@ export interface NonspatialResolve {
 
 export type Resolve = SpatialResolve | NonspatialResolve;
 
-// TODO: replace this with {[P in NonspatialScaleChannel]?: NonspatialResolve} & {[P in SpatialScaleChannel]?: SpatialResolve}; and make sure that the right schema is being generated
 export type ResolveMapping = {
-  // spatial channels
+  // Spatial channels
   x?: SpatialResolve
```
domoritz: Replace this with `{[P in SpatialScaleChannel]?: SpatialResolve}`?

kanitw: NonSpatialScaleChannel may grow over time, SpatialScaleChannel would not -- so I intentionally didn't change.  If you feel strong about feel free to go ahead and make the change.

domoritz: Cool. Makes sense.

PR #2554, opened 2017-06-22
Install a stable version of parallel

### LINKED ISSUE #2524: Travis error because of parallel
```
gpg: requesting key FFFFFFF1 from hkp server keys.gnupg.net
gpg: key FFFFFFF1: public key "Ole Tange <ole@tange.dk>" imported
gpg: key FFFFFFF1: public key "Ole Tange <ole@tange.dk>" imported
gpg: no ultimately trusted keys found
gpg: Total number processed: 2
gpg:               imported: 2  (RSA: 1)
gpg: requesting key 88888888 from hkp server keys.gnupg.net
gpgkeys: key 88888888 not found on keyserver
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
The signature on parallel-20170522.tar.bz2 is wrong. This may indicate that a criminal has changed the code.
THIS IS BAD AND THE CODE WILL NOT BE INSTALLED.
See http://git.savannah.gnu.org/cgit/parallel.git/tree/README for other installation methods.
The command "(wget -O - pi.dk/3 || curl pi.dk/3/ || fetch -o - http://pi.dk/3) | bash" failed and exited with 1 during .

```

NeelMohapatra:
Fixes #2524

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,11 @@ node_js:
   - "7"
   - "8"
 before_install:
-  - wget http://git.savannah.gnu.org/cgit/parallel.git/plain/src/parallel # Minimal installation
-  - chmod 755 parallel
+  - # Minimal installation as desdcribed in https://github.com/martinda/gnu-parallel
+  - wget http://git.savannah.gnu.org/cgit/parallel.git/plain/src/parallel
+  - chmod 700 parallel # Need rwx permissoins for user.
   - cp parallel sem
-  - sudo mv parallel sem /bin
+  - sudo mv parallel sem /usr/local/bin # echo $PATH for valid folders
 
 script:
   - npm run build
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ node_js:
 before_install:
   - # Minimal installation as desdcribed in https://github.com/martinda/gnu-parallel
   - wget http://git.savannah.gnu.org/cgit/parallel.git/plain/src/parallel
-  - chmod 700 parallel # Need rwx permissoins for user.
+  - chmod +x parallel
   - cp parallel sem
-  - sudo mv parallel sem /usr/local/bin # echo $PATH for valid folders
+  - sudo mv parallel sem /usr/local/bin
 
 script:
   - npm run build
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,11 @@ node_js:
   - "7"
   - "8"
 before_install:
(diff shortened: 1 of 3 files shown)


NeelMohapatra:
Someone needs to file a bug with apt-get and parallel to update their lists.

domoritz:
Can you explain why these commands are the way they are? Why is chmod necessary? How does this resolve the problems we saw?

NeelMohapatra:
I'm installing it by getting the files directly from their site. That prevents the need to match checksums. I think the chmod is necessary for the command to be executable. Then I'm just moving the program to \bin

domoritz:
Please just use `chmod +x` then. And to make the program findable, please put it in `/usr/local/bin` or change the `PATH`.

NeelMohapatra:
Why are these changes necessary?

domoritz:
Because I want to understand what's happening and we should make sure this is a clean install so that it doesn't keep breaking.

.travis.yml
```diff
@@ -4,10 +4,11 @@ node_js:
   - "7"
   - "8"
 before_install:
-  - wget http://git.savannah.gnu.org/cgit/parallel.git/plain/src/parallel # Minimal installation
-  - chmod 755 parallel
+  - # Minimal installation as desdcribed in https://github.com/martinda/gnu-parallel
+  - wget http://git.savannah.gnu.org/cgit/parallel.git/plain/src/parallel
+  - chmod 700 parallel # Need rwx permissoins for user.
```
domoritz: Not sure why you need write and read permissions. `chmod + x parallel` may be sufficient.

NeelMohapatra: Okay. I just put read/write permissions in case the system needed to alter files to execute use parallel.

domoritz: Read and write is for the file, not the executable.

scripts/generate-images.sh
```diff
@@ -8,4 +8,4 @@ set -e
 export NODE_PATH=.
 
 echo "Generating SVGs..."
-ls examples/specs/*.vl.json | parallel --eta --halt 1 "bin/vl2svg {} build/examples/images/{/.}.svg"
+ls examples/specs/*.vl.json |  --no-notice --eta --halt 1 "bin/vl2svg {} build/examples/images/{/.}.svg"
```
NeelMohapatra: @kanitw @domoritz 😂

NeelMohapatra: I can't believe we missed this.

domoritz: Ups.

PR #2556, opened 2017-06-23
Updated Vega-Embed on the site

yhoonkim:
Updated some pages of the site to use new Vega-Embed. 

My major concern of this PR is about `streaming.ts`. I changed the code of it under an assumption that we're going to use Vega-Embed with the promise feature which is not released yet. When testing, I linked Vega-Embed to the current master branch of Vega-Embed.

### DIFF
--- a/site/static/streaming.ts
+++ b/site/static/streaming.ts
@@ -18,7 +18,7 @@ export function runStreamingExample(eleId: string) {
 
   embed(eleId, vlSpec, {
     actions: false
-  }).then(function(res){
+  }).then(function(res) {
     const {view} = res;
 
     /**
--- a/site/static/streaming.ts
+++ b/site/static/streaming.ts
@@ -14,7 +14,7 @@ export function runStreamingExample(eleId: string) {
     }
   };
 
-  function cb(err: Error, res: {view: any, spec: any}) {
+  function cb(res: {view: any, spec: any}) {
     const {view} = res;
 
     /**
@@ -47,5 +47,5 @@ export function runStreamingExample(eleId: string) {
 
   embed(eleId, vlSpec, {
     actions: false
-  }, cb);
+  }).then(cb);
 }
--- a/site/static/streaming.ts
+++ b/site/static/streaming.ts
@@ -14,7 +14,11 @@ export function runStreamingExample(eleId: string) {
     }
   };
 
-  function cb(res: {view: any, spec: any}) {
+
+
+  embed(eleId, vlSpec, {
+    actions: false
+  }).then(function(res){
     const {view} = res;
 
     /**
@@ -43,9 +47,5 @@ export function runStreamingExample(eleId: string) {
       const changeSet = view.changeset().insert(valueGenerator()).remove((t: {x: number}) => t.x < minimumX);
       view.change('table', changeSet).run();
     }, 1000);
-  }
-
-  embed(eleId, vlSpec, {
-    actions: false
-  }).then(cb);
+  });
 }
--- a/site/tsconfig.json
+++ b/site/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
     "module": "commonjs",
     "moduleResolution": "node",
     "isolatedModules": false,
--- a/site/tsconfig.json
+++ b/site/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es6",
+    "target": "es5",
     "module": "commonjs",
     "moduleResolution": "node",
     "isolatedModules": false,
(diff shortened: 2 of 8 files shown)


site/static/streaming.ts
```diff
@@ -14,7 +14,7 @@ export function runStreamingExample(eleId: string) {
     }
   };
 
-  function cb(err: Error, res: {view: any, spec: any}) {
```
domoritz: Let's move this code inside the `then` call.

site/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
```
domoritz: Please don't change this! This means we generate es6 and this would break everyone who uses a browser that doesn't support es6.

site/usage/compile.md
```diff
@@ -20,14 +20,6 @@ If you want access to the compiled Vega spec from a Javascript program, you can
 var vgSpec = vl.compile(vlSpec).spec;
 ```
 
-Then, you can continue to use the [Vega runtime](https://github.com/vega/vega/wiki/Runtime)'s `vg.parse.spec` method to render your Vega spec.
```
domoritz: What's wrong with this?

site/usage/embed.md
```diff
@@ -9,22 +9,23 @@ permalink: /usage/embed.html
 
 The easiest way to use Vega-Lite on your own web page is with [Vega-Embed](https://github.com/vega/vega-embed), a library we built to make the process as painless as possible.
 
-To embed a Vega-Lite specification on your web page first load the required libraries (D3, Vega, Vega-Lite, **Vega-Embed**).
```
domoritz: Good catch.

site/usage/embed.md
```diff
@@ -35,15 +36,45 @@ The next step is to create a DOM element that the visualization will be attached
...
 ```
 
 Vega-Embed automatically adds links to export an image, view the source, and open the specification in the online editor. These links can be individually disabled. For more information, read the [Vega-Embed documentation](https://github.com/vega/vega-embed).
+
+Here is the fianl html file in the easiest way to embed Vega-Lite:
+```html
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Embedding Vega-Lite</title>
```
domoritz: Above, we show how to use cdnjs, which I think should be the default for beginners.

site/tutorials/getting_started.md
```diff
@@ -12,6 +12,14 @@ This tutorial will guide through the process of writing a visualization specific
 
 We suggest that you follow along the tutorial by building a visualization in the [online editor](https://vega.github.io/new-editor/?mode=vega-lite). Extend your specification in the editor as you read through this tutorial. If something does not work as expected, compare your specifications with ones inside this tutorial.
 
+## $schema
```
domoritz: out it in `

site/tutorials/getting_started.md
```diff
@@ -12,6 +12,14 @@ This tutorial will guide through the process of writing a visualization specific
...
+
+Before diving into components illustrating visualizations, we want to introduce "$schema" property that defines the [JSON schema](http://json-schema.org/) of the visualization specification in a certain version of Vega-Lite. It accepts value as an URL of the schema.
```
domoritz: **the** $schema property

site/tutorials/streaming.md
```diff
@@ -31,7 +31,7 @@ Putting it together, we have:
   view.change('table', changeSet).run();
 ```
 
-The view is given to us by the Vega [embed](https://github.com/vega/vega-embed) method, which help load Vega/Vega-Lite specs on your webpages. The callback argument to `embed` is invoked when embed is complete, with parameters: `error`, then an object containing the `view`, which is how we could access the `view.change` mentioned earlier, and `spec`, which is the compiled Vega spec from the original Vega-Lite spec given.
+The view is given to us by the Vega [embed](https://github.com/vega/vega-embed) method, which helps load Vega/Vega-Lite specs on your webpages. When `embed` completes successfully, `embed` returns a fulfilled [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) with an object containing the `view`, which is how we could access the `view.change` mentioned earlier, and `spec`, which is the compiled Vega spec from the original Vega-Lite spec given. When `embed` fails, it returns a rejected [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) with an error object.
```
domoritz: let's not call it the "Vega" embed method, because it is not part of vega.

site/usage/compile.md
yhoonkim: 1) `vg.parse.spec` method is deprecated. (I think it is replaced by `vega.parse`.)
2) I thought we should promote to use Vega-Embed to render the compiled spec too.
3) I felt that this info is not related to *Compile* and not necessary to be here.
So I just decided to delete this incorrect info.

yhoonkim:
Thanks @domoritz for reviews! I pushed a commit on this branch to reflect them.
(But I don't know why it is not shown in this thread...)

site/usage/compile.md
domoritz: Makes sense. Thanks for the explanation.

domoritz:
I pushed some small fixes. Can you also address https://github.com/vega/vega-lite/pull/2556/files#r123850983?

PR #2558, opened 2017-06-23
Exclude build from .vscode

kanitw:
This screws up imports all the time, I don't know why we need build inside vscode.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,8 @@
   "files.exclude": {
     "coverage/**/*": true,
     "_site/**/*": true,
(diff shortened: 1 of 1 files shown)


domoritz:
I often have to look into built. For example the schema.

kanitw:
If you only look at schema, can we only include only schema -- so we don't pollute things including import etc.

domoritz:
I sometimes also look at the generated js files. Not that often, though.

I still don't understand why this is necessary since I added

```json
"search.exclude": {
    "build": true
  },
```

PR #2564, opened 2017-06-23
[WIP] Merge layout signals  

### LINKED ISSUE #2198: Logic for merging size signal

kanitw:
Fix #2198

- [ ] correctly assemble for concat/repeat/facet

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,5 +11,8 @@
   },
   "files.trimTrailingWhitespace": true,
   "tsimporter.spaceBetweenBraces": false,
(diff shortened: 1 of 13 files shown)


kanitw:
Replaced with #2643

Issue #2555, opened 2017-06-23
Concat/Facet width/height signals reference out of scope scales


arvind:
- [x] **Concat**

With the following concatenated spec, the `concat_width` and `concat_height` signals are defined at the top-level. However, the corresponding scales referenced within their `update` expressions are not in scope (the scales are only defined within each `concat_group`). As a result, these top-level signals have a value of `0`. This is problematic for selections (which reference the size signals to clamp interactions correctly).

If the size signals are moved to be within the correct scopes (e.g., `concat_0_width`/`height` is moved to be a child of `concat_0_group`), the values are correctly calculated.

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {"url": "data/seattle-temps.csv"},
  "vconcat": [{
    "mark": "bar",
    "encoding": {
      "x": {
        "timeUnit": "month",
        "field": "date",
        "type": "temporal"
      },
      "y": {
        "aggregate": "mean",
        "field": "temp",
        "type": "quantitative"
      }
    } 
  }, {
    "mark": "bar",
    "encoding": {
      "x": {
        "timeUnit": "month",
        "field": "date",
        "type": "temporal"
      },
      "y": {
        "aggregate": "median",
        "field": "temp",
        "type": "quantitative"
      }
    }
  }],
  "config": {"cell": {"fill": "yellow"}}
}
```

kanitw:
Sounds like we have to implement a lot of signals that push value to outer scope instead?

arvind:
Pushing to outer scope is fairly easy. We do it with selections. Signals should have the same name, and the inner signals should be marked with `"push": "outer"`. 

> One observation -- this is only necessary for shared scales.

Not sure what "this" is referencing but selections need valid size signals to clamp interaction correctly. 

>

kanitw:
@arvind  -- This is a pretty quick reply for vacation day :p   Thanks!

> Not sure what "this" is referencing but selections need valid size signals to clamp interaction correctly.

Sorry this is more like a note to myself -- I mean we only have this problem for independent scales (actually not shared scale -- typo*) as shared scale would be moved to parent's scope.  

Currently, I plan to work on merging size signals first (#2198), which hopefully will make it easier for me to fix this one.

kanitw:
> Pushing to outer scope is fairly easy. We do it with selections. Signals should have the same name, and the inner signals should be marked with "push": "outer".

Yep.  Shouldn't be too bad

[...]

PR #2565, opened 2017-06-24
[WIP]Do not always move signals to the top-level

kanitw:
- [ ] Need to add duplicate signal and do push: outer  (or move scales outside)

Try to address #2555


kanitw:
I'll handle this differently in another PR.

PR #2567, opened 2017-06-24
Correctly merge legend properties & legend of the same field

### LINKED ISSUE #1844: Combine legends based on the same underlying variables into a single legend
For example, if we over-encode both size and color with the same variables, this should create one legend, rather than two.

I'm creating this to replace #821 to start a clean slate.  (The old #821 contains workaround while we haven't fixed this problem though.)

### LINKED ISSUE #2473: Correctly merge axis/legend
Forked from #2251 -- so #2251 can focus on scale 

- [x] Combine Title of the merged guide
- [x] Deal with grid_scale case https://github.com/vega/vega-lite/issues/2242
- [x] https://github.com/vega/vega-lite/issues/2334

kanitw:
Fix #2473 and Fix #1844

### DIFF
--- a/examples/specs/layered_color_legend_left.vl.json
+++ b/examples/specs/layered_color_legend_left.vl.json
@@ -1,0 +1,18 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "Stock price over time with color legend orient = left.",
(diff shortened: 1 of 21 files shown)


kanitw:
vega_editor

vega_editor

domoritz:
Bautiful!

PR #2570, opened 2017-06-25
fix table style to target on page content

sirahd:


domoritz:
Hmm, what do you think about adding https://developer.mozilla.org/en-US/docs/Web/CSS/isolation to the tooltip css? I think may people override tables.

Also, why do we use tables in the tooltips? Can we avoid that?

domoritz:
As explained above, I think a table is the wrong approach for tooltips. https://github.com/vega/vega-tooltip/issues/115

kanitw:
I disagree.  They are semantically table.

kanitw:
Plus, while people may override table, doing so is a bad practice anyway so I'm gonna merge this.

kanitw:
Btw, I don't think the `isolation` property is relevant here.

domoritz:
Hmm, I thought is isolates case scopes but I've never used it.

PR #2576, opened 2017-06-26
Triggers should register as scale dependencies not listeners.

arvind:
This prevents a glitch observed when brushing and panning within the same unit view. If triggers are registered as listeners to multiple scales, the brush will appear to "drift" as one scale registers and
fires an update before the other.


kanitw:
forgot to update unit test?

arvind:
Indeed I did. Updated now.

PR #2581, opened 2017-06-27
Sort facet values in ascending order as default

### LINKED ISSUE #2553: Facet labels are mismatched.
image

```json
{
    "data":{
        "values": [
            {"f1":"LL","f2":"A","f3":"L","f4":"++"},
            {"f1":"HM","f2":"C","f3":"H","f4":"-"},
            {"f1":"ML","f2":"C","f3":"L","f4":"-"},
            {"f1":"LM","f2":"B","f3":"X","f4":"-"}
        ]
    },
    "mark": "point",
    "encoding": {
        "x":{"field": "f1", "type": "nominal"},
        "y":{"field": "f4", "type": "nominal"},
        "row":{"field": "f3", "type": "nominal"},
        "column":{"field": "f2", "type": "nominal"},
        "size":{"field": "*", "aggregate":"count", "type": "quantitative"}
    }
}
```
As you can see, labels on `row` and `column` are incorrect. C should be swapped with B, and  H should be swapped with L.

yhoonkim:
Fix #2553 

(As a future work, I think we should allow user to sort facet as they want.)

### DIFF
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -103,7 +103,9 @@ export class FacetModel extends ModelWithField {
 
     // if we facet by two dimensions, we need to add a cross operator to the aggregation
     // so that we create all groups
-    const isCrossedFacet = this.channelHasField(ROW) && this.channelHasField(COLUMN);
+    const hasRow = this.channelHasField(ROW);
+    const hasColumn = this.channelHasField(COLUMN);
+
     this.component.mark = [{
       name: this.getName('cell'),
       type: 'group',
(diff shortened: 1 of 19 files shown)


src/compile/facet.ts
```diff
@@ -112,18 +114,24 @@ export class FacetModel extends ModelWithField {
...
+            hasColumn ? [this.field(COLUMN)] : []
           ),
-          ...(isCrossedFacet ? {aggregate: {
+          ...(hasRow && hasColumn ? {aggregate: {
             cross: true
           }}: {})
         }
       },
-      ...(isCrossedFacet ? {sort: {
+      ...(hasRow && hasColumn ? {sort: {
```
kanitw: You always produce `sort` and `order` so writing it this way is better:

```
{
  sort: {
    field: [].concat(
      hasRow: ? [this.field(ROW, {expr: 'datum'})] : [],
      hasColumn: ? [this.field(COLUMN, {expr: 'datum'})] : []
    )
    order: [].concat(
      hasRow: ? [ (this.facet.row.header && this.facet.row.header.sort) || 'ascending'] : [],
      hasColumn: ? [ ... ] : []
    )
}
```

src/compile/layout/header.ts
```diff
@@ -122,7 +124,13 @@ export function getHeaderGroup(model: Model, channel: HeaderChannel, headerType:
         name: model.getName(`${channel}_${headerType}`),
         type: 'group',
         role: `${channel}-${headerType}`,
-        ...(layoutHeader.fieldRef ? {from: {data: model.getName(channel)}} : {}),
+        ...(layoutHeader.facetFieldDef ? {
+          from: {data: model.getName(channel)},
+          sort: {
+            field: field(layoutHeader.facetFieldDef, {expr: 'datum'}),
+            order: 'ascending'
```
kanitw: (facetFieldDef.header && facetFieldDef.header.sort) || 'ascending'

kanitw:
I just fix a minor thing in the declaration.  Let's merge when test finishes running.

Issue #2579, opened 2017-06-26
Running build:images leads to "invalid format: null"


kanitw:
This might be relevant to the default format change

PR #2588, opened 2017-06-27
Fix `format` to use `''` instead of `null`

kanitw:
Fix https://github.com/vega/vega-lite/issues/2579


domoritz:
I thought I did that. Weird. 

Thanks for the fix!

kanitw:
lol  you actually changed `''` to `null` https://github.com/vega/vega-lite/pull/2510#discussion_r122355624 -- I think the comment in https://github.com/vega/vega-lite/issues/1539#issuecomment-303504194 really suggests null though but apparently it fails :p

PR #2589, opened 2017-06-27
Coerce interval extents to numbers for scale trigger test.

arvind:
This accounts for JS Date objects.


kanitw:
lol,  I just learn from this PR that `+(new Date())` works.  JS is so quirky!

PR #2590, opened 2017-06-27
Grouped bar chart vega-lite spec example

churtado:
This is not a fix for any issue, just submitting an example spec for a grouped bar chart. The data is embedded in the spec, so it's just one file.


kanitw:
Thanks a lot-- @domoritz just updated the website (for the next release) to include your example as a community example in https://github.com/vega/vega-lite/issues/2594.

churtado:
Thank you very much!

PR #2593, opened 2017-06-28
Customize brush style using "mark" property

arvind:
This PR introduces a `mark` property that can be defined on interval selections (and interval selection configs) to customize the appearance of the generated brush mark. The supported style properties include `fill`, `fillOpacity`, `stroke`, `strokeWidth`, `strokeOpacity`, `strokeDash`, and `strokeDashOffset`.

(I also made the type definitions more precise, which should help make Altair's selection API cleaner).

E.g., 

```json
{
  "selection": {
    "brush": {
      "type": "interval",
      "mark": {
        "fill": "red",
        "fillOpacity": 0.75,
        "stroke": "green",
        "strokeWidth": 4,
        "strokeDash": [10, 5],
        "strokeDashOffset": 3,
        "strokeOpacity": 1
      }
    }
  },
  "mark": "point",
  "encoding": {
    "color": {
      "condition": {"selection": "brush", "field": "Cylinders", "type": "ordinal"},
      "value": "grey"
    }
  }
}
```

download 3

### DIFF
--- a/src/compile/selection/interval.ts
+++ b/src/compile/selection/interval.ts
@@ -7,7 +7,6 @@ import {UnitModel} from '../unit';
 import {channelSignalName, ProjectComponent, SelectionCompiler, SelectionComponent, STORE, TUPLE} from './selection';
 import scales from './transforms/scales';
 
-
 export const BRUSH = '_brush';
 export const SCALE_TRIGGER = '_scale_trigger';
 
@@ -113,7 +112,10 @@ const interval:SelectionCompiler = {
     // not interefere with the core marks, but that the brushed region can still
     // be interacted with (e.g., dragging it around).
     const {fill, fillOpacity, ...stroke} = selCmpt.mark;
-    keys(stroke).forEach((k) => stroke[k] = {value: stroke[k]});
+    const vgStroke = keys(stroke).reduce((def, k) => {
+      def[k] = {value: stroke[k]};
+      return def;
+    }, {});
 
     return [{
       type: 'rect',
@@ -130,7 +132,7 @@ const interval:SelectionCompiler = {
       encode: {
         enter: {
           fill: {value: 'transparent'},
-          ...stroke
+          ...vgStroke
         },
         update: update
       }
--- a/src/compile/selection/interval.ts
+++ b/src/compile/selection/interval.ts
@@ -112,12 +112,15 @@ const interval:SelectionCompiler = {
     // Two brush marks ensure that fill colors and other aesthetic choices do
     // not interefere with the core marks, but that the brushed region can still
     // be interacted with (e.g., dragging it around).
+    const {fill, fillOpacity, ...stroke} = selCmpt.mark;
+    keys(stroke).forEach((k) => stroke[k] = {value: stroke[k]});
+
     return [{
       type: 'rect',
       encode: {
         enter: {
-          fill: {value: '#333'},
-          fillOpacity: {value: 0.125}
+          fill: {value: fill},
+          fillOpacity: {value: fillOpacity}
         },
         update: update
       }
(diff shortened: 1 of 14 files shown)


src/selection.ts
```diff
@@ -1,35 +1,55 @@
...
   // predicate?: string;
-  bind?: 'scales' | VgBinding | {[key: string]: VgBinding};
+  // domain?: SelectionDomain;
 
   // Transforms
   fields?: string[];
   encodings?: SingleDefChannel[];
+}
+
+export interface BaseSingleSelectionDef extends BaseSelectionDef {
```
kanitw: Should we name this `SingleSelectionConfig` so it shows up in the schema as `SingleSelectionConfig`?  (Same for multi/interval)

src/selection.ts
```diff
@@ -1,35 +1,55 @@
...
   'union_others' | 'intersect' | 'intersect_others';
 
 export interface BaseSelectionDef {
-  // domain?: SelectionDomain;
-  resolve?: SelectionResolutions;
   on?: any;
+  resolve?: SelectionResolutions;
   // predicate?: string;
-  bind?: 'scales' | VgBinding | {[key: string]: VgBinding};
+  // domain?: SelectionDomain;
```
kanitw: Why do we have some commented properties here?

src/compile/selection/transforms/toggle.ts
```diff
@@ -7,7 +7,8 @@ const TOGGLE = '_toggle';
 
 const toggle:TransformCompiler = {
   has: function(selCmpt) {
-    return selCmpt.toggle !== undefined && selCmpt.toggle !== false;
+    return selCmpt.type === 'multi' &&
```
kanitw: > `selCmpt.toggle !== undefined && selCmpt.toggle !== false;` 

so `null` behaves differently from `undefined/false`?

src/compile/selection/transforms/nearest.ts
```diff
@@ -1,11 +1,11 @@
 import {TransformCompiler} from './transforms';
 
-
 const VORONOI = 'voronoi';
 
 const nearest:TransformCompiler = {
   has: function(selCmpt) {
-    return selCmpt.nearest !== undefined && selCmpt.nearest !== false;
+    return selCmpt.type !== 'interval' &&
```
kanitw: so `null` behaves differently from `undefined/false`?

src/compile/selection/interval.ts
```diff
@@ -112,12 +112,15 @@ const interval:SelectionCompiler = {
     // Two brush marks ensure that fill colors and other aesthetic choices do
     // not interefere with the core marks, but that the brushed region can still
     // be interacted with (e.g., dragging it around).
+    const {fill, fillOpacity, ...stroke} = selCmpt.mark;
+    keys(stroke).forEach((k) => stroke[k] = {value: stroke[k]});
```
kanitw: I think this is actually breaking the type (as this line change stroke's value to be a valueRef). Maybe in the future, TS will notice and leads to compile error when we upgrade.  

Maybe we can do  this below instead

```
... (keys(stroke).reduce((vgStroke, k) => {
   vgStroke[k] = {value: stroke[k]};
   return vgStroke; 
}, {}); 
```

src/selection.ts
arvind: Because they're technically part of our model (as described in the paper) but not something we support yet in 2.0. I wanted to leave it in, but commented, to remind us should we consider them in the future.

kanitw: Okie. Maybe create an issue and add TODO link to the issue too?
(More likely that we will run into the issue than a commented code in the future.)

arvind:
@kanitw: Requested changes made!

PR #2619, opened 2017-07-02
Typescript 2.4 and a lot of fixes. 

domoritz:
Thanks to stronger typings, I found a large number of small bugs. You can see that I kept fixes in separate commits. Please do not squash.

### DIFF
--- a/src/compile/mark/mixins.ts
+++ b/src/compile/mark/mixins.ts
@@ -36,7 +36,7 @@ export function markDefProperties(mark: MarkDef, props: (keyof MarkDef)[]) {
   }, {});
 }
 
-export function valueIfDefined(prop: string, value: string): VgEncodeEntry {
+export function valueIfDefined(prop: string, value: string | number | boolean): VgEncodeEntry {
   if (value !== undefined) {
     return {[prop]: {value: value}};
   }
--- a/src/compile/mark/mixins.ts
+++ b/src/compile/mark/mixins.ts
@@ -36,7 +36,7 @@ export function markDefProperties(mark: MarkDef, props: (keyof MarkDef)[]) {
   }, {});
 }
 
-export function valueIfDefined(prop: string, value: VgValueRef): VgEncodeEntry {
+export function valueIfDefined(prop: string, value: string): VgEncodeEntry {
   if (value !== undefined) {
     return {[prop]: {value: value}};
   }
--- a/src/compile/repeat.ts
+++ b/src/compile/repeat.ts
@@ -32,7 +32,7 @@ export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: R
   return replaceRepeater(encoding, repeater);
 }
 
-type EncodingOrFacet<F> = Encoding<F> | Facet<F>;
+type EncodingOrFacet<F> = Encoding<F> & Facet<F>;
 
 /**
  * Replace repeater values in a field def with the concrete field name.
--- a/src/compile/repeat.ts
+++ b/src/compile/repeat.ts
@@ -25,14 +25,12 @@ export type RepeaterValue = {
 };
 
 export function replaceRepeaterInFacet(facet: Facet<Field>, repeater: RepeaterValue): Facet<string> {
-  return replaceRepeater(facet, repeater);
+  return replaceRepeater(facet, repeater) as Facet<string>;
 }
 
 export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: RepeaterValue): Encoding<string> {
-  return replaceRepeater(encoding, repeater);
+  return replaceRepeater(encoding, repeater) as Encoding<string>;
 }
-
-type EncodingOrFacet<F> = Encoding<F> & Facet<F>;
 
 /**
  * Replace repeater values in a field def with the concrete field name.
@@ -54,6 +52,8 @@ function replaceRepeaterInFieldDef(fieldDef: FieldDef<Field>, repeater: Repeater
     return fieldDef as FieldDef<string>;
   }
 }
+
+type EncodingOrFacet<F> = Encoding<F> | Facet<F>;
 
 function replaceRepeater(mapping: EncodingOrFacet<Field>, repeater: RepeaterValue): EncodingOrFacet<string> {
   const out: EncodingOrFacet<string> = {};
--- a/src/spec.ts
+++ b/src/spec.ts
@@ -310,6 +310,8 @@ function normalizeFacetedUnit(spec: FacetedCompositeUnitSpec, config: Config): F
     },
     spec: normalizeNonFacetUnit({
       mark,
+      ...(width ? {width} : {}),
+      ...(height ? {height} : {}),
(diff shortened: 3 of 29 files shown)


src/compile/mark/mixins.ts
```diff
@@ -36,7 +36,7 @@ export function markDefProperties(mark: MarkDef, props: (keyof MarkDef)[]) {
   }, {});
 }
 
-export function valueIfDefined(prop: string, value: VgValueRef): VgEncodeEntry {
+export function valueIfDefined(prop: string, value: string): VgEncodeEntry {
```
kanitw: value doesn't really have to be string -- number or boolean are also fine.

src/compile/repeat.ts
```diff
@@ -32,7 +32,7 @@ export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: R
   return replaceRepeater(encoding, repeater);
 }
 
-type EncodingOrFacet<F> = Encoding<F> | Facet<F>;
+type EncodingOrFacet<F> = Encoding<F> & Facet<F>;
```
kanitw: This fix seems incorrect?  Why is it "and"?

src/spec.ts
```diff
@@ -420,6 +420,7 @@ function normalizeOverlay(spec: UnitSpec, overlayWithPoint: boolean, overlayWith
         filled: true,
         role: 'pointOverlay'
       },
+      ...(selection ? {selection} : {}),
```
kanitw: Why only adding selection to the point layer?

test/compile/scale/type.test.ts
```diff
@@ -11,19 +11,21 @@ import {TimeUnit} from '../../../src/timeunit';
 import {NOMINAL, ORDINAL} from '../../../src/type';
 import * as util from '../../../src/util';
 
+const scaleConfig = defaultConfig.scale;
```
kanitw: defaultScaleConfig

src/vega.schema.ts
```diff
@@ -161,7 +161,7 @@ export type VgSignal = {
 
 export type VgEncodeChannel = 'x'|'x2'|'xc'|'width'|'y'|'y2'|'yc'|'height'|'opacity'|'fill'|'fillOpacity'|'stroke'|'strokeWidth'|'strokeOpacity'|'strokeDash'|'strokeDashOffset'|'cursor'|'clip'|'size'|'shape'|'path'|'innerRadius'|'outerRadius'|'startAngle'|'endAngle'|'interpolate'|'tension'|'orient'|'url'|'align'|'baseline'|'text'|'dir'|'ellipsis'|'limit'|'dx'|'dy'|'radius'|'theta'|'angle'|'font'|'fontSize'|'fontWeight'|'fontStyle';
 export type VgEncodeEntry = {
-  [k in VgEncodeChannel]?: VgValueRef | (VgValueRef & {test: string})[];
+  [k in VgEncodeChannel]?: VgValueRef | (VgValueRef | {test: string})[];
```
kanitw: This actually makes it incorrect!  The old was correct.

src/spec.ts
```diff
@@ -464,17 +464,10 @@ function fieldDefIndex<T>(spec: GenericSpec<GenericUnitSpec<any, any>>, dict: Di
...
   } else if (isConcatSpec(spec)) {
     const childSpec = isVConcatSpec(spec) ? spec.vconcat : spec.hconcat;
-    childSpec.forEach(child => {
-      if (isUnitSpec(child)) {
-        accumulate(dict, vlEncoding.fieldDefs(child.encoding));
-      } else {
-        fieldDefIndex(child, dict);
-      }
-    });
+    childSpec.forEach(child => fieldDefIndex(child, dict));
```
kanitw: Can you explain a bit more about this change?

src/compile/mark/mixins.ts
domoritz: It's not needed. I can add them but we can also add it when ts complains. VgValueRef was definitely wrong.

src/spec.ts
domoritz: How would the selection work on the line layer?

src/spec.ts
domoritz: The goal of this function is to collect the field defs. We can do this recursively so why not just go into the children instead of special casing for different children?

test/compile/scale/type.test.ts
domoritz: Done

src/vega.schema.ts
domoritz: The old one didn't work, though. Can you say what this is supposed to do or checkout the branch and try to correct this for yourself.

src/compile/repeat.ts
domoritz: We want an intersection type, which is **not** an intersection of the fields of those types.

```ts
interface A {
  a: number;
  b: number;
}

interface B {
  b: number;
  c: number;
}

const o: A & B;
```

Also, see https://twitter.com/domoritz/status/858899054507442176

src/vega.schema.ts
domoritz: I think that you might misunderstand intersection and union types.

kanitw: > I think that you might misunderstand intersection and union types.

Nope.  You actually might not understand how production rule works in Vega.  

I believe the correct fix is `(VgValueRef & {test?: string})[];`. As `test` is optional for the last clause -- but each element is not either `{test: string}` or `VgValueRef` -- each element must be VgValueRef and may have test.

src/spec.ts
kanitw: If the selection is on the categories (not individual points) then selection over the line is sensible.   (Maybe we need to add more condition here.)

src/compile/repeat.ts
kanitw: I already know about your tweet.  

That said, intersection type ("and") is not the thing we really want here. 
The right thing semantically would be 

`EncodingWithFacet<F> | Facet<F>`  as we will take mapping from either unit spec (`EncodingWithFacet`) or from facet spec (`Facet`).   

It just happens that all properties of Encoding and Facet are optional and thus `Encoding<F> & Facet<F>;`  makes TS stop complaining although it is not semantically correct.

src/spec.ts
kanitw: Thanks

src/compile/repeat.ts
domoritz: The problem is that I can't specify it differently because I cannot use generics as type parameters. I used `as` to make it work.

src/vega.schema.ts
domoritz: Fixed.

src/spec.ts
domoritz: Do you have an example?

kanitw: say you have colored multi-series line chart  and selection is projected over color.

domoritz: Okay, will add selection to both layers for now.

PR #2627, opened 2017-07-04
Use pan/zoom expression functions

arvind:
This PR updates the translate and zoom selection transforms to use corresponding expression functions (rather than embed the necessary calculation within the signal update function). We added these functions in vega/vega-parser@4d82d86 to support smoother panning/zooming of log and pow scales. Consequently, I've also added pan/zooming to the gapminder example!


domoritz:
Much smoother. I just tried it and it's awesome.

arvind:
@domoritz: Merged your PR #2628. Thanks! This PR should now be ready for merging as well.

PR #2632, opened 2017-07-06
Dom/typings

domoritz:


typings/vega-lite.d.ts
```diff
@@ -1,4 +1,3 @@
-declare module vl {
-  export var version: string;
-  export function compile(inputSpec: any, logger?: any): any;
-}
+export as namespace vl;
```
kanitw: This file is weird.  It seems a bit recursive.  but I guess we need it for docs & gallery.

domoritz: Not just that. Notice how I am copying it to build as well. It will be useful for people who include the bundle instead of using a bunder.

domoritz:
How can the schema change in this pr? https://travis-ci.org/vega/vega-lite/jobs/250828978

PR #2636, opened 2017-07-06
Add identifier transform for default point selections.

### LINKED ISSUE #2221: Eliminate _id reference from the Vega output
From discussion with Jeff today, this sounds like something we have to do before the final 2.0 release.  (This is okay for beta though.)

arvind:
Fixes #2221. Default point selections are projected over an ID field. However, the default `_id` is volatile -- if the spec is parsed again, different IDs may be generated based on global Vega state. As a result, breaks the ability to snapshot state via view.getState/setState.

With vega/vega-view#10, an identifier transform was introduced which persists its state in an internal signal. This transform is added whenever new tuples are derived (i.e., at the source and post-aggregation).

### DIFF
--- a/src/compile/data/assemble.ts
+++ b/src/compile/data/assemble.ts
@@ -1,6 +1,7 @@
 import {isUrlData, MAIN} from '../../data';
 import {every, flatten, vals} from '../../util';
 import {VgData} from '../../vega.schema';
+import {Model} from '../model';
 import {DataComponent} from './';
 import {AggregateNode} from './aggregate';
 import {BinNode} from './bin';
@@ -9,16 +10,20 @@ import {FacetNode} from './facet';
 import {ParseNode} from './formatparse';
 import {NonPositiveFilterNode} from './nonpositivefilter';
 import {NullFilterNode} from './nullfilter';
-import {iterateFromLeaves} from './optimizers';
 import * as optimizers from './optimizers';
+import {iterateFromLeaves} from './optimizers';
 import {OrderNode} from './pathorder';
 import {SourceNode} from './source';
 import {StackNode} from './stack';
 import {TimeUnitNode} from './timeunit';
 import {CalculateNode, FilterNode, LookupNode} from './transforms';
-
+// tslint:disable-next-line:ordered-imports
+// This tslint rule causes a circular dependency here for some reason.
+import {addIdentifierTransforms} from '../selection/selection';
 
 export const FACET_SCALE_PREFIX = 'scale_';
+
+export const SOURCE_PREFIX = 'source_';
 
 /**
  * Start optimization path from the root. Useful for removing nodes.
@@ -307,8 +312,8 @@ export function assembleFacetData(root: FacetNode): VgData[] {
  * @param  data array
  * @return modified data array
  */
-export function assembleData(dataCompomponent: DataComponent): VgData[] {
-  let roots: SourceNode[] = vals(dataCompomponent.sources);
+export function assembleData(model: Model, dataComponent: DataComponent): VgData[] {
+  let roots: SourceNode[] = vals(dataComponent.sources);
   const data: VgData[] = [];
 
   roots.forEach(removeUnnecessaryNodes);
@@ -331,7 +336,7 @@ export function assembleData(dataCompomponent: DataComponent): VgData[] {
   roots.forEach(root => {
     // assign a name if the source does not have a name yet
     if (!root.hasName()) {
-      root.dataName = `source_${sourceIndex++}`;
+      root.dataName = `${SOURCE_PREFIX}${sourceIndex++}`;
     }
 
     const newData: VgData = root.assemble();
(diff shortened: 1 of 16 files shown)


arvind:
Ping @domoritz.

src/compile/selection/selection.ts
```diff
@@ -202,6 +212,32 @@ export function assembleLayerSelectionMarks(model: LayerModel, marks: any[]): an
...
+  let identifier = false;
+  forEachSelection(model, (selCmpt) => {
+    identifier = identifier || selCmpt.project.some((proj) => proj.field === SELECTION_ID);
+  });
+
+  if (!identifier) {
+    return data;
+  }
+
+  data.forEach((def) => {
```
domoritz: We prefer `for (const def of data) {}`

src/compile/data/assemble.ts
```diff
@@ -307,8 +312,8 @@ export function assembleFacetData(root: FacetNode): VgData[] {
  * @param  data array
  * @return modified data array
  */
-export function assembleData(dataCompomponent: DataComponent): VgData[] {
-  let roots: SourceNode[] = vals(dataCompomponent.sources);
+export function assembleData(model: Model, dataComponent: DataComponent): VgData[] {
```
domoritz: Can we avoid the model as a parameter to this method?

src/compile/selection/selection.ts
```diff
@@ -202,6 +212,32 @@ export function assembleLayerSelectionMarks(model: LayerModel, marks: any[]): an
...
+    let index = -1;
+    const transform = def.transform || [];
+    const aggr = transform.some((t, i) => (index = i, t.type === 'aggregate'));
```
domoritz: What is happening here?

arvind: I'm emulating [`array.findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) which we don't seem to have typings for unfortunately. We need to add an `identifier` transform to the `source` dataset and any time new tuples are derived (which, with Vega-Lite, I think only happens after an `aggregate`).

src/compile/data/assemble.ts
arvind: Er, not really... It's passed to the `addIdentifierTransforms` method called at the end of this function. The `model` is needed to check the selection component to verify whether `identifier` transforms are actually needed or not (i.e., we do not add them if all selections are projected over other fields).

src/compile/selection/selection.ts
arvind: Ok, fixed.

src/compile/data/assemble.ts
arvind: Unless, of course, there's some other way I can get access to the model's selection components down this code path!

domoritz: Can we pass just the selection component? 

If not, I'd rather pass the whole model to this method.

src/compile/selection/selection.ts
domoritz: Why don't we do that when we parse data?

src/compile/data/assemble.ts
arvind: I'm not sure I understand what you're proposing. Would you rather the selection component be passed to `assembleData`. That seems awfully brittle, but I'm probably misunderstanding.

domoritz: > Would you rather the selection component be passed to assembleData

Yes, or just pass the whole model. But again, I am opposed to adding dataflow nodes in assemble. 

Can we add a dataflow node for selections that assembled the required transforms? We would add a node for each selection and then only actually insert the transform if it is needed.

src/compile/selection/selection.ts
arvind: This seems simpler. Identifier transforms are only needed to support default selection definitions. Doing it at parse time would require building out all the necessary dataflow architecture to support it, whereas this encapsulates that need in a single function.

src/compile/data/assemble.ts
```diff
@@ -9,17 +10,21 @@ import {FacetNode} from './facet';
...
 import * as optimizers from './optimizers';
+import {iterateFromLeaves} from './optimizers';
 import {OrderNode} from './pathorder';
 import {SourceNode} from './source';
 import {StackNode} from './stack';
 import {TimeUnitNode} from './timeunit';
 import {CalculateNode, FilterNode, LookupNode} from './transforms';
-
+// tslint:disable-next-line:ordered-imports
+// This tslint rule causes a circular dependency here for some reason.
```
domoritz: Can you try to resolve this? The order of imports really shouldn't matter.

src/compile/selection/selection.ts
domoritz: This is just inconsistent with how we do create the dataflow. We do add nodes to the dataflow and then in assemble, we create the vega data sources. I favor consistency over simplicity.

arvind: Ok. It sounds like this PR will not be accepted so I will close and open a new one.

PR #2637, opened 2017-07-06
Better error messaging for unrecognized selection names.

### LINKED ISSUE #2635: Uncaught TypeError: Cannot read property 'getSelectionComponent' of null
screen shot 2017-07-06 at 10 27 22

```json
{
  "data": {"url": "data/europe_world_bank.csv"},
  "hconcat": [
    {
      "repeat": {
        "column": ["Population_ages_15_64_of_total", "Population_ages_65_and_above_of_total"]
      },
      "spec": {
        "mark": "bar",
        "height": 700,
        "selection": {"brush": {"type": "interval", "encodings": ["y"]}},
        "encoding": {
          "y": {
            "field": "Country",
            "type": "ordinal",
            "sort": {
              "op": "min",
              "field": "Population_ages_15_64_of_total",
              "order": "descending"
            }
          },
          "x": {
            "field": "Population_ages_15_64_of_total",
            "type": "quantitative"
          },
          "color": {
            "condition": {"selection": "!brush", "value": "steelblue"},
            "value": "goldenrod"
          }
        }
      }
    },
    {
      "mark": "point",
      "width": 260,
      "height": 260,
      "selection": {"brush": {"type": "interval"}},
      "encoding": {
        "y": {
          "field": "Population_ages_15_64_of_total",
          "type": "quantitative",
          "scale": {"zero": false}
        },
        "x": {
          "field": "Population_ages_65_and_above_of_total",
          "type": "quantitative",
          "scale": {"zero": false}
        },
        "color": {
          "condition": {"selection": "!brush", "value": "steelblue"},
          "value": "goldenrod"
        }
      }
    }
  ]
}
```

arvind:
Fixes #2635.

### DIFF
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -5,6 +5,7 @@ import {Data, DataSourceType, MAIN, RAW} from '../data';
 import {forEach, reduce} from '../encoding';
 import {ChannelDef, field, FieldDef, FieldRefOption, getFieldDef, isFieldDef, isRepeatRef} from '../fielddef';
 import {Legend} from '../legend';
+import * as log from '../log';
 import {ResolveMapping} from '../resolve';
 import {hasDiscreteDomain, Scale} from '../scale';
 import {SortField, SortOrder} from '../sort';
@@ -426,8 +427,15 @@ export abstract class Model {
   /**
    * Traverse a model's hierarchy to get a particular selection component.
    */
-  public getSelectionComponent(name: string): SelectionComponent {
-    return this.component.selection[name] || this.parent.getSelectionComponent(name);
+  public getSelectionComponent(varName: string, origName: string): SelectionComponent {
+    let sel = this.component.selection[varName];
+    if (!sel && this.parent) {
+      sel = this.parent.getSelectionComponent(varName, origName);
+    }
+    if (!sel) {
+      throw new Error(log.message.selectionNotFound(origName));
+    }
+    return sel;
   }
 }
 
--- a/src/compile/selection/selection.ts
+++ b/src/compile/selection/selection.ts
@@ -213,9 +213,9 @@ const PREDICATES_OPS = {
 
 export function predicate(model: Model, selections: LogicalOperand<string>): string {
   function expr(name: string): string {
-    name = varName(name);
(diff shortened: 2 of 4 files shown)


domoritz:
Sweet. Can you add a test for it?

arvind:
Done!

domoritz:
Thanks! We use a different pattern for testing warnings in Vega-Lite. See https://github.com/vega/vega-lite/blob/master/test/fielddef.test.ts#L49 for example

```ts
it('should warn.', log.wrap((localLogger) => {
  foo();
  assert.equal(localLogger.warns[0], log.message.something(param));
}));
```

arvind:
Gotcha. But this PR isn't actually issuing a warning (`log.warn`) but rather throwing an error (`throw new Error`). If an unknown selection is provided, compilation needs to abort. Throwing errors (rather than warnings) seems to happen across the compiler but I could only find one [explicit test](https://github.com/vega/vega-lite/blob/master/test/compile/compile.test.ts#L12) of it. 

How would you like me to proceed?

domoritz:
How difficult would it be to completely ignore the selection and throw a warning instead? We try to be as forgiving as possible as resort to some reasonable alternative. However, in this particular case, I think it is okay to abort the compilation completely.

arvind:
If we wanted compilation to proceed with invalid selection names, it'd involve non-trivial additional code paths to test for empty predicate expression strings (or partial strings in the case of logical composition) and then remove the containing part of the specification. This is because empty expression strings are invalid (and throw errors) in Vega.

For example, both of the following Vega snippets (which Vega-Lite could potentially generate for an invalid selection name) throw errors. Based on the additional necessary logic to remove `FilterNodes` and branches of production rules, my preference would be to just throw an error and abort compilation.

```json
"transform": [{"type": "filter", "expr": ""}]
```

```json
"update": {
          "x": [
            {
              "test": "data(\"brush_store\").length && data(\"brush_store\")[0].unit === \"\"",
              "signal": "brush_x[0]"
            },
            {"value": 0}
          ],
```

domoritz:
Agreed. Thanks for the explanation.

PR #2639, opened 2017-07-06
Fix scale trigger coercion to only occur for continuous scales.

arvind:
#2589 coerced interval extents to numbers to support time scales. However, this [breaks with `band`](https://github.com/vega/vega-lite/issues/2635#issuecomment-313534244) (and presumably other discrete scale types). Thus, this PR only coerces for continuous scale types.


domoritz:
Thanks for the quick fix!

PR #2642, opened 2017-07-07
Make resolve a part of component & set default resolve in parse

kanitw:
Fix #2642

- This allows us to always have information whether an axis is merged (shared)
- Add compile/resolve.ts  (Instead of declaring them directly outside)
- I also make the scale resolve rule for concat and repeat consistent

This is a prep for making layout size merging correct.  
(If there are minor fixes todo, feel free to fix them as there might be some delay before I get to fix them.)


domoritz:
Rebased and fixed a bug

PR #2644, opened 2017-07-07
Add build:examples-quick, which is way faster for running build:examples

kanitw:
I know we already have too many commands but this is extremely faster and drain way more less battery when I work on a plane. So I'll definitely add it.   (Feel free to further optimize though.)


domoritz:
It doesn't run postbuild, right?

kanitw:
yeah that's why it's superfast

PR #2656, opened 2017-07-12
Add identifier transform for default point selections.

arvind:
This PR replaces #2636, adding the identifier transform via the dataflow architecture instead. 

#2636 describes our desired outcome: an identifier node added every time new tuples are derived -- i.e., at the source and after an aggregate transform (and if the two happen within the same dataset, only one identifier transform is added to reduce unnecessary computation).

Here's a simple unit spec to start testing with:

```json
{
  "description": "Select multiple points with the shift key.",
  "selection": {
    "paintbrush": {
      "type": "single"
    }
  },
  "mark": "point",
  "encoding": {
    "size": {
      "condition": {"selection": "paintbrush", "value": 300},
      "value": 50
    }
  }
}
```

Currently, while the identifier node is currently added, a number of things seem to happen during assembly/optimization that yield buggy or less ideal output:
1. When raw data values are embedded directly in the spec, the identifier transform is added to a new derived dataset from which all other datasets (e.g., in a layered or concatenated view) are then subsequently derived. The identifier transform should instead be merged up into the source dataset.
2. With a concatenated display, an identifier transform is correctly added to the `source` dataset, but also appears after parse transforms in last derived dataset.
3. D'oh, obvious error in the "problematic" spec. Once the selection is moved to the aggregate spec, the identifier transform appears correctly. ~~Aggregate plots, I'm unsure if identifier transforms are being dropped somewhere in the logic or whether there are just additional locations there are created from that I haven't tracked down.~~ 

    ~~For instance, with a simple aggregate plot, the identifier transform is appended correctly..~~

    ```json
    {
      "mark": "circle",
      "selection": {
        "pt": {"type": "single"}
      },
      "encoding": {
        "x": {"field": "IMDB_Rating", "type": "Q", "bin": true},
        "y": {"field": "Rotten_Tomatoes_Rating", "type": "Q", "bin": true},
        "size": {"aggregate": "count", "type": "Q"}
      }
    }
    ```

    ~~However, with a concatenated plot, the post-aggregate identifier disappears. Initially, I suspected this may be related to point (2) above (e.g., perhaps the identifier from the first dataset is being lifted to the source). But, you see it even when concatenating arbitrary numbers of normal plots and aggregated plots. I've been having trouble debugging further due to the unexpected ordering of datasets/concatenations.~~

    ```json
    {
      "hconcat": [{
        "selection": {
          "pt": {"type": "single"}
        },
        "mark": "point",
        "encoding": {
          "size": {
            "condition": {"selection": "pt", "value": 300},
            "value": 50
          }
        }
      }, {
        "mark": "point",
        "encoding": {
          "size": {
            "condition": {"selection": "pt", "value": 300},
            "value": 50
          }
        }
      }, {
        "mark": "point",
        "encoding": {
          "x": {"field": "Horsepower", "type": "quantitative", "bin": true},
          "y": {"field": "Miles_per_Gallon", "type": "quantitative", "aggregate": "count"},
          "size": {
            "condition": {"selection": "pt", "value": 300},
            "value": 50
          }
        }
      }]
    }
    ```
   
@domoritz: Given your deeper familiarity with the Vega-Lite dataflow architecture, I'm wondering whether you could take over? I'm a little worried about being a bull in a china shop here!

### DIFF
--- a/examples/vg-specs/paintbrush.vg.json
+++ b/examples/vg-specs/paintbrush.vg.json
@@ -29,6 +29,10 @@
             },
             "transform": [
                 {
+                    "type": "identifier",
+                    "as": "_vgsid_"
+                },
+                {
                     "type": "filter",
                     "expr": "datum[\"Horsepower\"] !== null && !isNaN(datum[\"Horsepower\"]) && datum[\"Miles_per_Gallon\"] !== null && !isNaN(datum[\"Miles_per_Gallon\"])"
                 }
@@ -65,7 +69,7 @@
                             "type": "mouseover"
                         }
                     ],
-                    "update": "datum && item().mark.marktype !== 'group' ? {unit: \"\", encodings: [], fields: [\"_id\"], values: [(item().isVoronoi ? datum.datum : datum)[\"_id\"]]} : null"
+                    "update": "datum && item().mark.marktype !== 'group' ? {unit: \"\", encodings: [], fields: [\"_vgsid_\"], values: [(item().isVoronoi ? datum.datum : datum)[\"_vgsid_\"]]} : null"
                 }
             ]
         },
--- a/examples/vg-specs/paintbrush_color.vg.json
+++ b/examples/vg-specs/paintbrush_color.vg.json
@@ -29,6 +29,10 @@
             },
             "transform": [
                 {
+                    "type": "identifier",
+                    "as": "_vgsid_"
+                },
+                {
                     "type": "filter",
                     "expr": "datum[\"Horsepower\"] !== null && !isNaN(datum[\"Horsepower\"]) && datum[\"Miles_per_Gallon\"] !== null && !isNaN(datum[\"Miles_per_Gallon\"])"
                 }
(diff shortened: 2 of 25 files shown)


arvind:
This is a higher priority than I realized -- the default for single/multi selections currently is to use the `_id` field. Besides the volatility problem (#2221), because view composition can often yield to multiple derived datasets, the `_id` fields do not correspond across views. If the `identifier` transform is introduced as described above, the `_vgsid_` field _will_ indeed correspond across views.

arvind:
I'm taking another crack at this!

arvind:
Once I realized that (3) above was a mistake in my testing, rather than an error in the code, I was most of the way there to the remaining points. I addressed (2) by adding a [`model.parent` check](https://github.com/vega/vega-lite/pull/2656/commits/d577d1d6047ad41335904bece3cadf2c1966f71f#diff-6ff9ac68ad5af5ba27bd1c51a6d58548R112) for the source identifier node. 

With regards to (1), I decided to leave the behavior as is to maintain our contract of never polluting user supplied data (especially as the `_id` field will [no longer be enumerable](https://vega-js.slackarchive.io/vega-dev/page-3)). 

So this PR should now be ready for review. Apologies for the delay for what ended up being pretty straightforward!

PR #2665, opened 2017-07-13
Consume wheel events to prevent accidental scrolling.

arvind:
This PR ensures that the page does not accidentally scroll when the user means to zoom the brush or scales.


domoritz:
Thanks!

PR #2667, opened 2017-07-15
[Refactor] Remove all occurences for row/column in scale compiler

kanitw:


kanitw:
(We no longer have scale for row/column)

PR #2672, opened 2017-07-17
Selection Documentation

arvind:
Apologies for including unrelated commits. It was less onerous for me to fix bugs as I found them over the course of writing documentation, rather than switching to separate branches each time. They're small/targeted bug fixes in atomic commits, to simplify reviewing. I'd prefer to not squash them to make tracking easier.


site/static/main.ts
```diff
@@ -29,8 +29,10 @@ function renderExample($target: Selection<any, any, any, any>, text: string) {
 
   const vis = $target.append('div').attr('class', 'example-vis');
 
+  // Decrease visual noise by removing $schema and description from code examples.
+  const textClean = text.replace(/(\s)+\"(\$schema|description)\": \".*?\",/g, '');
```
arvind: Want to call out this change I made. I found the `$schema` and `description` properties to add unnecessary noise as the code blocks are fairly narrow. Can revert if we'd rather keep them.

kanitw: I assume the `$schema` is still included when users click "Open in Vega Editor"?  If so, I think this is a good idea :)

arvind: Yup, that's right. `$schema` and `description` properties remain in the original example specifications. My intention is to just remove them when displaying them on documentation pages to reduce the clutter.

arvind:
As per Slack conversation with @kanitw, we decided to go ahead and merge this PR to get docs published asap, and create follow-up issues as needed.

Issue #2282, opened 2017-04-26
Axis title can be "jumpy" during pan/zoom interactions


jheer:
To reproduce, pan and zoom in this example:
https://vega.github.io/new-editor/?mode=vega-lite&spec=panzoom_scatter

To stabilize axes, we could adjust the default configuration:
- Set the y-axis minExtent to something reasonable (40px?) that accommodates a wider array of labels without the need to reposition the title.
- Include a meaningful maxExtent so that the title does not jump out of the display. Possibly even lock this to the same value as minExtent, so that the title will not ever move.

kanitw:
@jheer   Is this still the case?  I try to reproduce but I don't see a problem or maybe I overlook something.

kanitw:
Fixed in #2673

Issue #1193, opened 2016-02-28
Facet with Independent Scale


kanitw:
- [ ] correct `assembleSignals()` in `facet.ts`

kanitw:
@domoritz test this

domoritz:
It's broken

```json
{
  "facet": {
    "column": {
      "field": "numParts",
      "type": "ordinal"
    },
    "row": {
      "field": "regex",
      "type": "nominal"
    }
  },
  "spec": {
    "data": {
      "values": [
        {
          "regex": 0,
          "globalTime": 256,
          "numParts": 256
        },
        {
          "regex": 0,
          "globalTime": 512,
          "numParts": 512
        },
        {
          "regex": 0,
          "globalTime": 64,
          "numParts": 64
        },
        {
          "regex": 0,
          "globalTime": 1024,
          "numParts": 1024
        },
        {
          "regex": 0,
          "globalTime": 128,
          "numParts": 128
        }
      ]
    },
    "mark": "bar",
    "config": {
      "cell": {
        "width": 600,
        "height": 350
      }
    },
    "encoding": {
      "y": {
        "field": "globalTime",
        "type": "quantitative"
      },
      "x": {
        "field": "numParts",
        "type": "nominal"
      }
    }
  },
  "resolve": {
    "x": {
      "scale": "independent"
    }
  }
}
```


Error: `Uncaught Error: Undefined data set name: "scale_child_main"`

domoritz:
@kanitw What's odd is that the labels on the bottom are misaligned. Is there a way to fix this in the layout?

kanitw:
I think there is a way to set the width of the axis to be larger

kanitw:
I think you can use `minExtent` and `maxExtent` 

See  https://vega.github.io/vega/docs/axes/

Not sure if we really have to modify any rule -- maybe just setting the value for this spec is fine?

kanitw:
Or maybe we can set a fixed `min/maxExtent` https://github.com/vega/vega-lite/issues/2282

rsiebeling:
For what it's worth, I found myself wanting this today: specifically, a version with independent scales. This would be nice for comparing trends in similar but distinct data sets that have very different magnitudes without normalizing them (and losing information). 

Would this be solved?

kanitw:
@rsiebeling Yep, very soon

rsiebeling:
Great, can't wait till the final release. Keep up the good work ;-)

[...]

PR #2673, opened 2017-07-17
Rules for min/maxExtent

kanitw:
Fix https://github.com/vega/vega-lite/issues/2282

Please merge https://github.com/vega/vega-lite/pull/2674 first!

### DIFF
--- a/src/compile/axis/rules.ts
+++ b/src/compile/axis/rules.ts
@@ -53,11 +53,7 @@ export function labelOverlap(fieldDef: FieldDef<string>, specifiedAxis: Axis, ch
   return undefined;
 }
 
-export function minMaxExtent(specifiedExtent: number, isGridAxis: boolean, scaleType: ScaleType) {
-  if (specifiedExtent) {
-    return specifiedExtent;
-  }
-
+export function minMaxExtent(isGridAxis: boolean, scaleType: ScaleType) {
   // For quantitative scale, set extent to 25 by default to avoid jumpy axis title
   // (Fix https://github.com/vega/vega-lite/issues/2282)
   if (getScaleCategory(scaleType) === 'numeric') {
--- a/src/compile/axis/rules.ts
+++ b/src/compile/axis/rules.ts
@@ -1,4 +1,4 @@
-import {Axis} from '../../axis';
+import {Axis, AxisConfig} from '../../axis';
 import {binToString} from '../../bin';
 import {Channel, SpatialScaleChannel, X, Y} from '../../channel';
 import {Config} from '../../config';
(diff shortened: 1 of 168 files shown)


kanitw:
Don't review this yet -- I think I wanna fix https://github.com/vega/vega-lite/issues/1193 together in the same PR (but gotta board my connection to Seattle first!)

kanitw:
Actually NVM about #1193  -- I'm not sure what would be the best default yet -- so let's get this in first.

src/compile/axis/rules.ts
```diff
@@ -53,6 +53,24 @@ export function labelOverlap(fieldDef: FieldDef<string>, specifiedAxis: Axis, ch
...
+
+    // For quantitative scale, set extent to 25 by default to avoid jumpy axis title
+    // (Fix https://github.com/vega/vega-lite/issues/2282)
+    if (getScaleCategory(scaleType) === 'numeric') {
+      return 25;
+    }
+  }
+
+  return undefined;
+}
```
arvind: Should `minExtent` and `maxExtent` be added to `AXIS_PROPERTY_TYPE` in `axis.ts`? Would that simplify this function by dropping the `if` statement?

src/compile/axis/rules.ts
```diff
@@ -53,6 +53,24 @@ export function labelOverlap(fieldDef: FieldDef<string>, specifiedAxis: Axis, ch
...
+    // Only apply extent to non-grid axes as grid axes do not have labels and title.
+    // Otherwise, we would add unnecessary extent to faceted plots.
+    if (specifiedExtent) {
+      return specifiedExtent;
+    }
+
+    // For quantitative scale, set extent to 25 by default to avoid jumpy axis title
+    // (Fix https://github.com/vega/vega-lite/issues/2282)
+    if (getScaleCategory(scaleType) === 'numeric') {
+      return 25;
```
arvind: Why 25px? In my tests with zooming quantitative scales (e.g., the `panzoom_scatter` or the Gapminder `bubble_health_income` examples), with 25px extents, axis labels with one decimal place were flush against the title and with two decimal places overlapped it. 

30px seems to give us more breathing room (1dp still maintains padding while 2dps are flush).

arvind: Also, rather than hardcoding this, should we move it to a config value and/or something that can be customized by the user in their specs? E.g., `encoding.x.axis.extent = [min, max]`?

src/compile/axis/rules.ts
kanitw: > Should minExtent and maxExtent be added to AXIS_PROPERTY_TYPE in axis.ts?

Yes.  Great catch.  I actually added `AXIS_PROPERTY_TYPE` because I think I need it for `minExtent` and `maxExtent` but I stash and swap commit order and forgot to bring it back :p

src/compile/axis/rules.ts
kanitw: > Why 25px? In my tests with zooming quantitative scales (e.g., the panzoom_scatter or the Gapminder bubble_health_income examples), with 25px extents, axis labels with one decimal place were flush against the title and with two decimal places overlapped it.
> 30px seems to give us more breathing room (1dp still maintains padding while 2dps are flush).

Jeff actually originally suggests 40px but I felt like that's a bit too big so I decrease it (maybe I overdo it.)  I could change to `30` but wonder if you like 40 better also.  

> Also, rather than hardcoding this, should we move it to a config value and/or 

I'll add `config.axis.quantitativeExtent`

> something that can be customized by the user in their specs? E.g., encoding.x.axis.extent = [min, max]?

Users can already do `encoding.x.axis.max/minExtent`.  (This function only determines default rules when not specified.)

arvind: Yeah, I saw the 40px suggestion but agree that it seemed like too much. 30px seems like a good in-between though, enough breathing room for up to 2 decimal places (which, in our examples at least, amounts to a fair amount of zooming).

PR #2675, opened 2017-07-18
Move null filter to end 

### LINKED ISSUE #2504: Move NullFilter node to near the end (before filter NonPositiveFilter)?
I think this will make optimization way easier

AkshatSh:
Created a new data node for filtering invalid based on field. Replaces occurences of null filter and non positive filter and tested on gallery and in vega editor.

Fix #2504


src/compile/data/FilterInvalid.ts
```diff
@@ -0,0 +1,112 @@
...
+import {UnitModel} from './../unit';
+import {DataFlowNode} from './dataflow';
+import {isScaleChannel} from '../../channel'
+import {hasContinuousDomain} from '../../scale';
+import {QUANTITATIVE, TEMPORAL} from '../../type';;
+import {FieldDef} from '../../fielddef';
+import {Model, ModelWithField} from '../model';
+
+export class FilterInvalidNode extends DataFlowNode {
+  private _filter_null: Dict<FieldDef<String>>;
```
kanitw: 1)  We don't use underscore scale naming.   
2) Don't add unnecessary `_` prefix for private if not necessary.
3) I think you can have one dict that stores {fieldDef, scaleType} and apply correct filter based on the scaleType.

src/compile/data/FilterInvalid.ts
```diff
@@ -0,0 +1,112 @@
...
+            }
+            return nonPositiveComponent;
+        }
+        return nonPositiveComponent;
+      }, {} as Dict<FieldDef<String>>);
+
+    // create the filter of all the fields that cannot have null
+    let filter_nulls = {};
+    if (model instanceof ModelWithField) {
+      filter_nulls = model.reduceFieldDef((aggregator: Dict<FieldDef<string>>, fieldDef, channel) => {
```
kanitw: As we can only have one dict, you can just reduce all of them together 

`model.reduceFieldDef((aggregator: Dict<FieldDef<string>>, fieldDef, channel)`
and check if `isScaleChannel(channel)` should be pretty optimal.

kanitw:
I think we should not have to modify assemble at all.   Just literally moving it during the parse phase could be sufficient. 

Thus, I think you can drop https://github.com/vega/vega-lite/pull/2675/commits/5ab4609df053fd9360d21215cdd4d68e20b27a0a

src/compile/data/FilterInvalid.ts
```diff
@@ -0,0 +1,78 @@
...
+import {hasContinuousDomain, scaleCompatible, ScaleType} from '../../scale';
+import {QUANTITATIVE, TEMPORAL} from '../../type';
+import {contains, Dict, differ, differArray, duplicate, extend, hash, keys, stringValue} from '../../util';
+import {VgFilterTransform, VgTransform} from '../../vega.schema';
+import {Model, ModelWithField} from '../model';
+import {UnitModel} from './../unit';
+import {DataFlowNode} from './dataflow';
+
+export class FilterInvalidNode extends DataFlowNode {
+  private filterInvalid: Dict<ScaleType>;
```
kanitw: As we can only have one dict, you can just reduce all of them together

model.reduceFieldDef((aggregator: Dict<FieldDef<string>>, fieldDef, channel)
and check if isScaleChannel(channel) should be pretty optimal.

src/compile/data/parse.ts
```diff
@@ -156,6 +148,12 @@ export function parseData(model: Model): DataComponent {
       tu.parent = head;
       head = tu;
     }
+
+    const filter = FilterInvalidNode.make(model);
```
kanitw: Why is filter invalid now up here?  I thought we want to move it to the end?

src/compile/data/filterinvalid.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  // create the VgTransforms for each of the filtered fields
+  public assemble(): VgTransform[] {
+
+     return keys(this.filter).reduce((vegaFilters, field) => {
+      const fieldDef = this.fieldDefs[field];
+      const scaleCompType = this.filter[field];
+      const filters = [];
+
```
kanitw: I would declare 
```
const ref = fieldRef(fieldDef, {expr: 'datum'});
```

and 
`import {field as fieldRef}` from `src/fielddef.ts`

src/compile/data/parse.ts
```diff
@@ -248,6 +240,12 @@ export function parseData(model: Model): DataComponent {
       tu.parent = head;
       head = tu;
     }
+
+    // const filter = FilterInvalidNode.make(model);
```
kanitw: Remove commented code?

src/compile/data/filterinvalid.ts
```diff
@@ -0,0 +1,68 @@
...
+  constructor(fieldDefs: Dict<FieldDef<string>>) {
+   super();
+
+   this.fieldDefs = fieldDefs;
+  }
+
+  public static make(model: ModelWithField): FilterInvalidNode {
+
+    const filter = model.reduceFieldDef((aggregator: Dict<FieldDef<string>>, fieldDef, channel) => {
```
kanitw: Comparing to the old `nullfilter.ts`, somehow we no longer check `model.config.invalidValues === 'filter'` here.  
Let's add it back; otherwise we can't disable invalid filter.

src/compile/data/filterinvalid.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  get filter() {
+    return this.fieldDefs;
+  }
+
+  // create the VgTransforms for each of the filtered fields
+  public assemble(): VgTransform[] {
+
+     return keys(this.filter).reduce((vegaFilters, field) => {
```
kanitw: The old assemble in nullfilter.ts produces one filter per `FilterInvalidNode` (`NullFilterNode`), which is more efficient as it requires only one pass over the whole dataset.   We should do the same here. 

For example, we shouldn't see the diff in examples/compiled/trellis_row_column.vg.json.

examples/compiled/rect_binned_heatmap.vg.json
```diff
@@ -16,10 +16,6 @@
             },
             "transform": [
                 {
-                    "type": "filter",
```
kanitw: It looks like moving the filter affects the bin for this example.   Not sure why.  Maybe the bin operator does not handle invalid filter well?

examples/compiled/repeat_histogram.vg.json
```diff
@@ -20,10 +20,6 @@
                     "as": "Horsepower"
                 },
                 {
-                    "type": "filter",
```
kanitw: Looking at this one confirms that bin does not handle null filter well, treating them as zero.  (Look at the bin for horsepower.)

kanitw:
I'm glad that we have the SVG diff, which highlights my oversight.

Moving filter invalid to the end still provides benefits for the most cases though.  

Here is what I would do.

- [ ] Modify BinNode's assemble to prepend a filter transform to filter invalid values.  I think you may want to create a common function for producing a filter transform to exclude invalid values to use in both bin and filterinvalid.ts.  Be careful that for bin, we need to filter the pre-transformed data though. (Use `field()`  with `opt.nofn = true`.) 

- [ ] Since BinNode already filters data for itself, filterInvalid no longer needs to filter binned data.

kanitw:
I'm editing this part above:

- [ ]  Extract a common function for producing a filter predicate formula to exclude invalid values to use in both bin and filterinvalid.ts.  (We will re-use this function in other cases in the future too.)

src/compile/data/parse.ts
```diff
@@ -275,11 +273,13 @@ export function parseData(model: Model): DataComponent {
       stack.parent = head;
       head = stack;
     }
+  }
 
-    const nonPosFilter = NonPositiveFilterNode.make(model);
-    if (nonPosFilter) {
-      nonPosFilter.parent = head;
-      head = nonPosFilter;
+  if (isUnitModel(model) || isFacetModel(model)) {
```
kanitw: row / column channel in facet never have scales, so we don't really need to include `|| isFacetModel(model)` here.

examples/compiled/repeat_histogram.vg.json
kanitw: Actually it may be that the bin handles it okay, but linear scale isn't very smart about it too.

kanitw:
@AkshatSh @domoritz  This PR looks good for me except that I don't understand why the binned heatmap example has value changed when we move the filter.   Could two of you help investigate? 

(I looked at it already but I don't understand why)

domoritz:
@AkshatSh have you had a chance to look at the issue? I will look at it tonight to see what's going on.

domoritz:
There is a subtle change in the PR.

screen shot 2017-10-16 at 20 45 17

Above is the chart with filtering before everything and below the chart with filtering after everything. 

This is happening because we are filtering tuples that have a null for any field. 

Take these tuples: (10,null),(20,20),(30,30).
Extent of the first field is 10 - 30 and extent of the second field is 20 - 30.
But if we filter first, we remove (10,null). So then the extent is 20 - 30 for both.

PR #2676, opened 2017-07-18
Replace constant values in scale with signals

kanitw:


kanitw:
Since this PR changes pretty much every file, let me check the gallery too. 

(I think another pair of eyes at the test gallery would be helpful as well.)

kanitw:
Ok I looked at the gallery and all examples look reasonable to me.   (Except `lookup`, which is also broken on master.) 

Please double check the gallery though -- there are a lot of plots and I might have overlooked some of them.

Issue #2668, opened 2017-07-17
Merge the top-level width and height signal with the native width / height signals


kanitw:
Forked from https://github.com/vega/vega-lite/issues/2198

This is only applicable when the top-level spec is named.

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/2682

PR #2682, opened 2017-07-19
Merge the top-level spec's width/height signal with the top-level model

kanitw:
Fix https://github.com/vega/vega-lite/issues/2668

Please merge https://github.com/vega/vega-lite/pull/2676 first!


kanitw:
Yeah it's not really important but it seems easy and annoys me enough so I decide to just fix it last night :p

PR #2685, opened 2017-07-19
[WIP] Correct Facet independent scales with outer layout signals

kanitw:


kanitw:
https://github.com/vega/vega-lite/pull/2697 is a better approach

PR #2688, opened 2017-07-20
Always clip marks.

arvind:
This may be a little controversial but this PR proposes to always clip marks. This handles the case where marks would overflow if the user explicitly sets scale domains with a smaller extent than [min, max] (example spec below). Moreover, if we always clip marks, we also simplify selection code paths, which no longer need to test whether clipping is required + propagate it up the model hierarchy (e.g., a source of a bug with [panning concatenated displays](https://vega.github.io/vega-lite/docs/selection-bind.html#scale-binding)).

This change touches every example, but looking through the gallery I don't see anything adversely affected (since the default _is_ a [min, max] extent).

```json
{
  "mark": "circle",
  "encoding": {
    "x": {
      "field": "Horsepower",
      "type": "quantitative",
      "scale": {"domain": [75,150]}
    },
    "y": {
      "field": "Miles_per_Gallon",
      "type": "quantitative",
      "scale": {"domain": [20,40]}
    },
    "size": {"field": "Cylinders","type": "quantitative"},
    "color": {"field": "Origin","type": "nominal"}
  }
}
```

Before:
download 6

After:
download 7

### DIFF
--- a/examples/vg-specs/area.vg.json
+++ b/examples/vg-specs/area.vg.json
@@ -72,6 +72,7 @@
         {
             "name": "marks",
             "type": "area",
+            "clip": true,
             "from": {
                 "data": "source_0"
             },
--- a/examples/vg-specs/area_vertical.vg.json
+++ b/examples/vg-specs/area_vertical.vg.json
@@ -73,6 +73,7 @@
         {
             "name": "marks",
             "type": "area",
+            "clip": true,
             "from": {
                 "data": "source_0"
             },
--- a/examples/vg-specs/bar.vg.json
+++ b/examples/vg-specs/bar.vg.json
@@ -89,6 +89,7 @@
         {
             "name": "marks",
             "type": "rect",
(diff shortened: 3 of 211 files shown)


kanitw:
This change leads to many poorly clipped examples:

vega-lite_test_gallery

vega-lite_test_gallery

vega-lite_test_gallery


> if the user explicitly sets scale domains with a smaller extent than [min, max] (example spec below)

Not sure if clipping is *always* the right strategy here.  If users apply clamping, they they probably want to see full point marks on the border rather than half of them (as the other half gets clipped).

arvind:
Good catch, I hadn't noticed those. Perhaps a better option is to expose `clip` as an option in the mark definition object? That would correspond to where it exists in Vega (rather than the complexities of trying to put in in mark config, see vega/vega-parser#47).

kanitw:
Yeah i think its a reasonable property of mark def :)

PR #2692, opened 2017-07-20
Improve mark clipping

arvind:
This PR makes two changes:
1. It exposes `clip` as a MarkDef property to account for specs as in #2688.
2. Simplifies selection-based clipping determination to occur entirely at parse time, rather than during selection assembly. This approach fixes a couple of bugs where marks were either being erroneously clipped when layered, or were not being clipped when scales were shared across concatenated views.


src/compile/mark/mark.ts
```diff
@@ -60,12 +60,12 @@ function parsePathMark(model: UnitModel) {
   const details = detailFields(model);
 
   const role = model.markDef.role || markCompiler[mark].defaultRole;
-
+  const clip = model.markDef.clip != null ? model.markDef.clip : scaleClip(model);
```
kanitw: `model.markDef.clip !== undefined`?

site/docs/mark.md
```diff
@@ -41,7 +41,7 @@ For example, a bar chart has `mark` as a simple string `"bar"`.
 
 A mark definition object can contains the following values.
 
-{% include table.html props="type,role,filled,orient,interpolate,tension" source="MarkDef" %}
+{% include table.html props="type,role,filled,clip,orient,interpolate,tension" source="MarkDef" %}
```
kanitw: wonder if we should order every after `type` and `role` (or just `type`) alphabetically.

src/compile/mark/mark.ts
```diff
@@ -149,8 +149,8 @@ function detailFields(model: UnitModel): string[] {
 }
 
 function scaleClip(model: UnitModel) {
-  const xScaleDomain = model.scaleDomain(X);
-  const yScaleDomain = model.scaleDomain(Y);
-  return (xScaleDomain && isSelectionDomain(xScaleDomain)) ||
-    (yScaleDomain && isSelectionDomain(yScaleDomain)) ? true : false;
+  const xScale = model.getScaleComponent('x');
+  const yScale = model.getScaleComponent('y');
+  return (xScale && xScale.get('domainRaw')) ||
```
kanitw: Maybe add comment that we apply clipping if the interval selections is bound to scale domains,
which in turn produces `domainRaw` ? 

(It took me a while to match the commit message to this part of code.)

src/compile/selection/selection.ts
```diff
@@ -189,38 +189,27 @@ export function assembleUnitSelectionData(model: UnitModel, data: VgData[]): VgD
...
     });
   });
 
-  // In a layered spec, we want to clip all layers together rather than
-  // only the layer within which the selection is defined. Propagate
-  // our assembled state up and let the LayerModel make the right call.
-  if (model.parent && model.parent instanceof LayerModel) {
-    return [selMarks, clipMarks];
-  } else {
-    return clipGroup ? clipMarks(selMarks) : selMarks;
```
kanitw: Remove the now unused `clipMarks()` ?

src/compile/selection/interval.ts
```diff
@@ -135,6 +135,7 @@ const interval:SelectionCompiler = {
 
     return [{
       type: 'rect',
+      clip: true,
```
kanitw: Not really the main point of this PR, but it may be worth giving a name to this rect mark too?   (That will help skimming at the diff spec a lot.)

src/compile/mark/mark.ts
arvind: Since `model.markDef.clip` should be boolean, `null` is not technically a valid value. If the user specifies `null`, should we interpret that to be `false`?

site/docs/mark.md
arvind: I thought about ordering alphabetically. Chose to keep this ordering since `type` and `role` are closely related...

src/compile/selection/selection.ts
arvind: Good catch.

site/docs/mark.md
kanitw: Yeah, but `clip` should still come before `filled`?

src/compile/mark/mark.ts
kanitw: I only mention `model.markDef.clip !== undefined` because I thought you intend to use `scaleClip()` function only if clip is not specified (`undefined`).   (If users directly specify `clip`, then we should follow their instruction.)  

For `null` case, I think it's fair to interpret as `false`.  I also believe that passing `null` to Vega would be equivalent to false since underlying Vega probably would clip when `clip` is truthy?   If so, we might not need any special casing for `null`.

arvind:
@kanitw: Changes made!

PR #2694, opened 2017-07-20
Do not add empty/invalid selections to the store.

arvind:


kanitw:
Can you explain a bit more what was the problem that previous behavior cause?

"Do not add empty/invalid selections to the store." sounds reasonable, but I want to make sure I understand the problem.  :)

arvind:
Sorry, yes. Two issues are fixed here:
* If the user clicked in whitespace to clear the brush, previously, this would still add an interval to the backing store. Our predicate functions have a check to verify that if the extent coordinates match, the selection should be treated as empty. However, this fails when brushing over an ordinal/nominal domain because inverting a single point generates a single data value. So, it's easier (and more correct) to simply remove the interval entirely. 
* The signal definition for discrete (single/multi) selections would only add something to the store if a valid `datum` was present on interaction. However, with faceted plots, this check isn't sufficient as the group marks are also data-driven. Now we ensure that a `datum` is present and an actual mark has been interacted with in order to populate the selection.

arvind:
@kanitw: Good call. Done!

PR #2697, opened 2017-07-21
Correct incorrect width/height in Facet's encode block

kanitw:
Part of #2555 

- [x] facet of unit
  - [x] signals for the main subgroup
  - [x] signals for the header subgroups
    - [x] either `row` or `column`
    - [x] both `row` and `column`


kanitw:
Feel free to review code first, i'll add test later tomorrow.    (Too tired to do more)

domoritz:
Why are we not reusing the `column` and `row` output node created in `parse.ts`?

kanitw:
> Why are we not reusing the `column` and `row` output node created in `parse.ts`?

Which parse.ts?  If you talk about `data/parse.ts`, such code do not exist because these two output data sources are a part of the FacetNode so they do not have their own nodes. 

In any case, even if you think we should output separate nodes, the re-design of the part is beyond the scope of this PR.

domoritz:
Yes, we don't have the output nodes yet but I wonder whether they differ. If not too much, at least they should use a common code path, which is in the scope of this PR.

kanitw:
> Yes, we don't have the output nodes yet but I wonder whether they differ

Just in case you forgot, we previously discussed and agreed in person that we can have FacetNode output multiple relevant data sources instead of complicating the dataflow "tree" with more unnecessary nodes that are all a part of facet.  That's why the code was like this back then. 
So what I implemented is basically following this agreement.  

If you want to change this, then convince me why we should change and submit another PR?

kanitw:
I think the rationale back then was that they always get moved together, so keeping them in the same node is way easier to optimize.  

That said, thinking more -- we probably need to revise this soon due to an obscure nature of `layout` as discussed [here](https://github.com/vega/vega/issues/952#issuecomment-319720058).

domoritz:
@kanitw, I totally agree that this should be contained within a single node. I'm just asking whether we can have the code that generates the aggregate to be called from both places, the facet code and the code you wrote here.

kanitw:
Oh I see.  I think what I did here is correctly placed though.  
 
Basiclly, the two `aggregate` I added are different kind of `aggregate`.  

1) The `aggregate` in `compile/data/facet` (FacetNode) is an aggregate transform, which makes sense to be in FacetNode.

2) The `aggregate` in `compile/facet` is an aggregate operator inside a `facet` operator of a group mark (not a part of data source) -- thus it belongs in `compile/facet` (which is exactly where you did add `aggregate: {cross: true}` earlier -- I just extent the same object there in this PR).

kanitw:
FYI, since rebase would be a pain here, I'll use merge it manually.

kanitw:
(I need to merge it since I need this to rebase other things properly, otherwise it's getting very messy.)

PR #2698, opened 2017-07-21
Correct independent scale scope for concat/repeat

kanitw:
Do not nest them so that signals can still refer to these scales.

Part of #2555

### DIFF
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -241,9 +241,7 @@ export abstract class Model {
 
   public abstract assembleLayoutSignals(): VgSignal[];
 
-  public assembleScales(): VgScale[] {
-    return assembleScalesForModel(this);
-  }
+  public abstract assembleScales(): VgScale[];
 
   public assembleHeaderMarks(): VgMarkGroup[] {
     const {layoutHeaders} = this.component;
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -25,7 +25,7 @@ import {assembleLegends} from './legend/assemble';
 import {LegendComponentIndex} from './legend/component';
 import {parseMarkDef} from './mark/mark';
 import {RepeaterValue} from './repeat';
-import {assembleScale} from './scale/assemble';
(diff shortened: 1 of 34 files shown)


kanitw:
This now works for [this concat spec](https://github.com/vega/vega-lite/issues/2555#issue-238020243). 

vega_editor

kanitw:
Ready for review! --- Looking at each commit one-by-one would be easier :)

src/compile/model.ts
```diff
@@ -241,7 +241,7 @@ export abstract class Model {
   public abstract assembleLayoutSignals(): VgSignal[];
 
   public assembleScales(): VgScale[] {
-    return assembleScale(this);
+    return assembleScalesForModel(this);
   }
```
arvind: As far as I can tell, `Model` and every `Model` subclass has an identical `assembleScales` function? My vote would be to remove them from the subclasses to reduce the number of codepaths.

src/compile/model.ts
```diff
@@ -280,10 +280,10 @@ export abstract class Model {
     return assembleLegends(this);
   }
 
-  public assembleGroup(signals: VgSignal[] = []) {
+  public assembleGroup(mixins: {signals?: VgSignal[], scales: VgScale[]}) {
```
arvind: Perhaps worth adding a comment about why scales and signals are passed in (rather than assembled within this method as before)? Took me a second to understand why the change in logic was necessary (to allow for scale references from signals, which the previous approach did not).

src/compile/model.ts
kanitw: They are actually different.  I'm making it `abstract` then so it's clearer.

src/compile/model.ts
kanitw: I just found a way to make this part less convoluted! 
https://github.com/vega/vega-lite/pull/2698/commits/fbfc544565b0a248704bac95970d4734d7ada971

kanitw:
Just pushed a few fixes -- let me try to improve test coverage too!

PR #2705, opened 2017-07-24
Rename `assemble` to `assembleTopLevelModel` and add comment to explain

kanitw:


kanitw:
> assembleVega is perhaps more semantically accurate.

Everything in assemble is assembling Vega.  This one in particular assemble the top-level part of it.

kanitw:
So it's either just `assemble` (drop the commit) or just like what I did.

willium:
I'd keep assemble. My biggest issue with this its not clear what affect the various `assemble` steps have on the model when `parse` mutates the model. I'd actually err towards making parse not mutate the model--but that's not relevant to this PR. I think here, the comments are nice, the more explicit name isn't necessary.

kanitw:
>  My biggest issue with this its not clear what affect the various assemble steps have on the model when parse mutates the model.

Honest I don't think this is really a big deal and in fact is pretty straightforward as it "returns" something and it is the last step of the compilation.  

In general, for method that returns something, we avoid causing side-effect. 
Maybe you could add a note to contribute.md to note our coding style.  

(Even if it does, the code is slightly worse but it doesn't make it a critical issue. -- Just saying, but we didn't cause side effect here though.)

src/compile/compile.ts
```diff
@@ -55,7 +58,12 @@ function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config) {
   };
 }
 
-function assemble(model: Model, topLevelProperties: TopLevelProperties) {
+/*
+ * Assemble the top-level model.
+ * (This couldn't be `model.assemble()` since the top-level model needs some special treatment
```
domoritz: Remove the ()

src/compile/compile.ts
```diff
@@ -22,23 +22,26 @@ export function compile(inputSpec: TopLevelExtendedSpec, logger?: log.LoggerInte
...
     const spec = normalize(inputSpec, config);
 
     // 3. Instantiate the models with default config by doing a top-down traversal.
+    // This allows us to pass properties that child models derive from their parents via their constructors.
     const model = buildModel(spec, null, '', undefined, undefined, config);
 
-    // 4. Parse each part of the model to produce components that can be merged and assembled easily.
-    // We do a bottom-up traversal over the whole tree to
+    // 4. Parse parts of each model to produce components that can be merged
+    // and assembled easily as a part of model.
```
domoritz: part of a model

src/compile/compile.ts
```diff
@@ -22,23 +22,26 @@ export function compile(inputSpec: TopLevelExtendedSpec, logger?: log.LoggerInte
...
+    // In this phase, we do a bottom-up traversal over the whole tree to
```
domoritz: of the whole tree

src/compile/compile.ts
```diff
@@ -22,23 +22,26 @@ export function compile(inputSpec: TopLevelExtendedSpec, logger?: log.LoggerInte
...
     // parse once for each type of components (e.g., data, layout, mark, scale).
```
domoritz: to parse each component type once

src/compile/compile.ts
```diff
@@ -22,23 +22,26 @@ export function compile(inputSpec: TopLevelExtendedSpec, logger?: log.LoggerInte
...
     // By doing bottom-up, we starting parsing components of unit specs and
```
domoritz: this and the following lines don't make sense to me

PR #2710, opened 2017-07-26
Dom/overlaps

domoritz:
This makes especially binned plots with many bins much nicer. 

* always prevent overlapping
* remove hack that set the tick count to 5

screen shot 2017-07-25 at 17 37 10
screen shot 2017-07-25 at 17 35 09

### DIFF
--- a/src/compile/axis/parse.ts
+++ b/src/compile/axis/parse.ts
@@ -297,7 +297,7 @@ function getProperty<K extends keyof (Axis|VgAxis)>(property: K, specifiedAxis: 
     case 'orient':
       return getSpecifiedOrDefaultValue(specifiedAxis.orient, rules.orient(channel));
     case 'tickCount':
-      return getSpecifiedOrDefaultValue(specifiedAxis.tickCount, rules.tickCount());
+      return specifiedAxis.tickCount;
     case 'ticks':
       return rules.ticks(property, specifiedAxis, isGridAxis, channel);
     case 'title':
--- a/src/compile/axis/parse.ts
+++ b/src/compile/axis/parse.ts
@@ -297,7 +297,7 @@ function getProperty<K extends keyof (Axis|VgAxis)>(property: K, specifiedAxis: 
     case 'orient':
       return getSpecifiedOrDefaultValue(specifiedAxis.orient, rules.orient(channel));
     case 'tickCount':
-      return getSpecifiedOrDefaultValue(specifiedAxis.tickCount, rules.tickCount(channel, fieldDef)); // TODO: scaleType
+      return getSpecifiedOrDefaultValue(specifiedAxis.tickCount, rules.tickCount());
     case 'ticks':
       return rules.ticks(property, specifiedAxis, isGridAxis, channel);
     case 'title':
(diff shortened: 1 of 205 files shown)


kanitw:
From the code perspective, I think this is simpler.

However, I don't like when there are more ticks than labels like this:

screen shot 2017-07-25 at 17 35 09

The old output seems more visually pleasing

vega_editor


The Heatmap example is nice though, we should get it in another PR.

domoritz:
@kanitw That's just what happens in this one particular example. If you change the size, you won't have the overlap.

kanitw:
The heatmap example also has so many ticks!    I agree that `tickCount` = 5 is not a good idea for larger plots though (say 800 px)

domoritz:
I disagree that the binned plot has too many ticks because it's useful to know where the bins start and end. If you think there are too many ticks, then the only option is to only show ticks for where we have labels, which would require some changes in Vega (and I don't think it's a good idea).

kanitw:
> I disagree that the binned plot has too many ticks because it's useful to know where the bins start and end. 

Ah for bin, this makes sense.

test/compile/axis/rules.test.ts
```diff
@@ -47,15 +47,10 @@ describe('compile/axis', ()=> {
   });
 
   describe('tickCount', function() {
-    it('should return undefined by default for non-x', function () {
-      const tickCount = rules.tickCount('y', {field: 'a', type: 'quantitative'});
+    it('should return undefined by default', function () {
```
kanitw: Remove this

src/compile/axis/rules.ts
```diff
@@ -74,13 +76,7 @@ export function orient(channel: Channel) {
...
 }
 
-export function tickCount(channel: Channel, fieldDef: FieldDef<string>) {
-  // FIXME depends on scale type too
-  if (channel === X && !fieldDef.bin) {
-    // Vega's default tickCount often lead to a lot of label occlusion on X without 90 degree rotation
-    return 5;
-  }
-
+export function tickCount(): number {
```
kanitw: If there is no rule, better totally remove the function.

src/compile/axis/parse.ts
```diff
@@ -297,7 +297,7 @@ function getProperty<K extends keyof (Axis|VgAxis)>(property: K, specifiedAxis:
     case 'orient':
       return getSpecifiedOrDefaultValue(specifiedAxis.orient, rules.orient(channel));
     case 'tickCount':
-      return getSpecifiedOrDefaultValue(specifiedAxis.tickCount, rules.tickCount(channel, fieldDef)); // TODO: scaleType
+      return getSpecifiedOrDefaultValue(specifiedAxis.tickCount, rules.tickCount());
```
kanitw: Just remove this case as it only returns the specified property at the end

kanitw:
The question is whether we totally remove tickCount for non-bin.  Probably yes, and maybe we can make the grid lighter so it's less visually cluttered.

kanitw:
Also why are there two commits with the same name -- forgot to squash?

test/compile/axis/rules.test.ts
domoritz: Done

src/compile/axis/rules.ts
domoritz: Done

src/compile/axis/parse.ts
domoritz: Done

domoritz:
Rebased

Issue #2711, opened 2017-07-27
Limit selection resolution schemes in 2.0


arvind:
Our research paper introduced 6 selection resolution schemes: `global`, `independent`, `union`, `intersect`, `union_others`, `intersect_others`. I'm proposing that, for 2.0, we pare them back to just the `global`, `union`, and `intersect`. We can always re-introduce them in future versions if there's sufficient user demand.

Some reasoning:
* `global`, `union`, `intersect` are likely the most common use cases for resolving instances of the same selection. Unlayered cross filtering provided a motivating example for `intersect_others` and `independent` and `union_others` were included for completeness. 
  * Compared to the existing layered cross filtering example, unlayered cross filtering provides a less compelling interaction technique as it is difficult to hold the binning scheme constant within a repeated display. As a result, the number of bars in the histograms change as brushing occurs: 
![unlayeredcrossfilter](https://user-images.githubusercontent.com/42262/28682974-bc2b89fc-72b3-11e7-83b9-2d5222ae4af4.gif)
* `independent`, `union_others` and `intersect_others` requires identifying the unit view a selection is used in (i.e., passing the unit name to the selection's predicate expression function). As a result, to filter faceted displays with a selection, a nested dataset must be derived as the only way to identify one facet from another is using a corresponding `facet` signal that selections generate. This approach likely represents a non-trivial amount of code churn for a use case I am not yet convinced merits the effort.

Thoughts/objections @kanitw, @domoritz, @jheer, others?

kanitw:
I think lesser options is good.   Having too many options would make it more complicated to learn anyway.  For example, without description you provide above, I already cannot remember instantly what each resolution scheme does (I gotta think a bit about what they mean to understand).  

From a brief conversation today with @jheer,  I think he agrees to focus on more important things and do them well too. :)

kanitw:
@arvind I guess this is fixed in https://github.com/vega/vega-lite/issues/2712?

PR #2712, opened 2017-07-27
Remove unit-specific selection resolutions.

arvind:
This commit removes the `independent`, `union_others`, and `intersect_others` selection resolutions as per #2711. It relies on vega/vega-parser#50 which reorders the parameters of selection predicate expression functions, allowing us to reintroduce these resolutions at a future date.


kanitw:
@arvind  Won't this break since vega-parser#50 isn't included in the latest vega release yet?

arvind:
If we were to cut another Vega-Lite release prior to a Vega release, yes. But we've merged such tracked changes in the past, and I needed it to be able to continue moving forward with runtime tests.

kanitw:
> But we've merged such tracked changes in the past and I needed it to be able to continue moving forward with runtime tests.

I don't know if we  have done it intentionally in the past (more like mistake?).  

I think in general we should avoid it and consider a feature branch approach instead.  That said, let's ignore this for this time so it's easy for you.   :)

PR #2718, opened 2017-07-28
Split encoding.md into encoding, type, and stack

felixcodes:
TODO: Write intros for `type`, `stack`

Please verify how the docs look after split -- I wasn't able to compile and check with `npm run site`


kanitw:
Thanks! I'll take it from here.

PR #2721, opened 2017-07-31
Don’t apply transparent fill or stroke to legends.

### LINKED ISSUE #2719: shapes don't appear in legend (but do on plot)
Altair-generated Vega-Lite JSON, stored in https://owensgroup.github.io/gpustats/plots/Processing%20Power%20over%20Time.html

Using:
```
<script src="https://cdnjs.cloudflare.com/ajax/libs/vega/3.0.0-rc4/vega.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/vega-lite/2.0.0-beta.10/vega-lite.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/vega-embed/3.0.0-beta.19/vega-embed.js">
```

Shapes seem to work OK when there's only a shape, but not when there's shape+color.

domoritz:
Fixes #2719


kanitw:
It seems like this won't be a problem for the next version of Vega anyway?

PR #2724, opened 2017-07-31
Runtime/integration tests for selections

arvind:
This PR adds runtime/integration tests using Chrome via webdriver.io. A Vega-Lite spec is rendered, a variety of interactions are performed, and assertions are run against the interactive state as well as a snapshot of the SVG string.

Tests can be run via `npm run test:runtime` and is set to automatically run with `npm run test` as well as `npm run test:nocompile`. To generate new SVG string snapshots, run `npm run test:runtime:generate` instead. 

The generated SVG strings are current to Vega-Lite master 9f82014 but requires Vega 3.0.0-rc5 tuple ID changes as well several open vega-parser PRs to be merged. Although this PR is complete, I've marked it as WIP until the necessary Vega version is released.


package.json
```diff
@@ -54,10 +54,12 @@
...
 
     "lint": "tslint --project tsconfig.json -c tslint.json --type-check",
-    "test": "npm run tsc && npm run test:only && npm run lint",
+    "test": "npm run build:only && npm run test:only && npm run test:runtime && npm run lint",
     "posttest": "npm run schema && npm run data && npm run mocha:examples",
-    "test:nocompile": "npm run test:only && npm run lint && npm run mocha:examples",
+    "test:nocompile": "npm run test:only && npm run test:runtime && npm run lint && npm run mocha:examples",
     "test:only": "nyc --reporter=html --reporter=text-summary npm run mocha:test",
+    "test:runtime": "TZ=America/Los_Angeles wdio wdio.conf.js",
+    "test:runtime:generate": "rm -Rf test-runtime/resources && VL_GENERATE_TESTS=true npm run test:runtime",
```
kanitw: Quick Q -- When do we use `test:runtime:generate`?

arvind: It must be called manually to generate the SVG snapshots that get stored `test-runtime/resources`. The `test:runtime` script generates SVG snapshots during test cases and compares it to those in the `resources` folder.

kanitw: So I assume this is an equivalent to `build:examples` for the runtime test infrastructure right?

arvind: Exactly!

kanitw: Maybe it's worth updating PULL_REQUEST_TEMPLATE.md and CONTRIBUTING.md accordingly?

I also wonder if we should name this more similar to `build:examples` so the similarity is clearer.  (I don't have a good name to suggest though.)

arvind: The reason I chose this name is because generation of the SVG snapshots is inherently tied to the test cases (i.e., it is not a pure "build" step). One could imagine structuring the code to be the complement (i.e., only run assertions if a `TEST` environment variable is present), but that seems less clean than the current approach. 

I will update the two Markdown guides.

arvind:
This PR should now be ready for merging. It should be up-to-date with respect to Vega 3.0.0-rc7 and Vega-Lite 2.0.0-beta.11.

PR #2725, opened 2017-07-31
[WIP] Make `build-examples` also include static SVG output

kanitw:
Note that this won't pass test yet since we recently merged a code that actually requires a newer version of Vega.


kanitw:
I'll close this as we have #2726 that merges into the `hotfix` branch and we will later merge the branch to master.

PR #2726, opened 2017-07-31
[HotFix] Make `build-examples` also include static SVG output 

kanitw:
This is a mirror of #2725, but I'm merging this to a new `hotfix` branch, which branches from master before we introduce a Vega output that requires the unreleased `vega-rc5`.


domoritz:
I don't think we should have the SVGs in the repo. We should certinaly compile them as part of teh tests but not store them. The reason is that the repo gets larger, we introduce another build step before we can make even simple changes, and we will probably rarely look at the images because github doesn't preview svgs.

kanitw:
As mentioned on slack, it certainly makes repo larger and makes thing slightly slower, but certain way safer to protect us from introducing new bugs.  

FWIW, I already detect a bug (#2731) that we previously did not detect because of this additional infrastructure.  

Thus. I'll merge.

PR #2734, opened 2017-08-01
Geo

willium:
Add Geo to Vega-Lite
- [x] Add examples
- [x] Implement
- [x] Add tests
- [x] Add docs
- [ ] Add interaction support (@arvind)

choropleth
repeat
trellis
geo_layer
geo_point
geo_text
geo_line
image


src/compile/projection/parse.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+function parseUnitProjection(model: UnitModel, config: Config): ProjectionComponent {
+  const {projection, markDef, encoding} = model;
+
+    // TODO: make sure this actually works for both kinds of lookups, with GeoJSON transform, and with basic projection
+    const data = model.requestDataName(MAIN);
+
+    const width = model.getSizeSignalRef('width');
```
kanitw: I suggest using `{signal: model.getName('width')}` as I am about to restrict `getSizeSignalRef()` for using in assemble only.  

If you haven't, make sure that in `assemble`, you replace `size` signals with updated names.

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,65 @@
...
+    const height = model.getSizeSignalRef('height');
+    const size = [width.signal, height.signal];
+
+    const isGeoshapeMark = markDef && markDef.type === GEOSHAPE;
+    const hasEncodedProjection = encoding && [X, Y, SHAPE].some((channel) => isProjectionFieldDef(encoding[channel]));
+    const inheritedProjection = config && config.projection;
```
kanitw: `config` is never null -- you can just read config.projection when needed.

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export function parseProjection(model: Model, config: Config) {
+  if (model instanceof UnitModel) {
+    model.component.projection = parseUnitProjection(model, config);
+  } else {
+    const mergedProjection = parseNonUnitProjections(model);
+  }
+}
+
+function parseUnitProjection(model: UnitModel, config: Config): ProjectionComponent {
```
kanitw: No need to take `config` as parameter.  You can access it from `model.config`. 
The init part already passed `config` to all models in the model tree.

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  return undefined;
+}
+
+function parseNonUnitProjections(model: Model): ProjectionComponent {
+  const merged = model.children.reduce((projections, child) => {
+    const projection = child.component.projection;
+    if (projection) {
+      projections.forEach((other) => {
```
kanitw: Here you seems to always merge -- see `parseNonUnitScaleCore` how I make two passes over the children.  The first pass checks whether I can merge all of them.  If can merge, then make second pass to really merge.

src/compile/scale/parse.ts
```diff
@@ -40,16 +41,21 @@ function parseUnitScaleCore(model: UnitModel): ScaleComponentIndex {
   const {encoding, config} = model;
   const mark = model.mark();
 
-  return SCALE_CHANNELS.reduce((scaleComponents: ScaleComponentIndex, channel: ScaleChannel) => {
+  const scaleComponents: ScaleComponentIndex = {};
+  SCALE_CHANNELS.forEach((channel: ScaleChannel) => {
+    if (mark === GEOSHAPE && channel === SHAPE) {
```
kanitw: Add comment why geoshape's shape is always totally skipped?

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,65 @@
...
+import {VgProjection} from '../../vega.schema';
+import {Model} from '../model';
+import {UnitModel} from '../unit';
+import {ProjectionComponent} from './component';
+
+export function parseProjection(model: Model, config: Config) {
+  if (model instanceof UnitModel) {
+    model.component.projection = parseUnitProjection(model, config);
+  } else {
+    const mergedProjection = parseNonUnitProjections(model);
```
willium: ah right, I was going to check if undefined

src/compile/scale/parse.ts
```diff
@@ -45,15 +46,20 @@ function parseUnitScaleCore(model: UnitModel): ScaleComponentIndex {
...
+      return scaleComponents;
+    }
+
     let fieldDef: FieldDef<string>;
     let specifiedScale: Scale = {};
 
     const channelDef = encoding[channel];
 
-    if (isFieldDef(channelDef)) {
+    if (isFieldDef(channelDef) && isScaleFieldDef(channelDef)) {
```
willium: This was giving me errors in my editor (the linter), but now I wonder if this is what is causing my issues @kanitw

kanitw: What happen if scale is undefined?

willium: I dont think its causing the issues. I slightly rewrote it and the issues are the same

src/encoding.ts
```diff
@@ -77,11 +77,12 @@ export interface Encoding<F> {
...
 
   /**
-   * The symbol's shape (only for `point` marks). The supported values are
+   * For `point` marks the supported values are
    * `"circle"` (default), `"square"`, `"cross"`, `"diamond"`, `"triangle-up"`,
    * or `"triangle-down"`, or else a custom SVG path string.
+   * For `geoshape` marks it should be a fielddef of the geojson data
    */
-  shape?: Conditional<LegendFieldDef<F>, ValueDef<string>>; // TODO: maybe distinguish ordinal-only
+  shape?: Conditional<FieldDef<F>, ValueDef<string>>; // TODO: maybe distinguish ordinal-only
```
kanitw: This is the problematic line as shape can still be `LegendFieldDef` for point.

src/compile/compile.ts
```diff
@@ -71,6 +71,16 @@ function assembleTopLevelModel(model: Model, topLevelProperties: TopLevelPropert
   // Config with Vega-Lite only config removed.
   const vgConfig = model.config ? stripAndRedirectConfig(model.config) : undefined;
 
+  // Assemble data before projection
+  const data = [].concat(
```
willium: I don't see that on master... this needs to happen before projection, unless you mean thats no longer needed... I still prefer it up here...

src/compile/data/transforms.ts
```diff
@@ -48,6 +50,48 @@ export class FilterNode extends DataFlowNode {
...
 }
 
+export class GeoJSONNode extends DataFlowNode {
+  public clone() {
+    return new GeoJSONNode(duplicate(this.fields), this.geojson, this.signal);
+  }
+
+  public static make(model: ModelWithField): GeoJSONNode {
+    const geo = model.reduceFieldDef((rel, def, channel) => {
+      if ((contains([X, Y], channel) && contains([LATITUDE, LONGITUDE], def.type))
```
willium: Yes. that doesn't _really_ make any sense in Vega-Lite, but the GeoJSON transform does support that, so we might as well compile it correctly.

src/compile/legend/parse.ts
```diff
@@ -49,9 +53,12 @@ function getLegendDefWithScale(model: UnitModel, channel: Channel): VgLegend {
 }
 
 export function parseLegendForChannel(model: UnitModel, channel: NonspatialScaleChannel): LegendComponent {
-  const fieldDef = model.fieldDef(channel);
   const legend = model.legend(channel);
+  const fieldDef = model.fieldDef(channel);
 
+  if (!model.getScaleComponent(channel)) {
```
willium: doesn't this do that?

test/compile/compile.test.ts
```diff
@@ -43,6 +43,40 @@ describe('Compile', function() {
...
+        "data": {
+          "url": "data/us-10m.json",
+          "format": {
+            "type": "topojson",
+            "feature": "states"
+          }
+        },
+        "encoding": {}
+      }).spec;
+      console.log(spec);
```
willium: Remove this...

src/compile/legend/parse.ts
kanitw: It does. 

I mean move the logic to `parseUnitLegend` so that the structure of both `parseUnitLegend` and `parseUnitAxis` are similar.

src/compile/data/transforms.ts
```diff
@@ -45,6 +48,48 @@ export class FilterNode extends DataFlowNode {
...
 }
 
+export class GeoJSONNode extends DataFlowNode {
+  public clone() {
+    return new GeoJSONNode(duplicate(this.fields), this.geojson, this.signal);
+  }
+
+  public static make(model: ModelWithField): GeoJSONNode {
+    const geo = model.reduceFieldDef((rel, fieldDef, channel) => {
+      if ((contains([X, Y], channel) && contains([LATITUDE, LONGITUDE], fieldDef.type))
```
kanitw: From the earlier ([collaped conversaton](https://github.com/vega/vega-lite/pull/2734#discussion_r133342790))
I'm confused.  If you say this doesn't make any sense, why do you generate it? 

Looking at `geo_layer.vg.json`, it seems like both geojson and geopoint are required though.

src/compile/facet.ts
```diff
@@ -166,6 +166,10 @@ export class FacetModel extends ModelWithField {
     return assembleScalesForModel(this);
   }
 
+  public assembleProjections(): VgProjection[] {
```
kanitw: Let's get rid of this method -- just like https://github.com/vega/vega-lite/pull/2889

src/compile/data/transforms.ts
willium: Per: https://github.com/vega/vega-geo/issues/2#issuecomment-320533717
"The two params don't have to be mutually exclusive. The transform could handle both in one shot." So we might as well allow it. Otherwise we have to figure out which one they meant...

willium: Wait--I misread your original question. I think the answer to your original question is "no." It generates whichever is applicable to the channel for fieldDef being considered...

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,140 @@
...
+      };
+    } else {
+      // main source is geojson, so we can just use that
+      data = model.requestDataName(MAIN);
+    }
+
+    return new ProjectionComponent(model.projectionName(true), {
+      ...(config.projection || {}),
+      ...(specifiedProjection || {}),
+    } as Projection, [model.getSizeSignalRef('width'), model.getSizeSignalRef('height')], [data]);
```
kanitw: Why do you need to do casting (`as Projection`) here?

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+  });
+
+  // it cached one and all other children share the same projection,
+  if (nonUnitProjection && mergable) {
+    // so we can elevate it to the layer level
+    const name = model.projectionName(true);
+    const modelProjection = new ProjectionComponent(
+      name,
+      duplicate(nonUnitProjection.specifiedProjection),
```
kanitw: Do we really need to make duplicate for all of these (except `data`)?  (Since we don't really mutate them later?)

src/compile/legend/parse.ts
```diff
@@ -23,8 +26,10 @@ export function parseLegend(model: Model) {
 }
 
 function parseUnitLegend(model: UnitModel): LegendComponentIndex {
-  return [COLOR, SIZE, SHAPE, OPACITY].reduce(function(legendComponent, channel) {
-    if (model.legend(channel)) {
+  const {encoding} = model;
+  return [COLOR, SIZE, SHAPE, OPACITY].reduce(function (legendComponent, channel) {
+    const def = encoding[channel];
+    if (model.legend(channel) && !(isFieldDef(def) && (channel === SHAPE && def.type === GEOJSON))) {
```
kanitw: Since you already check the same thing in scale, maybe

`model.getScaleComponent[channel] && model.legend(channel)` is sufficient?

src/compile/unit.ts
```diff
@@ -56,7 +59,7 @@ export class UnitModel extends ModelWithField {
   public readonly stack: StackProperties;
 
   protected specifiedAxes: AxisIndex = {};
-
+  public specifiedProjection: Projection = {};
```
kanitw: Why is this one public, while others are protected?

src/fielddef.ts
```diff
@@ -134,8 +133,11 @@ export interface FieldDefBase<F> {
  */
 export interface FieldDef<F> extends FieldDefBase<F> {
   /**
-   * The encoded field's type of measurement. This can be either a full type
-   * name (`"quantitative"`, `"temporal"`, `"ordinal"`,  and `"nominal"`).
+   * The encoded field's type of measurement. This can be either a full basic type
+   * name (`"quantitative"`, `"temporal"`, `"ordinal"`,  and `"nominal"`)
+   * or an initial character of the basic type name (`"Q"`, `"T"`, `"O"`, `"N"`).
```
kanitw: Note that I hide 

> or an initial character of the basic type name (`"Q"`, `"T"`, `"O"`, `"N"`).

from master as it is not in the schema.

src/compile/projection/parse.ts
willium: I like it :D we don't need it

src/fielddef.ts
willium: Looks like my rebase missed that. Will fix

examples/specs/geo_choropleth.vl.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "width": 500,
+  "height": 300,
+  "mark": "geoshape",
```
kanitw: Stylistically, we normally put `mark` next to `encoding` for all examples.

examples/specs/geo_choropleth.vl.json
```diff
@@ -0,0 +1,38 @@
...
+  }],
+  "projection": {
+    "type": "albersUsa"
+  },
+  "encoding": {
+    "color": {
+      "field": "rate",
+      "type": "quantitative"
+    },
+    "opacity": {
```
kanitw: Once you write docs, you might wanna put

"opacity": {"value": 0.7}  

in one line so the spec isn't too long in the docs-- same for projection and other files.

examples/specs/geo_layer.vl.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    {
+      "data": {
+        "url": "data/airports.csv",
+        "format": {
+          "type": "csv"
+        }
+      },
+      "mark": "circle",
+      "projection": {
```
kanitw: Put projection before mark?

examples/specs/geo_point.vl.json
```diff
@@ -0,0 +1,28 @@
...
+    "url": "data/zipcodes.csv",
+    "format": {
+      "type": "csv"
+    }
+  },
+  "transform": [{"calculate": "substring(datum.zip_code, 0, 1)", "as": "digit"}],
+  "mark": "circle",
+  "encoding": {
+    "x": {
+      "field": "longitude",
```
kanitw: having both field and type represent the same thing is a bit redundant and verbose-- I wonder if we can offer a shorthand syntax (and automatically augment once of field or type)?

examples/specs/geo_point.vl.json
```diff
@@ -0,0 +1,28 @@
...
+    "type": "albersUsa"
+  },
+  "data": {
+    "url": "data/zipcodes.csv",
+    "format": {
+      "type": "csv"
+    }
+  },
+  "transform": [{"calculate": "substring(datum.zip_code, 0, 1)", "as": "digit"}],
+  "mark": "circle",
```
kanitw: Since you use circle mark, name this example `geo_circle`?

src/compile/projection/parse.ts
kanitw: Casting can cloak compile error for type mismatches. 
Thus, we generally avoid casting `as` unless it is really needed.  (e.g., when we run into TS bugs.)

examples/specs/geo_point.vl.json
willium: I don't think they really do... field is the data field, type is how we parse it in the projection. Firld could just as easily be "long" or "lng" or "city__longitude" or "barfoo"

kanitw: ah you're right -- I just looked at `zipcodes.csv` :)   Sorry!

src/compile/legend/parse.ts
willium: in cases like these, I'd prefer that we do this kind of scoping so the linter will catch type mismatch locally instead of having to chase it down

kanitw: I'm not sure why it helps in this case.

src/compile/unit.ts
willium: it needs to be publicized so I can access it from projection/parse.ts

willium:
I believe I addressed all your comments @kanitw.

src/compile/compile.ts
kanitw: It should no longer cause side effect, so no longer need to move it.  (Or at least, remove the comment that it has to come before projection

kanitw:
- [ ] Add example for lat/long type without explicit `projection`
- [ ] For `geoshape`'s default color range, I wonder we should use "heatmap" by default, like rect?

domoritz:
Can you rebase so that we can start the process of merging this PR?

willium:
Sure. I’ll try to get it a good place soon. Busy busy busy next couple of week.

domoritz:
No rush, I need to sleep for a bit first after this release :-)

eibanez:
I was wondering if there is a time table to including this in the official release.

Also, I opened a related issue in vega so that the `null` projection is recognized in the schema: https://github.com/vega/vega/issues/1086

I assume that would be relevant here too

domoritz:
I think we are just waiting for @willium to resolve the conflicts, will merge then, and then together with @arvind add selections. A full release with selections will need some time.

src/compile/data/geopoint.ts
```diff
@@ -0,0 +1,64 @@
...
+export class GeoPointNode extends DataFlowNode {
+  public clone() {
+    return new GeoPointNode(this.projection, duplicate(this.fields), duplicate(this.as));
+  }
+
+  constructor(private projection: string, private fields: string[], private as: string[]) {
+    super();
+  }
+
+  public static makeAll(model: ModelWithField): GeoPointNode[] {
```
willium: this is different. If they're using a line, I can pull out multiple GeoPoint transforms from one model. Is this a good way of doing that @domoritz?

willium:
@domoritz done. @kanitw I changed it to heatmap. However, no examples look very good with `mercator` which is the default projection. If you can come up with an example I can add it. I don't think it's necessary anyways.

src/compile/mark/mark.ts
```diff
@@ -10,6 +10,8 @@ import {UnitModel} from '../unit';
 import {area} from './area';
 import {bar} from './bar';
 import {MarkCompiler} from './base';
+import {geoshape} from './geoshape';
+import {normalizeMarkDef} from './init';
```
domoritz: Unused import

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,140 @@
...
+    // both have property and an equal value for property
+    if (first.explicit.hasOwnProperty(prop) &&
+      second.explicit.hasOwnProperty(prop) &&
+      JSON.stringify(first.get(prop)) === JSON.stringify(second.get(prop))) {
+      return true;
+    }
+    return false;
+  });
+
+  const size = JSON.stringify(first.size) === JSON.stringify
```
domoritz: size isn't just a number?

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,140 @@
...
+    (second.size);
+  if (size) {
+    if (properties) {
+      return first;
+    } else if (JSON.stringify(first.explicit) === JSON.stringify
```
domoritz: I don't think you need to stringify `{}`

test/compile/projection/parse.test.ts
```diff
@@ -0,0 +1,327 @@
...
+                "type": "topojson",
+                "feature": "states"
+              }
+            },
+            "encoding": {}
+          },
+          {
+            "data": {
+              "url": "data/airports.csv",
+              "format": {
```
domoritz: drop explicit type here and in other tests

src/compile/projection/parse.ts
willium: nope, could be a signalRef

src/compile/data/geopoint.ts
```diff
@@ -0,0 +1,63 @@
...
+        points.push(pair);
+      }
+    });
+
+    if (points.length <= 0 || keys(points[0]).length <= 0) { // no points found
+      return null;
+    }
+
+    const nodes: GeoPointNode[] = [];
+    points.forEach((coordinates) => {
```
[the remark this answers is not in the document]
willium: sure, good idea

src/compile/data/geopoint.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    interface TypeField {
+      [key: string]: string;
+    }
+
+    const points: TypeField[] = [];
+
+    const possiblePairs: SingleDefChannel[][] = [[X, Y], [X2, Y2]];
+    possiblePairs.forEach((posssiblePair) => {
```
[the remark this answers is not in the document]
willium: don't think this is as obviously a good idea, so going to not do it

src/compile/data/geopoint.ts
```diff
@@ -0,0 +1,63 @@
...
+  constructor(private projection: string, private fields: string[], private as: string[]) {
+    super();
+  }
+
+  public static makeAll(model: ModelWithField): GeoPointNode[] {
+    if (!model.projectionName()) {
+      return null;
+    }
+
+    interface TypeField {
```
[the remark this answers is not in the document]
willium: didn't know this existed! thanks

src/fielddef.ts
```diff
@@ -143,6 +143,7 @@ export interface FieldDefBase<F> {
 export interface FieldDef<F> extends FieldDefBase<F> {
   /**
    * The encoded field's type of measurement (`"quantitative"`, `"temporal"`, `"ordinal"`, or `"nominal"`).
+   * It can also be a geo type (`"latitude"`, `"longitude"`, and `"geojson"`) when applicable.
```
kanitw: Can you describe specifically what "when applicable" mean?  (So that users understand right away and do not have to try to find other docs to read.)

src/encoding.ts
```diff
@@ -75,10 +75,11 @@ export interface Encoding<F> {
   size?: FieldDefWithCondition<MarkPropFieldDef<F>> | ValueDefWithCondition<MarkPropFieldDef<F>>;
 
   /**
-   * The symbol's shape (only for `point` marks). The supported values are
+   * For `point` marks the supported values are
    * `"circle"` (default), `"square"`, `"cross"`, `"diamond"`, `"triangle-up"`,
    * or `"triangle-down"`, or else a custom SVG path string.
    * __Default value:__ If undefined, the default shape depends on [mark config](config.html#point-config)'s `shape` property.
+   * For `geoshape` marks it should be a fielddef of the geojson data
```
kanitw: No description should be added below "__Default value:__" line

Could you please move it above and add a line break between the main description and the  "__Default value:__"  (Yeah we should have added it prior to this PR.)


fielddef => field definition

src/fielddef.ts
```diff
@@ -321,10 +322,13 @@ export function isDiscrete(fieldDef: FieldDef<Field>) {
   switch (fieldDef.type) {
     case 'nominal':
     case 'ordinal':
+    case 'geojson':
       return true;
     case 'quantitative':
       return !!fieldDef.bin;
-    case 'temporal':
+      case 'latitude':
```
kanitw: inconsistent tabbing

src/projection.ts
```diff
@@ -0,0 +1,45 @@
...
+import {VgProjectionType} from './vega.schema';
+
+export type ProjectionType = VgProjectionType;
+
+export interface Projection {
+  /**
+   * The type of the projection. The default is "mercator".
+   */
+  type?: ProjectionType;
+  /**
```
kanitw: Please add line breaks between properties and JS Docs of the next properties

src/spec.ts
```diff
@@ -103,6 +104,11 @@ export interface GenericUnitSpec<E extends Encoding<any>, M> extends BaseSpec, L
    */
   encoding: E;
 
+  /**
+   * A geo projection
```
kanitw: An object defining properties of geographic projection.

Please also describe more if this can be null or something? 
Does it work only for certain type of marks? (If all marks, ignore my comment.) 

What's the `__Default value:__`, is there any case it is generated by default?

site/docs/projection.md
```diff
@@ -0,0 +1,38 @@
...
+---
+## Projection Overview
+Projection's are specified at the mark specification level, alongside encoding.
+
+{:#properties}
+## Projection Properties
+{% include table.html props="type,clipAngle,clipExtent,center,rotate,precision" source="Projection" %}
+
+
+*Note*: All [properties](#properties) of projections are **optional** with defaults as defined in the [Vega projection properties](https://vega.github.io/vega/docs/projections/#properties). Because of this, defined marks that don't have explicitly defined projections may implicitly derive a projection. Implicit projections will be added for any [`geoshape`](geoshape.html) mark or any mark that has fields of type [`latitude`](type.html#latitude), [`longitude`](type.html#longitude), or [`geojson`](type.html#geojson).
```
kanitw: > All [properties](#properties) of projections are **optional** with defaults as defined in the [Vega projection properties]

Can you just copy all default values and add them to the JSDocs in the interface (reduce the need for users to jump to Vega docs, which can be quite confusing as it's totally a different language.)


> defined marks that don't have explicitly defined projections may implicitly derive a projection. 

Unclear what is "defined marks"?

site/docs/projection.md
```diff
@@ -0,0 +1,36 @@
...
+
+{:#properties}
+## Projection Properties
+{% include table.html props="type,clipAngle,clipExtent,center,rotate,precision" source="Projection" %}
+
+*Note*: All [properties](#properties) of projections are **optional** with defaults as defined in the [Vega projection properties](https://vega.github.io/vega/docs/projections/#properties). Because of this, defined marks that don't have explicitly defined projections may implicitly derive a projection. Implicit projections will be added for any [`geoshape`](geoshape.html) mark or any mark that has fields of type [`latitude`](type.html#latitude), [`longitude`](type.html#longitude), or [`geojson`](type.html#geojson).
+
+
+## Projection Types
+For a list of all supported projection types, please see the [Vega projection types](https://vega.github.io/vega/docs/projections/#types).
```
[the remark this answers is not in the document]
kanitw: I would just copy the table from Vega's docs, reducing the need to jump around (and get lost by clicking Vega's menu instead of Vega after jumping there.)

site/docs/projection.md
```diff
@@ -0,0 +1,38 @@
...
+  ...,
+  "config": {          // Configuration Object
+    "projection": { ... },   // - Projection Configuration
+    ...
+  }
+}
+```
+
+The `projection` property of the [`config`](config.html) object determines the default properties and transformations applied to different types of [projections](projection.html).
+The projection config can contain any of the properties as specified above.
```
kanitw: > contain any of the properties as specified above

any projection properties

Please also add an anchor link to the header above.  :)

site/docs/mark/geoshape.md
```diff
@@ -0,0 +1,38 @@
...
+{
+  "data": ... ,
+  "mark": "geoshape",
+  "encoding": ... ,
+  ...
+}
+```
+
+The `geoshape` mark represents an arbitrary shapes whose geometry is determined by specified GeoJSON shape data.
+
```
kanitw: Need one or a few examples.  (All other mark pages have at least one example.)

site/docs/encoding/type.md
```diff
@@ -5,7 +5,12 @@ permalink: /docs/type.html
 ---
 
 If a field is specified, the channel definition **must** describe the encoded data's [type of measurement (level of measurement)](https://en.wikipedia.org/wiki/Level_of_measurement).
-The supported data types are: `"quantitative"`, `"temporal"`, `"ordinal"`, and `"nominal"`.
+The supported data types are: `"quantitative"`, `"temporal"`, `"ordinal"`, `"nominal"`, `"latitude"`, `"longitude"`, and `"geojson"`.
+
+{% include table.html props="type" source="FieldDef" %}
+
+**Note**:
```
kanitw: Thanks for adding this note.  This is definitely useful to add.  

A few thoughts:

- We should add it to `type.md` too

- For this page, I think maybe it's better to add this description to the JSDoc in the `FieldDef` interface so the explanation appears in the table.  Plus, Altair users can directly benefit from this explanation.

site/docs/projection.md
```diff
@@ -0,0 +1,38 @@
+---
+layout: docs
+menu: docs
+title: Projection
+permalink: /docs/projection.html
+---
+## Projection Overview
+Projection's are specified at the mark specification level, alongside encoding.
```
kanitw: I think we usually begin a docs by explaining the concept. You can copy some of the text from  https://vega.github.io/vega/docs/projections

> specified at the mark specification level, alongside encoding

Also, there is no concept of "mark specification".  I think you only mean specification.  Please also make it clear whether all types of specification support `projection`.  (Is it only unit and layer or all types of specification?)

site/docs/projection.md
```diff
@@ -0,0 +1,38 @@
...
+
+{:#properties}
+## Projection Properties
+{% include table.html props="type,clipAngle,clipExtent,center,rotate,precision" source="Projection" %}
+
+
```
kanitw: Missing example

site/docs/projection.md
```diff
@@ -0,0 +1,38 @@
...
+menu: docs
+title: Projection
+permalink: /docs/projection.html
+---
+## Projection Overview
+Projection's are specified at the mark specification level, alongside encoding.
+
+{:#properties}
+## Projection Properties
+{% include table.html props="type,clipAngle,clipExtent,center,rotate,precision" source="Projection" %}
```
kanitw: From Vega docs, 

> In addition to the shared properties above, the following properties are supported for specific projection types in the d3-geo-projection library: coefficient, distance, fraction, lobes, parallel, radius, ratio, spacing, tilt.

Do we support these properties too?

src/projection.ts
```diff
@@ -0,0 +1,45 @@
+
+import {VgProjectionType} from './vega.schema';
+
+export type ProjectionType = VgProjectionType;
+
+export interface Projection {
+  /**
+   * The type of the projection. The default is "mercator".
```
kanitw: Add link to the section that describe all supported projection type?

Also use `__Default value:__` to describe default value.

src/compile/data/geojson.ts
```diff
@@ -0,0 +1,48 @@
...
+export class GeoJSONNode extends DataFlowNode {
+  public clone() {
+    return new GeoJSONNode(duplicate(this.fields), this.geojson, this.signal);
+  }
+
+  public static make(model: ModelWithField): GeoJSONNode {
+    const geo = model.reduceFieldDef((rel, fieldDef, channel) => {
+      if ((contains([X, Y, X2, Y2], channel) && contains([LATITUDE, LONGITUDE], fieldDef.type))
+        || (channel === SHAPE && fieldDef.type === GEOJSON)) {
+        rel[fieldDef.type] = fieldDef.field;
```
kanitw: I wonder if using dictionary here would make it work when we want to use both x/y and y2/x2 (e.g., for line segment rules)?   

It seems like this logic will make fieldDef for x2 overrides x's

src/compile/data/geopoint.ts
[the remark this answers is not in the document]
kanitw: Yeah, I think map is a bad idea here.

src/compile/legend/parse.ts
```diff
@@ -51,6 +54,10 @@ export function parseLegendForChannel(model: UnitModel, channel: NonPositionScal
   const fieldDef = model.fieldDef(channel);
   const legend = model.legend(channel);
 
+  if (!model.getScaleComponent(channel)) {
```
kanitw: Move the logic to `parseUnitLegend` so that the structure of both `parseUnitLegend` and `parseUnitAxis` are similar?

src/compile/mark/geoshape.ts
```diff
@@ -0,0 +1,37 @@
...
+import {MarkCompiler} from './base';
+
+export const geoshape: MarkCompiler = {
+  vgMark: 'shape',
+  encodeEntry: (model: UnitModel) => {
+    const {config} = model;
+    return {
+      ...mixins.color(model),
+      ...mixins.text(model, 'tooltip'),
+      ...mixins.nonPosition('opacity', model)
```
kanitw: If we merge the href PR before we merge this one, we will have to add href here too :) 

https://github.com/vega/vega-lite/pull/3229

src/compile/mark/geoshape.ts
```diff
@@ -0,0 +1,37 @@
...
+import {VgGeoShapeTransform, VgPostEncodingTransform} from '../../vega.schema';
+import {getMarkConfig} from '../common';
+import {MarkCompiler} from './base';
+
+export const geoshape: MarkCompiler = {
+  vgMark: 'shape',
+  encodeEntry: (model: UnitModel) => {
+    const {config} = model;
+    return {
+      ...mixins.color(model),
```
kanitw: missing `...mixins.markDefProperties(model.markDef, true)`?

src/compile/mark/geoshape.ts
```diff
@@ -0,0 +1,37 @@
...
+      ...mixins.text(model, 'tooltip'),
+      ...mixins.nonPosition('opacity', model)
+    };
+  },
+  postEncodingTransform: (model: UnitModel): VgPostEncodingTransform[] => {
+    const {encoding, markDef} = model;
+    const shapeDef = encoding.shape;
+
+    if ((isFieldDef(shapeDef) && shapeDef.type === GEOJSON) || markDef.type === GEOSHAPE) {
```
kanitw: > markDef.type === GEOSHAPE

Can this function be called when mark type is not geoshape since this is in geoshape file?

kanitw: Restructure a bit to include warnings?
```
if (isFieldDef(shapeDef)) {
   if (shapeDef.type === GEOJSON) {
      ...
   } else {
      ... throw warning that field type is not compatible with geoshape
   }
} else {
   ...throw warning that value definition is not supported for `shape` of geoshape
} 
```

src/compile/mark/rule.ts
```diff
@@ -15,11 +16,23 @@ export const rule: MarkCompiler = {
       return {};
     }
 
+    let secondary = {};
```
kanitw: I added a cleaner fix for this in  https://github.com/vega/vega-lite/pull/3265.  
You can help review and merge the PR so you can rebase this branch.

src/compile/model.ts
```diff
@@ -229,7 +239,8 @@ export abstract class Model {
     this.renameTopLevelLayoutSize();
 
     this.parseSelection();
-    this.parseData(); // (pathorder) depends on markDef; selection filters depend on parsed selections.
+    this.parseProjection();
+    this.parseData(); // (pathorder) depends on markDef; selection filters depend on parsed selections; depends on projection.
```
kanitw: > depends on projection.

which part of data depends on projection? 

(Other text on this line is a bit more specific)

src/compile/projection/assemble.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function assembleProjectionForModel(model: Model): VgProjection[] {
+  const component = model.component.projection;
+  if (!component || component.merged) {
+    return [];
+  }
+
+  const projection = component.combine();
+  const {name, ...rest} = projection;
```
kanitw: not sure why you have to extract `name` and `rest` here, simply to merge them again at the end?

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,140 @@
...
+    return new ProjectionComponent(model.projectionName(true), {
+      ...(config.projection || {}),
+      ...(specifiedProjection || {}),
+    }, [model.getSizeSignalRef('width'), model.getSizeSignalRef('height')], [data]);
+  }
+
+  return undefined;
+}
+
+function dryMerge(first: ProjectionComponent, second: ProjectionComponent): ProjectionComponent {
```
kanitw: Can you add a comment to explain what dryMerge means?

src/compile/scale/parse.ts
```diff
@@ -50,15 +52,21 @@ function parseUnitScaleCore(model: UnitModel): ScaleComponentIndex {
 
     const channelDef = encoding[channel];
 
+    // if mark is geoshape scale channel encodes the geojson
```
kanitw: This comment reads quite confusing.  

I think you mean "If mark is geoshape, we use projection instead of scale?"

kanitw:
- [ ] Related to https://github.com/vega/vega-lite/pull/2734#discussion_r161414134, I think you have to update table in spec.md and perhaps layer.md to include the new `projection` property.  (Maybe there are other properties not added yet as well.)

src/projection.ts
```diff
@@ -0,0 +1,45 @@
+
+import {VgProjectionType} from './vega.schema';
+
+export type ProjectionType = VgProjectionType;
+
+export interface Projection {
```
kanitw: Looks like  descriptions of some properties in the Vega docs are a bit more comprehensive and have links to related bl.ock.org examples

src/spec.ts
```diff
@@ -388,7 +395,9 @@ function normalizeRangedUnit(spec: UnitSpec) {
 
 // FIXME(#1804): re-design this
 function normalizeOverlay(spec: UnitSpec, overlayWithPoint: boolean, overlayWithLine: boolean, config: Config): LayerSpec {
-  const {mark, selection, encoding, ...outerSpec} = spec;
+  // _ is used to denote a dropped property of the unit spec
+  // which should not be carried over to the layer spec
+  const {mark, selection, projection: _p, encoding, ...outerSpec} = spec;
```
kanitw: Why should we drop projection for point overlay over lines? 

(Suppose we draw lines between different cities and we want to overlay points over each city?)

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,140 @@
...
+  const properties = every(PROJECTION_PROPERTIES, (prop) => {
```
kanitw: `properties` needs a better variable name. 

Perhaps, `allPropertiesAreConsistent` or something?

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,140 @@
...
+    // neither has the poperty
+    if (!first.explicit.hasOwnProperty(prop) &&
+      !second.explicit.hasOwnProperty(prop)) {
+      return true;
+    }
+    // both have property and an equal value for property
+    if (first.explicit.hasOwnProperty(prop) &&
+      second.explicit.hasOwnProperty(prop) &&
+      hash(first.get(prop)) === hash(second.get(prop))) {
```
kanitw: Why do we have to compared the hashed value? Is that because some property can be an object? 
(If so, please add a comment.)

src/compile/projection/parse.ts
```diff
@@ -0,0 +1,140 @@
...
+      return first;
+    } else if (hash(first.explicit) === hash
+      ({})) {
+      return second;
+    } else if (hash(second.explicit) === hash
+      ({})) {
+      return first;
+    }
+  }
+  return null;
```
kanitw: So if properties are not consistent, then we don't merge the projection at all?  

(If this is intended, please add some comment + explain why this should be the behavior.)

src/compile/projection/parse.ts
kanitw: Perhaps, `mergeIfNoConflict` or `mergeIfPropertiesAreConsistent` might be a better name?

src/compile/mark/geoshape.ts
kanitw: FYI, to make this easier for the future -- I added https://github.com/vega/vega-lite/pull/3280#event-1433598724

site/docs/encoding/type.md
willium: Can you clarify what you mean? (also do you mean `types.md`?

site/docs/projection.md
willium: We don't currently. Do you want to? I can add them in `projection.ts`

src/fielddef.ts
willium: I don't think this is a good idea. Adding anything more is just weirdly specific. Here's what I came up with, but it's honestly more confusing.
```
   * It can also be a geo type (`"latitude"`, `"longitude"`, and `"geojson"`) when projecting geographical coordinates using a `point` mark (on `x`, and `y` channels), a `rule` mark (on `x`, `x2`, `y`, and `y2` channels) or projecting a GeoJSON shapefile on a `geoshape` mark (on the `shape` channel).
```

site/docs/encoding/type.md
kanitw: - There is type.md

- You can put the note in `FieldDef` interface in `fielddef.ts`  instead of here.

site/docs/projection.md
kanitw: I think so.

src/fielddef.ts
kanitw: Cut this part -- it's too much

> using a `point` mark (on `x`, and `y` channels), a `rule` mark (on `x`, `x2`, `y`, and `y2` channels) or projecting a GeoJSON shapefile on a `geoshape` mark (on the `shape` channel)." 

and then paraphrase this part to be slightly human-friendly and link to related concept

> when projecting geographical coordinates

src/compile/data/geojson.ts
willium: you've discovered a bug, but unfortunately it's not so easy to fix. The bug isn't about whether or not we can project `rule`, but instead how `fit` is computed (it appears to work since `fit` on X2 and Y2 is similar to X and Y). You're correct in your assessment that X2 would override X (my bad).

src/fielddef.ts
willium: I don't really understand what you're saying

src/compile/data/geojson.ts
willium: this is non trivial. I will have to think about it more.

src/fielddef.ts
kanitw: Sorry, I mean something like this:

> It can also be a geographic type (`"latitude"`, `"longitude"`, and `"geojson"`) for using with geographic projection

and link projection to projection docs 


("when applicable" is vague and doesn't convey much information :D)

src/compile/data/geojson.ts
willium: Resolved.

domoritz:
- [ ] Add example for connected scatterplot on a map.

site/docs/encoding/type.md
```diff
@@ -5,7 +5,12 @@ permalink: /docs/type.html
 ---
 
 If a field is specified, the channel definition **must** describe the encoded data's [type of measurement (level of measurement)](https://en.wikipedia.org/wiki/Level_of_measurement).
-The supported data types are: `"quantitative"`, `"temporal"`, `"ordinal"`, and `"nominal"`.
+The supported data types are: `"quantitative"`, `"temporal"`, `"ordinal"`, `"nominal"`, `"latitude"`, `"longitude"`, and `"geojson"`.
```
kanitw: Add anchor  links for the types?

src/compile/mark/rule.ts
willium: I think I still need this actually. Can you verify?

src/compile/projection/assemble.ts
willium: I think I do for some weird reason to do with types. if I remove them from the bottom and just extend `projection` with the return, the type check yells at me. Basically a vega projection is all the properties of the vega-lite projection + name, size, and fit (which are algorithmically decided)

examples/specs/geo_choropleth.vl.json
```diff
@@ -0,0 +1,32 @@
...
+    "lookup": "id",
+    "from": {
+      "data": {
+        "url": "data/unemployment.tsv"
+      },
+      "key": "id",
+      "fields": ["rate"]
+    }
+  }],
+  "projection": {
```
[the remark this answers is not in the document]
willium: nah. we can support that later maybe.

src/compile/projection/assemble.ts
domoritz: This is fine. I added a comment.

src/compile/mark/rule.ts
domoritz: I don't think it is needed. I removed your changes and the rule example still works.

domoritz:
Add documentation with example for geo point.

src/compile/mark/geoshape.ts
kanitw: This comment seems unaddressed?

src/compile/projection/assemble.ts
```diff
@@ -26,7 +24,7 @@ export function assembleProjectionForModel(model: Model): VgProjection[] {
   }
 
   const projection = component.combine();
-  const {name, ...rest} = projection;
+  const {name, ...rest} = projection;  // name is always defined
```
kanitw: > // name is always defined

This confuses me, but Dom just explained to me the real reason behind this. 

Please change to .. "we need to extract `name` so that it is always present in the output and pass TS type validation" or something similar.

src/compile/projection/parse.ts
```diff
@@ -66,25 +75,25 @@ function dryMerge(first: ProjectionComponent, second: ProjectionComponent): Proj
     // both have property and an equal value for property
     if (first.explicit.hasOwnProperty(prop) &&
       second.explicit.hasOwnProperty(prop) &&
+      // some properties might be signals or objects and require hashing for comparison
       hash(first.get(prop)) === hash(second.get(prop))) {
```
kanitw: If we want exact comparison, we should use stringify here more than hash?

src/fielddef.ts
```diff
@@ -143,7 +142,7 @@ export interface FieldDefBase<F> {
 export interface FieldDef<F> extends FieldDefBase<F> {
   /**
    * The encoded field's type of measurement (`"quantitative"`, `"temporal"`, `"ordinal"`, or `"nominal"`).
-   * It can also be a geo type (`"latitude"`, `"longitude"`, and `"geojson"`) when applicable.
+   * It can also be a geo type (`"latitude"`, `"longitude"`, and `"geojson"`) when projecting geographical coordinates using a `"point"` mark (on `"x"`, and `"y"` channels), a `"rule"` mark (on `"x"`, `"x2"`, `"y"`, and `"y2"` channels) or projecting a GeoJSON shapefile on a `"geoshape"` mark (on the `"shape"` channel).
```
kanitw: "when a geographic projection is applied"?

(and link to projection page)

src/projection.ts
```diff
@@ -5,29 +5,51 @@ export type ProjectionType = VgProjectionType;
...
    */
   rotate?: number[];
+
   /**
-   * The desired precision of the projection. If precision is not specified, returns the projection’s current resampling precision which defaults to √0.5 ≅ 0.70710….
+   * Sets the threshold for the projection’s [adaptive resampling](http://bl.ocks.org/mbostock/3795544) to the specified value in pixels. This value corresponds to the [Douglas–Peucker distance](http://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm). If precision is not specified, returns the projection’s current resampling precision which defaults to `√0.5 ≅ 0.70710…`.
    */
   precision?: String;
+
+  /** The following properties are all supported for specific types of projections. Consult the d3-geo-projection library for more information: https://github.com/d3/d3-geo-projection */
```
kanitw: If you want do it a lazy way like this, use `/*` not `/**` otherwise this comment will be attached to only `coefficient`, which is not what you want.

src/vega.schema.ts
```diff
@@ -150,6 +150,17 @@ export type VgProjection = {
    * Used in conjunction with fit, provides the width and height in pixels of the area to which the projection should be automatically fit.
    */
   size?: VgSignalRef | (number | VgSignalRef)[];
+
+  /** The following properties are all supported for specific types of projections. Consult the d3-geo-projection library for more information: https://github.com/d3/d3-geo-projection */
```
kanitw: `/*`

site/docs/mark/text.md
```diff
@@ -43,6 +43,13 @@ Mapping a field to `text` channel of text mark sets the mark's text value. For e
 <span class="vl-example" data-name="text_scatterplot_colored"></span>
 
 
+## Geo Text
+
+With the [types](type.html) `longitude` and `latitude` instead of `x` and `y` and a corresponding [projection](projection.html), we can show text at accurate locatins. The example below shows the name of every US state capital at the location of the capital.
```
kanitw: locations

src/compile/mark/geoshape.ts
willium: This is wontfix for me. I don’t see why your code is better, and mine works. Yours causes marks to not get a post encode transform that should (when they’re just a geoshape and not geojson).

kanitw: My code might be wrong, but the current code is confusing. 

1) I don't understand why you have to check for mark type here it's already in the geoshape file.  
It seems unnecessary.  If you really allow other mark type to call this function, then this function belongs to a wrong place in the codebase. 

2) You don't have to use exactly my code, but you should throw warnings for unsupported case.

site/docs/mark/text.md
domoritz: Done

src/compile/projection/parse.ts
willium: hash is stringify.

PR #2737, opened 2017-08-01
Use images on the website that we have already compiled

domoritz:


scripts/deploy-gh.sh
```diff
@@ -11,8 +11,6 @@ git merge master --no-edit
 
 # build
 npm run presite
```
kanitw: I try to track but can't find where  we run `build:examples`? (Maybe I'm just tired and overlook them.)

domoritz: We don't need it since the examples in the repo have to be updated. There is no way for them to change.

kanitw: Ah genius!

kanitw:
Hmm test is still failing

domoritz:
Fixed

package.json
```diff
@@ -26,7 +26,7 @@
     "build": "npm run build:only",
     "build:only": "npm run tsc && cp package.json build && browserify src/index.ts -p tsify -d -s vl | exorcist build/vega-lite.js.map > build/vega-lite.js",
     "postbuild": "node node_modules/uglify-js/bin/uglifyjs build/vega-lite.js -cm --source-map content=build/vega-lite.js.map,filename=build/vega-lite.min.js.map -o build/vega-lite.min.js && npm run schema",
-    "build:examples": "npm run build:only && npm run build:examples-only",
+    "build:examples": "npm run data && npm run build:only && npm run build:examples-only",
```
kanitw: Note: This seems to make build:examples slower, which is a bit annoying. I'll merge this for now, but I wonder if we should include `node_modules/vega-datasets` on jekyll instead of copying data.

domoritz: Hmm, I guess since we set the base url, it should be fine.

domoritz: Doesn't work because jekyll doesn't let us use node_modules.

PR #2740, opened 2017-08-01
Facet with independent scales for layered plot

### LINKED ISSUE #2736: Make facet with independent scale of layer/concat work
Example spec 

```
{
  "description": "A trellis bar chart showing the US population distribution of age groups and gender in 2000.",
  "data": { "url": "data/population.json"},
  "transform": [
    {"filter": "datum.year == 2000"},
    {"calculate": "datum.sex == 2 ? 'Female' : 'Male'", "as": "gender"}
  ],
  "facet": {"column": {"field": "gender", "type": "nominal"}},
  "spec": {
    "layer": [{
      "mark": "line",
        "encoding": {
        "y": {
            "aggregate": "sum", "field": "people", "type": "quantitative",
            "axis": {"title": "population"}
        },
        "x": {
            "field": "age", "type": "ordinal",
            "scale": {"rangeStep": 17}
        },
        "color": {
            "field": "gender", "type": "nominal",
            "scale": {"range": ["#EA98D2","#659CCA"]}
        }
      }

    },{
      "mark": "point",
        "encoding": {
        "y": {
            "aggregate": "sum", "field": "people", "type": "quantitative",
            "axis": {"title": "population"}
        },
        "x": {
            "field": "age", "type": "ordinal",
            "scale": {"rangeStep": 17}
        },
        "color": {
            "field": "gender", "type": "nominal",
            "scale": {"range": ["#EA98D2","#659CCA"]}
        }
      }

    }]
  },
  "resolve": {
    "x": {"scale": "independent"}
  },
  "config": {"cell": {"fill": "yellow"}}
}
```

kanitw:
Fix part of #2736 

Forked from #2697


src/compile/data/facet.ts
```diff
@@ -75,24 +64,47 @@ export class FacetNode extends DataFlowNode {
...
+    const childIndependentFieldsWithStep: ChildIndependentFieldsWithStep = {};
 
+    for (const channel of ['x', 'y'] as ScaleChannel[]) {
+      const childScaleComponent = this.childModel.component.scales[channel];
+      if (childScaleComponent && !childScaleComponent.merged) {
+        const type = childScaleComponent.get('type');
+        const range = childScaleComponent.get('range');
+
+        if (hasDiscreteDomain(type) && isVgRangeStep(range)) {
+          const domain = assembleDomain(this.childModel, channel);
```
kanitw: @domoritz -- even though I call this `assembleDomain` during data assemble already, it does not seem to get the final domain -- possibly because the child nodes have not been walked. 

I'm not sure what's the best way to restructure this to work correctly. 

One structural problem is that assembleData still cause side effect to the data component.
Thus, refactor to move all the side effect to parse may help, but I better let you take a look and decide.

kanitw:
FYI, rebased to latest `hotfix-facet`.

src/compile/data/facet.ts
domoritz: Split in https://github.com/vega/vega-lite/pull/2810

kanitw: OK thanks -- but as commented here -- https://github.com/vega/vega-lite/issues/2736#issuecomment-320533885  -- it's not really the root cause here.

domoritz: Yeah, you're not going to get the right domain here yet because we haven't finished assembling the data. We need to rename the data after we finish assembling the data.

kanitw: Can we make data assemble pure function? 
It shouldn't cause side effect anyway.

domoritz: The only side effect is that we can request the correct data names. Changing this is a major refactoring. 

Writing the dataflow stuff already took a a few weeks :-P

kanitw: Yeah but here we need to know the field of the child (if the child merge domains of its children), and the only way to know the field is from domain, but then domain cannot be merged correctly here because it depends on the side effect that hasn't happened.  

If you don't want to refactor, then tell me how would you fix this problem otherwise?

src/compile/scale/domain.ts
```diff
@@ -408,10 +409,45 @@ export function mergeDomains(domains: VgNonUnionDomain[]): VgDomain {
...
   if (isDataRefDomain(domain) && isString(domain.field)) {
     return domain.field;
+  } else if (isDataRefUnionedDomain(domain)) {
+    let field;
+    for (const nonUnionDomain of domain.fields) {
+      if (isDataRefDomain(nonUnionDomain) && isString(nonUnionDomain.field)) {
+        if (!field) {
+          field = nonUnionDomain.field;
+        } else if (field !== nonUnionDomain.field) {
+          log.warn('Detected faceted independent scales that union domain of multiple fields from different data sources.  We will use the first field.  The result cell size may be incorrect.');
```
kanitw: @domoritz I kinda feel obligated to throw warning when something might be incorrect, but this warning also looks scary.  Any thoughts?

kanitw: I think this should be fine given the dataflow will be optimized for this case soon.

kanitw:
image

domoritz:
LGTM

Issue #2717, opened 2017-07-28
Improve Axis / Grid Styles


kanitw:
- [x] Bring back `tickCount` = `Math.floor(width/40)`. 
- [x] Make default axis lines slightly lighter too
- [ ] ~~Make grid lines slightly lighter~~
- [ ] Varying the major and minor tick colors / length.

kanitw:
FYI, I updated this to become "Improve Axis / Grid Styles".

kanitw:
Mostly fixed in https://github.com/vega/vega-lite/pull/2745

I think we can't really vary major and minor ticks color/length in Vega 3.0, so I'll ignore it for now.

PR #2745, opened 2017-08-02
Improve Axis Styles

kanitw:
Fix of https://github.com/vega/vega-lite/issues/2717


arvind:
For my edification, could someone explain why `tickCount` is set to `width/40` or `height/40`?

PR #2750, opened 2017-08-02
Replace collect transform with mark sort

### LINKED ISSUE #2669: Replace Collect transform with mark sort
See vega/vega#930. While the Collect transform will sort input data tuples correctly, its sort order may not persist past the data join. Instead, we should be setting the `sort` property of the area/line marks. 

This is can be seen with the following Vega-Lite spec. If we add `"sort": {"field": "x"},` to the definition of the first area mark, the glitch is resolved.

areaglitches

```json
{
  "data": {"url": "data/sp500.csv"},
  "vconcat": [{
    "width": 480,
    "transform": [{"filter": {"selection": "brush"}}],
    "mark": "area",
    "encoding": {
      "x": {
        "field": "date",
        "type": "temporal",
        "axis": {"title": "", "labelAngle": 0}
      },
      "y": {"field": "price","type": "quantitative"}
    }
  },
  {
    "width": 480,
    "height": 60,
    "mark": "area",
    "selection": {
      "brush": {"type": "interval", "encodings": ["x"]}
    },
    "encoding": {
      "x": {
        "field": "date",
        "type": "temporal",
        "axis": {"format": "%Y", "labelAngle": 0}
      },
      "y": {
        "field": "price",
        "type": "quantitative",
        "axis": {"tickCount": 3, "grid": false}
      }
    }
  }]
}
```

kanitw:
Fix #2669


kanitw:
The output SVGs are still weird -- I've filed an issue in https://github.com/vega/vega/issues/954

kanitw:
They should look good now.  

I notice that some svg has changed data but they look pretty much the same.  

The changes are mainly because different lines got swapped layer order (as collect transform has effect on the order of the data points between groups too -- while mark's `sort` does not.) 

image

PR #2752, opened 2017-08-02
[WIP] Bin area issue

kanitw:


kanitw:
I think this is not the right solution

Issue #2663, opened 2017-07-13
stacked_area_binned broken


domoritz:
On master

<img width="363" alt="screen shot 2017-07-12 at 23 42 58" src="https://user-images.githubusercontent.com/589034/28153724-e1d447e2-675b-11e7-913d-7fc906837123.png">

kanitw:
https://github.com/vega/vega-lite/issues/2669 might be related?

kanitw:
https://github.com/vega/vega-lite/pull/2753

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/2753

PR #2753, opened 2017-08-03
Correct broken `stacked_area_binned broken` 

kanitw:
by:
-  Calculating `bin_mid` and imputing it instead of not a weird combination between the `bin_start` and `bin_end`
- For marks that are stacked with imputation, use bin_mid in (1)  encoding: midPoint of x and y (for dimension field) and (2) mark’s `sort` (edited)
- Decouple `facetby` from `groupby` so `facetby` fields are correctly both imputed and stacked

Fix https://github.com/vega/vega-lite/issues/2663


domoritz:
What does 

> Use bin_mid for marks that are stacked with imputatio
> * midPoint of x and y
> * sort field

mean?

kanitw:
@domoritz  Typo fixe

PR #2754, opened 2017-08-03
Support facet with nested column 

kanitw:
- If a facet's parent is a facet, calculate cardinality for column in facet instead of adding an extra data source
- Apply columns to group mark rather than layout operator as discussed in https://github.com/vega/vega/issues/952

Note: This PR only works for "facet of facet", but not "facet of concat of facet" -- However, I do not plan to support the latter in 2.0.  Feel free to submit a follow up PR if you think those are important.


examples/compiled/facet_column_facet_row_point.vg.json
```diff
@@ -67,9 +52,12 @@
             "row": 10,
             "column": 10
         },
+        "headerBand": {
+            "column": 0.5
```
kanitw: Maybe headerBand is bad for this one

kanitw:
Hmm -- I think I found a bug too!

kanitw:
This should be good now -- please review each commit separately

PR #2755, opened 2017-08-03
Upgrade to Vega 3.0.0-rc5

kanitw:


kanitw:
looks like we need to avoid applying stroke

PR #2766, opened 2017-08-04
Add version to homepage, add repeat and layer example, add notes on s…

domoritz:
…elections to homepage, upgrade vega datasets


kanitw:
build:examples?

domoritz:
Yeah, it took so long so I wanted to push first. Just amended but I still have this weird error where the avg changes. Can you build the examples for me?

PR #2770, opened 2017-08-04
[Bug] Add build-example for compiling only one example + Add Missing Example

kanitw:


kanitw:
hmm this is one is also buggy

kanitw:
@arvind, @domoritz -- I add this simple script for compiling one only one example, just in case you add/modify one example for the docs.

https://github.com/vega/vega-lite/commit/66395f22fb9567e8a66f0396e97630d0734be7fc

PR #2771, opened 2017-08-04
[WIP] My failed attempt to use build-example in build-examples

kanitw:
... give up for now ... not sure why -s doesn't get passed to `build-example`


domoritz:
What is `-s`?

kanitw:
It supposes to always compile SVG with that flag, but I just saw that #2770 has a bug for it

kanitw:
Let's not do this

PR #2779, opened 2017-08-05
Basic support for title

### LINKED ISSUE #431: Support adding title to a plot

kanitw:
Part of #431

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -29,6 +29,14 @@
       ],
       "type": "string"
     },
+    "Anchor": {
+      "enum": [
+        "start",
+        "middle",
+        "end"
+      ],
+      "type": "string"
+    },
     "AnyMark": {
       "anyOf": [
         {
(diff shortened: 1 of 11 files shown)


kanitw:
Hmm -- need to apply a few tricks -- according to https://github.com/vega/vega/issues/960

kanitw:
I think we can just wait for the fix in #960 and just merge this instead of inventing temporary hack.

One bug to fix is that width/height signal in this spec https://pastebin.com/NYJeSUgd should become child_width/child_height

kanitw:
Since this is the correct implementation, let's just hide it from the schema until we fix a relevant bug in Vega then.

domoritz:
Please add some tests so we cover the lines in the patch.

kanitw:
Those lines are unnecessary as we have duplicate codes -- https://github.com/vega/vega-lite/issues/2780

PR #2783, opened 2017-08-05
Example have changed on my machine too

kanitw:
This is weird, but let's take a look if the changes are all sensible.


kanitw:
@arvind @domoritz  -- I randomly look at the SVG diff and it seems like the SVG diff has improved.

For example, https://github.com/vega/vega-lite/pull/2783/files?short_path=41d56d1#diff-41d56d1fabeb9244e166d8b9e856b7ae is looking better.

However, it is still unclear to me why Travis is producing a different result.  (Maybe it still gets a different version from us.)

kanitw:
Also https://github.com/vega/vega-lite/pull/2783/files?short_path=1b5b2e4#diff-1b5b2e4c513d408ed5d64c5910cb05f2 is showing more labels like we expect

PR #2586, opened 2017-06-27
Consolidate parseLegend method 

kanitw:
This supposes to make the code shorter and more consistent but somehow leads to a mysterious error relaed to file importing

```
      ^
TypeError: Object prototype may only be an Object or null: undefined
    at setPrototypeOf (native)
    at Object.__extends (/Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/node_modules/tslib/tslib.js:53:9)
    at /Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/build/src/compile/layer.js:1:11626
    at Object.<anonymous> (/Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/src/compile/layer.ts:6:72)
    at Module._compile (module.js:570:32)
    at Module.replacementCompile (/Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/node_modules/nyc/node_modules/append-transform/index.js:58:13)
    at module.exports (/Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/node_modules/nyc/node_modules/default-require-extensions/js.js:8:9)
    at Object.<anonymous> (/Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/node_modules/nyc/node_modules/append-transform/index.js:62:4)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
...
```

PR #2785, opened 2017-08-05
Make BaseConcat to reduce redundant code + Fix Circular Dependencies

kanitw:
Note: We use type guards instead of `instanceof` as `instanceof` makes
  different parts of the compiler depend on the actual implementation of
  the model classes, which in turn depend on different parts of the compiler.
  Thus, `instanceof` leads to circular dependency problems.

  On the other hand, type guards only make different parts of the compiler
  depend on the type of the model classes, but not the actual implementation.


kanitw:
Removing `instanceof` also solve previous mysterious problem I have in https://github.com/vega/vega-lite/pull/2586

PR #2787, opened 2017-08-06
Update to rc6

kanitw:


yarn.lock
```diff
@@ -4885,6 +4984,33 @@ vega@^3.0.0-beta.31, vega@^3.0.0-rc5:
     vega-wordcloud "1"
     yargs "4"
 
+vega@^3.0.0-rc6:
```
kanitw: @domoritz  It's weird that yarn adds a new entry of Vega here but still leaves the old rc5 above!

domoritz: I think that is because veag-lite is a dep and by transitivity vega rc5.

kanitw:
I looked through all the SVG diff and they all look reasonable too.  So we should merge once we get Travis to work. 

However, Travis is still failing

(1) b17317d doesn't update selection test

(2) Travis seems to still get older version of Vega to compile SVGs.  (The diff is still outdated.)  And Travis UI seems to be dumb and does not know that its job has terminated when the log overflow due to SVG diff.

> "The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over)."

kanitw:
Pushed unit test fix -- let's see how it goes.

PR #2789, opened 2017-08-06
Multi-line Highlight Example

amitkaps:
Added an example of Multi Line Highlight using nearest selection.


domoritz:
Fantastic. Thank you!

Issue #2788, opened 2017-08-06
How to customise initial configuration for single selection?


amitkaps:
I created a mulit-line chart which highlights the nearest line, by changing the size of the line. See block here - https://bl.ocks.org/amitkaps/fe4238e716db53930b2f1a70d3401701

The chart works fine once the mouse moves and is over the chart view. However, when the chart first loads, the selection is treated as entire dataset and all the lines are highlighted. Is it possible to customise the behaviour, such that when the mouse is out of the view - 'mouseout', then the selection is treated as none?

I tried playing with event stream, to modify this like below, but this does not seem to work.
```
"on" : "`mouseover, !mouseout`"
```

Any suggestions on how to make this happen? And a broader question, on what the default behaviour of `mouseover` selection should be, when the mouse is not on the view?

domoritz:
Is this related to https://github.com/vega/vega-lite/issues/2696?

amitkaps:
Yes. Apologies, I had not seen that discussion thread, which addresses the design decision around `empty` vs `full` selection at the start. I had seen the documentation and the ability to use `and`, `or` and `not` for multiple selections, but it was not immediately apparent how to use it for single selection.

In case of single selection, the main example of [conditional encoding using selection](https://vega.github.io/vega-lite/docs/selection.html#conditional-encodings), should then be changed by using the `not` selection and reversing the values. It is not as intuitive (for a starter example), but makes more logical sense for this paintbrush highlight example and for single selections in general. 

```js
      "condition": {
        "selection": {"not": "paintbrush"},
        "value": 50
      },
      "value": 300
```
This principle simplifies the code for the multi-line highlight too. I have updated the same. Thanks @domoritz

amitkaps:
If you are ok with this logic, then I have sent a pull request to modify this example - https://github.com/vega/vega-lite/pull/2804

kanitw:
@domoritz  is this issue resolved?

domoritz:
@arvind is https://github.com/vega/vega-lite/pull/2804#issuecomment-321374998 addressed so that we can close this issue?

domoritz:
related: https://github.com/vega/vega-lite/issues/2321

pelotom:
Any update on how to do this without the `not` operator?

kanitw:
Looks like the example here https://vega.github.io/vega-lite/docs/selection.html#conditional-encodings already flipped the conditional logic and removed not?

pelotom:
[...]

PR #2804, opened 2017-08-08
Updated Selection Not and Paintbrush Simple example

amitkaps:
- Changed `selection.md` to explain the use of `not` operator for size encoding.
- Updated the paintbrush-simple example.

### DIFF
--- a/examples/specs/paintbrush_simple.vl.json
+++ b/examples/specs/paintbrush_simple.vl.json
@@ -10,9 +10,10 @@
     "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
     "size": {
       "condition": {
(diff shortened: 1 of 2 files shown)


arvind:
Thank you for the PR, @amitkaps. I'll soon have updates to the documentation to explain the initial state of selections to address your confusion in #2788, and tweak the paintbrush examples accordingly. I'd rather not use `not` in these examples as we do not introduce it as a valid operator until later in the documentation.

amitkaps:
Understood. It makes sense to keep the starter example simple and without 'not' operator.

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #2805, opened 2017-08-08
Automatically generate `parse: {field: 'date'}` for filtered field with timeUnit

### LINKED ISSUE #2756: Spec with a filter with timeUnit cannot be rendered
The following spec cannot be rendered.

```
{
  "data": {
    "url": "node_modules/vega-datasets/data/cars.json"
  },
  "transform": [
    {
      "filter": {
        "field": "Year",
        "timeUnit": "year",
        "range": [
          1969,
          1981
        ]
      }
    }
  ],
  "mark": "point",
  "encoding": {
    "y": {
      "field": "Origin",
      "type": "nominal"
    }
  },
  "config": {
    "overlay": {
      "line": true
    },
    "scale": {
      "useUnaggregatedDomain": true
    }
  }
}
```

kanitw:
Fix #2756

cc: @starry97

### DIFF
--- a/src/compile/data/formatparse.ts
+++ b/src/compile/data/formatparse.ts
@@ -1,9 +1,10 @@
 import {isCountingAggregateOp} from '../../aggregate';
 import {DateTime, isDateTime} from '../../datetime';
 import {FieldDef, isCount} from '../../fielddef';
-import {isEqualFilter, isOneOfFilter, isRangeFilter} from '../../filter';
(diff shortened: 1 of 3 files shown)


kanitw:
It renders fine now vega_editor.

domoritz:
I wonder whether this fix is incorrect. It adds the parse after the transforms, doesn't it?

PR #2807, opened 2017-08-08
Symlink data to `examples/` so all examples can be easily open in oth…

kanitw:
…er applications such as Vega Desktop and Voyager

### DIFF
--- a/scripts/build-examples.sh
+++ b/scripts/build-examples.sh
@@ -5,8 +5,8 @@ dir=${dir-"examples/compiled"}
 
 echo "compiling examples to $dir"
 
-rm -rf $dir
-mkdir $dir
+rm -f $dir/*.vg.json
+rm -f $dir/*.svg
 
 if type parallel >/dev/null 2>&1
 then
--- a/_config.yml
+++ b/_config.yml
@@ -17,7 +17,7 @@ baseurl: "/vega-lite" # the subpath of your site, e.g. /blog/
 url: "http://vega.github.io" # the base hostname & protocol for your site
 exclude: [bin, coverage, Gemfile, Gemfile.lock, lib/schema.json, node_modules,
   scripts, test, test-gallery, typings, package.json, bower.json, src, tsconfig.json,
-  tslint.json, nodemon.json, yarn.lock, CONTRIBUTING.md, README.md, LICENSE]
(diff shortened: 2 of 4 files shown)


domoritz:
Have you tried running the website? I just want to make sure that Jekyll doesn't choke on it.

kanitw:
Hmm weird, even though node_modules is in the exclude list, I still see this error:

```
Incremental build: disabled. Enable with --incremental
      Generating...
                    done in 8.154 seconds.
        ** ERROR: directory is already being watched! **

        Directory: /Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/examples/specs/data

        is already being watched through: /Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/node_modules/vega-datasets/data

        MORE INFO: https://github.com/guard/listen/wiki/Duplicate-directory-errors
 Auto-regeneration: enabled for '/Users/kanitw/Documents/_code/_idl/_visrec/vega-lite'
Configuration file: /Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/_config.yml
        ** ERROR: directory is already being watched! **

        Directory: /Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/node_modules/vega-datasets/data

        is already being watched through: /Users/kanitw/Documents/_code/_idl/_visrec/vega-lite/node_modules/vega-datasets/data
```

kanitw:
But the site is still running fine though

domoritz:
Do we have to manually create the symlink when we build the examples because the symlink gets deleted? Can you also move the normalized examples?

Other than that. LGTM.

kanitw:
Pushed update to script to remove generated specs, not the whole folder.

Move normalized spec is a separate sub-issue.  This PR and the squash commit should only deal with sym-linking.

scripts/build-examples.sh
```diff
@@ -5,8 +5,9 @@ dir=${dir-"examples/compiled"}
 
 echo "compiling examples to $dir"
 
-rm -rf $dir
-mkdir $dir
+mkdir -p $dir
```
domoritz: The mkdir is not necessary because we never delete the directory

kanitw: Good catch -- fixed!

Issue #2795, opened 2017-08-06
Avoid side effect in assembleData


kanitw:
Currently data component optimization happens in `assembleData()` and thus cause side effect in the model.component and requires that assembleData must be the first thing in assemble.  

I think we should have a separate optimizeData() method at the end of the parse.

kanitw:
This is at least causing problem in https://github.com/vega/vega-lite/issues/2736

domoritz:
So you would want assmebleData to come after some other assemble potentially?

kanitw:
It currently is not -- see https://github.com/vega/vega-lite/blob/master/src/compile/compile.ts#L93.

One potential quick fix is to call `const data = model.assembleData()` on the first line for `assembleTopLevelModel` and punt on refactoring this for now.

But in longer term, I think this refactor make sense since we do 
`return assembleTopLevelModel(model, getTopLevelProperties(inputSpec, config));`

as last line of `compile()`, which does not suggest that `assembleTopLevelModel` would have side effect on `model`.

kanitw:
FIxed in #2810

PR #2810, opened 2017-08-09
Split dataflow optimize from assemble.

domoritz:
Also refactored the code so we only call assemble data from the model. 

#2795

### DIFF
--- a/src/compile/compile.ts
+++ b/src/compile/compile.ts
@@ -5,12 +5,10 @@ import {Config, initConfig, stripAndRedirectConfig} from '../config';
 import * as log from '../log';
 import {normalize, TopLevel, TopLevelExtendedSpec} from '../spec';
 import {extractTopLevelProperties, TopLevelProperties} from '../toplevelprops';
-import {extend, keys} from '../util';
 import {buildModel} from './common';
-import {LayerModel} from './layer';
+import {assembleRootData} from './data/assemble';
+import {optimizeDataflow} from './data/optimize';
 import {Model} from './model';
-import {UnitModel} from './unit';
-
 
 export function compile(inputSpec: TopLevelExtendedSpec, logger?: log.LoggerInterface) {
   if (logger) {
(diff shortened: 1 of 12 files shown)


src/compile/model.ts
```diff
@@ -268,7 +270,14 @@ export abstract class Model {
   public abstract assembleSelectionSignals(): any[];
 
   public abstract assembleSelectionData(data: VgData[]): VgData[];
-  public abstract assembleData(): VgData[];
+  public assembleData(): VgData[] {
+     if (!this.parent) {
```
kanitw: If the rule is only assembleData in the root, why don't we just make sure that it is only called from `compile.ts`?

src/compile/model.ts
```diff
@@ -268,7 +270,14 @@ export abstract class Model {
...
+      // only assemble data in the root
+      optimizeDataflow(this.component.data);
```
kanitw: Hmm if we call it here, it may still happens after other things in compile.ts? 
Why don't we call it at the end of parse instead?

src/compile/model.ts
domoritz: Done.

src/compile/model.ts
domoritz: I moved it to compile.ts. I don't want to put it into parse because then I hate to check whether we are in the root again.

src/compile/compile.ts
```diff
@@ -67,6 +65,9 @@ function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config) {
 function assembleTopLevelModel(model: Model, topLevelProperties: TopLevelProperties) {
   // TODO: change type to become VgSpec
 
+  // optimize the datafow once
+  optimizeDataflow(model.component.data);
```
kanitw: Why here, not in the `compile` method above?  This feels like a separate phase between `parse` and `assemble`?

src/compile/compile.ts
```diff
@@ -86,7 +87,8 @@ function assembleTopLevelModel(model: Model, topLevelProperties: TopLevelPropert
     ...(style? {style} : {}),
     data: [].concat(
       model.assembleSelectionData([]),
-      model.assembleData()
+      // only assemble data in the root
+      assembleData(model.component.data)
```
kanitw: Yay.  I like that you move it here.  

Maybe rename to `assembleRootData` ? (So we have clear names for both `assembleRootData` and `assembleFacetData`)

src/compile/compile.ts
kanitw: (Not that this method returns value, so looking at it in `compile.ts` might still feel like `assembleTopLevelModel` shouldn't cause side-effect.

src/compile/compile.ts
domoritz: Done

src/compile/compile.ts
domoritz: Done

Issue #2418, opened 2017-05-24
Make Vega-Lite examples have proper relative paths


kanitw:
So that it works with Vega Desktop

- [ ] Update Gallery
- [ ] Update Vega-Editor

domoritz:
We actually correct the paths in the spec right now, which shows that there is a hack.

What's the corresponding vega issue?

kanitw:
> We actually correct the paths in the spec right now, which shows that there is a hack.

You mean baseUrl is a hack?

> What's the corresponding vega issue?

There is none yet.

domoritz:
> You mean baseUrl is a hack?

Even worse: https://github.com/vega/vega-lite/blob/master/site/static/main.ts#L38. We are not even using it. I don't understand how this even works for nested specs...

domoritz:
On second thought, I think the Vega desktop app should set the base URL. It's hard for us to use relative URLs because our specs appear on different pages at different nesting levels. See https://github.com/vega/vega-lite/pull/2722

Thoughts @kristw?

domoritz:
Reopen if you think we can use relative URLs and should use it instead of baseURLs.

kanitw:
How would the Vega Desktop app know what's the right base URL?

kanitw:
(It seems impossible to do that specifically for specs in Vega-Lite repos.)

Instead different pages that use different nested level should set correct base url.

domoritz:
The base URL is by default the URL of the app. For the Vega-Lite examples, it should be the base directory of the repo. 

I don't see how relative URLs would work.

kanitw:
> The base URL is by default the URL of the app.

But you cannot develop Voyager or Vega-Desktop to set the baseURL differently based on whether it is an example in the vega-lite repository.  I can't think of a reasonable way to ensure that. 

Suppose a spec in directory`/D` points to `data/xxx.json`,  Voyager/Vega-Desktop should better assume that the data is in `/D/data/xxx.json` rather than `~/Application/data/xxx.json` (For Mac).

domoritz:
I guess things work for the website and docs if we set the base URL for be the directory that has the specs.

domoritz:
Instead, we will
* Create a symlink to data in the example spec directory
* Move the normalized specs to `compiled`

kanitw:
@mattwchun  -- Would you have time to help submit a PR to move the normalized specs to `examples/compiled` instead of `examples/specs/normalized`?   Thanks.

PR #2811, opened 2017-08-09
Move the normalized specs to examples/compiled instead of examples/specs/normalized

mattwchun:
Refers to issue #2418. Move the normalized specs to examples/compiled instead of examples/specs/normalized


domoritz:
We're not removing vl specs in https://github.com/vega/vega-lite/blob/master/scripts/build-examples.sh#L8.

kanitw:
Thanks!

PR #2819, opened 2017-08-10
Three more examples

amitkaps:
- Scatter Nearest Rule
- Scatter Brush Rule
- Bar Brush Rule


domoritz:
Since I suggested the last one for the front page, we will add it to the official examples. The other two are great additions, though!

amitkaps:
@domoritz Makes sense. Feel free to pull more into official examples and delete from here. I am creating them on Vega datasets, so that we have a larger corpus of interactive examples for users to see the potential.

domoritz:
That's fantastic. I love these examples and they will be super useful to anyone getting started with Vega or Vega-Lite.

Issue #2763, opened 2017-08-04
Migrate to use style instead of role


kanitw:
Still need to produce a TODO based on discussion on slack...

kanitw:
- [ ] use style instead of role for marks
- [ ] make row-header/title use the right style
- [ ] cell style
- [ ] redirect mark-specific config to styles

kanitw:
- [ ] undo the docs here https://github.com/vega/vega-lite/pull/2670/files

kanitw:
- [ ] redirect `title`
- [ ] Make sure header labels output hardcoded orient, anchor, and offset

PR #2828, opened 2017-08-11
Make layout header uses the right style

kanitw:
Final part of #2763


kanitw:
Most of the SVG change should be mainly from `"black"` to `#000` that we use in Vega-parser

kanitw:
@arvind  -- Hmm, I regenerate in my machine but somehow it has no change.

test-runtime/translate.test.ts
```diff
@@ -126,7 +126,11 @@ import {
         it(`should work with shared scales in ${specType} views`, function() {
           for (let i = 0; i < hits[specType].length; i++) {
             embed(spec(specType, 0, {type, ...binding},
+<<<<<<< HEAD
```
domoritz: FIXME

Issue #2774, opened 2017-08-04
Facet default sort incorrect for binned facet


kanitw:
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {"url": "data/cars.json"},
  "facet": {"row": {"bin": {"maxbins": 4}, "field": "Displacement","type": "quantitative"}},
    "spec": {
      "mark": "point",
      "encoding": {
        "x": {"field": "Horsepower","type": "quantitative"},
        "y": {"field": "Acceleration","type": "quantitative"}
      }
    }
}
```

![vega_editor](https://user-images.githubusercontent.com/111269/28990470-9bef08e6-7932-11e7-9a5d-f0d4c28e282c.png)

kanitw:
Adding color shows that the only incorrect thing is labels:

![vega_editor](https://user-images.githubusercontent.com/111269/28991457-cef7f068-793c-11e7-9fc3-a95946ec9c62.png)

domoritz:
Aren't we adding the correct sort block?

```json
      "sort": {
        "field": "datum[\"bin_maxbins_4_Displacement\"]",
        "order": "ascending"
      },
```

yyyliu:
I also encountered this bug in Voyager ...

<img width="911" alt="screen shot 2017-08-10 at 12 40 50 pm" src="https://user-images.githubusercontent.com/25254084/29188890-797da146-7dc9-11e7-879e-21188c3d11b6.png">

kanitw:
@domoritz 

`bin_maxbins_4_Displacement` doesn't exist (we only have _range, _start, _end) and I think currently we only have range for the label, but we need to sort by `_start` (or `_end`) as range is string, not number.

domoritz:
I see. It is odd that we are sorting the cells after the range, not the start, though.

PR #2834, opened 2017-08-11
Move Facet's sort to FacetFieldDef, not header

### LINKED ISSUE #2829: header.sort should be a part of FacetFieldDef, not header.sort

kanitw:
to be consistent with FieldDef

Fix #2829


domoritz:
Does this also fix https://github.com/vega/vega-lite/issues/2774?

kanitw:
> Does this also fix #2774?

Nope -- this only moves things in the schema.

PR #2840, opened 2017-08-12
Rewrite Sort Docs

kanitw:


arvind:
Sorry but you could explain this change to me? Why are we removing FieldDef's `sort` in favor of reversing the scale? The latter always felt non-obvious to me. Does this contradict the change introduced in #2834 or, rather, why are FacetFieldDefs not sorted the same way?

kanitw:
Sure.  Given that this confuses you, maybe there is an opportunity to revise the design here.

Here I'm just updating the docs and examples to match previous changes.

In Vega-Lite v1, we were confused between reverse and sort descending.
We thought they are similar.  

However, these two different operations:

 - `sort` is applied to domain, which is only applicable for discrete scales since their order of the domains matter in the output visualizations.  For continuous scales, domain is simply the extent of the data, so ordering the domain does not affect the extent.    Note `sort` in VL becomes `scale.domain.sort` in Vega, which is only allowed for discrete scales.

- On the other hand, `reverse` is flipping the specified range so this is applicable for both discrete and continuous scales, but maybe less useful for discrete scale.

(For discrete scale, try applying sort and reverse to a color scale to see the difference.) 


Here are a few alternative designs:

a) make sort: "descending" works the same as `"reverse: true"` for quantitative scale, but I do not want to introduce multiple ways of doing the same thing.

b) Make `sort` works as is for discrete scale, but make sort: "descending" redirect to `reverse: true` in the underlying Vega continuous scale and remove `reverse` from VL scales.  Note that this would prevent users from flipping range of discrete scale.  The cons of this design are that (1) this no longer represents how scale works in Vega and (2) we can no longer reverse, but it may make it easier for end-user to flip continuous scale (by just using "sort").  I'm okay with this design, but if we go down this route, we might wanna make sure if there is a real world use case that users would need to `reverse` discrete scales.

kanitw:
Now that I revise the the docs more, the current design is pretty annoying. 

Basically in the `sort`'s docs, we have to note that this only works for discrete scales (See https://vega.github.io/vega-lite/docs/encoding.html#field-transform).  

I think either a) or b) is not bad.  Since we decouple `sort` to be a higher-level directive in Vega-Lite anyway, we do not really need to conform with D3 and Vega's underlying details.  
From end-user's perspective, reversing continuous scale is basically "sorting" it too. 

The question is whether we want to keep reverse.  In rare case, `reverse` allows users to use `category20`, starting from the last color in the palette.

Any thoughts? @domoritz @arvind

domoritz:
I like b)

kanitw:
Sort now supports continuous scales -- see https://github.com/vega/vega-lite/pull/2863.

PR #2841, opened 2017-08-12
Make points in connected scatterplot example filled

kanitw:


kanitw:
>  I wonder if this would be a more effective example if we styled it closer to the original (e.g. larger dimensions, axes styling/label formatting, etc.)

I think this is a good idea for a standalone example in the examples gallery.  Meanwhile, for using it in the documentation (for explaining `order` channel), maybe keeping it simpler is better? 

So I think, depending on whether customization makes the spec that much longer, we may modify it in the same file (or create two versions of it -- one for docs, one for example gallery?)

In any case, maybe we can merge this first, but please feel free to take a shot to improve it.

arvind:
Yeah, this makes sense as an example for the docs.

PR #2842, opened 2017-08-12
Update Bin Docs

kanitw:
Note that the bin transform is actually still broken, but that would be a separate issue.


site/docs/bin.md
```diff
@@ -53,26 +31,70 @@ The channel definition's `bin` property is for grouping quantitative, continuous
...
+You can directly bin an `encoding` field by using the `bin` property in a [field definition](encoding.html#field).
 
+{% include table.html props="bin" source="PositionFieldDef" %}
 
-#### Examples
+{:#histogram}
+### Example: Histogram
 
-Given a field with quantitative continuous data values
+<!-- Given a field with quantitative continuous data values
```
domoritz: Why do you remove this?

site/docs/bin.md
```diff
@@ -53,26 +31,70 @@ The channel definition's `bin` property is for grouping quantitative, continuous
...
+{% include table.html props="bin,field,as" source="BinTransform" %}
+
+### Example: Histogram with Bin Transform
+
+Instead of using the `bin` property of a field definition, you can also a bin transform
+to derive a new field (e.g., `bin_IMDB_Rating`), and encode the new field instead.
+
+**TODO: example**
+
+<!--
```
domoritz: I'm okay with having this here as a reminder that this is broken. I will fix the issue soon.

site/docs/bin.md
kanitw: I think it's making the page too long, but you can uncomment it.

domoritz: Then delete it ;-)

PR #2843, opened 2017-08-12
Added Unit Charts

amitkaps:
Added Unit Chart Examples 
- Rectangular
- Stacked
- Small Multiple


site/examples/gallery.md
```diff
@@ -23,5 +23,8 @@ Here we list great examples of Vega-Lite visualizations that were created by the
 * [Grouped Bar Chart by @churtado](https://bl.ocks.org/domoritz/f5abc519dd990bfcbc3f20f634658364)
 * [Multi Line Highlight by @amitkaps](https://bl.ocks.org/amitkaps/fe4238e716db53930b2f1a70d3401701)
 * [Slope graph by @g3o2](https://bl.ocks.org/g3o2/a6c539eacfb0b99eaf01e4f20b9f2897)
-* [Scatter Nearest Rule](https://bl.ocks.org/amitkaps/abfa7157d4366cc43cbbba55353d35d8)
-* [Scatter Brush Rule](https://bl.ocks.org/amitkaps/a484b94a7e1e0705c5ec865ba31f463c)
+* [Scatter Nearest Rule by @amitkaps](https://bl.ocks.org/amitkaps/abfa7157d4366cc43cbbba55353d35d8)
+* [Scatter Brush Rule by @amitkaps](https://bl.ocks.org/amitkaps/a484b94a7e1e0705c5ec865ba31f463c)
+* [Unit Chart Rectangular by @amitkaps](https://bl.ocks.org/amitkaps/d6648bd8ddb1c1e3706d7530126d1e2b)
+* [Unit Chart Stacked by @amitkaps](https://bl.ocks.org/amitkaps/cdc7dacd8f7d9f2a9cff4b10d3279b86)
+* [Unit Chart Small Multiple by @amitkaps](https://bl.ocks.org/amitkaps/67bd6dcb2af300a2b76f1e2351c1afdc)
```
domoritz: I wonder whether you could reuse the code from https://bl.ocks.org/amitkaps/d6648bd8ddb1c1e3706d7530126d1e2b and not add the offset manually if you used `facet`? Can you give it a try?

amitkaps: @domoritz The challenge when we create Unit Stacked Chart, is that I need to calculate the `y` position within that group. That would be the case even if I use `facet`. I would need access to a `window` function to do this - which is not there in Vega-Lite. See more details - https://github.com/vega/vega-lite/issues/2845#issuecomment-322007259

domoritz: Yes, you need to do that anyway for non-faceted unit chart, right?

amitkaps:
Yes, that is correct. I apologise, but I am not getting what you want me to try. Which offset are you referring to?

domoritz:
Never mind, I wasn't looking careful enough. You are already using the facet operator through the normalization.

PR #2847, opened 2017-08-12
Also group facet by bin start so we can sort marks correctly. Fixes #…

domoritz:
…2774


src/compile/data/facet.ts
```diff
@@ -30,13 +31,20 @@ export class FacetNode extends DataFlowNode {
     super();
 
     if (model.facet.column) {
-      this.columnField = model.field(COLUMN);
+      this.columnFields = [model.field(COLUMN)];
```
kanitw: Let's make `binSuffix: 'start'` explicit here.

src/compile/data/facet.ts
```diff
@@ -30,13 +31,20 @@ export class FacetNode extends DataFlowNode {
...
       this.columnName = model.getName('column_domain');
+      if (model.fieldDef(COLUMN).bin) {
+        this.columnFields.push(model.field(COLUMN, {binSuffix: 'range'}));
+      }
+
     }
 
     if (model.facet.row) {
-      this.rowField = model.field(ROW);
+      this.rowFields = [model.field(ROW)];
```
kanitw: Let's make `binSuffix: 'start'` explicit here.

src/compile/data/stack.ts
```diff
@@ -17,7 +17,7 @@ function getStackByFields(model: UnitModel): string[] {
 
     const scale = isScaleChannel(channel) ? model.getScaleComponent(channel) : undefined;
     const _field = field(fieldDef, {
-      binSuffix: scale && hasDiscreteDomain(scale.get('type')) ? 'range' : 'start'
+      binSuffix: 'start'
```
kanitw: Why is this change in "stack" related to facet sorting?

src/compile/model.ts
```diff
@@ -553,10 +553,9 @@ export abstract class ModelWithField extends Model {
   public field(channel: SingleDefChannel, opt: FieldRefOption = {}) {
     const fieldDef = this.fieldDef(channel);
 
-
-    if (fieldDef.bin) { // bin has default suffix that depends on scaleType
+    if (fieldDef.bin) {
       opt = extend({
-        binSuffix: this.hasDiscreteDomain(channel) ? 'range' : 'start'
+        binSuffix: 'start'
```
kanitw: Changing the default seems to affect a lot of place.  
Do we really never use `range` anymore?

kanitw:
Also, build examples?

src/compile/data/stack.ts
domoritz: It's not but I have no idea why we ever would use range instead of start. It's the same logic that I removed from the other function.

src/compile/model.ts
domoritz: We can explicitly call it but I don't see why you would ever want range unless it's used for labeling things.

src/compile/data/facet.ts
domoritz: It wasn't explicit before and it's not explicit anywhere else.

src/compile/model.ts
domoritz: I wonder whether this is relevant for setting the x scale to be ordinal but I cannot overrride it.

src/compile/data/facet.ts
kanitw: > it's not explicit anywhere else

This is definitely not true, I have been explicit in many places.

PR #2848, opened 2017-08-12
Restructure Encoding docs

kanitw:
The old encoding docs contains a huge that describes all group of encoding channels up front.  
It takes almost 2/3 of the page before it first mentions about Channel Def (Field/Value Def). 
Moreover, it incorrectly over-simplifies the types of channel def based on the channels, which will be also confusing for users (as they wouldn't distinguish their differences.)

Here I re-structure by having a short section upfront about basic of encoding channels and field/value def.  The rest of the page then describes different groups of encoding channels and the difference for their field/value def.  This allows us to have specialize table for each sub-type for channel defs. 

Please merge https://github.com/vega/vega-lite/pull/2849 first.


_data/link.yml
```diff
@@ -55,30 +55,46 @@ Type:
...
+  link: "aggregate.html"
+Axis:
+  link: axis.html
+Field:
+  link: field.html
+Legend:
+  link: legend.html
+Scale:
+  link: scale.html
+SortOrder:
```
kanitw: After applying `@nullable` to Sort, then it appears like this 
image

because our Jekyll engine doesn't handle `(SortOrder | SortField) |` Null well

site/docs/encoding.md
```diff
@@ -34,127 +50,177 @@ An integral part of the data visualization process is encoding data with visual
...
 
-### Mark Order Channels
+To map a constant value to an encoding channel, the channel's value definition must describe the `value` property.
+(See the [`value`](value.html) page for more examples.)
 
-`order` channel sorts the layer order or stacking order (for stacked charts) of the marks while `path` channel sorts the order of data points in line marks.
+{% include table.html props="value" source="NumberValueDef" %}
 
-{% include table.html props="order" source="Encoding" %}
+TODO: replace type with `Any` (as the type of `value` depends on channel.)
```
kanitw: @domoritz  Do we have a way to replace Number as Any just for `NumberValueDef`?

Alternatively, we can make this single table manually.

kanitw: I think I'll get rid of Generic ValueDef as it explodes the number of types too much.  This is just plain too confusing even for me to generate a docs.

kanitw:
@arvind and @domoritz  -- okay I spent the whole day one this PR -- Please feel free to make another pass.   At least it should be way less confusing than before.

_data/link.yml
domoritz: I'll fix it ina  follow up

PR #2849, opened 2017-08-12
Eliminate Generic property for ValueDef

kanitw:
... as it over-explodes the complexity of our JSON schema.
(It's even confuses me when I try to access them for the docs, I'm sure it will confuse other people.)


kanitw:
Note that this tiny of information costs 200 lines in the JSON schema.  If we really wanna check it, better do type validation in the compiler.

Issue #2839, opened 2017-08-12
Bin Transform broken


kanitw:
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "url": "data/movies.json"
  },
  "transform": [
    {
      "bin": true,
      "field": "IMDB_Rating",
      "as": "bin_IMDB_Rating"
    }
  ],
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "bin_IMDB_Rating",
      "type": "quantitative"
    },
    "y": {
      "aggregate": "count",
      "type": "quantitative"
    }
  }
}
```

kanitw:
- [ ] please add this back to bin.md once it is ready

kanitw:
Relevant cases to check for pre-binned field 
- [ ] Would it work correctly with stack?
- [ ] Would it work correctly with facet?

domoritz:
Problem is that `bin_IMDB_Rating` is not a field after the binning.

domoritz:
Hmm, I wonder how this ever worked. How do we track that a field is binned?

kanitw:
>  How do we track that a field is binned?

I don't remember seeing any of that, but that's exactly missing.

kanitw:
- [ ] Would it work with selection projection?

kanitw:
One temporary solution is remove `_start` suffix for bin_start.  This means that at very least, users will be able to use the bin_start value. 

I don't think I'm super happy with this but at least this will render some reasonable result (In fact, this is exactly what Tableau does.) 

Even if we have a bin component or something to track if a field is binned, I think we should still do this anyway, just in case some component isn't perfectly wired up with bin, we still always show some result.

domoritz:
> please add this back to bin.md once it is ready

Where was it removed?

domoritz:
Until we have a cleaner solution, ordinal gives better results than quantitative (but hides empty bins!). 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "url": "data/movies.json"
  },
  "transform": [
    {
      "bin": true,
      "field": "IMDB_Rating",
      "as": "bin_IMDB_Rating"
    }
  ],
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "bin_IMDB_Rating",
      "type": "ordinal"
    },
    "y": {
      "aggregate": "count",
      "type": "quantitative"
    }
  }
}
```

[...]

PR #2860, opened 2017-08-13
Eliminate `_start` suffix for bin's start value

kanitw:
This will alleviate #2839

### DIFF
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -553,10 +553,8 @@ export abstract class ModelWithField extends Model {
   public field(channel: SingleDefChannel, opt: FieldRefOption = {}) {
     const fieldDef = this.fieldDef(channel);
 
-    if (fieldDef.bin) {
-      opt = extend({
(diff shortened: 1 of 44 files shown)


src/compile/model.ts
```diff
@@ -553,10 +553,8 @@ export abstract class ModelWithField extends Model {
   public field(channel: SingleDefChannel, opt: FieldRefOption = {}) {
     const fieldDef = this.fieldDef(channel);
 
-    if (fieldDef.bin) {
-      opt = extend({
-        binSuffix: 'start'
-      }, opt);
+    if (!fieldDef) {
+      return undefined;
```
domoritz: Why return undefined and not `field(fieldDef, opt)`?

kanitw: There is no fieldDef here. Calling the function would lead to undefined error.

domoritz: Ups, didn't read the code... you're right

PR #2863, opened 2017-08-13
Make `sort` work with continuous scale + Revise Sort Docs

kanitw:


build/vega-lite-schema.json
```diff
@@ -1194,7 +1194,7 @@
               "type": "null"
             }
           ],
-          "description": "Sort order for a field with discrete domain.\nThis can be `\"ascending\"`, `\"descending\"`, `null`, or a [sort field definition object](sort.html#sort-field) for sorting the field by another field.\n\n__Default value:__ `\"ascending\"`\n\n__Note:__ For fields with continuous domain, please use `\"scale\": {\"reverse\": true}` to flip the scale instead."
+          "description": "Sort order for the encoded field.\nSuppoted `sort` values include `\"ascending\"`, `\"descending\"` and `null` (no sorting).\nFor fields with discrete domains, `sort` can also be a [sort field definition object](sort.html#sort-field) for sorting the field by another field.\n\n__Default value:__ `\"ascending\"`"
```
domoritz: include -> are

domoritz: for sorting the field by another field -> for sorting by an aggregate.

site/docs/aggregate.md
```diff
@@ -69,7 +69,8 @@ Otherwise, if none of the specified encoding channel contains `aggregate`, the r
 
 <span class="note-line"><sup>1</sup>The group-by fields are also known as [independent/condition variables](https://en.wikipedia.org/wiki/Dependent_and_independent_variables) in statistics and [dimensions](https://en.wikipedia.org/wiki/Dimension_(data_warehouse)) in Business Intelligence. Similarly, the aggregate fields are known as [dependent variables](https://en.wikipedia.org/wiki/Dependent_and_independent_variables) and [measures](https://en.wikipedia.org/wiki/Measure_(data_warehouse)). </span>
 
-#### Supported Aggregation Operations
+{:#ops}
```
domoritz: You probably need to fix the link file.

build/vega-lite-schema.json
kanitw: include is more precise ... (as the latter sentence describe other ones)

for sorting the field by another field -- the goal is to sort by another field, the fact that its requires aggregate is minor.

src/scale.ts
```diff
@@ -382,9 +382,13 @@ export interface Scale {
    */
   domain?: Domain;
 
+
+  // Hide because we might not really need this.
```
domoritz: Why? Just delete it if it is not needed.

build/vega-lite-schema.json
domoritz: but you are not sorting the field. So how about sorting the domain? Or the data?

site/docs/aggregate.md
kanitw: Force pushed

src/scale.ts
kanitw: As said on slack, it's too much work to remove it.  Plus, we may need it in the future. 
Totally not worth the time now.

build/vega-lite-schema.json
kanitw: To be precise, it would be "sorting the values of the field by values of another field", but it feels more awkward?

domoritz: Just read the sentence as it is right now. It repeats field like 5 times.

kanitw: Feel free to improve it.  I need more help with the docs anyway...

src/scale.ts
domoritz: So If I understand it correctly, you set the reverse property but don't want to remove it completely because then you would have to store it somewhere else.

build/vega-lite-schema.json
domoritz: Let me spend cycles on other docs rather than context switch then :-P

src/scale.ts
kanitw: Not just that.  Thinking more I think I prefer a) but don't want to make a call for it right now. 
By hiding it, we can make this decision in the future if we need to.

build/vega-lite-schema.json
kanitw: Ok I made it shorter.  Let's just merge this and focus on other issues..

Issue #2773, opened 2017-08-04
Enforce binned field to be quantitative type


kanitw:

kanitw:
Otherwise facet label can be wrong:

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {"url": "data/cars.json"},
  "facet": {"row": {"bin": {"maxbins": 4}, "field": "Displacement","type": "ordinal"}},
    "spec": {
      "mark": "point",
      "encoding": {
        "x": {"field": "Horsepower","type": "quantitative"},
        "y": {"field": "Acceleration","type": "quantitative"}
      }
    }
}
```

![vega_editor](https://user-images.githubusercontent.com/111269/28990443-6c1c59e8-7932-11e7-9c08-e1a7884cd825.png)

domoritz:
I think we can make type optional here.

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/2784

PR #2864, opened 2017-08-13
Support ordinal bin.

domoritz:
Before:
screen shot 2017-08-12 at 23 44 07
screen shot 2017-08-13 at 11 29 55

After:
screen shot 2017-08-13 at 11 27 56
screen shot 2017-08-12 at 23 53 51
screen shot 2017-08-13 at 11 28 50

Reverts https://github.com/vega/vega-lite/issues/2773


domoritz:
Let me try to tweak the second example a bit.

domoritz:
Ha, and now

screen shot 2017-08-13 at 11 47 43

src/compile/data/bin.ts
```diff
@@ -24,7 +24,7 @@ function rangeFormula(model: ModelWithField, fieldDef: FieldDef<string>, channel
 
       return {
         formulaAs: field(fieldDef, {binSuffix: 'range'}),
-        formula: `${numberFormatExpr(startField, guide.format, config)} + " - " + ${numberFormatExpr(endField, guide.format, config)}`
+        formula: `${startField} === null ? 'null' : ${numberFormatExpr(startField, guide.format, config)} + " - " + ${numberFormatExpr(endField, guide.format, config)}`
```
kanitw: What if NaN and undefined?   Better use the same expression we use for filtering invalid value?

kanitw:
I don't understand why [this commit](https://github.com/vega/vega-lite/pull/2864/commits/ebde6e7efe10f48cf2840dac5c6d6d2b50d1c5a6) help 
"support ordinal binned facet."  -- Updating test (and examples, if it is not too slow) within the commits rather than having "fix tests and update examples" commits at the end for the whole PR would help better.

src/compile/axis/rules.ts
```diff
@@ -78,11 +79,10 @@ export function orient(channel: Channel) {
...
 export function tickCount(channel: Channel, fieldDef: FieldDef<string>, scaleType: ScaleType, size: VgSignalRef) {
 
-  if (!fieldDef.bin && !hasDiscreteDomain(scaleType) && scaleType !== 'log') {
-    // Vega's default tickCount often lead to a lot of label occlusion on X without 90 degree rotation
-    // Thus, we set it to 5 for width = 200
-    // and set the same value for y for consistency.
-
+  if (!hasDiscreteDomain(scaleType) && scaleType !== 'log') {
+    if (fieldDef.bin) {
+      return {signal: `min(ceil(${size.signal}/40), ${(fieldDef.bin as Bin).maxbins})`};
```
kanitw: Why removing comments?  Instead, please add comments saying why it should be min of the two?

kanitw:
> I don't understand why [this commit](https://github.com/vega/vega-lite/pull/2864/commits/ebde6e7efe10f48cf2840dac5c6d6d2b50d1c5a6) help 
"support ordinal binned facet."  -- Updating test (and examples, if it is not too slow) within the commits rather than having "fix tests and update examples" commits at the end for the whole PR would help better.  


NVM.  I think I get it now.

src/compile/data/bin.ts
domoritz: Done

src/compile/axis/rules.ts
domoritz: Done.

kanitw:
1) Note that this PR introduces two things:
- Support binning with discrete position scales (band/point)
- Support binning with `type: ordinal`.

For the latter, I remember we discuss whether type should mean type before or after transforming, but I can't remember what's our conclusive agreement.  In this case, it would be "type after transforming".  I guess that's fine since `aggregate: "distinct"` requires `type: "quantitatitive"` too.  
I wonder if we have any case where type is type before transforming though.  If not, we are in good shape.  

2)Another question -- This PR seems to now allow `type: nominal` for bin.  Should we prevent/allow that? (or at least throw warning.

domoritz:
> Another question -- This PR seems to now allow type: nominal for bin. Should we prevent/allow that? (or at least throw warning.

I think nominal is totally fine to do. Updated the pr. 

screen shot 2017-08-13 at 14 02 31

domoritz:
idk, I might have a vega bug here but will need to investigate

```
{
      "scale": "y",
      "labelOverlap": true,
      "orient": "left",
      "tickCount": {"signal": "ceil(height/40)"},
      "title": "Number of Records",
      "zindex": 1
    },
...
 "legends": [
    {
      "fill": "color",
      "title": "BIN(IMDB_Rating)",
      "encode": {
        "symbols": {"update": {"shape": {"value": "square"}}}
      }
    }
  ],
```

domoritz:
Confirmed as vega bug (or we are using it wrong)

screen shot 2017-08-13 at 14 38 15

PR #2865, opened 2017-08-13
Updated documentation on stack

amitkaps:
Defining `stack` is now within encoding channel in v2 vs `config.mark.stack` in v1. Edits in documentation to reflect that.


kanitw:
Thanks!

PR #2869, opened 2017-08-13
"Doc Overview" + Minor Mark Revision

kanitw:
(I copy the style for "Doc Overview" from Vega --- it's better than having floating TOC.


site/examples/gallery.md
```diff
@@ -8,6 +8,7 @@ permalink: /examples/
 This page shows example specifications for different types of graphics.
 To see example code for embedding visualization in a webpage, please refer to [this page](../usage/embed.html).
 
+
```
domoritz: why a new line here

kanitw: no reason, but it doesn't change thing... so let's ignore it

PR #2876, opened 2017-08-14
Support Title with constraints for multi-views (anchor = start) only

kanitw:


src/compile/model.ts
```diff
@@ -170,7 +171,7 @@ export abstract class Model {
 
     // If name is not provided, always use parent's givenName to avoid name conflicts.
     this.name = spec.name || parentGivenName;
-    this.title = spec.title;
+    this.title = isString(spec.title) ? {text: spec.title} : spec.title;
```
willium: Is there ever a case where we would want to allow a numerical title?

src/title.ts
```diff
@@ -16,10 +21,17 @@ export interface TitleBase {
...
+   *
+   * __Default value:__ `"group-title"`.
+   */
+  style?: string | string[];
+
   // TODO: name, encode, interactive, zindex
 }
 
-export interface Title extends TitleBase {
+export interface TitleParams extends TitleBase {
```
willium: Why the rename? It doesn't seem to semantically fit as well, given that we don't use "Params" in other, similar cases

src/compile/model.ts
arvind: My guess is that this test isn't for numeric titles but rather a `spec.title` that can be either a [`string | TitleParams`](https://github.com/vega/vega-lite/pull/2876/files#diff-f0bc433010488ea7f9df55b35517b91bR39)

kanitw: I think it doesn't matter.  Feel free to change if you strongly think so.

src/title.ts
kanitw: We use `bin: boolean | BinParams` so that in the documentation it shows up as Bin is either a boolean or a BinParams object (not boolean or Bin, because boolean is Bin too (Bin = boolean | BinParams).

So I'm doing the same thing here.  (Title = string | TitleParams) but we don't define Title explicitly because we want both types to show up in the table in the docs.

willium: That makes sense.

kanitw:
Thanks for the comment.  Rebased and added example. 

Note that we still commit SVG changes to Github, we just stop asking Travis to check if the SVGs are up to date (as it will always say that we're out of date.)

PR #2889, opened 2017-08-17
Get rid of `model.assembleScales()`

kanitw:
so we have less methods in models + have all assembleScale tests in one place

(We couldn't do this earlier due to circular dependencies problem.)


kanitw:
@willium  Thanks for reviewing.  FYI, next time don't forget to delete a branch when you merge a PR :)

Issue #1487, opened 2016-08-03
More concise way to specify time unit


kanitw:
## 1) Day of month / week

Currently we follow Javascript:
- `date` means day of month
- `day` means day of week. 

But this also means that  YYYY-MM-DD = `yearmonthdate`.   (Currently, we only officially support `yearmonthdate` and  automatically converts `yearmonthday` to `yearmonthdate`.  So `yearmonthday` won't pass schema validation but will still behave okay with warning.) 

However, `yearmonthday` is more natural.  Date usually means all of year, month, and day -- not just the day.  For example, in [Wikipedia](https://en.wikipedia.org/wiki/Calendar_date#Gregorian.2C_year-day-month_.28YDM.29),  Year-Month-Day (YMD) is a date format. 

Plus, it's unnatural for typical Vega-Lite (except JS ninjas) that day of week = `day` and day of month = `date` as typically people would think `day` as day of month.   I can't remember how many times I type `yearmonthday` instead of `yearmonthdate` although I have wrote a part of this logic.  
### Proposal

Instead, I propose that we follow how [datalib's `dl.bins.date`](https://github.com/vega/datalib/wiki/Statistics#dl_bins_date) distinguishes between these two  and say that 
- `weekday` = day of week
- `day` = day of month

We can keep `date` similar to `day` for backward compatibility.  This way it would be less confusing. 
Plus, the only breaking change is the old `day` has to become `weekday`.  Other than that, everything behaves the same, but `yearmonthday` becomes the official one instead of `yearmonthdate` (but we'll keep supporting the latter for backward compatibility.) 

I don't think I'll make this change until we release 2.0 but I think we better remove this confusion!
## 2) Discrepancies between VL / DataLib

This is less important than 1) but worth considering.  

[Datalib's `dl.bins.date`](https://github.com/vega/datalib/wiki/Statistics#dl_bins_date) uses slight different scheme for time units. 

2.2) For periodic / single-part time Unit, in Vega-Lite we use singular for `month` / `day` but plural for `hours`, `minutes`, `seconds`, and `milliseconds` while Datalib uses plurals for all of them and have no `milliseconds`. 

| VL | DL |
| --- | --- |
| month / day | months / days |
| hours / minutes / seconds | hours / minutes / seconds |
| date* | dates |
| day* | weekdays |
| milliseconds | - |
| hoursminutes, hoursminutesseconds, minutes seconds, secondsmilliseconds | - |

[...]

PR #2893, opened 2017-08-17
Minor revision for timeUnit

kanitw:
Still more to be done, but I'll Dom take it from here.


site/docs/timeunit.md
```diff
@@ -51,34 +51,34 @@ A `timeUnit` transform in the `transform` array has the following properties:
...
 
-- Non-periodic Time Unit: `'year'`, `'yearmonth'`, `'yearmonthday'`, `'yearmonthdate'`, `'yearday'`, `'yeardate'`, `'yearmonthdayhours'`, `'yearmonthdayhoursminutes'`, `'utcyear'`, `'utcyearmonth'`, `'utcyearmonthday'`, `'utcyearmonthdate'`, `'utcyearday'`, `'utcyeardate'`, `'utcyearmonthdayhours'`, `'utcyearmonthdayhoursminutes'`.
+- Non-periodic Time Unit: `'year'`, `'yearmonth'`, `'yearmonthday'`, `'yearmonthdate'`, `'yearday'`, `'yeardate'`, `'yearmonthdayhours'`, `'yearmonthdayhoursminutes'`
   - For example the `yearmonth` of `April 4, 2016 11:52:34` is `April 2016`.
 
-- Periodic Time Unit: `'month'`, `'day'`, `'date'`, `'hours'`, `'minutes'`, `'seconds'`, `'milliseconds'`,  `'hoursminutes'`, `'hoursminutesseconds'`, `'minutesseconds'`, `'secondsmilliseconds'`, `'utcmonth'`, `'utcday'`, `'utcdate'`, `'utchours'`, `'utcminutes'`, `'utcseconds'`, `'utcmilliseconds'`,  `'utchoursminutes'`, `'utchoursminutesseconds'`, `'utcminutesseconds'`, `'utcsecondsmilliseconds'`.
+- Periodic Time Unit: `'month'`, `'day'`, `'date'`, `'hours'`, `'minutes'`, `'seconds'`, `'milliseconds'`,  `'hoursminutes'`, `'hoursminutesseconds'`, `'minutesseconds'`, `'secondsmilliseconds'`,
```
arvind: Can we explain what we mean by "periodic" and "non-periodic." Every time I see this, it always takes me a few seconds to figure out the difference. Perhaps there's a better classification?

kanitw: I agree.  It's pretty confusing.   Here are what we mean:

Non-periodic only means that if you have "year", then the output time is basically rounded value of the actual time.  

Periodic means that we only extract parts of the time without year, which makes the time circular/periodic. (Jan-Feb-...-Dec-Jan-...)

Wonder if you have any suggestion for better classification  (or should we just get rid of the classification?)

arvind: My vote would be to get rid of it. I think the `timeUnit` values by themselves are sufficiently descriptive that the periodic/non-periodic classification hampers understandability.

site/docs/timeunit.md
```diff
@@ -51,34 +51,34 @@ A `timeUnit` transform in the `transform` array has the following properties:
...
 
+By default, all time units represent date time using local time.  To use UTC time,
+you can add `utc` prefix (e.g., `utcyearmonth`).
```
domoritz: add *the* `utc` prefix, I think

kanitw:
I get rid of the table / classification -- I just list all of them, grouped by prefixes.

Looking at this table, I don't understand why hours, minutes, seconds need "s" but month and day do not.  I wonder if we should just get rid of "s" for all of them.  (Our logic would still work without s if we remove "s" from the keywords.)

arvind:
My vote would be to keep the singular/plural as is. They follow what I would expect colloquially, mirror the [JS Date API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) and, were we to change them, introduces an unnecessary breaking change.

site/docs/timeunit.md
```diff
@@ -49,37 +51,23 @@ A `timeUnit` transform in the `transform` array has the following properties:
...
-| `utchours`       | returns the hours component for a given date input, in utc time (number between `0` and `23`). | `11` |
-| `utcminutes`     | returns the minutes component for a given date input, in utc time (number between `0` and `59`). | `52` |
-| `utcseconds`     | returns the seconds component for a given date input, in utc time (number between `0` and `59`). | `34` |
-| `utcmilliseconds`| returns the milliseconds component for a given date input, in utc time  (number between `0` and `999`). | `201` |
+Vega-Lite supports the following time units:
+
+- `"year"`, `"yearquarter"`, `"yearquartermonth"`, `"yearmonth"`, `"yearmonthdate"`, `"yearmonthdatehours"`, `"yearmonthdatehoursminutes"`, , `"yearmonthdatehoursseconds"`.
+- `"quarter"`, `"quartermonth"`
+- `"month"`, `"monthdate"`
+- `"date"` (Day of month, i.e., January - December)
```
arvind: I think it would be clearer to say `(Day of month, i.e., 1-31)`

kanitw: lol right

arvind:
Also, see our prior discussion on pluralization: #1487, #78. [The "periodic time units" are the plurals.](https://github.com/vega/datalib/issues/78#issuecomment-237130277)

arvind:
Manually merged as 23f0ca3 to fix a typo.

Issue #2890, opened 2017-08-17
Add CSS to hide the first bulltet of a TOC


kanitw:
Otherwise we will get one bullet for "Documentation Overview"

PR #2894, opened 2017-08-17
Hide first item of TOC as it will always be Documentation Overview

kanitw:
Fix https://github.com/vega/vega-lite/issues/2890

(Let's agree on always adding "# Documentation Overview" so we don't have floating TOC)

### DIFF
--- a/site/static/main.css
+++ b/site/static/main.css
@@ -647,3 +647,8 @@ footer .edit-page {
     margin-top: 0;
   }
 }
+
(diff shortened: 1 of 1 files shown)


arvind:
Which page is this targeting? A quick glance at the main `/docs` where I see a top-level `Overview` bullet doesn't have a `ul#markdown-toc`.

kanitw:
All of these https://github.com/vega/vega-lite/pull/2869

arvind:
Thanks!

PR #2895, opened 2017-08-17
Separate transforms into individual files

willium:
Per https://github.com/vega/vega-lite/pull/2734#discussion_r133618235


kanitw:
Separate the test files too i think

willium:
They are separate... will fix tests

willium:
Tests should be passing now.

kanitw:
Thanks a lot :)

PR #2903, opened 2017-08-20
Rewrite scale docs

kanitw:


kanitw:
This should be ready for review now.  (After sinking ~ 15-20 hrs into it.) 

Since this is pretty much a complete re-write, checking out the branch and edit locally might be easier.

examples/compiled/rect_heatmap.vg.json
```diff
@@ -179,11 +179,6 @@
         },
         "axisY": {
             "minExtent": 30
-        },
```
domoritz: greenblue works much better for this example because 0 is white

kanitw: Sure, but I need an example to demonstrate what is our default ramp.  So you can either create a new example or tell Jeff to change the default.

domoritz: The default make sense for everything but the heatmap :-P I will change the example.

domoritz: What is the difference between ramp and heatmap? Is one for marks that fill the space and one for marks that may overlap and leave holes? If so, how is blues better than viridis?

kanitw: As mentioned on slack, I think the defaults are okay although I'm not a super big fan. 

The point is that I'm using the example to explain default color and thus it should not be changed.

PR #2921, opened 2017-08-25
Delete stacked_area_binned.vl.json

domoritz:
I can't tell why this example is binned.


kanitw:
If you mean it is not semantically sensible, I agree.  

Maybe it was created just to serve as a test case?  If so, maybe re-create a semantically sensible one with other datasets instead of removing?

domoritz:
We already have stacked area chart tests.

Issue #2878, opened 2017-08-14
Support timeUnit with ordinal 


kanitw:
Since you want to support bin with type: ordinal, we should support `timeUnit` with `type: ordinal` as well for consistency.  

- [x] In such case, we should use band/point/ordinal scale for the field.
- [x] Make formatParse use `timeUnit` as cue, in addition to type: temporal
- [x] whenever we check `type === TEMPORAL` in the rule base, we need to check if it has timeUnit too.   (We probably should just introduce a new `isTime(fieldDef: FieldDef)` method 

- [x] revise docs

kanitw:
Here is another conflicting design:

Right now for `'hours', 'day', 'month', 'quarter'`, we use band/point scale even though the field definition is  `{timeUnit: ..., field: ..., type: 'temporal'}`.  

![image](https://user-images.githubusercontent.com/111269/29294449-3e9d5b32-8104-11e7-9061-3d52a597fa27.png)

If we say `"type"` is post-transform, shouldn't we always use time scale instead?

We can always show warning if users use bar/tick marks with `timeUnit` + `type: "temporal"` to tell them to use `type: ordinal` instead.

kanitw:
- [ ] In such case, we have to change the rule for `tickCount` for `timeUnit`.

kanitw:
FYI, in the scale docs I'm re-writing, I already assume that we use always time scale for `temporal` fields and already get rid of `isDiscreteByDefault`.    Thus, we can derive a nice table like this:

![image](https://user-images.githubusercontent.com/111269/29483845-e2ba72b8-8465-11e7-8c95-2228694b710d.png)

Bring it back will make it too complex to be explained.

PR #2922, opened 2017-08-25
Make type: temporal always use time scale by default

kanitw:
Fix https://github.com/vega/vega-lite/issues/2878

The ticks for month are a little less nice, but this is definitely the right semantic since temporal supposes to be continuous, not ordinal. 

We should only use ordinal when it is explicitly "casted". 

The ticks can be improved in a later.  

(This is also what Tableau does by default.)


test-runtime/interval.test.ts
```diff
@@ -115,7 +115,7 @@ describe('interval selections at runtime in unit views', function() {
 
     embed(spec('unit', 1, {type, encodings: ['x']}, {values, x: {type: 'temporal', timeUnit: 'day'}}));
 
-    extents = [[1136188800000, 1136275200000, 1136361600000], [1136448000000, 1136534400000]];
+    extents = [[1136188800000, 1136361600000], [1136448000000, 1136534400000]];
```
kanitw: @arvind -- I have to make change to this line to make it work -- but not sure if it is really correct.  Could you please help double check?

test-runtime/interval.test.ts
```diff
@@ -115,7 +115,7 @@ describe('interval selections at runtime in unit views', function() {
 
     embed(spec('unit', 1, {type, encodings: ['x']}, {values, x: {type: 'temporal', timeUnit: 'day'}}));
 
-    extents = [[1136188800000, 1136275200000, 1136361600000], [1136448000000, 1136534400000]];
+    extents = [[1136187936000, 1136361600000], [1136447136000, 1136535264000]];
```
kanitw: @arvind -- I have to make change to this line to make it pass the runtime test -- but not sure if it is really correct. Could you please help double check?

(I also don't understand why the old extent should have 3 members.)

kanitw:
@domoritz  Lint fixed.  I don't know why saving files doesn't automatically fix them.

You might wanna check other occurrences of fielddef.type to make sure you agree with the ones I still leave.

test-runtime/interval.test.ts
arvind: LGTM.

test-runtime/interval.test.ts
arvind: LGTM. The old extents had 3 members because I think ordinal-point scales were being generated by timeUnits? Thus, inverting brush extents over a range returned the constituent points.

domoritz:
@kanitw We don't fix unused vars because of some tslint issues. I don't know why it complains on travis now, though.

kanitw:
Well that tslint issue is gone so i turn it on. (And you saw it.)

But i dont know why sometimes vscode doesnt automatically fix them in vega lite but in voyager it always fixes for me

domoritz:
Which tslint version are you using? I couldn't upgrade it in VL because of some issues with our enums.

PR #2923, opened 2017-08-25
Revise bin ranges and bin scales

domoritz:


src/compile/data/formatparse.ts
```diff
@@ -68,7 +68,7 @@ export class ParseNode extends DataFlowNode {
       model.forEachFieldDef(fieldDef => {
         if (fieldDef.type === TEMPORAL) {
           parse[fieldDef.field] = 'date';
-        } else if (fieldDef.type === QUANTITATIVE) {
+        } else if (fieldDef.type === QUANTITATIVE || fieldDef.bin) {
```
domoritz: https://github.com/vega/vega-lite/pull/2922 has a better solution for this but the test is useful.

src/compile/data/aggregate.ts
```diff
@@ -1,23 +1,20 @@
...
-import {isScaleChannel} from '../../channel';
+import {Channel, isScaleChannel} from '../../channel';
 import {field, FieldDef} from '../../fielddef';
 import * as log from '../../log';
 import {SummarizeTransform} from '../../transform';
-import {NOMINAL, ORDINAL} from '../../type';
 import {Dict, differ, duplicate, extend, keys, StringSet} from '../../util';
 import {VgAggregateTransform} from '../../vega.schema';
 import {UnitModel} from './../unit';
+import {binRequiresRange} from './bin';
```
kanitw: It's weird that aggregate node depends on bin node, should binRequiresRange be somewhere else such as common?

src/compile/data/bin.ts
```diff
@@ -1,18 +1,32 @@
...
 import {isUnitModel, Model, ModelWithField} from '../model';
 import {DataFlowNode} from './dataflow';
 
-function rangeFormula(model: ModelWithField, fieldDef: FieldDef<string>, channel: Channel, config: Config) {
-    const discreteDomain = model.hasDiscreteDomain(channel);
+/**
+ * Checks whether a fieldDef for a particular channel requires a computed bin range.
+ */
+export function binRequiresRange(fieldDef: FieldDef<string>, channel: Channel) {
+  if (!fieldDef.bin) {
```
kanitw: Wonder if this is necessary -- we could also just take type as input.

src/compile/data/bin.ts
```diff
@@ -1,18 +1,32 @@
...
+    console.warn('Only use this method with binned field defs');
+    return false;
+  }
+  if (!contains(NONSPATIAL_SCALE_CHANNELS, channel) && !contains(SPATIAL_SCALE_CHANNELS, channel)) {
```
kanitw: We have `NONSPATIAL_SCALE_INDEX` in channel.ts, maybe you wanna just add type guards and use that as it will be faster and reusable?

Note that currently we also don't have any channel scale without both axis and legend but some of future  nonspatial channel (e.g., fontWeight) _may_ not have legend (simply because it's a pain to implement).

src/compile/mark/valueref.ts
```diff
@@ -102,7 +102,7 @@ export function midPoint(channel: Channel, channelDef: ChannelDef<string>, scale
       if (channelDef.bin) {
         // Use middle only for x an y to place marks in the center between start and end of the bin range.
         // We do not use the mid point for other channels (e.g. size) so that properties of legends and marks match.
-        if (contains(['x', 'y'], channel)) {
+        if (contains(['x', 'y'], channel) && channelDef.type === 'quantitative') {
```
kanitw: Good catch.   Maybe add comment that we only impute bin Q but not bin T?

src/compile/data/bin.ts
kanitw: Thinking more, in theory, all non-spatial channel can have legends. So I don't really know if we really need this check.

src/compile/data/bin.ts
domoritz: I liked this better as fieldDef is more standard and this way I can add a sanity check.

src/compile/data/aggregate.ts
domoritz: Done

src/compile/mark/valueref.ts
domoritz: I think the code says that already.

src/compile/data/bin.ts
domoritz: Not detail and text, right?

kanitw: Typo:  all non spatial scale channels*

domoritz: So, all non-spatial scale channels.

kanitw: >  all non-spatial channel can have legends
All non-spatial scale channels can have legends

So the following clause
```
!contains(NONSPATIAL_SCALE_CHANNELS, channel) && !contains(SPATIAL_SCALE_CHANNELS, channel)
```
is equivalent to `!isScaleChannel(channel)`

kanitw: For tooltip, this is not really needed, but we might need to modify it bin combine start and `_end`.

src/compile/mark/valueref.ts
kanitw: sort of, but it doesn't say why and it's kinda complicated so doesn't hurt to note.

domoritz:
Pushed revisions.

src/compile/data/bin.ts
domoritz: That's what I'm using now.

src/compile/common.ts
```diff
@@ -176,3 +171,18 @@ export function titleMerger(v1: Explicit<string>, v2: Explicit<string>) {
...
+/**
+ * Checks whether a fieldDef for a particular channel requires a computed bin range.
+ */
+export function binRequiresRange(fieldDef: FieldDef<string>, channel: Channel) {
+  if (!fieldDef.bin) {
+    console.warn('Only use this method with binned field defs');
+    return false;
+  }
+  if (!isScaleChannel(channel)) {
+    // range is only needed for scales that have legends or axes
```
kanitw: This comment is pretty useless and confusing (as it doesn't covered the whole criteria).  I just push a fix to simplify it.

src/compile/mark/valueref.ts
kanitw: I don't understand why this predicate below for using range doesn't match `binRequiresRange`.  
```
return fieldRef(channelDef, scaleName, isBinScale(scale.get('type')) ? {} : {binSuffix: 'range'});
```

Should it be 
```
return fieldRef(channelDef, scaleName, binRequiresRange(channelDef, channel) ? {binSuffix: 'range'} : {});
``` 
instead?  (I'm not 100% sure)

kanitw:
- [ ] One related problem is that we currently impute line/area by imputing `bin_mid`, but for ordinal, we can't impute `bin_end` and thus can't make correct `bin_range`.  This can get quick tricky.  The solution maybe showing only `bin_start`.  I'm also okay if we file an issue for that case and ignore it for now.

domoritz:
I can use stack and bin at the same time. Yes, imputing doesn't make sense here. 

```json
{
  "mark": "bar",
  "width": 800,
  "encoding": {
    "x": {
      "bin": {"maxbins": 100},
      "field": "Horsepower",
      "type": "ordinal"
    },
    "color": {
      "field": "Origin",
      "type": "nominal"
    },
    "y": {
      "field": "Acceleration",
      "type": "quantitative",
      "aggregate": "sum"
    }
  }
}
```

image

src/compile/common.ts
domoritz: Thanks

src/compile/mark/valueref.ts
domoritz: The code doesn't produce different results in any cases I can construct and that make sense.  Changed.

PR #2930, opened 2017-08-26
Do not `round` scale by default	

kanitw:


kanitw:
@arvind  I think something is weird about the setup for `npm run test:runtime:generate` -- I already run it to make it pass locally (and double checked) but it somehow still doesn't pass online.

arvind:
Hmm, diffing the strings thrown by the assertions against the stored SVG, it looks like there are tiny floating point errors occuring. E.g., 

```svg
<path transform="translate(166.66666666666669,300.6666666666667)" 
```

vs.

```svg
<path transform="translate(166.66666666666669,299.66666666666663)"
```

I would've thought the output would be standard for Chrome regardless of the underlying platform, but perhaps this isn't true? For now, perhaps just disable the assertion in `test-runtime/util.ts:171:12` since there are other assertions that validate the selection state? We can always return to this post-2.0 launch.

kanitw:
Oh shit, just saw your new comment.

PR #2933, opened 2017-08-27
remove `# $schema` from `getting_started`

g3o2:
Technically, a vega-lite specification works without out the `$schema` property. While this may not be recommended for many (albeit unmentioned) reasons, starting with `schema` in this context just breaks the flow.

As a side node: the `vega` tutorials do not mention `$schema` explicitly either.

### DIFF
--- a/site/tutorials/getting_started.md
+++ b/site/tutorials/getting_started.md
@@ -14,14 +14,6 @@ We suggest that you follow along the tutorial by building a visualization in the
 * TOC
 {:toc}
 
-## `$schema`
-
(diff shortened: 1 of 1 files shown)


kanitw:
I think it's reasonable to cut it from here.  The question is where should we introduce it back?

Throughout the docs, we include `$schema` in the example specs.  Thus, it might make sense to still include it in the tutorial somewhere -- maybe at the end of this page?  
(So that when people move on to read docs, they won't see $schema as a surprise.)

arvind:
> Throughout the docs, we include $schema in the example specs. 

Although `$schema` appears in the example specs in the editor, it [no longer appears](https://github.com/vega/vega-lite/pull/2672#discussion_r127853373) in the specs shown inline/embedded in the documentation.

kanitw:
@arvind Oh right.  I guess we can just merge this then?

domoritz:
I'll send a follow-up PR with some more docs on $schema.

PR #2935, opened 2017-08-30
Refactor TimeUnit's type guard and TIMEUNITS list	

kanitw:


kanitw:
cc: @domoritz @arvind -- I need a part of this for CompassQL / Voyager so I'll merge.  Feel free to send follow up comments.

arvind:
Did this PR cause the example section on the front page to break? The first two steps show a broken (single-bar) histogram. The last/third example shows the proper histogram, but x-axis labels are broken.

PR #2937, opened 2017-08-31
Extend docs on $schema

domoritz:
Follow up for #2933

### DIFF
--- a/src/spec.ts
+++ b/src/spec.ts
@@ -20,7 +20,7 @@ import {contains, Dict, duplicate, hash, vals} from './util';
 
 export type TopLevel<S extends BaseSpec> = S & TopLevelProperties & {
   /**
-   * URL to JSON schema for this Vega-Lite specification.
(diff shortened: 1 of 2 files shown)


src/spec.ts
```diff
@@ -20,7 +20,7 @@ import {contains, Dict, duplicate, hash, vals} from './util';
 
 export type TopLevel<S extends BaseSpec> = S & TopLevelProperties & {
   /**
-   * URL to JSON schema for this Vega-Lite specification.
+   * URL to [JSON schema](http://json-schema.org/) for this Vega-Lite specification. Unless you have a reason to change this, use `https://vega.github.io/schema/vega-lite/v2.json`. Setting the `$schema` property allows automatic validation and autocomplete in editors that support JSON schema.
```
kanitw: "for a Vega-Lite specification"?

(This => a)

domoritz: I think this is correct here as you can have your own dialect of Vega-Lite or use a specific version.

kanitw: This will appear in the docs, and it will be weird which specification is "this"?

Issue #1953, opened 2017-02-16
[Refactor] Replace extend() with spread operator (...)


kanitw:
Spread operator is more readable.

PR #2939, opened 2017-08-31
Replace vega-utils.replace() with spread operator

cjiang:
This is the refactoring proposed in #1953.

The merge() in src/compile/data/bin.ts - is that intended?


domoritz:
> The merge() in src/compile/data/bin.ts - is that intended?

We use the merge method to combine nodes in the dataflow. Right now we don't optimize the dataflow very much but in the future, we will need these functions more.

src/compile/selection/selection.ts
```diff
@@ -90,10 +90,10 @@ export function parseUnitSelection(model: UnitModel, selDefs: Dict<SelectionDef>
     }
 
     name = varName(name);
-    const selCmpt = selCmpts[name] = extend({}, selDef, {
+    const selCmpt = selCmpts[name] = {...selDef,
```
kanitw: I would add a line break between `{` and `...`

cjiang:
To clarify, I mean in bins.ts, merge() is implemented as:
`this.bins = {...other.bins};`
which suspiciously doesn't involve this.bins.

src/compile/selection/selection.ts
cjiang: Done.

domoritz:
> this.bins = {...other.bins};

Good catch @cjiang! Yes, that should be `this.bins = {...this.bins, ...other.bins};`!

kanitw:
Thanks for the PR :)

PR #2949, opened 2017-09-05
Don’t hide toc entries via css but instead use {:.no_toc}

domoritz:
See https://kramdown.gettalong.org/converter/html.html#toc

This has two advantages. First, it makes it more explicit what is shown. Second, we don't even output the HTML.


kanitw:
Genius!

PR #2950, opened 2017-09-05
Add docs for rule mark. Fixes #2947

### LINKED ISSUE #2947: Rule docs
When reading #2946, I realized we don't have docs for the rule mark.

domoritz:


site/docs/rule.md
```diff
@@ -0,0 +1,43 @@
...
+```json
+{
+  "data": ... ,
+  "mark": "rule",
+  "encoding": ... ,
+  ...
+}
+```
+
+The `rule` mark represents each data point as a line. It can be used in two ways. First, as a line that spans the complete width or height of a view. This is useful for adding annotations. Second, a rule can be used to draw a line between two positions.
```
kanitw: as a line segment

site/docs/rule.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+### Usage similar to `tick`
+
+The rule mark can be used similar to a [`tick`](tick.html) mark. The difference is that a rule mark automatically spans the complete width or height of a single view such that no dimension is required. We can sue this for example to show the average price of different stocks.
```
kanitw: I know you try to describe how it normally behaves, but the header "usage similar to tick" is quite weird and it is not really an intended use.   Can you make it just a part (intro paragraph) of "Rules as annotations".

site/docs/rule.md
```diff
@@ -0,0 +1,43 @@
...
+Here is an example where we use the spec from above to show the average value of stocks alongside the price curve.
+
+<span class="vl-example" data-name="layer_line_color_rule"></span>
+
+
+### Rules with two positional encodings
+
+Rules can encode two positions for x and y. This is useful to show multiple measures for a single ordinal value.
+
+For example, we can show the extent of horsepowers for cars from different locations.
```
kanitw: I think using rule as error bar is a more commonly used example.  Maybe add error bar?

kanitw: We can show how we  use aggregate with rule to show different types of error bars similar to the error bar example in Vega?

site/docs/rule.md
domoritz: Done

site/docs/rule.md
domoritz: Done

site/docs/rule.md
domoritz: I wanted to keep this example as simple as possible. Min and max are pretty easy to understand and I don't want to use layering here.

domoritz:
Rebased and revised.

site/docs/rule.md
kanitw: I think we should keep this example, but it is important to show error bars as it is gonna be very commonly used.

domoritz: Okay, let's do that in a follow up.

domoritz: Done

kanitw:
I made a revision.  If you like it, let's merge squash this. :)

PR #2951, opened 2017-09-05
Docs consistency

domoritz:
Make docs more consistent by removing selection- preix and also grouping docs in folders.


kanitw:
One small problem with this renaming scheme is that selection project does not even have `project` name so I kinda feel like it's a bit confusing?

domoritz:
Yeah, there is no project keywords but that's a separate issue and I don't think it's a big issue.

kanitw:
Ok sure. There are a lot of conflicts for some reason though.

domoritz:
Fixed

PR #2954, opened 2017-09-05
Add formatting docs and fix links for formatting. Fixes #2852

### LINKED ISSUE #2852: Missing Docs for `format`
(Need to copy some details from Vega/D3)

domoritz:


site/docs/format.md
```diff
@@ -4,4 +4,19 @@ title: Format
 permalink: /docs/format.html
 ---
 
-TODO
+In Vega-Lite specifications you can customize the format of [text marks](text.html), [tooltips](tooltip.html#using-tooltip-channel), and [axis](axis.html) and [legend](legend.html) labels.
```
kanitw: Note that this page is currently linked under encoding.   So the main content should be about `encoding.text/tooltip/format`.  

If we mention axis/legend here, we need to provide a bit more context that format can be a part of text/tooltip field def and axis/legend?

The currently examples we show are examples for axis but not for text (which is the main content for this page).  

For axis/legend examples, I think we have a few options:

1) Show these examples in the page, but divide them into proper sections (for text/tooltip and for axis/legend) and make axis/legend pages link to this page. 

or 2) Show them in axis/legend pages and make a link from here to those pages.

I kinda prefer 1) so that axis/legend pages won't be too long.

kanitw: FYI, I'll add "See the [format documentation](format.html) for more information." to `format` in the table for Axis/Legend.

domoritz: Done

kanitw:
Why are there no SVGs for the new files?

domoritz:
I get very different results from the existing SVGs and also didn't want to use up all my battery :-P

kanitw:
lol but we should generate SVG before merging

domoritz:
Working on it...

Issue #2948, opened 2017-09-03
Legend does not have opacity.


yhoonkim:
`"opacity":{"value": 0.1}`
![image](https://user-images.githubusercontent.com/6269456/29999558-8388841c-9002-11e7-9571-0f0fb45bed8d.png)
`"opacity":{"value": 1}`
![image](https://user-images.githubusercontent.com/6269456/29999574-12b2555a-9003-11e7-9fa5-5fc19197ab3c.png)
Default opacity
![image](https://user-images.githubusercontent.com/6269456/29999583-360067b8-9003-11e7-9547-8c027d1ab559.png)

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "mark": "point",
  "encoding": {
    "x": {
      "field": "AWND",
      "type": "quantitative"

    },
    "color": {
      "field": "WDF5",
      "type": "quantitative"
    },
    "y": {
      "field": "name",
      "type": "nominal"
    }
  },
  "data": {
    "values": [
      {
        "name": "Texas",
        "AWND": 19.46138,
        "WDF5": 160
      },
      {
        "name": "Texas",
        "AWND": 9.39515,
        "WDF5": 310
      }
    ]
  }
}
```
User cannot look up the values represented by those colors if the legend's color scale is not matched. 

The default of opacity value is 0.7 and is also not matched with the default color legend. It seems error-prone.

domoritz:
Good catch. Do you want to send a PR?

PR #2955, opened 2017-09-05
Legend can have default/constant opacity.

yhoonkim:
FIx https://github.com/vega/vega-lite/issues/2948.


src/compile/legend/encode.ts
```diff
@@ -60,6 +60,9 @@ export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: Uni
       symbols.shape = {value: shapeDef.value};
     }
   }
+  if (channel !== OPACITY) {
+    symbols.opacity = model.encoding.opacity;
```
kanitw: It's weird that for shape we have to check if `model.encoding.shape` is a value def but here you don't check the same thing for opacity.

I don't know if the old code is wrong or your new code is wrong, but we definitely need to investigate what's right.  :)

src/compile/legend/parse.ts
```diff
@@ -176,5 +176,21 @@ export function mergeLegendComponent(mergedLegend: LegendComponent, childLegend:
     );
     mergedLegend.setWithExplicit(prop, mergedValueWithExplicit);
   }
+
+  if (mergedLegend.implicit.encode.symbols.update.opacity && mergedLegend.implicit.opacity) {
```
kanitw: Don't add this HACK here.  

Add the rule inside the `legend/encode.ts` for the symbols method. (If there is no `symbols` method, create one so we have a consistent pattern to determine default values of things in encode block.)

src/compile/legend/encode.ts
```diff
@@ -69,7 +70,18 @@ export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: Uni
   return keys(symbols).length > 0 ? symbols : undefined;
 }
 
-export function labels(fieldDef: FieldDef<string>, labelsSpec: any, model: UnitModel, channel: NonspatialScaleChannel) {
+export function gradient(fieldDef: FieldDef<string>, gradientSpec: any, model: UnitModel, channel: Channel, legendCmpt: LegendComponent) {
+  const legend = model.legend(channel);
+  let gradient:any = {};
+
+  if (legendCmpt.explicit.type === 'gradient' || legendCmpt.implicit.type === 'gradient') {
```
kanitw: `if (legendCmpt.get('type') === 'gradient')`

src/compile/legend/encode.ts
```diff
@@ -6,9 +6,10 @@ import {keys, without} from '../../util';
 import {applyMarkConfig, timeFormatExpression} from '../common';
 import * as mixins from '../mark/mixins';
 import {UnitModel} from '../unit';
+import {LegendComponent} from './component';
 
 
-export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: UnitModel, channel: Channel) {
+export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: UnitModel, channel: Channel, legendCmpt: LegendComponent) {
```
kanitw: Why do you need to add a parameter here given you're not modifying the method?

test/compile/legend/encode.test.ts
```diff
@@ -32,6 +32,27 @@ describe('compile/legend', function() {
...
+    });
+  });
+
+  describe('encode.gradient', function() {
+    it('should have default opacity', function() {
+      const gradient = encode.gradient({field: 'a', type: 'quantitative'}, {}, parseUnitModelWithScale({
+          mark: "point",
+          encoding: {
+            x: {field: "a", type: "quantitative"}}
+        }), COLOR, {"implicit":{"type":"gradient"}} as LegendComponent);
```
kanitw: Please don't do unnecessary  `as` here.   You should instead properly create an instance of LegendComponent.

examples/compiled/diverging_color_points.vg.json
```diff
@@ -174,7 +174,23 @@
         {
             "stroke": "color",
             "title": "Weight_in_lbs",
-            "type": "gradient"
+            "type": "gradient",
+            "encode": {
+                "symbols": {
```
domoritz: we don't need this for gradients

kanitw: The logic is wrong.  Why would you need to set opacity for both symbols and gradient?

examples/compiled/facet_independent_scale_layer.vg.json
```diff
@@ -506,7 +506,12 @@
     "legends": [
         {
             "stroke": "color",
-            "title": "gender"
+            "title": "gender",
+            "encode": {
```
kanitw: Why adding an empty encode block?

examples/compiled/interval_mark_style.vg.json
```diff
@@ -874,7 +874,15 @@
         {
             "fill": "color",
             "title": "Number of Records",
-            "type": "gradient"
+            "type": "gradient",
+            "encode": {
```
kanitw: Empty block again.

src/compile/legend/encode.ts
```diff
@@ -1,14 +1,19 @@
...
 import {keys, without} from '../../util';
 import {applyMarkConfig, timeFormatExpression} from '../common';
 import * as mixins from '../mark/mixins';
 import {UnitModel} from '../unit';
+import {LegendComponent} from './component';
 
 
-export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: UnitModel, channel: Channel) {
+export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: UnitModel, channel: Channel, legendCmpt: LegendComponent) {
+  if (((legendCmpt.explicit || {}).type === 'gradient') || ((legendCmpt.implicit || {}).type === 'gradient')){
```
kanitw: `legendCmpt.get('type') ===  'gradient'`

(Maybe you miss my comment earlier)

src/compile/legend/encode.ts
```diff
@@ -61,12 +66,33 @@ export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: Uni
...
   symbols = {...symbols, ...symbolsSpec};
 
   return keys(symbols).length > 0 ? symbols : undefined;
 }
 
-export function labels(fieldDef: FieldDef<string>, labelsSpec: any, model: UnitModel, channel: NonspatialScaleChannel) {
+export function gradient(fieldDef: FieldDef<string>, gradientSpec: any, model: UnitModel, channel: Channel, legendCmpt: LegendComponent) {
+  let gradient:any = {};
+
+  if (((legendCmpt.explicit || {}).type === 'gradient') || ((legendCmpt.implicit || {}).type === 'gradient')) {
```
kanitw: legendCmpt.get('type') === 'gradient'

test/compile/legend/encode.test.ts
```diff
@@ -58,7 +88,7 @@ describe('compile/legend', function() {
       });
 
       const fieldDef = {field: 'a', type: TEMPORAL, timeUnit: TimeUnit.QUARTER};
-      const label = encode.labels(fieldDef, {}, model, COLOR);
+      const label = encode.labels(fieldDef, {}, model, COLOR,  legendCompt);
```
kanitw: one space after comma is sufficient

test/compile/legend/parse.test.ts
```diff
@@ -37,6 +40,12 @@ describe('compile/legend', function() {
         const model = parseUnitModelWithScale(s);
 
         const def = legendParse.parseLegendForChannel(model, channel).combine();
+
+        if (channel !== OPACITY){
+          assert.equal(def.encode.symbols.update.opacity.value, "0.7");
```
kanitw: It should be `0.7`, not `"0.7"`?

src/util.ts
```diff
@@ -238,3 +238,18 @@ export function logicalExpr<T>(op: LogicalOperand<T>, cb: Function): string {
 // Omit from http://ideasintosoftware.com/typescript-advanced-tricks/
 export type Diff<T extends string, U extends string> = ({[P in T]: P } & {[P in U]: never } & { [x: string]: never })[T];
 export type Omit<T, K extends keyof T> = {[P in Diff<keyof T, K>]: T[P]};
+
+// delete an object's empty key until it is not empty.
```
kanitw: /**
 * Delete nested property of an object, and delete the ancestors of the property if they become empty.
 */ 

Name it `deleteNestedProperty`.

PR #2961, opened 2017-09-06
Update axis/legend docs

### LINKED ISSUE #2855: Revise Axis / Legend Docs
- Scale 
  - [x] revise the whole page
  - [x] check all occurrences of "ordinal scale"
  - [x] scale and range config

- Axis / Legend 
  - [ ] Read if description for each property makes sense.
  - [ ] Please reorder properties and subsections by alphabetical order
  - Legend only 
    - [ ] please help reorganize the sections so it's more consistent with axis
    - [ ] Legend -- just like axis, we now use `null` to disable legend, not false.

kanitw:
Fix #2855


kanitw:
Rebased

PR #2965, opened 2017-09-06
Some SVG changes due to changes in vega 3.0.2

kanitw:


kanitw:
some changes are probably due to previous changes in specs  too.

PR #2966, opened 2017-09-06
Add separate documentation for encoding "condition" property.

### LINKED ISSUE #2853: Missing Docs for `condition`
We better describe about `condition` a bit in its own page as we want to list all properties of channel def on the left side under `encoding`.  This page could be very brief but explain the rules for the `condition` property. 

```
{
   condition: {... fielddef or valuedef}
   ...valuedef
}
```

```
{
   condition: {... valuedef only}
   ...fielddef
}
```

(I already explained this a bit in the encoding page, but we better give a bit more examples in this page.)

arvind:
Resolves #2853.


kanitw:
This looks solid :)

PR #2970, opened 2017-09-07
Custom translate trigger for multiple intervals.

### LINKED ISSUE #2706: Can move wrong selection
In the example in the selection docs, I can move the brush that requires shift to be clicked when it is dragged out without pressing shift. 

editor

arvind:
This prevents the wrong selection from being translated (resolves #2706).


kanitw:
Since @domoritz raises the issue, I'll let him review this then.

domoritz:
Hmm, I can still move the red selection even though it should only work when shift is clicked. 

editor

arvind:
Hmm, which example were you testing? I'm not able to reproduce this behavior with the three examples included in this PR (but perhaps I've forgotten to update other examples in the docs). For example, based on the GIF, I'm guessing it's `interval_mark_style.vl.json` but here's what I see with VL 2.0.0-beta.21 and VG 3.0.2:

brushshift

domoritz:
I used the example from https://vega.github.io/vega-lite/docs/selection.html#interval-mark. I guess that one behaves correctly as we are not specifying the trigger to move the selection. 

I noticed another issue. When I create an overlapping red and grey brush and I want to move the grey brush, I can only move it if the brush is not being covered by the red brush. 

editor

domoritz:
> We can open a feature request against vega-scenegraph, to address this in the future.

Thanks for the clarifications. I approved the PR. Please merge when we filed a feature request in vega-scenegraph.

arvind:
Done: vega/vega-scenegraph#53.

PR #2973, opened 2017-09-08
Rename config.cell => config.view + Add docs.

kanitw:
"cell" is a weird keyword that is coming from nowhere. 

Naming it "View Config" makes it fit nicely in the View Specification page. 

This is obviously a breaking change, but I strongly think we should make it right as it is the last chance.

### DIFF
--- a/_data/link.yml
+++ b/_data/link.yml
@@ -89,6 +89,9 @@ TickConfig:
 VgTitleConfig:
   name: TitleConfig
   link: "title.html#config"
+ViewConfig:
+  name: ViewConfig
+  link: "spec.html#config"
 
 # Encoding
 EncodingWithFacet:
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1029,55 +1029,6 @@
       ],
       "type": "object"
     },
-    "CellConfig": {
-      "additionalProperties": false,
-      "properties": {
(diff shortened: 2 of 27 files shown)


arvind:
I agree that `cell` makes no sense. We use `view` to mean both unit and composed views (we explicitly define this as such in the paper, but it's implicitly true in the existing documentation as well). Would `unitView` be clearer?

kanitw:
Thanks for raising the point.  I had similar concern too. 

Here is my reasoning for choosing `config.view`.   (I don't think it's the perfect name, but it seems least annoying among the choice I have considered.)  

Basically, in the docs, we only call unit view a single view specification.  So if we want a slightly more descriptive name that match the docs, it should be `config.singleView`. 

However, this config also customizes sizes and styles of layered views.  So `singleView` is not perfect accurately either.  I also think that, if we introduce other config options specific to each composite view in the future, we will probably name them `config.facet/concat/repeat/layer` (shorter and match operator names) rather than `config.facet/concat/repeat/layerView`.   
So I don't want to just make `singleView` the only thing with -View suffix. 

Thus, I think maybe choose a brief name like `config.view` and just explain what is it for in the documentation might be okay.

arvind:
Cool, that reasoning makes sense to me. Thanks!

PR #2974, opened 2017-09-08
Hide composite marks from schema and do not validate examples. Fixes #2837.

### LINKED ISSUE #2837: @hide Composite Mark
We discussed that we wanna hide box from the schema so it is an experimental feature. 

- [x] hide `CompositeMark` for mark type
- [x] hide `CompositeAggregate` for `aggregate`
- [x] mark all box examples as experimental and filter them from JSON schema validation.

domoritz:


src/mark.ts
```diff
@@ -116,7 +116,13 @@ export interface MarkDef {
   clip?: boolean;
 }
 
-export type AnyMark = CompositeMark | CompositeMarkDef | Mark | MarkDef;
+/** @hide */
+export type hiddenComposite = CompositeMark | CompositeMarkDef;
```
kanitw: Should this be HiddenComposite (Capital H?)

examples/specs/box-plot_minmax_1D_horizontal.vl.json
```diff
@@ -1,13 +0,0 @@
-{
```
kanitw: As said on slack, let's not remove examples so we have test to prevent regression and ensure they still produce vega and SVG outputs fine.

src/mark.ts
domoritz: Fixed

examples/specs/box-plot_minmax_1D_horizontal.vl.json
domoritz: One sec

domoritz:
Fixed and squashed

arvind:
LGTM, but I still see one removed example?

domoritz:
@arvind fixed

examples/specs/normalized/layer_box_plot_circle_normalized.vl.json
```diff
@@ -1,155 +0,0 @@
-{
```
kanitw: I think you forgot to re-compile the normalize script?

kanitw:
It should be good now, I just recompiled examples and push a fix.  :)

examples/specs/normalized/layer_box_plot_circle_normalized.vl.json
domoritz: I used npm run build:example... is that not enough?

PR #2977, opened 2017-09-08
Make facet's child either layer or unit specs for now 

kanitw:


examples/examples.test.ts
```diff
@@ -43,23 +43,28 @@ function validateVega(spec: TopLevelExtendedSpec) {
...
 
     describe(example, function() {
       it('should be valid vega-lite with proper $schema', function() {
+        if (
+          // Do not validate overlay example until we have redesigned it
+          example.indexOf('overlay') >= 0 ||
+          // Also ignore box-plot examples until we support selections
+          example.indexOf('box-plot') >= 0 ||
+          // Also ignore all examples with "_future" suffix
+          example.lastIndexOf('_future.vl.json') === example.length - suffixLength
```
arvind: This is an interesting way to make this test. An alternative might be a regular expression to verify that the string ends with the given suffix.

PR #2979, opened 2017-09-08
Clean up for Rc0

### LINKED ISSUE #2975: Remove axis domainColor/tickColor
(We already use the same default color in Vega, so no need to output them.)

kanitw:
Fix #2975


kanitw:
These are minor changes anyway.  I'll merge this as I wanna make a release.  

Feel free to send follow-up comments.

PR #2980, opened 2017-09-08
Reduce memory usage of hashes of large objects

domoritz:
I had an issue with large datasets passed directly into Vega-Lite. This reduces the memory usage.


domoritz:
Made some more improvements. This also cleans up the data section in Vega-Lite.

domoritz:
@kanitw I think I made significant changes that need to be reviewed again.

domoritz:
6b3d20f completely avoids the problems of large embedded datasets

domoritz:
Tests are failing because of https://github.com/vega/vega-parser/pull/60

domoritz:
Tests passing.

src/compile/data/source.ts
```diff
@@ -20,35 +20,37 @@ export class SourceNode extends DataFlowNode {
...
+        let defaultExtension = /(?:\.([^.]+))?$/.exec(data.url)[1];
+        if (!contains(['json', 'csv', 'tsv', 'topojson'], defaultExtension)) {
+          defaultExtension = 'json';
+        }
+
+        if (defaultExtension !== 'json') {
+          if (!data.format) {
+            data.format = {};
+          }
+          data.format.type = defaultExtension as DataFormatType;
```
kanitw: Why do we need `as` here?  Please add a comment to explain.

src/data.ts
```diff
@@ -12,15 +13,31 @@ export interface DataFormat {
...
+  /**
+   * Type of input data: `"json"`, `"csv"`, `"tsv"`.
+   * The default format type is determined by the extension of the file URL.
+   * If no extension is detected, `"json"` will be used by default.
+   */
+  type?: DataFormatType;
+}
 
+export interface CsvDataFormat extends DataFormatBase {
+  type?: 'csv' | 'tsv';
```
kanitw: Shouldn't this get the same JSDocs comment as the original type it overrides?

(So that the editor always get comment during autocompletion.)

src/compile/data/source.ts
```diff
@@ -20,35 +20,37 @@ export class SourceNode extends DataFlowNode {
...
+        }
       }
-      const dataFormat = data.format || {};
-
-      // For backward compatibility for former `data.formatType` property
```
kanitw: Make sure to add this to the release note.

examples/compiled/embedded_csv.vg.json
```diff
@@ -0,0 +1,126 @@
+{
+    "$schema": "https://vega.github.io/schema/vega/v3.0.json",
+    "autosize": "pad",
+    "padding": 5,
+    "style": "cell",
+    "data": [
+        {
+            "name": "source_0",
+            "values": "a\n1\n2\n3\n4",
```
kanitw: So Vega actually supports CSV as inline value too?

domoritz: Yes. This example works.

src/compile/data/source.ts
domoritz: Done

src/compile/data/source.ts
domoritz: defaultExtension has type string but we ensure that it is DataFormatType above

done

src/data.ts
domoritz: No. That's the nice thing about the change. Only supported properties are autocompleted. 

screen shot 2017-09-10 at 13 39 56

domoritz:
Revised and increased test coverage

test/compile/data/source.test.ts
```diff
@@ -20,9 +20,10 @@ describe('compile/data/source', function() {
         assert.deepEqual(source.data.values, [{a: 1, b:2, c:3}, {a: 4, b:5, c:6}]);
       });
 
-      it('should have no source.format.type', function(){
-        assert.deepEqual(source.data.format, undefined);
-      });
+      // https://github.com/vega/vega-parser/pull/60
```
kanitw: So this can be revert after https://github.com/vega/vega-parser/pull/60 is included in the next version of Vega?

domoritz: Probably, yes

src/data.ts
kanitw: I mean adding `type?: 'csv' | 'tsv';` here is good .
But currently, if you mouseover "type" in the editor, you won't get JSDocs comment like you hover other properties such as "format". 

Thus, I wonder if copying JSDocs from the original to all specific `type` you override would include JSDocs comment for `type` when users hover?

domoritz: Done

Issue #2946, opened 2017-09-01
1-D line mark type doesn't render nicely


jtalbot:
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"a": "A","b": 28}, {"a": "B","b": 55}, {"a": "C","b": 43},
      {"a": "D","b": 91}, {"a": "E","b": 81}, {"a": "F","b": 53},
      {"a": "G","b": 19}, {"a": "H","b": 87}, {"a": "I","b": 52}
    ]
  },
  "mark": "line",
  "encoding": {
    "y": {"field": "b", "type": "quantitative"}
  }
}
```
renders as:
<img width="53" alt="screen shot 2017-09-01 at 3 20 20 pm" src="https://user-images.githubusercontent.com/638466/29989643-187b910c-8f29-11e7-9680-38e2f69cd4d3.png">

It should look like:
<img width="94" alt="screen shot 2017-09-01 at 3 30 40 pm" src="https://user-images.githubusercontent.com/638466/29989850-a19d24cc-8f2a-11e7-891c-1facf24115e1.png">

There might be 2 issues here. The line mark rendering on the left edge instead of in the middle seems to be a Vega-lite issue; for other mark types the mark mapping to the x and y axes is different and better behaved. The incorrect padding seems to be a Vega issue caused by the following code:
```
"sort": {
    "order": "descending"
},
```
If this is removed the axis and padding render correctly.

g3o2:
At first, I wondered why would one want to use a 1D `line` instead of `rule`.

On second thought, I realized the beauty in this idea: one could use this line as a min-max `domain` through concatenation (and some tweaking) in a scatter plot. Very nice!

domoritz:
@g3o2 You can use x and x2 with the role mark.

g3o2:
Yes, I know the rule mark. However, with the line mark, you don't need any transform on the user side to determine x and x2.

kanitw:
@jtalbot   -- I wonder if 

replacing
```
...mixins.pointPosition('x', model, 'zeroOrMin'),
...mixins.pointPosition('y', model, 'zeroOrMin'),
```
in `src/compile/mark/line.ts`

with 

```
...mixins.pointPosition('x', model, ref.mid(width)),
...mixins.pointPosition('y', model, ref.mid(height)),
```

[...]

PR #2983, opened 2017-09-08
Don't sort by missing fields

jtalbot:
Fixes the other half of #2946. 1D lines will now render correctly in Firefox and Safari.


kanitw:
Thanks :)

PR #2989, opened 2017-09-10
Interactive dashboard example

domoritz:
screen shot 2017-09-10 at 14 01 12


kanitw:
The fact that the rect's color and the point's size encode the same thing is a bit weird.
I think using filled circles (or squares) with size for both layer might be better:

```
{
  "description": "A dashboard with cross-highlighting.",
  "vconcat": [
    {
      "layer": [{
        "mark": "circle",
        "encoding": {
          "x": {
            "bin": {"maxbins": 10},
            "field": "IMDB_Rating",
            "type": "quantitative"
          },
          "y": {
            "bin": {"maxbins": 10},
            "type": "quantitative"
          },
          "size": {
            "aggregate": "count",
            "type": "quantitative",
            "legend": {
              "title": "All Movies"
            }
          },
          "opacity": {"value": "0.4"}
        }
      }, {
        "transform": [{
          "filter": {"selection": "pts"}
        }],
        "mark": "circle",
        "encoding": {
          "x": {
            "bin": {"maxbins": 10},
            "field": "IMDB_Rating",
            "type": "quantitative"
          },
          "y": {
            "bin": {"maxbins": 10},
            "type": "quantitative"
          },
          "size": {
            "aggregate": "count",
            "type": "quantitative",
            "legend": {
              "title": "In Selected Category"
            }
          }
        }
      }]
    }, {
      "width": 330,
      "height": 120,
      "mark": "bar",
      "selection": {
        "pts": {"type": "single", "encodings": ["x"]}
      },
      "encoding": {
        "x": {"field": "Major_Genre", "type": "nominal", "axis": {"labelAngle": -40}},
        "y": {"aggregate": "count", "type": "quantitative"},
        "color": {
          "condition": {
            "selection": "pts",
            "value": "steelblue"
          },
          "value": "grey"
        }
      }
    }
  ],
  "resolve": {
    "legend": {
      "size": "independent"
    }
  },
  "config": {
    "range": {
      "heatmap": {
        "scheme": "greenblue"
      }
    }
  }
}
```

image

kanitw:
I think this should facilitate comparison better?

domoritz:
Size and color are easier to separate. The original example makes it easier to look at the variables in isolation.

kanitw:
Yeah, but it doesn't facilitate comparison at all.  What's the point of overlaying them if you don't want comparison?  In such case, better show only filtered value.

domoritz:
I want comparison but also at the same time see only the highlighted value. See https://www.cs.rpi.edu/~cutler/classes/visualization/F10/papers/ware_texton.pdf

kanitw:
From skimming the screenshot it doesn't seem like your example here exactly match the paper?

What's the goal of this example?  If the goal is to demonstrate that users can create dashboard, I think the uncommon visual encoding design would distract readers from focusing on what it is design to communicate?

_data/examples.json
```diff
@@ -217,6 +217,10 @@
     {
       "name": "interactive_splom",
       "title": "Scatterplot Matrix"
+    },
+    {
+      "name": "concat_layer.vl.json",
+      "title": "Interactive Dashboard"
```
kanitw: Do you have a name for the upper plot?

"Coordinated bar chart and XXX"? 

(Interactive Dashboard seems pretty generic?)

domoritz: Cross highlight

PR #2990, opened 2017-09-10
Add normalized BoxPlot specs to the gallery + Add Tukey BoxPlot example

kanitw:


_data/examples.json
```diff
@@ -154,6 +154,14 @@
       "style": "background-size: auto 55%; background-position-y: center !important;"
     },
     {
+      "name": "normalized/box-plot_minmax_2D_vertical_normalized",
```
kanitw: Currently we need to add normalized example since our "box" syntax is not official yet.  However, the example link does not work with path (e.g., `normalized/`).

There are 3 ways to fix this:
1. Make Vega Editor support this
2. Temporarily add these two normalized examples to `examples/specs`
3. Move all normalized examples to `examples/specs` (This will require editing so many other specs.) 

I prefer 2.  -- @domoritz -- Any opinion?

domoritz: 2 sounds good.

Issue #2503, opened 2017-06-14
Investigating if we can support autosize: 'fit'


kanitw:
From slack conversation

> What `fit` does: (1) calculates the degree to which the current view exceeds the user-specified width / height, (2) updates the internal `width`/`height` signals to accommodate the excess (as much as possible), (3) re-runs the dataflow.
> Whether or not layout works depends on to what degree sub-elements respond to changes in the top-level `width`/`height` signals. So in general, layout may not work. Nor will any other visualization that does not respond to the top-level `width`/`height` signals.

It might be possible to support `autosize: fit`. 

We could investigate if we can support this as a minor patch (post 2.0).

PR #2991, opened 2017-09-10
Autosize Fit

### LINKED ISSUE #2981: Revise if `autoResize` is a name that we will not change forever
@domoritz I remember you add `autoResize`, which is basically `autosize.resize` in Vega.

I can't remember exactly why they should have different names, except that we do not support other `autosize` parameters in Vega-Lite (as least for now).  However, I am not sure if this will remain true forever.  Thus, I wonder if we should either revise this naming or not `@hide` it until we're confident about it. 

(I also can't remember the use case for it.)

domoritz:
Fixes #2981, #2503

CC @nyurik This addresses https://github.com/nyurik/kibana-vega-vis/issues/4


kanitw:
Unfortunately, this doesn't seem to generally work yet. 

For example, fit doesn't work for the following unit spec:

```
{
  "description": "A simple bar chart with embedded data that fits exactly into 300px width.",
  "width": 300,
  "autosize": {
    "type": "fit",
    "contains": "padding"
  },
  "data": {
    "values": [
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "b", "type": "quantitative"},
    "color": {"field": "b", "type": "quantitative"}
  }
}
```

image

I think it also won't work for concat/facet/repeat.

domoritz:
> I think it also won't work for concat/facet/repeat.

Yes, it won't. I made this very clear in the documentation and also added a warning.

src/compile/compile.ts
```diff
@@ -65,23 +69,43 @@ function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config) {
...
+    return true;
+  });
+
+  if (autosize === 'fit' || (autosize && isAutosize(autosize) && autosize.type === 'fit')) {
+    if (!topLevelProps.width || !topLevelProps.height) {
+      log.warn(log.message.FIT_DATA_DRIVEN);
+    }
+
+    if (!(model instanceof UnitModel) && !(model instanceof LayerModel)) {
+      log.warn(log.message.FIT_NON_SINGLE);
```
kanitw: Using this fit thing with facet makes the visualization output nothing.  Won't it be better to drop it? 

In fact, if possible, we may want to make the autosize fit only available for unit / layer in the schema?

domoritz:
Works for me

screen shot 2017-09-10 at 17 21 36

src/compile/compile.ts
```diff
@@ -65,23 +69,43 @@ function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config) {
...
+  layoutSignals = layoutSignals.filter(signal => {
+    if ((signal.name === 'width' || signal.name === 'height') && signal.value !== undefined) {
+      topLevelProps[signal.name] = signal.value;
+      return false;
+    }
+    return true;
+  });
+
+  if (autosize === 'fit' || (autosize && isAutosize(autosize) && autosize.type === 'fit')) {
+    if (!topLevelProps.width || !topLevelProps.height) {
```
kanitw: Instead of warning, I think we better modify scale range logic to support this case.

domoritz: I'll do both. We should support this case but also show a warning that we don't determine the size in a data driven way.

src/compile/compile.ts
domoritz: I could do it in the schema but that requires changing all instances of TopLevelProperties and TopLevel and probably more. I'll drop it in the code.

src/compile/compile.ts
kanitw: I guess we should only warn when users explicitly specify both autosize and range step?

(Otherwise, why would we need to warn?)

src/compile/compile.ts
kanitw: > changing all instances of TopLevelProperties and TopLevel and probably more

I doubt if it is really that much work.

src/compile/compile.ts
domoritz: Because the behavior is different from when autosize is not fit

src/compile/compile.ts
domoritz: I'd have to make TopLevelProperties a generic. Seems overkill in this case.

src/compile/scale/range.ts
```diff
@@ -45,13 +45,19 @@ function parseUnitScaleRange(model: UnitModel) {
 
     // Read if there is a specified width/height
     const sizeType = channel === 'x' ? 'width' : channel === 'y' ? 'height' : undefined;
-    const specifiedSize = sizeType ? model.component.layoutSize.get(sizeType) : undefined;
+    let sizeSpecified = sizeType ? !!model.component.layoutSize.get(sizeType) : undefined;
+
+    // if autosize is fit, size cannot be data driven
+    if (sizeType && model.fit && !sizeSpecified) {
```
kanitw: If neither width/height or rangeStep is specified, we shouldn't throw warning.  

Basically, if user either specify width/height or fit, that means users intend to fit the range step.  Thus the warning would be confusing for users.   

I think we should instead
- [ ] update relevant docs (`size.md` + JSDocs for range step and width/height) to explain about this.
- [ ] throw warning only if users specify both rangeStep and fit.

src/log.ts
```diff
@@ -98,6 +98,11 @@ export function debug(..._: any[]) {
 export namespace message {
   export const INVALID_SPEC = 'Invalid spec';
 
+  // FIT
+  export const FIT_NON_SINGLE = 'Autosize fit only works well for single views that are not composed of other views except if layered.';
```
kanitw: 'Autosize "fit" only works for single views and layered views'

src/toplevelprops.ts
```diff
@@ -20,15 +21,45 @@ export interface TopLevelProperties {
...
+   * Object values can additionally specify parameters for content sizing and automatic resizing. See the autosize section below for more.
+   *
+   * __Default value__: `pad`
+   */
+  autosize?: AutosizeType | Autosize;
+}
+
+export type AutosizeType = 'pad' | 'fit' | 'none';
+
+export interface Autosize {
```
kanitw: Name this `AutoSizeParams`?  

Technically, `Autosize = AutosizeType | AutoSizeParams`.  
(This will be consistent with `bin: boolean: BinParams`, etc.)

src/toplevelprops.ts
```diff
@@ -20,15 +21,45 @@ export interface TopLevelProperties {
...
+  /**
+   * Required. The sizing format type. One of `"pad"` (default), `"fit"` or `"none"`.
+   *
+   * __Default value__: `pad`
+   */
+  type?: AutosizeType;
+
+  /**
+   * A boolean flag indicating if autosize layout should be re-calculated on every view update.
```
kanitw: is view update specific to stream data?  If so may expand the description a bit?

site/docs/view/size.md
```diff
@@ -54,6 +54,24 @@ For example, the following plot use `21` as a default height.
...
+The specified dimensions of a chart as explained above set the size of the data rectangle (plotting) dimensions. You can override this behavior by setting the autosize property in [the top level specification](spec.html#top-level-specifications).
+
+The autosize object has the following properties:
+
+{% include table.html props="type,resize,contains" source="Autosize" %}
+
+In order to fit a chart into specified dimensions, it has to satisfy two properties.
+
+* Only single and layered views are supported.
+* The width and height of the chart cannot be data driven. This means that if you use a discrete scale for `x` or `y`, you have to set the corresponding size (`width` for `x`, `height` for `y`) proeprty explicitly.
```
kanitw: This is outdated.

site/docs/view/size.md
```diff
@@ -54,6 +54,24 @@ For example, the following plot use `21` as a default height.
 
 <span class="vl-example" data-name="bar_1d_rangestep_config"></span>
 
+### Autosize
+
+The specified dimensions of a chart as explained above set the size of the data rectangle (plotting) dimensions. You can override this behavior by setting the autosize property in [the top level specification](spec.html#top-level-specifications).
```
kanitw: - [ ] Please add
  - the `autosize` property
  - , which can be either a string or an autosize parameter object.

- [ ] Missing description about what none, pad, and fit mean.  (Better copy content from https://vega.github.io/vega/docs/specification/#autosize).  

> The specified dimensions of a chart as explained above set the size of the data rectangle (plotting) dimensions.

I think this is an important point that is kinda hidden in this section.  Maybe add this to the intro so we provide better overview how size works?

src/compile/scale/range.ts
kanitw: Also, currently the logic here throw warning even for height when y is continuous. 

image

(There are two of these warnings in the screenshot.)

src/log.ts
```diff
@@ -98,6 +98,11 @@ export function debug(..._: any[]) {
...
+
+  export const FIT_DATA_DRIVEN = 'Autosize fit does not work with data driven width or height.';
```
kanitw: We never talk about "data driven" width/height before (except in the new part of docs you added) so this is unclear that this means using range step. 

A better warning would be `'Cannot use "rangeStep" when "autosize" is "fit"'`.

src/spec.ts
```diff
@@ -207,6 +206,10 @@ export function isHConcatSpec(spec: GenericSpec<GenericUnitSpec<any, any>>): spe
   return spec['hconcat'] !== undefined;
 }
 
+export function autoSizeFitSpec(spec: TopLevelExtendedSpec) {
```
kanitw: This method is weirdly named and it seems unclear to me why we need it. 

We can just call `autoSizeFit(spec.autosize)` directly.

src/toplevelprops.ts
```diff
@@ -58,6 +58,10 @@ export function isAutosize(a: Autosize | AutosizeType): a is Autosize {
   return !isString(a);
 }
 
+export function autoSizeFit(autosize: AutosizeType | Autosize) {
```
kanitw: `isAutoSizeFit`?

src/compile/compile.ts
```diff
@@ -65,23 +69,43 @@ function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config) {
...
+      log.warn(log.message.FIT_NON_SINGLE);
+    }
+  }
+
   const output = {
     $schema: 'https://vega.github.io/schema/vega/v3.0.json',
     ...(model.description ? {description: model.description} : {}),
-    // By using Vega layout, we don't support custom autosize
-    autosize: topLevelProperties.autoResize ? {type: 'pad', resize: true} : 'pad',
+    autosize: autosize && isAutosize(autosize) ? {...{type: 'pad'}, ...autosize} : autosize || 'pad',
```
kanitw: I think we should just normalize autoSize to have an object form from the beginning, so we don't need complicated logic here nor  helper methods like `autosizeFit`?  (Just read `autosize.type === 'fit'`). 

(We mostly do the same thing for other similar structures with a short form such as mark and bin.)

src/toplevelprops.ts
```diff
@@ -20,15 +21,49 @@ export interface TopLevelProperties {
...
+   * __Default value__: `"content"`
+   */
+  contains?: 'content' | 'padding';
+}
+
+export function isAutosize(a: Autosize | AutosizeType): a is Autosize {
+  return !isString(a);
+}
+
+export function autoSizeFit(autosize: AutosizeType | Autosize) {
```
kanitw: See this [comment](https://github.com/vega/vega-lite/pull/2991#discussion_r137980261)

src/compile/scale/range.ts
domoritz: Good catch. Chrome just showed me a 2 in front of the message that I missed.

src/log.ts
domoritz: Done

src/log.ts
domoritz: Done

src/toplevelprops.ts
domoritz: Done

src/toplevelprops.ts
domoritz: No, it can also happen with interactions.

site/docs/view/size.md
domoritz: Done

site/docs/view/size.md
domoritz: Done

src/spec.ts

src/compile/compile.ts
domoritz: Unfortunately this doesn't quite work with the typings but I am adding normalization as it makes the code a bit easer.

src/toplevelprops.ts
domoritz: I don't like `(inputSpec.autosize as AutoSizeParams).type === 'fit'`

src/compile/compile.ts
kanitw: The trick is to declare 
```
const autoSize = normalizeAutoSize(topLevelProperties.autoSize); 
```
then for the rest of the code `autoSize` is `AutoSizeParams`

domoritz: I have a lot of places where I only have a spec. This works in one place only.

domoritz: I guess I can also do `normalizeAutoSize(spec.autosize).type === 'fit'`

domoritz:
Revised

site/docs/view/size.md
```diff
@@ -54,6 +54,34 @@ For example, the following plot use `21` as a default height.
...
+- `pad`: Automatically increase the size of the view such that all visualization content is visible. This is the default _autosize_ setting, and ensures that axes, legends and other items outside the normal width and height are included. The total size will often exceed the specified width, height, and padding.
+- `fit`: Automatically adjust the layout in an attempt to force the total visualization size to fit within the given width, height and padding values. This setting causes the plotting region to be made smaller in order to accommodate axes, legends and titles. As a result, the value of the _width_ and _height_ signals may be changed to modify the layout. Though effective for many plots, the `fit` method can not always ensure that all content remains visible. For example, if the axes and legends alone require more space than the specified width and height, some of the content will be clipped. Similar to `none`, by default the total width will be `width + padding.left + padding.right`, relative to the original, unmodified _width_ value. If _autosize.contains_ is set to `"padding"`, the total width will instead be the original _width_.
+
+### Limitations
```
kanitw: Why is this the same heading level of autosize?

domoritz: Ups

site/docs/view/size.md
```diff
@@ -54,6 +54,34 @@ For example, the following plot use `21` as a default height.
...
 
+### Autosize
+
+The specified dimensions of a chart as explained above set the size of the data rectangle (plotting) dimensions. You can override this behavior by setting the autosize property in [the top level specification](spec.html#top-level-specifications). Please note the [limitations below](#limitations).
+
+The autosize property can be a string or an object with the following properties:
+
+{% include table.html props="type,resize,contains" source="AutoSizeParams" %}
+
+The total size of a Vega-Lite visualization may be determined by multiple factors: specified _width_, _height_, and _padding_ values, as well as content such as axes, legends, and titles. The support different use cases, there are three different _autosize_ types for determining the final size of a visualization view:
```
nyurik: "To support", not "The support"

domoritz: Thanks!

src/toplevelprops.ts
domoritz: Fixed differently

site/docs/view/size.md
```diff
@@ -54,6 +54,34 @@ For example, the following plot use `21` as a default height.
...
+- `fit`: Automatically adjust the layout in an attempt to force the total visualization size to fit within the given width, height and padding values. This setting causes the plotting region to be made smaller in order to accommodate axes, legends and titles. As a result, the value of the _width_ and _height_ signals may be changed to modify the layout. Though effective for many plots, the `fit` method can not always ensure that all content remains visible. For example, if the axes and legends alone require more space than the specified width and height, some of the content will be clipped. Similar to `none`, by default the total width will be `width + padding.left + padding.right`, relative to the original, unmodified _width_ value. If _autosize.contains_ is set to `"padding"`, the total width will instead be the original _width_.
+
+#### Limitations
+
+In order to `fit` a chart into specified dimensions, it has to satisfy two things.
```
nyurik: "two requirements".  "things" is too informal imo

domoritz: Agreed

site/docs/view/size.md
```diff
@@ -54,6 +54,34 @@ For example, the following plot use `21` as a default height.
...
+
+* Only single and layered views are supported.
```
nyurik: Would this be better?  "The view must be either single or layered."   I was told passive is bad ))  Regardless, I have no clue what this means. Please link to definition of those terms.

domoritz: I improved the wording and also added links to the corresponding docs.

site/docs/view/size.md
domoritz: Fixed

domoritz:
Ready for merge

tslint.json
```diff
@@ -1,4 +1,8 @@
 {
+  "tslint.options": {
```
kanitw: @domoritz  Can you explain why we need this part? (It seems irrelevant to this PR?)

src/compile/compile.ts
```diff
@@ -88,11 +93,18 @@ function assembleTopLevelModel(model: Model, topLevelProperties: TopLevelPropert
     return true;
   });
 
+  if (autosize.type === 'fit') {
```
kanitw: This logic should be consolidate as a part of `normalizeAutoSize` and `normalizeAutoSize` should be called only once in `compile.ts`.

src/compile/compile.ts
```diff
@@ -57,10 +55,11 @@ export function compile(inputSpec: TopLevelExtendedSpec, logger?: log.LoggerInte
 }
 
 
-function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config) {
+function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config, autosize: AutoSizeParams) {
   return {
+    autosize: keys(autosize).length === 1 ? autosize.type : autosize,
```
kanitw: Note that the ternary operator here is not really necessary but I keep it to minimize the spec diffs.

kanitw:
Have one question about lint -- other than that I have improved the changed logic (it is way more consolidated now). I think it is good to go – @domoritz

tslint.json
domoritz: This just enables better linting is vscode because it knows the arguments we pass to tslint.

PR #2993, opened 2017-09-11
Do not output sort for 1D line

### LINKED ISSUE #2985: Do not generate sort without fields
https://github.com/vega/vega-scenegraph/issues/52

We should enforce that through typings.

kanitw:
Fix #2985


kanitw:
Looks like @jtalbot already did this in https://github.com/vega/vega-lite/pull/2983

Issue #2986, opened 2017-09-09
Test whether people find how to look at the different examples on the homepage. 


domoritz:
We might need to highlight the picker, add a "next" button, or auto advance. 

cc @arvind

PR #2996, opened 2017-09-11
Say "Click the links to see each step of the example" in the index page

kanitw:
Address https://github.com/vega/vega-lite/issues/2986

Given that you worry that some people might not click the links, we should just say explicitly that they should click.  (This is probably easier than "testing if they really click".


kanitw:
(I think next button might be better, but this seems easiest to add for now.)

domoritz:
There is already code for a button (my first implementation used a button). I'd prefer that.

PR #3001, opened 2017-09-11
Fix broken example in the master

kanitw:


domoritz:
Hmm, should we automatically convert values to strings?

kanitw:
This is actually the opposite.

PR #3002, opened 2017-09-11
Config Docs Revision

kanitw:


build/vega-lite-schema.json
```diff
@@ -448,7 +448,7 @@
           "type": "number"
         },
         "color": {
-          "description": "Default color.  Note that `fill` and `stroke` have higher precedence than `color` and will override `color`.\n\n__Default value:__ <span style=\"color: #4682b4;\">&#9632;</span> `\"#4682b4\"`",
+          "description": "Default color.  Note that `fill` and `stroke` have higher precedence than `color` and will override `color`.\n\n__Default value:__ <span style=\"color: #4682b4;\">&#9632;</span> `\"#4682b4\"`\n\n__Note:__ This property cannot be used in a [style config](mark.html#style-config).",
```
domoritz: Can we enforce that in the schema and only write it once above a table in the docs?

kanitw: It is actually already enforced in the schema.  I just do not want to have two tables for MarkConfig and VLMarkConfig in the docs. (The former do not have filled and color.)

PR #3006, opened 2017-09-13
Support repeat in conditions and sort

domoritz:
@sorenmacbeth this fixes your issue.

### DIFF
--- a/src/compile/repeater.ts
+++ b/src/compile/repeater.ts
@@ -1,16 +1,7 @@
 import {Encoding} from '../encoding';
 import {Facet} from '../facet';
-import {
-  ConditionalFieldDef,
-  ConditionalValueDef,
-  Field,
-  FieldDef,
-  isConditionalDef,
-  isFieldDef,
-  isRepeatRef,
-  LegendFieldDef,
-} from '../fielddef';
-import {ChannelDef} from '../fielddef';
+import {Field, FieldDef, isConditionalDef, isFieldDef, isRepeatRef} from '../fielddef';
+import {ChannelDef, ScaleFieldDef} from '../fielddef';
 import * as log from '../log';
 import {isSortField} from '../sort';
 import {isArray} from '../util';
@@ -29,42 +20,41 @@ export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: R
 }
 
 /**
+ * Replaces repeated value and returns if the repeated value is valid.
+ */
+function replaceRepeat<T extends {field?: Field}>(o: T, repeater: RepeaterValue): T {
+  if (isRepeatRef(o.field)) {
+    if (o.field.repeat in repeater) {
+      // any needed to calm down ts compiler
+      return {...o as any, field: repeater[o.field.repeat]};
+    } else {
+      log.warn(log.message.noSuchRepeatedValue(o.field.repeat));
+      return undefined;
+    }
+  }
+  return o;
+}
+
+/**
  * Replace repeater values in a field def with the concrete field name.
  */
-function replaceRepeaterInFieldDef(fieldDef: LegendFieldDef<Field>, repeater: RepeaterValue): LegendFieldDef<string> | null {
-  const field = fieldDef.field;
-  if (isRepeatRef(field)) {
-    if (field.repeat in repeater) {
-      fieldDef = {
-        ...fieldDef,
-        field: repeater[field.repeat]
-      };
-    } else {
-      log.warn(log.message.noSuchRepeatedValue(field.repeat));
-      // return null, meaning that the field def should be ignored
-      return null;
-    }
+function replaceRepeaterInFieldDef(fieldDef: ScaleFieldDef<Field>, repeater: RepeaterValue): ScaleFieldDef<string> {
+  fieldDef = replaceRepeat(fieldDef, repeater);
+
+  if (fieldDef === undefined) {
+    // the field def should be ignored
+    return undefined;
   }
 
-  const sort = fieldDef.sort;
-  if (sort && isSortField(sort) && isRepeatRef(sort.field)) {
-    if (sort.field.repeat in repeater) {
-      fieldDef = {
-        ...fieldDef,
-        sort: {
-          ...sort,
-          field: repeater[sort.field.repeat]
-        }
-      };
-    } else {
-      log.warn(log.message.noSuchRepeatedValue(sort.field.repeat));
-      // remove sort as it uses unknown field
-      const {sort: _s, ...newFieldDef} = fieldDef;
-      fieldDef = newFieldDef;
-    }
+  if (fieldDef.sort && isSortField(fieldDef.sort)) {
+    const sort = replaceRepeat(fieldDef.sort, repeater);
+    fieldDef = {
+      ...fieldDef,
+      ...(sort ? {sort} : {})
+    };
   }
 
-  return fieldDef as LegendFieldDef<string>;
+  return fieldDef as ScaleFieldDef<string>;
 }
 
 type EncodingOrFacet<F> = Encoding<F> | Facet<F>;
@@ -77,10 +67,10 @@ function replaceRepeater(mapping: EncodingOrFacet<Field>, repeater: RepeaterValu
 
       if (isArray<FieldDef<Field>>(channelDef)) {
         out[channel] = channelDef.map(fd => replaceRepeaterInFieldDef(fd, repeater))
-          .filter(fd => fd !== null);
+          .filter(fd => fd);
       } else if (isConditionalDef(channelDef) && isFieldDef(channelDef.condition)) {
         const fd = replaceRepeaterInFieldDef(channelDef.condition, repeater);
-        if (fd !== null) {
+        if (fd) {
           out[channel] = {
             ...channelDef,
             condition: fd
@@ -88,7 +78,7 @@ function replaceRepeater(mapping: EncodingOrFacet<Field>, repeater: RepeaterValu
         }
       } else if (isFieldDef(channelDef)) {
         const fd = replaceRepeaterInFieldDef(channelDef, repeater);
-        if (fd !== null) {
+        if (fd) {
           out[channel] = fd;
         }
       } else {
--- a/src/compile/repeater.ts
+++ b/src/compile/repeater.ts
@@ -66,9 +66,8 @@ function replaceRepeaterInChannelDef(channelDef: ChannelDef<Field>, repeater: Re
         condition: fd
       } as ChannelDef<string>;
     } else {
-      return {
-        value: (channelDef as ValueDef).value
-      };
+      const {condition, ...channelDefWithoutCondition} = channelDef;
+      return channelDefWithoutCondition as ChannelDef<string>;
     }
   } else if (isFieldDef(channelDef)) {
     const fd = replaceRepeaterInFieldDef(channelDef, repeater);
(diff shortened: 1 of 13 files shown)


sorenmacbeth:
👍

src/compile/repeater.ts
```diff
@@ -20,40 +30,70 @@ export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: R
...
   if (isRepeatRef(field)) {
     if (field.repeat in repeater) {
-      return {
+      fieldDef = {
         ...fieldDef,
         field: repeater[field.repeat]
       };
     } else {
       log.warn(log.message.noSuchRepeatedValue(field.repeat));
+      // return null, meaning that the field def should be ignored
```
kanitw: Normally we use undefined for the rest of the code base.

src/compile/repeater.ts
```diff
@@ -20,40 +30,70 @@ export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: R
 /**
  * Replace repeater values in a field def with the concrete field name.
  */
-function replaceRepeaterInFieldDef(fieldDef: FieldDef<Field>, repeater: RepeaterValue): FieldDef<string> | null {
+function replaceRepeaterInFieldDef(fieldDef: LegendFieldDef<Field>, repeater: RepeaterValue): LegendFieldDef<string> | null {
```
kanitw: Why is this a LegendFieldDef?  PositionFieldDef can be sorted too.  This is more likely `ScaleFieldDef`?

src/compile/repeater.ts
```diff
@@ -20,40 +30,70 @@ export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: R
...
+      };
+    } else {
+      log.warn(log.message.noSuchRepeatedValue(sort.field.repeat));
+      // remove sort as it uses unknown field
+      const {sort: _s, ...newFieldDef} = fieldDef;
+      fieldDef = newFieldDef;
+    }
+  }
+
+  return fieldDef as LegendFieldDef<string>;
```
kanitw: Do we really need casting here, especially if this become ScaleFieldDef.

src/compile/repeater.ts
```diff
@@ -20,40 +30,70 @@ export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: R
...
       return null;
     }
-  } else {
-    // field is not a repeat ref so we can just return the field def
-    return fieldDef as FieldDef<string>;
   }
+
+  const sort = fieldDef.sort;
+  if (sort && isSortField(sort) && isRepeatRef(sort.field)) {
+    if (sort.field.repeat in repeater) {
```
kanitw: This part is redundant with above, we should extract a more general `replaceRepeat` method that takes any object with a property that can be a `RepeatRef`. 

Something like
```
function replaceRepeat<T>(o: T, prop: keyof T, repeater: Repeater) {...}
```
may work (I haven't tried myself).

src/compile/repeater.ts
```diff
@@ -20,40 +30,70 @@ export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: R
...
 type EncodingOrFacet<F> = Encoding<F> | Facet<F>;
 
+type FieldDefTypes<F> = LegendFieldDef<F> | FieldDef<F>[] | ConditionalFieldDef<FieldDef<F>> | ConditionalValueDef<FieldDef<F>>;
+
 function replaceRepeater(mapping: EncodingOrFacet<Field>, repeater: RepeaterValue): EncodingOrFacet<string> {
   const out: EncodingOrFacet<string> = {};
   for (const channel in mapping) {
     if (mapping.hasOwnProperty(channel)) {
-      const fieldDef: FieldDef<Field> | FieldDef<Field>[] = mapping[channel];
+      const fieldDef: FieldDefTypes<Field> = mapping[channel];
```
kanitw: This is a channelDef, not always fieldDef.

src/compile/repeater.ts
```diff
@@ -44,16 +73,26 @@ function replaceRepeater(mapping: EncodingOrFacet<Field>, repeater: RepeaterValu
...
+        out[channel] = channelDef.map(fd => replaceRepeaterInFieldDef(fd, repeater))
+          .filter(fd => fd !== null);
+      } else if (isConditionalDef(channelDef) && isFieldDef(channelDef.condition)) {
+        const fd = replaceRepeaterInFieldDef(channelDef.condition, repeater);
+        if (fd !== null) {
+          out[channel] = {
+            ...channelDef,
+            condition: fd
+          };
+        }
```
kanitw: In the else case, don't we have to at least do
```
out[channel] = channelDef;
```
? 

(Drop only the condition, not the whole thing)

domoritz: We could, I guess. Since there is a value, right?

kanitw: Yes, I think we definite should.

src/sort.ts
```diff
@@ -3,13 +3,13 @@ import {AggregateOp} from './aggregate';
 
 export type SortOrder = 'ascending' | 'descending' | null;
 
-export interface SortField {
+export interface SortField<Field> {
```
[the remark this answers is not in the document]
domoritz: Yes

src/spec.ts
```diff
@@ -110,17 +110,17 @@ export interface GenericUnitSpec<E extends Encoding<any>, M> extends BaseSpec, L
   selection?: {[name: string]: SelectionDef};
 }
 
-export type UnitSpec = GenericUnitSpec<Encoding<Field>, Mark | MarkDef>;
+export type UnitSpec = GenericUnitSpec<Encoding<string | RepeatRef>, Mark | MarkDef>;
```
domoritz: Why not leave it as `Field` and have a link to one place where we explain what Field is? I think we should completely remove `Field` if you replace it here.

kanitw: We never use type as a union type in the documentation elsewhere. 

 It's quite a jump for readers to comprehend IMHO. 

This is the old docs, which makes the `Field` not really match the text.  

pasted_image_9_13_17__10_26_am

>  I think we should completely remove Field if you replace it here.

We still do the same in many other places. (Use the full union type in the schema, but keep the shorthand for references in the code.)

src/compile/repeater.ts
domoritz: Done

src/compile/repeater.ts
domoritz: Done

src/compile/repeater.ts
domoritz: Yes, because we start with `ScaleFieldDef<Field>` and end with `ScaleFieldDef<string>`

src/compile/repeater.ts
domoritz: Done

domoritz:
Pushed revisions

src/compile/repeater.ts
```diff
@@ -44,16 +63,26 @@ function replaceRepeater(mapping: EncodingOrFacet<Field>, repeater: RepeaterValu
   const out: EncodingOrFacet<string> = {};
   for (const channel in mapping) {
     if (mapping.hasOwnProperty(channel)) {
-      const fieldDef: FieldDef<Field> | FieldDef<Field>[] = mapping[channel];
+      const channelDef: ChannelDef<Field> = mapping[channel];
```
kanitw: Hmm, `detail` and `order` can be an array of fieldDef too.

src/compile/repeater.ts
kanitw: Is this case handled somehow? (Doesn't look like it?)

src/compile/repeater.ts
```diff
@@ -18,24 +20,41 @@ export function replaceRepeaterInEncoding(encoding: Encoding<Field>, repeater: R
...
-  if (isRepeatRef(field)) {
-    if (field.repeat in repeater) {
-      return {
-        ...fieldDef,
-        field: repeater[field.repeat]
-      };
+function replaceRepeat<T extends {field?: Field}>(o: T, repeater: RepeaterValue): T {
+  if (isRepeatRef(o.field)) {
+    if (o.field.repeat in repeater) {
+      // any needed to calm down ts compiler
```
kanitw: lol

src/compile/repeater.ts
domoritz: Done. There was also a missing case for reversed conditionals.

src/compile/repeater.ts
domoritz: Can they be conditionals?

src/compile/repeater.ts
domoritz: Done

domoritz:
Revisions sent

src/compile/repeater.ts
```diff
@@ -57,32 +57,50 @@ function replaceRepeaterInFieldDef(fieldDef: ScaleFieldDef<Field>, repeater: Rep
...
+function replaceRepeaterInChannelDef(channelDef: ChannelDef<Field>, repeater: RepeaterValue): ChannelDef<string> {
+  if (isConditionalDef(channelDef) && isFieldDef(channelDef.condition)) {
+    const fd = replaceRepeaterInFieldDef(channelDef.condition, repeater);
+    if (fd) {
+      return {
+        ...channelDef,
+        condition: fd
+      } as ChannelDef<string>;
+    } else {
+      return {
```
kanitw: better do
```
const {condition, ...channelDefWithoutCondition} = channelDef;
```

and 

`return channelDefWithoutCondition` 

as this will (1) not require `as` and (2) be more sustainable in long term should we add `datum` or additional property to a value def.

src/compile/repeater.ts
```diff
@@ -57,32 +57,50 @@ function replaceRepeaterInFieldDef(fieldDef: ScaleFieldDef<Field>, repeater: Rep
...
+    const fd = replaceRepeaterInFieldDef(channelDef, repeater);
+    if (fd) {
+      return fd;
+    } else if (isConditionalDef(channelDef)) {
+      return {value: channelDef.condition.value};
+    } else {
+      return undefined;
+    }
+  } else {
+    return channelDef as ValueDef;
```
kanitw: I think you do not need `as` if you organize the `if/else` like this:

```
if (isFieldDef(channelDef)) {
   ...
} else { // Definitely ValueDef
  if (isConditionalDef(channelDef) && isFieldDef(channelDef.condition)) {
    ...
  }
  return channelDef;
}

File: src/compile/repeater.ts (hunk 11 above)
Comment (domoritz) [reply src/compile/repeater.ts]: Lol, I had exactly that and changed it back because I wanted to make it explicit that the second thing has to be a value.

File: src/compile/repeater.ts (hunk 12 above)
Comment (domoritz) [reply src/compile/repeater.ts]: That would add additional runtime checks, wouldn't it?

Comment (kanitw) [reply src/compile/repeater.ts]: How?  I think in general, try to separate a branch for a field def and a value def makes it easier to read.
The currently logic here as value def in two separate branches.

File: src/compile/repeater.ts [hunk 13]
```diff
@@ -80,6 +79,7 @@ function replaceRepeaterInChannelDef(channelDef: ChannelDef<Field>, repeater: Re
       return undefined;
     }
   } else {
+    // Definitely ValueDef
     return channelDef as ValueDef;
```
kanitw: See https://github.com/vega/vega-lite/pull/3006/commits/f6e83df5c3e4a24b2f5e0811990c8584a8402d87#r142015425

kanitw:
Fix the remaining comment & make test pass?

src/compile/repeater.ts
domoritz: I see. Will update.

domoritz: https://github.com/vega/vega-lite/pull/3006/commits/1815cec3f3d4dce93251fc2fbe5e6be183222dc1

Issue #3004, opened 2017-09-13
Nearest with lines


domoritz:
I wanted to highlight the nearest line in a multi series line chart. Is this supposed to work?

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {"url": "data/stocks.csv"},
  "mark": "line",
  "selection": {
    "line": {"type": "single", "on": "mouseover", "nearest": true}
  },
  "encoding": {
    "x": {"field": "date", "type": "temporal", "axis": {"format": "%Y"}},
    "y": {"field": "price", "type": "quantitative"},
    "color": {"field": "symbol", "type": "nominal"},
    "opacity": {
      "condition": {"selection": "line", "value": 1},
      "value": 0.2
    }
  }
}
```

amitkaps:
My understanding is that nearest only works with discrete points and not lines...  Below is how you get this to work currently. It is there in the community examples.

https://bl.ocks.org/amitkaps/fe4238e716db53930b2f1a70d3401701

arvind:
A couple of things are happening here. First, Vega's line marks vary from the others in that properties besides `x` and `y` do not have an effect on individual points along the line (but are rather based on the values set on the first scene graph item). So, in your above example, the selection is being populated with a particular `id` for the nearest point, and the corresponding opacity in the scenegraph is being set, but has no effect on the line overall. Instead, if you project your selection over the `symbol` field, you should see an effect. 

The second issue, however, is related to what @amitkaps describes. Right now, the voronoi diagram is calculated via the reactive geometry. In the case of multi series line charts, however, the line mark is nested within a faceted group causing the voronoi too as well. What this means is that a voronoi diagram is calculated for each individual line, and they get layered one on top of the other. 

As far as I can tell, there isn't an easy fix for this -- if we were to lift the voronoi transform out of the faceted group, we would need an additional derived dataset to calculate its input `x` and `y` coordinates. @amitkaps' example is a good workaround in the interim, and I will add it to the documentation.

domoritz:
> As far as I can tell, there isn't an easy fix for this

Is there an easy way to detect this case so that we can show a warning?

PR #3009, opened 2017-09-13
Nearest selection transform does not support continuous mark types.

arvind:
Addresses issues raised in #3004 which identified a bug the `nearest` transform and continuous mark types (i.e., `line` and `area`). For now, this PR drops support for these mark types (which has the added effect of simplifying code paths) and the compiler issues a warning. This limitation is also noted in the documentation, with a workaround shown in an example.


site/docs/selection/nearest.md
```diff
@@ -16,3 +16,7 @@ In the scatterplot below, points <select onchange="changeSpec('paintbrush_neares
 The `nearest` transform also respects any [spatial encoding projections](project.html) applied to the selection. For instance, in the example below, moving the mouse cursor back-and-forth snaps the vertical rule and label to the nearest `date` value.
 
 <div id="paintbrush_nearest" class="vl-example" data-name="stocks_nearest_index"></div>
+
+## Current Limitations
+
+* The `nearest` transform is not supported for continuous mark types (i.e., `line` and `area`). For these mark types, consider layering a discrete mark type (e.g., `point`) with a 0-value `opacity` as in the last example above.
```
domoritz: The example above doesn't use 0 opacity. How about "consider layering a discrete mark type (e.g., `point`) as in the last example above but use a 0-value `opacity`."

arvind: Not sure what you mean -- [here](https://github.com/vega/vega-lite/pull/3009/files#diff-41b1e96b395ea6629b4c649c14778f7bR28) is the 0-value `opacity` on the `point` mark.

domoritz:
Nearest is also behaving a little funny with bars:

```
{
  "selection": {
    "paintbrush": {
      "type": "multi",
      "on": "mouseover",
      "nearest": true
    }
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "Horsepower", "type": "quantitative", "bin": true},
    "y": {"aggregate": "mean", "field": "Miles_per_Gallon", "type": "quantitative"},
    "color": {
      "condition": {
        "selection": "paintbrush",
        "field": "Cylinders", "type": "nominal"
      },
      "value": "grey"
    }
  }
}
```

I understand that the voronoi tesselation is using the corners but it's still a bit odd to use. 

Regardless of this PR, I wonder whether the definition of nearest could be different when the domain in ordinal. Think of a normal bar chart. If I want to use it for cross-highlighting, I could have to hover over a bar to add it to the selection. This can be hard for bars that are small. Instead, a better method is to use the background as the hit target (I did this in https://www.domoritz.de/papers/2017-TrustButVerify-CHI.pdf). Nearest could take on this semantic in this case. Just a thought, no action item.

site/docs/selection/nearest.md
domoritz: Oh, I missed that update. 👼

domoritz: It's also not quite the right semantics but the best we can do. What I mean is that the closest line may not be the one that has the closest control point.

PR #3027, opened 2017-09-21
Next button on homepage. Fixes #2986

### LINKED ISSUE #2986: Test whether people find how to look at the different examples on the homepage.

domoritz:
screen shot 2017-09-21 at 10 20 59

### DIFF
--- a/site/index.md
+++ b/site/index.md
@@ -47,7 +47,9 @@ Read our [introduction article to Vega-Lite v1 on Medium](https://medium.com/p/4
 ## Example
 
 <div id="carousel" class="carousel">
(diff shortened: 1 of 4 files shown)


arvind:
I'm not sure I understand the motivation for an explicit next step button? The links provide the same functionality, but with more information scent (i.e., you know what the next step is).

domoritz:
@arvind I've shown the homepage to a few friends and noticed that they didn't see that you can interact with the links or indicators. I'm open to other suggestions that make it more explicit that you can switch the example (e.g. autoplay).

domoritz:
> Blue link color is pretty standard though, so I don't have a better suggestion for now.

Maybe links don't have the same affordance as a button. Link = I can go somewhere else, button = I can do something. Having said that, we are breaking that with the two big buttons to the tutorial and editor on the homepage. 

We can also remove the indicators as they are redundant with the links.

domoritz:
Also, note discussion in #2996 with @kanitw

arvind:
> We can also remove the indicators as they are redundant with the links.

My vote would be to keep them. They have a small footprint, and provide guidance for how many steps there are.

PR #3029, opened 2017-09-22
Read autosize from config. Fixes #3025

### LINKED ISSUE #3025: Is autosize really in config?
https://vega.github.io/vega-lite/docs/config.html#top-level-config

cc @znation

domoritz:


src/toplevelprops.ts
```diff
@@ -57,25 +57,25 @@ export interface AutoSizeParams {
...
-    return {
-      type: 'pad'
-    };
-  }
-  const autoSizeParams: AutoSizeParams = {
+function normalizeAutoSize(autosize: AutosizeType | AutoSizeParams) {
+  return isString(autosize) ? {type: autosize} : autosize || {};
+}
+
+export function normalizeAutoSizes(topLevelAutosize: AutosizeType | AutoSizeParams, configAutosize: AutosizeType | AutoSizeParams, isUnitOrLayer: boolean = true): AutoSizeParams {
```
kanitw: The plural form of `normalizeAutoSizes` reads a bit funny to me, but I guess that's fine.

kanitw: Wonder if we should rename this to `normalizeAutoSize` (since this one is a public API, its name is more important) and rename the other to `normalizeSingleAutoSize` or `_normalizeAutoSize` or something.

domoritz: Done

PR #3032, opened 2017-09-22
Dom/lookup docs

domoritz:


site/docs/transform/lookup.md
```diff
@@ -0,0 +1,35 @@
...
+  ],
+  ...
+}
+```
+
+For each data object in the main data source, the transform tries to find a matching objects in the secondary data source. An object matches if the value in the field specified by `lookup` is the same as the field specified in the `from.key`.
+
+{% include table.html props="lookup,from,as,default" source="LookupTransform" %}
+
+The secondary data reference (set with `from`) is an object that specifies how the lookup key should be matched to a second data source and what fields should be added.
```
kanitw: Missing an anchor (and link from "LookUpData" in the table).

domoritz:
Done

domoritz:
Fixed the build failure.

domoritz:
Rebased

PR #3042, opened 2017-09-27
Replace a lot of any with VgSignal, add log message to log.ts

domoritz:


src/compile/compile.ts
```diff
@@ -82,7 +82,7 @@ function assembleTopLevelModel(model: Model, topLevelProperties: TopLevelPropert
   // move width and height signals with values to top level
   layoutSignals = layoutSignals.filter(signal => {
     if ((signal.name === 'width' || signal.name === 'height') && signal.value !== undefined) {
-      topLevelProperties[signal.name] = signal.value;
+      topLevelProperties[signal.name] = +signal.value;
```
kanitw: Shoudn't signal values already be a number?

src/vega.schema.ts
```diff
@@ -59,32 +59,32 @@ export function isVgSignalRef(o: any): o is VgSignalRef  {
...
+  sort?: VgUnionSortField;
+}
 
 // TODO: add vg prefix
-export type FieldRefUnionDomain = {
-  data: string,
-  fields: VgFieldRef[],
-  sort?: VgUnionSortField
-};
+export interface FieldRefUnionDomain {
```
kanitw: Just do the Vg prefix now ?

src/compile/compile.ts
domoritz: Can't they be strings or bools?

kanitw: By type, yes.  By how we initialize width and height signals, should be no (unless we have a bug).

src/vega.schema.ts
domoritz: Done

src/compile/compile.ts
domoritz: Right, this is just a type thing. I guess we could say `as number`

PR #3043, opened 2017-09-27
Add config parameter to compile function. Fixes #3028

### LINKED ISSUE #3028: compile should have config parameter
Should it be the first or second (after logger)? What's the precedence?

domoritz:


kanitw:
This is okay, but I still think that making it `{config, logger}` would prevent breaking changes in the future.  (We are creating a breaking change now anyway, better just do it once and for all.)

domoritz:
I prefer it the way it is in this PR and think it will be good for a while.

kanitw:
>  it will be good for a while.

I _strongly_ don't like this part.  If we can make it work forever without hurting much, let's just do it so we don't regret later.  For example, if we want to provide title formatter function, we already have to extend it here.

domoritz:
Okay, I don't feel strong enough about it and changed it.

domoritz:
Done

site/usage/compile.md
```diff
@@ -20,10 +20,16 @@ If you want access to the compiled Vega spec from a Javascript program, you can
 var vgSpec = vl.compile(vlSpec).spec;
 ```
 
-By default, warnings and errors are printed to the JavaScript console. To customize how errors and warnings are handled, you can pass a customize logger to the compile function. 
+### Customized Configuration
```
kanitw: It isn't very clear that compile has  two parameters the second one is `opt` and Opt can have config/logging.  

https://vega.github.io/vega/docs/api/view/#view is one way to do it.

domoritz: You're right. I improved it.

Issue #1486, opened 2016-08-02
Replicate Vega examples that can be done now, with ranged mark and layering


kanitw:
For example
- http://localhost:1234/?mode=vega&spec=weather

domoritz:
https://vega.github.io/vega-editor/?mode=vega&spec=weather

domoritz:
- [ ] Add examples from the infovis paper.

domoritz:
- [ ] https://uber.github.io/react-vis/#/examples/charts/showcase has some interesting examples that we can do as well.

domoritz:
- [ ] https://app.datawrapper.de/chart/Q9Lfq/visualize

PR #3049, opened 2017-09-29
Add Diverging Stacked Bar Chart Example

haldenl:
Add an example showing vega-lite implementation of a diverging stacked bar chart, centered along a 0 axis.

Part of issue #1486


package.json
```diff
@@ -127,6 +121,7 @@
   "dependencies": {
     "json-stable-stringify": "^1.0.1",
     "tslib": "^1.7.1",
+    "typescript-compiler": "^1.4.1-2",
```
willium: Why do we need this dependency?

examples/specs/bar_diverging_stacked.vl.json
```diff
@@ -0,0 +1,376 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "A diverging stacked bar chart for sentiments towards a set of eight questions, displayed as percentages with neutral responses straddling the 0% mark",
+  "data": {
+    "values":
```
willium: Should we be inlining the data in this example? Is there an existing dataset we can use to show diverging stacked bars?

haldenl: The data is specific to this example.

package.json
haldenl: Oh hm, this (and all the newline removals) might have been an artifact of my workspace? Thanks for pointing this out! @kanitw would you have any ideas?

willium: Try this:
1. `git reset --soft HEAD^`
2. re-add/commit only the relevant files
3. git push origin +hl/example_diverging_stacked_bar (force push, be careful)

examples/specs/bar_diverging_stacked.vl.json
willium: Okay! @domoritz perhaps we should factor out and add to https://github.com/vega/vega-datasets. It's kind of verbose to be included in the spec like this IMO

kanitw: Extracting the dataset is a good idea, but we should not do that yet in this PR since this example currently include manually derived fields that would be possible once @AkshatSh add supports for window transform.  (At that point, it is okay to add raw data to the vega-dataset repo.)

package.json
haldenl: Perfect, should be good to go. Thanks Will!

examples/specs/bar_diverging_stacked.vl.json
domoritz: Agreed with Ham.

kanitw:
Nice job.  Thanks!

PR #3053, opened 2017-09-30
Automatically convert constant value to valueDef  (e.g., `x:5` to `x:…

kanitw:
… {value: 5}`)

I think I mistakenly forgot to wrap constant with `{value: ...}` a few times already.  This would be helpful to users to inform them about the right syntax. 

We may revise if we wanna include this in the schema in the future, but I'm not sure if we should do that yet, but I'm pretty sure that having some warning & conversion is helpful.

### DIFF
--- a/src/fielddef.ts
+++ b/src/fielddef.ts
@@ -1,5 +1,6 @@
 // utility for a field definition object
 
+import {isNumber} from 'vega-util';
 import {AggregateOp, isAggregateOp, isCountingAggregateOp} from './aggregate';
 import {Axis} from './axis';
 import {autoMaxBins, BinParams, binToString} from './bin';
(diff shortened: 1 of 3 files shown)


src/fielddef.ts
```diff
@@ -374,6 +375,12 @@ export function getFieldDef<F>(channelDef: ChannelDef<F>): FieldDef<F> {
  * Convert type to full, lowercase type, or augment the fieldDef with a default type if missing.
  */
 export function normalize(channelDef: ChannelDef<string>, channel: Channel): ChannelDef<any> {
+  if (isString(channelDef) || isNumber(channelDef)) {
+    const type = isString(channelDef) ? 'string' : 'number';
```
domoritz: What about booleans?

kanitw: Not sure if boolean makes sense for any channel.  But if you think I should add I can add. Let me know.

domoritz:
I think we should support it.

kanitw:
Done

domoritz:
Fix tests

kanitw:
@domoritz  done -- merge?

PR #3056, opened 2017-10-01
Add example of a diverging stacked bar chart

cjiang:
This implements one of the suggestions in #1486


kanitw:
@cjiang -- Thanks for submitting.   However, @haldenl just beat you to that. 

Note that I also think that the https://github.com/vega/vega-lite/pull/3049 is a better example since it is clear which value is negative, neutral or positive.  (While for weather, the positivity/negativity may be subjective.)

PR #3060, opened 2017-10-02
Make some spec fields possibly-null in the typings

pelotom:


kanitw:
We already use @nullable to annotate nullable properties.  

I think the problem with using `Axis | null` is that that it doesn't work with our JSON schema generator.  (Not 100% sure, but @domoritz would know more about this since he wrote the JSON schema generator.)  

In any case, you might want to re-run `npm run schema` and push the schema just in case I'm wrong.

pelotom:
Hm, ok. Well right now the TypeScript types are wrong, it would be nice if both could be correct.

domoritz:
@pelotom They are not wrong since we did not enable strict null checks. It would be good to enable that option someday soon.

pelotom:
Unfortunately that decision means consumers of your library can’t use strict null checks either, or more likely they will have to use casts to circumvent the typing.

domoritz:
Yes, we are aware of this. Unfortunately, we had to focus on other issues first before turning on strict null checks. We hope to get to it post 2.0.

domoritz:
@pelotom I think for now we will have to fix the ones where we notice that `| null` is missing but the right long-term solution is to enable it in the compiler.

pelotom:
@domoritz I'm not sure what you're suggesting for next steps... does that mean this PR or some version of it can be merged?

domoritz:
Yes, we will merge after a review. We are currently all out of town so
reviews are a bit stalled until next week.

kanitw:
In any case, I wonder why the test is failing.

domoritz:
@pelotom You need to rebuild and commit the schema.

domoritz:
Merged.

PR #3065, opened 2017-10-04
Add Ranged Dot Plot example to example specs and gallery

haldenl:
Added an example ranged dot plot that uses 'layer' to convey changing life expectancy for the five most populous countries (between 1955 and 2000). Part of issue #1486

Result shown below:

Ranged Dot Plot Example


kanitw:
Very well done!  @haldenl

examples/specs/layer_ranged_dot.vl.json
```diff
@@ -0,0 +1,85 @@
...
+          "encoding": {
+              "x": {
+                  "field": "life_expect",
+                  "type": "quantitative"
+              },
+              "y": {
+                  "field": "country",
+                  "type": "nominal"
+              },
+              "opacity": {
```
kanitw: Not sure why you are mapping a field here (while setting range to be constant). 

I think you actually wanna use `{value: 0.2}` ?

examples/specs/layer_ranged_dot.vl.json
```diff
@@ -0,0 +1,85 @@
...
+                      "#911a24"]
+                  },
+                  "legend": {
+                      "title": "Year"
+                  }
+              }
+          }
+      },
+      {
+          "mark": "line",
```
kanitw: since this is only a line between two dots, maybe using rule is slightly better

examples/specs/layer_ranged_dot.vl.json
```diff
@@ -0,0 +1,85 @@
...
+                  "scale": {
+                      "range": [0.2, 0.2]
+                  },
+                  "legend": null
+              }
+          }
+      }
+  ],
+  "config": {
+      "point": {
```
kanitw: Since there is only one layer of point and line, perhaps using
`mark:{ type: 'point', filled: true}`
and `size: {value: 100}` and `color: {"value": "#db646f"}` are better.

examples/specs/layer_ranged_dot.vl.json
kanitw: actually line makes more sense for this case

examples/specs/layer_ranged_dot.vl.json
haldenl: I used 'field' as a workaround to group points by 'country' such that the lines would be drawn between points of the same country.

Using just 'value', I get the following:
```              
"opacity": {
    "value": 0.2
}
```
screen shot 2017-10-06 at 11 16 45 am

Maybe there is a different workaround?

kanitw: ah I see, the right way to separate line would be using `detail` channel. 

However, a better alternative would be using `rule` (line segment) instead .

PR #3068, opened 2017-10-05
Horizon graph example

kanitw:


domoritz:
Although it's not great that the number of layers is not data-driven.

PR #3072, opened 2017-10-10
[WIP] Add Axis's `labelBound` and `labelFlush`

### LINKED ISSUE #2880: x-axis labels can go beyond the axis bound
In https://github.com/vega/vega-lite/issues/2825, we found a similar issue is problematic for column layout. It seems like this is actually also problematic for legends.

```
{
  "data": {
    "url": "node_modules/vega-datasets/data/population.json"
  },
  "transform": [],
  "mark": "bar",
  "encoding": {
    "color": {
      "field": "sex",
      "type": "nominal"
    },
    "x": {
      "aggregate": "sum",
      "field": "people",
      "type": "quantitative"
    }
  }
}
```

image

kanitw:
Also set default labelBound to `21` (I just need some nice number that's higher than 10 so I set to 21 just like rangeStep)

Fix #2880


kanitw:
Using 21 doesn't work well.  I think making both `labelBound` and `labelFlush` true by default is slightly better (although not perfect yet).

PR #3080, opened 2017-10-11
Make legend have non-0 value when having condition

### LINKED ISSUE #3010: Legend marks when 0-value opacity
In the following spec, the legend marks disappear because the second unit spec has an `opacity` of `0`. 

```json
{
  "width": 600,
  "height": 300,
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "x": {"field": "date", "type": "temporal", "axis": null},
      }
    },
    {
      "mark": "point",
      "encoding": {
        "x": {"field": "date", "type": "temporal", "axis": null},
        "color": {"field": "symbol", "type": "nominal"},
        "opacity": {"value": 0}
      }
    }
  ]
}
```

In the following spec, the opacity is set interactively but still produces 0-value legend mark opacities:

```json
{
  "width": 600,
  "height": 300,
  "selection": {"brush": {"type": "interval"}},
  "mark": "point",
  "encoding": {
    "x": {"field": "date","type": "temporal","axis": null},
    "y": {"field": "price","type": "quantitative"},
    "color": {"field": "symbol","type": "nominal"},
    "opacity": {
      "value": 0,
      "condition": {"selection": "brush","value": 1}
    }
  }
}
```

Besides these examples, perhaps we should always set legend opacity to some non-0 value? Are there cases where an invisible legend mark is desirable?

yhoonkim:
Fix #3010

### DIFF
--- a/src/compile/legend/encode.ts
+++ b/src/compile/legend/encode.ts
@@ -1,5 +1,5 @@
 import {Channel, COLOR, NonPositionScaleChannel, OPACITY, SHAPE} from '../../channel';
-import {FieldDef, hasConditionValueDef, isTimeFieldDef, isValueDef} from '../../fielddef';
+import {ConditionalFieldDef, ConditionalValueDef, FieldDef, hasConditionValueDef, isTimeFieldDef, isValueDef, LegendFieldDef} from '../../fielddef';
 import {AREA, BAR, CIRCLE, FILL_STROKE_CONFIG, LINE, POINT, SQUARE, TEXT, TICK} from '../../mark';
 import {ScaleType} from '../../scale';
 import {keys, without} from '../../util';
@@ -66,13 +66,9 @@ export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: Uni
   }
 
   if (channel !== OPACITY) {
-    const opacityDef = model.encoding.opacity;
-    if (isValueDef(opacityDef)) {
-      if (hasConditionValueDef(opacityDef)) {
-        symbols.opacity = {value: Math.max(opacityDef.condition.value as number, opacityDef.value as number)};
-      } else {
-        symbols.opacity = {value: opacityDef.value};
-      }
+    const opacity = getOpacityValue(model.encoding.opacity);
+    if (opacity) { // only apply opacity if it is zero or undefined
+      symbols.opacity = {value: opacity};
     }
   }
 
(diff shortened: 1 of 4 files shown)


kanitw:
Could you revert your changes to package.json and remove package-lock.json?  
(I believe you use `npm` rather than `yarn` and thus have created these unnecessary files.)

src/compile/legend/encode.ts
```diff
@@ -68,7 +68,11 @@ export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: Uni
   if (channel !== OPACITY) {
     const opacityDef = model.encoding.opacity;
     if (isValueDef(opacityDef)) {
-      symbols.opacity = {value: opacityDef.value};
+      if (hasConditionValueDef(opacityDef)) {
```
kanitw: 1) Please extract method `getOpacityValue` so we can reuse below. 

2) How should we deal with `{condition: { value: ... }, field: ..., ...}` and `{condition: { field: ... }, value: ..., ...}`?  I think two of these should be consistent but given the current code, the latter will get fixed opacity. 
  - If you deal with the case where condition is a fielddef too, maybe you do not need to add the new `hasConditionValueDef`.

test/compile/legend/encode.test.ts
```diff
@@ -43,6 +43,20 @@ describe('compile/legend', function() {
...
+
+    it('should return the maximum value when there is a condition', function() {
+
+      const symbol = encode.symbols({field: 'a', type: 'nominal'}, {}, parseUnitModelWithScale({
+          mark: "point",
+          encoding: {
+            x: {field: "a", type: "nominal"},
+            opacity: {
+              value: 0,
+              condition: {selection: "brush", value: 1}
```
kanitw: Please put condition above value (so it reads more like a typical `if/else`)

src/compile/legend/encode.ts
```diff
@@ -66,9 +66,9 @@ export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: Uni
   }
 
   if (channel !== OPACITY) {
-    const opacityDef = model.encoding.opacity;
-    if (isValueDef(opacityDef)) {
-      symbols.opacity = {value: opacityDef.value};
+    const opacity = getOpacityValue(model.encoding.opacity);
+    if (opacity) { // only apply opacity if it is zero or undefined
```
kanitw: "if it is neither zero or undefined"

I think this is a typo that I made earlier :P

src/compile/legend/encode.ts
```diff
@@ -81,9 +81,9 @@ export function gradient(fieldDef: FieldDef<string>, gradientSpec: any, model: U
   let gradient:any = {};
 
   if (type === 'gradient') {
-    const opacityDef = model.encoding.opacity;
-    if (isValueDef(opacityDef)) {
-      gradient.opacity = {value: opacityDef.value};
+    const opacity = getOpacityValue(model.encoding.opacity);
+    if (opacity) { // only apply opacity if it is zero or undefined
```
kanitw: "if it is neither zero or undefined"

I think this is a typo that I made earlier :P

src/compile/legend/encode.ts
```diff
@@ -112,3 +112,13 @@ export function labels(fieldDef: FieldDef<string>, labelsSpec: any, model: UnitM
   return keys(labels).length > 0 ? labels : undefined;
 }
 
+function getOpacityValue(opacityDef: ConditionalFieldDef<LegendFieldDef<string>> | ConditionalValueDef<LegendFieldDef<string>>): number {
+  if (isValueDef(opacityDef)) {
+    if (hasConditionValueDef(opacityDef)) {
```
kanitw: wonder if we can just do 
```
if (opacityDef.condition && isValueDef(opacityDef))
```

so we do not have to increase the number of type guards

Issue #3077, opened 2017-10-10
When possible, make `resolve` work with `vconcat`


mstone:
I'm making an interactive timeline of two related but independent timeseries via `vconcat` of two independent views but I'd like both views to share a common x-axis scale, labels, etc.

Question 1: is `resolve` intended to be the right tool to unify these scales?

Question 2: if so, is it supposed to work, given a spec like this one:

```
var spec = {
	"$schema": "https://vega.github.io/schema/vega-lite/v2.json",
	"description": "Timeline",
	"height": 30,
	"config": {
		"padding": 0,
	},
	"data": {
		"values": values,
	},
	"resolve": {
		"x": {
			"scale": "shared",
			"axis": "shared",
		},
	},
	"vconcat": [{
		"layer": [{
			"selection": {
				"index": {
					"type": "single",
					"on": "mousemove",
					"encodings": ["x"],
					"nearest": true
				}
			},
			"mark": "rule",
			"encoding": {
				"x": {
					"field": "date",
					"type": "temporal",
					"axis": {
						"grid": false,
						"title": "",
						"labelAngle": 0,
						"format": "%b %e, %H:%M  ",
						"tickCount": 8,
					}
				},
				"color": {
					"value": "rgba(0,0,180,0.3)"
				},
				"size": {
					"value": 1
				}
			}
		}, {
			"transform": [{
				"filter": {
					"and": ["index.date", {
						"selection": "index"
					}]
				}
			}],
			"mark": "rule",
			"encoding": {
				"x": {
					"field": "date",
					"type": "temporal",
					"axis": null
				},
				"color": {
					"value": "blue"
				},
				"size": {
					"value": 2
				},
			}
		}, {
			"data": {
				"name": "currentMsg",
			},
			"mark": {
				"type": "rule",
			},
			"encoding": {
				"x": {
					"field": "date",
					"type": "temporal",
					"axis": null
				},
				"color": {
					"value": "rgba(0,180,0,0.3)"
				},
				"size": {
					"value": 3
				}
			}
		}]
	}, {
		"data": {
			"values": cases,
		},
		"mark": "rule",
		"encoding": {
			"x": {
				"field": "date",
				"type": "temporal",
				"axis": {
					"grid": false,
					"title": "",
					"labelAngle": 0,
					"format": "%b %e, %H:%M  ",
					"tickCount": 8,
				}
			},
			"color": {
				"value": "rgba(180,0,0,0.3)"
			},
			"size": {
				"value": 1
[...]

PR #3082, opened 2017-10-11
Clarify resolve docs. #3077

domoritz:


kanitw:
LGTM.  I just correct positional channels to position channels (either is fine but we use "position channels" elsewhere in the docs.  

Maybe add warnings for the logger too? (So users can notice without needing to look at the document.)

domoritz:
In https://github.com/vega/vega-lite/blob/master/site/docs/composition/resolve.md we say positional.

kanitw:
Then let's change all occurrences of positional to position.
The reason is that we have "Mark Property Channel" / etc. which are compound nouns too, so "Position Channels" is more consistent.

domoritz:
It still sounds odd to me but I'm not a native speaker either. What does
@jheer use in the Vega docs?

kanitw:
In Vega, there is no distinction between them since they all have the same types.

domoritz:
I think "position channel" is already taken for the x and y location. "Positional" imo includes length. 

Cleveland and McGill

image

PR #3091, opened 2017-10-12
Support setting field title formatter

kanitw:

Issue #1977, opened 2017-02-24
Title formatter


kanitw:
Many datasets have `field` name as lower case strings with underscore or dashes (e.g., "curb-weight").

Oftentimes, users would want the title to nicely have title case e.g., (`"Curb Weight"`). 
It's probably nice if we have titleFormat support in the config.

danyelf:
Adding to this:  I might want to title my axis "Sum of Area", rather then SUM(area)." Can you give me the ability to provide a raw "Title:" field and have it respected? (That might also address your underlying issue.)

kanitw:
@danyelf Thanks.  I think Sum of Area is really nicer than `SUM(area)`. 

I think it's even a better default, but that will be a breaking change though.

In any case, maybe we should support title as either `string` or an object describing title parameter:

```
{
  text: string;
  field: string; // This can be the text for just the field part 
  style: 'parenthesis' | 'of';
}
```

@domoritz @arvind @jheer  Any thoughts? 

(This is only for VL.)

domoritz:
I'd say we can support titles with variable such as `Sum of {}` or something similar. I also like `Sum of X` better than `Sum(X)` and would be supportive of breaking it now rather than for v3. I do not like the object idea above. There will always be some other alternative that people will ask for.

kanitw:
> I do not like the object idea above

How do you specify that you want `SUM()` then?   For example, in Voyager, we would want the old style. 

Another possibility is to have `config.guide.title.style: 'parenthesis' | 'of'`; (or `config.guideTitle`) but `config.title` is for the view-level `title`.

kanitw:
Note that the object style is consistent with how `title`, `bin` and many other properties that support param objects work.

domoritz:
You're right, it is the way we do it in a lot of other places. I'm just thinking that it is not very extensible. For example, what if I want `Sum(X)` instead of `SUM(X)` or even a localized `Summe von X` or `Average of X` instead of `MEAN(X)`?

kanitw:
I think one extensible way is allow `compile` function to take more parameter in the future. 
e.g., maybe users can specify `titleFormat` function .

kanitw:
I don't think we would ever have a cycle to implement that though.  
But I also do not want to reimplement everything in Voyager.

kanitw:
Actually, the compile function way would actually work for Voyager very well.

kanitw:
Done in https://github.com/vega/vega-lite/pull/3092

PR #3092, opened 2017-10-12
Adjusting default title format

kanitw:
Please merge https://github.com/vega/vega-lite/pull/3091 first!

From discussion in  https://github.com/vega/vega-lite/issues/1977, maybe the current title format looks too much like programmatically generated. 

For example, given a field "`field`" we change:

- `SUM(field)` => `Sum of field`
- `BIN(field)` => `Bin of field`
- `YEARMONTH(field)` => `Year-month of field`


examples/compiled/area.vg.json
```diff
@@ -156,7 +156,7 @@
             "tickCount": {
                 "signal": "ceil(height/40)"
             },
-            "title": "SUM(count)",
+            "title": "Sum of count",
```
domoritz: This is just count, right? We could set this manually in the spec.

domoritz: Or even better, use plain.

kanitw:
@domoritz Btw,  I really wish I can use the gallery to revise all the changes, but right now the gallery is broken ><"

domoritz:
I fixed the gallery.

examples/compiled/area.vg.json
```diff
@@ -122,7 +122,7 @@
             "tickCount": {
                 "signal": "ceil(width/40)"
             },
-            "title": "Year-month of date",
+            "title": "date (year-month)",
```
domoritz: I think this is worse for most examples like

`date (month)` or `date (date)` instead of `month of date` or `date of date`.

kanitw: I actually think the opposite way that's why I make the change.

For aggregate, I think having "Sum/mean/median of" is critical information so it should not be in parentheses. 
Meanwhile, for time unit, we often see people just shows time data without specifying units.  
Thus, putting time unit in parentheses might better signify that they are units that we display.

I don't see why `month of date` is better than `date (month)`.  Meanwhile, the confusing `year of year` definitely happens more than `date of date`.  Doing `year (year)` signifies that the `year` in parentheses is a unit.  I agree that `date (date)` is confusing, but it should rarely happen and`date of date` is not really less confusing anyway.

domoritz: Okay, I don't have strong opinions here. Let's merge it.

PR #3099, opened 2017-10-15
Dom/default label angle

domoritz:
I haven't changed the SVGs because I'd change all of them.


src/compile/axis/encode.ts
```diff
@@ -63,8 +63,7 @@ export function labelAngle(axis: Axis, channel: Channel, fieldDef: FieldDef<stri
     // Make angle within [0,360)
     return ((axis.labelAngle % 360) + 360) % 360;
   } else {
-    // auto rotate for X
-    if (channel === X && (contains([NOMINAL, ORDINAL], fieldDef.type) || !!fieldDef.bin ||  isTimeFieldDef(fieldDef))) {
+    if (channel === X && contains([NOMINAL, ORDINAL], fieldDef.type)) {
```
kanitw: I think you should only remove bin?

Time definitely should be rotated too.  So maybe add TEMPORAL here?

kanitw: Actually it's quite tricky.  I can see why you do not want to rotate it for year.  However, for month and longer timeUnits, rotating is better ...

domoritz: Well, we actually have a lot of cases where we manually rotated the label. I think as a default it makes more sense to not rotate (consistency with quantitative).

kanitw: > Well, we actually have a lot of cases where we manually rotated the label.

I notice that you remove manual rotation in many cases. 

However, if you compile SVGs, then you will see that there are many other specs (e.g., when we use month) that we have to add manual rotation due to the default changes. So I don't think this is necessary overall better. (It's better for some, worse for some.) 

I think this is quite not satisfactory.  I w

> I think as a default it makes more sense to not rotate (consistency with quantitative).
I disagree.  Month and Weekday by default are strings so they are more like nominal/ordinal. 

I wonder if one solution is to only rotate if the time unit includes month or (week) day then that should solve most of the problem.

domoritz: I don't claim that not rotating is better in every case. It certainly isn't. however, it is more consistent with quantitative scales and also I think the default should be to not do anything unless we have good reasons to rotate the labels. For quantitative and temporal scales it is okay if we have fewer labels because you can always infer the points in between.

kanitw: > For quantitative and temporal scales it is okay if we have fewer labels because you can always infer the points in between

Good point.  I'm convinced.

PR #3101, opened 2017-10-15
Warn if users try to share scales in concat view. Fixes #3086

### LINKED ISSUE #3086: Warning that axis resolve for concat / repeat is not supported

domoritz:


src/compile/concat.ts
```diff
@@ -17,6 +18,12 @@ export class ConcatModel extends BaseConcatModel {
   constructor(spec: ConcatSpec, parent: Model, parentGivenName: string, repeater: RepeaterValue, config: Config) {
     super(spec, parent, parentGivenName, config, spec.resolve);
 
+    for (const t of ['axis', 'scale']) {
+      if (spec.resolve && spec.resolve[t] && (spec.resolve[t].x === 'shared' || spec.resolve[t].y === 'shared')) {
+        log.warn(log.message.CONCAT_CANNOT_SHARE_SCALE_OR_AXIS);
```
kanitw: I think you can share scale, but not axis

domoritz: Replaced the commit.

Issue #3107, opened 2017-10-16
Too few ticks for binned axis


domoritz:
![image](https://user-images.githubusercontent.com/589034/31592460-472dbbca-b1dd-11e7-8bfa-c2b27529a4df.png)

kanitw:
Perhaps the right thing to do is to generate less ticks only for binned field.

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/3127

PR #3113, opened 2017-10-16
Dom/more ticks

domoritz:


domoritz:
@kanitw I don't think I actually want this but this but here you can see how we could fix #3107. e can also just close it.

PR #3115, opened 2017-10-16
Added marked-vega

eterna2:
Added `marked-vega` to list of application for Vega-Lite.

`marked-vega` is a Polymer webcomponent to parse image/code markdowns into Vega or Vega-Lite charts.


kanitw:
@eterna2 very nice.  Thank you for making this!

PR #3121, opened 2017-10-17
Dom/more ticks

domoritz:


kanitw:
Looking at the output -- i have quite a mixed feeling about this change.  

I think it certainly is better for `bar_aggregate_count.svg` but for other cases, I wonder if we are adding too many ticks. 

image

image

image

Since having less ticks that number of bins is not that bad, and there are more plots without bin,
I wonder if we should just close this branch and ignore #3107?

PR #3128, opened 2017-10-17
Update ranged dot plot example and add to documentation for `detail` channel

haldenl:
Update ranged dot plot example to use `detail` channel and make small style changes. Add ranged dot plot to documentation as an example of using `detail` channel in this way.


kanitw:
- [ ] add to "Multi-series Line Chart with the Detail Channel"
- [ ] add https://vega.github.io/vega-lite/docs/encoding.html#detail

domoritz:
And fix the build ;-)

domoritz:
Should I expect that this gets fixed or not?

haldenl:
Sorry I missed this! The build has been fixed.

domoritz:
Awesome, could you address @kanitw's comments in https://github.com/vega/vega-lite/pull/3128#issuecomment-337386431 and rebase?

domoritz:
I think git got your rebase wrong. I can help if you want. 

screen shot 2017-10-27 at 15 38 49

haldenl:
Yes, I was having issues with that... Advice would be appreciated!

examples/specs/layer_ranged_dot.vl.json
```diff
@@ -3,83 +3,77 @@
...
-      },
-      {
-          "filter": {
-              "field": "year",
-              "oneOf": [1955, 2000]
-          }
+    {
+      "filter": {
+        "field": "country",
+          "oneOf": ["China", "India", "United States", "Indonesia", "Brazil"]
```
kanitw: Can you make the tabs align here?

examples/specs/layer_ranged_dot.vl.json
```diff
@@ -3,83 +3,77 @@
...
-                      "#911a24"]
-                  },
-                  "legend": {
-                      "title": "Year"
-                  }
-              }
-          }
+    {
+      "mark": {
+          "type": "point",
```
kanitw: I think the points should be above the lines (not below).

haldenl:
Ah yes of course, fixed!

kanitw:
Ah I just realize that the default point opacity is 0.7 (mainly for scatterplots) -- but we should set it to `1`. 

@domoritz  -- since `point`s will be now be used in many scenarios (rather than just single view plots in the past) -- I wonder if we should eliminate automatic opacity.

My rationale is that we should only do automatic stuff when it works _most of the time_.  
But right now, whenever we try to layer/overlay points on top of line/rule/area/etc.  The automatic opacity now always cause problems in all of these cases.  Alternatively, we may want to revise the "overlay" syntax sooner than later and make them available.  (And make sure the overlay use opacity=1.)

kanitw:
@haldenl -- In any case, you may want to add "opacity: {value: 1}" to the points for now so we can merge :)

PR #3145, opened 2017-10-19
Add mark config properties to mark definition

kanitw:
Rationale: now that config is a top-level property -- there is no direct way to customize each individual single view.  

(The only possible way is to use `style`, but that is pretty indirect and make more sense when one want to apply the same styles to multiple parts.) 

An alternative solution would be implementing encoding channels for all of these properties. However, that requires careful thinking about scale and scale ranges + legends for all of the properties we may add.  Given that's a non-trivial amount of design work, I don't think it's gonna happen anytime soon so this seems like the most natural thing to do.


site/docs/mark/mark.md
```diff
@@ -62,9 +62,9 @@ For example, a bar chart has `mark` as a simple string `"bar"`.
 }
 ```
 
-A mark definition object can contains the following values.
+A mark definition object can contains the following properties and all properties of the mark config:
```
domoritz: Would it make sense to link to it or include all properties in the table?

kanitw: Good catch.  I'll add a link

domoritz: Ping me when you added it.

kanitw: Done

Issue #3040, opened 2017-09-26
Support multiple conditions in an encoding (if/else/else...)


pelotom:
Correct me if I'm wrong, but it seems like currently an encoding can only take one of at most two branches, based on a single condition:

```json
{
  "color": {
    "condition": { "selection": "focused", "value": "red" },
    "value": "blue"
  }
}
```

but it would be nice if one could check multiple conditions until one succeeds, e.g. something like

```json
{
  "color": [
    { "condition": { "selection": "focused", "value": "red" } },
    { "condition": { "selection": "pinned", "value": "green" } },
    { "value": "blue" }
  ]
}
```

kanitw:
You're right, currently we intentionally support only two branches, so that we can limit that one encoding channel can have only one `field`.  

That said, it is possible to extend the design to have multiple branches with different `value`s -- as long as there is only one `field` for each encoding channel (relaxing the latter constraint introduce way more complexity to the system and is definitely not worth supporting).   

I could see it work with the following syntax: 
```
{
  "color": {
    "condition": [
       { "selection": "sel1", "value": "red" },
       { "selection": "sel2", "value": "green" },
    ],
    "value": "blue"
  }
}
```
where the `condition` array can only be a [ConditionValueDef](https://vega.github.io/vega-lite/docs/condition.html#value). 

However, we haven't prioritized it since we have not found an example visualization that would really need this.  If you have a convincing use case, please feel free to submit one. 

For example, If you have a clear use case to have "pinned" data and "focused" data, please feel free to describe.

pelotom:
That syntax looks good to me.

Ok, let me describe my use case. I want the user to be able to hover the mouse over various elements in the chart, whereupon they will acquire a "focused" effect... maybe in a scatterplot they become slightly larger than the other points, for example. Then if the user clicks on an element it becomes "pinned", at which point it becomes even more emphasized... maybe in addition to being larger than the other points it acquires a red color, for example. Now as the user hovers over other items they acquire the focus effect, _but not the pinned effect_, so that it's clear that the old element is still pinned, whereas the things they are hovering over are only transiently focused and _potentially_ pinnable. This chart might be accompanied by an external details view which listens only for "pin" events and shows details about the pinned element, so it needs to be clear which element that view is referring to.

kanitw:
Will be fixed in https://github.com/vega/vega-lite/pull/3146 (pending review)

PR #3146, opened 2017-10-19
Array of condition value defs

kanitw:
Fix https://github.com/vega/vega-lite/issues/3040


site/docs/encoding/condition.md
```diff
@@ -8,9 +8,11 @@ For [mark property channels](encoding.html#mark-prop) as well as [text and toolt
 
 There are two ways to use the `condition` property:
 
-1) Combining a [conditional field definition](#field) with a value definition
+1) Combining a [conditional field definition](#field) with a value definition.
+
+2) Combining a [conditional value definition](#value) or an array of conditional value definitions with a  field definition or a value definition.
```
arvind: This wording is confusing. Here's what I'd suggest:

```markdown
1. Combining one conditional field definition with one value definition.
2. Combining one or more conditional value definitions with a single field or value definition. 
```

I'd favor some brevity here since "conditional [blah] definition" is already a mouthful (I tried to brainstorm alternatives here but couldn't come up with anything better... maybe we drop "definition"?). We link to the two cases anyways, so the fact that (2) should be expressed as an array will be made explicit later on.

kanitw:
Added example and simplify the docs a bit like suggested.  The example also reveals an issue that's beyond the scope of this PR.  https://github.com/vega/vega-lite/issues/3154

domoritz:
@arvind @kanitw Please merge this when you think it's ready.

kanitw:
I think it's ready for review.  (I'm not supposed to merge my own PR.) 

I also rename interfaces https://github.com/vega/vega-lite/pull/3146/commits/01980c306768079f6253b4fcabae0f51322421f2

PR #3147, opened 2017-10-19
Pad continuous scale if it is a dimension of bar  

### LINKED ISSUE #2988: Pad continuous scales for bar
Currently, we use `continuousBandSize` for bar = 2.  

This means that bars on continuous scales (such as linear and time scales) are very thin.

```
{
  "mark": "bar",
  "encoding": {
    "x": {"timeUnit": "month", "field": "date", "type": "temporal"},
    "y": {"aggregate": "mean", "field": "price", "type": "quantitative"}
  }
}
```

image
 
Looking at Tableau, they actually set default bandSize on quantitative scale to be quite big.

image

However, note that they also automatically pad the domain of the x-scale so that the bar fits within the plot.  

 In VL, if we only increase  `continuousBandSize`, we will get something like this:

image

It seems like we should determine a method to expand the x-domain by a little bit akin to Tableau.

- Maybe we can do this by calculating extent and set the domain to slightly expanded extent.  But I wonder if we should make [Vega scale domain's data reference](https://vega.github.io/vega/docs/scales/#domain) support this feature.
- Moreover, we need a good rule of thumb for how much should we expand the domain by default.  The rule might vary for quantitative and temporal data and for binning/timeUnit. 

Note that Tableau also consistently expand the scale domain even for scatter plots -- probably for consistency when users transition between different marks. 

image

cc: @jheer @arvind @domoritz -- Any thoughts?

kanitw:
Fix #2988


kanitw:
(Need to merge https://github.com/vega/vega-lite/pull/3148 and rebase this on top of that branch)

domoritz:
I rejected #3148 because I don't see the changes. Can you try again?

domoritz:
Rebased and rebuilt examples for you.

kanitw:
Thanks -- I just make sure this only get applied to non-binned field.  This should be good to go.

domoritz:
Shouldn't the end be padded as well?

screen shot 2017-10-20 at 11 58 29

kanitw:
Look closely -- its already padded

PR #3148, opened 2017-10-19
Update examples for 3.0.6

kanitw:
@domoritz  -- did clean node modules before re-compiling examples (I just did but still get different results)


domoritz:
I just rebuilt all the examples on master with a fresh node_modules. No changes! Please reopen if you still think that yours is correct.

PR #3149, opened 2017-10-19
Remove vg specs

domoritz:
Are these used anywhere? @mattwchun


kanitw:
I think these are left-over from a very old "move null filter to end branch" that we just merged recently.

PR #3151, opened 2017-10-19
Updates for new Vega-Embed

domoritz:


domoritz:
Updated to latest Vega-Embed

domoritz:
Good catch. The problem is that Vega doesn’t provide any typings. I’ll fix it.

domoritz:
I don't understand yet why but the vega loader doesn't load the right specs...

screen shot 2017-10-22 at 21 53 40

PR #3156, opened 2017-10-23
We don’t use vega extend.

domoritz:
Removing this prevents us from accidentally using it.


kanitw:
If you wanna move this to vega-util repo soon, we will have to undo this anyway?  That said if u wanna merge just merge.

domoritz:
I don't think we will move it soon.

PR #3162, opened 2017-10-25
Add django-chartflo to applications tools list

synw:
Add [django-chartflo](https://github.com/synw/django-chartflo) to the tools section of applications


domoritz:
Thank you!

PR #3167, opened 2017-10-25
String should now use @minimum. It also doesn’t make any sense here.

domoritz:


kanitw:
I guess you mean "should NOT" lol

domoritz:
Yes

PR #3180, opened 2017-10-30
Update bin.md

domoritz:


site/docs/transform/bin.md
```diff
@@ -59,6 +61,8 @@ You can use binning to discretize color scales. Vega-Lite automatically creates
 {:#transform}
 ## Bin Transform
 
+While Vega-Lite supports binning in transforms, the preferred way is to use binning in the encoding field.
```
kanitw: I really think this is quite inaccurate and sounds stupid (if one is more preferred, why implement another?) 

More accurate description would be that the current implementation for binning in encoding is better.

domoritz: Fair

PR #3186, opened 2017-11-03
Cleaner schema

domoritz:
Needs https://github.com/vega/ts-json-schema-generator/issues/19

@jakevdp, before I merge this, can you check that this schema in this PR works for you and doesn't have unused definitions.

### DIFF
--- a/src/data.ts
+++ b/src/data.ts
@@ -13,12 +13,6 @@ export interface DataFormatBase {
    * For Specific date formats can be provided (e.g., `{foo: 'date:"%m%d%Y"'}`), using the [d3-time-format syntax](https://github.com/d3/d3-time-format#locale_format). UTC date format parsing is supported similarly (e.g., `{foo: 'utc:"%m%d%Y"'}`). See more about [UTC time](timeunit.html#utc)
    */
   parse?: 'auto' | object;
-  /**
-   * Type of input data: `"json"`, `"csv"`, `"tsv"`.
-   * The default format type is determined by the extension of the file URL.
-   * If no extension is detected, `"json"` will be used by default.
-   */
-  type?: DataFormatType;
 }
 
 export interface CsvDataFormat extends DataFormatBase {
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -464,44 +464,6 @@
       },
       "type": "object"
     },
-    "AxisConfigMixins": {
-      "additionalProperties": false,
-      "properties": {
-        "axis": {
-          "$ref": "#/definitions/AxisConfig",
-          "description": "Axis configuration, which determines default properties for all `x` and `y` [axes](axis.html). For a full list of axis configuration options, please see the [corresponding section of the axis documentation](axis.html#config)."
-        },
-        "axisBand": {
-          "$ref": "#/definitions/VgAxisConfig",
-          "description": "Specific axis config for axes with \"band\" scales."
-        },
-        "axisBottom": {
-          "$ref": "#/definitions/VgAxisConfig",
-          "description": "Specific axis config for x-axis along the bottom edge of the chart."
-        },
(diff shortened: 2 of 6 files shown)


domoritz:
hmm, looks like `Config` is missing.

jakevdp:
There are still a handful of unneeded definitions:
```
{'BoxPlotConfig',
 'BoxPlotConfigMixins',
 'DataFormatType',
 'FacetedCompositeUnitSpecAlias'}
```

domoritz:
Hmm, the boxplot defs are there because we use @hide. I have to check where `FacetedCompositeUnitSpecAlias` comes from.

domoritz:
`DataFormatType` is subtle. The issue is that we are extending an abstract interface but in the extensions, we override the type.

domoritz:
`CompositeMarkConfigMixins` is also unused, isn't it?

domoritz:
There is a bug in the schema generator that causes some other classes to be included.

external.ts
```ts
export interface CC {
    c: number;
}

export interface C extends CC {
}

// or: export type C = CC;
```

test.ts
```ts
import { C } from "./external";

export interface MyObject extends C {
    foo: number;
}
```

domoritz:
I figures out everything except for `FacetedCompositeUnitSpecAlias`.

src/data.ts
```diff
@@ -13,12 +13,6 @@ export interface DataFormatBase {
    * For Specific date formats can be provided (e.g., `{foo: 'date:"%m%d%Y"'}`), using the [d3-time-format syntax](https://github.com/d3/d3-time-format#locale_format). UTC date format parsing is supported similarly (e.g., `{foo: 'utc:"%m%d%Y"'}`). See more about [UTC time](timeunit.html#utc)
    */
   parse?: 'auto' | object;
-  /**
-   * Type of input data: `"json"`, `"csv"`, `"tsv"`.
-   * The default format type is determined by the extension of the file URL.
-   * If no extension is detected, `"json"` will be used by default.
-   */
-  type?: DataFormatType;
```
kanitw: Why is `type` removed from here?

test/compile/selection/nearest.test.ts
```diff
@@ -21,7 +21,7 @@ function getModel(markType: any) {
   model.component.selection = selection.parseUnitSelection(model, {
     "one": {"type": "single", "nearest": true},
     "two": {"type": "multi", "nearest": true},
-    "three": {"type": "interval", "nearest": true},
+    "three": {"type": "interval"},
```
kanitw: Why are these get removed? We will need input from @arvind if removing these make sense.

src/data.ts
kanitw: Please make sure that the type displayed in the table in data.md is correct.

domoritz: Because we override it in all cases anyway. I fixed the table in data.md.

test/compile/selection/nearest.test.ts
arvind: I’m guessing these were removed because they are actually not valid Vega-Lite statements. I just had them here to verify that we did the right thing (eg ignore or throw errors as appropriate) if a user specified them. Will go through the other selection test changes too and comment if something looks amiss.

domoritz: And now the typescript compiler is smart enough to notice.

domoritz:
@jakevdp Let us know if the schema is clean now. @kanitw feel free to merge when you think it's ready.

kanitw:
Do you want to wait for @jakevdp's input? If not, feel free to merge.

domoritz:
The schema is clean except for `FacetedCompositeUnitSpecAlias` but I can't figure out where that's coming from right now.

jakevdp:
Thanks!

PR #3189, opened 2017-11-04
Added gulp-marked-vega

eterna2:
Added `gulp-marked-vega` to libraries available for Vega and Vega-Lite markdowns. 

`gulp-marked-vega` is a gulp plugin and a cli tool to replace `marked-vega` markdown syntax with base64 embedded image tags so that standard markdown parser can render the charts without any additional plugins or modification to the rendering rules.


domoritz:
Thanks!

PR #3191, opened 2017-11-06
Add note about escaping fields to schema.

domoritz:
I intentionally kept it short so that users actually read it.

Follow up for https://github.com/vega/vega-lite/commit/0a68ff175c11939dcf833c61c953aca50aad8925

### DIFF
--- a/site/docs/encoding/field.md
+++ b/site/docs/encoding/field.md
@@ -31,9 +31,9 @@ For example, we can set `field` to `"precipitation"` to map it to `x` position.
 
 <span class="vl-example" data-name="tick_dot"></span>
 
-Valid JavaScript object access paths using either dot (`.`) or bracket (`foo['bar']`) notation will be converted into lookups on nested objects. Make sure you use proper escape characters if that's not desired. For example, to specify field names that contain dots but are not nested lookups, escape the dot like in `"my\\.field"`. Similarly, escape `[` and `]` as `\\[` and `\\]`.
+Valid JavaScript object access paths using either dot (e.g., `foo.bar`) or bracket with quotes (e.g., `foo['bar']`) notations can be used in `field` to perform nested object lookups. If field names contain dots or brackets but are not nested, use `\\` to escape dots and brackets.  For example, if the field name is `a.b`, use `a\\.b`.  Similarly, if the field name is `a[0]`, use `a\\[0\\]`.
 
-Further note that in JSON, you have to escape backslashes. Hence, a single backslash`\` is escaped as `\\\\` in JSON. You also need to escape quotes (`\"`) and use special escape characters for newlines (`\n`), carriage returns (`\r`), and tabs (`\t`).
+Further note that in JSON, you have to escape backslashes. Hence, a single backslash (`\`) is escaped as `\\\\` in JSON. You also need to escape quotes (`\"`) and use special escape characters for newlines (`\n`), carriage returns (`\r`), and tabs (`\t`).
 
 {:#repeat-ref}
 **2) An object defining iterated values from the `repeat` operator**
--- a/site/docs/encoding/field.md
+++ b/site/docs/encoding/field.md
@@ -31,7 +31,9 @@ For example, we can set `field` to `"precipitation"` to map it to `x` position.
 
 <span class="vl-example" data-name="tick_dot"></span>
 
-Take care to make sure you use proper escape characters as needed. Valid JavaScript object access paths using either dot (`.`) or bracket (`foo['bar']`) notation will be converted into lookups on nested objects. For example, to specify field names that contain dots but are not nested lookups, escape the dot like in `"my\\.field"`. Similarly, escape `[` and `]` as `\\[` and `\\]`.
+Valid JavaScript object access paths using either dot (`.`) or bracket (`foo['bar']`) notation will be converted into lookups on nested objects. Make sure you use proper escape characters if that's not desired. For example, to specify field names that contain dots but are not nested lookups, escape the dot like in `"my\\.field"`. Similarly, escape `[` and `]` as `\\[` and `\\]`.
+
+Further note that in JSON, you have to escape backslashes. Hence, a single backslash`\` is escaped as `\\\\` in JSON. You also need to escape quotes (`\"`) and use special escape characters for newlines (`\n`), carriage returns (`\r`), and tabs (`\t`).
 
 {:#repeat-ref}
 **2) An object defining iterated values from the `repeat` operator**
--- a/src/fielddef.ts
+++ b/src/fielddef.ts
@@ -101,6 +101,10 @@ export interface FieldDefBase<F> {
    * __Required.__ A string defining the name of the field from which to pull a data value
    * or an object defining iterated values from the [`repeat`](repeat.html) operator.
    *
+   * __Note:__ Dots (`.`) and brackets (`[` and `]`) can be used to access nested objects.
+   * Make sure you use proper escape characters if dots or brackets are part of the field name.
+   * You can find details about escaping in the [field docs](field.html).
+   *
    * __Note:__ `field` is not required if `aggregate` is `count`.
    */
   field?: F;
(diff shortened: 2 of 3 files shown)


kanitw:
You can also remove "take care to".

domoritz:
Escaping is really hard :-P. Try escaping `hello.foo\foo[bar\.baz\\`. 

I'll add some more details.

domoritz:
Pushed

src/fielddef.ts
```diff
@@ -100,6 +100,8 @@ export interface FieldDefBase<F> {
   /**
    * __Required.__ A string defining the name of the field from which to pull a data value
    * or an object defining iterated values from the [`repeat`](repeat.html) operator.
+   * Make sure you use proper escape characters as needed. `.`, `[`, and ']' have to
```
kanitw: I think it's worth saying 
```
__Note:__  Dots (`.`) and brackets (`[` and `]`) can be used to access nested objects.  
To access field names with dots or brackets, make sure to escape them using `\\`. 
For example, use `"field": "a\\.b` for a field named "a.b".
```

(Preview)

__Note:__ Dots (`.`) and brackets (`[` and `]`) can be used to access nested objects.  
To access field names with dots or brackets, make sure to escape them using `\\`. 
For example, use `"field": "a\\.b` for a field named "a.b".

kanitw:
> Try escaping `hello.foo\foo[bar\.baz\\`. 

Not sure what you mean

domoritz:
You have to escape a \ with \\\\

kanitw:
I still do not understand what you mean.   Why some \\ are escaped and some are not?  


Can you please explain in details and make sure these details are included in the docs.  It's okay if the `field` TSDocs contains a brief version but more details should be in the `field` page.   If the `field` page has more info, make sure to link from `field` TSDocs to the `field` page.

domoritz:
As you wished.

site/docs/encoding/field.md
```diff
@@ -31,7 +31,9 @@ For example, we can set `field` to `"precipitation"` to map it to `x` position.
 
 <span class="vl-example" data-name="tick_dot"></span>
 
-Take care to make sure you use proper escape characters as needed. Valid JavaScript object access paths using either dot (`.`) or bracket (`foo['bar']`) notation will be converted into lookups on nested objects. For example, to specify field names that contain dots but are not nested lookups, escape the dot like in `"my\\.field"`. Similarly, escape `[` and `]` as `\\[` and `\\]`.
+Valid JavaScript object access paths using either dot (`.`) or bracket (`foo['bar']`) notation will be converted into lookups on nested objects. Make sure you use proper escape characters if that's not desired. For example, to specify field names that contain dots but are not nested lookups, escape the dot like in `"my\\.field"`. Similarly, escape `[` and `]` as `\\[` and `\\]`.
```
kanitw: - Unclear what "proper escape character" mean since it's partly defined by us.  
- Also unclear what "that's not desired mean"

site/docs/encoding/field.md
```diff
@@ -31,7 +31,9 @@ For example, we can set `field` to `"precipitation"` to map it to `x` position.
...
+
+Further note that in JSON, you have to escape backslashes. Hence, a single backslash`\` is escaped as `\\\\` in JSON. You also need to escape quotes (`\"`) and use special escape characters for newlines (`\n`), carriage returns (`\r`), and tabs (`\t`).
```
kanitw: This is still very confusing.  I don't understand why a single backslash`\` is escaped as `\\\\` in JSON.

src/fielddef.ts
```diff
@@ -101,6 +101,10 @@ export interface FieldDefBase<F> {
    * __Required.__ A string defining the name of the field from which to pull a data value
    * or an object defining iterated values from the [`repeat`](repeat.html) operator.
    *
+   * __Note:__ Dots (`.`) and brackets (`[` and `]`) can be used to access nested objects.
+   * Make sure you use proper escape characters if dots or brackets are part of the field name.
```
kanitw: Unclear what "proper" means.

site/docs/encoding/field.md
kanitw: I read this and it's still unclear to me if newline should be `\n` or `\\n`.

kanitw:
@domoritz This part still reads confusing to me

> Further note that in JSON, you have to escape backslashes. Hence, a single backslash`\` is escaped as `\\\\` in JSON. You also need to escape quotes (`\"`) and use special escape characters for newlines (`\n`), carriage returns (`\r`), and tabs (`\t`).

domoritz:
> This part still reads confusing to me

It makes complete sense to me :-P

kanitw:
Why does single backslash become four backslashes while other things just need one backslash prefix?

domoritz:
```json
{
  "data": {
    "values": {
      "hits": {
        "hits": [
          {"source.reco\\": 2 ,"source.yes": 1},
          {"source.reco\\": 3 ,"source.yes": 4}
        ]
      }
    },
    "format": {"property": "hits.hits"}
  },
  "mark": "point",
  "encoding": {
    "x": {"field": "source\\.reco\\\\", "type": "quantitative"},
    "y": {"field": "source\\.yes", "type": "quantitative"}
  }
}
```

```
{
  "data": {
    "values": {
      "hits": {
        "hits": [
          {"source.reco\n": 2 ,"source.yes": 1},
          {"source.reco\n": 3 ,"source.yes": 4}
        ]
      }
    },
    "format": {"property": "hits.hits"}
  },
  "mark": "point",
  "encoding": {
    "x": {"field": "source\\.reco\n", "type": "quantitative"},
    "y": {"field": "source\\.yes", "type": "quantitative"}
  }
}
```

domoritz:
When you type `\`, the JS parser expects an escape sequence. So you need to escape the `\` itself.   This is why a single \ disappears in JavaScript. 

screen shot 2017-11-18 at 08 36 24

Yes, this is super annoying and confusing. I fully agree.

kanitw:
And most people would not know about all of these explanations you replied unless you put it in the docs page (probably too long to put in `field` docs).  :)

domoritz:
@kanitw Yeah, it's good that it's in the field page now. Can we merge it?

kanitw:
No.  I think the current markdown file is still very confusing.  I think we need more explanations like you reply above in this thread.

domoritz:
Hmm, any suggestions or concrete problems for this text:

Valid JavaScript object access paths using either dot (e.g., `foo.bar`) or bracket with quotes (e.g., `foo['bar']`) notations can be used in `field` to perform nested object lookups. If field names contain dots or brackets but are not nested, use `\\` to escape dots and brackets.  For example, if the field name is `a.b`, use `a\\.b`.  Similarly, if the field name is `a[0]`, use `a\\[0\\]`.

Further note that in JSON, you have to escape backslashes. Hence, a single backslash (`\`) is escaped as `\\\\` in JSON. You also need to escape quotes (`\"`) and use special escape characters for newlines (`\n`), carriage returns (`\r`), and tabs (`\t`).

kanitw:
> Further note that in JSON, you have to escape backslashes. Hence, a single backslash (\) is escaped as \\\\ in JSON. You also need to escape quotes (\") and use special escape characters for newlines (\n), carriage returns (\r), and tabs (\t).

This part is too brief and lack explanation like you do above. Just saying "you have to escape backslashes" doesn't justify why you need four backslash for a single one.  

That's why I asked:

> Why does single backslash become four backslashes while other things just need one backslash prefix?

domoritz:
I think the double encoding is necessary because Vega generates expressions. So when you write JSON, you need to escape `\` as `\\`. If Vega gets `\` and generates an expression from it, it would just disappear. Hence, we need to write `\\\\`. For `\n` and other special escape characters, this is not necessary because the `\\` does not become `\`. 

@jheer can you confirm that this is correct? If so, shouldn't Vega hide this from users? It looks like we used to do some escaping in Vega: https://github.com/vega/vega/blob/46ab30c3810eae58fce543d9a2f0b32a072da6f1/docs/releases/v1.4.3/vega.js#L160.

domoritz:
Will follow up in #3213.

Issue #2488, opened 2017-06-12
Support `window` in the transform array


kanitw:
Given the new [`joinaggregate`](https://vega.github.io/vega/docs/transforms/joinaggregate/) transform, we should also support this in the `transform` array.  

One possible syntax is to have a `join: boolean` flag in the `SummarizeTransform`.

kanitw:
Since @domoritz agree with the `join: boolean` flag, @AkshatSh you can directly modify AggregateNode and the type definition for `SummarizeTransform`. 

I also notice that `SummarizeTransform` doesn't support `cross` and `drop` in [Vega's aggregate transform](https://vega.github.io/vega/docs/transforms/aggregate/) yet, so probably worth adding them too.   Note that `cross` and `drop` are in `aggregate` but not in `joinaggregate` so we will probably have to make some minor type adjustment.

kanitw:
Right now AggregateNode always output type: 'aggregate', but with `join: true` -- we wanna do `joinaggregate`.

kanitw:
@AkshatSh   Let's not add the `join` flag for now as Jeff is working on a new `window` transform and will deprecate `joinaggregate`.    That said,  supporting `cross` and `drop` in aggregate could still be useful :)

kanitw:
Note that now a `window` transform has landed.  You may want to add a window transform to VL. :)

kanitw:
I think `window` transform syntax should be pretty similar to Vega, but should not have `type` (`type: window`) just like our summarize transform does not have `type: 'aggregate'`.  

Please propose a syntax and we can discuss :)

kanitw:
Good cases to demonstrate in examples:
- Percent of total
- Different from total/mean
- Percent difference from total/mean

kanitw:
Another example:
- [ ] rank chart as in https://stats.stackexchange.com/a/178344

kanitw:
Will be fixed in https://github.com/vega/vega-lite/pull/3199

PR #3199, opened 2017-11-15
[WIP] Revise `window` transform + Add Docs / Examples

AkshatSh:
Implemented Window Transform for Vega Lite
Issue #2488


src/compile/data/window.ts
```diff
@@ -0,0 +1,52 @@
...
+import {WindowFieldDef, WindowTransform} from '../../transform';
+import {duplicate} from '../../util';
+import {VgWindowTransform} from '../../vega.schema';
+import {DataFlowNode} from './dataflow';
+import {WindowOnlyOp} from '../../window';
+import {AggregateOp} from '../../aggregate';
+
+export class WindowTransformNode extends DataFlowNode {
+  public clone(): WindowTransformNode {
+      return new WindowTransformNode(duplicate(this.transform));
```
kanitw: de-tab

src/compile/data/window.ts
```diff
@@ -0,0 +1,52 @@
...
+      operations.push(window.op);
+      as.push(window.as);
+      params.push(window.param);
+      fields.push(window.field);
+    }
+
+    const result: VgWindowTransform = {
+      type: 'window',
+      params: params,
+      as: as,
```
kanitw: just do as

src/compile/data/window.ts
```diff
@@ -0,0 +1,52 @@
...
+    for (const window of this.transform.window) {
+      operations.push(window.op);
+      as.push(window.as);
+      params.push(window.param);
+      fields.push(window.field);
+    }
+
+    const result: VgWindowTransform = {
+      type: 'window',
+      params: params,
```
kanitw: just do 
```
params,
as,
ops,  // rename operations to ops
fields 
```

src/transform.ts
```diff
@@ -97,6 +98,57 @@ export interface AggregatedFieldDef {
   as: string;
 }
 
+export interface WindowFieldDef {
+  /**
+   * The operations supported for the window aggregation
```
kanitw: Need to link to the table that lists all supported operations in the docs.

src/transform.ts
```diff
@@ -97,6 +98,57 @@ export interface AggregatedFieldDef {
...
+   */
+  op: AggregateOp | WindowOnlyOp;
+
+  /**
+   * The optional parameters for the operation
```
kanitw: Add full stop. 

Explain a bit more that look at the operator table to see parameter for each operation.

src/transform.ts
```diff
@@ -97,6 +98,57 @@ export interface AggregatedFieldDef {
...
+   */
+  param?: Number;
+
+  /**
+   * The field that will be used in the operation, some operations
```
kanitw: The field that will be used in the operation.  Note that some operations ...

src/transform.ts
```diff
@@ -97,6 +98,57 @@ export interface AggregatedFieldDef {
...
+   * do not require fields
+   */
+  field?: string;
+
+  /**
+   * The name for the new field in the window transform
```
kanitw: full stop.. can you add it all for the rest of the docs.

src/transform.ts
```diff
@@ -97,6 +98,57 @@ export interface AggregatedFieldDef {
...
+}
+
+export interface WindowTransform {
+  /**
+   * Array of objects that summarize the fields that will be aggregated over the window
+   */
+  window: WindowFieldDef[];
+
+  /**
+   * The frame for the window, if none is set the default is [null, 0] everything before the
```
kanitw: use back tick around code

src/transform.ts
```diff
@@ -97,6 +98,57 @@ export interface AggregatedFieldDef {
...
+   * current item
+   */
+  frame?: Number[];
+
+  /**
+   * Whether to ignoreThePeers during the comparison in the window
```
kanitw: Why camelcase for `ignoreThePeers`?

src/transform.ts
```diff
@@ -139,10 +191,26 @@ export interface LookupTransform {
   default?: string;
 }
 
+export interface Comparator {
```
kanitw: Wonder if naming this SortField or WindowSortField make more sense?
(But not sure if there is another thing already called SortField -- but this is definitely slightly different from `sort` 's SortField(Def?).

src/compile/data/window.ts
```diff
@@ -0,0 +1,52 @@
+import {add} from '../../compositemark/index';
+import {WindowFieldDef, WindowTransform} from '../../transform';
+import {duplicate} from '../../util';
+import {VgWindowTransform} from '../../vega.schema';
+import {DataFlowNode} from './dataflow';
+import {WindowOnlyOp} from '../../window';
+import {AggregateOp} from '../../aggregate';
+
+export class WindowTransformNode extends DataFlowNode {
```
kanitw: WindowNode?

test/compile/data/parse.test.ts
```diff
@@ -55,5 +57,29 @@ describe('compile/data/parse', () => {
       assert.isTrue(result.first instanceof BinNode);
       assert.isTrue(result.last instanceof AggregateNode);
     });
+
+    it ('should return a WindowTransform Node', () => {
+      const transform: Transform = {
+        window: [
```
kanitw: probably to add another test (or modify this test) to include other optional properties of window transform

src/vega.schema.ts
```diff
@@ -1183,3 +1184,22 @@ export interface VgTitleConfig {
    */
   orient?: TitleOrient;
 }
+
+export type VgComparatorOrder = 'ascending' | 'descending';
+
+export interface VgComparator {
+  field?: string | string[];
+  order?: VgComparatorOrder;
```
kanitw: shouldn't this allow array too?

src/compile/data/window.ts
```diff
@@ -0,0 +1,52 @@
...
+      params: params,
+      as: as,
+      ops: operations,
+      fields: fields,
+      ignorePeers: this.transform.ignorePeers,
+      groupby: this.transform.groupby,
+      sort: this.transform.sort
+    };
+
+    return result;
```
kanitw: minor -- but I guess you can just return the line above

kanitw:
@domoritz  – Can you help @AkshatSh figure why the Travis says that new specs in this PR do not pass validation?  I tried testing in the local machine myself, but it's passing ok for me.

examples/examples.test.ts
```diff
@@ -33,7 +33,7 @@ function validateVL(spec: TopLevelExtendedSpec) {
 }
 
 function validateVega(spec: TopLevelExtendedSpec) {
-  const vegaSpec = compile(spec).spec;
+  const vegaSpec = JSON.parse(JSON.stringify(compile(spec).spec));
```
kanitw: @domoritz  -- any idea why this makes some validation passes?

kanitw: @AkshatSh -- this commit should be separated from other changes as it's a controversial change.

src/compile/data/window.ts
AkshatSh: Resolved.

src/compile/data/window.ts
AkshatSh: Resolved.

src/transform.ts
AkshatSh: Resolved.

src/transform.ts
AkshatSh: Resolved.

src/transform.ts
AkshatSh: Resolved.

src/transform.ts
AkshatSh: Resolved.

src/transform.ts
AkshatSh: I made a couple changes and this should be reflective of the Comparator defined here now: https://vega.github.io/vega/docs/types/#Compare

src/vega.schema.ts
AkshatSh: Assuming this is referring to the field parameter, it is resolved.

src/transform.ts
kanitw: Um, I haven't looked how you do it -- but seeing that you mention that you make it like "Compare" in Vega.  I would like to mention that it might be more important to make it more consistent with rest of Vega-Lite language than the underlying Vega.

examples/examples.test.ts
domoritz: This removes undefines and is not something we should do in tests. I am strongly against this change.

AkshatSh: Oh I see, what happens when you stringify something that is undefined?

examples/specs/window_transform.vl.json
```diff
@@ -0,0 +1,46 @@
+{
+    "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+    "description": "Bar graph showing the row number for each of the cells",
+    "width": 300,
+    "height": 50,
+    "data": {
+        "values": [
+            { "x": 1, "y": 28 }, { "x": 2, "y": 55 },
```
kanitw: Why do the data need "x" here at all?

Also, maybe rename this file to `window_row_number`?

examples/specs/window_transform_average_rating.vl.json
```diff
@@ -0,0 +1,44 @@
+{
+    "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+    "description": "Bar graph showing how each film differs from the average rating for that year",
```
kanitw: I think this example is quite confusing.   
I took me a while to get that "Good" "Bad" "Average" are movie names.  Maybe better put actual movie names in there?  

Also, it's kinda weird that you calculate average ratings for one movie in 2015. 

Perhaps, try to re-create examples using the movies data instead? (Feel free to play with it in  Voyager a bit.)

examples/specs/window_transform_time_spent_in_class.vl.json
```diff
@@ -0,0 +1,53 @@
+{
+    "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+    "description": "Bar graph showing how each film differs from the average rating for that year",
```
kanitw: outdated description?

examples/specs/window_transform_time_spent_in_class.vl.json
```diff
@@ -0,0 +1,53 @@
...
+            "x": {
+                "field": "Class",
+                "type": "nominal",
+                "scale": { "rangeStep": 12 },
+                "axis": { "title": "" }
+            },
+            "y": {
+                "field": "TimeInClass",
+                "type": "quantitative",
+                "axis": { "title": "Fatigue", "grid": false }
```
kanitw: "fatigue" as a title axis is quite confusing to me.

examples/specs/window_transform_time_spent_in_class.vl.json
kanitw: I'm also not sure what this plot try to communicate.

kanitw:
Btw, please make sure to demonstrate these cases in your doc page: 
- [ ] Percent of total
- [ ] Different from total/mean
- [ ] Percent difference from total/mean

kanitw:
@AkshatSh 

We're hoping to release 2.2 (https://github.com/vega/vega-lite/pull/3461) some times next quarter.  
I wonder if you would have time over spring break to finish this.  (No worry about it this week of course,  I know it's the dead week.)

examples/specs/window_transform_average_rating.vl.json
AkshatSh: I liked this example because it could group by year. I changed the movie names to be Move One etc. I also added another example to use the movies.json file and do something similar there.

examples/specs/window_transform_time_spent_in_class.vl.json
AkshatSh: I was trying to make a more complicated window transform example. The purpose was to show how long each student has been in class until the current class. So if student A and B both show up to a class, the graph will show that class along with the total time each student has been at school, to show how tired (or the fatigue) of a student during that class. For example, something like this could be used to see the affects of taking an exam after a lot of other classes vs taking an exam at the beginning of the day or when there are a few classes before hand.

I changed the name of the axis to be `Time Spent in Class` instead to hopefully make this more clear. I have also updated the description.

src/compile/data/window.ts
```diff
@@ -0,0 +1,84 @@
...
+    const ignorePeers = this.transform.ignorePeers;
+
+    const result: VgWindowTransform = {
+      type: 'window',
+      params,
+      as,
+      ops,
+      fields,
+      sort,
+    };
```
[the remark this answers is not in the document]
AkshatSh: I tried this, but I was not able to get it to compile properly:

```typescript
const result: VgWindowTransform = {
      type: 'window',
      params,
      as,
      ops,
      fields,
      sort,
ignorePeers !== undefined ? {ignorePeers} : {},
    };
```

site/docs/transform/window.md
```diff
@@ -0,0 +1,131 @@
...
+  ],
+  ...
+}
+```
+
+## Window Transform Definition
+
+| Property    | Type | Description  |
+| :----------- | :-------: | :------------|
+| window | WindowFieldDef[] | The definition of the fields in the window, and what calculations to use. |
```
kanitw: Use a special tag and make sure that you have the JSDocs in the code include comments for all of these 

`{% include table.html props="window,frame,ignorePeers,groupBy,sort" source="WindowTransform" %}`


(Window Only Operation Reference can't use this table thing though as it's a list of valid values, not property table.)

site/docs/transform/window.md
```diff
@@ -0,0 +1,131 @@
...
+| first_value  | _None_    | Assigns a value from the first data object in the current sliding window frame. This operation must have a corresponding entry in the _fields_ parameter array.|
+| last_value   | _None_    | Assigns a value from the last data object in the current sliding window frame. This operation must have a corresponding entry in the _fields_ parameter array.|
+| nth_value    | {% include type t="Number" %} | Assigns a value from the nth data object in the current sliding window frame. If no such object exists, assigns `null`. Requires a non-negative integer parameter that indicates the offset from the start of the window frame. This operation must have a corresponding entry in the _fields_ parameter array.|
+
+## Examples
+To replace any of the examples from total to `mean`, the operation just needs to be changed from `sum` to `mean`.
+
+### Percent of Total
+
+```json
```
kanitw: Can you add actual Vega-Lite specs for all of these?   Also make sure to include other good examples that you're adding.  

`<div class="vl-example" data-name="example-file-name-without-extension"></div>`

src/compile/data/window.ts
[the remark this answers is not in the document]
kanitw: Weird... What's the error?

src/transform.ts
```diff
@@ -143,10 +198,23 @@ export interface LookupTransform {
   default?: string;
 }
 
+
+/**
+ * A compartor for fields within the window transform
+ */
+export interface WindowSortField {
+  field: string;
+  order?: ('ascending' | 'descending');
```
kanitw: No need for parentheses?

src/transform.ts
```diff
@@ -101,6 +102,60 @@ export interface AggregatedFieldDef {
...
 
+export interface WindowFieldDef {
+  /**
+   * The operations supported for the window aggregation. See the list of supported operations here:
+   *   https://vega.github.io/vega-lite/docs/transforms/window.html
+   */
+  op: AggregateOp | WindowOnlyOp;
+
+  /**
+   *  Parameter values for the window functions. Parameter value can be null for operations that do not accept a
```
kanitw: "See the list of operations for each operation's supported parameter."?

src/window.ts
```diff
@@ -0,0 +1,2 @@
+export type WindowOnlyOp = 'row_number' | 'rank' | 'dense_rank' | 'percent_rank' | 'cume_dist'
```
kanitw: Why not adding this to transform.ts?  

(It's kinda weird that you add this file only for this.)  

If you wanna have window.ts,  better have one file for each transform type. (I don't think we should though -- at least for now.)

src/transform.ts
```diff
@@ -101,6 +102,60 @@ export interface AggregatedFieldDef {
...
+  window: WindowFieldDef[];
+
+  /**
+   * The frame for the window, if none is set the default is `[null, 0]` everything before the
+   * current item.
+   */
+  frame?: (null | number)[];
+
+  /**
+   * Will indicate whether to ignore peer values (items with the same rank) in the window. The default value is `False`.
```
kanitw: `false`

src/transform.ts
kanitw: Also, for operations that do not accept a parameter -- should param really be `null`?   Why not `undefined` (omitted)?

examples/specs/window_transform_movie_mean_difference.vl.json
```diff
@@ -0,0 +1,35 @@
...
+            },
+            {
+                "calculate": "datum.IMDB_Rating - datum.AverageRating",
+                "as": "DifferenceFromMean"
+            }
+        ],
+        "mark": { "type": "bar", "clip": true },
+        "encoding": {
+            "x": {
+                "field": "Title",
```
kanitw: I think there are definitely too many titles here to be comprehensible  -- pick a subset of 10 names using `filter`?

examples/specs/window_transform_average_rating.vl.json
kanitw: I'd add another movie for 2015 to make it less weird.   Also, looks like the bar is acting weird here.  (Why isn't it starting from zero?)  -- Use point for now and file a bug?

examples/specs/window_transform_row_number.vl.json
```diff
@@ -0,0 +1,46 @@
...
+            { "y": 87 }, { "y": 66 },
+            { "y": 17 }, { "y": 27 },
+            { "y": 68 }, { "y": 16 },
+            { "y": 49 }, { "y": 15 }
+        ]
+    },
+    "layer": [{
+        "transform": [{
+                "calculate": "datum.y - 50",
+                "as": "ny"
```
kanitw: What is this ny for?

examples/specs/window_transform_row_number.vl.json
```diff
@@ -0,0 +1,46 @@
+{
```
kanitw: Why is this example important?  
(Is it important for explaining any concept in the documentation?)

examples/specs/window_transform_time_spent_in_class.vl.json
```diff
@@ -0,0 +1,53 @@
+{
```
kanitw: Following the discussion here: https://github.com/vega/vega-lite/pull/3199#discussion_r175275403

I'm still not convinced that this is a good example.  

If you can explain a specific analysis question that this plot can answer well, I can be more convinced. (Why would you want to stack the time spent in school for the two students?)

examples/specs/window_transform_row_number.vl.json
kanitw: I'd say rank chart as in https://stats.stackexchange.com/a/178344  is a more compelling use case of `row_number`

examples/specs/window_transform_movie_mean_difference.vl.json
kanitw: Btw, this one is a clearly reasonable example for "### Different from Mean"  Example in the docs.

(You can remove different from total -- diff from mean is way more common)

site/docs/transform/window.md
```diff
@@ -0,0 +1,131 @@
...
+    "frame": [null, null]
+  },
+  {
+    "calculate": "datum.x/datum.total",
+    "as": "Percent of Total"
+  }
+```
+
+
+### Difference from Total
```
kanitw: An improved window_transform_movie_mean_difference.vl.json is a fair game for this one. 

I'd put this one as the first example in the docs, as it seems the most common.

src/compile/data/window.ts
```diff
@@ -9,31 +9,35 @@ import {DataFlowNode} from './dataflow';
...
 
-  constructor(private transform: WindowTransform) {
-    super();
+  constructor(parent: DataFlowNode, private transform: WindowTransform) {
+    super(parent);
   }
 
   public producedFields() {
     const out = {};
     this.transform.window.forEach(element => {
```
kanitw: element => windowFieldDef?

src/compile/data/window.ts
```diff
@@ -9,31 +9,35 @@ import {DataFlowNode} from './dataflow';
...
-      out[element.as] = true;
+      out[this.getDefaultName(element)] = true;
     });
 
     return out;
   }
 
+  private getDefaultName(window: WindowFieldDef): string {
```
kanitw: window => windowFieldDef

src/compile/data/window.ts
```diff
@@ -43,7 +47,7 @@ export class WindowTransformNode extends DataFlowNode {
     const sortFields: string[] = [];
     const sortOrder: VgComparatorOrder[] = [];
     if (this.transform.sort !== undefined) {
-      for (const compField of this.transform.sort.compare) {
+      for (const compField of this.transform.sort) {
```
kanitw: compField => sortField

examples/specs/window_transform_movie_mean_difference.vl.json
kanitw: Idea:  try to pick titles that are academic award winners?

examples/specs/window_transform_average_rating.vl.json
kanitw: Btw, this example seems redundant with window_transform_movie_mean_difference.vl.json? 

An improved window_transform_movie_mean_difference.vl.json would be way more interesting, so remove this one? 

(Still, please file a bug that bar for this one is wrong.)

src/transform.ts
AkshatSh: That was my mistake. I copied the definition from the vega docs, but here it can just be omitted. Updated to remove that.

src/window.ts
AkshatSh: I actually didn't realize this was in a separate file. I moved it into transform.ts

AkshatSh: Actually, I had done this because WindowOnlyOp is necessary for vega.schema.ts. I think I would create a circular dependency if I had it in transform.ts, so I moved it to transform.d.ts

site/docs/transform/window.md
AkshatSh: Oh thats pretty cool. I didn't know about that tag. Just updated it to use that for the 3 object definitions (WindowTransform, WindowSortField, WIndowFieldDef).

src/transform.ts
```diff
@@ -110,20 +111,17 @@ export interface WindowFieldDef {
...
-   *  Parameter values for the window functions. Parameter value can be null for operations that do not accept a
-   *  parameter.
+   *  Parameter values for the window functions. Parameter value can be omitted for operations that do not accept a parameter.
    */
   param?: number;
 
   /**
-   * The data fields for which to compute aggregate or window functions. Field can be omitted for operations that do not
-   * operate over a specific data field, including count, rank, and dense_rank.
+   * The data field for which to compute the aggregate or window function. This can be null for functions that do not operate over a field such as `count`, `rank`, `dense_rank`.
```
kanitw: null => omitted

src/transform.ts
```diff
@@ -110,20 +111,17 @@ export interface WindowFieldDef {
...
    */
   field?: string;
 
   /**
-   * The output name for each field. If none is defined will use the format op_field. For example, count_field for count,
-   *  and sum_field for sum.
+   *  The output name for each field. If non specified will use the format `window_op_field` for example, `count_field` for count and `sum_field` for sum.
```
kanitw: Can you make as required by removing question mark below?   (Yes  -- we should  still autogenerate it if `as` is missing, but JSON schema to enforce people to specify it, so that people write better specs.)

src/transform.ts
```diff
@@ -146,12 +143,12 @@ export interface WindowTransform {
   ignorePeers?: boolean;
 
   /**
-   * The fields to group by.
+   * The names of the data fields to partioin the objects into seprate windows. If not specified, everything will be in a single group.
```
kanitw: The data fields for partitioning the data objects into separate windows.  If unspecified, all data points will be a single group.

src/transform.ts
```diff
@@ -146,12 +143,12 @@ export interface WindowTransform {
...
    */
   groupby?: string[];
 
   /**
-   * The definitions of how to sort each of the fields in the window.
+   * A definition for sorting the objects within the window. Equivalent objects are considered a peer (Look at ignorePeers). If left undefined, the order of items in the window is undefined.
```
kanitw: A definition for sorting the data objects  (add data)

if left undefined => If undefined

src/transform.ts
```diff
@@ -110,20 +111,17 @@ export interface WindowFieldDef {
   op: AggregateOp | WindowOnlyOp;
 
   /**
-   *  Parameter values for the window functions. Parameter value can be null for operations that do not accept a
-   *  parameter.
+   *  Parameter values for the window functions. Parameter value can be omitted for operations that do not accept a parameter.
```
kanitw: Please refer to the operation / parameter table.  

Also, I want to be able to read this and know if it is ever required. 

- For ones that don't have param, then obviously this property will be "ignored" -- not just "can be omitted"
- For ones that do have param, are they ever required, or are they always optional?

Please make this clear  (Note that Vega docs isn't very clear in this aspect too, please experiment and help jeff fix Vega docs :p)

src/transform.ts
```diff
@@ -4,8 +4,9 @@ import {Data} from './data';
 import {LogicalOperand, normalizeLogicalOperand} from './logical';
 import {normalizePredicate, Predicate} from './predicate';
 import {TimeUnit} from './timeunit';
-import {WindowOnlyOp} from './window';
 
+export type WindowOnlyOp = 'row_number' | 'rank' | 'dense_rank' | 'percent_rank' | 'cume_dist'
```
kanitw: Put this right above `WindowTransform`?

src/transform.ts
```diff
@@ -135,8 +133,7 @@ export interface WindowTransform {
   window: WindowFieldDef[];
 
   /**
-   * The frame for the window, if none is set the default is `[null, 0]` everything before the
-   * current item.
+   * A two element specification about how large the sliding window is. The first element indicates the start and the second element indicates the end. If `null` is specified for the start, it will include everything before the current point. If `null` is specified for the end, it will include everything after the endpoint. For example a frame of `[-5,5]` says the window should include 5 previous objects and 5 after objects. The default is `[null, 0]`, which means include everything in the window. `[null, null]` would mean include everything in the window.
```
kanitw: > The default is `[null, 0]`, which means include everything in the window. `[null, null]` would mean include everything in the window. 

This is inaccurate.  Your text basically says that `[null, 0]` and `[null,null]` are the same.   Why not just copy text from  Vega docs https://vega.github.io/vega/docs/transforms/window/.

- [ ] Maybe use "data objects" over just "objects" though
- [ ] Another thing to change is that we always use the following format for default values:

```

__Default value:__:  `[null, 0]` (includes the current object and all preceding objects)
```

(Please make sure that you use this format for all properties.)

site/docs/transform/window.md
```diff
@@ -0,0 +1,131 @@
...
+| Property    | Type | Description  |
+| :----------- | :-------: | :------------|
+| op | AggregateOp or WindowOnlyOp | See supported operations in the table below.  |
+| param | number | Parameter values for the window functions. Parameter value can be null for operations that do not accept a parameter. |
+| field | string | The data field for which to compute the aggregate or window function. This can be null for functions that do not operate over a field such as `count`, `rank`, `dense_rank`. |
+| as | string | The output name for each field. If non specified will use the format `window_op_field` for example, `count_field` for count and `sum_field` for sum. |
+
+## <a name="ops"></a> Window Only Operation Reference
+
+The valid operations include all [valid aggregate operations](../aggregate/#ops) plus the following window operations.
```
kanitw: broken link  -- I think `aggregate.html#ops` is the right one, but please run `yarn run site` to check.

kanitw:
Note that I change merge target to a new feature branch, but some comments (like [this one]) got collapsed, but still need to be addressed.

examples/specs/window_transform_row_number.vl.json
AkshatSh: I dropped this file, and added a new one which compares the ranks of the top 5 students on an exam.

examples/specs/window_transform_time_spent_in_class.vl.json
AkshatSh: Dropped this example.

src/transform.ts
AkshatSh: Sorry, I thought the default was `[null, null]` realized it was `[null, 0]` so I quickly changed it, but forgot to update the description. I'll update this.

AkshatSh:
Resolved all the comments. I also added a couple examples that will hopefully show case this transform better. I have also displayed all the examples, and their vega-lite specs in the window.md documentation file.

Let me know your thoughts!

examples/specs/window_transform_movie_mean_difference.vl.json
AkshatSh: Tried implementing this, by only displaying movies where the rating was 2.5 points higher than the average rating. This significantly reduced the number of titles being displayed.

src/transform.ts
```diff
@@ -103,6 +100,20 @@ export interface AggregatedFieldDef {
   as: string;
 }
 
+
+export type WindowOnlyOp =
+  |'row_number'
```
kanitw: Why starting with |

src/transform.ts
```diff
@@ -111,19 +122,19 @@ export interface WindowFieldDef {
   op: AggregateOp | WindowOnlyOp;
 
   /**
-   *  Parameter values for the window functions. Parameter value can be omitted for operations that do not accept a parameter.
+   *  Please refer to the operation/parameter table. Parameter values for the window functions. Parameter value can be omitted for operations that do not accept a parameter. The spec will be invalid if any function that requires a parameter is unspecified.
```
kanitw: > Please refer to the operation/parameter table. 

It's weird description to say refer first.  Better describe what it is first and say this later. 

> The spec will be invalid if any function that requires a parameter is unspecified.

This seems too extreme.  It should just be ignored without failing.

src/transform.ts
```diff
@@ -133,22 +144,26 @@ export interface WindowTransform {
...
-   * A two element specification about how large the sliding window is. The first element indicates the start and the second element indicates the end. If `null` is specified for the start, it will include everything before the current point. If `null` is specified for the end, it will include everything after the endpoint. For example a frame of `[-5,5]` says the window should include 5 previous objects and 5 after objects. The default is `[null, 0]`, which means include everything in the window. `[null, null]` would mean include everything in the window.
+   * A frame specification as a two-element array indicating how the sliding window should proceed. The array entries should either be a number indicating the offset from the current data object, or null to indicate unbounded rows preceding or following the current data object. The default value is `[null, 0]`, indicating that the sliding window includes the current object and all preceding objects. The value `[-5, 5]` indicates that the window should include five objects preceding and five objects following the current object. Finally, `[null, null]` indicates that the window frame should always include all data objects.
```
kanitw: >  The default value is `[null, 0]`, indicating 

For example, `[null, 0]` indicates that ...

(No need to repeat "default" twice -- If you haven't do `yarn run site` to check the page, please do as read through the docs as if you're a user and see if everything make sense.)

src/transform.ts
```diff
@@ -133,22 +144,26 @@ export interface WindowTransform {
...
+   *
+   * __Default value:__:  `[null, 0]` (includes the current object and all preceding objects)
    */
   frame?: (null | number)[];
 
   /**
-   * Will indicate whether to ignore peer values (items with the same rank) in the window. The default value is `False`.
+   * Will indicate whether to ignore peer values (items with the same rank) in the window.
```
kanitw: The Vega docs is more descriptive.  Not sure why you have to shorten here.

src/transform.ts
```diff
@@ -133,22 +144,26 @@ export interface WindowTransform {
...
 
   /**
-   * The names of the data fields to partioin the objects into seprate windows. If not specified, everything will be in a single group.
+   * The data fields for partitioning the data objects into separate windows. If unspecified, all data points will be a single group.
    */
   groupby?: string[];
 
   /**
-   * A definition for sorting the objects within the window. Equivalent objects are considered a peer (Look at ignorePeers). If left undefined, the order of items in the window is undefined.
+   * A definition for sorting the data objects within the window. Equivalent objects are considered a peer (Look at ignorePeers). If undefined, the order of items in the window is undefined.
```
kanitw: Why "Look at ignorePeers" here?  (Looking at ignorePeers doesn't seem to make me understand `sort` better.)

Also, the description of this field in Vega docs is more descriptive.

src/transform.ts
```diff
@@ -133,22 +144,26 @@ export interface WindowTransform {
...
-   * Will indicate whether to ignore peer values (items with the same rank) in the window. The default value is `False`.
+   * Will indicate whether to ignore peer values (items with the same rank) in the window.
+   *
+   * __Default value:__ `false'
    */
   ignorePeers?: boolean;
 
   /**
-   * The names of the data fields to partioin the objects into seprate windows. If not specified, everything will be in a single group.
+   * The data fields for partitioning the data objects into separate windows. If unspecified, all data points will be a single group.
```
kanitw: If unspecified, a single group containing all data objects will be used.

examples/specs/window_transform_average_rating.vl.json
```diff
@@ -7,7 +7,8 @@
         "values": [
             { "name": "Movie 1", "Rating": 9, "Year": 2016 },
             { "name": "Movie 2", "Rating": 3, "Year": 2016 },
-            { "name": "Movie 3", "Rating": 5, "Year": 2015 }
+            { "name": "Movie 3", "Rating": 5, "Year": 2015 },
+            { "name": "Movie 4", "Rating": 2, "Year": 2015 }
         ]
     },
     "layer": [{
```
kanitw: Not sure why you use `layer` and only have one layer. 

That said, a better plot would be overlaying it with ticks showing `AverageYearRating`.

examples/specs/window_transform_movie_mean_difference.vl.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
-    "description": "Bar graph showing how each movie is different from the average movie rating",
+    "description": "Bar graph showing how each of the 'really good' movies is different from the average movie rating. Really good is defined to differ from the average movie rating by more than 2.5 points.",
```
kanitw: This plot would benefit from another layer of rule that show global average rating (and perhaps another layer of text to label the average)

site/docs/transform/window.md
```diff
@@ -69,52 +69,171 @@ To replace any of the examples from total to `mean`, the operation just needs to
...
         "field": "x",
-        "as": "total"
+        "as": "average"
     }],
     "frame": [null, null]
   },
   {
-    "calculate": "(datum.total - datum.x)/dataum.total",
-    "as": "Percent difference from total"
+    "calculate": "(datum.total - datum.x)/dataum.average * 100",
```
kanitw: "dataum"

also the example above doesn't have field "x"

Please make sure to organize sections in a sensible way.

site/docs/transform/window.md
```diff
@@ -69,52 +69,171 @@ To replace any of the examples from total to `mean`, the operation just needs to
...
+    "as": "Percent difference from average"
   }
-```
\ No newline at end of file
+```
+
+## Other examples
+
+Here is an example where the window_transform can be used to get the average rating for a film during different years.
```
kanitw: Why is this "other example"?   This is another example of difference from mean.   Seems like the  DIV tag is also pointing to the same example as above.  

I'd say group this in the `### Difference from mean` section.  

In fact, I'd say that remove the "Movie 1, Movie 2, Movie 3, Movie 4" example as it's as interesting as the real one. 

If your point is to show that you can group by year, you can create multiple variations of the  same example like:

- [ ] no group by and use the value to filter
- [ ] group by year and use the value to filter
- [ ] no group by, but visualize residual  (rating - mean_rating) instead of just filtering. 

This way users can easily see the changes without having to totally learn about a totally different example.

src/transform.ts
AkshatSh: I was following the example for aggregate op in transform.d.ts, to make the style more consistent.

```typescript
export type AggregateOp =
  | 'argmax'
  | 'argmin'
  | 'average'
  | 'count'
  | 'distinct'
  | 'max'
  | 'mean'
  | 'median'
  | 'min'
  | 'missing'
  | 'q1'
  | 'q3'
  | 'ci0'
  | 'ci1'
  | 'stderr'
  | 'stdev'
  | 'stdevp'
  | 'sum'
  | 'valid'
  | 'values'
  | 'variance'
  | 'variancep';
```

kanitw: Interesting.   I don't know what's the rationale for `transform.d.ts` though. 
I'd argue that this is a bad style, so we shouldn't follow it.

domoritz: That's the style in vega-typings. It might be automatically generated by prettier over there. You can ignore the style for Vega-Lite.

kanitw:
Test now failing

examples/specs/window_transform_mean_difference_by_year.vl.json
```diff
@@ -0,0 +1,45 @@
...
+            "calculate": "datum.IMDB_Rating - datum.AverageYearRating",
+            "as": "RatingDelta"
+        },
+        {
+            "filter": "datum.RatingDelta > 2.5"
+        }
+    ],
+    "mark": { "type": "bar", "clip": true },
+    "encoding": {
+        "x": {
```
kanitw: Try either transposing the plot (swap x and y) or adjust x's axis labelAngle so it's easier to read.  (Same for the other plot.)

kanitw:
- [ ] Please manually reformat examples to have nice tabbing and only line breaks if the line would be too long if we don’t.   (Autoformatter usually add too much news lines, making it harder to read.)

- [ ] Don’t add space before and after braces (edited)

- [ ] Some texts are missing — Make some introduction before showing specs.

(The examples intro right now is “To replace any of the examples from total to mean, the operation just needs to be changed from sum to mean.” — this is more like a note that better comes at the end, not an intro.  Also, you can’t just replace sum with mean.  For example,  what is “percentage of mean”? )
- [ ] Percent Difference from mean — it’s nice that you add this section with diff — but why not also adding another example?

kanitw:
- [ ] Rename files

Why `window_transform_movie_mean_difference`  has `movie_` in the name, but `window_transform_mean_difference_by_year` does not -- just remove `movie_` for shorter names?


Also, change `window_transform` to `window_` ?  (Right now the names are SO long.)

Issue #3204, opened 2017-11-20
Sizing facet label size with guide-label fails


eibanez:
I am trying to format all titles and labels in the plots, using `guide-title` and `guide-label` as recommended in the documentation.

I haven't done exhaustive checks but from the example below you can see that the size for facet row and column label names do not respond to the size property.

![image](https://user-images.githubusercontent.com/438494/33019588-95522562-cdc0-11e7-9e38-c9c03474b952.png)

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "name": "trellis_barley",
  "data": {"url": "data/barley.json"},
  "mark": "point",
  "encoding": {
    "row": {
      "field": "site", "type": "ordinal"
    },
    "column": {
      "field": "year", "type": "ordinal"
    },
    "x": {
      "field": "yield", "type": "quantitative",
      "scale": {"zero": false}
    },
    "y": {
      "field": "variety", "type": "ordinal",
      "scale": {"rangeStep": 12}
    },
    "color": {"field": "year", "type": "nominal"}
  },
  "config": {
    "style": {
      "guide-title": {
            "fill": "red",
            "fontSize": 15,
            "fontStyle": "italic"
        },
      "guide-label": {
            "fill": "blue",
            "fontSize": 15,
            "fontStyle": "italic"
        }
    }
  }
}
```

kanitw:
Nice catch / thanks for reporting.  I will make a PR to fix this.

kanitw:
Will be fixed in https://github.com/vega/vega-lite/pull/3205

eibanez:
Thanks for the quick fix!

PR #3205, opened 2017-11-21
Fix bug that prevent users from using `guide-label` style to customize header labels

kanitw:
Fix https://github.com/vega/vega-lite/issues/3204

### DIFF
--- a/src/compile/layout/header.ts
+++ b/src/compile/layout/header.ts
@@ -3,6 +3,7 @@
  */
 import {FacetFieldDef} from '../../facet';
 import {field} from '../../fielddef';
+import {keys} from '../../util';
 import {AxisOrient, VgAxis} from '../../vega.schema';
 import {formatSignalRef} from '../common';
 import {Model} from '../model';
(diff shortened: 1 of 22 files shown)


src/compile/layout/header.ts
```diff
@@ -96,13 +96,11 @@ export function getHeaderGroup(model: Model, channel: HeaderChannel, headerType:
...
         encode: {
           update: {
-            fontWeight: {value: 'normal'},
-            angle: {value: 0},
-            fontSize: {value: 10}, // default label font size
             ... (channel === 'row' ? {
               align: {value: 'right'},
               baseline: {value: 'middle'}
             } : {})
+            // TODO: apply label* (e.g, labelAlign, labelBaseline) here
```
domoritz: Is there an issue for this?

domoritz:
Some tests are failing

screen shot 2017-11-22 at 08 58 31

kanitw:
Hmm, this is weird.  I wonder if `keys(update).length` behave differently for linux.

kanitw:
Should be good now.  The lower diff hit is actually beyond the scope of this PR.

PR #3211, opened 2017-11-29
Fix ref: supported-aggregation-operations → ops

devgru:
Just fixing ref in documentation:
https://vega.github.io/vega-lite/docs/aggregate.html#ops


domoritz:
Thank you @devgru!

PR #3214, opened 2017-11-30
Fix encoding/sort doc, replace "none" with "null" (as in code)

devgru:
Another typo in docs, compare with
https://github.com/vega/vega-lite/blob/master/src/sort.ts#L4


kanitw:
@devgru Thanks so much!

PR #3219, opened 2017-12-01
DOCS: Update transform.ts

nyurik:
doc update - hope the link works


kanitw:
@domoritz -- maybe we should move "Vega expression" to [the parameter types](https://vega.github.io/vega-lite/docs/types.html) page and add a special roles to such strings?  

Having it in "filter" page is a bit weird as it's used in many places.

domoritz:
@kanitw I agree.

domoritz:
Done in #3288

PR #3221, opened 2017-12-03
Correctly disable sorting when sort = null

### LINKED ISSUE #3215: Cannot disable sorting layers
From https://stackoverflow.com/questions/47570551/sorting-for-unioned-domains-in-vega-lite

In the spec below, we should not sort the unioned domain. 

```json
{
  "width": 500,
  "height": 300,
  "data": {
        "values": [
            {
                "a": "A",
                "b": 70
            },
            {
                "a": "Z",
                "b": 50
            },
            {
                "a": "B",
                "b": 34
            }
        ]
    },
    "layer": [
        {
            "mark": "bar",
            "encoding": {
                "y": {
                    "field": "a",
                    "type": "ordinal",
                    "sort": null
                },
                "x": {
                    "field": "b",
                    "type": "quantitative"
                },
                "color": {
                    "field": "a",
                    "type": "nominal"
                }
            }
        },
        {
            "mark": "text",
            "encoding": {
                "y": {
                    "field": "a",
                    "type": "ordinal",
                    "sort": null
                },
                "text": {
                    "field": "b",
                    "type": "quantitative"
                }
            }
        }
    ]
}
```

domoritz:
Fixes #3215

### DIFF
--- a/src/compile/scale/domain.ts
+++ b/src/compile/scale/domain.ts
@@ -208,7 +208,7 @@ function parseSingleChannelDomain(scaleType: ScaleType, domain: Domain, model: U
         // Use range if we added it and the scale does not support computing a range as a signal.
         field: model.field(channel, binRequiresRange(fieldDef, channel) ? {binSuffix: 'range'} : {}),
         // we have to use a sort object if sort = true to make the sort correct by bin start
-        sort: sort === true || !isSortField(sort) ? {
+        sort: sort !== false && (sort === true || !isSortField(sort)) ? {
           field: model.field(channel, {}),
           op: 'min' // min or max doesn't matter since we sort by the start of the bin range
         } : sort
@@ -232,7 +232,7 @@ function parseSingleChannelDomain(scaleType: ScaleType, domain: Domain, model: U
         }];
       }
     }
-  } else if (sort) {
+  } else if (sort !== undefined) {
     return [{
       // If sort by aggregation of a specified sort field, we need to use RAW table,
       // so we can aggregate values for the scale independently from the main aggregation.
@@ -249,7 +249,7 @@ function parseSingleChannelDomain(scaleType: ScaleType, domain: Domain, model: U
 }
 
 
-export function domainSort(model: UnitModel, channel: ScaleChannel, scaleType: ScaleType): true | SortField<string> {
+export function domainSort(model: UnitModel, channel: ScaleChannel, scaleType: ScaleType): boolean | SortField<string> {
   if (!hasDiscreteDomain(scaleType)) {
     return undefined;
   }
@@ -273,8 +273,8 @@ export function domainSort(model: UnitModel, channel: ScaleChannel, scaleType: S
     return true;
   }
 
-  // sort === 'none'
-  return undefined;
+  // sort == null
+  return false;
 }
 
 
--- a/src/compile/scale/domain.ts
+++ b/src/compile/scale/domain.ts
@@ -1,4 +1,3 @@
-import {isString} from 'vega-util';
 import {SHARED_DOMAIN_OP_INDEX} from '../../aggregate';
 import {binToString} from '../../bin';
 import {isScaleChannel, ScaleChannel} from '../../channel';
@@ -16,7 +15,7 @@ import {
   VgFieldRefUnionDomain,
   VgNonUnionDomain,
   VgSortField,
-  VgUnionSortField,
+  VgUnionSortField
 } from '../../vega.schema';
 import {isDataRefUnionedDomain, isFieldRefUnionDomain} from '../../vega.schema';
 import {binRequiresRange} from '../common';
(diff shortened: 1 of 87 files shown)


src/compile/scale/domain.ts
```diff
@@ -208,7 +208,7 @@ function parseSingleChannelDomain(scaleType: ScaleType, domain: Domain, model: U
         // Use range if we added it and the scale does not support computing a range as a signal.
         field: model.field(channel, binRequiresRange(fieldDef, channel) ? {binSuffix: 'range'} : {}),
         // we have to use a sort object if sort = true to make the sort correct by bin start
-        sort: sort === true || !isSortField(sort) ? {
+        sort: sort !== false && (sort === true || !isSortField(sort)) ? {
```
kanitw: This is weird.  Shouldn't sort be true by default? 
If so, shouldn't this be `sort !== false && !isSortField(sort)`?

kanitw: Actually if sort is true by default and we should allow using null and false (just in case of user error), I guess this should be `sort !== null && sort !== false && !isSortField(sort)`?

domoritz: The first comment doesn't work because in `sort !== false && !isSortField(sort)` the second `sort` could be `true` and that doesn't pass type checking. 

Also no to the second comment! Because `sort` comes from `domainSort` and is thus already cleaned up.

src/compile/scale/domain.ts
```diff
@@ -273,8 +273,8 @@ export function domainSort(model: UnitModel, channel: ScaleChannel, scaleType: S
     return true;
   }
 
-  // sort === 'none'
-  return undefined;
+  // sort == null
+  return false;
```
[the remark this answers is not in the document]
domoritz: Good point. I'll fix that.

domoritz:
Hmm, okay. So making `sort: false` explicit makes it much easier for me to merge domains. The problem is that if we use `undefined`, then I would automatically sort domains even if the user explicitly disables sorting for one domain. I guess that's okay but just flagging this.

domoritz:
Updated. Note that I removed sorting from all continuous scales.

domoritz:
I don't know why the SVGs change.

domoritz:
Ping @kanitw

kanitw:
Please isolate the actual svg changes

domoritz:
Done. I am setting this PR to be against a new branch.

test/compile/scale/domain.test.ts
```diff
@@ -474,86 +474,80 @@ describe('compile/scale', () => {
...
+      assert.deepEqual(domain, {
         data: 'foo',
-        fields: ['a', 'b'],
-        sort: true
+        fields: ['a', 'b']
       });
     });
 
-    it('should maintain sort even if false', () => {
+    it('should ignore order: ascending', () => {
```
kanitw: Can you explain why it should ignore ascending?

test/compile/scale/domain.test.ts
```diff
@@ -474,86 +474,80 @@ describe('compile/scale', () => {
...
+        fields: ['a', 'b']
       });
     });
 
     it('should merge domains with different data', () => {
       const domain = mergeDomains([{
         data: 'foo',
-        field: 'a'
+        field: 'a',
+        sort: true
```
kanitw: why are `sort: true` ignored here?  

Perhaps, update the test description (`it`) to explain why?

test/compile/scale/domain.test.ts
domoritz: Because it is the default.

test/compile/scale/domain.test.ts
domoritz: How are they ignored? The sorted domains are still sorted.

domoritz:
> Also, I think these issues are quite delicate so it's better to make this PR pass test coverage too.

Agreed!

domoritz:
The test coverage change is unrelated to this PR. I filed https://github.com/vega/vega-lite/issues/3248 separately.

src/compile/scale/domain.ts
```diff
@@ -431,10 +423,10 @@ export function getFieldFromDomain(domain: VgDomain): string {
     }
     log.warn('Detected faceted independent scales that union domain of identical fields from different source detected.  We will assume that this is the same field from a different fork of the same data source.  However, if this is not case, the result view size maybe incorrect.');
     return field;
-  } else if (isFieldRefUnionDomain(domain) && isString) {
+  } else if (isFieldRefUnionDomain(domain) && util.isString(domain)) {
```
domoritz: @kanitw Note that this was definitely wrong before.

domoritz:
Ping @kanitw

src/compile/scale/domain.ts
kanitw: lol

PR #3222, opened 2017-12-03
Replace ‘none’ -> null in stacking tests.

domoritz:
We don't support stacking: 'none' anymore.


kanitw:
Doesnt hurt to keep the test for backward compatibility? Idk.  It still works, but i dont include it in schema anymore

domoritz:
I got a bit confused as we are not testing `null`, though. Also, it's nice to be able to search for `"none"` and not find anything ;-)

kanitw:
You can  just add null?

domoritz:
Done

Issue #3220, opened 2017-12-01
tooltip encoding breaks line mark


nyurik:
This graph works, but try changing mark to `"line"`, and everything disappears.  Also, try `"area"` -- there is some strange drawing artifacts.

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "url": "https://gist.githubusercontent.com/nyurik/daa888a0dfa7d1a3a15b2e57e2b1c4e4/raw/5c8904f85e2674bffb364c5049d6e5627ddb2172/logstashrandom.json",
    "format": {
      "property": "aggregations.time_buckets.buckets",
      "type": "json"
    }
  },
  "transform": [
    {
      "calculate": "if(datum.doc_count > 0, format(datum.doc_count, ',d') + ' documents', 'no documents')",
      "as": "tip"
    }
  ],
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "key",
      "type": "temporal"
    },
    "y": {
      "field": "doc_count",
      "type": "quantitative"
    },
    "tooltip": {
      "field": "tip"
    }
  },
  "autosize": {
    "type": "fit",
    "contains": "padding"
  }
}
```

domoritz:
Tooltip is adding another group by clause. I don't think there is anything we can do in Vega-Lite. The easiest solution for you is to add a second layer with point marks and only add tooltips to the point marks.

domoritz:
Closing as there is no immediate action here.

kanitw:
I disagree.  Since this is a raw plot, I think we have a bug in terms of determining line groupings

kanitw:
Will be fixed in #3224

PR #3224, opened 2017-12-06
Prevent tooltip from accidentally grouping lines + Refactor Path grouping Field

kanitw:


kanitw:
Example in #3220 is fixed:

image

kanitw:
Actually, I start wondering if this is a good idea.  Basically, the way Vega currently works, one group of line always produce one actual mark (and thus can only have one tooltip. 

There are two options:

a) Make tooltip do not group lines (like this PR does), but throw warning that tooltip will use value from the first item in the line grouping. 

b) Make tooltip group line like before, but throw warning that line grouping is happening due to tooltip.

I still think a) is a better option as Vega might become smarter with line marks in the future. 

Any thoughts?

domoritz:
Hmm, this is a tough one but I think you are right that a) is less surprising. I'd still throw a warning that we only use the first value. And fix the tests :-P

kanitw:
I just make the test passed.  However, I now start wondering if the error should be thrown at the Vega level (instead of Vega-Lite) when the tooltip channel is used for line marks and there are multiple tooltip values for each line. 

cc: @jheer

domoritz:
Approved. Merge if you feel that this is right.

PR #3228, opened 2017-12-08
Reduce size of npm packages

### LINKED ISSUE #3227: Vega lite release package is humongous
The node_modules/vega-lite dir is 34.1 MB! For comparison, Vega is 2.1MB, and most other well known packages are a tiny fraction of that. Recently, github reduced the size of the release files, which means I can no longer publish release packages that contain the entire node_modules.   BTW, this also means that anyone using vega lite lib has to download the whole thing when installing it, even though most of the code is just a few hundred KB.  

image

domoritz:
Fixes #3227 

```
$ du -sh package
7.5M	package

$ du -sh package/*
284K	package/_data
 12K	package/bin
5.3M	package/build
1.2M	package/examples
4.0K	package/LICENSE
8.0K	package/package.json
4.0K	package/README.md
788K	package/src
 12K	package/typings
```


domoritz:
With the last commit

```
$ du -sh package
6.1M	package

$du -sh package/*
 12K	package/bin
5.3M	package/build
4.0K	package/LICENSE
8.0K	package/package.json
4.0K	package/README.md
788K	package/src
 12K	package/typings
```

domoritz:
We use git anyway beacuse Vega doesn't ship examples either: https://github.com/vega/editor/blob/master/package.json#L73.

> and everything else that depends on these examples
That's why I'm asking you because I wasn't sure who uses examples.

kanitw:
Would this work with the old editor, which we still use as the editor for CompassQL.

domoritz:
> Would this work with the old editor, which we still use as the editor for CompassQL?

It won't copy the examples but you can do that manually. Alternatively, just switch to github urls like we do in the new editor.

PR #3229, opened 2017-12-09
Support href and cursor properties for marks

domoritz:


src/vega.schema.ts
```diff
@@ -1056,6 +1056,18 @@ export interface VgMarkConfig {
...
+   * A URL to load upon mouse click. If defined, the mark acts as a hyperlink.
+   *
+   * @format uri
+   */
+  href?: string;
+
+  /**
+   * The mouse cursor used over the mark. Any valid [CSS cursor type](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#Values) can be used.
+   */
+  cursor?: string;
```
kanitw: The link you provided lists all cursor type.  Perhaps, make it an enum here?

auto | default | none | context-menu | help | pointer | progress | wait | cell | crosshair | text | vertical-text | alias | copy | move | no-drop | not-allowed | e-resize | n-resize | ne-resize | nw-resize | s-resize | se-resize | sw-resize | w-resize | ew-resize | ns-resize | nesw-resize | nwse-resize | col-resize | row-resize | all-scroll | zoom-in | zoom-out | grab | grabbing

src/vega.schema.ts
```diff
@@ -1084,10 +1096,11 @@ const VG_MARK_CONFIG_INDEX: Flag<keyof VgMarkConfig> = {
   font: 1,
   fontSize: 1,
   fontWeight: 1,
-  fontStyle: 1
+  fontStyle: 1,
+  href: 1,
+  cursor: 1,
```
kanitw: I'm not sure why previously list `cursor` as "vg channel that do not have mark config".  Can you test if cursor works below for sure?

Btw, does it really make sense to have `href` as a mark config too?

site/docs/mark/mark.md
```diff
@@ -125,11 +125,16 @@ The rest of this section describe groups of properties supported by the `mark` c
...
 {% include table.html props="opacity,fillOpacity,strokeOpacity" source="MarkConfig" %}
 
-
 ### Stroke Style
 
 {% include table.html props="strokeWidth,strokeDash,strokeDashOffset" source="MarkConfig" %}
 
+### Link Properties
+
+Marks can act as hyperlinks when the `href` property is defined.
```
kanitw: - `href` property -> `href` channel

- Maybe also add:

A `cursor` channel can also be provided to serve as affordance for the links.

domoritz: Done

src/vega.schema.ts
domoritz: Okay. I don't know whether browser vendors will add more in the future, though.

src/vega.schema.ts
domoritz: Working on it

domoritz:
I will finish this in a few days. Thanks for the review.

src/vega.schema.ts
domoritz:
```
{
  "data": {
    "values": [
    ]
  },
  "mark": "bar",
  "encoding": {
  },
  "config": {
    "bar": {
      "cursor": "pointer"
    }
  }
}
```

works

domoritz:
Ready

site/docs/mark/mark.md
```diff
@@ -125,11 +125,16 @@ The rest of this section describe groups of properties supported by the `mark` c
...
 {% include table.html props="opacity,fillOpacity,strokeOpacity" source="MarkConfig" %}
 
-
 ### Stroke Style
 
 {% include table.html props="strokeWidth,strokeDash,strokeDashOffset" source="MarkConfig" %}
 
+### Link Properties
+
+Marks can act as hyperlinks when the `href` channel is defined. A `cursor` channel can also be provided to serve as affordance for the links.
```
kanitw: If you don't add them as channels, why the docs say that they are channels here?

domoritz: Ups.

site/docs/encoding.md
```diff
@@ -180,27 +180,27 @@ In addition to the constant `value`, [value definitions](#value-def) of mark pro
 See [the `condition`](condition.html) page for examples how to specify condition logic.
 
 {:#text}
-## Text and Tooltip Channels
+## Text, Tooltip, and HREF Channels
```
kanitw: href should be on its own category -- it doesn't have `format` property

src/encoding.ts
```diff
@@ -98,6 +98,11 @@ export interface Encoding<F> {
    */
   tooltip?: FieldDefWithCondition<TextFieldDef<F>> | ValueDefWithCondition<TextFieldDef<F>>;
 
+  /**
+   * A URL to load upon mouse click.
+   */
+  href?: FieldDefWithCondition<TextFieldDef<F>> | ValueDefWithCondition<TextFieldDef<F>>;
```
kanitw: It shouldn't be TextFieldDef?

site/docs/encoding.md
domoritz: Hmm, I was thinking that one might want to format a number to be a URL like `www.site.com?page=${id}`.

src/encoding.ts
domoritz: What if you want a conditional url?

kanitw: Well, TextFieldDef and conditional are orthogonal.  TextFieldDef has `format` that href shouldn't

site/docs/encoding.md
kanitw: Yeah but having `format` wouldn't help you do that anyway?

https://github.com/d3/d3-format#locale_format

kanitw: oh I see -- you can use "fill"

kanitw: Actually you can't .. fill is for a slightly different purpose

domoritz: You're right. I was thinking of dates where you can do something like `foo %H bar`

src/encoding.ts
domoritz: Fixed and rebased.

site/docs/encoding.md
```diff
@@ -204,6 +208,29 @@ In addition to the constant `value`, [value definitions](#value-def) of `text` a
 
 {% include table.html props="condition" source="TextValueDefWithCondition" %}
 
+{:#href}
+## Hyperlink Channel
+
+By setting the `href` channel, a mark becomes a hyperlink. The specified URL is loaded upon a muse click. The `cursor` mark property can be set to `pointer` to serve as affordance for hyperlinks.
```
kanitw: add a link for "The `cursor` mark property"?

test/compile/mark/mark.test.ts
```diff
@@ -120,6 +120,21 @@ describe('Mark', function() {
         assert.equal(markGroup[0].encode.update.tooltip.value, 'foo');
       });
     });
+
+    describe('Bar with href', () => {
```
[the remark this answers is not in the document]
domoritz: Okay. I'm moving the tooltip and href tests.

domoritz:
Done with revisions

src/channel.ts
```diff
@@ -128,7 +128,8 @@ export type PositionScaleChannel = typeof POSITION_SCALE_CHANNELS[0];
 // NON_POSITION_SCALE_CHANNEL = SCALE_CHANNELS without X, Y
 const {
   // x2 and y2 share the same scale as x and y
-  // text, tooltip, and href have format instead of scale
+  // text and tooltip have format instead of scale,
+  // href has neother format, nor scale
```
kanitw: neither

domoritz:
cc @rdeline

PR #3230, opened 2017-12-09
Fix title alignment

domoritz:


domoritz:
No idea why we get all these minor diffs in the SVGs.

kanitw:
With the SVGs diff, it's pretty much impossible to review.  Can you isolate the SVG changes?

domoritz:
The PR now has fewer SVG changes

Issue #3232, opened 2017-12-10
Copy disaster chart to examples


domoritz:
https://bl.ocks.org/domoritz/8afa8f0ce5e03b43497e821f7a9cf216. It's a great demonstration of an ordinal bubble chart.

kanitw:
You add to blocks in #3233, but if you want this to be core examples, why not adding it to examples folder?

domoritz:
I thought this would be a great task for one of our ugrads to learn how to add data and specs.

PR #3233, opened 2017-12-10
Add some of my block examples

domoritz:
I think the disaster chart should become one of our core examples: https://github.com/vega/vega-lite/issues/3232.


arvind:
Agreed re:disaster chart.

Issue #3166, opened 2017-10-25
Clamp whiskers to min and max


domoritz:
See https://github.com/vega/vega-lite/issues/3163

kanitw:
@jwoLondon @domoritz  -- Can you post a reference that says that Tukey's original box plot supposes to do this clamping? I don't remember seeing one.

mattwchun:
@kanitw it says here: https://www.graphpad.com/support/faq/five-ways-to-plot-whiskers-in-box-and-whisker-plots/

kanitw:
@mattwchun  Thanks.  This means that it makes sense for you to fix this issue then :)

kanitw:
- [ ] This also makes Tukey box plot a better default than min-max box-plot
- [ ] I think it also makes sense to add the outliers by default.  (Perhaps add a flag called `outliers: boolean` in the BoxPlot definition to allow disabling the outliers layer.)

kanitw:
Actually, the outlier is a bit more complicated to do, so let's ignore that comment for now.

g3o2:
outliers are best added "manually" using filter transform on a separate layer...

kanitw:
> outliers are best added "manually" using filter transform on a separate layer...

Yes, but you have to filter by the aggregated value, and there is no easy way to do that automatically yet until we have the (pending) window transform.

kanitw:
Done in #3237 -- other subtasks are now forked as new issues.

PR #3237, opened 2017-12-19
Clamps the whiskers of the Boxplot

mattwchun:
Refers to issue #3166. 

upperwhisker = min(iqr *k + q3, max of continuous axis field)
lowerwhisker = max(q1 - iqr *k, min of continuous axis field)

### DIFF
--- a/examples/compiled/box-plot_minmax_2D_horizontal_IQR.vg.json
+++ b/examples/compiled/box-plot_minmax_2D_horizontal_IQR.vg.json
@@ -21,33 +21,39 @@
                     "ops": [
                         "q1",
                         "q3",
-                        "median"
+                        "median",
+                        "min",
+                        "max"
                     ],
                     "fields": [
+                        "people",
+                        "people",
                         "people",
                         "people",
                         "people"
                     ],
                     "as": [
-                        "lowerBox",
-                        "upperBox",
-                        "midBox"
+                        "lower_box_people",
+                        "upper_box_people",
+                        "mid_box_people",
+                        "min_people",
+                        "max_people"
                     ]
                 },
                 {
                     "type": "formula",
-                    "expr": "datum.upperBox - datum.lowerBox",
-                    "as": "IQR"
+                    "expr": "datum.upper_box_people - datum.lower_box_people",
+                    "as": "iqr_people"
                 },
                 {
                     "type": "formula",
-                    "expr": "datum.lowerBox - datum.IQR * 1.5",
-                    "as": "lowerWhisker"
+                    "expr": "min(datum.upper_box_people + datum.iqr_people * 1.5, datum.max_people)",
+                    "as": "upper_whisker_people"
                 },
                 {
                     "type": "formula",
-                    "expr": "datum.upperBox + datum.IQR * 1.5",
-                    "as": "upperWhisker"
+                    "expr": "max(datum.lower_box_people - datum.iqr_people * 1.5, datum.min_people)",
+                    "as": "lower_whisker_people"
                 }
             ]
         },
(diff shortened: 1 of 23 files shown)


kanitw:
- [ ] Another question -- do we now have tukey box plot as the default box plot?  If not, we should change the default too. :)

mattwchun:
@kanitw No, currently the minmax boxplot is the default. I'll make the tukey box plot the default in another PR.

src/compositemark/boxplot.ts
```diff
@@ -285,30 +285,40 @@ function boxParams(spec: GenericUnitSpec<Encoding<string>, BOXPLOT | BoxPlotDef>
...
         calculate: 'datum.lowerBox - datum.IQR * ' + kIQRScalar,
-        as: 'lowerWhisker'
+        as: 'iqr_lowerWhisker'
       },
       {
         calculate: 'datum.upperBox + datum.IQR * ' + kIQRScalar,
+        as: 'iqr_upperWhisker'
+      },
+      {
+        calculate: 'min(datum.iqr_upperWhisker, datum.minmax_upperWhisker)',
```
kanitw: You can calculate `iqr_lowerWhisker` inline here:

`calculate: 'min(datum.upperBox + datum.IQR * ${kIQRScalar},  datum.minmax_upperWhisker)'`

examples/compiled/box-plot_minmax_2D_horizontal_IQR.vg.json
```diff
@@ -21,17 +21,23 @@
...
                         "people",
                         "people",
                         "people"
                     ],
                     "as": [
                         "lowerBox",
                         "upperBox",
-                        "midBox"
+                        "midBox",
+                        "minmax_lowerWhisker",
```
kanitw: Can this be named as min_fieldname

examples/compiled/box-plot_minmax_2D_horizontal_IQR.vg.json
```diff
@@ -21,17 +21,23 @@
...
                     "fields": [
+                        "people",
+                        "people",
                         "people",
                         "people",
                         "people"
                     ],
                     "as": [
                         "lowerBox",
                         "upperBox",
```
kanitw: Can we name it upper/lower/midBox_fieldName to avoid name collision?

examples/compiled/box-plot_minmax_2D_horizontal_IQR.vg.json
```diff
@@ -41,13 +47,13 @@
                 },
                 {
                     "type": "formula",
-                    "expr": "datum.lowerBox - datum.IQR * 1.5",
-                    "as": "lowerWhisker"
+                    "expr": "min(datum.upperBox + datum.IQR * 1.5, datum.minmax_upperWhisker)",
+                    "as": "upperWhisker"
```
kanitw: upperWhisker_fieldname

examples/compiled/box-plot_minmax_2D_horizontal_IQR.vg.json
mattwchun: @kanitw should IQR should have fieldname appended? So, it would be "IQR_people".

kanitw:
Thanks -- actually i now wonder if upper_box_fieldname is better than upperBox_fieldname.  (Sorry for being picky but it should be easy to change though.)

Maybe lets just make iqr lowercase too (iqr_fieldname).

After that feel free to fork another branch to make tukey box plot the default

mattwchun:
@kanitw the config properties (boxWhisker and boxMid) I presume you want changed to box_whisker and box_mid as well?

kanitw:
> @kanitw the config properties (boxWhisker and boxMid) I presume you want changed to box_whisker and box_mid as well? If I do it will follow a different Capitalization Styles than the other properties like axisTop.

Your hunch is right -- for config, camelCase is the way to go.  We shouldn't change that.

I only want to change the field naming style to be consistent.  (We do name field `mean_xxx` for normal aggregate so box should be consistent -- not to mention that boxWhisker_xxx mixes two style so it's obviously bad.)

kanitw:
Excellent.  Thank you for all the changes :)

akngs:
Tukey box plot's whiskers shouldn't be clamped to min/max bound. Please see #4568

Issue #3234, opened 2017-12-13
"null" is missing from invalidValues attribute in vega-lite 2 scheme


mattijn:
In the vega-lite 2 [scheme ](https://vega.github.io/schema/vega-lite/v2.json)for `invalidValues` it is only mentioned that `"filter"` is accepted while in the [documentation](https://vega.github.io/vega-lite/docs/config.html#top-level-config) `"null"` is also a valid option (which also works in the editor as documented)

kanitw:
Thanks for reporting.  This will be fixed in https://github.com/vega/vega-lite/pull/3240. :)

PR #3240, opened 2017-12-19
Fix schema for `config.invalidValues`

kanitw:
Fix https://github.com/vega/vega-lite/issues/3234

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1011,7 +1011,10 @@
           "enum": [
             "filter"
           ],
-          "type": "string"
+          "type": [
+            "string",
+            "null"
+          ]
         },
         "legend": {
           "$ref": "#/definitions/LegendConfig",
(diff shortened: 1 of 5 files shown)


build/vega-lite-schema.json
```diff
@@ -1011,7 +1011,10 @@
           "enum": [
             "filter"
           ],
-          "type": "string"
+          "type": [
+            "string",
+            "null"
```
kanitw: @domoritz  The schema generator actually still generate incorrect results here.

(The `enum` property should include `null` above -- otherwise, JSON schema validator would still throw warning.)

domoritz: Hmm, so you want it to be 

```
enum: "filter", null
type: string
```

?

kanitw: I think it should be 


```
enum: ["filter", null]
type: ["string", "null"]
```

domoritz: https://spacetelescope.github.io/understanding-json-schema/reference/null.html isn't clear about this but it seems that you are right: https://github.com/json-schema-org/json-schema-spec/issues/258. We do need to add the type and also add `null` to the enum. 

Fixed in https://github.com/vega/ts-json-schema-generator/issues/22

domoritz: Fixed. I added a commit with the new schema.

PR #3245, opened 2017-12-26
Improve `layer_rect_extent` example with better title

kanitw:


kanitw:
I wonder if our axis merge logic should be somewhat smarter about this?

domoritz:
How would we be smarter about this?

PR #3246, opened 2017-12-26
Update applications.md -- PoleStar is now part of Voyager 

kanitw:
This might be a better version of https://github.com/vega/vega-lite/pull/3236

cc: @nyurik

### DIFF
--- a/site/usage/applications.md
+++ b/site/usage/applications.md
@@ -20,8 +20,7 @@ This is an incomplete list of integrations, applications, and extensions of the 
 
 ## Tools
 
-* [Polestar](https://github.com/vega/polestar), lightweight Tableau-style interface for visual analysis
-* [Voyager](https://github.com/vega/voyager), visualization browser for open-ended data exploration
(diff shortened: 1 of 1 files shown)


nyurik:
+1.  I also think that we should group all Vega-3 compliant tools in one section, vs the legacy ones that don't support it (yet).  thx!

kanitw:
>  I also think that we should group all Vega-3 compliant tools in one section, vs the legacy ones that don't support it (yet). thx!

lol that's quite an amount of work and can get outdated very easily so I'd pass on this idea.

nyurik:
@kanitw how do you mean?  Vega3 is fairly different from previous versions, so either a tool supports it, or it doesn't. I think we should have a section for the "current" tools, and everything else. Otherwise it misleads people into thinking there are all these projects even though the projects might no longer be maintained. I'm sure the community will help you with this list ;)

kanitw:
Okie.  Feel free to submit another PR after we merge this PR.

PR #3254, opened 2018-01-06
Remove an error thrown for condition-only def

kanitw:
For example, this spec should not get error:

```
{
  "description": "Select multiple points with the shift key.",
  "selection": {
    "paintbrush": {
      "type": "multi", "on": "mouseover",
      "nearest": true
    }
  },
  "mark": "point",
  "encoding": {
    "size": {
      "condition": {"selection": "paintbrush", "value": 300}
    }
  }
}
```


domoritz:
Can you add a test?

Issue #3242, opened 2017-12-21
Make tukey box plot the default box plot


kanitw:
Forked from https://github.com/vega/vega-lite/issues/3166 -- https://github.com/vega/vega-lite/issues/3166#issuecomment-340120960

PR #3262, opened 2018-01-13
Default Boxplot to Tukey boxplot

mattwchun:
Refers to #3242.

Changes:
- Added to config property box a property called extent to make the default extent value to be 1.5
- Added test for boxplot mark type for tukey box plot

### DIFF
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -47,7 +47,10 @@ export const BOXPLOT_STYLES: BoxPlotStyle[] = ['boxWhisker', 'box', 'boxMid'];
 export interface BoxPlotConfig extends MarkConfig {
   /** Size of the box and mid tick of a box plot */
   size?: number;
-  /** Same as BoxPlotDef extent property */
+  /** The default extent, which is used to determine where the whiskers extend to. The options are
+   * - `"min-max": min and max are the lower and upper whiskers respectively.
+   * - `"number": A scalar (integer or floating point number) that will be multiplied by the IQR and the product will be added to the third quartile to get the upper whisker and subtracted from the first quartile to get the lower whisker.
+   */
   extent?: 'min-max' | number;
 }
 
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -1,4 +1,4 @@
-import {isNumber} from 'vega-util';
+import {isNumber} from 'util';
 import {Channel} from '../channel';
 import {Config} from '../config';
 import {reduce} from '../encoding';
@@ -32,8 +32,8 @@ export interface BoxPlotDef {
   /**
    * Extent is used to determine where the whiskers extend to. The options are
    * - `"min-max": min and max are the lower and upper whiskers respectively.
-   * - `"number": A scalar (integer or floating point number) that will be multiplied by the IQR and the product will be added to the third quartile to get the upper whisker and subtracted from the first quartile to get the lower whisker.
-   * __Default value:__ `"min-max"`.
+   * -  A scalar (integer or floating point number) that will be multiplied by the IQR and the product will be added to the third quartile to get the upper whisker and subtracted from the first quartile to get the lower whisker.
+   * __Default value:__ `"1.5"`.
    */
   extent?: 'min-max' | number;
 }
@@ -47,6 +47,8 @@ export const BOXPLOT_STYLES: BoxPlotStyle[] = ['boxWhisker', 'box', 'boxMid'];
 export interface BoxPlotConfig extends MarkConfig {
   /** Size of the box and mid tick of a box plot */
   size?: number;
+  /** Same as BoxPlotDef extent property */
+  extent?: 'min-max' | number;
 }
 
 export interface BoxPlotConfigMixins {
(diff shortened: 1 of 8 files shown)


examples/compiled/area.vg.json
```diff
@@ -177,6 +177,7 @@
         }
     ],
     "config": {
+        "boxExtent": 1.5,
```
kanitw: Hmm. `boxExtent` is not a valid Vega config.  In any case, it shouldn't be output in all output Vega specs.

src/compositemark/boxplot.ts
```diff
@@ -33,7 +32,7 @@ export interface BoxPlotDef {
    * Extent is used to determine where the whiskers extend to. The options are
    * - `"min-max": min and max are the lower and upper whiskers respectively.
    * - `"number": A scalar (integer or floating point number) that will be multiplied by the IQR and the product will be added to the third quartile to get the upper whisker and subtracted from the first quartile to get the lower whisker.
-   * __Default value:__ `"min-max"`.
+   * __Default value:__ `"number"`.
```
kanitw: 1.5

src/compositemark/boxplot.ts
```diff
@@ -95,11 +96,11 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<string>, BOXPLOT
   // TODO: use selection
   const {mark, encoding, selection, ...outerSpec} = spec;
 
-  let kIQRScalar: number = undefined;
+  let kIQRScalar: number = (config.box && config.box.extent) ? config.box.extent : 1.5;
```
kanitw: if you already put 1.5 in the default config, you won't need redundancy here.  

The more important part is to make sure extent is a number.

src/compositemark/boxplot.ts
```diff
@@ -33,7 +33,7 @@ export interface BoxPlotDef {
    * Extent is used to determine where the whiskers extend to. The options are
    * - `"min-max": min and max are the lower and upper whiskers respectively.
    * - `"number": A scalar (integer or floating point number) that will be multiplied by the IQR and the product will be added to the third quartile to get the upper whisker and subtracted from the first quartile to get the lower whisker.
```
kanitw: `"number"` is not a valid value here  --  better just say
```
-  A scalar value (integer or floating point number) that ...
```

src/compositemark/boxplot.ts
```diff
@@ -47,6 +47,8 @@ export const BOXPLOT_STYLES: BoxPlotStyle[] = ['boxWhisker', 'box', 'boxMid'];
 export interface BoxPlotConfig extends MarkConfig {
   /** Size of the box and mid tick of a box plot */
   size?: number;
+  /** A scalar that will be multiplied by the IQR */
```
kanitw: Shouldn't this have the same description as `extent` in `BoxPlotDef`?

mattwchun: No, I use this config extent property just when extent isn't specified in the mark def

kanitw: Yes, but say if you put "min-max" here it should work too.   So at least it should say about min-max here.

src/compositemark/boxplot.ts
```diff
@@ -47,6 +47,8 @@ export const BOXPLOT_STYLES: BoxPlotStyle[] = ['boxWhisker', 'box', 'boxMid'];
 export interface BoxPlotConfig extends MarkConfig {
   /** Size of the box and mid tick of a box plot */
   size?: number;
+  /** Same as BoxPlotDef extent property */
```
kanitw: This is not an acceptable explanation that will show up in the documentation.  
```
    *  The default extent, which is used to determine where the whiskers extend to. The options are
    * - `"min-max": min and max are the lower and upper whiskers respectively.	    * - `"min-max": min and max are the lower and upper whiskers respectively.
-   * - `"number": A scalar (integer or floating point number) that will be multiplied by the IQR and the product will be added to the third quartile to get the upper whisker and subtracted from the first quartile to get the lower whisker.	+   * -  A scalar (integer or floating point number) that will be multiplied by the IQR and the product will be added to the third quartile to get the upper whisker and subtracted from the first quartile to get the lower whisker.
```


Also, if you push fixes, please do not squash and force push as I won't notice.

mattwchun: So do you want what you posted instead of the following?
/** Same as BoxPlotDef extent property */

kanitw: Yes.   There is no "BoxPlotDef" appear in the documentation, so this wouldn't be understandable.

mattwchun: I assume:
```"min-max": min and max are the lower and upper whiskers respectively.	```
showing up twice in what you typed above is a typo

Comment (kanitw) [reply src/compositemark/boxplot.ts]: Yes, that's a typo. 

Not sure what to update on line 35.  I copy this from line 35 for you. 

The only exception is that I say "the default extent" here as this is a config. 
The term the default extent wouldn't apply in line 35.

File: src/compositemark/boxplot.ts [hunk 7]
```diff
@@ -1,4 +1,4 @@
-import {isNumber} from 'vega-util';
+import {isNumber} from 'util';
```
ydlamba: `util.isNumber` is deprecated. If `vega-util` has `isNumber` function, Is there is a reason why you are using it from `util`?

mattwchun: Good catch! #3446

PR #3265, opened 2018-01-14
Fix rule mark to support arbitrary (possibly oblique) line segments

### LINKED ISSUE #3263: Oblique rule mark
Currently a bounded `rule` mark can be either horizontal (if `x` and `x2` are defined) or vertical (if `y` and `y2` are defined).

Would it be possible to allow both x and y bounds to be defined for the same rule so rules can be oblique? (And as an aside, the current wording in the documentation ["a rule can be used to draw a line segment between two positions"](https://vega.github.io/vega-lite/docs/rule.html) does not make it clear that those positions have to share either a horizontal or vertical coordinate).

This would allow parallel coordinates plots (PCPs) to be easily built such as in the following:

    {
      "width":100,
      "data": { "url": "data/cars.json"},
      "mark": "rule",
      "encoding": {
        "x": { "value": 0},
        "y": {"field": "Horsepower","type": "quantitative"},
        "x2": { "value": 100},
        "y2": {"field": "Weight_in_lbs","type": "quantitative"}
      }
    }

While PCPs can be partially simulated using the `line` mark and `detail`, this only works if all variables in the PCP share a common scale. This means that PCPs that mix measurement types and ranges have to be rescaled with a data transform rather than relying on Vega-Lite's built-in scaling of any measurement type.

kanitw:
Fix #3263

image

I don't think this imperfect parallel coordinate is a good example for this yet -- if you have a good idea about a good example, please feel free to submit a separate PR.

### DIFF
--- a/src/compile/mark/init.ts
+++ b/src/compile/mark/init.ts
@@ -65,6 +65,12 @@ function orient(mark: Mark, encoding: Encoding<string>, specifiedOrient: Orient)
 
   switch (mark) {
     case RULE:
+      // return undefined for line segment rule
+      if (xIsRange && yIsRange) {
+        return undefined;
+      }
+      /* tslint:disable */
+      // intentional fall through
     case BAR:
     case AREA:
       // If there are range for both x and y, y (vertical) has higher precedence.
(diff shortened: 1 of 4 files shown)


domoritz:
@jwoLondon, can you send me the spec for https://github.com/vega/vega-lite/issues/3263#issuecomment-357555321 so I can try it on this branch?

kanitw:
The spec on the top works though, so the spec shouldn't block this merge.

domoritz:
I just want to try https://user-images.githubusercontent.com/1846999/34922544-5c3c768e-f989-11e7-8f4e-45e3cd0ae799.png.

domoritz:
Okay, let's add it later as an example.

PR #3239, opened 2017-12-19
Make encoding support non-selection predicate 

kanitw:
Please merge https://github.com/vega/vega-lite/pull/3275 first

Fix https://github.com/vega/vega-lite/issues/2529

Issue #2901, opened 2017-08-19
Improve Docs for Logical operand


kanitw:
Right now, we don't really document logical operand well.

kanitw:
- [ ] Logical operand (for `selection` predicate and for `filter`)

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/5741

PR #3275, opened 2018-01-17
Refactor: Distinguish between FieldPredicate and SelectionFilter

kanitw:
(FieldPredicate can be useful in other scenario besides filter -- e.g., for condition in https://github.com/vega/vega-lite/pull/3239 or for selecting data for annotation (@starry97's project)

cc: @starry97

### DIFF
--- a/_data/link.yml
+++ b/_data/link.yml
@@ -42,9 +42,8 @@ Bin:
   link: bin.html
 BinParams:
   link: "bin.html#bin-parameters"
-FilterOperand:
-  name: Filter
-  link: "transform.html#filter"
+"LogicalOperand<Predicate>":
+  link: "transform.html#predicate"
 AggregateOpDef:
   link: "transform.html#aggregate-op-def"
 TimeUnit:
--- a/site/docs/transform/filter.md
+++ b/site/docs/transform/filter.md
@@ -19,7 +19,7 @@ The filter transform removes objects from a data stream based on a provided filt
 }
 ```
 
-Vega-Lite filter transforms must have the `filter` property.
+Vega-Lite filter transforms must have the `filter` property describing the predicate for the filtering condition.
 
 {% include table.html props="filter" source="FilterTransform" %}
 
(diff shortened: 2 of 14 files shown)


site/docs/transform/filter.md
```diff
@@ -68,7 +67,9 @@ For example, to check if the `car_color` field's value is equal to `"red"`, we c
 For example, `{"filter": {"field": "car_color", "oneOf":["red", "yellow"]}}` checks if the `car_color` field's value is `"red"` or `"yellow"`.
 
 {:#selectionfilter}
-### Selection Filter
+## Selection Filter
+
+For a selection filter object, a [`selection` name](#selectionfilter)  must be provided.
```
domoritz: This links to itself.

site/docs/transform/filter.md
```diff
@@ -30,13 +30,12 @@ Vega-Lite filter transforms must have the `filter` property.
 For a [Vega Expression](https://vega.github.io/vega/docs/expressions/) string, each datum object can be referred using bound variable `datum`. For example, setting `filter` to `"datum.b2 > 60"` would make the output data includes only items that have values in the field `b2` over 60.
 
 
-## Filter Object
+## Field Predicate Object
```
domoritz: This is a bit inconsistent. Below, we call the filter "Selection Filter". I'd prefer to keep the name (it is also less confusing) and say that the "Filter Object" has to contain a "Filter Predicate". Alternatively, we could name it "Field Filter Object". The predicate is the expression that is used to filter.

kanitw: I will change selection filter to selection predicate.  We should not call them filters because they are also use for condition syntax in a follow-up PR.  Field predicate is also more applicable in more places so we need to distinguish them.  The old text is confusing anyway.

site/docs/transform/filter.md
kanitw: Nice catch!

src/filter.ts
```diff
@@ -9,25 +9,30 @@ import {isArray, isString, logicalExpr} from './util';
 
 
 export type Filter =
```
kanitw: Actually Let me rename this whole thing to predicate

src/transform.ts
```diff
@@ -1,19 +1,23 @@
...
+   * The `filter` property must be one of the predicate definitions:
+   * (1) a [Vega Expression](filter.html#expression) string,
+   * where `datum` can be used to refer to the current data object;
+   * (2) one of the field predicates: [equal predicate](filter.html#equal-predicate);
+   * [range precidate](filter.html#range-predicate), [one-of predicate](filter.html#one-of-predicate);
+   * (3) a [selection predicate](filter.html#selection-predicate);
+   * or (4) a logical operand that combines (1), (2), or (3).
    */
-  filter: LogicalOperand<Filter>;
+  // TODO: https://github.com/vega/vega-lite/issues/2901
```
domoritz: How much work is https://github.com/vega/vega-lite/issues/2901?

kanitw: An hour or two -- in any case, it is NOT the scope of this PR.  I'm just noting down.

_data/link.yml
```diff
@@ -42,9 +42,8 @@ Bin:
   link: bin.html
 BinParams:
   link: "bin.html#bin-parameters"
-FilterOperand:
-  name: Filter
-  link: "transform.html#filter"
+"LogicalOperand<Predicate>":
+  link: "transform.html#predicate"
```
domoritz: do you want to give this a better name?

kanitw: This is good since there is LogicalOperand<String> too


So having generic should allow us to write about it only once

domoritz: I see. Go ahead and merge then.

PR #3279, opened 2018-01-19
Make sort descending work correctly with header

### LINKED ISSUE #3278: Facet headers do not sort
When adding `sort` to a facet like `row` or `column`, the charts are sorted but the headers are not:
```
{
  "mark": "bar",
  "encoding": {
    "x": {
      "bin": {"maxbins": 15},
      "field": "Horsepower",
      "type": "quantitative"
    },
    "y": {
      "aggregate": "count",
      "type": "quantitative"
    },
    "row": {"field": "Origin", "type": "nominal", "sort": "ascending"}
  }
}
```
Change `sort` to `descending`:
image

kanitw:
Fix #3278

### DIFF
--- a/src/compile/layout/header.ts
+++ b/src/compile/layout/header.ts
@@ -4,7 +4,7 @@
 import {FacetFieldDef} from '../../facet';
 import {vgField} from '../../fielddef';
 import {keys} from '../../util';
-import {AxisOrient, VgAxis} from '../../vega.schema';
+import {AxisOrient, VgAxis, VgMarkGroup} from '../../vega.schema';
 import {formatSignalRef} from '../common';
 import {Model} from '../model';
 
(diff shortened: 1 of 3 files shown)


src/compile/layout/header.ts
```diff
@@ -84,11 +84,25 @@ export function getTitleGroup(model: Model, channel: HeaderChannel) {
   };
 }
 
-export function getHeaderGroup(model: Model, channel: HeaderChannel, headerType: HeaderType, layoutHeader: LayoutHeaderComponent, headerCmpt: HeaderComponent) {
+export function getHeaderGroups(model: Model, channel: HeaderChannel): VgMarkGroup[] {
+  const {layoutHeaders} = model.component;
+  const layoutHeader = layoutHeaders[channel];
```
domoritz: const layoutHeader = model.component.layoutHeaders[channel];

kanitw: Done

Issue #3281, opened 2018-01-20
Aggregate encoding and aggregate transform give different results


amitkaps:
Here is a small test case based on `stocks.csv` dataset

1. Using aggregate encoding 
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "width": 600,
  "height": 300,
  "data": {"url": "data/stocks.csv"},
  "mark": "point",
  "encoding": {
    "x": {"field": "date", "type": "temporal", "axis": {"format": "%Y"}},
    "y": {"field": "price", "type": "quantitative", "aggregate": "max"}
  }
}
```
![agg_encoding](https://user-images.githubusercontent.com/878499/35179805-92256016-fdc7-11e7-8d8b-a3654a9f3c55.png)

2. Using aggregate transform
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "width": 600,
  "height": 300,
  "data": {"url": "data/stocks.csv"},
  "transform": [{
    "aggregate": [{"op": "max", "field": "price", "as": "max_price"}],
    "groupby": ["date"]
  }],
  "mark": "point",
  "encoding": {
    "x": {"field": "date", "type": "temporal", "axis": {"format": "%Y"}},
    "y": {"field": "max_price", "type": "quantitative"}
  }
}
```
![agg_transform](https://user-images.githubusercontent.com/878499/35179806-96b5c332-fdc7-11e7-9989-575e7eb40ab8.png)

On further investigation, it is appears to be a result of different handling of null values in the initial data or as a result of the aggregation. The compiled specs for both are different. Here is the output for `vgSpec.data` using the following code. 

```js
var vgSpec = vl.compile(vlSpec).spec;
console.log(JSON.stringify(vgSpec.data, 0, 2))
```
1. Using aggregate encoding 
```json
[
  {
    "name": "source_0",
    "url": "data/stocks.csv",
    "format": {
      "type": "csv",
      "parse": {
        "date": "date",
        "price": "number"
      }
    },
    "transform": [
      {
        "type": "aggregate",
        "groupby": [
          "date"
        ],
        "ops": [
          "max"
        ],
        "fields": [
          "price"
        ],
        "as": [
          "max_price"
        ]
      },
      {
        "type": "filter",
        "expr": "datum[\"date\"] !== null && !isNaN(datum[\"date\"])"
      }
    ]
  }
]
```
2. Using aggregate transform
```json
[
  {
    "name": "source_0",
[...]

PR #3285, opened 2018-01-21
Added example in Aggregate Transform

amitkaps:
- Added explicit example on how the aggregate transform works
- Added note on need to parse dataset for explicit type inference 

Reference: https://github.com/vega/vega-lite/issues/3281


examples/specs/bar_aggregate_transform.vl.json
```diff
@@ -0,0 +1,19 @@
+{
```
kanitw: Please run `yarn run build:example bar_aggregate_transform` so the compiled files are included.

site/docs/transform/aggregate.md
```diff
@@ -14,7 +14,7 @@ To aggregate data in Vega-Lite, users can either use use an `aggregate` property
 
 
 {:#encoding}
-## Aggregate (Encoding)
+## Aggregate Encoding
```
kanitw: Let's call this "Aggregate in Encoding Field Definition"

(Please feel free to do the same correction for bin and timeUnit)

site/docs/transform/aggregate.md
```diff
@@ -111,3 +114,5 @@ The supported **aggregation operations** are:
 | max       | The maximum field value.|
 | argmin    | An input data object containing the minimum field value.|
 | argmax    | An input data object containing the maximum field value.|
+
+Note: It is important you [`parse`](data.html#format) your data types explicitly, especially if you are likely to have `null` values in your dataset and automatic type inference will fail.
```
kanitw: I think this note should be listed in the subsection for aggregate transform.  (Since this subsection is for supported aggregation operations, which are applicable for aggregate in either encoding or transform. )

examples/specs/bar_aggregate_transform.vl.json
amitkaps: Done.

site/docs/transform/aggregate.md
amitkaps: Done for all.

site/docs/transform/aggregate.md
amitkaps: Agreed. Changed.

PR #3287, opened 2018-01-21
Domain fix

### LINKED ISSUE #3284: Invalid data source name in output scale domain for faceted layer
Here is a minimal spec to reproduce the problem. 

```
{
  "facet": {
    "row": {"field": "symbol", "type": "nominal"}
  },
  "spec": {
    "layer": [{
      "data": {"values": [{"ref": 100}]},
      "mark": "rule",
      "encoding": {
        "y": { "field":"ref", "type": "quantitative"},
        "size": {"value": 2}
      }
    }] 
  }
}
```

This produces the following error: 

```
Error: Undefined data set name: "scale_child_layer_0_main"
```

Originally reported by @eibanez in https://github.com/vega/vega-lite/issues/1601#issuecomment-358978898

domoritz:
Fixes #3284

### DIFF
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -8,7 +8,7 @@ import {ChannelDef, FieldDef, FieldRefOption, getFieldDef, vgField} from '../fie
 import * as log from '../log';
 import {Resolve} from '../resolve';
 import {hasDiscreteDomain} from '../scale';
-import {BaseSpec} from '../spec';
+import {BaseSpec, isFacetSpec} from '../spec';
 import {extractTitleConfig, TitleParams} from '../title';
 import {normalizeTransform, Transform} from '../transform';
 import {contains, Dict, keys, varName} from '../util';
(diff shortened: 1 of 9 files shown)


src/compile/model.ts
```diff
@@ -186,7 +186,9 @@ export abstract class Model {
         sources: parent ? parent.component.data.sources : {},
         outputNodes: parent ? parent.component.data.outputNodes : {},
         outputNodeRefCounts: parent ? parent.component.data.outputNodeRefCounts : {},
-        ancestorParse: parent ? {...parent.component.data.ancestorParse} : {}
+        ancestorParse: parent ? {...parent.component.data.ancestorParse} : {},
+        // data is faceted if the spec is a facet spec or the parent has faceted data and no data is defined
+        isFaceted: isFacetSpec(spec) || (parent && parent.component.data.isFaceted && !spec.data)
```
kanitw: If I understand correctly, this will also be correct for facet of layer right?  (Just asking to make sure)

kanitw:
Test failing

src/compile/model.ts
domoritz: Yeah, it works with my example.

domoritz:
Revised and added some more tests.

Issue #3289, opened 2018-01-21
Tie breaker fails with objects


domoritz:
Load `facet_independent_scale_layer.vl.json`. 

```
vega-util.js:132 WARN Conflicting scale property "range" ("{"step":17}" and "{"step":17}").  Using "{"step":17}".
```

kanitw:
Will be fixed in #3291

PR #3291, opened 2018-01-22
Make `mergeValuesWithExplicit` work with object

kanitw:
Fix https://github.com/vega/vega-lite/issues/3289


src/compile/split.ts
```diff
@@ -122,7 +123,7 @@ export function mergeValuesWithExplicit<S, T>(
     return v1;
   } else if (v2.explicit && !v1.explicit) {
     return v2;
-  } else if (v1.value === v2.value) {
+  } else if (stringify(v1.value) === stringify(v2.value)) {
```
domoritz: Please use `utils.hash`

domoritz:
Done.

Issue #3022, opened 2017-09-19
Support value: "end" for equivalent Vega "group": "width"/"height"


kanitw:
This way users can use maximum/minimum value of the view frame.

kanitw:
@starry97 -- can you remember the use case that we need this for?

PR #3314, opened 2018-01-28
Add a new example with specific data domain values and labels

starry97:
Add example

image


examples/specs/layer_bar_annotations.vl.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "A simple bar chart with annotation.",
+  "data": {
+    "values": [
+      {"Day": "D01", "Value": "50"},
```
kanitw: Can we just have day as number from 1,2,3,...?

examples/specs/layer_bar_annotations.vl.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "A simple bar chart with annotation.",
```
kanitw: I think this is more specific than just  "a simple bar chart with annotation".  

Do you have source of the original example?  It would be worth linking to giving the original example credits.

examples/specs/layer_bar_annotations.vl.json
```diff
@@ -0,0 +1,60 @@
...
+      {"Day": "D07", "Value": "140"},
+      {"Day": "D08", "Value": "130"},
+      {"Day": "D09", "Value": "390"},
+      {"Day": "D10", "Value": "200"},
+      {"Day": "D11", "Value": "70"},
+      {"Day": "D12", "Value": "50"},
+      {"Day": "D13", "Value": "100"},
+      {"Day": "D14", "Value": "150"},
+      {"Day": "D15", "Value": "130"},
+      {"Threshold": "hazardous", "ThresholdValue": "300"}
```
kanitw: I don't think you should make the data table heterogenous.  In a way, you should think about this as one data source for the bar and perhaps another with only one data point for the threshold.
In a way, we might not really need the second data source, as it's more like constant value.  

In any case, please remove this line from the main data source?

examples/specs/layer_bar_annotations.vl.json
```diff
@@ -0,0 +1,60 @@
...
+    }, {
+      "mark": "bar",
+      "encoding": {
+        "x": {"field": "Day", "type": "ordinal"},
+        "y": {"field": "Value", "type": "quantitative"}
+      }
+    }, {
+      "mark": {
+        "type": "text",
+        "align": "left",
```
kanitw: It should be align right and dx = plot width?

It's probably worth implementing https://github.com/vega/vega-lite/issues/3022 too

examples/specs/layer_bar_annotations.vl.json
```diff
@@ -0,0 +1,60 @@
...
+      {"Day": "D11", "Value": "70"},
+      {"Day": "D12", "Value": "50"},
+      {"Day": "D13", "Value": "100"},
+      {"Day": "D14", "Value": "150"},
+      {"Day": "D15", "Value": "130"},
+      {"Threshold": "hazardous", "ThresholdValue": "300"}
+    ]
+  },
+  "transform": [{
+    "calculate": "datum.Value > 300 ? 300 : datum.Value", "as": "diff"
```
kanitw: Why is this called "diff"?  The formula doesn't read like diff to me.  
Instead of calculating a new field like this, I think your red bar layer can just "filter" if value is more than 300 and encode data from y=300 to y=field: value.

examples/specs/layer_bar_annotations.vl.json
kanitw: (We can just implement #3022 later, and update this example to be nicer.)

starry97: How can I get the plot width?

kanitw: Just stick the number for now (in this PR).

But I think you should also implement #3022 so you get a cleaner syntax for width.

examples/specs/layer_bar_annotations.vl.json
starry97: I used ChartAccent tool to create the original example. Should I mention that where the dataset is from?

examples/specs/layer_bar_annotations.vl.json
```diff
@@ -0,0 +1,77 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "The PM2.5 value of Beijing observed 15 days, highlighting the days when PM2.5 level is hazardous to human health",
```
kanitw: Please credit data source :)

examples/specs/layer_bar_annotations.vl.json
```diff
@@ -0,0 +1,77 @@
...
+      "mark": "bar",
+      "encoding": {
+        "x": {"field": "Day", "type": "ordinal"},
+        "y": {"field": "Value", "type": "quantitative"}
+      }
+    },
+    {
+      "mark": "bar",
+      "transform": [{
+        "calculate": "datum.Value >= 300 ? 300 : datum.Value",
```
kanitw: Technically this group of marks still generates N bars under the hood, which is semantically wrong.  We only want to draw one bar here. 

Please use filter and calculate baseline = 300 without condition.

examples/specs/layer_bar_annotations.vl.json
```diff
@@ -0,0 +1,77 @@
...
+    {
+      "mark": {
+        "type": "text",
+        "align": "left",
+        "dx": 215,
+        "dy": -5
+      },
+       "data": {
+         "values": [
+          {"ThresholdValue": 300, "Threshold": "hazardous"}
```
kanitw: this data source can be shared between rule and text

kanitw:
One last request -- rotate x axis label?

starry97:
It seems like I have to overwrite all x-axis label to overwrite its default angle. I'm not sure if this is intentional.

kanitw:
> It seems like I have to overwrite all x-axis label to overwrite its default angle. I'm not sure if this is intentional.

Sounds like you found a bug.  Please file an issue with detailed description. ;)

_data/examples.json
```diff
@@ -159,6 +159,10 @@
       "title": "Simple Bar Chart with Labels",
       "style": "background-size: auto 70%; background-position-y: center;"
     },
+    {
+      "name": "layer_bar_annotations",
+      "title": "Bar chart with highlighting annotation overlay"
```
kanitw: Be a bit more specific: e.g., "Bar chart that highlights a value beyond a threshold"?

Also add a `description` field trying to describe the example -- this will appear in your example page like in https://vega.github.io/vega-lite/examples/bar.html. 

(You can use markdown for the `description` field.)

_data/examples.json
```diff
@@ -161,7 +161,8 @@
     },
     {
       "name": "layer_bar_annotations",
-      "title": "Bar chart with highlighting annotation overlay"
+      "title": "Bar chart that highlights values beyond a threshold",
+      "description": "A bar chart encodes quantitative values, that marks a user specified data domain value, and highlights the area above this value."
```
kanitw: Description could be multiple sentences.  (Think about it as if you have to teach someone briefly about tricks you use in this visualization, how would you explain it to other people.)

Something like: 

To create a bar chart that highlights values beyond a threshold, we use two `layer`s of `bar` marks. The lower layer shows all the bars while the upper layer shows data with values above the threshold in red (#...colorno..).  We then `layer` a `rule` and a `text` mark to annotate the threshold on top of the bars.

kanitw:
Please feel free to check / revise my edits and merge squash.

PR #3327, opened 2018-01-30
Fix type of RangeConfig for strict null checks

pelotom:
With `strictNullChecks` enabled in TS 2.6, this definition wasn't type checking; each other field under `RangeConfig` was being flagged with a variant of this error:

```
Property 'category' of type 'VgScheme | string[] | undefined' is not assignable to string index type 'RangeConfigValue'.
```

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -102,7 +102,7 @@ export const defaultViewConfig: ViewConfig = {
 
 export type RangeConfigValue = (number|string)[] | VgScheme | {step: number};
 
-export interface RangeConfig {
+export interface RangeConfig extends Partial<Record<string, RangeConfigValue>> {
   /**
    * Default range for _nominal_ (categorical) fields.
    */
@@ -132,8 +132,6 @@ export interface RangeConfig {
    * Default range palette for the `shape` channel.
    */
   symbol?: string[];
-
-  [name: string]: RangeConfigValue | undefined;
 }
 
 export interface VLOnlyConfig {
--- a/src/config.ts
+++ b/src/config.ts
@@ -133,7 +133,7 @@ export interface RangeConfig {
    */
   symbol?: string[];
 
-  [name: string]: RangeConfigValue;
+  [name: string]: RangeConfigValue | undefined;
 }
 
 export interface VLOnlyConfig {


src/config.ts
```diff
@@ -133,7 +133,7 @@ export interface RangeConfig {
    */
   symbol?: string[];
 
-  [name: string]: RangeConfigValue;
+  [name: string]: RangeConfigValue | undefined;
```
kanitw: @domoritz  -- I wonder if the right thing to do is adding `?` to `name`?

domoritz: Yes, add ? instead.

pelotom: That's not valid TypeScript syntax. What about this?

```ts
export interface RangeConfig extends Partial<Record<string, RangeConfigValue>> {
    /**
     * Default range for _nominal_ (categorical) fields.
     */
    category?: string[] | VgScheme;
    /**
     * Default range for diverging _quantitative_ fields.
     */
    diverging?: string[] | VgScheme;
    /**
     * Default range for _quantitative_ heatmaps.
     */
    heatmap?: string[] | VgScheme;
    /**
     * Default range for _ordinal_ fields.
     */
    ordinal?: string[] | VgScheme;
    /**
     * Default range for _quantitative_ and _temporal_ fields.
     */
    ramp?: string[] | VgScheme;
    /**
     * Default range palette for the `shape` channel.
     */
    symbol?: string[];
}
```

domoritz: `[name: string]?: RangeConfigValue` is not?

pelotom: No.

domoritz: Ahh, I misremembered and confused it with this:

```ts
type Bar = 'a' | 'b';

export type Foo = {
  [C in Bar]?: boolean
};
```

How come that one works but not the other?

Anyway, looking at this again, it doesn't really matter. `[name: string]: RangeConfigValue` already means that the object can have no props defined.

domoritz: Here are some experiments: https://www.typescriptlang.org/play/#src=type%20Foo1%20%3D%20%7B%0A%20%20%5Bx%3A%20string%5D%3F%3A%20boolean%0A%7D%0A%0Atype%20Bar%20%3D%20'a'%20%7C%20'b'%3B%0A%0Atype%20Foo2%20%3D%20%7B%0A%20%20%5BC%20in%20Bar%5D%3F%3A%20boolean%0A%7D%3B%0A

pelotom:
>Ahh, I misremembered and confused it with this:
>
>```ts
>type Bar = 'a' | 'b';
> 
>export type Foo = {
>  [C in Bar]?: boolean
>};
>```
>
>How come that one works but not the other?

I'm not sure, that's just the language 🤷‍♂️ But you can't use the mapped types syntax as part of an interface:

```ts
// Invalid:
interface Foo {
  [C in Bar]?: boolean
}
```

> Anyway, looking at this again, it doesn't really matter. `[name: string]: RangeConfigValue` already means that the object can have no props defined.

I'm not sure what you mean, but this library cannot currently be depended on by a project which is compiling with `strictNullChecks` because the other props in the interface are not compabitle with  `[name: string]: RangeConfigValue`.

domoritz:
> I'm not sure what you mean, but this library cannot currently be depended on by a project which is compiling with strictNullChecks because the other props in the interface are not compabitle with [name: string]: RangeConfigValue.

That's a different issue. We could address this in a cleaner way by using an intersection type.

pelotom:
The intersection form is equivalent, but I’m happy to use that if you find it preferable.

domoritz:
I tried the alternative approach in https://github.com/vega/vega-lite/pull/3329. @kanitw will decide what is better. 

Thank you for the contributions @pelotom by the way.

pelotom:
Thanks. Since your version is passing CI and mine isn't, I'd be inclined to close this in favor of it.

PR #3332, opened 2018-01-31
Clarify top-level vs view-level in transform.md

LukasGrebe:
The View spec.mp differentiates between "Common Properties of Specifications" and "Top-Level Specifications".
The documentation for transform.md was updated to follow this differentiation.


domoritz:
Thank you!

kanitw:
Thank you so much :)

PR #3334, opened 2018-01-31
Link to observable and observable notebook.

domoritz:


site/usage/applications.md
```diff
@@ -26,6 +26,7 @@ This is an incomplete list of integrations, applications, and extensions of the
 -   [Lyra](https://github.com/vega/lyra), an interactive, graphical Visualization Design Environment (VDE)
 -   [JupyterLab](https://github.com/jupyterlab/jupyterlab), an extensible environment for interactive and reproducible computing, based on the Jupyter Notebook and Architecture.
 -   [nteract](https://github.com/nteract/nteract), interactive notebook application with Vega and Vega-Lite renderer.
+-   [Observable](https://beta.observablehq.com/), an interactive JavaScript notebook.
```
kanitw: (Maybe later) but I think we should distinguish between 
- platforms that you can use Vega-Lite on directly (Jupyter, nteract, Observable, data.world) 
- Tools for authoring Vega-Lite (Voyager 2)
- Tools that leverage Vega-Lite in their implementation (Lyra, Turi Create, Django Chartflo, Iris)

domoritz: Hmm, we only have one authoring tool. What's the difference between the last two in your list?

domoritz: Let's do that in a follow up PR.

Issue #3340, opened 2018-02-01
Encoding is not always required


domoritz:
See the first layer in https://github.com/vega/vega-lite/issues/3335. However, our schema requires an encoding in any case.

kanitw:
Fixed in one of the pending PR

PR #3342, opened 2018-02-01
Improve geo examples. Fixes #3309

### LINKED ISSUE #3309: Increase contrast in airport example

domoritz:
Including one suggestion by @jwoLondon.


domoritz:
Fails because of #3340

PR #3345, opened 2018-02-01
Map with point and text

domoritz:
Depends on #3515


kanitw:
Now doable with https://github.com/vega/vega-lite/pull/3515?

domoritz:
Yes

image

examples/specs/geo_text.vl.json
```diff
@@ -36,9 +32,22 @@
         "y": {
           "field": "lat",
           "type": "latitude"
+        }
+      },
+      "layer": [{
+        "mark": "circle",
```
kanitw: Why not using color in mark?  (more concise)

domoritz: Will do when I update the spec for you lat/long PR.

PR #3356, opened 2018-02-05
Clean title output

domoritz:


domoritz:
Ready for review @kanitw

src/compile/axis/parse.ts
```diff
@@ -317,7 +317,8 @@ function getProperty<K extends keyof VgAxis>(property: K, specifiedAxis: Axis, c
     case 'ticks':
       return properties.ticks('ticks', specifiedAxis, isGridAxis, channel);
     case 'title':
-      return getSpecifiedOrDefaultValue(specifiedAxis.title, properties.title(specifiedAxis.titleMaxLength, fieldDef, model.config));
+      const title = getSpecifiedOrDefaultValue(specifiedAxis.title, properties.title(specifiedAxis.titleMaxLength, fieldDef, model.config));
+      return title === null || title === '' ? undefined : title;
```
kanitw: what if title is `undefined`? 

Why not just 
```
return getSpecifiedOrDefaultValue(specifiedAxis.title, properties.title(specifiedAxis.titleMaxLength, fieldDef, model.config)) || undefined;
```

src/compile/legend/parse.ts
```diff
@@ -96,7 +96,8 @@ function getProperty(property: keyof (Legend | VgLegend), specifiedLegend: Legen
       // We don't include temporal field here as we apply format in encode block
       return numberFormat(fieldDef, specifiedLegend.format, model.config);
     case 'title':
-      return getSpecifiedOrDefaultValue(specifiedLegend.title, fieldDefTitle(fieldDef, model.config));
+      const title = getSpecifiedOrDefaultValue(specifiedLegend.title, fieldDefTitle(fieldDef, model.config));
+      return title === null || title === '' ? undefined : title;
```
kanitw: Why not just 
```
return getSpecifiedOrDefaultValue(specifiedAxis.title, properties.title(specifiedAxis.titleMaxLength, fieldDef, model.config)) || undefined;
```

src/compile/axis/parse.ts
domoritz: Then we want to use the default, right?

kanitw: I see, but then title is never undefined here I guess. 

Still, I think just do `|| undefined` makes it less confusing.

domoritz: I agree

PR #3366, opened 2018-02-07
Add new example in Bl.ock to Community Examples

melissatdiamond:
New example based on vega example: https://vega.github.io/vega-editor/?mode=vega&spec=weather

### DIFF
--- a/site/examples/index.md
+++ b/site/examples/index.md
@@ -54,3 +54,4 @@ Here we list great examples of Vega-Lite visualizations that were created by the
 * [Interactive Exploration of Seattle Weather](https://bl.ocks.org/jakevdp/5d1915d808d3d91ce86f0bc3ca066d48) by @jakevdp
 * [Waterfall Chart](https://bl.ocks.org/italo-batista/5f93b3c9f87dffcf01bb489d90f60652/ef01271c72c57dc940bb357019be1c6cdf9eb51a) by @italo-batista
 * [Bar, Small Multiple, Heatmap, Gantt Charts: Exploring NYC Event Permits](https://medium.com/enigma-engineering/exploring-new-york-city-event-permits-with-vega-lite-f83178ff9a8d) by [@hydrosquall](https://twitter.com/hydrosquall)
+* [Layered Floating Bar Chart](https://bl.ocks.org/melissatdiamond/014a58aead0b70c72f851938fefb6be4) by @melissatdiamond


domoritz:
Looks great! Can you add a legend to explain what the different colors mean? Also, could you change the axis labels to be the weekday, rather than a number.

melissatdiamond:
I changed the axis labels to be the weekdays, which required some tweaks to the data so made it inline. However, I'm not sure about how to make a legend. The colors only distinguish the two layers, one that shows the range of record temps and one that shows the range of normal temps. The colors are not scaled by the temperatures themselves. Is there a way to make a legend that reflects this?

melissatdiamond:
In the data there are multiple objects with the same value for day (since there is more than a week of data and some days, like Saturday and Sunday, both use the same letter). If I use day for the x axis, then the data overlaps and there is only an entry for each unique day value. I could not find a way to base the axis on one field but have the tick labels be another field in the object. I found this post https://stackoverflow.com/questions/45814950/vega-lite-use-one-data-field-for-the-axis-and-another-for-the-label that seems to deal with the same topic, however this example was regarding dates so the solution doesn't seem to apply to the weather data. Is there another way? 

Besides the record and normal temps that are currently included, the first half of the objects in the data have an "actual" temp while the second half have a "forecast" temp instead. I did not include these because when I added a layer with one of these fields, the half of the data that has the opposite the field was removed.  I did not see a way to layer data that only applies to some of the fields in the lower layers. 

I added a description to the text of the Block as in the Vega example.

domoritz:
> In the data there are multiple objects with the same value for day (since there is more than a week of data and some days, like Saturday and Sunday, both use the same letter). If I use day for the x axis, then the data overlaps and there is only an entry for each unique day value.

This would not be an issue if the data had the full date instead of just the day. A workaround is to use the id to determine the x location and concatenate the layered chart with a chart that uses a text mark for the days. 

> Besides the record and normal temps that are currently included, the first half of the objects in the data have an "actual" temp while the second half have a "forecast" temp instead. I did not include these because when I added a layer with one of these fields, the half of the data that has the opposite the field was removed.

Could you paste a spec here so I can play with this for a bit and see whether I can find a way and give you a hint.

melissatdiamond:
{
    "name": "10 Day Historical Weather",
    "description": "A layered bar chart with floating bars",
    "data":{ "url": "https://raw.githubusercontent.com/vega/vega-datasets/gh-pages/data/weather.json" },
    "layer": [
        {
            "mark": {
                "type": "bar", "style": "box"
            },
            "encoding": {
                "y": {
                    "field": "record.low",
                    "type": "quantitative",
                    "scale": {"domain": [10,70]},
                    "axis": {"title": "Temperature"}
                },
                "y2": {
                    "field": "record.high",
                    "type": "quantitative"
                },
                "x": {
                    "field": "id",
                    "type": "ordinal",
                    "axis": {"title": "Day"}
                },
                 "size": {"value": 15},
                 "color": {"value": "orange"}
            }
        },
        {
            "mark": {
                "type": "bar", "style": "box"
            },
            "encoding": {
                "y": {
                    "field": "normal.low",
                    "type": "quantitative"
                },
                "y2": {
                    "field": "normal.high",
                    "type": "quantitative"
                },
                "x": {
                    "field": "id",
                    "type": "ordinal"
                },
                "size": {"value": 15}
            }
        }
    ]
}

Thanks!

domoritz:
Can you paste the spec that has the forecast data in it?

Just TYI, you can edit a comment if you want to update the spec in place.
screen shot 2018-02-07 at 00 27 54

melissatdiamond:
Updated the comment above (not sure if it notifies you when I edit).

domoritz:
Thanks for the message. No, it doesn't.

domoritz:
@melissatdiamond, you found an issue in the Vega-Lite compiler. I filed it at https://github.com/vega/vega-lite/issues/3368. Unfortunately, there is no way for you to proceed right now until we resolve that issue.

melissatdiamond:
Good to know. For now, I added a layer with a text mark of the days and removed the numbers labels, so the data is url and not inline.

chanwutk:
You can also use color. I think that would be easier than using line and text.

melissatdiamond:
I could definitely use color. As of now I was attempting to replicate the Vega example, which doesn't differentiate between predicted and observed temps besides through the shape of the bars. Should I stick with issue #1486 and just do replication or should I change it more?

melissatdiamond:
Great! Removed the style tag.

domoritz:
Lgtm.

domoritz:
:tada

PR #3367, opened 2018-02-08
Add offset as mark properties

### LINKED ISSUE #3328: offset as mark definition properties
Forked from https://github.com/vega/vega-lite/issues/396, we discussed earlier today that it might make more sense to provide support for `offsetX`, `offsetY`, `offsetX2`, and `offsetY2` in [mark definition](https://vega.github.io/vega-lite/docs/mark.html#mark-def) first.

- [x] Add these properties (`offsetX`, `offsetY`, `offsetX2`, and `offsetY2`) to `MarkDef`

- [x] Use them in related position mixins in https://github.com/vega/vega-lite/blob/master/src/compile/mark/mixins.ts –– there are many mixins methods with Position in the name.  Make sure the offset are correctly included in all the cases.  (Modifying some of the help refs method in https://github.com/vega/vega-lite/blob/master/src/compile/mark/valueref.ts may actually make it work for all cases. 

- [x] Write test for different scales on x,y (linear, bin-linear, band, point, time, etc.)

starry97:
close #3328

### DIFF
--- a/src/compile/mark/valueref.ts
+++ b/src/compile/mark/valueref.ts
@@ -13,6 +13,7 @@ import {
   TextFieldDef,
   vgField,
 } from '../../fielddef';
+import {MarkDef} from '../../mark';
 import {hasDiscreteDomain, ScaleType} from '../../scale';
 import {StackProperties} from '../../stack';
 import {LATITUDE, LONGITUDE, QUANTITATIVE} from '../../type';
@@ -49,6 +50,20 @@ export function stackable2(channel: 'x2' | 'y2', aFieldDef: ChannelDef<string>, 
     return fieldRef(aFieldDef, scaleName, {suffix: 'start'});
   }
   return midPoint(channel, a2fieldDef, scaleName, scale, stack, defaultRef);
+}
+
+
+
+export function getOffset(channel: 'x' | 'y' | 'x2' | 'y2', markDef: MarkDef) {
+  const offsetChannel = channel + 'Offset';
+  // TODO: in the future read from encoding channel too
+
+  const markDefOffsetValue = markDef[offsetChannel];
+  if (markDefOffsetValue) {
+    return markDefOffsetValue; // Should this be value or not
+  }
+
+  return undefined;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/mark.ts
```diff
@@ -79,6 +79,26 @@ export interface MarkConfig extends VgMarkConfig {
    * __Note:__ This property cannot be used in a [style config](mark.html#style-config).
    */
   color?: string;
+
+  /**
+   * Offset for x channel in pixels. The offset is applied before rotation by the _angle_ property.
```
kanitw: Angle isn't supported for all marks, so I would just cut the part about angle to avoid confusion.

src/compile/mark/valueref.ts
```diff
@@ -51,6 +52,20 @@ export function stackable2(channel: 'x2' | 'y2', aFieldDef: ChannelDef<string>,
...
 
+
+
+export function getOffset(channel: 'x' | 'y' | 'x2' | 'y2', markDef: MarkDef) {
+  const offsetChannel = channel + 'Offset';
+  // TODO: in the future read from encoding channel too
+
+  const markDefOffsetValue = markDef[offsetChannel];
+  if (markDefOffsetValue) {
+    return markDefOffsetValue; // Should this be value or not
```
kanitw: Why leaving a comment here?

src/mark.ts
```diff
@@ -81,22 +81,22 @@ export interface MarkConfig extends VgMarkConfig {
   color?: string;
 
   /**
-   * Offset for x channel in pixels. The offset is applied before rotation by the _angle_ property.
+   * Offset for x channel in pixels.
```
kanitw: actually, one more nitpick --  
Can we say 
 "Offset for x-position"

instead of x channel

kanitw:
- Can you also add the new properties to the mark.md (in the [mark definition object section](https://vega.github.io/vega-lite/docs/mark.html#mark-def))
- I'm also changing the merge target to "2.x-annotation"

kanitw:
test failing - I think you forgot to update schema.  But it's otherwise good.

Issue #3150, opened 2017-10-19
Composite examples missing


domoritz:
See https://vega.github.io/vega-lite/docs/compositemark.html

domoritz:
Hey @mattwchun, are you fixing this?

mattwchun:
@domoritz I won't be able to get to it for a couple of weeks but I can definitely fix it.

domoritz:
Since we don't ship this in 2.0, it's fine. Just wanted to make sure
somebody is looking at it.

PR #3377, opened 2018-02-13
Fixed Missing Boxplot Examples

mattwchun:
Made change to the filename that was wrong in all the examples. Refers to #3150.


domoritz:
Thanks!

PR #3379, opened 2018-02-13
Update PULL_REQUEST_TEMPLATE.md

domoritz:
Examples is only necessary in a few cases.


kanitw:
Instead of removing it entirely, better describe when they are needed.  (Otherwise, people might not know they're supposed to run it.)

domoritz:
Updated. Feel free to squash.

PR #3381, opened 2018-02-14
Consistently use font weight.

domoritz:


src/vega.schema.ts
```diff
@@ -1276,7 +1278,7 @@ export interface VgTitleConfig {
   /**
    * Font weight for title text.
    */
-  fontWeight?: FontWeight | FontWeightNumber;
+  fontWeight?: FontWeight;
```
kanitw: Take this chance to update the description to says that this could be either string e.g,. ("bold") or a number (100, 200, 300, ..., 900 where normal = 400 and bold = 700)?

domoritz: Good idea.

domoritz:
Pushed update

PR #3385, opened 2018-02-15
Add Fremont Bridge bicycle count visualization

jakevdp:


kanitw:
Another nice example :)  Thank you!

domoritz:
I wonder whether it would make sense to ad some opacity or make the lines thinner.

jakevdp:
@domoritz do you mean in the bottom panel in particular?

domoritz:
Yes. The top one can be zoomed so it doesn't matter much.

PR #3386, opened 2018-02-15
Do not group by cursor

domoritz:


kanitw:
Cursor isn't an encoding channel  (yet) so not sure what's this for?

Issue #3387, opened 2018-02-15
Legend title=false is shown as "false"


nyurik:
In an encoding channel, using `"legend": {"title": false}` shows a string "false" instead of the expected no title.

kanitw:
`title` has a string type,  so `false` would be coerced to a string.   Please use `""` or `null` instead.

nyurik:
Thanks, makes sense. Could VL  throw an error, or at least a warning?  If the value has unexpected type, it is not a valid spec, and this would allow users to self-correct things much more rapidly?

kanitw:
Actually, let me just add a check for false too.  It's kinda dumb to coerce type here anyway.

kanitw:
(I won't add boolean to the schema / docs, but I think it's a reasonable fallback not to show anything)

domoritz:
Agreed with @kanitw. This is a good starter task.

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/3388#pullrequestreview-97052085

PR #3388, opened 2018-02-16
Make `title=false` means title is hidden

kanitw:
Fix https://github.com/vega/vega-lite/issues/3387

### DIFF
--- a/src/compile/axis/parse.ts
+++ b/src/compile/axis/parse.ts
@@ -317,7 +317,12 @@ function getProperty<K extends keyof VgAxis>(property: K, specifiedAxis: Axis, c
     case 'ticks':
       return properties.ticks('ticks', specifiedAxis, isGridAxis, channel);
     case 'title':
-      return getSpecifiedOrDefaultValue(specifiedAxis.title, properties.title(specifiedAxis.titleMaxLength, fieldDef, model.config)) || undefined;
+      return getSpecifiedOrDefaultValue(
+        // For falsy value, keep undefined so we use default,
+        // but use null for '', null, and false to hide the title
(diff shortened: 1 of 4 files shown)


src/compile/axis/parse.ts
```diff
@@ -317,7 +317,12 @@ function getProperty<K extends keyof VgAxis>(property: K, specifiedAxis: Axis, c
     case 'ticks':
       return properties.ticks('ticks', specifiedAxis, isGridAxis, channel);
     case 'title':
-      return getSpecifiedOrDefaultValue(specifiedAxis.title, properties.title(specifiedAxis.titleMaxLength, fieldDef, model.config)) || undefined;
+      return getSpecifiedOrDefaultValue(
+        // For falsy value, keep undefined so we use default,
+        // but use '' for '' and false to hide the title
+        specifiedAxis.title || (specifiedAxis.title === undefined ? undefined : ''),
+        properties.title(specifiedAxis.titleMaxLength, fieldDef, model.config)
+      ) || undefined;
```
domoritz: I think this `|| undefined` is not needed anymore.

test/compile/axis/parse.test.ts
```diff
@@ -113,20 +113,22 @@ describe('Axis', function() {
...
 
-    it('should not set title if title  = null', function() {
-      const model = parseUnitModelWithScale({
-        mark: "point",
-        encoding: {
-          x: {
-            field: "a",
-            type: "quantitative",
-            axis: {title: null}
+    it('should not set title if title  = null, \'\', or false', function () {
```
domoritz: use "" instead of \'\'

PR #3391, opened 2018-02-17
Use varName to replace . with _ in bin signals. 

### LINKED ISSUE #3390: Property "step" in histogram produces no output when "step" < 1
I am fairly new to Vega-Lite, so this may be due to my relative ignorance, but there's something I stumbled upon while playing around in the [online Vega-Lite Editor](https://vega.github.io/editor/#/custom/vega-lite) that may be a bug or at perhaps something that could use a note in the documentation.

I've been trying to create a histogram while specifying the step-size with the `step` property. 

The following will produce an output as expected:

```
{
  "transform": [{
    "bin": {"step":0.5},
    "field": "IMDB_Rating",
    "as": "bin_IMDB_Rating"
  }],
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "bin_IMDB_Rating",
      "type": "ordinal"
    },
    "y": {
      "aggregate": "count",
      "type": "quantitative"
    }
  }
}
```

screen shot 2018-02-16 at 2 46 23 pm

However, when I specify `step` in the `x` property within `encoding`, it will *only* create a plot when `step` is *integer*:

```
{
  "data": {
    "url": "data/movies.json",
    "format": {
      "type": "json"
    }
  },
  "mark": "bar",
  "encoding": {
    "x": {
      "bin": {"step": 0.5},
      "field": "IMDB_Rating",
      "type": "quantitative"
    },
    "y": {
      "aggregate": "count",
      "type": "quantitative"
    }
  }
}
```
This produces:

screen shot 2018-02-16 at 2 50 33 pm

Is this expected behaviour? Does the property `step` mean different things when used within `encoding` than when it does for `transform`?

domoritz:
Fixes #3390

### DIFF
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -31,32 +31,18 @@ function binKey(bin: BinParams, field: string) {
   return `${binToString(bin)}_${field}`;
 }
 
-function isModelParams(p: {model: Model} | {signal?: string, extentSignal?: string}): p is {model: Model} {
-  return !!p['model'];
-}
-
-function getSignalsFromParams(
-  params: {model: Model} | {signal?: string, extentSignal?: string},
-  key: string
-) {
-  if (isModelParams(params)) {
-    const model = params.model;
-    return {
-      signal: model.getName(`${key}_bins`),
-      extentSignal: model.getName(`${key}_extent`)
-    };
-  }
-  return params;
+function getSignalsFromModel(model: Model, key: string) {
+  return {
+    signal: model.getName(`${key}_bins`),
+    extentSignal: model.getName(`${key}_extent`)
+  };
 }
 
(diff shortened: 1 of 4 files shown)


src/compile/data/bin.ts
```diff
@@ -126,7 +126,7 @@ export class BinNode extends DataFlowNode {
    * Creates a bin node from BinTransform.
    * The optional parameter should provide
    */
-  public static makeFromTransform(t: BinTransform, params: {model: Model} | {signal?: string, extentSignal?: string}) {
```
domoritz: @kanitw note that I just removed this and things still compile. Not sure why this was needed.

kanitw: someone forgot to remove it?

kanitw: Use blame to check?

PR #3394, opened 2018-02-20
use vega-util's error instead of util.error

rgbkrk:
[`util.error` is the deprecated predecessor to `console.error`](https://nodejs.org/api/util.html#util_util_error_strings).

As suggested in this PR, this uses `vega-util`'s error, which _will_ throw rather than erroring in the console.

Can this also get backported to the 1.x line?

### DIFF
--- a/src/compile/projection/assemble.ts
+++ b/src/compile/projection/assemble.ts
@@ -1,4 +1,4 @@
-import {error} from 'util';
+import {error} from 'vega-util';
 import {contains} from '../../util';
 import {isVgSignalRef, VgProjection, VgSignalRef} from '../../vega.schema';
 import {isConcatModel, isLayerModel, isRepeatModel, Model} from '../model';


kanitw:
Thank for proposing.  However, our util is not node.js util.   Using our util means that users can redirect the error/warnings to other log listener instead of the console.  Thus, we won't accept this change.

kanitw:
Actually, this one uses node.js util by mistake.  So this is a valid point.

kanitw:
But we should replace it with our util's error, not console.error.

rgbkrk:
Happy to switch it over to your util's error too, gimme a sec here.

kanitw:
Actually there are in here: https://github.com/vega/vega-lite/blob/master/src/log.ts, but I can't remember why we only have warn, info, but not error.  @domoritz any idea?

rgbkrk:
Ok, just let me know which to use and I'm happy to switch this up.

domoritz:
I think we just throw exceptions when we have an error as there is no way to recover from errors.

jheer:
Yes, the vega-util [`error`](https://github.com/vega/vega-util/blob/master/src/error.js) method is simply a convenience for throwing an `Error`. It is handy as it enables use within a ternary expression, plus I believe using it (ever so slightly) reduces the code size.

rgbkrk:
Switched this over to `vega-util`s `error` to keep it consistent, updating title + description now.

kanitw:
Thanks

rgbkrk:
Same with #3395, can you backport this change to 1.x?

kanitw:
Sure. Please submit another PR(s) with target branch = `1.x`.  You might want to fork from `1.x` though as the code has changed a lot since then.

PR #3395, opened 2018-02-20
switch from node's util to vega-util for isString

rgbkrk:
`util.isString` [is deprecated](https://nodejs.org/api/util.html#util_util_isstring_object) and `vega-util` provides `isString` as well.

I'd like to get rid of all usage of node's `util` in this project (also see #3394), to make it simpler to work with webpack and conflicting definitions of browser side `util` versions. Unfun story at the moment.

I'd also love to see this change as part of the 1.x line.

### DIFF
--- a/src/toplevelprops.ts
+++ b/src/toplevelprops.ts
@@ -1,4 +1,5 @@
-import {isString} from 'util';
+import {isString} from 'vega-util';
+
 import * as log from './log';
 
 /**


src/toplevelprops.ts
```diff
@@ -58,7 +57,7 @@ export interface AutoSizeParams {
 }
 
 function _normalizeAutoSize(autosize: AutosizeType | AutoSizeParams) {
-  return isString(autosize) ? {type: autosize} : autosize || {};
+  return typeof autosize === 'string' ? {type: autosize} : autosize || {};
```
kanitw: Let's use isString in our local util instead?

rgbkrk: Done, and rebased.

rgbkrk:
Addressed the linting nit triggered by CI as well.

kanitw:
Thanks ;)

rgbkrk:
No prob, can you also backport this change to 1.x?

PR #3397, opened 2018-02-20
Update to JSON schema draft 6.

domoritz:


kanitw:
Does VSCode (and its open source) module support draft 6 yet?   If not, we shouldn't migrate?

domoritz:
Autocomplete works and I haven't seen any problems. Not even with the new meta schema for draft 06. I think it's ready to go.

domoritz:
Agreed

PR #3400, opened 2018-02-21
State nested field issue in argmin/argmax example description and aggregate page

chanwutk:
**In Vega-lite Doc:**
- In Aggregate page: add note about nested field issue on `argmin`/`argmax` and provide solution in the bottom of the page and link to the issue page and the example page for solution.

**In example page:**
- Add not in the description about the solution of the nested field issue on `argmin`/`argmax` and link to the issue page

### DIFF
--- a/site/docs/transform/aggregate.md
+++ b/site/docs/transform/aggregate.md
@@ -118,4 +118,4 @@ The supported **aggregation operations** are:
 | argmin    | An input data object containing the minimum field value.|
 | argmax    | An input data object containing the maximum field value.|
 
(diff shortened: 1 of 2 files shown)


site/docs/transform/aggregate.md
```diff
@@ -117,3 +117,5 @@ The supported **aggregation operations** are:
 | max       | The maximum field value.|
 | argmin    | An input data object containing the minimum field value.|
 | argmax    | An input data object containing the maximum field value.|
+
+__Note:__ When accessing aggregated argmax/argmin fields, the aggregated fields must be flatten using calculate, due to [issue about nested field #3361](https://github.com/vega/vega-lite/issues/3361),  like [this example](../../../vega-lite/examples/layer_line_co2_concentration.html).
```
kanitw: Thanks for adding this.  

One nitpick: can we have a simpler relative path (without "vega-lite")?

PR #3401, opened 2018-02-21
Added composite mark callout

starry97:


src/compositemark/callout.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export interface CalloutConfig extends MarkConfig {
+  calloutOffset: {x: number, y: number};
+  calloutSize: {x: number, y: number};
+  labelOffset:{x: number, y: number};
+}
+
+export function normalizeCallout(spec: GenericUnitSpec<Encoding<string>, CALLOUT>): LayerSpec {
+  const {mark: mark, selection: _sel, projection: _p, encoding, ...outerSpec} = spec;
```
kanitw:
```
// TODO:  determine what's the general rule for applying selection for composite marks
```

Add todo (we don't have good general rule yet)

src/compositemark/callout.ts
```diff
@@ -0,0 +1,74 @@
...
+  let calloutOffset: {x: number, y: number} = undefined;
+  let calloutSize: {x: number, y: number} = undefined;
+  let labelOffset: {x: number, y: number} = undefined;
+
+  const defaultOffset = {x: 0, y: 0};
```
kanitw: I have a few issues here

1.  Why should default offset for call out and label be the same?  More importantly, calloutSize shouldn't be zero. 

Instead of inventing offset out of nowhere here.  We should have each of these in `config.callout` with default values already set (just like we have `size` for box -- and read the values from there).

src/compositemark/callout.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (isCalloutDef(mark)) {
+    calloutOffset = mark.calloutOffset ? mark.calloutOffset : defaultOffset;
+    calloutSize = mark.calloutSize ? mark.calloutSize : defaultOffset;
+    labelOffset = mark.labelOffset ? mark.labelOffset : defaultOffset;
+  }
+  const {text: textEncoding, ...encodingWithoutText} = encoding;
+  if (!textEncoding) {
+    // throw error since it must provides an text
```
kanitw: an text => a text

src/compositemark/callout.ts
```diff
@@ -0,0 +1,74 @@
...
+    throw new Error('Must have a text encoding in composite mark callout');
```
kanitw: Why throwing an error?  Imagine you're still typing the spec in the editor.  Instead of getting an intermediate stage without text (text =""), this will unnecessarily crash the plot, which is not an ideal user experience.

src/compositemark/callout.ts
```diff
@@ -0,0 +1,74 @@
...
+          xOffset: labelOffset.x + calloutSize.x + calloutOffset.x,
+          yOffset: labelOffset.y + calloutSize.y + calloutOffset.y
+        },
+        encoding
+      }, { // callout
+        mark: {
+          type: 'rule',
+          xOffset: calloutOffset.x,
+          yOffset: calloutOffset.y,
+          x2Offset: calloutSize.x,
```
kanitw: you have to add calloutOffset too.

src/compositemark/callout.ts
```diff
@@ -0,0 +1,74 @@
...
+          y2Offset: calloutSize.y
+        },
+        encoding: {
+          x: encoding.x,
```
kanitw: why do you have to repeat x and y,  aren't they included in `encodingWithoutText` already?

`x2` and `y2` makes sense though.

src/compositemark/callout.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (!textEncoding) {
+    // throw error since it must provides an text
+    throw new Error('Must have a text encoding in composite mark callout');
+  }
+  const returnedSpec: LayerSpec = {
+    ...outerSpec,
+    layer: [
+      { // label
+        mark: {
+          type: 'text',
```
kanitw: missing style -- maybe "callout-label" ?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,74 @@
...
+          xOffset: labelOffset.x + calloutSize.x + calloutOffset.x,
+          yOffset: labelOffset.y + calloutSize.y + calloutOffset.y
+        },
+        encoding
+      }, { // callout
+        mark: {
+          type: 'rule',
```
kanitw: missing style -- maybe "callout-rule" ?

src/compositemark/callout.ts
starry97: Add calloutOffset in mark property?

kanitw: As discussed in person, the math here is a bit incorrect.  :)

kanitw:
@starry97  -- one possibility for all of these offset thing is having `calloutAngle` and then makes all of the `calloutOffset`, `calloutSize` (maybe rename to `calloutLength`), `labelOffset` just `number` rather than `{x,y}` -- then you can do the math to calculate things.

kanitw:
Rebase onto `2.x-annotation`?  (The first commit is redundant)

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+export const CALLOUT: 'callout' = 'callout';
+export type CALLOUT = typeof CALLOUT;
+
+export interface CalloutDef {
+  type: CALLOUT;
+  calloutAngle?: number;
+  calloutOffset?: number;
+  calloutLength?: number;
+  labelOffset?: number;
+  callout?: MarkDef;
```
kanitw: MarkDef is a bit too much?   You don't want them to change the mark type of the `callout`.

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+  label?: MarkDef;
+}
+
+export function isCalloutDef(mark: CALLOUT | CalloutDef): mark is CalloutDef {
+  return !!mark['type'];
+}
+
+export type CalloutStyle = 'callout-label' | 'callout-rule';
+
+export const CALLOUT_STYLES: CalloutStyle[] = ['callout-label', 'callout-rule'];
```
kanitw: We should decide if we want dash or camelCase and be consistent across all composite marks.  

- camelCase is good as this will be a property `config.style.calloutLabel` -- having dash will be awkward?  
- but normally we should dash for property names.  Also we should just redirect `config.callout.label` to `config.style.calloutLabel` / `config.style["callout-label"]` anyway 
   - the question is do we do `config.callout.rule` or just `config.callout` for the part that will get redirected to `config.style.calloutRule` / `config.style["callout-rule"]` 

@domoritz  -- any opinion?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export interface CalloutConfig extends MarkConfig {
```
kanitw: Why does this extend mark config?  

For example, if I set "config.callout.size", would this affect the callout in some way?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  }
+
+  const returnedSpec: LayerSpec = {
+    ...outerSpec,
+    layer: [
+      { // label
+        mark: {
+          type: 'text',
+          style: 'callout-label',
```
kanitw: Need to decide if we use dash or camelCase throughout the project

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+          xOffset: labelOffsetCoordinate.x,
+          yOffset: labelOffsetCoordinate.y
+        },
+        encoding
+      }, { // callout
+        mark: {
+          type: 'rule',
+          style: 'callout-rule',
```
kanitw: Need to decide if we use dash or camelCase throughout the project

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+          ...encodingWithoutText
+        }
+      }
+    ]
+  };
+  return returnedSpec;
+}
+
+export function getCoordinateFromAngleAndLength(angle: number, length: number): {x: number, y: number} {
+  // tan(angle) = y/x
```
kanitw: I don't understand why this function has to be this complicated with so many corner cases. 

Can't you just use sine and cosine?

src/compositemark/index.ts
```diff
@@ -21,23 +22,25 @@ export function remove(mark: string) {
...
-export type CompositeMark = BOXPLOT | ERRORBAR;
+export type CompositeMark = BOXPLOT | ERRORBAR | CALLOUT;
 
-export type CompositeMarkDef = BoxPlotDef;
+export type CompositeMarkDef = BoxPlotDef | CalloutDef;
 
 export type CompositeAggregate = BOXPLOT;
 
-export const COMPOSITE_MARK_STYLES = BOXPLOT_STYLES;
+export const COMPOSITE_MARK_STYLES = BOXPLOT_STYLES || CALLOUT_STYLES;
```
kanitw: Why `||`?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+  const calloutAngle: number = mark.calloutAngle ? mark.calloutAngle : config.callout.calloutAngle;
+  const calloutOffset: number = mark.calloutOffset ? mark.calloutOffset : config.callout.calloutOffset;
+  const calloutLength: number = mark.calloutLength ? mark.calloutLength : config.callout.calloutLength;
+  const labelOffset: number = mark.labelOffset ? mark.labelOffset : config.callout.labelOffset;
+
+  const calloutOffsetCoordinate: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset);
+  const calloutLengthCoordinate: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset + calloutLength);
+  const labelOffsetCoordinate: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset + calloutLength + labelOffset);
+
+  const {text: textEncoding, ...encodingWithoutText} = encoding;
```
kanitw: I'm not sure if we should just copy size for both of them:

`size` for `text` is `fontSize`
`size` for `rule` is invalid

only copy size to text?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+    return null;
+  }
+
+  const calloutAngle: number = mark.calloutAngle ? mark.calloutAngle : config.callout.calloutAngle;
+  const calloutOffset: number = mark.calloutOffset ? mark.calloutOffset : config.callout.calloutOffset;
+  const calloutLength: number = mark.calloutLength ? mark.calloutLength : config.callout.calloutLength;
+  const labelOffset: number = mark.labelOffset ? mark.labelOffset : config.callout.labelOffset;
+
+  const calloutOffsetCoordinate: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset);
+  const calloutLengthCoordinate: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset + calloutLength);
```
kanitw: The name is misleading.   This is not coordinate of "calloutLength`.   


Call them `calloutOffsetCoor1` and `calloutOffsetCoor2` ?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+  const labelOffsetCoordinate: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset + calloutLength + labelOffset);
```
kanitw: this is slightly from the input labelOffset though,  maybe `labelTotalOffsetCoor`?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+import {MarkConfig, MarkDef} from '../mark';
+import {GenericUnitSpec, LayerSpec} from '../spec';
+
+
+export const CALLOUT: 'callout' = 'callout';
+export type CALLOUT = typeof CALLOUT;
+
+export interface CalloutDef {
+  type: CALLOUT;
+  calloutAngle?: number;
```
kanitw: You should share things that are similar between `CalloutDef` and `CalloutConfig` so that you don't have the write documentation twice.  

Also, missing documentation for all of these properties here.

src/compositemark/callout.ts
```diff
@@ -0,0 +1,131 @@
...
+          y2: encoding.y,
+          ...encodingWithoutText
+        }
+      }
+    ]
+  };
+  return returnedSpec;
+}
+
+export function getCoordinateFromAngleAndLength(angle: number, length: number): {x: number, y: number} {
```
kanitw: Does this function have to be this complicated with so many corner cases?

I think you should be able to just use sine and cosine?  (If you don't get it, revise about vector projection)

src/compositemark/callout.ts
```diff
@@ -1,30 +1,25 @@
...
-export interface CalloutDef {
+export interface CalloutDef extends CalloutConfig {
   type: CALLOUT;
-  calloutAngle?: number;
-  calloutOffset?: number;
-  calloutLength?: number;
-  labelOffset?: number;
-  callout?: MarkDef;
-  label?: MarkDef;
+  callout?: MarkConfig;
```
kanitw: callout mark is already the whole thing so I don't think we should call the line segment part a `callout`.   After all, this should be consistent with config, and `config.callout.callout` is confusing.

kanitw: I wonder if it makes sense to support `style` and `clip` too, but definitely not type?

(Maybe not -- I'm just asking)

src/compositemark/callout.ts
```diff
@@ -39,12 +34,16 @@ export interface CalloutConfigMixins {
...
+  calloutLabel?: MarkConfig;
+  calloutRule?: MarkConfig;
 }
 
 export const VL_ONLY_CALLOUT_CONFIG_PROPERTY_INDEX: {
   [k in keyof CalloutConfigMixins]?: (keyof CalloutConfigMixins[k])[]
 } = {
-  callout: ['calloutAngle', 'calloutOffset', 'calloutLength', 'labelOffset']
+  callout: ['calloutAngle', 'calloutOffset', 'calloutLength', 'labelOffset'],
+  calloutLabel: ['color'],
```
kanitw: why is color in here?

src/compositemark/callout.ts
```diff
@@ -39,12 +34,16 @@ export interface CalloutConfigMixins {
    * @hide
    */
   callout?: CalloutConfig;
+  calloutLabel?: MarkConfig;
```
kanitw: let's remove `config.calloutLabel`, `config.calloutRule` and make `config.callout.label/rule`.

src/compositemark/callout.ts
```diff
@@ -1,30 +1,25 @@
...
+  callout?: MarkConfig;
+  label?: MarkConfig;
 }
 
 export function isCalloutDef(mark: CALLOUT | CalloutDef): mark is CalloutDef {
   return !!mark['type'];
 }
 
-export type CalloutStyle = 'callout-label' | 'callout-rule';
+export type CalloutStyle = 'calloutLabel' | 'calloutRule';
```
kanitw: Wonder why you prefer camelcase?   (I start liking dash more at this point, but wanna hear your rationale too.)

src/compositemark/callout.ts
kanitw: These two (`label` and whatever you gonna call `rule/line/callout`) should be in CalloutConfig anyway.

src/compositemark/callout.ts
kanitw: My rationale is we most use camelCase for property keys, but dash-case for property values.

Since style name is a value, then dash-case is better.

src/compositemark/callout.ts
```diff
@@ -0,0 +1,114 @@
...
+  callout?: CalloutConfig;
+}
+
+
+export function normalizeCallout(spec: GenericUnitSpec<Encoding<string>, Callout | CalloutDef>, config: Config): LayerSpec {
+  // TODO:  determine what's the general rule for applying selection for composite marks
+  const {mark, selection: _sel, projection: _p, encoding, ...outerSpec} = spec;
+  const markDef = isMarkDef(mark) ? mark : {type: mark};
+
+  const calloutAngle: number = markDef.calloutAngle || config.callout.calloutAngle;
```
kanitw: This is probably a better pattern

```
const markDef = {
    ...config.callout
    ...isMarkDef(mark) ? mark : {type: mark}
};

const {calloutAngle, calloutOffset, calloutLength, labelOffset} = markDef;
```


(If this is applicable for boxplot, please submit a PR to refactor too.)

src/compositemark/callout.ts
```diff
@@ -0,0 +1,114 @@
...
+   */
+  labelOffset?: number;
+}
+
+export interface CalloutDef extends GenericMarkDef<Callout>, CalloutConfig {}
+
+export interface CalloutConfigMixins {
+  /**
+   * Callout Rule Config
+   * @hide
```
kanitw: No need to hide anymore

src/compositemark/callout.ts
```diff
@@ -0,0 +1,114 @@
...
+   * __Default value: `"45"`
+   */
+  calloutAngle?: number;
+  /**
+   * Offset of callout line.
+   * __Default value: `"0"`
+   */
+  calloutOffset?: number;
+  /**
+   * Length of callout length.
```
kanitw: Length of callout

src/compositemark/callout.ts
```diff
@@ -0,0 +1,114 @@
...
+};
+
+export interface CalloutConfig extends CalloutPartsMinxins {
+  /**
+   * Angle of callout line.
+   * __Default value: `"45"`
+   */
+  calloutAngle?: number;
+  /**
+   * Offset of callout line.
```
kanitw: Offset distance between the data point and the callout line

src/compositemark/callout.ts
```diff
@@ -0,0 +1,114 @@
...
+   * __Default value: `"0"`
+   */
+  calloutOffset?: number;
+  /**
+   * Length of callout length.
+   * __Default value: `"30"`
+   */
+  calloutLength?: number;
+  /**
+   * Offset of callout label.
```
kanitw: Offset distance between callout line and label

src/compositemark/callout.ts
```diff
@@ -0,0 +1,114 @@
...
+  // TODO:  determine what's the general rule for applying selection for composite marks
+  const {mark, selection: _sel, projection: _p, encoding, ...outerSpec} = spec;
+  const markDef = isMarkDef(mark) ? mark : {type: mark};
+
+  const calloutAngle: number = markDef.calloutAngle || config.callout.calloutAngle;
+  const calloutOffset: number = markDef.calloutOffset || config.callout.calloutOffset;
+  const calloutLength: number = markDef.calloutLength || config.callout.calloutLength;
+  const labelOffset: number = markDef.labelOffset || config.callout.labelOffset;
+
+  const calloutOffsetCoor1: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset);
```
kanitw: Do you really need `: {x: number, y: number}` for all of these?  Seems a bit unnecessarily verbose.

src/compositemark/callout.ts
```diff
@@ -0,0 +1,114 @@
...
+  const calloutOffsetCoor2: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset + calloutLength);
+  const labelTotalOffsetCoor: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset + calloutLength + labelOffset);
+
+  const {text: textEncoding, size: sizeEncoding, ...encodingWithoutTextAndSize} = encoding;
+  if (!textEncoding) {
+    log.warn('callout mark should have text encoding');
+  }
+
+  const returnedSpec: LayerSpec = {
```
kanitw: just return it?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,114 @@
...
+  const calloutAngle: number = markDef.calloutAngle || config.callout.calloutAngle;
+  const calloutOffset: number = markDef.calloutOffset || config.callout.calloutOffset;
+  const calloutLength: number = markDef.calloutLength || config.callout.calloutLength;
+  const labelOffset: number = markDef.labelOffset || config.callout.labelOffset;
+
+  const calloutOffsetCoor1: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset);
+  const calloutOffsetCoor2: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset + calloutLength);
+  const labelTotalOffsetCoor: {x: number, y: number} = getCoordinateFromAngleAndLength(calloutAngle, calloutOffset + calloutLength + labelOffset);
+
+  const {text: textEncoding, size: sizeEncoding, ...encodingWithoutTextAndSize} = encoding;
```
kanitw: `const {text, size, ...encodingWithoutTextAndSize} = encoding` is more concise?

src/config.ts
```diff
@@ -270,6 +270,16 @@ export const defaultConfig: Config = {
     whisker: {}
   },
 
+
+  callout: {
+    calloutAngle: 45,
+    calloutOffset: 0,
```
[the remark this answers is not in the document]
starry97: Should we add any offset between data points and callout? I feel like by default we should just point at data points.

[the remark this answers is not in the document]
kanitw: try to draw that by hand on paper and see if you still think the same :)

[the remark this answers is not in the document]
kanitw: Thinking more, there is some value for calloutOffset = 0 for line marks.   So I'd say try to make examples and see if this is a default that you're happy with.

kanitw:
Sorry for going back and forth.  I google callout and only see cloud style callout.  This makes me wonder if `calloutline` is a better mark name.  In any case, changing name is not that hard, I'd still keep the name `callout` and address other issues / make examples.

src/compositemark/callout.ts
```diff
@@ -0,0 +1,111 @@
...
+export type CalloutPartsMinxins = {
+  [part in CalloutPart]?: MarkConfig
+};
+
+export interface CalloutConfig extends CalloutPartsMinxins {
+  /**
+   * Angle of callout line.
+   * __Default value: `45`
+   */
+  calloutAngle?: number;
```
kanitw: The mark is called callout anyway.  This could be just the `angle`?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * Offset distance between the data point and the callout line.
+   * __Default value: `0`
+   */
+  calloutOffset?: number;
```
kanitw: You are calling the rule "line",  so `lineOffset`?

src/compositemark/callout.ts
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * Length of callout.
+   * __Default value: `30`
+   */
+  calloutLength?: number;
```
kanitw: lineLength?

src/compositemark/callout.ts
kanitw: (This is `config.callout.angle` anyway.)

src/compositemark/callout.ts
```diff
@@ -0,0 +1,111 @@
...
+   * __Default value: `45`
+   */
+  calloutAngle?: number;
+  /**
+   * Offset distance between the data point and the callout line.
+   * __Default value: `0`
+   */
+  calloutOffset?: number;
+  /**
+   * Length of callout.
```
kanitw: Length of a callout line

src/config.ts
```diff
@@ -270,6 +270,16 @@ export const defaultConfig: Config = {
...
+    calloutLength: 30,
+    labelOffset: 5,
```
kanitw: 5 is a bit too much IMHO.  I think 2 is better.  In any case, your documentation says default is `0`, which doesn't match the number here.

kanitw:
I'm playing with it a little bit.  Here is another example you could polish and consider adding:
```
{
  "layer": [{
    "mark": "line",
    "encoding": {
    }

  }, {
    "transform": [{
      "filter": {
        "field": "date",
        "range": [
          {"year": 2006, "month": 1, "day": 1},
          {"year": 2006, "month": 1, "day": 1}
        ]
      }
    }],
    "mark": {
      "type": "callout",
      "calloutAngle": 90,
      "label": {
        "align": "center"
      },
      "labelOffset": 2
    },
    "encoding": {
      "text": {"value": "high horsepower"},
      "color": {"value": "red"}
    }
  }, {
    "transform": [{
      "filter": {
        "field": "date",
        "range": [
          {"year": 2006, "month": 1, "day": 1},
          {"year": 2006, "month": 1, "day": 1}
        ]
      }
    }],
    "mark": {
      "type": "circle",
      "color": "red"
    },
    "encoding": {
      "text": {"value": "high horsepower"}
    }
  }]
}
```

kanitw:
image

kanitw:
- [ ] Besides `align`, `baseline` is important too.   You might wanna look at @NeelMohapatra's PR recently to see if you can reuse similar rules. 

image

g3o2:
What would this callout look like if it was supposed to represent several data points instead of just one ? Should there be a similar restriction as with `selection`, that is `point` and `interval`?

kanitw:
@g3o2  you can draw multiple callouts just like any other marks.  This is just a macro for drawing line segment(s) and text(s) using just one (composite) mark.

kanitw:
@starry97  -- any update on this?  (Do you still have time to clean it up for a release?)

domoritz:
What's left in this PR? I think this would be awesome to have but I won't have the cycles to finish it.

PBI-David:
Is there any chance of this being resurrected? It looks like a fantastic feature to have.

Issue #3402, opened 2018-02-21
Setting domain for line plot should clip values outside domain


gregnordin:
This is my problem - when the original data extends outside the specified domain, it still gets plotted:
![temp](https://user-images.githubusercontent.com/4380622/36506880-1c991d12-1715-11e8-9f9d-b898e8c97b22.gif)
Shouldn't setting the domain clip the line to stay within the range of the plot? Is this a bug or am I misunderstanding how Vega-Lite should operate? Or perhaps I have implemented something incorrectly? Any insight would be much appreciated.

Here is the JSON that creates the above plot:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "values": [
      {"wavelength": 250, "normalized_power": 1},
      {"wavelength": 300, "normalized_power": 2},
      {"wavelength": 330, "normalized_power": 3},
      {"wavelength": 390, "normalized_power": 4},
      {"wavelength": 420, "normalized_power": 2},
      {"wavelength": 450, "normalized_power": 1.8},
      {"wavelength": 500, "normalized_power": 1.1}
    ]
  },
  "encoding": {
    "x": {
      "axis": {
        "title": "Wavelength (nm)"
      },
      "field": "wavelength",
      "scale": {
        "domain": [
          300,
          450
        ]
      },
      "type": "quantitative"
    },
    "y": {
      "axis": {
        "orient": "left",
        "title": "Normalized LED Power"
      },
      "field": "normalized_power",
      "type": "quantitative"
    }
  },
  "height": 250,
  "mark": "line",
  "width": 400
}
```

kanitw:
Hi, 

This is a desired behavior -- changing scale domain should only change the "domain" of the scale function, but doesn't change the data that get passed through the scale. 

You could instead use a filter transform to filter the data:
```
transform: [{filter: {"field": "wavelength", "range": [300, 450]}}]
```

domoritz:
I was about to post this. You might want to add `scale.zero: false`. 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "values": [
      {"wavelength": 250, "normalized_power": 1},
      {"wavelength": 300, "normalized_power": 2},
      {"wavelength": 330, "normalized_power": 3},
      {"wavelength": 390, "normalized_power": 4},
      {"wavelength": 420, "normalized_power": 2},
      {"wavelength": 450, "normalized_power": 1.8},
[...]

PR #3403, opened 2018-02-21
Add examples to the Scale Domain docs using clip and transform

gregnordin:
As requested in #3402, I have added an example for the `Scale Domains` section of `site/docs/encoding/scale.md` that uses 2 methods to ensure that a plotted line does not go out of bounds of the plot area when restricting the domain of `x` data. One method uses `clip` and the other a `transform` filter. I put a markdown file containing the text to be added and two images in a new temporary folder, `site/docs/encoding/scale_domain`, since I don't know how to directly add these to `scale.md` with the workflow used in this project.


domoritz:
Merged in #3425

PR #3408, opened 2018-02-22
Add downloads example

domoritz:


domoritz:
Hmm, right. We need argmax, I believe. It probably also doesn't make sense to show the downloads for the ongoing month. Maybe we can filter that out somehow. I will wait for argmax to ship before fixing this.

domoritz:
Maybe something like https://twitter.com/domoritz/status/968912322583412736 is more appropriate.

site/examples/index.md
```diff
@@ -37,4 +37,6 @@ Here we list great examples of Vega-Lite visualizations that were created by the
 * [Horizontal Stacked Bar Chart with Labels](https://bl.ocks.org/pratapvardhan/00800a4981d43a84efdba0c4cf8ee2e1) by @pratapvardhan
 * [Interactive stacked time-series](https://bl.ocks.org/jakevdp/1643ebb6853e76c32e47a969f415f3ea) by @jakevdp
 * [Bicycle Count Time-series with Dynamic Scale](https://bl.ocks.org/jakevdp/b511d09ed4e2797234bd6236d7b428f7) by @jakevdp
+* [Vega-Lite downloads](https://bl.ocks.org/domoritz/81008b55ae2e2649eb42f600440f87d2) by @domoritz
+* [Multi-Line Tooltip](https://bl.ocks.org/jakevdp/a414950f61e4b224765f2439dd1f09b9) by @jakevdp
```
kanitw: Why adding multi-line tooltip back?

domoritz: Ups

Issue #1418, opened 2016-06-08
Create bar with negative value example


kanitw:
We need an easy way to move axis in the middle where there are negative values. 

<img width="993" alt="bar_chart_with_negative_values_-_bl_ocks_org" src="https://cloud.githubusercontent.com/assets/111269/15909465/5de920e4-2d7b-11e6-872f-80d4b6221a42.png">

kanitw:
We can manually draw rule or using condition as in https://github.com/vega/vega-lite/pull/3239

kanitw:
Actually, let's make this an issue to create this example once #3239 is merged

pratapvardhan:
Example appears to be added to gallery. Issue is good to be closed?

domoritz:
It's not in the official gallery yet, though.

kanitw:
I'm adding a better version, which uses conditional grid color in https://github.com/vega/vega-lite/pull/6105

PR #3419, opened 2018-02-25
Add bar chart example with Negative values (#1418)

digi0ps:
Adds a bar chart example with negative values with the axis in the middle.


digi0ps:
@kanitw

PR #3420, opened 2018-02-25
yarn tsc works without special script

domoritz:


package.json
```diff
@@ -64,14 +63,14 @@
...
-    "test:debug-examples": "npm run tsc && npm run schema && mocha --recursive --debug-brk --inspect build/examples",
+    "test:debug": "tsc && mocha --recursive --debug-brk --inspect build/test",
+    "test:debug-examples": "tsc && npm run schema && mocha --recursive --debug-brk --inspect build/examples",
     "mocha:test": "mocha --require source-map-support/register --reporter dot --recursive build/test",
     "mocha:examples": "mocha --require source-map-support/register --reporter dot --recursive build/examples",
 
     "codecov": "nyc report --reporter=json && codecov -f coverage/*.json",
     "watch:build": "watchify src/index.ts -p tsify -v -d -s vl -o 'exorcist build/vega-lite.js.map > build/vega-lite.js'",
-    "watch:tsc": "npm run tsc -- -w",
+    "watch:tsc": "tsc -- -w",
```
kanitw: do you still need `--` here?

domoritz: Nope. It's not needed anymore.

domoritz: Should I just delete this entry then? You can just call `yarn tsc -w`.

kanitw: I never use it so fine with me.

Issue #2915, opened 2017-08-23
Custom sort order


kanitw:
Currently, for unit encoding, we can specify custom order for normal (non-facet) encoding channel by specifying `scale.domain` directly.

However, we don't have similar support for 
1) facet channel
2) stack (https://github.com/vega/vega-lite/issues/1734)

Since facet does not have scale domain, we can't use similar syntax.  But facet has `sort` property like other encoding channel.  Thus, I think we should make `sort` support a custom array, which derives a formula like [this](https://github.com/vega/vega-lite/issues/1734#issuecomment-263481177) so that it's safer than setting domain directly (as domain would still include omitted values).  

For stack, we can then also make stack order orders by color/detail (or any other stack-by field) by default.

curran:
Great idea! Being able to specify facet ordering seems like an essential feature.

FWIW, as a Vega-Lite noob, I tried the following:

```
        "column": {
          "field": "variable",
          "type": "ordinal",
          "scale": {
            "type": "ordinal",
            "domain": ["Online news sites","Social media","Printed newspapers","TV"]
          }
        },
```

Then I tried this (your proposal here):

```
        "column": {
          "field": "variable",
          "type": "ordinal",
          "sort": ["Online news sites","Social media","Printed newspapers","TV"]
        },
```

I think the above syntax would be really great.

As a separate thought, if the notion of "ordinal" is there in Vega, and is distinct from "nominal", wouldn't it be nice to be able to specify _in the data definition_ the natural ordering of a column?

kanitw:
The same data field can be treated as ordinal or nominal so it makes more sense to specify it as `sort` of a data field definition like proposed above.

curran:
Makes sense. Thanks for your reply.

domoritz:
Column used to be implemented with scales but now we use Vega layouts (which is more flexible and produces nicer output). Does the code snippet with scale pass JSON schema validation in the editor?

curran:
@domoritz No, it does not pass schema validation.

I discovered the validation in the editor only after trying the configuration (rather blindly) in my code.

domoritz:
Okay, that's what I wanted. When we implement this feature, you'll use a sort order on the `header` property.

curran:
Sounds good!

If I had some time to dig and try to implement this, where would be the place to start?

domoritz:
[...]

PR #3423, opened 2018-02-27
Custom Sort Order

sirahd:
Please:
- [ ] Make your pull request atomic, fixing one issue at a time unless there are many relevant issues that cannot be decoupled.
- [ ] Provide a test case & update the documentation under `site/docs/`
- [ ] Make lint and test pass. (Run `yarn lint` and `yarn test`.  If your change affects Vega outputs of some examples, please run `yarn build:example EXAMPLE_NAME` to re-compile a specific example or `yarn build:examples` to re-compile all examples.)
- [ ] Make sure you have rebased onto the `master` branch.
- [ ] Provide a concise title so that we can just copy it to our release note.

Pro-Tip: https://medium.com/@greenberg/writing-pull-requests-your-coworkers-might-enjoy-reading-9d0307e93da3 is a nice article about writing a nice pull request.

- feature request (#2915)


kanitw:
Thanks.  Before we review, exclude SVG changes first?  If the vega files don't change, don't include SVG changes.

sirahd:
@kanitw Done!

src/compile/data/calculate.ts
```diff
@@ -15,6 +18,32 @@ export class CalculateNode extends DataFlowNode {
     super();
   }
 
+  public static makeAllFromSort(model: ModelWithField): CalculateNode[] {
```
kanitw: The name is quite vague for its very specific purpose. ("Make all" of what?   This is actually a special purposed calculate node, not a generic calculate node.) 

Perhaps,  `makeAllForSortIndex` is better? 

(Also feel free to suggest better name.)

examples/specs/bar_custom_sort.vl.json
```diff
@@ -0,0 +1,16 @@
...
+  "data": {
+    "values": [
+      {"a": "A","b": 28}, {"a": "B","b": 55}, {"a": "C","b": 43},
+      {"a": "D","b": 91}, {"a": "E","b": 81}, {"a": "F","b": 53},
+      {"a": "G","b": 19}, {"a": "H","b": 87}, {"a": "I","b": 52}
+    ]
+  },
+  "mark": "bar",
+  "encoding": {
+    "x": {"field": "a", "type": "ordinal", "sort": ["D", "G", "F"]},
```
kanitw: Maybe it's better distinguish between (1) the example that show how sort order normally works when all items are included and (2) another example to explain behavior when some values is "accidentally" ignored?


The first can just have three values B, A, C. 
The second can have 2-3 more (D,E,F?).  
No need for 9 categories -- only make it more confusing.

src/compile/data/calculate.ts
```diff
@@ -15,6 +18,32 @@ export class CalculateNode extends DataFlowNode {
...
+    // get all the encoding with sort fields from model
+    const nodes = model.reduceFieldDef((acc: CalculateNode[], fieldDef: ChannelDef<any>) => {
+      if (isScaleFieldDef(fieldDef) && isSortArray(fieldDef.sort)) {
+        const transform: CalculateTransform = {
+          calculate: CalculateNode.calculateExpressionFromSortField(fieldDef.field, fieldDef.sort),
+          as: `${fieldDef.field}_sort_index`
```
kanitw: Should channel be a part of the name? 
(e.g., `${channel}_${fieldDef.field}_sort_index`.)

Imagine a troll case where I have x and y encode the same field, but use different `sort` array.  Your current code will behave incorrectly.

src/compile/data/calculate.ts
kanitw: I also start wondering that this is quite a special purposed one, so I wonder if having a distinct type of node for this will help with optimization in longer term? 
(But since this is already working, maybe it's waste of time to change it now -- I'm just asking.) 

@domoritz  -- any thoughts?

src/compile/scale/domain.ts
```diff
@@ -266,6 +266,15 @@ export function domainSort(model: UnitModel, channel: ScaleChannel, scaleType: S
 
   const sort = model.sort(channel);
 
+  // if the sort is specified with array, use the created data
+  if (isSortArray(sort)) {
+    return {
+      op: 'min',
+      field: `${model.vgField(channel)}_sort_index`,
```
kanitw: See https://github.com/vega/vega-lite/pull/3423/files#diff-3cdb83c52621809d58d32cb4e52787ffR27

src/compile/scale/domain.ts
```diff
@@ -266,6 +266,15 @@ export function domainSort(model: UnitModel, channel: ScaleChannel, scaleType: S
 
   const sort = model.sort(channel);
 
+  // if the sort is specified with array, use the created data
```
kanitw: use the derived sort index field

_config.yml
```diff
@@ -49,7 +49,7 @@ kramdown:
   syntax_highlighter: rouge
   auto_id_stripping: true
 
-plugins:
+plugins_dir:
```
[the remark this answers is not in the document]
domoritz: Let's put this in a separate PR.

src/compile/data/calculate.ts
domoritz: I don't think it is that special and I can't think of a special optimization that we would do. Let's leave it for now.

src/compile/data/parse.ts
```diff
@@ -249,6 +249,11 @@ export function parseData(model: Model): DataComponent {
       tu.parent = head;
       head = tu;
     }
+
+    for (const calculate of CalculateNode.makeAllFromSort(model)) {
```
kanitw: Also update visual diagram on the top.

site/docs/encoding/sort.md
```diff
@@ -55,21 +55,22 @@ If the channel has a discrete scale (`band`, `point` or `ordinal`), the field's
 
 3) Unsorted – `null` – The field is not sorted. This is equivalent to specifying `sort: false` in [Vega's scales](https://vega.github.io/vega/docs/scales/#sort).
 
-4) Specify custom order by:
-  1) providing `sort` with an array that lists valid input domains
-  2) providing custom `scale`'s [`domain`](scale.html#domain)
+4) Specify custom order by providing custom `scale`'s [`domain`](scale.html#domain). (In this case, you don't need to use `sort` property.)
```
kanitw: As mentioned earlier, this shouldn't be a "suggested" way for sorting.   Plus, putting unpreferred way on top of preferred way is weird. 

Please just make this like a note that it can achieve similar effect but has some pitfall(s).    (The `sort` array can be 4. and this should become just a note.)

Actually, we should also mention about `sort` in scale's `domain` documentation too.

site/docs/encoding/sort.md
```diff
@@ -55,21 +55,22 @@ If the channel has a discrete scale (`band`, `point` or `ordinal`), the field's
...
 
-#### Example: Sorting Ordinal Scale by Another Field
+5) Sorting by preferred order of values by providing `sort` with array specifying preferred order of the values. Unspecified values will assume their original orders in the data source, preceded by the orders in the sort array.
```
kanitw: > Sorting by preferred order of values by providing `sort` with array specifying preferred order of the values.

"preferred order of values" showing up twice, quite redundant.

"providing `sort` as an  array specifying preferred order of the values." ?

src/fielddef.ts
```diff
@@ -169,8 +169,9 @@ export interface ScaleFieldDef<F> extends FieldDef<F> {
 
   /**
    * Sort order for the encoded field.
-   * Supported `sort` values include `"ascending"`, `"descending"`, `"array of field's domains"` and `null` (no sorting).
+   * Supported `sort` values include `"ascending"`, `"descending"`, `"array specifying the preferred order of values"` and `null` (no sorting).
```
kanitw: `"array specifying the preferred order of values"` suggests that this is an enum array, so it's a bit misleading. 

`"ascending"`, `"descending"`, `null`, or an array specifying the preferred order of values

src/fielddef.ts
```diff
@@ -169,8 +169,9 @@ export interface ScaleFieldDef<F> extends FieldDef<F> {
...
    * For fields with discrete domains, `sort` can also be a [sort field definition object](https://vega.github.io/vega-lite/docs/sort.html#sort-field).
+   * For `sort` as an "array specifying the preferred order of values", the sort order will obey the values in the array, followed by any unspecified values in their original order.
```
kanitw: Why quoting?

kanitw: Also add link to the section that you provide an example for sort array.

site/docs/encoding/sort.md
```diff
@@ -55,21 +55,22 @@ If the channel has a discrete scale (`band`, `point` or `ordinal`), the field's
...
+__Note__: This sorting approach is more preferable to specifying custom `scale`'s `domain` because it only requires the array of preferred order of values, while `scale`'s `domain` requires array containing full domain.
 
-The following example sorts x by mean of Horsepower.
+In the case that sort array contains every field value, the sort order will follow the specified values in the array.
+<div class="vl-example" data-name="bar_custom_sort_full"></div>
```
kanitw: line break before DIV?

site/docs/encoding/sort.md
```diff
@@ -55,21 +55,22 @@ If the channel has a discrete scale (`band`, `point` or `ordinal`), the field's
...
 
-<div class="vl-example" data-name="histogram_sort_mean"></div>
+Sort order will precede by the specified values in the array, even though some value is ignored. Note how `D`, `E` and `F` still follows their original order, while `B`, `A` and `C` follows the order in sort array.
+<div class="vl-example" data-name="bar_custom_sort_partial"></div>
```
kanitw: line break before DIV?

kanitw:
Oh and please make tests pass. :)

site/docs/encoding/sort.md
sirahd: I think we already have that in `domain` documentation (https://vega.github.io/vega-lite/docs/scale.html#domain)

kanitw: Yes, but it didn't say what's the different between domain array and sort array yet so briefly mention about it and link to the full example in sort page would be nice.

site/static/main.ts
```diff
@@ -72,7 +72,7 @@ function embedExample($target: any, spec: TopLevelExtendedSpec, actions=true, to
   });
 
   if (tooltip) {
-    vegaLite(view, spec);
```
kanitw: can you fix this in a separate PR that we can undo easily

sirahd: Yup

sirahd:
@kanitw I just rebased. It should be good to go!

examples/specs/bar_custom_sort_partial.vl.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "A simple bar chart with embedded data.",
+  "data": {
+    "values": [
+      {"a": "A","b": 28}, {"a": "B","b": 55}, {"a": "C","b": 43},
+      {"a": "D","b": 91}, {"a": "E","b": 81}, {"a": "F","b": 53}
```
kanitw: Actually, change this to Z, Y, X? 

I believe the result would be B, A, C, Z, Y, X.       
(The reason we should change is that using  B, A, C, D, E, F, it won't be clear whether it's D E F because of the original order or because D is naturally before E and F.)

src/compile/data/parse.ts
```diff
@@ -202,6 +204,10 @@ export function parseData(model: Model): DataComponent {
     }
 
     head = TimeUnitNode.makeFromEncoding(head, model) || head;
+
+    for (const calculate of CalculateNode.makeAllForSortIndex(head, model)) {
```
kanitw: Do we really still need for loop here?  Can't we just set head to the last item in the array?

src/compile/scale/domain.ts
```diff
@@ -267,6 +267,15 @@ export function domainSort(model: UnitModel, channel: ScaleChannel, scaleType: S
 
   const sort = model.sort(channel);
 
+  // if the sort is specified with array, use the derived sort index field
+  if (isSortArray(sort)) {
+    return {
+      op: 'min',
+      field: `${channel}_${model.vgField(channel)}_sort_index`,
```
kanitw: Here you use `model.vgField(channel)` but in the calculate.ts file you use `${channel}_{fieldDef.field}_sort_index`.  

For the most cases, this should be fine, but I wonder if there are cases where we apply aggregation function for categorical value -- in such case `model.vgField(channel)` would return different result from `fieldDef.field`.

I think we should have a helper method in `calculate.ts` for generating this field name (from a `fieldDef`?) and make both calculate.ts and the code here call the same method for safety in the future?

src/fielddef.ts
```diff
@@ -182,12 +182,13 @@ export interface ScaleFieldDef<F> extends FieldDef<F> {
 
   /**
    * Sort order for the encoded field.
-   * Supported `sort` values include `"ascending"`, `"descending"` and `null` (no sorting).
+   * Supported `sort` values include `"ascending"`, `"descending"`, `array specifying the preferred order of values` and `null` (no sorting).
```
kanitw: The `sort` property can be `"ascending"`, `"descending"`, `null`, or an array specifying the preferred order of values.

(Remove back ticks for the array case too since you have currently have inside backticks is not the exact string you gonna put in the JSON.)

site/docs/encoding/sort.md
```diff
@@ -55,7 +55,18 @@ If the channel has a discrete scale (`band`, `point` or `ordinal`), the field's
...
+{:#sort-array}
+
+4) Sorting by preferred order of values by providing `sort` an array of values that specify the order. Unspecified values will assume their original orders in the data source, preceded by the orders in the sort array.
+__Note__: It is also possible to sort by providing custom `scale`'s [`domain`](scale.html#domain). However, it is not preferable compared to using `sort` array since `domain` requires every possible value to be present in the array.
+
+In the case that sort array contains every field value, the sort order will follow the specified values in the array.
+
+<div class="vl-example" data-name="bar_custom_sort_full"></div>
+
+Sort order will precede by the specified values in the array, even though some value is ignored. Note how `D`, `E` and `F` still follows their original order, while `B`, `A` and `C` follows the order in sort array.
```
kanitw: Z, Y, X as suggested?

examples/specs/bar_custom_sort_partial.vl.json
kanitw: (Please double check to make sure I'm not crazy :p)

kanitw:
I also made a pass over the docs. :)

src/compile/data/parse.ts
sirahd: So it seems like the constructor and assemble code for `CalculateNode` doesn't handle the case where it has to generate multiple nodes from one instance. I could change it so that the constructor doesn't require for loop, but that would affect the code we already have for general `CalculateNode`

kanitw: I don't understand.    Why would the following code be different from the for loop you have here?

```
const sortIndexNodes = CalculateNode.makeAllForSortIndex(head, model);
head = sortIndexNodes[sortIndexNodes.length - 1];
```

Can you explain more if I overlook anything.

sirahd: I think what I'm trying to do might be different here. I saw that @domoritz made some changes to the way `bin` and `timeUnit` nodes are created and assembled. Previously we had to iterate over for loop to construct nodes returned from methods like `makeAll`, but the new changes made it so that `makeAll` returns only one node with all the components inside a dictionary with field name as a key(instead of multiple nodes with separate components). So the constructor of `bin` and `timeUnit` now takes the parent and the dictionary as params.

Anyway, I will follow your suggestion since I think it doesn't make sense for `CalculateNode` to have the same behavior as `bin` and `timeUnit` node as I explained above.

src/compile/scale/domain.ts
sirahd: Should we stick to `vgField` or just `fieldDef.field` as a field name?

kanitw: `vgField(fieldDef)` ?

src/compile/data/parse.ts
kanitw: > Previously we had to iterate over for loop to construct nodes returned from methods like makeAll, but the new changes made it so that makeAll returns only one node with all the components inside a dictionary with field name as a key(instead of multiple nodes with separate components). So the constructor of bin and timeUnit now takes the parent and the dictionary as params.

I think the old loop is updating head like you do down here in the past, but it's kinda annoying because you need to update head and keep passing current head to the next one.  

> returns only one node with all the components inside a dictionary with field name as a key

I think your case is more like geopoint/ geojson transform.  Which basically updating parent within the `parseAll` method.   So I think you could follow the pattern and do 

```
head = CalculateNode.parseAllSortIndex(head, model);
```

where your `parseAllSortIndex` returns just the final head (parent)

cc: @domoritz  Please correct me if I'm wrong.

sirahd:
@kanitw Hopefully this fixes the issue with for loop in node parsing. I move that iteration into how `CalculateNode` is constructed in the first place.

src/compile/data/calculate.ts
```diff
@@ -15,6 +19,30 @@ export class CalculateNode extends DataFlowNode {
     super(parent);
   }
 
+  public static parseAllForSortIndex(parent: DataFlowNode, model: ModelWithField) {
+    // get all the encoding with sort fields from model
+    model.forEachFieldDef((fieldDef: ScaleFieldDef<string>, channel: SingleDefChannel) => {
+      if (isScaleFieldDef(fieldDef) && isSortArray(fieldDef.sort)) {
+        const transform: CalculateTransform = {
+          calculate: CalculateNode.calculateExpressionFromSortField(fieldDef.field, fieldDef.sort),
+          as: sortArrayIndexField(model, channel)
```
kanitw: It's weird that you're still using `model` and `channel` given that this is already under `model.forEachFieldDef`.  


I'd say `sortArrayIndexField` should just accept a `fieldDef`

sirahd: If that's the case, how can we know what the channel the given `fieldDef` represents?

sirahd: I think we should still leave `channel` as function param

sirahd: Sorry for the long reply, since this function is also used in `domain.ts` to look up the sort field, given that only `model` and `channel` is provided in that code, I think accepting `model` and `channel` here is reasonable, even though it seems redundant in `model.forEachFieldDef`.

PR #3426, opened 2018-02-28
Add Invalid Encoding Channel Warning

### LINKED ISSUE #3415: Invalid encoding channel should fail more gracefully
For example, if I type `_y` instead of `y`, everything crashes.

```
{
  "mark": "bar",
  "encoding": {
    "x": {
      "bin": true,
      "field": "IMDB_Rating",
      "type": "quantitative"
    },
    "_y": {
      "aggregate": "count",
      "type": "quantitative"
    }
  }
}
```

image

invokesus:
Please:
- [X] Make your pull request atomic, fixing one issue at a time unless there are many relevant issues that cannot be decoupled.
- [x] Provide a test case & update the documentation under `site/docs/`
- [X] Make lint and test pass. (Run `yarn lint` and `yarn test`.  If your change affects Vega outputs of some examples, please run `yarn build:example EXAMPLE_NAME` to re-compile a specific example or `yarn build:examples` to re-compile all examples.)
- [X] Make sure you have rebased onto the `master` branch.
- [X] Provide a concise title so that we can just copy it to our release note.

Fixes #3415.

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -175,6 +175,10 @@ export namespace message {
     return `${channel} dropped as it is incompatible with "${markOrFacet}"${when ? ` when ${when}` : ''}.`;
   }
 
+  export function invalidEncodingChannel(channel: string) {
+    return `${channel}-encoding is dropped as ${channel} is not a valid encoding channel.`;
+  }
+
   export function facetChannelShouldBeDiscrete(channel: string) {
     return `${channel} encoding should be discrete (ordinal / nominal / binned).`;
   }
--- a/test/compile/unit.test.ts
+++ b/test/compile/unit.test.ts
@@ -19,6 +19,16 @@ describe('UnitModel', function() {
         assert.equal(localLogger.warns[0], log.message.incompatibleChannel(SHAPE, BAR));
       }));
 
+    it('should drop invalid channel and throws warning', log.wrap((localLogger) => {
+        const _model = parseUnitModel({
+          mark: 'bar',
+          encoding: {
+            _y: {type: 'quantitative'}
+          }
+        } as any); // To make parseUnitModel accept the model with invalid encoding channel
+        assert.equal(localLogger.warns[0], log.message.invalidEncodingChannel('_y'));
+      }));
+
     it('should drop channel without field and value and throws warning', log.wrap((localLogger) => {
         const model = parseUnitModel({
           mark: 'bar',
--- a/src/encoding.ts
+++ b/src/encoding.ts
@@ -1,6 +1,6 @@
 
 import {isArray} from 'vega-util';
-import {Channel, CHANNELS, supportMark} from './channel';
+import {Channel, CHANNELS, isChannel, supportMark} from './channel';
 import {FacetMapping} from './facet';
 import {
   ChannelDef,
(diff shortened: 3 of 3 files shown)


domoritz:
Thank you for the pull request.

Please make this a warning and delete the encoding instead of throwing an error. Also, let's add a test.

invokesus:
The warning and deletion of encoding is already implemented in [encoding.ts](https://github.com/vega/vega-lite/blob/ecc47de34e2603a87540bd2e70059225236bd4e6/src/encoding.ts#L146-L153).
I was thinking of adding a case for unrecognized Channel [here](https://github.com/vega/vega-lite/blob/ecc47de34e2603a87540bd2e70059225236bd4e6/src/channel.ts#L216). So that, `getSupportedMark` returns an empty dictionary, causing `supportMark` to return false and finally the warning being triggered in `normalizeEncoding`. 
Is this in the right direction?

domoritz:
I'd trace what causes the error in #3415 and make sure that it's handled either by the code you found or something new you write.

invokesus:
Updated the PR. 🙂

src/channel.ts
```diff
@@ -214,6 +214,8 @@ export function getSupportedMark(channel: Channel): SupportedMark {
       return {point: true, geoshape: true};
     case TEXT:
       return {text: true};
+    default:
```
kanitw: I wonder if this is the right place to correct this?  

If the channel you have isn't a valid `Channel` at all, then it doesn't satisfy the signature of this method. 

A better check would be using `isChannel` method to check if a key of the encoding object  is a channel.

test/compile/unit.test.ts
```diff
@@ -19,6 +19,18 @@ describe('UnitModel', function() {
...
 
+    it('should drop invalid channel and throws warning', log.wrap((localLogger) => {
+        const INVALID_CHANNEL = '_y';
+        const model = parseUnitModel({
+          mark: 'bar',
+          encoding: {
+            _y: {type: 'quantitative'}
+          }
+        } as any); // To make parseUnitModel accept the model with invalid encoding channel
+        assert.equal(model.encoding.shape, undefined);
```
kanitw: I'm not sure why this is related to `model.encoding.shape`.

invokesus: Removed the assert.

src/channel.ts
invokesus: Made the changes. Thanks for the help!

src/log.ts
```diff
@@ -175,6 +175,10 @@ export namespace message {
     return `${channel} dropped as it is incompatible with "${markOrFacet}"${when ? ` when ${when}` : ''}.`;
   }
 
+  export function invalidEncodingChannel(encoding: string) {
```
kanitw: `encoding` normally refer to encoding mapping, which is an object.

Rename to `channel` would be less misleading.

src/log.ts
```diff
@@ -175,6 +175,10 @@ export namespace message {
...
+    return `${encoding} dropped as it is invalid.`;
```
kanitw: `${channel}-encoding is dropped as ${channel} is not a valid encoding channel.`

test/compile/unit.test.ts
```diff
@@ -19,6 +19,17 @@ describe('UnitModel', function() {
         assert.equal(localLogger.warns[0], log.message.incompatibleChannel(SHAPE, BAR));
       }));
 
+    it('should drop invalid channel and throws warning', log.wrap((localLogger) => {
+        const INVALID_CHANNEL = '_y';
```
kanitw: remove this line and just use `"_y"` below?

src/log.ts
invokesus: Done.

domoritz:
Thank you @invokesus!

PR #3434, opened 2018-03-02
Re-exports from utils removed

### LINKED ISSUE #3433: Remove re-exports from utils
It's weird to export vega-utils functions from utils. We should just import form vega-utils directly.

ydlamba:
**Re-exports from utils removed**

- The functions re-exporting from utils are removed.

- The inconsistency of using same functions from utils or vega-utils fixed.

Fixes #3433.


domoritz:
Thank you!

PR #3438, opened 2018-03-03
Add multi-line tooltip example

jakevdp:


domoritz:
That's pretty cool. Super useful and actually fairly easy to implement. I think this is one example that we could put into the main example gallery.

jakevdp:
It's even easier to implement using Altair :smile: https://github.com/altair-viz/altair/blob/master/altair/vegalite/v2/examples/multiline_tooltip.py

domoritz:
Would be cool to have in pdvega :-)

kanitw:
This is very cool!  I wonder if it would be a bit nicer with `filled: true`?

PR #3458, opened 2018-03-04
[Fix] Make sure `mark.size` are correctly redirected for bar, text, and tick

kanitw:


kanitw:
Please manually merge to master and push to avoid unnecessary rebases in dependent branches.

src/compile/mark/mixins.ts
```diff
@@ -34,19 +34,21 @@ export function color(model: UnitModel) {
   return e;
 }
 
-export function baseEncodeEntry(model: UnitModel, ignoreOrient: boolean) {
+export type Ignore = Record<'size' | 'orient', 'ignore' | 'include'>;
```
domoritz: Didn't know about this one.

src/compile/mark/mixins.ts
```diff
@@ -121,7 +123,7 @@ export function bandPosition(fieldDef: FieldDef<string>, channel: 'x'|'y', model
   const scaleName = model.scaleName(channel);
   const sizeChannel = channel === 'x' ? 'width' : 'height';
 
-  if (model.encoding.size) {
+  if (model.encoding.size || model.markDef.size !== undefined) {
```
domoritz: Why are you just checking for `model.encoding.size` and not `model.encoding.size !== undefined`?

test/compile/mark/tick.test.ts
```diff
@@ -161,7 +177,7 @@ describe('Mark: Tick', function() {
       'data': {'url': 'data/cars.json'},
     });
     const props = tick.encodeEntry(model);
-    it('height should change with size field', function() {
+    it('maps height ot size', function() {
```
domoritz: ot -> not

src/compile/mark/mixins.ts
kanitw: `markDef.size` is number so `0` means something

`encoding.size` can only be an object.

test/compile/mark/tick.test.ts
kanitw: lol it's actually "to"

src/compile/mark/mixins.ts
domoritz: I see. I'd be okay with checking against undefined in both cases.

PR #3454, opened 2018-03-04
Refactor Composite Mark Infrastructure

kanitw:
Fix https://github.com/vega/vega-lite/issues/3398 and address part of https://github.com/vega/vega-lite/issues/3443

@starry97 @chanwutk  -- I'm refactoring this so it becomes more what obvious what to do in your PRs.

cc: @mattwchun

PR #3459, opened 2018-03-04
Rename  Boxplot's `mid` to `median`

kanitw:
"Mid" can't be a noun but "center" can so I like center better.  

~~We should merge https://github.com/vega/vega-lite/pull/3454 first.~~


kanitw:
Thinking a bit more, we can just call the mid tick "median" tick.

PR #3465, opened 2018-03-05
Use `Partial<Record<P, MarkConfig>>` to define PartMixins

kanitw:
(Currently failing as json-schema-generator doesn't work with `Partial<Record<P, MarkConfig>>`) 

cc: @domoritz  -- this is not urgent at all, but im posting so you know there is a bug

### DIFF
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -26,11 +26,7 @@ const BOXPLOT_PART_INDEX: Flag<BoxPlotPart> = {
 
 export const BOXPLOT_PARTS = keys(BOXPLOT_PART_INDEX);
 
-// TODO: Currently can't use `PartsMixins<BoxPlotPart>`
-// as the schema generator will fail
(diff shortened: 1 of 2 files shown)


domoritz:
If you have time, could you file a big over at https://github.com/vega/ts-json-schema-generator?

domoritz:
Depends on https://github.com/vega/ts-json-schema-generator/issues/24

domoritz:
Try again after https://github.com/vega/vega-lite/pull/3502 has been merged.

build/vega-lite-schema.json
```diff
@@ -5379,6 +5379,7 @@
     },
     "PartsMixins<BoxPlotPart>": {
       "additionalProperties": false,
+      "description": "Make all properties in T optional",
```
kanitw: lol this also get the description from TS interface

kanitw:
This is no longer a problem now.  I don't know if `Partial<Record<..., ...>>` is better than existing code though.

PR #3473, opened 2018-03-06
Add stderr to Aggregate

chanwutk:
`stderr` is added into `aggregate`.

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -21,6 +21,7 @@
         "q3",
         "ci0",
         "ci1",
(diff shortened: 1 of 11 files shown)


domoritz:
Can you instead replace the aggregate type by the vega type?

chanwutk:
What do you mean by replacing the aggregate type by the vega type?

domoritz:
https://github.com/vega/vega-typings/blob/3a127100a350cedfabf47e7e093123fa8a977910/types/spec/transform.d.ts#L65

PR #3486, opened 2018-03-07
use as any to suppress warning

sirahd:


kanitw:
Actually, I wonder if we can release a minor patch to fix the API instead of using any to cheat?

kanitw:
(Sorry, didn't think about that at first.)

domoritz:
I suggested this to not block development in VL. But this should be fixed properly in tooltip.

sirahd:
As of now, I'm not sure what's the root cause of this. The type definitions in tooltip are not modified recently, so I'm not sure if the error occurs from recent change in `vega-lite`. 

@domoritz Do you know what cause this error?

kanitw:
Perhaps, update Vega-Lite tooltip to use latest version of Vega-Lite and release?

domoritz:
I mainly did this: https://github.com/vega/vega-lite/commit/8781e08be3e43f3942d77ae5919bf0c58a005bb5

kanitw:
> Perhaps, update Vega-Lite tooltip to use latest version of Vega-Lite and release?

But this is easy, so let's see if it works?  if it doesn't really work then I'm fine with this.

kanitw:
It doesn't look like this is really blocking anything and we shouldn't really do this so I'll close this.

PR #3490, opened 2018-03-08
Add seattle weather example

jakevdp:


jakevdp:
OK, I made the bar charts a multi selection that filters points in the top panel. Might take a few minutes to percolate through to the block.

domoritz:
The default spacing in blocks isn't great. Can you make the upper chart smaller? 

screen shot 2018-03-08 at 11 51 49

domoritz:
This would be a cool example for the weather tutorial. Right now, none of our tutorials use selections.

jakevdp:
Made the upper chart smaller. Thanks!

PR #3491, opened 2018-03-08
Some top level specs require data. Fixes #3489

### LINKED ISSUE #3489: Should TopLevel<FacetSpec> require data?
This is a valid spec according to the schema, but it results in no plot because the data is not defined at the top level:

```
{
  "facet": {
    "column": {
      "field": "species",
      "type": "nominal"
    }
  },
  "spec": {
    "mark": "point",
    "data": {
      "url": "https://vega.github.io/vega-datasets/data/iris.json"
    },
    "encoding": {
      "color": {
        "field": "species",
        "type": "nominal"
      },
      "x": {
        "field": "petalLength",
        "type": "quantitative"
      },
      "y": {
        "field": "sepalLength",
        "type": "quantitative"
      }
    },
    "height": 150,
    "width": 150
  }
}
```

Should the schema be changed such that the ``"data"`` property is required within ``TopLevel<FacetSpec>``? I can't seem to create any valid spec that doesn't have a top-level data property.

domoritz:
Fixes #3489

### DIFF
--- a/scripts/rename-schema.sh
+++ b/scripts/rename-schema.sh
@@ -25,6 +25,3 @@ perl -pi -e s,'LogicalOperand<string>','SelectionOperand',g build/vega-lite-sche
 perl -pi -e s,'LogicalAnd<string>','SelectionAnd',g build/vega-lite-schema.json
 perl -pi -e s,'LogicalOr<string>','SelectionOr',g build/vega-lite-schema.json
 perl -pi -e s,'LogicalNot<string>','SelectionNot',g build/vega-lite-schema.json
-
-perl -pi -e s,'TopLevelFacetedUnitSpec','TopLevel<FacetedUnitSpec>',g build/vega-lite-schema.json
-perl -pi -e s,'TopLevelFacetSpec','TopLevel<FacetSpec>',g build/vega-lite-schema.json
--- a/scripts/rename-schema.sh
+++ b/scripts/rename-schema.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 perl -pi -e s,'<\(string\|RepeatRef\)>','',g build/vega-lite-schema.json
 perl -pi -e s,'<\(string\|RepeatRef\)\,','<',g build/vega-lite-schema.json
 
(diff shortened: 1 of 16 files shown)


jakevdp:
I'm not familiar enough with the typescript source to review this... but one question: would it be possible to somehow sort the keys in the schema JSON so that the diffs are more informative in the future?

domoritz:
I think we don't do that on purpose because we sometimes set the order on purpose. But @kanitw can correct me.

domoritz:
I added you as a reviewer so you can look at the schema. I don't expect you to review the typescript code.

jakevdp:
Ah, sounds good :smile: It's difficult from the diff here to see what this actually changes in the schema... could you give me the executive summary?

domoritz:
The summary is that top level unit specs and top level facet specs require data to be present.

kanitw:
> I think we don't do that on purpose because we sometimes set the order intentionally. But @kanitw can correct me.

We only set the order of examples manually as a lot of people would read the specs.  
However, for JSON schema, we mostly only read diff.  So stable stringify would be awesome!

src/spec.ts
```diff
@@ -59,6 +59,14 @@ export interface BaseSpec {
   transform?: Transform[];
 }
 
+export type DataRequired = {
```
kanitw: Make this `DataMixins` and re-use it elsewhere with `Partial<DataMixins>` to make sure that the documentation is always consistent if we update them in the future?

domoritz: That requires some non-trivial changes in the schema compiler. I will do that after the deadline.

domoritz:
Let's merge this as is for now. I will come up with a proper fix for the ts issue later.

domoritz:
Here is the clean version for later: https://github.com/vega/vega-lite/pull/3502

scripts/rename-schema.sh
```diff
@@ -25,3 +25,6 @@ perl -pi -e s,'LogicalOperand<string>','SelectionOperand',g build/vega-lite-sche
 perl -pi -e s,'LogicalAnd<string>','SelectionAnd',g build/vega-lite-schema.json
 perl -pi -e s,'LogicalOr<string>','SelectionOr',g build/vega-lite-schema.json
 perl -pi -e s,'LogicalNot<string>','SelectionNot',g build/vega-lite-schema.json
+
+perl -pi -e s,'TopLevelFacetedUnitSpec','TopLevel<FacetedUnitSpec>',g build/vega-lite-schema.json
+perl -pi -e s,'TopLevelFacetSpec','TopLevel<FacetSpec>',g build/vega-lite-schema.json
```
kanitw: Why do you have to do this?  This seems misleading.

domoritz: I didn't want to change the names. We can remove this if @jakevdp is okay with it.

domoritz: Done

domoritz:
We already used stable stringify but stable stringify doesn't sort arrays. It's actually not obvious how to sort arrays with objects.

kanitw:
> We already used stable stringify but stable stringify doesn't sort arrays. It's actually not obvious how to sort arrays with objects.

We shouldn't re-sort array.  


Another question.  If you want consistency, one idea would be to rename all `TopLevel<XXX>` to `TopLevelXXX` ?

domoritz:
I'm in favor of that renaming.

kanitw:
Thinking more -- this doesn't help fix all the problem though. 

For example, a concat of two facet spec is still valid without data at all.  It seems like we should provide some user-friendly compile time error.  

That said, I think we should still merge this.

domoritz:
This change only makes the code slightly safer. We won't be able to catch all errors in the schema (neither do we want to).

PR #3498, opened 2018-03-10
Replace `type: 'latitude/longitude'` with channels `latitude/longitude/latitude2/.../`

### LINKED ISSUE #3351: Nested fields in Geo
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.1.json",
  "width": 500,
  "height": 500,
  "layer": [
    {
      "data": {
        "url": "https://unpkg.com/world-atlas@1/world/110m.json",
        "format": {
          "type": "topojson",
          "feature": "countries"
        }
      },
      "projection": {
        "type": "orthographic"
      },
      "mark": {
        "type": "geoshape",
        "fill": "mintcream",
        "stroke": "black",
        "strokeWidth": 0.35
      }
    },
    {
      "data": {
        "url": "https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_week.geojson",
        "format": {
          "type": "json",
          "property": "features"
        }
      },
      "projection": {
        "type": "orthographic"
      },
      "mark": {
        "type": "circle",
        "opacity": 0.25,
        "fill": "red"
      },
      "encoding": {
        "x": {"field": "geometry.coordinates[0]", "type": "longitude"},
        "y": {"field": "geometry.coordinates[1]", "type": "latitude"},
        "size": {"field": "properties.mag", "type": "quantitative"}
      }
    }
  ]
}
```

### LINKED ISSUE #3353: Better way to specify latitude/longtitude mapping?
@jwoLondon has pointed out that an geographic encoding of the form:

```
"projection": ...
“encoding" : {
  "x": { "field": "lng", "type": “longitude" },
  "y": { "field": "lat", "type": “latitude" }
} 
```

is slightly misleading in implying two independent mappings while in fact both x and y are functions of longtitude and latitude values.  

Note that the design was in part motivated by the fact that for a basic projection like mercator, x predominantly represents longtitude and y predominantly represents latitude so this should be easy for most people to understand. (Tableau's UI for similar uses the same approach to represent mapping between lat/long to "row"/"column" shelves.) 

Another related note is that distinguishing between longitude and latitude types represents a level of
semantic detail that is at odds with quantitative , ordinal etc.

## Possible Proposal

If we want to make this strictly accurate for all projection (which I personally think we should), one possible change is making latitude and longitude (and latitude2/longitude2 for ranged marks) encoding channels that are only used for projections.  Given a field lng and lat in a data table, you can do:

```
"projection": ...,
“encoding" : {
  "longitude": {"field": "lng", "type": “quantitative" },
  "latitude": {"field": "lat", "type": “quantitative" }
} 
```

Thus, this should read naturally as  'encode the field “lat” as “latitude"' instead of 'encode the field “lat” as “y”' in the current approach we have.  Note that both fields are simply quantitative types -- no longer need special lat/long type anymore.   I think this proposal will actually address @jwoLondon's another concern that type `latitude` and `longitude` are not necessary too. 

However, one drawback of this approach that other members in the team are concerning with is that lat and long are not direct properties of the marks so it might be hard for users to understand.

I personally don't think this is an issue for a few reasons.
1. In my interpretation, encoding channel are target properties that users can map data fields (or constant values) to.  Most of the time, they should be mark properties but they don't have to.  We actually have a precedence for this like the "detail" channel.  "lat" and "long" could be target properties for encoding marks with projections.
2. I think it actually makes it easier to teach people that having projection means that we have input locations in lat/long.  The projection then converts lat/long into output x, y positions on the visualization.   This is a very critical concept to teach people when we explain about projection. 

However, we don't have a consensus that this proposal is a clear win over existing approach, especially regarding the concern that "lat" and "long" are not direct properties of the marks" so I would post it here for further discussion that we can revisit later.  

Should we reach agreement for this approach (or other better approach) in the future, 
we will introduce the new way as a new "preferred" approach while maintaining backward compatibility support for the approach introduced in 2.1.  So this won't be a breaking change.

kanitw:
Fix #3353   Fix #3351

Please merge https://github.com/vega/vega-lite/pull/3503 first

Note that I first remove the type so that I correctly replace the logic everywhere else, then I bring them back with a normalize for backward compatibility.  We should be able to revert https://github.com/vega/vega-lite/pull/3498/commits/ba8dde7cd8cccbef705d76d7d0d29efd8f69a395 in 3.0.   (So please _manually merge this PR_ so we can easily revert exactly the same hash in the future.) 

TODO:
- [x] Update Documentation 

cc: @jwoLondon -- if you have anything you want to add to the doc, feel free to comment / submit a PR.


src/compile/scale/parse.ts
```diff
@@ -52,9 +52,11 @@ function parseUnitScaleCore(model: UnitModel): ScaleComponentIndex {
 
     const channelDef = encoding[channel];
 
-    // If mark has a projection (potentially implicitly), there is no need to generate a scale.
-    if (isFieldDef(channelDef) && ((mark === GEOSHAPE && isFieldDef(channelDef) && channel === SHAPE && channelDef.type === GEOJSON)
-    || (contains([X, Y, X2, Y2], channel) && contains([LATITUDE, LONGITUDE], channelDef.type)))) {
+    // Don't generate scale for shape of geoshape
+    if (
```
kanitw: The logic for scale here is way simpler now as we separate them.

src/encoding.ts
```diff
@@ -48,6 +49,27 @@ export interface Encoding<F> {
...
+   */
+  longitude?: FieldDef<F>;
+
+  /**
+   * Latitude position of geographically projected marks.
+   */
+  latitude?: FieldDef<F>;
+
+  /**
+   * Longitude-2 position for geographically projected  ranged  `"area"`, `"bar"`, `"rect"`, and  `"rule"`.
```
domoritz: `  ranged  ` => ` ranged `

domoritz: Remove space around ranged.

test/compile/mark/mixins.test.ts
```diff
@@ -49,4 +50,33 @@ describe('compile/mark/mixins', () => {
       assert.propertyVal(colorMixins.fill, 'value', "transparent");
     });
   });
+
+  describe('midPoint()', function () {
+    it('should return correctly for lat/lng', function () {
```
domoritz: should return correct mod point for lat/lng

src/fielddef.ts
```diff
@@ -557,6 +557,18 @@ export function channelCompatibility(fieldDef: FieldDef<Field>, channel: Channel
...
       return COMPATIBLE;
 
+    case 'longitude':
+    case 'longitude2':
+    case 'latitude':
+    case 'latitude2':
+      if (fieldDef.type !== QUANTITATIVE) {
+        return {
+          compatible: false,
+          warning: `Channel ${channel} should not be used with${fieldDef.type} field.`
```
domoritz: missing space after with

src/compile/projection/parse.ts
```diff
@@ -27,16 +27,16 @@ function parseUnitProjection(model: UnitModel): ProjectionComponent {
   const {specifiedProjection, markDef, config, encoding} = model;
 
   const isGeoShapeMark = markDef && markDef.type === GEOSHAPE;
-  const isGeoPointOrLineMark = encoding && [X, Y, X2, Y2].some(
+  const isGeoPointOrLineMark = encoding && GEOPOSITION_CHANNELS.some(
     (channel) => {
       const def = encoding[channel];
-      return isFieldDef(def) && contains([LATITUDE, LONGITUDE], def.type);
+      return isFieldDef(def);
```
domoritz: `isFieldDef(encoding[channel])`

PR #3502, opened 2018-03-10
Define data only once

domoritz:
Depends on https://github.com/vega/ts-json-schema-generator/issues/24

### DIFF
--- a/src/spec.ts
+++ b/src/spec.ts
@@ -33,7 +33,7 @@ export type TopLevel<S extends BaseSpec> = S & TopLevelProperties & {
   config?: Config;
 };
 
-export interface BaseSpec {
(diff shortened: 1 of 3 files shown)


domoritz:
No longer blocked

src/spec.ts
```diff
@@ -33,7 +33,7 @@ export type TopLevel<S extends BaseSpec> = S & TopLevelProperties & {
   config?: Config;
 };
 
-export interface BaseSpec {
+export interface BaseSpecWithoutData {
```
kanitw: BaseSpecWithoutData isn't used any where. 

Maybe better do  the following?
```
export type BaseSpec = Partial<DataMixins> & { 
  ...
}
```

Issue #1855, opened 2017-01-18
Add Stroke Dash Channel


kanitw:
Add Stroke Dash Channel

- [x] Add channel + update encoding docs
- [x] Add them to relevant marks 
- [ ] Add default range + update scale docs
- [ ] Add unit test
- [ ] Test if they really work in the editor

kanitw:
- [ ] Before we implement, need to determine good default `strokeDash` range first. 

@willium @mathisonian  Any suggestions? I'm sure you two have seen good conventions.

mathisonian:
I like `4, 3` or `5, 5`

kanitw:
Yeah, but this should be a range, so a collection of strokeDash.  :) 

Note: @mcorrell suggests us looking at photoshop.  It's not pretty though as we don't need this for 2.0.

kanitw:
Yeah, but this should be a range, so a collection of strokeDash. :)

Note: @mcorrell suggests us looking at photoshop. It's not pressing though as we don't need this for 2.0.

mcorrell:
Just as a point of reference (in addition to weird options), word has the following sort of horizontal rules:

<img width="163" alt="screen shot 2017-01-18 at 7 01 02 pm" src="https://cloud.githubusercontent.com/assets/8633790/22092180/9d640078-ddb0-11e6-88ed-baa803bec789.png">

Illustrator is a little odder in that it assumes that strokes can be specified by a series of gaps and dash widths:

<img width="243" alt="screen shot 2017-01-18 at 6 57 49 pm" src="https://cloud.githubusercontent.com/assets/8633790/22092181/9d646216-ddb0-11e6-8753-be202d83be9d.png">

Illustrator also will also give you the option of being faithful to the dash/gap scheme you specify, or allow it to futz around with the start and end points of dashes such that corners and lines end on dashes rather than spaces (which could make shapes harder to determine).

Different stroke weights and styles are a pretty standard way of encoding a categorical variable on a line chart for cases where you can't use color, so having a channel for it would be useful.

willium:
yeah. I believe what @mathisonian was specifying (or at least how I interpreted it was)
Dash: 4
gap: 3
I'd opt for perhaps even allowing that level of specificity within the stroke properties with sensible defaults.

g3o2:
[...]

PR #3503, opened 2018-03-11
Add fill & stroke channels

kanitw:
Part of #1855 

Note that if you specify fill/stroke, the default stroke/fill for mark isn't automatically removed.
I think this is a less confusing rule as adding fill/stroke encoding will only cause one change to the visualization.

That said, an alternative rule is to ignore mark.color, config.mark.color when fill/stroke channels are specified.  But it's also unclear what to do then users specify color channel with fill/stroke channel.  The current approach I take is simply including everything (defaults + specified encodings), and make fill/stroke have higher precedence.

### DIFF
--- a/site/docs/encoding.md
+++ b/site/docs/encoding.md
@@ -163,7 +163,7 @@ In addition, definitions of mark property channels can include the `condition` p
 
 Here are the list of mark property channels:
 
-{% include table.html props="color,opacity,shape,size" source="Encoding" %}
+{% include table.html props="color,fill,stroke,opacity,shape,size" source="Encoding" %}
 
 {:#mark-prop-field-def}
 ### Mark Property Field Definition
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1610,7 +1610,7 @@
               "$ref": "#/definitions/MarkPropValueDefWithCondition"
             }
           ],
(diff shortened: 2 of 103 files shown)


domoritz:
image

kanitw:
Did you build?

pasted_image_3_11_18__2_12_pm

image

domoritz:
`strokeWidth` doesn't seem to work

```
{
  "mark": {"type": "point", "strokeWidth": 0},
  "encoding": {
    "fill": {"field": "Origin", "type": "nominal"},
    "shape": {"field": "Origin", "type": "nominal"}
  }
}
```

image

domoritz:
Not a bug per se but something I wanted to point out. Stroke doesn't work well for areas because strokes of neighboring areas overlap.

image

kanitw:
> strokeWidth doesn't seem to work

File a separate bug?  It's not something caused by this PR.

domoritz:
Cool, then only docs are missing for this to get merged.

site/docs/encoding.md
```diff
@@ -163,7 +163,7 @@ In addition, definitions of mark property channels can include the `condition` p
 
 Here are the list of mark property channels:
 
-{% include table.html props="color,opacity,shape,size" source="Encoding" %}
+{% include table.html props="color,fill,stroke,opacity,shape,size" source="Encoding" %}
```
kanitw: What do you mean doc missing?

domoritz: Ahh, I just noticed that the note about setting both stroke and fill will show up there as well. 

👍

PR #3505, opened 2018-03-11
Re-design AxisComponent -- do not split between `main` and `grid` until assemble phase

### LINKED ISSUE #3034: Consolidate main and grid axis in the axis component
Currently, we separate main and grid axis in the axis component. 

However, this causes a number of issues. 

1. `grid: false` does not get merged correctly (#3020)
2. Layering mean value rule over histogram produce two axes - https://github.com/vega/vega-lite/issues/2963 (the root cause is also that `grid: false` doesn't merge correctly.) 

We should consolidate them in one axis component and split them during assemble axis instead.

kanitw:
Fix #3034

Let's merge https://github.com/vega/vega-lite/pull/3506 first


kanitw:
Although this PR reorders properties output is almost every file, there is online one SVG change and it's actually fixing an unnoticed bug.

Issue #3449, opened 2018-03-04
Selection does not work with nested layer


kanitw:
Error: Undefined data set name: "tooltip_store"

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "Credits to @jakevdp. The example is forked/modified from his block.",
  "data": {
    "url": "data/stocks.csv",
    "format": {
      "type": "csv"
    }
  },
  "width": 800,
  "height": 400,
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "color": {
          "type": "nominal",
          "field": "symbol"
        },
        "x": {
          "type": "temporal",
          "field": "date"
        },
        "y": {
          "type": "quantitative",
          "field": "price"
        }
      }
    },
    {
      "mark": "point",
      "encoding": {
        "color": {
          "type": "nominal",
          "field": "symbol"
        },
        "opacity": {
          "value": 0,
          "condition": {
            "selection": "tooltip",
            "value": 1
          }
        },
        "x": {
          "type": "temporal",
          "field": "date"
        },
        "y": {
          "type": "quantitative",
          "field": "price"
        }
      },
      "selection": {
        "tooltip": {
          "type": "single",
          "nearest": true,
          "on": "mouseover",
          "encodings": [
            "x"
          ],
          "empty": "none",
          "resolve": "global"
        }
      }
    },
    {
      "transform": [
        {
          "filter": {
            "selection": "tooltip"
          }
        }
      ],
      "layer": [
         {
          "mark": {
           "type": "rule",
           "color": "gray"
         },
         "encoding": {
           "x": {
             "type": "temporal",
             "field": "date"
           }
         }
        },
        {
         "mark": {
           "type": "text",
           "align": "left",
           "dx": 5,
           "dy": -5
         },
         "encoding": {
           "color": {
             "type": "nominal",
             "field": "symbol"
           },
           "text": {
             "type": "quantitative",
             "field": "price"
           },
           "x": {
             "type": "temporal",
             "field": "date"
           },
           "y": {
             "type": "quantitative",
             "field": "price"
           }
         }
        }
      ]
    }
  ]
}
```

kanitw:
I suspect that this is because we add nested layer after we implement selection.

jakevdp:
Ha, I just ran into this trying the same thing :smile:

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/3507

Issue #2177, opened 2017-04-16
Merge parse and identical transforms at forks + allow sorting x/y-scale for layered plots


domoritz:
Started in https://github.com/vega/vega-lite/tree/dom/parse-merge but it's still not correct.

kanitw:
- [ ] overlay_line_short.vl.json is one example that needs this

kanitw:
- [ ] the example in https://github.com/vega/vega-lite/issues/2964 also needs this

domoritz:
In #2964 the two nodes are actually equivalent. We should definitely handle that case as it is trivial.

g3o2:
Given that there a few issues dealing with sorting in layered plots, some of them closed, I am bit confused regarding the status: is sorting x/y scale for layered plots work in progress ?

domoritz:
@g3o2 I think it's working. Do you have an example where it doesn't work?

g3o2:
@domoritz  maybe I'm doing something wrong ...

```javascript
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"a": "A","b": 28},
      {"a": "B","b": 55},
      {"a": "C","b": 43}
    ]
  },
  "layer": [{
    "mark": "bar",
    "encoding": {
      "y": {
        "field": "a", "type": "ordinal", 
        "sort": {"field": "b", "op": "mean", "order": "descending" }
      },
      "x": {"field": "b", "type": "quantitative"}
    }
  }, {
    "mark": {
      "type": "text",
      "align": "left",
      "baseline": "middle",
      "dx": 3
    },
    "encoding": {
      "y": {
        "field": "a", "type": "ordinal", 
        "sort": {"field": "b", "op": "mean", "order": "descending" }
      },
      "x": {"field": "b", "type": "quantitative"},
      "text": {"field": "b", "type": "quantitative"}
    }
  }]
}
```

![image](https://user-images.githubusercontent.com/12123506/36909289-e54f4b1c-1e3d-11e8-806c-420435cf00a0.png)

domoritz:
[...]

PR #3507, opened 2018-03-11
Fix bug in nested layer selection.

kanitw:
(Error: Undefined data set name: "tooltip_store")

The root problem seems to be that assembleSelectionData for both BaseConcat and Layer doesn't use the array that's passed in

Fix https://github.com/vega/vega-lite/issues/3449


domoritz:
The fix looks reasonable to me. Why is this WIP?

domoritz:
works

```
{
  "data": {"url": "data/stocks.csv","format": {"type": "csv"}},
  "width": 800,
  "height": 400,
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "color": {"type": "nominal","field": "symbol"},
        "x": {"type": "temporal","field": "date"},
        "y": {"type": "quantitative","field": "price"}
      }
    },
    {
      "mark": "point",
      "encoding": {
        "color": {"type": "nominal","field": "symbol"},
        "opacity": {
          "condition": {"selection": "tooltip","value": 1},
          "value": 0
        },
        "x": {"type": "temporal","field": "date"},
        "y": {"type": "quantitative","field": "price"}
      },
      "selection": {
        "tooltip": {
          "type": "single",
          "nearest": true,
          "on": "mouseover",
          "encodings": ["x"],
          "empty": "none"
        }
      }
    },
    {
      "layer": [
        {
          "transform": [{"filter": {"selection": "tooltip"}}],
          "mark": {"type": "rule","color": "gray"},
          "encoding": {"x": {"type": "temporal","field": "date"}}
        },
        {
          "transform": [{"filter": {"selection": "tooltip"}}],
          "mark": {"type": "text","align": "left","dx": 5,"dy": -5},
          "encoding": {
            "text": {"type": "quantitative","field": "price"},
            "color": {"type": "nominal","field": "symbol"},
            "x": {"type": "temporal","field": "date"},
            "y": {"type": "quantitative","field": "price"}
          }
        }
      ]
    }
  ]
}
```

does not work

```
{
  "data": {"url": "data/stocks.csv","format": {"type": "csv"}},
  "width": 800,
  "height": 400,
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "color": {"type": "nominal","field": "symbol"},
        "x": {"type": "temporal","field": "date"},
        "y": {"type": "quantitative","field": "price"}
      }
    },
    {
      "mark": "point",
      "encoding": {
        "color": {"type": "nominal","field": "symbol"},
        "opacity": {
          "condition": {"selection": "tooltip","value": 1},
          "value": 0
        },
        "x": {"type": "temporal","field": "date"},
        "y": {"type": "quantitative","field": "price"}
      },
      "selection": {
        "tooltip": {
          "type": "single",
          "nearest": true,
          "on": "mouseover",
          "encodings": ["x"],
          "empty": "none"
        }
      }
    },
    {
      "transform": [{"filter": {"selection": "tooltip"}}],
      "layer": [
        {
          "mark": {"type": "rule","color": "gray"},
          "encoding": {"x": {"type": "temporal","field": "date"}}
        },
        {
          "mark": {"type": "text","align": "left","dx": 5,"dy": -5},
          "encoding": {
            "text": {"type": "quantitative","field": "price"},
            "color": {"type": "nominal","field": "symbol"},
            "x": {"type": "temporal","field": "date"},
            "y": {"type": "quantitative","field": "price"}
          }
        }
      ]
    }
  ]
}
```

But they should be equivalent.

domoritz:
Here is the problem. Selections are not working because the dates are not parsed before `(vlSingle(\"tooltip_store\", datum))`. I'm not sure how @arvind wanted this to work but I see three fixes:
1. The selection knows what data type it expects because of the encoding where it is defined. We then parse data before filtering in `vlSingle`.
2. We always push parse up to the root. That means it will never be possible to use the same field with different types. This would simplify a bunch of other things in our codebase but it is a major change (although not really breaking as we don't officially support using the same data with multiple types).
3. We push parse through single filters. This should fix this particular example but may not work for example if the encoding where the selection filter is applied does not imply a type parse.

Thoughts?

domoritz:
Ha, never mind. The third fix would have been easy but doesn't work because the parse node hits a fork before moving past the selection filter. https://github.com/vega/vega-lite/issues/2177 would fix this.

domoritz:
Here is an alternative that works

```
{
  "data": {"url": "data/stocks.csv","format": {"type": "csv"}},
  "width": 800,
  "height": 400,
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "color": {"type": "nominal","field": "symbol"},
        "x": {"type": "temporal","field": "date"},
        "y": {"type": "quantitative","field": "price"}
      }
    },
    {
      "mark": "point",
      "encoding": {
        "color": {"type": "nominal","field": "symbol"},
        "opacity": {
          "condition": {"selection": "tooltip","value": 1},
          "value": 0
        },
        "x": {"type": "temporal","field": "date"},
        "y": {"type": "quantitative","field": "price"}
      },
      "selection": {
        "tooltip": {
          "type": "single",
          "nearest": true,
          "on": "mouseover",
          "encodings": ["x"],
          "empty": "none"
        }
      }
    },
    {
      "transform": [{
        "calculate": "toDate(datum[\"date\"])", "as": "date"
      },{
        "filter": {"selection": "tooltip"}
      }],
      "layer": [
        {
          "mark": {"type": "rule","color": "gray"},
          "encoding": {"x": {"type": "temporal","field": "date"}}
        },
        {
          "mark": {"type": "text","align": "left","dx": 5,"dy": -5},
          "encoding": {
            "text": {"type": "quantitative","field": "price"},
            "color": {"type": "nominal","field": "symbol"},
            "x": {"type": "temporal","field": "date"},
            "y": {"type": "quantitative","field": "price"}
          }
        }
      ]
    }
  ]
}
```

domoritz:
Or even easier

```
  "data": {
    "url": "data/stocks.csv",
    "format": {
      "type": "csv",
      "parse": {"date": "date"}
    }
  },
```

Issue #1601, opened 2016-10-20
Encoding's `datum` for constant values in data domain that still pass through scales


kanitw:
`value` in fieldDef set the range value directly. 

If users wanna provide domain value that needs to be scaled, we need to add another property.
From a conversation long time ago, we decided that this property is `datum`. 

To clarify, a `value` property in VL produces a Vega ValueRef `{value: value}`. 
Meanwhile, a `datum` property in VL should produces a Vega Value Ref `{scale:..., value: datum}` and use scale from relevant layers.  

cc: @swojit

kanitw:
- [ ] Change Dropping condition in `unit.ts`

kanitw:
{
  "data": {"url": "data/stocks.csv"},
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "x": {"field": "date", "type": "temporal"},
        "y": {"field": "price", "type": "quantitative"},
        "color": {"field": "symbol", "type": "nominal"}
      }
    },
    {
      "mark": "rule",
      "encoding": {
        "y": { "datum": 100},
        "size": {"value": 2}
      }
    }
  ]
}

kanitw:
Since this is just a syntactic sugar and there is already a good workaround, I don't think we will get to do it soon. Just in case any one wanna help, this would involve:

1) Add a new `DatumDef` as a new type of ChannelDef
2) Modify scale creation logic in `parseUnitScaleCore` in `compile/scale/parse.ts` to also create scales for datum.  
3) Fix all the compile error due to new type of ChannelDef as DatumDef.   This should include
- rules for creating scales
- rules in `compile/mark/valueref.ts` and `compile/mark/mixins.ts` for how to apply datum with scale.  (This part should be quite straightforward)

4) The complicated part is deciding how we deal with axes and legends for datum.

eibanez:
@kanitw here is a related issue in the compiled vega spec.

I started with your example above, but added facetting:

```{
  "name": "test",
  "data": {"url": "data/stocks.csv"},
  "facet": {
    "row": {"field": "symbol", "type": "nominal"}
  },
  "spec": {
  "layer": [
    {
      "mark": "line",
      "encoding": {
        "x": {"field": "date", "type": "temporal"},
        "y": {"field": "price", "type": "quantitative"},
        "color": {"field": "symbol", "type": "nominal"}
      }
    },
    {
      "data": {"values": [{"ref": 100}]},
      "mark": "rule",
      "encoding": {
[...]

PR #3511, opened 2018-03-11
Support scale disabling

### LINKED ISSUE #2400: Allow disabling scale with `scale: null`
Basically, if values of a data field are already color names, we should be able to disable scale.  
(Possibly with `"scale": null`)

For example, we should be able to reproduce this spec:
```
{
  "data": {
    "values": [
      {"a": "red","b": 28}, {"a": "green","b": 55}, {"a": "blue","b": 43}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "a", "type": "nominal"},
    "color": {"field": "a", "type": "nominal", "scale": {"range": ["blue", "green", "red"]}}
  }
}
```

using
```
{
  "data": {
    "values": [
      {"a": "red","b": 28}, {"a": "green","b": 55}, {"a": "blue","b": 43}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "a", "type": "nominal"},
    "color": {"field": "a", "type": "nominal", "scale": null}
  }
}
```

kanitw:
Fix #2400


kanitw:
> I think another useful example to add is one with a rule at zero. This change makes it really easy to do.

For that one, I think you need datum #1601, not this one.

domoritz:
> For that one, I think you need datum #1601, not this one.

I see. Okay then.

kanitw:
(That one is "applying scale to constant value, not disabling scale for data field)

PR #3513, opened 2018-03-11
Pass parent to dataflow node constructor. 

### LINKED ISSUE #2197: Why isn't each type of Node constructor include parent?
It seems like we always know its parent when we build things.

Including parent in the constructor will make the code shorter and further avoid typo bugs.

domoritz:
Fixes #2197.

### DIFF
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -88,7 +88,7 @@ export class BinNode extends DataFlowNode {
     super(parent);
   }
 
-  public static makeBinFromEncoding(parent: DataFlowNode, model: ModelWithField) {
+  public static makeFromEncoding(parent: DataFlowNode, model: ModelWithField) {
     const bins = model.reduceFieldDef((binComponentIndex: Dict<BinComponent>, fieldDef, channel) => {
       if (fieldDef.bin) {
         const {key, binComponent} = createBinComponent(fieldDef, model);
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -81,14 +81,14 @@ export interface BinComponent {
 
 export class BinNode extends DataFlowNode {
   public clone() {
-    return new BinNode(duplicate(this.bins));
+    return new BinNode(null, duplicate(this.bins));
   }
 
-  constructor(private bins: Dict<BinComponent>) {
-    super();
+  constructor(parent: DataFlowNode, private bins: Dict<BinComponent>) {
+    super(parent);
   }
 
-  public static makeBinFromEncoding(model: ModelWithField) {
+  public static makeBinFromEncoding(parent: DataFlowNode, model: ModelWithField) {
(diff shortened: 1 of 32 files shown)


src/compile/data/parse.ts
```diff
@@ -179,129 +157,88 @@ Description of the dataflow (http://asciiflow.com/):
...
-    for (const geojson of GeoJSONNode.makeAll(model)) {
-      geojson.parent = head;
-      head = geojson;
-    }
-
-    for (const geopoint of GeoPointNode.makeAll(model)) {
-      geopoint.parent = head;
-      head = geopoint;
-    }
+    head = GeoJSONNode.makeAll(head, model);
```
kanitw: makeAll gives a feel like factory pattern, but I think it's better named `parseAll()` ? (More consistent with `parseTransformArray` above.)

src/compile/data/bin.ts
```diff
@@ -81,14 +81,14 @@ export interface BinComponent {
...
   }
 
-  constructor(private bins: Dict<BinComponent>) {
-    super();
+  constructor(parent: DataFlowNode, private bins: Dict<BinComponent>) {
+    super(parent);
   }
 
-  public static makeBinFromEncoding(model: ModelWithField) {
+  public static makeBinFromEncoding(parent: DataFlowNode, model: ModelWithField) {
```
kanitw: Rename to makeFromEncoding?

src/compile/data/parse.ts
```diff
@@ -179,129 +157,88 @@ Description of the dataflow (http://asciiflow.com/):
...
 
   if (model.transforms.length > 0) {
-    const {first, last} = parseTransformArray(model);
-    first.parent = head;
-    head = last;
+    head = parseTransformArray(head, model);
   }
 
-  const parse = ParseNode.make(model);
+  const parse = ParseNode.make(head, model);
```
kanitw:
```
head = ParseNode.make(head, model) || head;  
```

?

src/compile/data/stack.ts
```diff
@@ -56,16 +56,16 @@ export class StackNode extends DataFlowNode {
...
 
   public clone() {
-    return new StackNode(duplicate(this._stack));
+    return new StackNode(null, duplicate(this._stack));
   }
 
-  constructor(stack: StackComponent) {
-    super();
+  constructor(parent: DataFlowNode, stack: StackComponent) {
+    super(parent, );
```
kanitw: remove comma

src/compile/data/parse.ts
domoritz: Done

src/compile/data/bin.ts
domoritz: Done

src/compile/data/parse.ts
domoritz: Nope, we need parse below.

src/compile/data/stack.ts
domoritz: Done

PR #3525, opened 2018-03-12
Remove all field: '*' in tests

kanitw:


jakevdp:
Does this mean ``field:*`` is no longer supported? Or is it allowed but optional?

domoritz:
It became optional in 2.0. In fact, we ignore any field when you use a count aggregate.

PR #3539, opened 2018-03-14
Make zindex for axis work again

### LINKED ISSUE #3537: zindex appears broken as of 2.3.x
As of version 2.3.x, [`zindex`](https://vega.github.io/vega-lite/docs/axis.html#general) no longer works as expected. 

In the following example spec, it would be expected that the (purposefully) black grid lines would be in front of the bar mark. No matter whether one puts 0 or 1, nothing happens.

```javascript
{
  "data": {
    "values": [
    ]
  },
  "mark": "bar",
  "encoding": {
    "y": {"field": "b", "type": "quantitative", "axis":{"zindex":1}}
  },
  "config": {
    "axis": {
      "gridColor": "black"
    }
  }
}
```

image

kanitw:
Fix #3537

### DIFF
--- a/src/compile/axis/assemble.ts
+++ b/src/compile/axis/assemble.ts
@@ -21,7 +21,7 @@ export function assembleAxis(
     header: boolean // whether this is called via a header
   } = {header: false}
 ): VgAxis {
-  const {orient, scale, title, ...axis} = axisCmpt.combine();
+  const {orient, scale, title, zindex, ...axis} = axisCmpt.combine();
 
   // Remove properties that are not valid for this kind of axis
   keys(axis).forEach((key) => {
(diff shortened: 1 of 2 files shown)


src/compile/axis/assemble.ts
```diff
@@ -21,7 +21,7 @@ export function assembleAxis(
     header: boolean // whether this is called via a header
   } = {header: false}
 ): VgAxis {
-  const {orient, scale, title, ...axis} = axisCmpt.combine();
+  const {orient, scale, title, zindex, ...axis} = axisCmpt.combine();
```
domoritz: How do we ensure that we don't forget anything here in the future?

kanitw: If you look at the code below, that's probably the only thing left that's not specific to grid / main axis.

PR #3540, opened 2018-03-14
DOC: Update interactive_panzoom_vconcat_shared example

jakevdp:
Change the shared scale from ``y`` to ``x``, which makes the example better match the documentation text (see https://vega.github.io/vega-lite/docs/bind.html#scale-binding)


jakevdp:
I think that should do it... let's see what Travis says.

jakevdp:
BTW, I found that in a fresh clone I needed to first run ``yarn build:examples`` before I could run the command to build a single example.

arvind:
Thanks @jakevdp!

domoritz:
It's sufficient to run `yarn build` (`yarn build:examples` takes a few minutes unless you have a beefy machine). I would love to have a more robust build system but it's tricky to get right.

PR #3545, opened 2018-03-15
Fix color precedence for mark and legend

### LINKED ISSUE #3463: Stroke in legend even if mark doesn't use it
Is there a reason that we show the border?

https://vega.github.io/vega-lite/examples/circle_binned.html

screen shot 2018-03-04 at 19 59 08

### LINKED ISSUE #3543: Incorrect legend for combined fill, stroke and color encodings
Looking at the new fill and stroke encodings, I think there is an inconsistency in legend generation. If marks are encoded with both the `color` and `fill` channels, `fill` takes precedence (as indicated in the documentation). However, the legend shows both the color (which should be ignored) and fill encodings:

```javascript
{
  "width": 200,
  "height": 200,
  "data": {
    "values": [
      { "x": 10, "y": 1, "cat": "a" },
      { "x": 20, "y": 2, "cat": "b" },
      { "x": 30, "y": 3, "cat": "c" },
      { "x": 36, "y": 4, "cat": "d" }
    ]
  },
  "encoding": {
    "x":     { "field": "x", "type": "quantitative" },
    "y":     { "field": "y", "type": "quantitative" },
    "size":  { "value": 2000 },
    "color": { "field": "cat", "type": "nominal" },
    "fill":  { "field": "y", "type": "ordinal" }
  },
  "mark": { "type": "circle", "stroke": "black" }
}
```

color and fill legend

This also applies if `fill` is set to a constant, this time (correctly) not showing the fill legend at all, but incorrectly showing the color legend:

```javascript
{
  "width": 200,
  "height": 200,
  "data": {
    "values": [
      { "x": 10, "y": 1, "cat": "a" },
      { "x": 20, "y": 2, "cat": "b" },
      { "x": 30, "y": 3, "cat": "c" },
      { "x": 36, "y": 4, "cat": "d" }
    ]
  },
  "encoding": {
    "x":     { "field": "x", "type": "quantitative" },
    "y":     { "field": "y", "type": "quantitative" },
    "size":  { "value": 2000 },
    "color": { "field": "cat", "type": "nominal" },
    "fill":  { "value": "red" }
  },
  "mark": { "type": "circle", "stroke": "black" }
}
```

color and constant fill

As a workaround one can obviously remove the color encoding entirely, but I think it would be preferable to follow the (consistent) behaviour of the `stroke` rather than `fill` channel:

screen shot 2018-03-14 at 19 14 37

As a more minor inconsistency, when encoding the `stroke` channel the stroke is not applied to the legend symbols when it is a constant:

screen shot 2018-03-14 at 19 15 02

kanitw:
Fix #3543 

Fix #3463


domoritz:
We need a more robust way to handle size legends.

screen shot 2018-03-15 at 12 38 32

domoritz:
In other cases, the legend improves.

screen shot 2018-03-15 at 12 39 56

domoritz:
This one is also better than before

screen shot 2018-03-15 at 12 41 11

domoritz:
I don't think line legends need to change. 

screen shot 2018-03-15 at 12 41 39

domoritz:
What's up with this example?

screen shot 2018-03-15 at 12 42 24

kanitw:
The stacked_bar one is correct.  Unnecessary stroke is remove.  

The line color is currently using stroke by default -- the only change you see here is the one that's inconsistent.   (See https://vega.github.io/vega-lite/examples/line_color.html.)

Although I agree that fill looks better for this case, it requires a special complex casing that's beyond the scope of this PR. Moreover, I'm not sure if it's worth adding complexity to the codebase.  

The only case that's worth fixing is `selection_zoom_brush_wheel.vg.json`.

domoritz:
👍

PR #3554, opened 2018-03-18
[2.4] Add window transform 

kanitw:


kanitw:
This feature branch is ready.  

cc: @domoritz @arvind  @jheer 

Please make a final review of the window syntax when you have time (after VIS deadline is fine.)

kanitw:
Let's manually merge offline without rebasing as another feature branch is depending on this branch so we better avoid rebasing.

domoritz:
By show above you mean sort the keys? I think that makes sense but that's more important is to make things clear in the docs.

kanitw:
> By show above you mean sort the keys?

Yes. 

>  I think that makes sense but that's more important is to make things clear in the docs.

It is pretty clear at this point.  I also just make one more pass. 

Please manually merge this PR to master to avoid problems with many other dependent branches.

kanitw:
> By show above you mean sort the keys? I think that makes sense but that's more important is to make things clear in the docs.

Should be pretty clear at this point.

src/transform.ts
```diff
@@ -143,10 +211,30 @@ export interface LookupTransform {
...
+export interface WindowSortField {
+  /**
+   * The name of the field to sort.
+   */
+  field: string;
+
+  /**
+   * Whether to sort the field in ascending or descending order.
+   */
+  order?: 'ascending' | 'descending';
```
domoritz: Why not use VgComparatorOrder?

src/transform.ts
```diff
@@ -101,6 +100,75 @@ export interface AggregatedFieldDef {
...
+   'nth_value';
+
+export interface WindowFieldDef {
+  /**
+   * The window or aggregation operations to apply within a window, including `rank`, `lead`, `sum`, `average` or `count`. See the list of all supported operations [here](https://vega.github.io/vega-lite/docs/window.html#ops).
+   */
+  op: AggregateOp | WindowOnlyOp;
+
+  /**
+   *  Parameter values for the window functions. Parameter values can be omitted for operations that do not accept a parameter.
```
domoritz: remove double space

src/transform.ts
```diff
@@ -101,6 +100,75 @@ export interface AggregatedFieldDef {
...
+   */
+  param?: number;
+
+  /**
+   * The data field for which to compute the aggregate or window function. This can be omitted for functions that do not operate over a field such as `count`, `rank`, `dense_rank`.
+   */
+  field?: string;
+
+  /**
+   *  The output name for each field.
```
domoritz: DOuble space. Also, what does this comment mean? Please explain a bit more.

PR #3555, opened 2018-03-18
Fix broken link to site examples gallery at CONTRIBUTE.md

italo-batista:
Hi, this is my first PR in this repository. I read at CONTRIBUTE.md that for small fixes would be possible to just create the PR instead of creating the relate ISSUE before. If I have done something wrong, just tell me (:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ You can find [tasks with the "Help wanted" label in the issue tracker](https://g
 
 ### Help Create New Examples
 
-To submit a new example, fork our [example Block](https://bl.ocks.org/domoritz/455e1c7872c4b38a58b90df0c3d7b1b9) and send us a [pull request to add a link](https://github.com/vega/vega-lite/edit/master/site/examples/gallery.md) to it to our [example gallery](https://vega.github.io/vega-lite/examples/).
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -22,7 +22,7 @@ You can find [tasks with the "Help wanted" label in the issue tracker](https://g
 
 ### Help Create New Examples
 
-To submit a new example, fork our [example Block](https://bl.ocks.org/domoritz/455e1c7872c4b38a58b90df0c3d7b1b9) and send us a [pull request to add a link](https://github.com/vega/vega-lite/edit/master/site/examples/gallery.md) to it to our [example gallery](https://vega.github.io/vega-lite/examples/).
```
kanitw: Not sure why you need to change edit to blob?

italo-batista: Actually I noticed this only when I checked the commit diff. I chose to keep it because I wanted to hear from you before. Since the link opens to master branch, it would not be better to use the link to just points to the collaborator in what file she/he would add the link? I will make as you desire! (:

domoritz: https://github.com/vega/vega-lite/edit/master/site/examples/index.md get's you straight into edit mode so I prefer that. 

Thank you for sending the PR @italo-batista!

italo-batista:
I rebased previous commit to condense changes in only one commit.

kanitw:
Thank you :)

PR #3558, opened 2018-03-20
 Make bar starting point correct when a custom domain that include zero is specified

### LINKED ISSUE #3556: Throw warning for bar with zero = false or with domain that doesn't include zero or with log scale

### LINKED ISSUE #3557: Bar does not start from 0 if has negative values.
In this example, I am trying to use the window transform to show how each of these 4 movies is different from the average, by graphing currentValue - mean. Some of the values are negative. However all the graphs should start from 0 and go either up or down. Not start from -10 and go to the appropriate value.

```
{
    "description": "Bar graph showing how each film differs from the average rating for that year",
    "width": 300,
    "height": 50,
    "data": {
        "values": [
            { "name": "Movie 1", "Rating": 9, "Year": 2016 },
            { "name": "Movie 2", "Rating": 3, "Year": 2016 },
            { "name": "Movie 3", "Rating": 5, "Year": 2015 },
            { "name": "Movie 4", "Rating": 2, "Year": 2015 }
        ]
    },
    "layer": [{
        "transform": [{
                "window": [{
                    "op": "mean",
                    "field": "Rating",
                    "as": "AverageYearRating"
                }],
                "groupby": [
                    "Year"
                ],
                "frame": [null, null]
            },
            {
                "calculate": "datum.Rating - datum.AverageYearRating",
                "as": "RatingDelta"
            }
        ],
        "mark": { "type": "bar", "clip": true },
        "encoding": {
            "x": {
                "field": "name",
                "type": "ordinal"
            },
            "y": {
                "field": "RatingDelta",
                "type": "quantitative",
                "scale": { "domain": [-10, 10] },
                "axis": { "title": "Rating Delta" }
            }
        }
    }]
}
```

Image:
image

kanitw:
Fix #3557 and Fix #3556

### DIFF
--- a/src/compile/mark/valueref.ts
+++ b/src/compile/mark/valueref.ts
@@ -1,6 +1,7 @@
 /**
  * Utility files for producing Vega ValueRef for marks
  */
+import {isArray} from 'util';
 import {Channel, X, X2, Y, Y2} from '../../channel';
 import {Config} from '../../config';
 import {
@@ -16,7 +17,7 @@ import {
 import {hasDiscreteDomain, ScaleType} from '../../scale';
 import {StackProperties} from '../../stack';
 import {QUANTITATIVE} from '../../type';
-import {contains} from '../../util';
+import {contains, some} from '../../util';
 import {VgSignalRef, VgValueRef} from '../../vega.schema';
 import {binRequiresRange, formatSignalRef} from '../common';
 import {ScaleComponent} from '../scale/component';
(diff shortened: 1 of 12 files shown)


src/compile/mark/valueref.ts
```diff
@@ -180,71 +164,56 @@ export function mid(sizeRef: VgSignalRef): VgValueRef {
...
-      };
-    }
-  }
-  // Put the mark on the x-axis
-  return {value: 0};
-}
-
 /**
- * @returns {VgValueRef} base value if scale exists and return max value if scale does not exist
+ * Whether the scale definitely include zero in the domain
```
domoritz: include -> includes

PR #3561, opened 2018-03-20
Add lt, lte, gt, gte filters

### LINKED ISSUE #3452: lessThan / moreThan Filter
For timeUnit, we can't just use expression.   Using range with null values make spec not very readable. 

So I think we better support `lessThan` / `moreThan` filter.    
And perhaps, `lessThanOrEqual` / `moreThanOrEqual`.

invokesus:
Will fix #3452 .

### DIFF
--- a/site/docs/transform/filter.md
+++ b/site/docs/transform/filter.md
@@ -32,7 +32,7 @@ For an [expression](types.html#expression) string, each datum object can be refe
 
 ## Field Predicate
 
-For a filter predicate, a `field` must be provided along with one of the predicate properties: ([`equal`](#equal-predicate), [`less than`](#lt-predicate), [`less than or equals`](#lte-predicate), [`greater than`](#gt-predicate), [`greater than or equals`](#gte-predicate), [`range`](#range-predicate), or [`oneOf`](#one-of-predicate)). Values of these operators can be primitive types (string, number, boolean) or a [DateTime definition object](types.html#datetime) to describe time. In addition, `timeUnit` can be provided to further transform a temporal `field`.
+For a field predicate, a `field` must be provided along with one of the predicate properties: [`equal`](#equal-predicate), [`lt`](#lt-predicate) (less than), [`lte`](#lte-predicate) (less than or equal), [`gt`](#gt-predicate) (greater than), [`gte`](#gte-predicate)(greater than or equal), [`range`](#range-predicate), or [`oneOf`](#one-of-predicate). Values of these operators can be primitive types (string, number, boolean) or a [DateTime definition object](types.html#datetime) to describe time. In addition, `timeUnit` can be provided to further transform a temporal `field`.
 
 {:#equal-predicate}
 ### Field Equal Predicate
--- a/site/docs/transform/filter.md
+++ b/site/docs/transform/filter.md
@@ -76,7 +76,7 @@ For example, to check if the `Year` field's value is less than or equals to `"20
 {% include table.html props="gt" source="FieldGTPredicate" %}
 
 To check if the `state` field's value is greater than `"Arizona"` by string comparison, we can use the following filter:
-(Note: Standard Javascript string comparison is done, ie., "A" < "a")
+(Note: Standard Javascript string comparison is done, ie., "A" < "B", but "B" < "a")
 
 {: .suppress-error}
 ```json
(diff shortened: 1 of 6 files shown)


invokesus:
I keep getting a [`RangeError: Maximum call stack size exceeded`](https://travis-ci.org/vega/vega-lite/jobs/366041785#L543) when trying to build a new spec. The tests are passing but I haven't been able to build without an error.😕 I need some help in figuring out what I need to change.

domoritz:
Sounds like recursion without a base case. Can you see where the last call is?

kanitw:
@invokesus any update on this?

src/predicate.ts
```diff
@@ -56,6 +56,103 @@ export function isFieldEqualPredicate(predicate: any): predicate is FieldEqualPr
...
+   */
+  timeUnit?: TimeUnit;
+
+  /**
+   * Field to be filtered.
+   */
+  field: string;
+
+  /**
+   * The value that the field should be equal to.
```
kanitw: "less than"  -- similar fixes for all other added filters.

kanitw:
FYI, test still failing

kanitw:
- [ ] Please also add docs :)

invokesus:
Have exams in a couple of days 😓. Will add the docs, an example and change the command to build the spec accordingly by 3rd May.

kanitw:
No worries -- take your time!

src/predicate.ts
```diff
@@ -58,6 +58,104 @@ export function isFieldEqualPredicate(predicate: any): predicate is FieldEqualPr
   return predicate && !!predicate.field && predicate.equal !== undefined;
 }
 
+export interface FieldLTPredicate {
```
kanitw: Can you make all of these interfaces extend `FieldPredicateBase` so `timeUnit` and `field` don't have to be declared repeatedly?

site/docs/transform/filter.md
```diff
@@ -46,6 +46,54 @@ For example, to check if the `car_color` field's value is equal to `"red"`, we c
...
+{:#gt-predicate}
+### Field Greater Than Predicate
+
+{% include table.html props="field,gt,timeUnit" source="FieldGTPredicate" %}
+
+For example, to check if the `state` field's value is greater than `"Arizona"`, we can use the following filter:
+
+{: .suppress-error}
+```json
+{"filter": {"field": "state", "gt": "Arizona"}}
```
kanitw: Why would you compared a state field to Arizona?

site/docs/transform/filter.md
```diff
@@ -46,6 +46,54 @@ For example, to check if the `car_color` field's value is equal to `"red"`, we c
 {"filter": {"field": "car_color", "equal": "red"}}
 ```
 
+{:#lt-predicate}
+### Field Less Than Predicate
```
kanitw: @domoritz  I'm wondering if we should really add 4 sections for lt,lte,gt,gte.  To me, they are pretty much the same.  So adding 4 sections maybe a bit too much.  That said, may be it's fine.

domoritz: Let's have a single section.

kanitw: Ok.  We chatted and let's keep it this way.  There are many other things we should focus more on.

site/docs/transform/filter.md
kanitw: Alphabetical order?   If that's the case, better note that `"B" < "a"` so users are more careful.

site/docs/transform/filter.md
kanitw: Actually, let's name the section 

### Field Comparator Predicates

and simplify the content?

kanitw: (You can omit the tables and say that instead of `equal`, you can use `lt`, `gt`, ... for less than , greater than, ...)

kanitw:
- [ ] Can you also remove `{"filter": {"field": "x", "range": [null, 5]}}` from the "Field Range Predicate" section (now that we have proper gt, lt, gte, lte?

site/docs/transform/filter.md
invokesus: Should `equal` be a part of `Field Comparator Predicates`?

kanitw:
Merged into feature/filter (so I can checkout / make final pass easily.  Please see minor corrections that I do.

PR #3563, opened 2018-03-21
Update examples/index.md adding Waterfall Chart example

italo-batista:
Add [link](http://blockbuilder.org/italo-batista/5f93b3c9f87dffcf01bb489d90f60652) to a Waterfall chart in Community Examples list.


domoritz:
Thank you!

PR #3567, opened 2018-03-23
DOC: Change default nominal scheme reference

eitanlees:
I believe that there is a typo in the documentation.

Based on the defaults listed in the [vega-parser](https://github.com/vega/vega-parser/blob/master/src/config.js#L188) it looks like categorical scales use the `tableau10 ` color scheme by default. 

```
    // defaults for scale ranges
    range: {
      category: {
        scheme: 'tableau10'
      },
```


domoritz:
Absolutely! Thank you for the fix.

PR #3571, opened 2018-03-26
Window Transform Documentation and Examples

AkshatSh:
Implemented Window Transform for Vega Lite
Issue #2488

Continuing from #3199


examples/specs/window_mean_difference.vl.json
```diff
@@ -34,7 +34,8 @@
           "aggregate": "average",
           "field": "AverageRating",
           "type": "quantitative"
-        }
+        },
+        "y": {"field": "Title", "type": "ordinal"}
```
kanitw: Why adding this?

kanitw: Obviously this is making it wrong

image

site/docs/transform/window.md
```diff
@@ -0,0 +1,91 @@
...
+
+Another example is to show the best movies for the year they were released. Here best is defined by having a score that is 2.5 points higher than the average for the year it was released in.
+
+<div class="vl-example" data-name="window_transform_mean_difference_by_year"></div>
+
+Rather than filtering the above two examples we can also show a residual graph using the window transform.
+
+<div class="vl-example" data-name="window_transform_residual_graph"></div>
+
+To replace any of the examples from total to `mean`, the operation just needs to be changed from `sum` to `mean`.
```
kanitw: This sentence is appearing twice.

Also, I don't understand why this sentence is necessary here especially since residual graph already uses "mean". 

As mentioned earlier on slack, we better just remove it.

site/docs/transform/window.md
```diff
@@ -0,0 +1,91 @@
...
+Below are some common use cases for the window transform. Shown are the examples along side the Vega Lite spec.
+
+### Percent of Total
+
+<div class="vl-example" data-name="window_activities"></div>
+
+### Difference from Mean
+
+
+Another example is to show the best movies for the year they were released. Here best is defined by having a score that is 2.5 points higher than the average for the year it was released in.
```
kanitw: This is out of place.  This is the explanation for the second example, not the first example.

PR #3573, opened 2018-03-26
[WIP] Adding more Examples

AkshatSh:
Examples Added
* Top K
  * Should I add a slider to select the Top K, or since this is for window transform would it be better to leave it as is, so the focus is on the window transform are the spec is minimal?
* Running Average
  * Currently I show how the global average changes each year, would it make more sense to show how the average for each year changes? Right now the value at the point 2015, shows the average from the start of time to 2015, should it only be the average for the year 2015?

Note for running average: I added the domain for the y axis and changed the width, because originally the graph looked like
image

But I changed it to look like this so the changes are more readable
image


examples/specs/window_running_average.vl.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "A line graph showing how the average rating for a movie has changed over the years.",
+  "data": {"url": "data/movies.json"},
+  "width": 500,
+  "transform": [
+    {
+      "window": [{
```
kanitw: 1) Don't you have to sort the data before hand.  
2) Shouldn't you calculate year first so you group the data by year, not by individual time in seconds.

examples/specs/window_running_average.vl.json
```diff
@@ -0,0 +1,39 @@
...
+        "field": "IMDB_Rating",
+        "as": "AverageRating"
+      }],
+      "frame": [
+        null,
+        0
+      ]
+    },
+    {"filter": "datum.AverageRating != null"},
+    {"filter": "datum.AverageRating > 0"}
```
kanitw: Not sure what you need to check average rating > 0 here?

examples/specs/window_running_average.vl.json
```diff
@@ -0,0 +1,39 @@
...
+  ],
+  "layer": [
```
kanitw: Why using layer with only one layer?  

A better plot would be showing both raw data and moving average.

site/docs/transform/window.md
```diff
@@ -96,3 +96,15 @@ Rather than filtering the above two examples we can also calculate a residual by
...
+### Top K
+
+Here we use window transform to derive the total number of students along with the rank of the current student to determine the top K students and display their score.
+
+<div class="vl-example" data-name="window_top_k"></div>
+
+
+### Running Average
+
+Here we use window transform to visualize how the average movie rating has been changing over the years.
```
kanitw: Please update this based on new dataset.

kanitw:
> Should I add a slider to select the Top K, or since this is for window transform would it be better to leave it as is, so the focus is on the window transform are the spec is minimal?

We don't support a slider yet anyway.  

> Currently I show how the global average changes each year, would it make more sense to show how the average for each year changes? Right now the value at the point 2015, shows the average from the start of time to 2015, should it only be the average for the year 2015?

Yes -- apply `timeUnit` to derive a new field first.

examples/specs/window_running_average.vl.json
```diff
@@ -0,0 +1,42 @@
...
+    {"filter": "datum.Miles_per_Gallon !== null"},
+    {"timeUnit": "year", "field": "Year", "as": "year"},
+    {
+      "window": [{
+        "op": "mean",
+        "field": "Miles_per_Gallon",
+        "as": "Average_MPG"
+      }],
+      "frame": [null, null],
+      "groupby": ["year"]
```
kanitw: I think you have to sort by year without groupBy and make frame [-2, 2] to make it 5 year moving average. 

The current setup is simply annual average.

examples/specs/window_top_k.vl.json
```diff
@@ -0,0 +1,46 @@
...
+    ]
+  },
+  "transform": [
+  {
+    "window": [{
+      "op": "rank",
+      "as": "rank"
+    }],
+    "sort": [{ "field": "score", "order": "descending" }],
+    "groupby": [
```
kanitw: The result SVG seems correct, but I wonder why do we need `groupby` here?   

Isn't the ranking global ranking?  (Do I miss something?)

examples/specs/window_running_average.vl.json
AkshatSh: You're right I need to update this example because it is no longer a running average. However I cannot figure out why, but there are multiple values for each year if I use something in the future (i.e. `[-2,2]`), however if I use all the values before this year `[null, 0]` there are no problems.

So I updated the example to show a running average of the MPG of all the cars over the years. Which I believe is still relevant because you can see how out of all the cars on the road for a given year, the average tends to get higher.

examples/specs/window_top_k.vl.json
AkshatSh: Sorry this was a mistake, I removed the count definition, and forgot to remove this part. The SVG output did not change.

examples/specs/window_running_average.vl.json
kanitw: > However I cannot figure out why, but there are multiple values for each year if I use something in the future (i.e. [-2,2]), however if I use all the values before this year [null, 0] there are no problems.

I think we better figure out otherwise it might mean we don't support a very common use case.  (Feel free to ask in Vega channel for more info.)

kanitw: > So I updated the example to show a running average of the MPG of all the cars over the years. Which I believe is still relevant because you can see how out of all the cars on the road for a given year, the average tends to get higher.

This makes sense, but it's a different example.  So rename?

kanitw: "Accumulative average"?

examples/specs/window_top_k.vl.json
```diff
@@ -0,0 +1,43 @@
...
+    ]
+  },
+  "transform": [
+  {
+    "window": [{
+      "op": "rank",
+      "as": "rank"
+    }],
+    "sort": [{ "field": "score", "order": "descending" }],
+    "frame": [null, null]
```
kanitw: If `frame` doesn't affect `rank`, do we really have to specify `frame` here?

site/docs/transform/window.md
```diff
@@ -101,3 +101,16 @@ Rather than filtering the above two examples we can also calculate a residual by
...
 <div class="vl-example" data-name="window_rank"></div>
+
+### Top K
+
+Here we use window transform to derive the total number of students along with the rank of the current student to determine the top K students and display their score.
+
+<div class="vl-example" data-name="window_top_k"></div>
+
+
+### Running Average
```
kanitw: This is "Cumulative Running Average", not just random moving average, isn't it? 

Maybe a bit clearer?  

https://en.wikipedia.org/wiki/Moving_average#Cumulative_moving_average

PR #3581, opened 2018-03-29
Add window_rank example 

kanitw:


kanitw:
You can play one match at a time. Match 2 comes after match 1.  This is "over time" too.

kanitw:
I'll rename match to matchday, if that's less confusing.

PR #3583, opened 2018-03-29
Add top level ref back to spec. Fixes #3582

### LINKED ISSUE #3582: TopLevelSpec missing from schema
The [release notes](https://github.com/vega/vega-lite/releases/tag/v2.3.0) for v2.3.0 say that ``TopLevelExtendedSpec`` was renamed to ``TopLevelSpec`` in the JSON schema, but in fact it just seems to be gone, i.e. I don't see anything for ``TopLevelSpec`` in the JSON schema.

Which is a problem for [VegaLite.jl](https://github.com/fredo-dedup/VegaLite.jl) because that is the starting point for us :)

domoritz:


domoritz:
I missed in the other changes. The problem was that I refactored the schema generator and used the argument parser library incorrectly. Unfortunately, I don't have tests for that part of the schema library.

PR #3584, opened 2018-03-30
Window Markdown Updates

AkshatSh:
Updated the window mark down to include:

* Added `sort` and `ignorePeers` to the initial transform spec
* Updated documentation for frame to contain which operators are affected.


site/docs/transform/window.md
```diff
@@ -41,19 +46,24 @@ The window transform performs calculations over sorted groups of data objects. T
 
 ## Window Transform Definition
 
+Note: First, the tuples are partitioned according to the groupby fields. Each partition is then sorted. Finally, the window calculations are performed over the sorted partitions.
```
kanitw: Writing-wise, it's weird to have "Note" at the beginning of a section.

src/transform.ts
```diff
@@ -145,7 +145,7 @@ export interface WindowTransform {
...
+   * A frame specification as a two-element array indicating how the sliding window should proceed. The array entries should either be a number indicating the offset from the current data object, or null to indicate unbounded rows preceding or following the current data object. The default value is `[null, 0]`, indicating that the sliding window includes the current object and all preceding objects. The value `[-5, 5]` indicates that the window should include five objects preceding and five objects following the current object. Finally, `[null, null]` indicates that the window frame should always include all data objects. The only operators affected are the aggregation operations and the first_value, last_value, and nth_value window operations.
```
kanitw: > The only operators affected are the aggregation operations and the first_value, last_value, and nth_value window operations.

So other window ops are not affected by frame?   If so, state that explicitly?

site/docs/transform/window.md
AkshatSh: I agree Note is weird at the beginning, but I did want this to be highlighted at the start. I wasn't quite sure where to put this, so I put it at the very beginning now right after the description of the window transform and removed the "Note:".

Issue #3212, opened 2017-11-29
Design composite mark for error bars


alshain:
Vega-lite should provide first-class support for error bars.

[Currently](https://vega.github.io/vega-lite/examples/layer_error_bars_dev.html), I can force vega-lite to aggregate data in some way, produce new fields, further process these fields to produce even more new fields, and use *these* new fields to draw a line, which just happens to represent an error bar. This is rather imperative and laborious.

Is there any plan to support error-bars natively, or "declaratively"?

Why not provide the ability to produce an error-measure when aggregating the data?

One idea:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {"url": "data/barley.json"},

  "transform": [
    {
      "aggregate": [
        {"op": "mean", "field": "yield", "as": "mean", "error": "stdev"}, // here
      ],
      "groupby": ["variety"]
    },
  ],
  "layer": [{
    "mark": {"type": "point", "filled": true},
    "encoding": {
      "x": {
        "field": "mean", "type": "quantitative",
        "scale": {"zero": false},
        "axis": {"title": "Barley Yield"}
      },
      "y": {
        "field": "variety", "type": "ordinal"
      },
      "color": {"value": "black"}
    }
  }]
}
```

This could automatically calculate the standard deviation and and display the corresponding error bar.

I don't know how this feature should interact and compose with the rest of the grammar, but I think it may be worth exploring this.

I have found a couple issues (#269 #1401) that mention error bars, but still, there seems to be no such feature except for visual markers that can be used to render error bars and the `x2` functionality which allows to draw a line that represents an errorbar.

kanitw:
Good point.  Thanks for suggesting.

We have some plan for composite marks (basically, macro for composing multiple marks).
For example we have an experimental feature for box plot like in [this example](https://github.com/vega/vega-lite/blob/master/examples/specs/box-plot_tukey_2D_vertical.vl.json). 

We probably will provide a composite error-bar macro to make it slightly easier to produce error bar.  

Probably something like this:

```
{
  mark: {
    "type": "errorbar", 
    "center": "mean" | "median", 
    "extent": "stdev" | "stderr" | "ci",
    "ticks": true | false, // whether error bar contain end ticks -- not sure if ticks is the best name though
    "centerMark": "point" | "circle" | "square" | "bar" | "line"
}
```

That said, for now, like you describe, you can compose the layer manually.  Here is one example:
[...]

PR #3588, opened 2018-03-31
Implement Error bar

chanwutk:
Implement macro for error bar from the issue #3212


kanitw:
I'm changing merge target to `2.x-composite-mark` (see above).

chanwutk:
@kanitw I have done most of the features of `errorbar`, there are some issue that I still need to fix.
- setting `color` and `size` in the` errorbar` mark def still does not work.

kanitw:
> setting color and size in theerrorbar mark def still does not work.

I think we shouldn't have to support "size" since it will be redundant with `mark.whisker.size`?  

For color, it may make more sense as we want to apply the same color for all layers.  That should be pretty easy though as you can just manually specify it for each layer?

chanwutk:
I also want to talk to you about the properties `size` and `color` in the `errorbar` mark def. Would it be nice to have `size` property to set the size of both `ticks` and `whiskers`, so the user does not have to type the same size twice?

kanitw:
> Would it be nice to have size property to set the size of both ticks and whiskers, so the user does not have to type the same size twice?

But size for tick and whiskers should not be the same so this is not helpful anyway.
(If ticks have the same size as whiskers, they are equivalent to having no ticks.)

kanitw:
Can you exclude `.nyc_output`?

src/config.ts
```diff
@@ -254,6 +254,20 @@ export const defaultConfig: Config = {
...
   },
 
+  errorbar: {
+    center: 'mean',
+    extent: 'stderr',
+    whisker: true,
+    point: {
+      opacity: 1,
+      filled: true,
+      size: 20
```
kanitw: Do we really need different size from default point size?

test/spec.test.ts
```diff
@@ -77,114 +77,133 @@ describe('normalize()', function () {
     it('should produce correct layered specs for mean point and vertical error bar', () => {
       assert.deepEqual(normalize({
         "description": "A error bar plot showing mean, min, and max in the US population distribution of age groups in 2000.",
-        "data": {"url": "data/population.json"},
-        "transform": [{"filter": "datum.year == 2000"}],
+        "data": {
```
kanitw: Instead of removing errorbar.test.ts, better move these tests there.

src/compositemark/common.ts
```diff
@@ -26,7 +27,7 @@ function getMarkDefMixins<P extends PartsMixins<any>>(
   markDef: GenericMarkDef<any> & P, part: keyof P, compositeMarkConfig: P
 ) {
   const mark = markDef.type;
-  const partMarkDef: MarkConfig = markDef[part];
+  const partMarkDef: MarkConfig = isBoolean(markDef[part]) ? {} : <MarkConfig>markDef[part];
```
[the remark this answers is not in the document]
chanwutk: It has an error, if I don't have the `<MarkConfig>`.

chanwutk:
There is still an issue when setting opacity at `MarkDef`, all the parts of `errorbar`'s opacity are set except the `point` because `point` has their own default opacity.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+  /**
+   * The options for extent when the center is median. Available options include:
+   * - `"ci": Extend the rule to the confidence interval of the mean.
+   * - `"stderr": The size of rule are set to the value of standard error, extending from the center.
+   * - `"stdev": The size of rule are set to the value of standard deviation, extending from the center.
+   * - `"iqr": Extend the rule to the q1 and q3.
+   *
+   * __Default value:__ `"iqr"`.
+   */
+  medianExtent?: ErrorBarExtent;
```
kanitw: I don't know if we should really add this because it's not useful.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+
+  // TODO(https://github.com/vega/vega-lite/issues/3702): add selection support
+  if (selection) {
+    log.warn(log.message.selectionNotSupported('boxplot'));
+  }
+
+  const center: ErrorBarCenter = markDef.center || config.errorbar.center;
+  const extent: ErrorBarExtent = markDef.extent || ((center === 'mean') ? config.errorbar.meanExtent : config.errorbar.medianExtent);
+
+  // add warning check in the test in this format.
```
kanitw: Not sure why you have commented test code here. Please clean up

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+
+const markTypeOfErrorBarPart: { [key: string]: Mark; }= {
+  bar: 'bar',
+  line: 'line',
+  point: 'point',
+  ticks: 'tick',
+  rule: 'rule'
+};
+
+export function getMarkNameOfErrorBarPart(key: ErrorBarPart): Mark {
```
kanitw: why exporting this?

src/compositemark/common.ts
```diff
@@ -21,7 +21,7 @@ export function partLayerMixins<P extends PartsMixins<any>>(
   const {color, opacity} = markDef;
 
   const mark = markDef.type;
-  const partMarkDef: MarkConfig = markDef[part];
+  const partMarkDef: MarkConfig = isBoolean(markDef[part]) ? {} : <MarkConfig>markDef[part];
```
kanitw: This is incorrect as `false` would not disable the part.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
-  if (!encoding.x2 && !encoding.y2) {
-    throw new Error('Neither x2 or y2 provided');
+  const orient: Orient = errorBarOrient(spec);
+  const {transform, continuousAxisChannelDef, continuousAxis, groupby, encodingWithoutContinuousAxis} = errorBarParams(spec, orient, center, extent);
+
+  const {size: _s, ...encodingWithoutSizeAndContinuousAxis} = encodingWithoutContinuousAxis;
+
+  const {scale, axis} = continuousAxisChannelDef;
+
+  function errorBarSinglePointPartMarksSpec(partName: ErrorBarPart, startingPosition: string = center, endingPosition:string = undefined) {
```
kanitw: `errorBarSinglePointPartMarksSpec` is so verbose

`errorBarPartSpec`?

kanitw: `:string` => `: string`

src/compositemark/errorbar.ts
kanitw: Rename to `getErrorBarPartMark`, same for the constant.

src/compositemark/errorbar.ts
kanitw: Thinking more about it, nested function make it really hard to read / know which variables from the outer scope you need in this function.  Better extract this as a separate function?  

alternatively you can make a function factory outside this method:

```
function makeErrorPartFactory(... /*list all outer variables you need here*/) {
  return (partName: ErrorBarPart, startingPosition: string = center, endingPosition:string = undefined) => {
     ...
  }
}
```

then you can call

`const errorBarSinglePointPartMarksSpec = makeErrorPartFactory(...)`

then keep the rest of this method.  

This way it will be easier to read.

src/compositemark/common.ts
kanitw: Also, `...partMarkDef` below will apply `...true`, which is not correct too.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+    const markProperties: boolean | object = config.errorbar[partName];
```
kanitw: `partLayerMixins` already applies config (and also apply it before `opacity`). 

Remove these two constants and set

```
mark: getMarkNameOfErrorBarPart(partName)
```

below should solve your opacity problem.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+    continuousAxisChannelDef = continuousAxisWithoutAggregate;
+  }
+
+  return {
+    continuousAxisChannelDef,
+    continuousAxis
+  };
+}
+
+function isExtentCI(extent: ErrorBarExtent): extent is 'ci' {
```
kanitw: You shouldn't have to declare type guards for string literals.  Please remove both of this.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+  return extent === 'ci';
+}
+
+function isExtentIqr(extent: ErrorBarExtent): extent is 'iqr' {
+  return extent === 'iqr';
+}
+
+function errorBarSpecialExtents(extent: 'ci' | 'iqr', fieldName: string): AggregatedFieldDef[] {
```
kanitw: What make these "special"?  

I think you either have extent that need simple aggregate (ci, iqr) or extent that needs to calculate delta from center (stdev, stderr).  None of these are "special".

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+    {
+      op: center,
+      field: continuousFieldName,
+      as: center + '_' + continuousFieldName
+    }
+  ];
+
+  let postAggregateCalculates: CalculateTransform[] = [];
+
+  if (isExtentCI(extent)) {
```
kanitw: You don't need two separate conditions for CI and IQR.  

`if (extent === 'ci' || extent === 'iqr')`  is sufficient. 

You can also just remove the `errorBarSpecialExtents` method and move its logic here.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+        type: channelDef.type
+      };
+    } else {
+      // For value def, just copy
+      encodingWithoutContinuousAxis[channel] = encoding[channel];
+    }
+  });
+
+  return {
+    transform: [].concat(
```
kanitw:
```
transform: [
  ...bins,
  ...timeUnits,
  {aggregate, groupby},
  postAggregateCalculates
]
```

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+    // y is continuous but x is not
+    return 'vertical';
+  } else {
+    // Neither x nor y is continuous.
+    throw new Error('Need a valid continuous axis for errorbars');
+  }
+}
+
+
+function errorBarContinousAxis(spec: GenericUnitSpec<Encoding<string>, ErrorBar | ErrorBarDef>, orient: Orient) {
```
kanitw: This method is exactly the same as boxplot's (except the expected `aggregate`)? 

If so, extract this method to `common.ts` and reuse.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
+      ...errorBarSinglePointPartMarksSpec('line'),
+      ...errorBarSinglePointPartMarksSpec('ticks', 'lower_rule'),
+      ...errorBarSinglePointPartMarksSpec('ticks', 'upper_rule'),
+      ...errorBarSinglePointPartMarksSpec('rule', 'lower_rule', 'upper_rule'),
+      ...errorBarSinglePointPartMarksSpec('point')
+    ]
+  };
+}
+
+function errorBarOrient(spec: GenericUnitSpec<Encoding<Field>, ErrorBar | ErrorBarDef>): Orient {
```
kanitw: This method is exactly the same as boxplot's (except the expected `aggregate`)? 

If so, extract this method to `common.ts` and reuse.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
-          x: encoding.x,
-          y: encoding.y2,
-          ...encodingWithoutX_X2_Y_Y2
+        calculate: `datum.${center}_${continuousFieldName} - datum.extent_${continuousFieldName}`,
+        as: 'lower_rule_' + continuousFieldName
+    }];
+  }
+
+
+  const groupby: string[] = [];
```
kanitw: This part for producing `groupBy, bins, timeUnit`  seems exactly the same as boxplot?  Extract a method to `common.ts`.

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,395 @@
...
-  const {mark: _m, selection: _sel, projection: _p, encoding, ...outerSpec} = spec;
-  const {size: _s, ...encodingWithoutSize} = encoding;
-  const {x2: _x2, y2: _y2, ...encodingWithoutX2Y2} = encoding;
-  const {x: _x, y: _y, ...encodingWithoutX_X2_Y_Y2} = encodingWithoutX2Y2;
+  const {mark, encoding, selection, projection: _p, ...outerSpec} = spec;
+  const markDef: ErrorBarDef = isMarkDef(mark) ? mark : {type: mark};
+
+  // TODO(https://github.com/vega/vega-lite/issues/3702): add selection support
+  if (selection) {
+    log.warn(log.message.selectionNotSupported('boxplot'));
```
kanitw: boxplot => errorbar

src/compositemark/errorbar.ts
kanitw: I think it might be better to remove these configs and hardcode the default for now.

src/compositemark/common.ts
chanwutk: isn't that this line
```if (markDef[part] || (markDef[part] === undefined && compositeMarkConfig[part])) {```
has already handled the case?

my `markDef[part]` in errorbar also has to be `boolean` or `MarkConFig`.

Comment (kanitw) [reply src/compositemark/common.ts]: Ah my bad.

File: src/compositemark/errorbar.ts (hunk 16 above)
Comment (kanitw) [reply src/compositemark/errorbar.ts]: Yes. Sorry didn’t call catch that

File: .nyc_output/0f1dbb21a7b2b647e4fb060e81f675dd.json [hunk 18]
```diff
@@ -0,0 +1 @@
+{}
```
kanitw: Please exclude output from NYC.

src/compositemark/common.ts
```diff
@@ -21,35 +29,167 @@ export function partLayerMixins<P extends PartsMixins<any>>(
...
+    continuousAxisChannelDef = continuousAxisWithoutAggregate;
+  }
+
+  return {
+    continuousAxisChannelDef,
+    continuousAxis
+  };
+}
+
+export function compositeMarkOrient(spec: GenericUnitSpec<Encoding<Field>, CompositeMark | CompositeMarkDef>, compositeMarkName: 'errorbar' | 'boxplot'): Orient {
```
kanitw: compositeMarkName => compositeMark

kanitw: (same for all other methods)

test/compositemark/errorbar.test.ts
```diff
@@ -1,97 +1,634 @@
...
+import {isFieldDef} from '../../src/fielddef';
+import * as log from '../../src/log';
+import {isMarkDef} from '../../src/mark';
+import {isLayerSpec, isUnitSpec, normalize} from '../../src/spec';
+import {isAggregate} from '../../src/transform';
+import {some} from '../../src/util';
 import {defaultConfig} from '.././../src/config';
 
-
-describe("normalizeErrorBar", () => {
```
kanitw: Please do not drop this wrapping `describe`?

src/compositemark/errorbar.ts
```diff
@@ -1,42 +1,215 @@
...
+  const orient: Orient = compositeMarkOrient(spec, ERRORBAR);
+  const {transform, continuousAxisChannelDef, continuousAxis, encodingWithoutContinuousAxis} = errorBarParams(spec, orient, center, extent);
+
+  const {size: _s, ...encodingWithoutSizeAndContinuousAxis} = encodingWithoutContinuousAxis;
 
-  if (!encoding.x2 && !encoding.y2) {
-    throw new Error('Neither x2 or y2 provided');
+  const {scale, axis} = continuousAxisChannelDef;
+
+  function errorBarPartSpec(partName: ErrorBarPart, startingPosition: string = center, endingPosition: string = undefined) {
```
kanitw: This function is still not extracted as commented in https://github.com/vega/vega-lite/pull/3588#discussion_r186473899

PR #3590, opened 2018-04-01
Tukey Boxplot Outliers

mattwchun:
Adds outliers to boxplot so proper tukey boxplot can be created.

@kanitw I'm still debugging a bug where it fails because it thinks the mark is 'box-plot' even though it should be 'boxplot'. However, nowhere in the codebase is 'box-plot'.


src/compositemark/boxplot.ts
```diff
@@ -171,6 +173,30 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<string>, BoxPlot
       }
     ]
   };
+
+  if (!isMinMax) {
+    // add outliers
+    const upperBoxField: string = 'upper_box_' + continuousAxisChannelDef.field;
+    const lowerBoxField: string = 'lower_box_' + continuousAxisChannelDef.field;
+    result.layer.push({
```
kanitw: if you just push a layer, this will be incorrect as it takes the same aggregated data source. 

```
data: original data source
layer:
  layer 
     ...all other layers
  new unit spec for outlier
```

src/compositemark/boxplot.ts
```diff
@@ -332,11 +359,13 @@ function boxParams(spec: GenericUnitSpec<Encoding<string>, BoxPlot | BoxPlotDef>
...
   });
 
+  const window: AggregatedFieldDef[] = aggregate;
+
   return {
     transform: [].concat(
       bins,
       timeUnits,
-      [{aggregate, groupby}],
+      [isMinMax ? {aggregate, groupby} : {window, groupby, frame}],
```
kanitw: You can't just change aggregate to window as the core (non-outliers) layer will render to many data points instead.

kanitw:
>  I'm still debugging a bug where it fails because it thinks the mark is 'box-plot' even though it should be 'boxplot'. However, nowhere in the codebase is 'box-plot'.

Where does it fails?   In the editor?  If so, sounds like you just need to link them properly :)

src/compositemark/boxplot.ts
mattwchun: Do I need to have the same transform in every non-outlier mark unit spec to calculate (q1, q3, median, iqr, etc)? I also then would need to have a transform with window and filter to calculate q1 and q3 and filter out values for outliers point mark spec?

kanitw:
@mattwchun FYI, test is still failing.  (Just in case you don't see.)

src/compositemark/boxplot.ts
```diff
@@ -171,6 +173,44 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<string>, BoxPlot
...
+    layer: [
+      { // boxplot
+        transform,
+        layer: result.layer
+      }, {
+        transform: [
+          {
+            window: transform[0].aggregate.slice(0, 2),
+            frame: [null, null],
+            groupby: transform[0].groupby
```
kanitw: `transform[0]` is similarly incorrect here.

src/compositemark/boxplot.ts
```diff
@@ -171,6 +173,44 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<string>, BoxPlot
...
+          }, {
+            filter: `(datum.${continuousAxisChannelDef.field} < datum.lower_box_${continuousAxisChannelDef.field} - ${extent} * (datum.upper_box_${continuousAxisChannelDef.field} - datum.lower_box_${continuousAxisChannelDef.field})) || (datum.${continuousAxisChannelDef.field} > datum.upper_box_${continuousAxisChannelDef.field} + ${extent} * (datum.upper_box_${continuousAxisChannelDef.field} - datum.lower_box_${continuousAxisChannelDef.field}))`
+          }
+        ],
+        mark: {
+          type: 'point',
+          ...(sizeValue ? {size: sizeValue} : {}),
```
kanitw: Why applying size here?   Size of bar = width, size of point = area.  Applying the same number would only lead to problematic result.

src/compositemark/boxplot.ts
```diff
@@ -171,6 +173,44 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<string>, BoxPlot
...
+          ...getMarkDefMixins<BoxPlotPartsMixins>(markDef, 'outliers', config.boxplot)
+        },
+        encoding: {
+          [continuousAxis]: {
+            field: continuousAxisChannelDef.field,
+            type: continuousAxisChannelDef.type
+          },
+          ...encodingWithoutSizeColorAndContinuousAxis,
+          ...(size ? {size} : {})
```
kanitw: Why applying size here? Size of bar = width, size of point = area. Applying the same number would only lead to problematic result.

src/compositemark/boxplot.ts
```diff
@@ -276,29 +290,33 @@ function boxContinousAxis(spec: GenericUnitSpec<Encoding<string>, BoxPlot | BoxP
...
 }
 
-function boxParams(spec: GenericUnitSpec<Encoding<string>, BoxPlot | BoxPlotDef>, orient: Orient, extent: 'min-max' | number) {
-
-  const {continuousAxisChannelDef, continuousAxis} = boxContinousAxis(spec, orient);
-  const encoding = spec.encoding;
-
-  const isMinMax = !isNumber(extent);
-  const aggregate: AggregatedFieldDef[] = [
+function boxParamsQuartiles(continousAxisField: string): AggregatedFieldDef[] {
```
kanitw: continuous*

src/compositemark/boxplot.ts
```diff
@@ -276,29 +290,33 @@ function boxContinousAxis(spec: GenericUnitSpec<Encoding<string>, BoxPlot | BoxP
...
+}
+
+function boxParams(spec: GenericUnitSpec<Encoding<string>, BoxPlot | BoxPlotDef>, orient: Orient, extent: 'min-max' | number) {
+
+  const {continuousAxisChannelDef, continuousAxis} = boxContinousAxis(spec, orient);
+  const encoding = spec.encoding;
+
+  const isMinMax = !isNumber(extent);
+  const aggregate: AggregatedFieldDef[] = boxParamsQuartiles(continuousAxisChannelDef.field);
+  aggregate.push({
```
kanitw: better just concat

src/compositemark/boxplot.ts
```diff
@@ -100,119 +99,134 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<string>, BoxPlot
...
-          ...encodingWithoutSizeColorAndContinuousAxis,
-          ...(size ? {size} : {})
+          ...encodingWithoutSizeColorAndContinuousAxis
         }
       }
     ]
   };
 }
 
+function getGroupbyBoxParams(transform: Transform[]): string[] {
```
kanitw: This O(n) search is quite inefficient and hard to understand why it is correct.  

Better extract a method for boxAggregate.

src/compositemark/boxplot.ts
```diff
@@ -100,119 +99,134 @@ export function normalizeBoxPlot(spec: GenericUnitSpec<Encoding<string>, BoxPlot
...
   if (isMinMax) {
-    return result;
+    return {
+      ...outerSpec,
+      transform,
+      layer: boxLayer
+    };
   }
 
+  const lowerBoxStr: string = 'datum.lower_box_' + continuousAxisChannelDef.field;
```
kanitw: Change `Str` to `Expr`?   Since these are Vega expression.

PR #3599, opened 2018-04-07
Implement Error Area

chanwutk:
Like error bar, the `error area` is the area of error (`stdev`, `stderr`, or `ci`) when using mark `line`.
screen shot 2018-04-07 at 00 46 10


kanitw:
Let's name this `errorband`.    To save you time, let's focus on getting `errorbar` right first?

kanitw:
Note: without  x and x2  or y and y2 at all, should `errorband` use `rect` mark under the hood instead?  

https://vega.github.io/vega-lite/examples/layer_global_mean_dev.html

kanitw:
I'm gonna close this for now as you better reimplement it by forking from the other branch.

Issue #3532, opened 2018-03-12
(Macro) adding shape to line should automatically add another point layer 


kanitw:
This may be the way to replace overlay config #1804 -- that said -- to override overlay config we have to override this too.

kanitw:
The spec in https://github.com/vega/vega-lite/issues/3530#issuecomment-372383322 is kinda tedious. 

Ideally, 

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A scatterplot showing horsepower and miles per gallons for various cars.",
  "data": {"url": "data/cars.json"},
  "mark": {"type": "line", "color": "black"},
    "encoding": {
      "x": {"field": "Year","type": "temporal"},
      "y": { "aggregate":"mean", "field": "Miles_per_Gallon","type": "quantitative"},
      "detail": {"field": "Origin", "type": "nominal"}
    }
  }
}
```

Should produce 


![image](https://user-images.githubusercontent.com/111269/37297738-a0f378b2-25db-11e8-8f49-d7038c21b1fe.png)

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/3604/commits/79f4cd6d4336c129cdc1ee51357c119f14a15fe5

PR #3604, opened 2018-04-12
Redesign config for overlaying points on top of lines/points and lines on top of area.

### LINKED ISSUE #1804: Redesign overlay config

kanitw:
Fix #1804 

- `config.overlay.line = true` => `config.line.point = {...}` // mark properties object
- `config.overlay.area = "line"` => `config.area =  {line: {...}, ...}`
- `config.overlay.area = "linepoint"` => `config.area =  {line: {...}, point: {...}, ...}`

This way we can also customize the styles of the overlaying line and area.

Here are some example specs:

```
{
  "mark": "line",
  "encoding": {
  },
  "config": {
    "line": {"point": {"color": "red"}}
  }
}
```

```
{
  "mark": "area",
  "encoding": {
  },
  "config": {
    "area": {"point": {}, "line": {}, "opacity": 0.5}
  }
}
```

This could also be mark definition property:

```
{
  "mark": {"type": "area", "point": {}, "line": {}, "opacity": 0.5},
  "encoding": {
  }
}
```

## TODO:

- [x]  whether we want to allow these properties to be boolean? (`config.line.point = true`)
- [x] If there is shape for line, automatically add point and drop shape from line?


kanitw:
@domoritz -- there are a few questions above -- let me know if you have any comment.

g3o2:
Wouldn't a composite mark (e.g. `point-line`) approach be more transparent about its intention and features?

kanitw:
@g3o2  That's an idea worth considering.  

The number of marks will increase a bit though (`line-point`, `area-line`, `area-line-point`) and I prefer to keep the number of marks relatively small. 

That said, I think it really comes down to whether we want point/line overlay to be a part of "theming". 
I think some organization may always want to have line overlay on top of area / points on top of line).  Thus, if we want to support this to be part of theming, then it should at least be in config.  (But perhaps can be in mark definition too.)

g3o2:
Do you have any specific use case in mind for this way of overlaying?

It should certainly be useful for area and line marks in conjunction with tooltip, for the latter sort of requires a (most often transparent) shape overlay to function properly.

Other than the above use case, doesn't specifying the additional shape as a separate layer provide more control and wouldn't it be more explicit when using specs for communication purposes? That's why I tend to interpret this feature more as a layered or composite mark than part of theming.

kanitw:
> no to 3 (too much magic).

@domoritz  Can you be consistent about your opinion?  -- You liked it here https://github.com/vega/vega-lite/issues/3532#issuecomment-372386426. lol  

(Let me know if you still like/don't like it.) 


> Other than the above use case, doesn't specifying the additional shape as a separate layer provide more control 

That feature is still available.  This overlay thing is more like a shortcut.  

One common use case is considering tools that only support single view specification like Voyager.  (Voyager doesn't support `layer` so doing another separate layer isn't an option for the tool.  
We also don't want to flood the list of mark types with line-point, area-line, area-line-point and so on.)

kanitw:
Also, given that we currently don't use dash in mark names (`boxplot`, `errorbar`) so we have to be consistent here. I don't like  `linepoint`, `arealine`, `arealinepoint`  that much.

domoritz:
What was option 3?

kanitw:
> What was option 3?

If there is shape for line, automatically add point and drop shape from line?

domoritz:
> If there is shape for line, automatically add point and drop shape from line?

It feels magical but it's useful. I'm okay with adding it or also not adding it as it's not too hard to use the mark config. Should we show a warning?

kanitw:
>  Should we show a warning?

Nope

jakevdp:
One concern with the mark config route: since config is only at the top level, does it mean that it would not be possible to have a compound chart where lines are plotted both with and without points?

kanitw:
> Mark config is an object that is used instead of the mark string. It can appear anywhere in the spec, not just at the top level.

To clarify about terminology, "an object that is used instead of the mark string" is called [mark definition (object)](https://vega.github.io/vega-lite/docs/mark.html#mark-def).    So  having only "mark config" will have the issue that @jakevdp mentioned. 

But that's why I make `point` and `line` available as a property of the mark definition too. For example, we can set `mark` to `{"type": "area", "point": {}, "line": {}, "opacity": 0.5}` anywhere in the spec. (No need to be top-level.)

PR #3615, opened 2018-04-17
Add `trail` mark support

### LINKED ISSUE #3044: Support Trail Mark
https://vega.github.io/vega/docs/marks/trail/

This should be pretty similar to line, but use fill instead of stroke, and do not support interpolation.

kanitw:
Fix #3044 

(This is not really important, but definitely an easy one so I just did it on the flight.)


kanitw:
Nope, because it will break line interpolation too.

kanitw:
Plus,   color of line = stroke  while color of trail = fill.

PR #3623, opened 2018-04-18
Change Vega-Lite's default strokeCap to square so any data points tha…

kanitw:
…t is disconnected (surrounded by invalid values) still appear


kanitw:
From discussion with Jeff, we decided not to do this, but I'm creating this PR so that we can cherrypick this commit if we change our mind later.

PR #3625, opened 2018-04-19
Make line/area overlay transparent by default

kanitw:


kanitw:
Given the new tooltip logic for line/area in Vega 4 is better. We don't really need this!

PR #3629, opened 2018-04-20
Follow up Fixes / Refactor for Overlay Logic

kanitw:
More docs will follow in another PR


kanitw:
@domoritz  --- added docs / example.  You might want to re-review last commit before merging.

PR #3632, opened 2018-04-20
Make `aggregate` transform do not require field

kanitw:


build/vega-lite-schema.json
```diff
@@ -63,7 +63,7 @@
           "type": "string"
         },
         "field": {
-          "description": "The data field for which to compute aggregate function.",
+          "description": "The data field for which to compute aggregate function. (This is required for most aggregation operations except `\"count\"`.)",
```
domoritz: most -> all

PR #3634, opened 2018-04-21
Make tooltip support tooltip array for specifying multiple fields

kanitw:


test/compile/mark/mixins.test.ts
```diff
@@ -201,6 +201,22 @@ describe('compile/mark/mixins', () => {
     });
   });
 
+  describe('tootlip()', () => {
```
domoritz: tooltip

domoritz:
Works now! 

screen shot 2018-04-21 at 23 00 57

Issue #3488, opened 2018-03-08
Small non-zero values for size scale are not visible


jakevdp:
In Vega-Lite 2.x, if you encode a quantity with size and that quantity has elements with value zero, those points do not appear on the plot with the default scale:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "url": "https://vega.github.io/vega-datasets/data/seattle-weather.csv"
  },
  "encoding": {
    "color": {
      "field": "weather",
      "type": "nominal"
    },
    "size": {
      "field": "precipitation",
      "type": "quantitative"
    },
    "x": {
      "field": "temp_min",
      "type": "quantitative"
    },
    "y": {
      "field": "temp_max",
      "type": "quantitative"
    }
  },
  "mark": "point"
}
```
![vega](https://user-images.githubusercontent.com/781659/37137533-025f4a4c-225b-11e8-8f32-9cef28b571ee.png)

In Vega-Lite 1.x, this was not the case; here's the result of the above spec with VL1:

![vega 2](https://user-images.githubusercontent.com/781659/37137666-9ed20fd6-225b-11e8-878c-50ad71edcc7b.png)

In VL2, you can fix this by setting the size scale domain to start at a negative number:

```python
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "url": "https://vega.github.io/vega-datasets/data/seattle-weather.csv"
  },
  "encoding": {
    "color": {
      "field": "weather",
      "type": "nominal"
    },
    "size": {
      "field": "precipitation",
      "type": "quantitative",
      "scale": {"domain": [-1, 50]}
    },
    "x": {
      "field": "temp_min",
      "type": "quantitative"
    },
    "y": {
      "field": "temp_max",
      "type": "quantitative"
    }
  },
  "mark": "point"
}
```
![vega 1](https://user-images.githubusercontent.com/781659/37137535-0598ee5c-225b-11e8-91db-75fa094cbd06.png)

 I think that silently hiding valid data should be considered a bug, and that the domain should default to something like what is in the final panel.

domoritz:
I also consider this a bug. We should map the range to [0..max] but 0 should map to a circle that doesn't have an empty area inside.

domoritz:
Instead of setting the domain, we should set the range. 

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "width": 500,
  "height": 300,
[...]

PR #3638, opened 2018-04-22
Set smallest circle size to 1

domoritz:
image


domoritz:
Hmm, I wonder whether we should do this only for marks with a stroke. If you have a fill, size 0 should really not be visible.

src/compile/scale/range.ts
```diff
@@ -214,6 +214,10 @@ export function defaultRange(
 
 function sizeRangeMin(mark: Mark, zero: boolean, config: Config) {
   if (zero) {
+    if (util.contains(['point', 'square', 'circle'], mark)) {
+      // the smallest size for points should not be zero as the marks become invisible
```
kanitw: But this also make scale incorrect. 

I'd say, if you have "zero" value, better not use size.

kanitw: Maybe 0.01 or 0.1 is palatable

domoritz: Try it out, you still have white inside the circle but it looks worse.

kanitw: Already tried and it's terrible.  In general, using size to encode field with zero values is a bad idea anyway, so I don't think we should make size less accurate for this bad use case.

domoritz: The change in accuracy is small but https://github.com/vega/vega-lite/issues/3488 seems problematic. If you disagree, we should close the issue and this PR.

kanitw: Let's try production rule? https://github.com/vega/vega-lite/issues/3488#issuecomment-383388466

PR #3640, opened 2018-04-22
Make build:examples skip SVG if there is no Vega diff & Split build-examples from build-example

kanitw:
(This makes build-examples takes 1/3 total time)


package.json
```diff
@@ -28,9 +28,10 @@
     "build:only": "tsc && cp package.json build && browserify src/index.ts -p tsify -d -s vl | exorcist build/vega-lite.js.map > build/vega-lite.js",
     "postbuild": "node node_modules/uglify-js/bin/uglifyjs build/vega-lite.js -cm --source-map content=build/vega-lite.js.map,filename=build/vega-lite.min.js.map -o build/vega-lite.min.js && npm run schema",
     "build:examples": "npm run data && npm run build:only && npm run build:examples-only",
-    "build:examples-only": "npm run build:example && npm run build:examples-normalized",
+    "build:examples-only": "TZ=America/Los_Angeles scripts/build-examples.sh && npm run build:examples-normalized",
+    "build:examples-full": "TZ=America/Los_Angeles scripts/build-examples.sh 1",
```
kanitw: Instead of adding one more command, we should output generated vega version and check the version instead. (We only need to force re-build when we change vega version.)

scripts/build-example.sh
```diff
@@ -0,0 +1,17 @@
...
+echo "Compiling example(s) to $dir"
+
+for name in "$@"
+do
+  echo "Building $name"
+  rm -f examples/compiled/$name.vg.json
+  bin/vl2vg -p examples/specs/$name.vl.json > examples/compiled/$name.vg.json
+  ((! git diff --exit-code HEAD -- $dir/$name.vg.json) && node_modules/vega/bin/vg2svg --seed 123456789 examples/compiled/$name.vg.json > examples/compiled/$name.svg -b .) || true
+
+  # if ! git diff --exit-code HEAD -- $dir/$name.vg.json
```
domoritz: Remove comments?

scripts/build-examples.sh
```diff
@@ -5,31 +5,35 @@ dir=${dir-"examples/compiled"}
...
+  ls examples/specs/*.vl.json | parallel --eta --no-notice --plus --halt 1 "rm -f examples/compiled/{/..}.vg.json && bin/vl2vg -p {} > examples/compiled/{/..}.vg.json && ( ( ! git diff --no-patch HEAD -- examples/compiled/{/..}.vg.json || $forcesvg ) && rm -f examples/compiled/{/..}.svg && node_modules/.bin/vg2svg --seed 123456789 examples/compiled/{/..}.vg.json examples/compiled/{/..}.svg -b . || true )"
 else
-  for name in "$@"
-  do
-    echo "Building $name"
-    rm -f examples/compiled/$name.vg.json
-    bin/vl2vg -p examples/specs/$name.vl.json > examples/compiled/$name.vg.json
-    rm -f examples/compiled/$name.svg
-    node_modules/vega/bin/vg2svg --seed 123456789 examples/compiled/$name.vg.json > examples/compiled/$name.svg -b .
+  echo "Parallel not found! Sequentially generate vega specs from examples."
```
domoritz: Should we just delete this and reply on parallel?

Issue #3643, opened 2018-04-22
Make test run faster


kanitw:
Consider `jest` or some mocha parallel solutions

kanitw:
Actually the real test is pretty fast, but the tsc and schema before hand are not

kanitw:
The runtime test is pretty slow so parallelization may still help.

domoritz:
I don't think we can really speed up tsc and schema generation. However, we can use `tsc -w`.

domoritz:
Working on it

PR #3644, opened 2018-04-22
Test shouldn't require build, (only tsc) is fine?

kanitw:
Not 100% sure yet, need to test more

part of https://github.com/vega/vega-lite/issues/3643


kanitw:
Also run `lint` before `runtime:test`

PR #3646, opened 2018-04-22
Use rollup, webpack, and jest. 

### LINKED ISSUE #3190: Re-consider rollup
Vega uses rollup. We could consistently use rollup in all our projects. Would be great if somebody tried it and sent a pr :-)

### LINKED ISSUE #3643: Make test run faster

domoritz:
Remove test gallery. Fixes #3643, fixes #3190

Vega-Lite is now only 162K.


package.json
```diff
@@ -23,47 +24,33 @@
...
+    "schema": "ts-json-schema-generator --path src/index.ts --type TopLevelSpec > build/vega-lite-schema.json && npm run renameschema && cp build/vega-lite-schema.json _data/",
     "renameschema": "scripts/rename-schema.sh",
     "presite": "npm run prebuild && npm run data && npm run build:site && npm run build:toc && npm run build:versions && scripts/create-example-pages",
     "site": "bundle exec jekyll serve --incremental",
     "lint": "tslint -p .",
-    "test": "npm run build:only && npm run test:only && npm run test:runtime && npm run lint",
-    "posttest": "npm run schema && npm run data && npm run mocha:examples",
-    "test:nocompile": "npm run test:only && npm run test:runtime && npm run lint && npm run mocha:examples",
-    "test:only": "nyc --reporter=html --reporter=text-summary npm run mocha:test",
+    "test": "npm run lint && npm run schema && jest && npm run test:runtime",
```
kanitw: Can we run 

1) unit test (without example schema test)
2) lint
3) schema
4) example schema test
5) runtime test

This would be the most optimal for workflow as we normally want to test logic first before linting.

domoritz: Sure. I will change it. The cool thing about running all tests in one step is that they can run in parallel.

.vscode/settings.json
```diff
@@ -4,7 +4,6 @@
   "files.exclude": {
     "coverage/**/*": true,
     "_site/**/*": true,
-    ".nyc_output": true
```
kanitw: why removing this?

examples/specs/window_residual_graph.vl.json
```diff
@@ -8,20 +8,20 @@
...
-            "field": "IMDB_Rating",
-            "as": "AverageRating"
-        }],
-        "frame": [null, null]
-      },
-      {
-        "calculate": "datum.IMDB_Rating - datum.AverageRating",
-        "as": "RatingDelta"
-      }
+    {"filter": "datum.IMDB_Rating != null"},
```
kanitw: Why are these tabbings related to build changes?

package.json
```diff
@@ -9,9 +9,10 @@
   ],
   "homepage": "https://vega.github.io/vega-lite/",
   "description": "Vega-Lite is a concise high-level language for interactive visualization.",
-  "main": "build/src/index.js",
+  "main": "build/vega-lite.js",
```
kanitw: This may break a lot of things?

package.json
```diff
@@ -22,48 +23,34 @@
...
     "build:example": "TZ=America/Los_Angeles scripts/build-example.sh",
     "build:toc": "bundle exec jekyll build -q && scripts/generate-toc",
-    "build:site": "browserify site/static/main.ts -p [tsify -p site] -d | exorcist build/site/main.js.map > build/site/main.js",
+    "build:site": "tsc -p site && webpack --config site/webpack.config.js",
     "build:versions": "scripts/update-version.sh",
-    "build:test-gallery": "browserify test-gallery/main.ts -p [tsify -p test-gallery] -d > build/test-gallery/main.js",
     "check:examples": "scripts/check-examples.sh",
     "check:schema": "scripts/check-schema.sh",
-    "clean": "rm -rf build && rm -f vega-lite.* & find -E src test site examples -regex '.*\\.(js|js.map|d.ts)' -delete && rm -f examples/compiled/*.png && find site/examples ! -name 'index.md' -type f -exec rm -f {} +",
+    "clean": "rm -rf build && rm -f examples/compiled/*.png && find site/examples ! -name 'index.md' -type f -delete",
```
kanitw: Why are there `.png` included?

src/util.ts
```diff
@@ -1,7 +1,8 @@
-import * as stableStringify from 'json-stable-stringify';
+import * as stableStringify_ from 'json-stable-stringify';
 import {isArray, isNumber, isString, splitAccessPath, stringValue} from 'vega-util';
 import {isLogicalAnd, isLogicalNot, isLogicalOr, LogicalOperand} from './logical';
 
+const stableStringify = stableStringify_['default'] || stableStringify_;
```
kanitw: Why do we need to do this?  Shouldn't it be just one of these two?

tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
-    "outDir": "./build",
+    "outDir": "./build/src/",
```
kanitw: Can you explain why you need to add `src/`?

.vscode/settings.json
domoritz: We don't use nyc anymore

examples/specs/window_residual_graph.vl.json
domoritz: They are not. I got an error in the file and saw it.

package.json
domoritz: This is consistent with Vega, though. https://github.com/vega/vega/blob/master/package.json

package.json
domoritz: Because we need to have the geo examples as PNGs.

src/util.ts
domoritz: This is just to be safe. Rollup can change how modules are exposed depending on what plugin you use. This always works.

tsconfig.json
domoritz: Because I copy the package.json into build/src

domoritz:
I confirmed that this works with Vega-Tooltip, Vega-Embed, and the Vega-Editor.

tsconfig.json
domoritz: Oh, and otherwise everything would end up in the build directory and that's breaking compassql.

PR #3668, opened 2018-04-26
Fix example scripts

kanitw:


domoritz:
Did you try it?

kanitw:
Yes, in callout branch.  This is just copying it here.

kanitw:
See https://github.com/vega/vega-lite/pull/3669 for example

PR #3675, opened 2018-04-27
Resolve conflict between scale.domain and bin.extent

### LINKED ISSUE #3575: Binned x/y encoding ignores scale domain parameter
When using a binned spatial encoding, custom scale domain settings seem to be ignored. The generated Vega pulls the scale domain from the data, not the provided custom domain.

Here is an example spec:
```json
{
  "data": {"values": [0.1, 0.6, 2.3]},
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "data",
      "type": "quantitative",
      "bin": {"extent": [0, 10], "step": 0.5},
      "scale": {"domain": [0, 10]}
    },
    "y": {
      "aggregate": "count",
      "type": "quantitative",
      "scale": {"domain": [0, 2]}
    }
  }
}
```

I expect the x-axis domain to be `[0, 10]`, but instead it is `[0, 2.5]`.

yhoonkim:
Fix #3575

- If `scale.domain` is specified, use `scale.domain` despite `bin`.
- If `scale.domain` is specified, axis' ticks are ranged over `scale.domain` instead of the signal of the `bin.extent`.
- If `bin.extent` is specified, the scale domain is `bin.extent` by default.


kanitw:
Thanks so much @yhoonkim!

PR #3676, opened 2018-04-27
Flatten nested fields

### LINKED ISSUE #3251: Data parsing does not support nested fields
As described in https://github.com/vega/vega-parser/issues/64#issuecomment-355054535 - if the source data has nested fields, e.g.
```
data: {
 xxx: [
  { a: { b: '2018-01-01T00:15:07.5235123+00:00' }},
  { a: { b: '2018-01-02T00:15:07.5235123+00:00' }},
  { a: { b: '2018-01-03T00:15:07.5235123+00:00' }}
] }
```

The data is not parsed with this format:

```
      "format": {
        "property": "xxx",
        "type":"json",
        "parse": { "a.b": "date" }
```

### LINKED ISSUE #3368: Nested data accesses break output
This spec (https://github.com/vega/vega-lite/pull/3366#issuecomment-363974114) should generate the Vega below. However, it breaks because of two issues

1. Parsing json as a number removes the nesting. Consequently, data cannot be accessed as nested properties
2. Records always only have either `forecast` or `actual` but never both. We try to access the nested properties, which causes the Vega runtime to throw its hands up. 

I'm not too concerned about the second issue but the first one should be fixed. 

Vega-Lite spec:

```json
{
    "description": "A layered bar chart with floating bars",
    "data":{ "url": "https://raw.githubusercontent.com/vega/vega-datasets/gh-pages/data/weather.json" },
    "layer": [
        {
            "mark": {
                "type": "bar", "style": "box"
            },
            "encoding": {
                "y": {
                    "field": "record.low",
                    "type": "quantitative",
                    "scale": {"domain": [10,70]},
                    "axis": {"title": "Temperature"}
                },
                "y2": {
                    "field": "record.high",
                    "type": "quantitative"
                },
                "x": {
                    "field": "id",
                    "type": "ordinal",
                    "axis": {"title": "Day"}
                },
                 "size": {"value": 15},
                 "color": {"value": "orange"}
            }
        },
        {
            "mark": {
                "type": "bar", "style": "box"
            },
            "encoding": {
                "y": {
                    "field": "normal.low",
                    "type": "quantitative"
                },
                "y2": {
                    "field": "normal.high",
                    "type": "quantitative"
                },
                "x": {
                    "field": "id",
                    "type": "ordinal"
                },
                "size": {"value": 15}
            }
        },
    {
      "mark": {"type": "bar", "style": "box"},
      "encoding": {
        "y": {
          "field": "actual.low",
          "type": "quantitative"
        },
        "y2": {
          "field": "actual.high",
          "type": "quantitative"
        },
        "x": {
          "field": "id",
          "type": "ordinal"
        },
        "size": {"value": 10}
      }
    },
    {
      "mark": {"type": "bar", "style": "box"},
      "encoding": {
        "y": {
          "field": "forecast.low.low",
          "type": "quantitative"
        },
        "y2": {
          "field": "forecast.low.high",
          "type": "quantitative"
        },
        "x": {
          "field": "id",
          "type": "ordinal"
        },
        "size": {"value": 10}
      }
    },
    {
      "mark": {"type": "bar", "style": "box"},
      "encoding": {
        "y": {
          "field": "forecast.low.high",
          "type": "quantitative"
        },
        "y2": {
          "field": "forecast.high.low",
          "type": "quantitative"
        },
        "x": {
          "field": "id",
          "type": "ordinal"
        },
        "size": {"value": 3}
      }
    },
    {
      "mark": {"type": "bar", "style": "box"},
      "encoding": {
        "y": {
          "field": "forecast.high.low",
          "type": "quantitative"
        },
        "y2": {
          "field": "forecast.high.high",
          "type": "quantitative"
        },
        "x": {
          "field": "id",
          "type": "ordinal"
        },
        "size": {"value": 10}
      }
     }
    ]
}
```

Working Vega:

```json
{
  "$schema": "https://vega.github.io/schema/vega/v3.0.json",
  "description": "A layered bar chart with floating bars",
  "autosize": "pad",
  "padding": 5,
  "height": 200,
  "style": "cell",
  "data": [
    {
      "name": "source_0",
      "url": "https://raw.githubusercontent.com/vega/vega-datasets/gh-pages/data/weather.json",
      "format": {
        "type": "json"
      }
    },
    {
      "name": "data_0",
      "source": "source_0",
      "transform": [
        {
          "type": "formula",
          "expr": "toNumber(datum[\"record\"][\"low\"])",
          "as": "record.low"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"record\"][\"high\"])",
          "as": "record.high"
        }
      ]
    },
    {
      "name": "data_1",
      "source": "source_0",
      "transform": [
        {
          "type": "formula",
          "expr": "toNumber(datum[\"normal\"][\"low\"])",
          "as": "normal.low"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"normal\"][\"high\"])",
          "as": "normal.high"
        }
      ]
    },
    {
      "name": "data_2",
      "source": "source_0",
      "transform": [
        {
          "type": "filter",
          "expr": "datum[\"actual\"]"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"actual\"][\"low\"])",
          "as": "actual.low"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"actual\"][\"high\"])",
          "as": "actual.high"
        }
      ]
    },
    {
      "name": "data_3",
      "source": "source_0",
      "transform": [
        {
          "type": "filter",
          "expr": "datum[\"forecast\"]"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"forecast\"][\"low\"][\"low\"])",
          "as": "forecast.low.low"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"forecast\"][\"low\"][\"high\"])",
          "as": "forecast.low.high"
        }
      ]
    },
    {
      "name": "data_4",
      "source": "source_0",
      "transform": [
        {
          "type": "filter",
          "expr": "datum[\"forecast\"]"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"forecast\"][\"low\"][\"high\"])",
          "as": "forecast.low.high"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"forecast\"][\"high\"][\"low\"])",
          "as": "forecast.high.low"
        }
      ]
    },
    {
      "name": "data_5",
      "source": "source_0",
      "transform": [
        {
          "type": "filter",
          "expr": "datum[\"forecast\"]"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"forecast\"][\"high\"][\"low\"])",
          "as": "forecast.high.low"
        },
        {
          "type": "formula",
          "expr": "toNumber(datum[\"forecast\"][\"high\"][\"high\"])",
          "as": "forecast.high.high"
        }
      ]
    }
  ],
  "signals": [
    {
      "name": "x_step",
      "value": 21
    },
    {
      "name": "width",
      "update": "bandspace(domain('x').length, 0.1, 0.05) * x_step"
    }
  ],
  "marks": [
    {
      "name": "layer_0_marks",
      "type": "rect",
      "style": [
        "bar",
        "box"
      ],
      "from": {
        "data": "data_0"
      },
      "encode": {
        "update": {
          "fill": {
            "value": "orange"
          },
          "xc": {
            "scale": "x",
            "field": "id",
            "band": 0.5
          },
          "width": {
            "value": 15
          },
          "y": {
            "scale": "y",
            "field": "record\\.low"
          },
          "y2": {
            "scale": "y",
            "field": "record\\.high"
          }
        }
      }
    },
    {
      "name": "layer_1_marks",
      "type": "rect",
      "style": [
        "bar",
        "box"
      ],
      "from": {
        "data": "data_1"
      },
      "encode": {
        "update": {
          "fill": {
            "value": "#4c78a8"
          },
          "xc": {
            "scale": "x",
            "field": "id",
            "band": 0.5
          },
          "width": {
            "value": 15
          },
          "y": {
            "scale": "y",
            "field": "normal\\.low"
          },
          "y2": {
            "scale": "y",
            "field": "normal\\.high"
          }
        }
      }
    },
    {
      "name": "layer_2_marks",
      "type": "rect",
      "style": [
        "bar",
        "box"
      ],
      "from": {
        "data": "data_2"
      },
      "encode": {
        "update": {
          "fill": {
            "value": "black"
          },
          "xc": {
            "scale": "x",
            "field": "id",
            "band": 0.5
          },
          "width": {
            "value": 10
          },
          "y": {
            "scale": "y",
            "field": "actual\\.low"
          },
          "y2": {
            "scale": "y",
            "field": "actual\\.high"
          }
        }
      }
    },
    {
      "name": "layer_3_marks",
      "type": "rect",
      "style": [
        "bar",
        "box"
      ],
      "from": {
        "data": "data_3"
      },
      "encode": {
        "update": {
          "fill": {
            "value": "black"
          },
          "xc": {
            "scale": "x",
            "field": "id",
            "band": 0.5
          },
          "width": {
            "value": 10
          },
          "y": {
            "scale": "y",
            "field": "forecast\\.low\\.low"
          },
          "y2": {
            "scale": "y",
            "field": "forecast\\.low\\.high"
          }
        }
      }
    },
    {
      "name": "layer_4_marks",
      "type": "rect",
      "style": [
        "bar",
        "box"
      ],
      "from": {
        "data": "data_4"
      },
      "encode": {
        "update": {
          "fill": {
            "value": "black"
          },
          "xc": {
            "scale": "x",
            "field": "id",
            "band": 0.5
          },
          "width": {
            "value": 3
          },
          "y": {
            "scale": "y",
            "field": "forecast\\.low\\.high"
          },
          "y2": {
            "scale": "y",
            "field": "forecast\\.high\\.low"
          }
        }
      }
    },
    {
      "name": "layer_5_marks",
      "type": "rect",
      "style": [
        "bar",
        "box"
      ],
      "from": {
        "data": "data_5"
      },
      "encode": {
        "update": {
          "fill": {
            "value": "black"
          },
          "xc": {
            "scale": "x",
            "field": "id",
            "band": 0.5
          },
          "width": {
            "value": 10
          },
          "y": {
            "scale": "y",
            "field": "forecast\\.high\\.low"
          },
          "y2": {
            "scale": "y",
            "field": "forecast\\.high\\.high"
          }
        }
      }
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "band",
      "domain": {
        "fields": [
          {
            "data": "data_0",
            "field": "id"
          },
          {
            "data": "data_1",
            "field": "id"
          },
          {
            "data": "data_2",
            "field": "id"
          },
          {
            "data": "data_3",
            "field": "id"
          },
          {
            "data": "data_4",
            "field": "id"
          },
          {
            "data": "data_5",
            "field": "id"
          }
        ],
        "sort": true
      },
      "range": {
        "step": {
          "signal": "x_step"
        }
      },
      "paddingInner": 0.1,
      "paddingOuter": 0.05
    },
    {
      "name": "y",
      "type": "linear",
      "domain": {
        "fields": [
          [
            10,
            70
          ],
          {
            "data": "data_1",
            "field": "normal\\.low"
          },
          {
            "data": "data_1",
            "field": "normal\\.high"
          },
          {
            "data": "data_2",
            "field": "actual\\.low"
          },
          {
            "data": "data_2",
            "field": "actual\\.high"
          },
          {
            "data": "data_3",
            "field": "forecast\\.low\\.low"
          },
          {
            "data": "data_3",
            "field": "forecast\\.low\\.high"
          },
          {
            "data": "data_4",
            "field": "forecast\\.low\\.high"
          },
          {
            "data": "data_4",
            "field": "forecast\\.high\\.low"
          },
          {
            "data": "data_5",
            "field": "forecast\\.high\\.low"
          },
          {
            "data": "data_5",
            "field": "forecast\\.high\\.high"
          }
        ]
      },
      "range": [
        {
          "signal": "height"
        },
        0
      ],
      "nice": true,
      "zero": false
    }
  ],
  "axes": [
    {
      "title": "Day",
      "scale": "x",
      "orient": "bottom",
      "labelOverlap": true,
      "zindex": 1,
      "encode": {
        "labels": {
          "update": {
            "angle": {
              "value": 270
            },
            "align": {
              "value": "right"
            },
            "baseline": {
              "value": "middle"
            }
          }
        }
      }
    },
    {
      "title": "Temperature",
      "scale": "y",
      "orient": "left",
      "labelOverlap": true,
      "tickCount": {
        "signal": "ceil(height/40)"
      },
      "zindex": 1
    },
    {
      "scale": "y",
      "orient": "left",
      "domain": false,
      "grid": true,
      "labels": false,
      "maxExtent": 0,
      "minExtent": 0,
      "tickCount": {
        "signal": "ceil(height/40)"
      },
      "ticks": false,
      "zindex": 0,
      "gridScale": "x"
    }
  ],
  "config": {
    "axisY": {
      "minExtent": 30
    }
  }
}
```

### LINKED ISSUE #3369: Issues with nesting
* #3361
* #3368
* #3351
* #3251 
* #3441

I thought I had fixed most of the issues in https://github.com/vega/vega-lite/pull/3141 so it would be good to look into this patch. 

There are two ways forward. One is to clean up these issues with proper fixes. The other is to just flatten nested accesses. I feel the second option might make sense since we flatten nested data anyway when we parse it as a number.

### LINKED ISSUE #3441: Unable to filter out null objects if referenced by encoding
```json
{
  "description": "Unable to truly filter data",
  "data": {
    "values":[
      {
        "date": "2000-01-01"
      },
      {
        "date": "2000-02-01",
        "options": {
          "price": 16 
        }
      },
      {
        "date": "2000-03-01",
        "options": {
          "price": 17 
        }
      } 
    ]
  },
  "transform": [{"filter": "datum.options != null"}],
  "mark": "line",
  "encoding": {
    "y": {"field": "options.price", "type": "quantitative"}
  }
}
```

This will throw "ERROR TypeError: Cannot read property 'price' of undefined" in the console.
I'm not sure why "options.price" would ever try to be evaluated for the record which was supposed to be filtered out.

domoritz:
Always flattens nested fields. Flattening happens when data is parsed. This is similar to what we are doing already since an expression always returns flattened data and we often generate expressions instead of suing vega's native format.parse. 

Fixes #3369
Fixes #3368
Fixes #3251
Fixes #3441

The logic is now as follows: the input to any transform could be nested. After the transforms, all data is flat and should be accessed as such.


domoritz:
@nyurik do you have a few test cases that I can try?

nyurik:
Not readily available, no.

domoritz:
In expressions, we require users to write the correct, non-nested, version.

```json
{
  "description": "Unable to truly filter data",
  "data": {
    "values":[
      {
        "rank": "1",
        "options": {
          "price": 10
        }
      },
      {
        "rank": "2",
        "options": {
          "price": 16 
        }
      },
      {
        "rank": "3",
        "options": {
          "price": 17 
        }
      } 
    ]
  },
  "transform": [{"filter": "datum['options.price'] > 10"}],
  "mark": "line",
  "encoding": {
    "x": {"field": "rank", "type": "ordinal"},
    "y": {"field": "options.price", "type": "quantitative"}
  }
}
```

However, for all other filters, aggregates, and so on where users specify a field, we automatically o the right thing. 

```json
{
  "description": "Unable to truly filter data",
  "data": {
    "values":[
      {
        "rank": "1",
        "options": {
          "price": 10
        }
      },
      {
        "rank": "2",
        "options": {
          "price": 16 
        }
      },
      {
        "rank": "3",
        "options": {
          "price": 16
        }
      } 
    ]
  },
  "transform": [{"filter": {"field": "options.price", "equal": 16}}],
  "mark": "line",
  "encoding": {
    "x": {"field": "rank", "type": "ordinal"},
    "y": {"field": "options.price", "type": "quantitative"}
  }
}
```

src/compile/data/formatparse.ts
```diff
@@ -74,6 +79,11 @@ export class ParseNode extends DataFlowNode {
             return;
           }
           parse[fieldDef.field] = 'number';
+        } else if (countAccessPath(fieldDef.field) > 1) {
```
kanitw: countAccessPath is a pretty confusing name.  

`accessPathDepth`?

kanitw: Also why is this case checked after `isTimeFieldDef` / `isNumberFieldDef`? 

What if a number field def has nested field?

kanitw:
> In expressions, we require users to write the correct, non-nested, version.

Why is that the case?  What if you never have an encoding that uses the field `"options.price"`?  

Your example is correct because the encoding makes `options.price` get flattened so `datum['options.price']` in the expression is fine. 

but in general, isn't it more correct to filter by the nested expression? 

```
{
  "description": "Unable to truly filter data",
  "data": {
    "values":[
      {
        "rank": "1",
        "options": {
          "price": 10
        }
      },
      {
        "rank": "2",
        "options": {
          "price": 16 
        }
      },
      {
        "rank": "3",
        "options": {
          "price": 17 
        }
      } 
    ]
  },
  "transform": [{"filter": "datum.options.price > 10"}],
  "mark": "point",
  "encoding": {
    "x": {"field": "rank", "type": "ordinal"}
  }
}
```

Note that what you do here is not literally flattening, but rather augment each data point with a flattened field, while the old nested field still remains.

examples/specs/test_subobject_missing.vl.json
```diff
@@ -0,0 +1,29 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "Unable to truly filter data",
```
kanitw: This description is very confusing.   

The description suggests that this spec does NOT work, but it seems to be working.  So I don't understand what's the point you're trying to make here.   (What is "truly filter" data?)

src/compile/data/formatparse.ts
domoritz: Then we flatten the field already. Note that every expression flattens.

examples/specs/test_subobject_missing.vl.json
domoritz: Ups, copy pasta.

domoritz:
@kanitw The case you described is okay. Yes, you don't gave to use the flattened field and there may be fields that we don't know how to flatten.

kanitw:
> Yes, you don't gave to use the flattened field and there may be fields that we don't know how to flatten.

I don't understand this sentence.  Can you explain more?

domoritz:
The idea with this PR is that as long as you only use fields and no expressions, everything will just work even with nested fields because we automatically flatten them. However, if you use an expression, we don't parse them and thus you will be able to access the nested fields as well. This is okay and I don't think we want to prevent it.

examples/compiled/test_subobject_nested.vg.json
```diff
@@ -84,17 +84,17 @@
             "transform": [
                 {
                     "type": "formula",
-                    "expr": "toNumber(datum[\"source\"][\"reco\"])",
+                    "expr": "toNumber(datum[\"source\"] && datum[\"source\"][\"reco\"])",
```
kanitw: if source is missing, what this does output?

kanitw:
In https://github.com/vega/vega-lite/issues/3369, there are links to many other related issues (mostly in the field nesting milestones). 

Does this PR fix each of those, if so, listed them so we can close once we merge?

examples/compiled/test_subobject_nested.vg.json
domoritz: `null`

screen shot 2018-04-29 at 10 59 01

src/compile/data/formatparse.ts
```diff
@@ -74,6 +80,10 @@ export class ParseNode extends DataFlowNode {
             return;
           }
           parse[fieldDef.field] = 'number';
+        } else if (accessPathDepth(fieldDef.field) > 1) {
+          // For non-date/non-number (strings and booleans), derive a flattened field for a referenced nested field. 
+          // (Parsing numbers / dates already flatten numeric and temporal fields.)
+          parse[fieldDef.field] = 'flatten';
```
kanitw: @domoritz   Don't you have to add the same logic above for `transform`?

kanitw: Side issue: I also wonder if it is behaving correctly for standalone timeUnit transform. 
(the field in `timeUnit` won't get included in parse -- see `line_timeunit_transform` for example.)

src/compile/data/formatparse.ts
```diff
@@ -113,7 +123,13 @@ export class ParseNode extends DataFlowNode {
     other.remove();
   }
   public assembleFormatParse() {
-    return this._parse;
+    const formatParse = {};
```
kanitw: Clarifying question:  when do we still use format parse? 

(It seems like for the most case, we use expression. -- maybe add comment to clarify?)

src/compile/data/formatparse.ts
domoritz: > Don't you have to add the same logic above for transform -- for all of aggregate, timeUnit, window, filter.

Hmm, yes. I'm surprised we don't do that already. That seems like an orthogonal bug, doesn't it?

domoritz: > Side issue: I also wonder if the original logic above is behaving correctly for standalone timeUnit transform.

It looks correct although we parse the generates date as date again (which is not necessary but we need to track fields to remove it). 

```json
{
  "mark": "point",
  "transform": [{
    "timeUnit": "month",
    "field": "foo.date",
    "as": "month"
  }],
  "encoding": {
    "x": {
      "field": "month",
      "type": "temporal"
    }
  }
}
```

```json
{
      "name": "source_0",
      "format": {"type": "csv"},
      "transform": [
        {
          "type": "formula",
          "as": "month",
          "expr": "datetime(0, month(datum[\"foo\"] && datum[\"foo\"][\"date\"]), 1, 0, 0, 0, 0)"
        },
        {
          "type": "formula",
          "expr": "toDate(datum[\"month\"])",
          "as": "month"
        },
        {
          "type": "filter",
          "expr": "datum[\"month\"] !== null && !isNaN(datum[\"month\"])"
        }
      ]
    }
```

domoritz:
> Does this PR fix each of those, if so, listed them so we can close once we merge?

Fixes #3368
Fixes #3251
Fixes #3441

kanitw:
> > @domoritz Don't you have to add the same logic above for transform -- for all of aggregate, timeUnit, window, filter.

> Hmm, yes. I'm surprised we don't do that already. That seems like an orthogonal bug, doesn't it?

It's not orthogonal since you didn't add the new nested field logic (this PR's concern) to above?  This means that nested `"field"` in `aggregate`, `timeUnit`, `filter`, `window`, `bin`  may not work?  

(It might, but you better check)

domoritz:
test

- [x] TimeUnit (works as shown above)
- [x] Calculate (we already had that)
- [ ] Bin
- [ ] Aggregate
- [ ] Filter
- [ ] Window

PR #3679, opened 2018-04-28
Restructure Gallery

kanitw:


_data/examples.json
```diff
@@ -193,37 +237,24 @@
       "title": "Trellis Area"
     }
   ],
-  "Layering": [
+  "Using Layer to Add Labels": [
```
domoritz: why capitalize add?

_data/examples.json
```diff
@@ -134,15 +166,27 @@
...
+      "name": "rect_binned_heatmap",
+      "title": "Table Binned heatmap"
+    },
+    {
+      "name": "circle_github_punchcard",
+      "title": "Table Bubble Plot (Github Punch Card)"
     }
   ],
-  "Window Transform": [
+  "Advanced Calculations": [
```
domoritz: Let's leave this as window transforms until we have more.

_data/examples.json
```diff
@@ -233,20 +264,45 @@
...
+      "title": "Candlestick Chart"
+    },
+    {
+      "name": "layer_histogram_global_mean",
+      "title": "Histogram with a Global Mean Overlay"
     },
     {
       "name": "layer_falkensee",
-      "title": "Falkensee",
+      "title": "Line chart with Hilighted Rectangles",
```
domoritz: typo

_data/examples.json
kanitw: I think a lot of people who come to the gallery could be novice who do not know what window is.  Naming the section "window" makes it more confusing.

_data/examples.json
kanitw: chart => Chart? 

(Falkensee is a terrible name as it doesn't suggest what this example is for at all.

_data/examples.json
kanitw: It's a heading.

_data/examples.json
kanitw: Plus, a lot of these examples are using window and calculate to achieve the result, not purely "window".

_data/examples.json
domoritz: Hilighted

domoritz: > Falkensee is a terrible name

Also, thanks.

_data/examples.json
domoritz: I think this heading is inconsistent with the others. The heading are about what, now what for.

kanitw: Well, using layer to label is different from when you really layer graphics.  Plus, some people will be looking around to see how to add labels.

PR #3690, opened 2018-05-02
Add oz to list of libs in docs

metasoarous:
Add note of more fully featured Clojure library oz (fork of vizard) to docs, as suggested here: https://github.com/metasoarous/oz/issues/11


domoritz:
Thank you! Can you mention that it is a fork?

metasoarous:
OK; This is now noted.

PR #3693, opened 2018-05-03
Allow making line uses original order in the data

### LINKED ISSUE #1020: Allow line’s path to be ordered by the original order of the data.
Currently, we always sort line's path based on fields of the discrete axis (x or y) or specified `order` channel.   However, users might want to order line based on the original order in the data.

kanitw:
Fix #1020


kanitw:
Bad, but working example. 

image

PR #3697, opened 2018-05-04
Disable format parsing and track parsed fields

### LINKED ISSUE #3282: Parse is added after transform
The parsing should be done before the aggregate transform. Note that this is not true in general for all parses. The parse of `max_price` should happen after the aggregation. 

```json
{
  "width": 600,
  "height": 300,
  "data": {
    "url": "data/stocks.csv",
    "format": {"parse": {"price": "number"}}
  },
  "transform": [{
    "aggregate": [{"op": "max", "field": "price", "as": "max_price"}],
    "groupby": ["date"]
  }],
  "mark": "point",
  "encoding": {
    "y": {"field": "max_price", "type": "quantitative"}
  }
}
```

### LINKED ISSUE #3665: Disable format parse
Please:
- [x] Provide an example spec in JSON and, if applicable, screenshots, GIF videos (e.g. using [LICEcap](https://www.cockos.com/licecap/)), or working example (e.g. based on [this Block](https://bl.ocks.org/domoritz/455e1c7872c4b38a58b90df0c3d7b1b9)).
- [x] Please ensure all code snippets and error messages are formatted in appropriate code blocks. See [Creating and highlighting code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks/).

Is it possible to get vega lite to compile vega that doesn't try to transform the data to the right type? For example, in [the bar example](https://vega.github.io/editor/#/examples/vega-lite/bar), it has this transform:

```json
    {
      "name": "data_0",
      "source": "source_0",
      "transform": [
        {"type": "formula", "expr": "toNumber(datum[\"b\"])", "as": "b"},
        {
          "type": "filter",
          "expr": "datum[\"b\"] !== null && !isNaN(datum[\"b\"])"
        }
      ]
    }
```

I know I can remove the `null` checks by setting `invalidValues` to `null` in the config, but I am still left with the `toNumber` conversion. What if I already know it will be a number, isn't this conversion superfluous? Is there a way to disable it and have the vega that is produced only create one data source?

I am asking because [the mapd implementation of transform](https://www.mapd.com/docs/latest/mapd-core-guide/vegaReference/#vegarefdatatrans) only supports `formula` transforms against previously aggregated columns and [I am trying to render vega lite with mapd](https://github.com/Quansight/jupyter-mapd-renderer/pull/7).

domoritz:
Fixes #3665 
Fixes #3282

CC @saulshanabrook

### DIFF
--- a/src/compile/data/formatparse.ts
+++ b/src/compile/data/formatparse.ts
@@ -90,11 +90,15 @@ export class ParseNode extends DataFlowNode {
 
     // Custom parse should override inferred parse
     const data = model.data;
-    if (data && data.format && data.format.parse) {
+    if (data && data.format) {
       const p = data.format.parse;
-      keys(p).forEach(field => {
-        parse[field] = p[field];
-      });
+      if (p === null) {
+        return null;  // disable parsing completely
+      } else if (p) {
+        keys(p).forEach(field => {
+          parse[field] = p[field];
+        });
+      }
     }
 
     // We should not parse what has already been parsed in a parent
@@ -106,6 +110,13 @@ export class ParseNode extends DataFlowNode {
         delete parse[field];
       }
     });
+
+    // filter out parse == null (do not parse the field)
+    for (const field of keys(parse)) {
+      if (parse[field] === null) {
+        delete parse[field];
+      }
+    }
 
     if (keys(parse).length === 0) {
       return null;
--- a/src/compile/data/formatparse.ts
+++ b/src/compile/data/formatparse.ts
@@ -1,15 +1,14 @@
 import {toSet} from 'vega-util';
-
+import {AncestorParse} from '.';
 import {isCountingAggregateOp} from '../../aggregate';
 import {isNumberFieldDef, isTimeFieldDef} from '../../fielddef';
 import * as log from '../../log';
-import {forEachLeaf} from '../../logical';
-import {isFieldPredicate} from '../../predicate';
-import {isCalculate, isFilter, Transform} from '../../transform';
 import {accessPathDepth, accessPathWithDatum, Dict, duplicate, keys, removePathFromField, StringSet} from '../../util';
 import {VgFormulaTransform} from '../../vega.schema';
 import {isFacetModel, isUnitModel, Model} from '../model';
+import {Split} from '../split';
 import {DataFlowNode} from './dataflow';
+
 
 /**
  * @param field The field.
@@ -27,6 +26,8 @@ function parseExpression(field: string, parse: string): string {
     return `toDate(${f})`;
   } else if (parse === 'flatten') {
     return f;
+  } else if (parse === null || parse === 'derived') {
+    return null;  // not not parse field
   } else if (parse.indexOf('date:') === 0) {
     const specifier = parse.slice(5, parse.length);
     return `timeParse(${f},${specifier})`;
@@ -40,7 +41,7 @@ function parseExpression(field: string, parse: string): string {
 }
 
 export class ParseNode extends DataFlowNode {
-  private _parse: Dict<string> = {};
+  private _parse: Dict<string>;
 
   public clone() {
     return new ParseNode(null, duplicate(this.parse));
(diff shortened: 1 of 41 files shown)


src/data.ts
```diff
@@ -3,16 +3,21 @@
...
+   * If set to `"auto"` (the default), perform automatic type inference to determine the desired data types.
+   * If set to `null`, disable type inference based on the spec and only use type inference based on the data.
+   * Alternatively, a parsing directive object can be provided for explicit data types. Each property of the object corresponds to a field name, and the value to the desired data type (one of `"number"`, `"boolean"`, `"date"`, or null (do not parse the field)).
    * For example, `"parse": {"modified_on": "date"}` parses the `modified_on` field in each input record a Date value.
    *
    * For `"date"`, we parse data based using Javascript's [`Date.parse()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse).
    * For Specific date formats can be provided (e.g., `{foo: 'date:"%m%d%Y"'}`), using the [d3-time-format syntax](https://github.com/d3/d3-time-format#locale_format). UTC date format parsing is supported similarly (e.g., `{foo: 'utc:"%m%d%Y"'}`). See more about [UTC time](timeunit.html#utc)
    */
-  parse?: 'auto' | object;
+  parse?: 'auto' | Parse | null;
```
kanitw: What's the behavior for  providing an object for only a subset of fields?

(Do we do "auto" for all other fields unmentioned?)

domoritz:
I updated the PR to properly track parsed fields. This is something we have not done correctly until now.

src/compile/data/formatparse.ts
```diff
@@ -52,77 +53,85 @@ export class ParseNode extends DataFlowNode {
...
+          log.warn(log.message.differentParse(field, explicit[field], parsedAs));
+        }
       }
     }
 
-    if (keys(parse).length === 0) {
+    const parse = new Split(explicit, implicit);
+
+    // add the format parse from this model so that children don't parse the same field again
+    ancestorParse.copyAll(parse);
```
kanitw: So you merge this into parent, but still create a local parse node? 

I'm pretty confused.

src/compile/data/index.ts
```diff
@@ -43,7 +44,23 @@ export interface DataComponent {
...
 
   /**
-   * Parse properties passed down from ancestors.
+   * Parse properties passed down from ancestors. Helps us to keep track of what has been parsed or is derived.
    */
-  ancestorParse: Dict<string>;
+  ancestorParse?: AncestorParse;
+}
+
+export class AncestorParse extends Split<Dict<string>> {
```
kanitw: Why call it AncestorParse?  I don't see anything that is specific about Ancestor. 

ParseComponent is probably more appropriate and consistent with the rest of the code base.

src/compile/data/parse.ts
```diff
@@ -160,8 +191,15 @@ Formula From Sort Array
 export function parseData(model: Model): DataComponent {
   let head = parseRoot(model, model.component.data.sources);
 
-  const outputNodes = model.component.data.outputNodes;
-  const outputNodeRefCounts = model.component.data.outputNodeRefCounts;
+  const {outputNodes, outputNodeRefCounts} = model.component.data;
+  const ancestorParse = model.parent ? model.parent.component.data.ancestorParse.clone() : new AncestorParse();
```
kanitw: I don't understand why do you have to clone here?

src/predicate.ts
```diff
@@ -32,9 +32,7 @@ export function isSelectionPredicate(predicate: LogicalOperand<Predicate>): pred
   return predicate && predicate['selection'];
 }
 
-export interface FieldEqualPredicate {
-  // TODO: support aggregate
-
+export interface FieldPredicateBase {
```
kanitw: This is a nice refactor, but seems irrelvant to this PR.

src/compile/data/formatparse.ts
```diff
@@ -52,77 +53,85 @@ export class ParseNode extends DataFlowNode {
...
-        keys(p).forEach(field => {
-          parse[field] = p[field];
-        });
+    return this.makeWithAncestors(parent, {}, implicit, ancestorParse);
+  }
+
+  /**
+   * Creates a parse node from two sets of explicit and implicit parse and updates ancestorParse.
+   */
+  public static makeWithAncestors(parent: DataFlowNode, explicit: Dict<string>, implicit: Dict<string>, ancestorParse: AncestorParse) {
```
kanitw: Why do you need to pass `ancestorParse` here?   Shouldn't it be a part of `parent`?  

(Sorry this is very complicated so it's hard to get why.)

src/compile/data/formatparse.ts
domoritz: ancestorParse is not a parse node, it's an object to track what has been parsed already

src/compile/data/index.ts
domoritz: This already existed: https://github.com/vega/vega-lite/blob/master/src/compile/data/index.ts#L48. Let's rename things in a new PR.

src/predicate.ts
domoritz: I had to use it at some point but then it ended up being automatically inferred by Typescript.

src/compile/data/formatparse.ts
```diff
@@ -52,77 +53,85 @@ export class ParseNode extends DataFlowNode {
...
-        return null;  // disable parsing completely
-      } else if (p) {
-        keys(p).forEach(field => {
-          parse[field] = p[field];
-        });
+    return this.makeWithAncestors(parent, {}, implicit, ancestorParse);
+  }
+
+  /**
+   * Creates a parse node from two sets of explicit and implicit parse and updates ancestorParse.
```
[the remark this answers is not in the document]
domoritz: explicit = data.format.parse
implicit = inferred from transforms and field types

src/compile/data/formatparse.ts
domoritz: No, the ancestor parse applies to the current model as well. I could simplify this and not make a copy, I guess.

src/compile/data/formatparse.ts
```diff
@@ -52,77 +53,85 @@ export class ParseNode extends DataFlowNode {
...
 
-    // We should not parse what has already been parsed in a parent
-    const modelParse = model.component.data.ancestorParse;
-    keys(modelParse).forEach(field => {
-      if (parse[field] !== modelParse[field]) {
-        log.warn(log.message.differentParse(field, parse[field], modelParse[field]));
+    for (const field of keys(explicit)) {
+      const parsedAs = ancestorParse.get(field);
+      if (parsedAs !== undefined) {
+        if (parsedAs === explicit[field]) {
```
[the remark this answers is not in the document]
domoritz: Done

src/compile/data/parse.ts
domoritz: I don't have to. I guess we could just reuse the same object and modify it as we go down the tree.

src/compile/split.ts
```diff
@@ -68,6 +68,17 @@ export class Split<T extends object> {
       this.set(key, s[key], true);
     }
   }
+
+  /**
+   * Merge split object this split object. Properties from the other split
```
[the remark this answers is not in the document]
domoritz: Done

src/compile/split.ts
```diff
@@ -68,6 +68,17 @@ export class Split<T extends object> {
...
+   * overwrite properties from this split.
+   */
+  public copyAll(other: Split<T>) {
```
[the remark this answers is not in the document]
domoritz: I looked at `copyKeyFromObject` and `copyKeyFromObject` in the same class. What do you suggest? 

Yes, I want side effects because I don't want to have to return a copy everywhere.

domoritz:
Hmm, the last commit broke some things. Looking into it.

src/compile/data/index.ts
kanitw: Your link only shows an old `Dict<string>`, but this class is introduced in this PR so naming it property should be in the scope of this PR.

src/compile/data/formatparse.ts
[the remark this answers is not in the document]
kanitw: The above part has

> // We always ignore derived fields even if they are implicitly defined because we expects users to create the right types.

The logic for this part is different, but doesn't have any comment

src/compile/data/formatparse.ts
kanitw: If you can, better do it now.

src/compile/data/formatparse.ts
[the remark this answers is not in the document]
kanitw: Better remove "two sets of". 

"two sets of explicit and implicit parse" sounds like you have 2 x explicit parse and 2 x implicit parse

src/compile/data/formatparse.ts
domoritz: Actually, that's incorrect when we have forks because every subtree needs to track parsed fields separately.

src/compile/data/formatparse.ts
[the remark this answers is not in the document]
domoritz: Because the logic isn't special here. If a field was parsed explicitly, we don't ignore what it has been parsed as.

src/compile/data/index.ts
domoritz: ParseComponent is not more appropriate, though because we never generate a parse object from this. I added a comment to make this clear.

src/compile/data/parse.ts
domoritz: I was wrong. We have to as explained earlier.

src/compile/data/formatparse.ts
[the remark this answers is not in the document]
kanitw: There are two cases -- each of them is different.  So better comment so it's easier for reader.

src/compile/data/parse.ts
kanitw: > We have to as explained earlier.

Can you explain again?  I don't know where you explain it as this PR is pretty complicated.

The rest of the "parse" codebase produce something as as part of component. 

Isn't it weird that this `ancestorParse` seems to be just floating within the parse codebase?

kanitw: At very least, this line deserves a better comment why this is being cloned.

src/compile/data/index.ts
```diff
@@ -43,7 +44,28 @@ export interface DataComponent {
...
   /**
-   * Parse properties passed down from ancestors.
+   * Parse properties passed down from ancestors. Helps us to keep track of what has been parsed or is derived.
    */
-  ancestorParse: Dict<string>;
+  ancestorParse?: AncestorParse;
+}
+
+/**
+ * Class to track interesting properties (see https://15721.courses.cs.cmu.edu/spring2016/papers/graefe-ieee1995.pdf)
```
kanitw: Is this paper really needed?  It's another 8 page people have to read. 

Better provide a TLDR explanation.

src/compile/data/index.ts
```diff
@@ -43,7 +44,28 @@ export interface DataComponent {
...
+ * about how fields have been parsed or whether they have been derived in a transforms. We use this to not parse the
+ * same field again (or differently).
+ */
+export class AncestorParse extends Split<Dict<string>> {
```
kanitw: If you want to rename it later, better rename it in this PR as this class is introduced in the PR.

src/compile/data/formatparse.ts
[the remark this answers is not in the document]
domoritz: ok

src/compile/data/formatparse.ts
kanitw: I don't understand your explanation here yet... 

Please explain "Why do you need to pass ancestorParse here?"

src/compile/data/index.ts
domoritz: I like the name.

src/compile/data/index.ts
domoritz: This goes into the depths of query optimization.

src/compile/data/formatparse.ts
domoritz: Oh, my response was to making a copy.

`ancestorParse` is currently not in nodes but passed as a separate object. I was thinking about adding this information to the nodes themselves but that makes a lot of the merging logic more complicated. I created an issue for this in https://github.com/vega/vega-lite/issues/3715

src/compile/data/parse.ts
```diff
@@ -48,56 +49,81 @@ function parseRoot(model: Model, sources: Dict<SourceNode>): DataFlowNode {
...
+            parse[filter.field] = 'date';
+          }
         }
-      }
+      });
 
-      if (keys(parse).length > 0) {
-        parent = new ParseNode(parent, parse);
+      if (keys(parse).length) {
+        parent = ParseNode.makeWithAncestors(parent, {}, parse, ancestorParse) || parent;
```
kanitw: I find the name `makeImplicit` a bit misleading since this parse from transform is "implicit" too, but doesn't call that `makeImplicit` method.  

May be rename `makeImplicit` to be`makeImplicitFromEncoding` and extract the logic above here to be `makeImplicitFromFilterTransform` (or just FromTransform).   This way you can make `makeWithAncestor` a `private` method too. 


One question that I have is don't we have to do the same thing with TimeUnit and Bin, in addition to Filter?

Issue #3694, opened 2018-05-03
Consider making ValueDef a generic again


kanitw:
So we can enforce data type of value def.  For example, `order`'s value can only be `null`. 

That said, we should see if this will be problematic for Altair.

kanitw:
This will allow us to have separate documentation section for different encoding channels. (Each of them can have separate documents.)

PR #3700, opened 2018-05-04
Adding support for value = "width" or "height"  for x/x2/y/y2  channels.

kanitw:
We don't have a way to say `{"field": {"group": "width"/ "height"}}`  in Vega-Lite yet. 

I propose that we support `{x/x2/y/y2: {"value": "min" / "max"}}` for this. 

Another alternative keyword could be "start" / "end"?  Any thoughts?

Fix #https://github.com/vega/vega-lite/issues/3022

cc: @jheer @arvind


kanitw:
image

```
{
  "layer": [
    {
      "mark": "bar",
      "encoding": {
        "x": {
          "timeUnit": "month",
          "field": "date",
          "type": "ordinal"

        },
        "y": {
          "aggregate": "mean",
          "field": "precipitation",
          "type": "quantitative"
        }
      }
    },
    {
      "mark": "rule",
      "encoding": {
        "y": {
          "value": "min"
        },
        "color": {"value": "red"},
        "size": {"value": 3}
      }
    }
  ]
}
```

domoritz:
I think min and max can be confusing when you reverse the quantitative axis, right?

kanitw:
For that case, wouldn't "start" and "end" also confusing?

domoritz:
Probably but less so. Maybe `top`, `bottom`, `left`, `right` are better (but can't be transposed).

kanitw:
How about `{"value": "width" / "height"}`?   (We actually only need one-side -- the other side `{value: 0}` is fine.)

domoritz:
```json
{
  "data": {
    "values": [
      {"a": "A","b": 28},
      {"a": "B","b": 55},
      {"a": "C","b": 43}
    ]
  },
  "encoding": {
    "x": {
      "field": "b",
      "type": "quantitative",
      "sort": "descending"
    }
  },
  "layer": [{
    "mark": "bar"
  }, {
    "mark": {
      "type": "text",
      "align": "right",
      "baseline": "middle",
      "dx": -4,
      "color": "white"
    },
    "encoding": {
      "text": {"field": "b", "type": "quantitative"},
      "x": {"value": "width"}
    }
  }]
}
```

image

domoritz:
https://github.com/vega/vega-lite/issues/3775 could be a good example.

kanitw:
layer_bar_annotation is the one that need this

kanitw:
I think documenting this is a bit tricky since we have have specific section for value def for different type of marks.  So let's punt on that until we do https://github.com/vega/vega-lite/issues/3694, but at least now we have an example that uses this!

examples/specs/layer_bar_annotations.vl.json
```diff
@@ -54,11 +54,14 @@
...
-          "align": "left",
-          "dx": 215,
-          "dy": -5
+          "align": "right",
+          "dx": -2,
+          "dy": -4
         },
         "encoding": {
+          "x": {
+            "value": "width"
```
domoritz: :+1:

PR #3701, opened 2018-05-04
Box Refactor

kanitw:


kanitw:
cc: @chanwutk -- you probably want to use `CompositeGenericMarkDef`  for ErrorBar too

Issue #3702, opened 2018-05-04
Apply selection to composite marks & mark overlays


kanitw:
Composite marks ([pending PR](https://github.com/vega/vega-lite/pull/3461)) and mark overlays (e.g., [area with line and point overlay](https://vega.github.io/vega-lite/docs/area.html#area-chart-with-overlaying-lines-and-point-markers)) generally produce multiple layers.  It's unclear if specified `selection` should be applied to all layers or only some of them.

It's possible that it's better to apply to only some of them and the rule could depend on selection type. 
For example, for line overlayed with points, "single" selection is probably applied to the overlayed points, but not the line (since we can't select individual points in the line).   However, for `boxplot`, it's unclear what's the appropriate behavior. 
  
Note:  Since we've not thoroughly determine how to do this for all composite marks/overlays yet, in our first release of composite mark will totally ignore selections as the behavior is not well defined.  (It is better to not support them yet instead of adding partially broken support and then introduce breaking changes in the future.) 

cc: @arvind, @domoritz, @jheer

PR #3703, opened 2018-05-04
For mark overlays, do not apply selection to the augmented layers yet.

kanitw:
(We will properly do that later in https://github.com/vega/vega-lite/issues/3702)

### DIFF
--- a/src/spec.ts
+++ b/src/spec.ts
@@ -560,6 +560,7 @@ function normalizePathOverlay(spec: NormalizedUnitSpec, config: Config = {}): No
   }
 
   const layer: NormalizedUnitSpec[] = [{
+    ...(selection ? {selection} : {}),
     // Do not include point / line overlay in the normalize spec
     mark: dropLineAndPoint({
       ...markDef,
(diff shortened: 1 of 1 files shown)


domoritz:
What problem does this fix?

kanitw:
We don't have a clear policy how to do it yet.  This resets the behavior to be "doing nothing".  The old code was weird anyway.

kanitw:
We don't have a clear policy how to do selection for overlay / composite yet #3702 .  This resets the behavior to be "doing nothing".  The old code was weird anyway.

domoritz:
Why was the old code weird? I don't think it's good to break selections for overlaid plots. Can you provide some examples where the old code was incorrect or what won't work anymore when we merge this?

kanitw:
The old code remove selection from the area layer and add selection to both point and line overlays.  Isn't that weird?

kanitw:
This PR makes the logic for overlay supposes to just add new overlay layers, but preseves selection to be in the original layer.  I think this is a reasonable standard logic.  We could have more sophisticated policy later.

PR #3713, opened 2018-05-06
Implement DSV data format. Fixes #2602

### LINKED ISSUE #2602: Add missing `delimiter` to DataFormat
(for DSV)

domoritz:


src/data.ts
```diff
@@ -17,16 +17,33 @@ export interface DataFormatBase {
...
   type?: 'csv' | 'tsv';
 }
 
+export interface DsvDataFormat extends DataFormatBase {
+  /**
+   * Type of input data: `"json"`, `"csv"`, `"tsv"`, `"dsv"`.
+   * The default format type is determined by the extension of the file URL.
+   * If no extension is detected, `"json"` will be used by default.
+   */
+  type?: 'dsv';
```
kanitw: Better move `type` to DataFormatBase and make it generic so there is only single source of description

domoritz: I tried this before but it didn't work with the schema generator.

PR #3714, opened 2018-05-06
Make sort logic reads from conditional fieldDef too

### LINKED ISSUE #3712: allow `sort` object within `field condition`
While it is possible in an encoding channel such as `color` with a simple field encoding or with a value condition to specify a `sort` object (sorting on another field), a `sort` object is no longer allowed when using `sort` within a `condition` which itself has a field definition. In the latter case, `sort` can only take the values `null`, `ascending` and `descending`.

This impacts the presentation of the legend, whose values can no longer be sorted by another field. As a matter of fact, the legend can also only be turned or off in case of condition, no customization allowed.

A workaround to this restriction: one can define a value condition together with a brushing selection, so as to "turn off" the undesired marks, e.g. by setting their color to a value of `gray`, instead of using the brush to colorize the selected marks, which would require a field definition and which would come with the above restriction. However, this workaround does not really seem intuitive.

kanitw:
Also Remove model.sort()

Fix #3712

### DIFF
--- a/src/compile/mark/mark.ts
+++ b/src/compile/mark/mark.ts
@@ -79,35 +79,37 @@ function parsePathMark(model: UnitModel) {
 }
 
 export function getSort(model: UnitModel) {
-  const order = model.encoding.order;
+  const {encoding, stack, mark, markDef} = model;
+  const order = encoding.order;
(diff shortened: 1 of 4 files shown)


g3o2:
examples added to the issue

PR #3717, opened 2018-05-07
Add bullet chart example

kanitw:
~~TODO: still need to customize facet header size before merging~~


kanitw:
Note that there is a bug if we change `{..., "axis": {"title": ""}}` to `{..., "title": ""}` too.

kanitw:
Note that we don't support headerBand yet -- otherwise I would have moved the label a bit below.

PR #3723, opened 2018-05-09
Make fieldDef.title produces explicit title for axis/legend

kanitw:
(Fixing old comment in https://github.com/vega/vega-lite/pull/3717#issuecomment-387136239)


domoritz:
Lgtm

PR #3728, opened 2018-05-09
Flatten domain sort. Fixes #3727

### LINKED ISSUE #3727: Can't sort by subobject field
Can't sort by field stored in properties of geojson object.
Vega-light processes fields with dot in name but not in "sort field definition object"

### Simplified example to reproduce situation :

```javascript
{
    "$schema": "https://vega.github.io/schema/vega-lite/v2.4.1.json",
    "data": {
        "values": [
            {"name": "a0", "a": {"i": 0}}, 
            {"name": "a1", "a": {"i": 1}},
            {"name": "a2", "a": {"i": 2}}
        ]
    },
    "mark": "bar",
    "encoding": {
        "x": {"field": "a.i", "type": "quantitative"},
        "y": {"field": "name", "type": "ordinal",
            "sort": {
                "order": "descending",
                "op": "sum",
                "field": "a.i"//  "a\\.i" will work
            }
        }
    }
}
```
this example on [bl.ocks.org](https://bl.ocks.org/iliatimofeev/309b8529c14886fed6f9e05d124a99eb/4f257d0fee1ce1d56ff95893fdec6f3caf355423)

domoritz:
Fixes #3727

### DIFF
--- a/src/compile/data/formatparse.ts
+++ b/src/compile/data/formatparse.ts
@@ -1,12 +1,12 @@
-import {isString} from 'util';
-import {isNumber, toSet} from 'vega-util';
+import {isNumber, isString, toSet} from 'vega-util';
 import {AncestorParse} from '.';
 import {isCountingAggregateOp} from '../../aggregate';
 import {DateTime, isDateTime} from '../../datetime';
-import {isNumberFieldDef, isTimeFieldDef} from '../../fielddef';
+import {isNumberFieldDef, isScaleFieldDef, isTimeFieldDef} from '../../fielddef';
 import * as log from '../../log';
 import {forEachLeaf} from '../../logical';
 import {isFieldEqualPredicate, isFieldOneOfPredicate, isFieldPredicate, isFieldRangePredicate} from '../../predicate';
+import {isSortField} from '../../sort';
 import {FilterTransform} from '../../transform';
 import {accessPathDepth, accessPathWithDatum, Dict, duplicate, keys, removePathFromField, StringSet} from '../../util';
 import {VgFormulaTransform} from '../../vega.schema';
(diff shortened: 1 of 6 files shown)


src/compile/data/formatparse.ts
```diff
@@ -122,14 +122,20 @@ export class ParseNode extends DataFlowNode {
         if (isTimeFieldDef(fieldDef)) {
           implicit[fieldDef.field] = 'date';
         } else if (isNumberFieldDef(fieldDef)) {
-          if (isCountingAggregateOp(fieldDef.aggregate)) {
-            return;
+          if (!isCountingAggregateOp(fieldDef.aggregate)) {
+            implicit[fieldDef.field] = 'number';
```
kanitw: 1) Why is this related to sort?

2) What about distinct and some other operations that do not require numeric input?

src/compile/data/formatparse.ts
```diff
@@ -141,12 +147,12 @@ export class ParseNode extends DataFlowNode {
    * Creates a parse node from "explicit" parse and "implicit" parse and updates ancestorParse.
    */
   private static makeWithAncestors(parent: DataFlowNode, explicit: Dict<string>, implicit: Dict<string>, ancestorParse: AncestorParse) {
-    // We should not parse what has already been parsed in a parent (explicitly or implicitly) or what has been derived (maked as "derived").
+    // We should not parse what has already been parsed in a parent (explicitly or implicitly) or what has been derived (maked as "derived") and we don't need to flatten again.
```
kanitw: this sentence is now super complicated.  I don't understand how "or" and "and" are being grouped.

src/compile/data/formatparse.ts
```diff
@@ -141,12 +147,12 @@ export class ParseNode extends DataFlowNode {
...
     for (const field of keys(implicit)) {
       const parsedAs = ancestorParse.getWithExplicit(field);
       if (parsedAs.value !== undefined) {
         // We always ignore derived fields even if they are implicitly defined because we expect users to create the right types.
-        if (parsedAs.explicit || parsedAs.value === implicit[field] || parsedAs.value === 'derived') {
+        if (parsedAs.explicit || parsedAs.value === implicit[field] || parsedAs.value === 'derived' || implicit[field] === 'flatten') {
```
kanitw: Why do you need to delete flatten?

src/compile/data/formatparse.ts
domoritz: I wanted to make it consistent with the code below and also I don't like to use return when we might switch to a proper for loop in the future. It's a disaster bound to happen.

Re distinct. Yes, that's incorrect but unrelated to this PR.

src/compile/data/formatparse.ts
domoritz: If we already parse a field, there is no need to flatten (flatten is just using a calculate with no change to the data).

src/compile/data/formatparse.ts
domoritz: Agreed and fixed

domoritz:
Are we good here?

src/compile/data/formatparse.ts
kanitw: > If we already parse a field

Isn't this checked from `parseAs` -- why is it related to the new `|| implicit[field] === 'flatten'`?

domoritz: The field was parsed before. So now the question is, do we parse it again as something different (e.g. string to number). If the new parse is flatten, we don't need to do anything.

PR #3736, opened 2018-05-10
Add Index for supported Scale Types given encoding channels

### LINKED ISSUE #3725: Create method for Supported Scale Types
With Voyager now supporting custom encoding, users can change Scale Types. It would be nice if Vega-Lite could have a `supportedScaleType` method that would determine the possible valid scale types given the Field Definition and encoding channel. This would avoid having tedious if/else logic trees within libraries that use Vega-Lite, like the code snippet below in Voyager.

```
function getSupportedScaleTypes(shelfId: ShelfId, fieldDef: ShelfFieldDef): string[] {
  switch (fieldDef.type) {
    case ExpandedType.QUANTITATIVE:
      if (contains([Channel.X, Channel.Y], shelfId.channel)) {
        return ["linear", "log", "pow", "sqrt"];
      } else if (shelfId.channel === Channel.COLOR) {
        return ["linear", "pow", "sqrt", "log", "sequential"];
      } else if (shelfId.channel === Channel.SIZE) {
        return ["linear", "pow", "sqrt", "log"];
      }
    case ExpandedType.ORDINAL:
    case ExpandedType.NOMINAL:
      if (contains([Channel.X, Channel.Y], shelfId.channel)) {
        return ["point", "band"];
      } else if (shelfId.channel === Channel.COLOR) {
        return ["ordinal"];
      } else if (shelfId.channel === Channel.SIZE) {
        return ["point", "band"];
      }
    case ExpandedType.TEMPORAL:
      if (contains([Channel.X, Channel.Y], shelfId.channel)) {
        return ["time", "utc"];
      } else if (shelfId.channel === Channel.COLOR) {
        return ["time", "utc", "sequential"];
      } else if (shelfId.channel === Channel.SIZE) {
        return ["time", "utc"];
      }
    default:
      return [];
  }
}
```

Both CompassQL and Voyager can benefit from this method!

ssharif6:
Fix #3725 

Expose method `getSupportedScaleType` that returns the valid scaleTypes given an encoding channel, FieldDef type, and optional bin parameter. Function calls generated index for pre-computed valid scale types.

### DIFF
--- a/src/scale.ts
+++ b/src/scale.ts
@@ -1,7 +1,9 @@
 import {toSet} from 'vega-util';
-import {Channel, isColorChannel} from './channel';
+import {BinParams} from './bin';
+import {Channel, CHANNELS, isColorChannel} from './channel';
 import {DateTime} from './datetime';
 import * as log from './log';
+import {Type, TYPE_INDEX} from './type';
 import {contains, Flag, flagKeys, keys} from './util';
 import {ScaleInterpolate, ScaleInterpolateParams} from './vega.schema';
 
@@ -587,6 +589,8 @@ const {type, domain, range, rangeStep, scheme, ...NON_TYPE_DOMAIN_RANGE_VEGA_SCA
 
 export const NON_TYPE_DOMAIN_RANGE_VEGA_SCALE_PROPERTIES = flagKeys(NON_TYPE_DOMAIN_RANGE_VEGA_SCALE_PROPERTY_INDEX);
 
+export const SCALE_TYPE_INDEX = generateScaleTypeIndex();
+
 export function scaleTypeSupportProperty(scaleType: ScaleType, propName: keyof Scale) {
   switch (propName) {
     case 'type':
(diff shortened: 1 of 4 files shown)


src/scale.ts
```diff
@@ -660,6 +664,21 @@ export function channelScalePropertyIncompatability(channel: Channel, propName:
   throw new Error(`Invalid scale property "${propName}".`);
 }
 
+export function fieldDefMatchScaleType(specifiedType: ScaleType, fieldDefType: Type, bin?: boolean|BinParams):boolean {
```
kanitw: rename to `scaleTypeSupportDataType`  (there is no "fieldDef" anymore) and make `bin` required

src/scale.ts
```diff
@@ -681,3 +700,43 @@ export function channelSupportScaleType(channel: Channel, scaleType: ScaleType):
...
+
+// generates ScaleTypeIndex where keys are encoding channels and values are list of valid ScaleTypes
+export function generateScaleTypeIndex() {
+  const index: ScaleTypeIndex = {};
+  for (const channel of CHANNELS) {
+    for (const fieldDefType of keys(TYPE_INDEX)) {
+      for (const scaleType of SCALE_TYPES) {
+        const key = generateScaleTypeIndexKey(channel, fieldDefType);
+        if (channelSupportScaleType(channel, scaleType) && fieldDefMatchScaleType(scaleType, fieldDefType)) {
+          index[key] ? index[key].push(scaleType) : index[key] = [scaleType];
```
kanitw: I don't like standalone ternary operator.  This is better.

```
index[key] = index[key] || [];
index[key].push(scaleType);
```

src/scale.ts
```diff
@@ -681,3 +700,43 @@ export function channelSupportScaleType(channel: Channel, scaleType: ScaleType):
...
+      if (channelSupportScaleType(channel, scaleType) && fieldDefMatchScaleType(scaleType, Type.QUANTITATIVE, true)) {
+        index[key] ? index[key].push(scaleType) : index[key] = [scaleType];
+      }
+    }
+  }
+
+  return index;
+}
+
+export function generateScaleTypeIndexKey(channel: Channel, fieldDefType: Type, bin?: boolean) {
```
kanitw: make bin required so it's more readable.

src/scale.ts
```diff
@@ -681,3 +700,43 @@ export function channelSupportScaleType(channel: Channel, scaleType: ScaleType):
...
+        const key = generateScaleTypeIndexKey(channel, fieldDefType);
+        if (channelSupportScaleType(channel, scaleType) && fieldDefMatchScaleType(scaleType, fieldDefType)) {
+          index[key] ? index[key].push(scaleType) : index[key] = [scaleType];
+        }
+      }
+    }
+  }
+
+  // add quantitative binned keys to index
+  for (const channel of CHANNELS) {
```
kanitw: Instead of writing it twice, just add another loop   

```
for (const bin of [false, true]) {
   ...
}
```

test/scale.test.ts
```diff
@@ -75,4 +83,70 @@ describe('scale', () => {
       }
     });
   });
+
+  describe('generateScaleTypeIndexKey', () => {
```
kanitw: These two methods are internal implementation of `getSupportedScaleType`, but not what users will use directly. 

Better test `getSupportedScaleType`  and make these two methods private (Don't export them.) 

In a way, even if we change the implementation, the test should preserve the result of `getSupportedScaleType`.

PR #3741, opened 2018-05-14
Rename `SortField` => `EncodingSortField`, `WindowSortField` => `SortField`

kanitw:
Rationale: WindowSortField will be used in Stack too and this is like a more general version of SortField while the encoding one is specific to one place.  

(Note that I opt for this simpler change that we discussed earlier as SortField are used in many places and they are not exactly the same thing anyway.)

### DIFF
--- a/_data/link.yml
+++ b/_data/link.yml
@@ -62,7 +62,7 @@ Day:
 # Transform - Window
 WindowFieldDef:
   link: window.html#field-def
-WindowSortField:
+SortField:
   link: window.html#sort-field-def
 WindowOnlyOp:
   link: window.html#ops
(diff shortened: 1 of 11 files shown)


invokesus:
Maybe update the documentation too?
- [window.md](https://github.com/vega/vega-lite/blob/master/site/docs/transform/window.md#window-sort-field-definition)
- [link.yml](https://github.com/vega/vega-lite/blob/40f18b9c594b32d991b3cd8171d9dacaa0ada8d4/_data/link.yml#L65)

kanitw:
Actually this is quite tricky once we write the documentation as there will be two "sort field" in the documentation and TransformSortField might not be the most descriptive name.  -- need to think a bit more

jakevdp:
Just noticed this when updating Altair to VL 2.5. Since Python class names are generated based on definition names in the schema, this is a breaking change for the Python API.

Issue #3724, opened 2018-05-09
Examples on the website are not interactive


domoritz:
e.g. https://vega.github.io/vega-lite/examples/selection_layer_bar_month.html

jheer:
Is this running Vega 3.3.1? If so, perhaps try bumping your bundle to 4.0.0-rc.1. Or alternatively lock vega-scenegraph to 2.4.2 and see if that helps? I'm worried there may be an inadvertent version incompatibility here...

domoritz:
According to the yarn.lock, we use Vega scenegraph 2.5. Could that be a problem? I can't test it right now unfortunately.

domoritz:
I'm having a hard time figuring out exactly what's going wrong but with an older version of all vega libraries, the examples work so I suspect that that's the issue.

domoritz:
Fixed with https://github.com/vega/vega-lite/commit/40f18b9c594b32d991b3cd8171d9dacaa0ada8d4

PR #3743, opened 2018-05-14
Update to Vega 4

### LINKED ISSUE #3719: Update to Vega 4
This is only a minor version change in Vega-Lite since the schema doesn't change much.

### LINKED ISSUE #3739: compiled vega source contains outdated `$schema` property
currently contains:

`"$schema": "https://vega.github.io/schema/vega/v3.0.json"`

instead of the more general:

`"$schema": "https://vega.github.io/schema/vega/v3.json"`

domoritz:
Fixes #3739, fixes #3719


domoritz:
I was hoping it would fix #3724 but then found the problem somewhere else. Since vega is just a dev dependency, I think we could update but not yet switch the version of the generated specs yet.

domoritz:
Updated the PR. I think this is good to merge now.

PR #3747, opened 2018-05-14
Fix default BoxPlot title

kanitw:


kanitw:
@chanwutk  wonder if you can copy one of your error-bar test and modify it for boxplot test here? (Shouldn't take too much time - I hope)

chanwutk:
Which errorbar test do you want me to copy?

kanitw:
I assume you have one test for the title?

chanwutk:
I don't have a test for titles, but I can create one.

kanitw:
Let's close this since the title fix is in the `makeCompositeAggregatePartFactory` method that we gonna re-use.

PR #3749, opened 2018-05-15
Adding Waterfall Chart

chanwutk:


chanwutk:
@kanitw I saw that there is already a waterfall chart in the vega-lite example page. Should I continue making this example?

examples/specs/layer_bar_waterfall_chart.vl.json
```diff
@@ -0,0 +1,67 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "data": {
+    "values": [
+      {"label": "Start", "amount": 4000, "order": 1},
```
kanitw: Instead of including order in the data, use window's `row_number` operation?

examples/specs/layer_bar_waterfall_chart.vl.json
```diff
@@ -0,0 +1,67 @@
...
+    },
+    {
+      "calculate": "datum.label == 'End' || datum.label == 'Start' ? 'Start/End' : datum.amount >= 0 ? 'increasing' : 'decreasing'",
+      "as": "color"
+    },
+    {"calculate": "datum.sum - 100", "as": "bottomline"}
+  ],
+  "layer": [
+    {
+      "mark": "bar",
```
kanitw: Why is there only one layer? Add labels like the community example in the gallery?

kanitw:
> @kanitw I saw that there is already a waterfall chart in the vega-lite example page. Should I continue making this example?

Yes, the other example didn't use window transform.  Also, I think community example doesn't count.  :)

examples/specs/window_sum_waterfall_chart.vl.json
```diff
@@ -0,0 +1,104 @@
...
+      {"label": "Oct", "amount": 1205},
+      {"label": "Nov", "amount": -1388},
+      {"label": "Dec", "amount": 1492},
+      {"label": "End", "amount": 0}
+    ]
+  },
+  "width": 700,
+  "height": 450,
+  "transform": [
+    {"window": [{"op": "row_number", "as": "row"}]},
```
kanitw: Do we really need to run window twice here?  (Wouldn't the default sort order be by row number?)

examples/specs/window_sum_waterfall_chart.vl.json
```diff
@@ -0,0 +1,104 @@
...
+      "as": "sum_start_end"
+    }
+  ],
+  "encoding": {
+    "y": {
+      "field": "previous_amount",
+      "type": "quantitative",
+      "axis": {"title": "Amount"}
+    },
+    "y2": {"field": "sum", "type": "quantitative"},
```
kanitw: Why are y and y2 shared?  I see that you override y in many of them and y2 are ignored by the text layers. 

Please avoid overriding encoding as it makes it harder to read examples.

examples/specs/window_sum_waterfall_chart.vl.json
```diff
@@ -0,0 +1,104 @@
...
+          "scale": {"range": ["#f7e0b6", "#f78a64", "#93c4aa"]}
+        }
+      }
+    },
+    {
+      "mark": {"type": "text", "dy": -5},
+      "encoding": {
+        "y": {"field": "sum_inc", "type": "quantitative"},
+        "text": {"field": "sum_inc", "type": "nominal"},
+        "color": {"value": "#234f39"}
```
kanitw: Can set these colors in mark property -- no need to use color.value

examples/specs/window_sum_waterfall_chart.vl.json
```diff
@@ -0,0 +1,104 @@
...
+    {
+      "calculate": "datum.Color == 'Start/End' ? datum.sum : ''",
+      "as": "sum_start_end"
+    }
+  ],
+  "encoding": {
+    "y": {
+      "field": "previous_amount",
+      "type": "quantitative",
+      "axis": {"title": "Amount"}
```
kanitw: let's replace `"axis": {"title": "Amount"}` with `"title": "Amount"`

examples/specs/window_sum_waterfall_chart.vl.json
```diff
@@ -0,0 +1,104 @@
...
+      "sort": {"field": "row", "op": "min"},
+      "axis": {"title": "Months"}
+    }
+  },
+  "layer": [
+    {
+      "mark": "bar",
+      "encoding": {
+        "color": {
+          "field": "Color",
```
kanitw: Instead of deriving a color field, I think it might be better to use conditions to set the color directly.

domoritz:
What does it looks like if you make all the text just black?

chanwutk:
screen shot 2018-06-15 at 11 18 06
It would look like this. 
By the way, I added ticks between each bar and add the amount inside each bar. I also removed the label on the right because I think it would be cleaner to just show the amount inside the bar.

For the Start and End, because the value above the bar is the same as the one inside, should I remove it?

examples/specs/window_sum_waterfall_chart.vl.json
```diff
@@ -0,0 +1,133 @@
...
+        }
+      }
+    },
+    {
+      "mark": {
+        "type": "tick",
+        "color": "#404040",
+        "thickness": 1.5,
+        "opacity": 1,
+        "xOffset": 28.5
```
kanitw: Why is this offset `28.5`?

chanwutk: It's the value that the ticks look the most centered to me. Is there other way that I can set the ticks, so that they are at the middle between 2 bars?

examples/specs/window_sum_waterfall_chart.vl.json
```diff
@@ -96,8 +111,8 @@
         "color": "#404040"
       },
       "encoding": {
-        "y": {"field": "sum_inc", "type": "quantitative"},
```
kanitw: Not sure why do you have to remove all whitespaces after comma.  It's actually a good thing to have them.

chanwutk: I'm pretty sure that all I did was changing tick to rule, adding window, and adding offsets...

kanitw: Hmm, some autoformatter must be acting weird.

chanwutk: I figured it out. I was trying to use the old editor, and it gave me that format. Now, I use the new editor, and it works fine.

Issue #3756, opened 2018-05-16
Support size for band scale


kanitw:
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A simple bar chart with ranged data (aka Gantt Chart).",
  "data": {
    "values": [
      {"task": "A","start": 1, "end": 3},
      {"task": "B","start": 3, "end": 8},
      {"task": "C","start": 8, "end": 10}
    ]
  },
  "mark": "bar",
  "encoding": {
    "y": {"field": "task", "type": "ordinal"},
    "x": {"field": "start", "type": "quantitative"},
    "x2": {"field": "end", "type": "quantitative"},
    "size": {"field": "start", "type": "quantitative"}
  }
}
```

![image](https://user-images.githubusercontent.com/111269/40135250-b6c16d56-58f9-11e8-98b9-812032e6f17d.png)

## Workaround

For now, users can set the y scale to point and it will work

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A simple bar chart with ranged data (aka Gantt Chart).",
  "data": {
    "values": [
      {"task": "A","start": 1, "end": 3},
      {"task": "B","start": 3, "end": 8},
      {"task": "C","start": 8, "end": 10}
    ]
  },
  "mark": "bar",
  "encoding": {
    "y": {"field": "task", "type": "ordinal", "scale": {"type": "point"}},
    "x": {"field": "start", "type": "quantitative"},
    "x2": {"field": "end", "type": "quantitative"},
    "size": {"field": "start", "type": "quantitative"}
  }
}

![image](https://user-images.githubusercontent.com/111269/40135285-cf59dac4-58f9-11e8-80be-a3a7fe9bc6fe.png)

```

PR #3762, opened 2018-05-18
Improve log that size on band scale is not supported, better use point

kanitw:
https://github.com/vega/vega-lite/issues/3756 -- since we won't have time to fix that -- better make the warning suggests a good workaround.


kanitw:
Actually -- fixing this is quite easy.  I'll just fix it.

Issue #1734, opened 2016-11-29
Allow specifying specific order of stack in stacked charts


kanitw:
We currently do not have a way to specify manual order of stacking group (e.g., `color` or `detail` channel) in stacked charts.  

Note: As seen in https://github.com/vega/vega-lite/issues/1732#issuecomment-263454554, users might expect that color's scale domain also affect how stack is ordered. 

For example, in the following spec. 

```
{
  "data": {
    "url": "https://raw.githubusercontent.com/dhimmel/biorxiv-licenses/4ad9659e4fc823f4c491e13be4505248df5c1ab6/figure/license-vs-time/vega-lite-data.json"
  },
  "config": {
    "mark": {
      "stacked": "normalize"
    }
  },
  "encoding": {
    "color": {
      "field": "license",
      "scale": {
        "domain": [
          "CC BY",
          "CC BY-ND",
          "CC BY-NC",
          "CC BY-NC-ND",
          "None"
        ],
        "type": "ordinal"
      },
      "type": "nominal"
    },
    "x": {
      "axis": {
        "axisWidth": 0.0,
        "format": "%Y",
        "labelAngle": 0.0,
        "tickSize": 0.0
      },
      "field": "date",
      "scale": {
        "nice": "month"
      },
      "timeUnit": "yearmonth",
      "title": " ",
      "type": "temporal"
    },
    "y": {
      "aggregate": "count",
      "axis": {
        "axisWidth": 0.0,
        "grid": false,
        "labels": false,
        "ticks": 0.0
      },
      "field": "*",
      "title": "Proportion of Preprints",
      "type": "quantitative"
    }
  },
  "mark": "area"
}
```

The scale domain only determines how the domain values are mapped to first 5 color in the default color palette, but does not affect order of the stacking.

![vega](https://cloud.githubusercontent.com/assets/1117703/20693613/91366726-b5ad-11e6-9f33-d0e6e7c5c2fe.png)

TODOs:
- [ ] design how we support manual order for stacking 
   - Under the hood, one possible way to do this is to add a formula transform that add index for each domain value and use the index as `sortby` of `stack` transform. 
   - [ ] syntax

- [ ] decide if we want to use `color`/`size`/`detail`'s `sort` by default. 

cc: @dhimmel

kanitw:
A workaround for now would be to use `calculate` to derive a new field (`license_index`) and map the field to `order` channel.  

```js
{
  "data": {
[...]

PR #3771, opened 2018-05-18
Add Stack Transform

### LINKED ISSUE #3311: Document & unhide stack transform
- [ ] improve diverging stack bar example

Recently, I wanted to reuse this [example](https://vega.github.io/vega-lite/examples/bar_diverging_stacked.html) only to realize that its dataset appears to have undergone some preprocessing outside of vega-lite, thus rendering the example theoretical in a pure vega-lite context. 

The fields in question are `percentage_start` and `percentage_end`. I can see how to calculate those two fields in vega using the `stack` transform. 

But is it currently feasible at all to calculate the said fields using only vega-lite? Alternatively, would it be possible to showcase this example using vega-lite's `layer` and `stack` instead?

```javascript
{
        "question": "Question 1",
        "type": "Strongly disagree",
        "value": 24,
        "percentage": 0.7,
        "percentage_start": -19.1,
        "percentage_end": -18.4
}
```

invokesus:
Fix #3311 

To be done:
- [ ] ~~Documentation~~
- [ ] Examples
  - [ ] Basic Example 
  - [ ] Diverging Stacked Bar
  - [ ] Mosaic / Marimekko Chart
- [ ] Test coverage
- [x] Let's add @hide to the description of `StackTransform` to hide it from the schema for now (and re-compile schema) so we can merge this after you address the unaddressed comments.
- [ ] For examples you wanna add, please add `_future` suffix to the example files so they can be included even though stack transform wouldn't be in the schema yet.

### DIFF
--- a/src/compile/data/parse.ts
+++ b/src/compile/data/parse.ts
@@ -87,7 +87,11 @@ export function parseTransformArray(head: DataFlowNode, model: Model, ancestorPa
         ancestorParse.set(field, 'derived', false);
       }
     } else if (isStack(t)) {
-      const stack= head = StackNode.makeFromTransform(head, t);
+      const stack = head = StackNode.makeFromTransform(head, t);
+
+      for (const field of keys(stack.producedFields())) {
+        ancestorParse.set(field, 'derived', false);
+      }
     } else {
       log.warn(log.message.invalidTransformIgnored(t));
       return;
--- a/src/compile/data/parse.ts
+++ b/src/compile/data/parse.ts
@@ -225,7 +225,7 @@ export function parseData(model: Model): DataComponent {
       }
     }
 
-    head = StackNode.make(head, model) || head;
+    head = StackNode.makeFromEncoding(head, model) || head;
   }
 
   if (isUnitModel(model)) {
--- a/src/compile/data/parse.ts
+++ b/src/compile/data/parse.ts
@@ -1,6 +1,6 @@
 import {MAIN, RAW} from '../../data';
 import * as log from '../../log';
-import {isAggregate, isBin, isCalculate, isFilter, isLookup, isTimeUnit, isWindow} from '../../transform';
+import {isAggregate, isBin, isCalculate, isFilter, isLookup, isStack, isTimeUnit, isWindow} from '../../transform';
 import {Dict, keys} from '../../util';
 import {isFacetModel, isLayerModel, isUnitModel, Model} from '../model';
 import {requiresSelectionId} from '../selection/selection';
@@ -86,6 +86,8 @@ export function parseTransformArray(head: DataFlowNode, model: Model, ancestorPa
       for (const field of keys(window.producedFields())) {
         ancestorParse.set(field, 'derived', false);
       }
+    } else if (isStack(t)) {
+      const stack= head = StackNode.makeFromTransform(head, t);
     } else {
       log.warn(log.message.invalidTransformIgnored(t));
       return;
--- a/src/compile/data/stack.ts
+++ b/src/compile/data/stack.ts
@@ -23,7 +23,7 @@ function getStackByFields(model: UnitModel): string[] {
 
 export interface StackComponent {
   // TODO it would be cleaner to compose the two interfaces to create this one.
-  // TODO remove Used in *comments
+  // TODO Rename make to makeFromTransform
   /**
    * Faceted field. Used in makeFromEncoding.
    */
@@ -67,9 +67,9 @@ export interface StackComponent {
    */
   groupby?: string[];
   /**
-   * Output field names of each stack field. TODO write about the defaults
+   * Output field names of each stack field.
    */
-  as?: string | string[];
+  as: string[];
 
 }
 
@@ -92,8 +92,16 @@ export class StackNode extends DataFlowNode {
 
   public static makeFromTransform(parent: DataFlowNode, model: StackTransform) {
     const offset = model.offset || 'zero';
-    const sort = model.sort || {'field': model.stack, 'order': 'ascending'}; // refactor possible?
-
+    const sort = model.sort || {'field': model.stack, 'order': 'ascending'};
+    const as = model.as;
+    let normalizedAs: Array<string> | undefined;
+    if (isAsValidArray(as)) {
+      normalizedAs = as;
+    } else if(typeof as === 'string') {
+      normalizedAs = [as, as + '_end'];
+    } else {
+      normalizedAs = [model.stack + '_start', model.stack + '_end'];
+    }
 
 
     return new StackNode (parent, {
@@ -101,8 +109,9 @@ export class StackNode extends DataFlowNode {
       groupby: model.groupby,
       offset,
       sort,
-      as: model.as
+      as: normalizedAs
     });
+
   }
   public static make(parent: DataFlowNode, model: UnitModel) {
 
@@ -132,7 +141,13 @@ export class StackNode extends DataFlowNode {
         return s;
       }, {field:[], order: []});
     }
-
+    // Refactored to add as in the make phase so that we can get producedFields
+    // from the as property
+    const field = model.vgField(stackProperties.fieldChannel);
+    const as = [
+      field + '_start',
+      field + '_end'
+    ];
     return new StackNode(parent, {
       dimensionFieldDef,
       field: model.vgField(stackProperties.fieldChannel),
@@ -141,6 +156,7 @@ export class StackNode extends DataFlowNode {
       sort,
       offset: stackProperties.offset,
       impute: stackProperties.impute,
+      as
     });
   }
 
@@ -166,11 +182,10 @@ export class StackNode extends DataFlowNode {
   }
 
   public producedFields() {
-    const out = {};
-
-    out[this._stack.field + '_start'] = true;
-    out[this._stack.field + '_end'] = true;
-
+    const out = this._stack.as.reduce((result, item) => {
+      result[item] = true;
+      return result;
+    }, {});
     return out;
   }
 
@@ -243,15 +258,7 @@ export class StackNode extends DataFlowNode {
       return transform;
     } else {
       const {stack, groupby, offset, sort, as} = this._stack;
-      let normalizedAs: Array<string>;
-      if (isAsValidArray(as)) {
-        normalizedAs = as;
-      } else if(typeof as === 'string') {
-        normalizedAs = [as, as + '_end'];
-      } else {
-        normalizedAs = [this._stack + '_start', this._stack + '_end'];
-      }
-      return [{type: 'stack',groupby,field: stack, sort, as: normalizedAs, offset}];
+      return [{type: 'stack', groupby,field: stack, sort, as, offset}];
     }
   }
 }
--- a/src/compile/data/stack.ts
+++ b/src/compile/data/stack.ts
@@ -172,11 +172,10 @@ export class StackNode extends DataFlowNode {
   }
 
   public producedFields() {
-    const out = this._stack.as.reduce((result, item) => {
+    return this._stack.as.reduce((result, item) => {
       result[item] = true;
       return result;
     }, {});
-    return out;
   }
 
   private getGroupbyFields() {
--- a/src/compile/data/stack.ts
+++ b/src/compile/data/stack.ts
@@ -3,7 +3,7 @@ import {FieldDef, isFieldDef, vgField} from '../../fielddef';
 import {StackOffset} from '../../stack';
 import {StackTransform} from '../../transform';
 import {duplicate} from '../../util';
-import {VgSort, VgTransform} from '../../vega.schema';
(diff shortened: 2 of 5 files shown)


src/compile/data/parse.ts
```diff
@@ -86,6 +86,12 @@ export function parseTransformArray(head: DataFlowNode, model: Model, ancestorPa
       for (const field of keys(window.producedFields())) {
         ancestorParse.set(field, 'derived', false);
       }
+    } else if (isStack(t)) {
+      const stack = head = StackNode.makeFromTransform(head, t);
+
+      for (const field of keys(stack.producedFields())) {
```
kanitw: This code is called repeatedly through this method.  In a separate PR, we should refactor this part https://github.com/vega/vega-lite/issues/3772.

invokesus:
> Why does this address #1734?
Sorry, I misunderstood the issue. Removed it from the checklist.

src/compile/data/stack.ts
```diff
@@ -21,35 +22,59 @@ function getStackByFields(model: UnitModel): string[] {
...
 
-  dimensionFieldDef: FieldDef<string>;
+  dimensionFieldDef?: FieldDef<string>;
 
   /**
-   * Stack measure's field
+   * Stack measure's field. Used in makeFromEncoding.
    */
-  field: string;
+  field?: string;
```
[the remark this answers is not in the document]
invokesus: Will keep either `stack` or `field` and make that compulsory.

src/compile/data/stack.ts
```diff
@@ -21,35 +22,59 @@ function getStackByFields(model: UnitModel): string[] {
...
   /**
-   * Whether to impute the data before stacking.
+   * Whether to impute the data before stacking. Used only in makeFromEncoding.
+   */
+  impute?: boolean;
+
+  /**
+   * Stack measure's field. Used in makeFromTransform
+   */
+  stack?: string;
```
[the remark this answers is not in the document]
kanitw: We can also name it `stackField`. 

(We later do rename it in the assemble method anyway):

```
const {facetby, field: stackField, dimensionFieldDef, impute, offset, sort, stackby} = this._stack;
```

src/transform.ts
```diff
@@ -102,6 +102,32 @@ export interface AggregatedFieldDef {
...
+   * Mode for stacking marks. Default is 'zero'
+   */
+  offset?: 'zero' | 'center' | 'normalize';
+  /**
+   * Field that determines the order of leves in the stacked charts.
+   */
+  sort?: SortField ;
+  /**
+   * Output field names.
+   * y2 or x2 is optional if only one name is provided the end will be “<name>_end”
```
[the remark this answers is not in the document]
invokesus: > If a single string is provided, the end field will be "_end".

Right now if a single string(eg. "values") is provided the end field is "values_end".

src/compile/data/stack.ts
```diff
@@ -90,30 +86,30 @@ export class StackNode extends DataFlowNode {
...
 
-  public static makeFromTransform(parent: DataFlowNode, model: StackTransform) {
-    const offset = model.offset || 'zero';
-    const sort = model.sort || {'field': model.stack, 'order': 'ascending'};
-    const as = model.as;
-    let normalizedAs: Array<string> | undefined;
+  public static makeFromTransform(parent: DataFlowNode, stackTransform: StackTransform) {
+
+    const {stack, groupby, as, offset='zero'} = stackTransform;
+    const sort = stackTransform.sort || {'field': stack, 'order': 'ascending'};
```
kanitw: I'm curious if this default is needed?

src/compile/data/stack.ts
```diff
@@ -211,58 +204,54 @@ export class StackNode extends DataFlowNode {
...
+      transform.push({
+        type: 'impute',
+        field,
+        groupby: stackby,
+        key: dimensionField,
+        method: 'value',
+        value: 0
+      });
+    }
+    if(facetby) {
```
kanitw: Why adding this `if(facetby) {`?

src/transform.ts
```diff
@@ -112,19 +112,21 @@ export interface StackTransform {
...
    */
   sort?: SortField ;
   /**
-   * Output field names.
-   * y2 or x2 is optional if only one name is provided the end will be “<name>_end”
-   * Otherwise as will be ["StackField_start","StackField_end"]
+   * Output field names. This can be either a string or an array of strings with
+   * two elements denoting the name for the fields for stack start and stack end
+   * respectively.
+   * If a single string(eg."val") is provided, the end field will be "val_end".
```
kanitw: > string(eg."val")

string (e.g., "val")

src/compile/data/stack.ts
```diff
@@ -21,35 +21,53 @@ function getStackByFields(model: UnitModel): string[] {
 }
 
 export interface StackComponent {
+
   /**
-   * Faceted field.
+   * Faceted field. Used in makeFromEncoding.
```
kanitw: See https://github.com/vega/vega-lite/pull/3771/commits/6bbd2d63c51e7da9dfcaa5bb25925a5c76d1c8e8#r189408463

src/compile/data/stack.ts
```diff
@@ -21,35 +21,53 @@ function getStackByFields(model: UnitModel): string[] {
...
+  groupby?: string[];
+  /**
+   * Output field names of each stack field.
+   */
+  as: string[];
+
+}
+// TODO Better Name for this function
+function isAsValidArray(as: string[] | string): as is string[] {
+  return isArray(as) && as.every(s => isString(s)) && as.length ===2;
```
kanitw: If you really wanna keep this method, check if array.length > 1.  (The 3rd, 4th, ... elements can be ignored.)

src/transform.ts
```diff
@@ -102,6 +102,34 @@ export interface AggregatedFieldDef {
...
+  groupby: string[];
+  /**
+   * Mode for stacking marks.
+   * __Default value:__ `"zero"`
+   */
+  offset?: 'zero' | 'center' | 'normalize';
+  /**
+   * Field that determines the order of leaves in the stacked charts.
+   */
+  sort?: SortField ;
```
kanitw: SortField;

src/compile/data/stack.ts
```diff
@@ -65,7 +83,30 @@ export class StackNode extends DataFlowNode {
     this._stack = stack;
   }
 
-  public static make(parent: DataFlowNode, model: UnitModel) {
+  public static makeFromTransform(parent: DataFlowNode, stackTransform: StackTransform) {
+
+    const {stack, groupby, as, offset='zero'} = stackTransform;
+    const sort = stackTransform.sort || {'field': stack, 'order': 'ascending'};
```
kanitw: I'm curious why this default is needed?

kanitw:
Note that Github collapses a few of my comments again ><" -- make sure to look at them :)

kanitw:
1) Let's add @hide to the description of `StackTransform` to hide it from the schema for now (and re-compile schema) so we can merge this after you address the unaddressed comments.
2) For examples you wanna add, please add `_future` suffix to the example files so they can be included even though stack transform wouldn't be in the schema yet.

src/transform.ts
kanitw: Actually shouldn't it be `SortField[]` like window?

src/compile/data/stack.ts
invokesus: I'm not sure about this since I haven't read how `facetby` is generated, but the general logic behind the `if` is that there are 3 code segments in assemble: 
* Segment 1: when `this._stack` has been created by `makeFromEncoding` where `impute`: `true`
* Segment 2: when `this._stack` has been created by `makeFromEncoding`
* Segment 3. when `this._stack` has been created by `makeFromTransform`

The existing code structure led me to believe that Segment 1 and Segment 2 are not disjoint and both can be executed for a single input spec.

To maintain a flat structure, initially `this._stack` with `impute:true` is handled, then all the `this._stack` created by `makeFromEncoding`(**including** the ones operated on in the previous step), then finally the ones created by `makeFromTransform` which is disjoint to Segment 1 and 2.

`facetby` is `undefined` for any `this._stack` created by `makeFromTransform` . So those cases skip the `if(facetby) {` block.

src/compile/data/stack.ts
```diff
@@ -21,35 +21,53 @@ function getStackByFields(model: UnitModel): string[] {
...
+   */
+  groupby?: string[];
+  /**
+   * Output field names of each stack field.
+   */
+  as: string[];
+
+}
+
+function isAsValidArray(as: string[] | string): as is string[] {
```
kanitw: isValidAsArray

src/compile/data/stack.ts
```diff
@@ -180,27 +233,32 @@ export class StackNode extends DataFlowNode {
...
         type: 'impute',
-        field: stackField,
+        field,
         groupby: stackby,
         key: dimensionField,
         method: 'value',
         value: 0
       });
     }
+    if(facetby) {
```
kanitw: In general, you should look at your overall PR diff and think why you do certain things. 
Here you add `if (facetby)` to wrap the main code for the transform assemble logic (as it creates the stack transform).  Having a case that we can skip the main logic is quite questionable. 

From your [earlier rationale](https://github.com/vega/vega-lite/pull/3771#discussion_r189981663), you say that facetby is `undefined`  in `makeFromTransform`.  But why should that be undefined? (That's actually something you set yourself.)

FYI, previously `facetby` was required but somehow you make it optional in this PR. (Why?)
I actually mentioned earlier that the facetby should be an empty array  `[]` in either case.  
Thus, facetby shouldn't be optional and then you don't have to add the if here. 

Just FYI, it's okay to forget some details I mentioned because this is quite a complicated code base.  But feel free to ask again if you are unsure about  something.  :)

A good case to test whether this logic is now correct for facetby is to try to see if you can modify 
https://github.com/vega/vega-lite/blob/master/examples/specs/normalized/trellis_stacked_bar_normalized.vl.json
such that you have the stack transform inside the inner `spec`.

```
{
  "data": {
    "url": "data/barley.json"
  },
  "facet": {
    "column": {
      "field": "year",
      "type": "ordinal"
    }
  },
  "spec": {
    "transform": [{aggregate: ...}, {stack: ...}],
    "mark": "bar",
    "encoding": {
      ...
    }
  }
}



```

src/compile/data/stack.ts
```diff
@@ -65,7 +83,41 @@ export class StackNode extends DataFlowNode {
...
+      }
+    }
+    const sort: VgSort = {
+      field: sortFields,
+      order: sortOrder,
+    };
+    let normalizedAs: Array<string>;
+    if (isAsValidArray(as)) {
+      normalizedAs = as;
+    } else if(typeof as === 'string') {
```
kanitw: if(isString(as))

src/compile/data/stack.ts
invokesus: Thanks for the review, Ham. I think now I understand where I was wrong. I made `facetby` a required field, and refactored `assemble`. 
I must have missed your comment about trying out `stack` inside a `faceted` spec. It's working now.
screen shot 2018-05-23 at 12 01 30 am

src/compile/data/stack.ts
```diff
@@ -206,12 +207,12 @@ export class StackNode extends DataFlowNode {
...
       return [vgField(dimensionFieldDef)];
     }
-    return [];
+    return groupby || [];
   }
 
   public assemble(): VgTransform[] {
     const transform: VgTransform[] = [];
-    const {facetby, dimensionFieldDef, stackField: field, stackby, sort, offset, impute, groupby, as} = this._stack;
+    const {facetby, dimensionFieldDef, stackField: field, stackby, sort, offset, impute, as} = this._stack;
```
invokesus: Note: this commit changes the behavior of `getGroupbyFields`. 
An alternative would be to leave `getGroupbyFields` unchanged and instead do:
`const {facetby, dimensionFieldDef, stackField: field, stackby, sort, offset, impute, groupby=[], as} = this._stack;`
and then further down do `groupby: this.getGroupbyFields().concat(groupby, facetby),`

src/compile/data/stack.ts
```diff
@@ -152,15 +207,14 @@ export class StackNode extends DataFlowNode {
       }
       return [vgField(dimensionFieldDef)];
     }
-    return [];
+    return groupby || [];
```
kanitw: I'm ok with this for now.  

Once you decouple Impute from stacking (), we should make `dimensionFieldDef` no longer a part of StackComponent and just have `StackNode.makeFromEncoding()` create a `groupby`, using this method.  At that point, we can basically remove `facetBy` and just have push new dimension in the facet to `groupby` instead.

(Basically, if you look at the code, the final groupby in the stack transform = groupBy that's from dimension fieldDef + facetBy)

kanitw:
Since the code looks pretty stable at this point, I'll merge to a `feature/stack` branch for now. Please create another PR to merge to the feature branch.

PR #3777, opened 2018-05-19
Use latest VSCode's organize import on save

kanitw:
Caveat: it will introduce braces on save -- l saw that this will be [fixed in the next version of VSCode](https://github.com/Microsoft/vscode/issues/47654)


domoritz:
Let's wait for the release then.

kanitw:
It's now good!

domoritz:
:tada:

PR #3778, opened 2018-05-19
Support legend properties from Vega 4.

domoritz:
Support direction and columns. 

```
{
  "mark": "rect",
  "width": 300,
  "height": 200,
  "encoding": {
    "x": {
      "bin": {"maxbins":60},
      "field": "IMDB_Rating",
      "type": "quantitative"
    },
    "y": {
      "bin": {"maxbins": 40},
      "type": "quantitative"
    },
    "color": {
      "field": "US_Gross",
      "bin": {"maxbins": 20},
      "aggregate": "max",
      "type": "quantitative",
      "legend": {
        "direction": "vertical"
      }
    }
  },
  "config": {
    "range": {
      "heatmap": {
        "scheme": "greenblue"
      }
    },
    "view": {
      "stroke": "transparent"
    }
  }
}
```


src/compile/legend/parse.ts
```diff
@@ -172,6 +172,17 @@ export function mergeLegendComponent(mergedLegend: LegendComponent, childLegend:
     // Cannot merge due to inconsistent orient
     return undefined;
   }
+
+  const mergedDirection = mergedLegend.getWithExplicit('direction');
```
domoritz: Is this really necessary (and the one above) since we could just use the properties of the first legend instead.

domoritz:
Closed in favor of #3825

PR #3783, opened 2018-05-21
implement tickStep and update unit tests

sirahd:
- `tickStep` will auto-generate:
    - `tickCount` = `"signal": "(domain('channel')[1] - domain('channel')[0]) / tickStep + 1"`
    - `values` = `"signal": "sequence(domain('channel')[0], domain('channel')[1] + 1, 2)"`

### DIFF
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -49,9 +49,9 @@ export interface Axis extends VgAxisBase, Guide {
   tickCount?: number;
 
   /**
-   * A desired step size for prebinned data. This property will generate the appropriate `tickCount` and `values` for the prebinned field
+   * A desired step size for ticks. This property will generate the corresponding `tickCount` and `values` and can be useful for data that is prebinned before importing to Vega-Lite
    *
-   * __Default value__: `0`
+   * __Default value__: `undefined`
    */
   tickStep?: number;
 
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -49,7 +49,7 @@ export interface Axis extends VgAxisBase, Guide {
   tickCount?: number;
 
   /**
-   * A desired step size for ticks. This property will generate the corresponding `tickCount` and `values` and can be useful for data that is prebinned before importing to Vega-Lite
+   * A desired step size for ticks. This property will generate the corresponding `tickCount` and `values`. It can be useful for data that is binned before importing into Vega-Lite.
    *
    * __Default value__: `undefined`
    */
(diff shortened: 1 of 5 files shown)


src/compile/axis/properties.ts
```diff
@@ -95,13 +98,19 @@ export function values(specifiedAxis: Axis, model: UnitModel, fieldDef: FieldDef
...
+      const domain = model.scaleDomain(channel);
+      if (domain && domain !== 'unaggregated' && !isSelectionDomain(domain)) { // explicit value
+        return vals;
+      }
+      const signal = model.getName(`${binToString(fieldDef.bin)}_${fieldDef.field}_bins`);
+      return {signal: `sequence(${signal}.start, ${signal}.stop + ${signal}.step, ${signal}.step)`};
+    }
+    if (specifiedAxis.tickStep) {
+      const step = specifiedAxis.tickStep;
+      return {signal: `sequence(domain('${channel}')[0], domain('${channel}')[1] + 1, ${step})`};
```
kanitw: > `+ 1 `

`+ ${step}` ?

src/compile/axis/properties.ts
```diff
@@ -67,9 +67,12 @@ export function orient(channel: PositionScaleChannel) {
   throw new Error(log.message.INVALID_CHANNEL_FOR_AXIS);
 }
 
-export function tickCount(channel: PositionScaleChannel, fieldDef: FieldDef<string>, scaleType: ScaleType, size: VgSignalRef) {
+export function tickCount(specifiedAxis: Axis, channel: PositionScaleChannel, fieldDef: FieldDef<string>, scaleType: ScaleType, size: VgSignalRef) {
```
kanitw: Do you need the whole specifiedAxis,  seems `tickStep` is enough?  Also, maybe move this `tickStep` to last parameter?

sirahd:
@kanitw I think we should have a pattern for function signature, for example, `(model, channel, fieldDef)` that's consistent across modules (eg. axis) for most of the utility functions under them. Otherwise if we need some parameters inside the function, we'll have to add new params to the function and go back to edit existing tests we have.

src/axis.ts
```diff
@@ -49,9 +49,9 @@ export interface Axis extends VgAxisBase, Guide {
   tickCount?: number;
 
   /**
-   * A desired step size for prebinned data. This property will generate the appropriate `tickCount` and `values` for the prebinned field
+   * A desired step size for ticks. This property will generate the corresponding `tickCount` and `values` and can be useful for data that is prebinned before importing to Vega-Lite
```
kanitw: prebinned is not a standard term.   That's why I previously suggested 

"data that are already binned before importing into Vega-Lite". 

Also, this is a run-on sentence.  Break it into two sentences?

PR #3786, opened 2018-05-21
Fix all docs url to use absolute url

kanitw:
cc: @jakevdp


kanitw:
@domoritz  -- I start wondering if we should do this in the rename script so that we won't forget about it anymore and make the JS code simpler.

domoritz:
How do we automate this?

kanitw:
Replace `\]\(([^/.]*).html` with `](https://vega.github.io/vega-lite/docs/$1.html`

domoritz:
Although I like to have all URLs absolute as it also means our typescript typings have helpful links. I think we should enforce absolute URLs instead.

kanitw:
Ok i add check script instead.

jakevdp:
Thanks!

PR #3807, opened 2018-05-24
Allow naming any dataset. Fixes #3806

### LINKED ISSUE #3806: Support naming data sources with only values
This is useful when you wanna have a predictable (and stable) name. 

```
{
  "data": {
    "values": [
    ],
    "name": "source"
  },
  "mark": "bar",
  "encoding": {
  }
}
```

Same for datasets with a source URL.

domoritz:


domoritz:
Note that this will require a minor version bump.

kanitw:
> Note that this will require a minor version bump.

We haven't released 2.5, so no.

kanitw:
> My point is that if we merge, we have to make 2.5 the next release.

Yes, but there are already many new features. So 2.5 is already the next release lol

PR #3809, opened 2018-05-24
Add lt, lte, gt, gte predicates (#3561)

### LINKED ISSUE #3452: lessThan / moreThan Filter

kanitw:
Fix #3452


kanitw:
cc: @invokesus a few minor fixes here too

kanitw:
🎉  congrats!

PR #3810, opened 2018-05-24
 Fix logic for window’s default as

kanitw:
`_field` was wrong as multiple fields can have the same aggregate op

### DIFF
--- a/src/compile/data/window.ts
+++ b/src/compile/data/window.ts
@@ -27,8 +27,7 @@ export class WindowTransformNode extends DataFlowNode {
   }
 
   private getDefaultName(windowFieldDef: WindowFieldDef): string {
-    const {as, op} = windowFieldDef;
-    return as || vgField(windowFieldDef, {op});
+    return windowFieldDef.as || vgField(windowFieldDef);
   }
 
   public assemble(): VgWindowTransform {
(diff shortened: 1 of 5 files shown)


src/compile/data/window.ts
```diff
@@ -26,7 +26,8 @@ export class WindowTransformNode extends DataFlowNode {
   }
 
   private getDefaultName(windowFieldDef: WindowFieldDef): string {
-    return windowFieldDef.as === undefined ? String(windowFieldDef.op) + '_field' : windowFieldDef.as;
+    const {as, op, field} = windowFieldDef;
+    return as || (field ? `${op}_${field}` : op);
```
domoritz: Does this work with nested fields? Why not use `vgField`?

src/compile/data/window.ts
```diff
@@ -26,7 +27,8 @@ export class WindowTransformNode extends DataFlowNode {
   }
 
   private getDefaultName(windowFieldDef: WindowFieldDef): string {
-    return windowFieldDef.as === undefined ? String(windowFieldDef.op) + '_field' : windowFieldDef.as;
+    const {as, op} = windowFieldDef;
+    return as || vgField(windowFieldDef, {op});
```
kanitw: Actually this still seems hacky.... let me refactor more

PR #3811, opened 2018-05-25
Make warnings a bit more precise (and take `sort` into account)

kanitw:


src/fielddef.ts
```diff
@@ -579,7 +579,7 @@ export function channelCompatibility(fieldDef: FieldDef<Field>, channel: Channel
       if (type !== QUANTITATIVE) {
         return {
           compatible: false,
-          warning: log.message.channelShouldNotEncodeType(channel, type)
+          warning: `Channel ${channel} should be used with a quantitative field only, not ${fieldDef.type} field.`
```
domoritz: I still think it's good to keep the messages in `log.ts`

PR #3812, opened 2018-05-25
Add Flatten transform

invokesus:
TODO:
- [x] Use vega-typings
- [x] Docs
- [x] Example
- [x] Add docstrings in transform.ts


src/compile/data/flatten.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.transform.flatten.forEach(field => {
+      out[this.getDefaultName(field)] = true;
+    });
+    return out;
+  }
+
+  private getDefaultName(field: string): string {
+    const index = this.transform.flatten.indexOf(field);
+    // Returns the "as" entry corresponding to field if it exists else returns field
+    return (this.transform.as === undefined || this.transform.as[index] === undefined) ? field : this.transform.as[index];
```
invokesus: Right now this follows Vega Defaults. Not sure whether to change this and introduce different behavior when `as` is not specified.

domoritz: Can't we just pass nothing and let Vega handle the default case?

invokesus: @kanitw mentioned in slack that we need to set `as`, so that `producedFields` gives the correct output.  But here, since the default is the `field` value, I think passing nothing might be ok.

domoritz: Try it out and I'd always prefer to output less.

kanitw: You can output `as=undefined` if you modify `producedFields` to output `field` name instead if `as` is undefined

src/compile/data/flatten.ts
```diff
@@ -0,0 +1,48 @@
...
+    const as = [];
+    for (const field of this.transform.flatten) {
+      fields.push(field === undefined ? null : field);
+      as.push(this.getDefaultName(field));
+    }
+
+    const result: VgFlattenTransform = {
+      type: 'flatten',
+      fields,
+      as
```
kanitw: I'd remove the loop above and just do something like:

```
const {as=[], flatten} = this.transform;

return {
  type: 'flatten',
  fields: flatten, 
  as: flatten.map((f, i) => as[i] || f[i])
}
```

src/compile/data/flatten.ts
```diff
@@ -0,0 +1,48 @@
...
+  public clone() {
+    return new FlattenTransformNode(this.parent, duplicate(this.transform));
+  }
+
+  constructor(parent: DataFlowNode, private transform: FlattenTransform) {
+    super(parent);
+  }
+
+  public producedFields() {
+    const out = {};
```
kanitw: Better just use `reduce`

```
return this.transform.flatten.reduce((index, field, i) => {
  out[as[i] || field] = true;
  return out
}, {});
```

src/compile/data/flatten.ts
kanitw: I think this method is pretty unnecessary.

kanitw: > Try it out and I'd always prefer to output less.

I just check the rest of transform are all explicit though, so doing so would be quite inconsistent.  So let's make them explicit for consistency for now. 

If @domoritz doesn't like it, he can refactor all of the transforms later.

src/compile/data/flatten.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public producedFields() {
+    return this.transform.flatten.reduce((out, field, i) => {
+      out[this.transform.as[i]] = true;
+      return out;
+    }, {});
+  }
+
+  private getNames() {
```
kanitw: Remove this method and move the logic above?

examples/specs/vconcat_flatten.vl.json
```diff
@@ -0,0 +1,64 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "Show a line chart corresponding to selected points.",
+  "data": {
+    "values": [
+      { "id": "001",
+        "ra": "243.35",
```
kanitw: Do we need to put these numbers as strings? Same for "time".

examples/specs/vconcat_flatten.vl.json
```diff
@@ -0,0 +1,64 @@
...
+        "lc": [{"time": "1", "mag": 19.5}, {"time": "2", "mag": 20}]
+      },
+      { "id": "003",
+        "ra": "143.35",
+        "dec": "+33.6",
+        "lc": [{"time": "1", "mag": 19}, {"time": "2", "mag": 18}]
+      }
+    ]
+  },
+  "transform": [
```
kanitw: Why transforming it up here?  Doing so require you to apply aggregate to basically unflatten it for the upper plot.

examples/specs/vconcat_flatten.vl.json
```diff
@@ -0,0 +1,64 @@
...
+    {"flatten": ["lc"]}
+  ],
+  "vconcat": [
+    {
+      "width": 300,
+      "height": 200,
+      "title": "Sky position",
+      "transform": [{"aggregate": [], "groupby": ["ra", "dec", "id"]}],
```
kanitw: We shouldn't have to aggregate here.

examples/specs/vconcat_flatten.vl.json
```diff
@@ -0,0 +1,64 @@
...
+      "encoding": {
+        "x": {"field": "ra", "type": "quantitative", "scale": {"zero": false}},
+        "y": {"field": "dec", "type": "quantitative"},
+        "color": {
+          "condition": {"selection": "pts", "value": "steelblue"},
+          "value": "grey"
+        },
+        "size": {"value": 100}
+      },
+      "mark": "circle"
```
kanitw: We always put mark before encoding in our examples.

examples/specs/vconcat_flatten.vl.json
```diff
@@ -0,0 +1,64 @@
...
+        "x": {
+          "field": "lc.time",
+          "type": "quantitative",
+          "scale": {"zero": false}
+        },
+        "y": {"field": "lc.mag", "type": "quantitative"},
+        "color": {"value": "steelblue"},
+        "detail": {"field": "id", "type": "nominal"}
+      },
+      "mark": "line"
```
kanitw: We always put mark before encoding in our examples.

examples/specs/vconcat_flatten.vl.json
```diff
@@ -0,0 +1,64 @@
...
+      "transform": [{"filter": {"selection": "pts"}}],
+      "encoding": {
+        "x": {
+          "field": "lc.time",
+          "type": "quantitative",
+          "scale": {"zero": false}
+        },
+        "y": {"field": "lc.mag", "type": "quantitative"},
+        "color": {"value": "steelblue"},
+        "detail": {"field": "id", "type": "nominal"}
```
kanitw: Why do we need detail here? (Given the selection is a single selection.)

site/docs/transform/flatten.md
```diff
@@ -0,0 +1,31 @@
...
+Below are some examples to demonstrate the usage of the flatten transform.
+
+### Basic Example
+
+In this example, `flatten` is used on two fields simultaneously. `null` values are added to the shorter array.
+
+<div class="vl-example" data-name="circle_flatten"></div>
+
+
+### Concat Circle and Line Marks with Nested Data
```
kanitw: ### Advanced Example: Coordinated View with Nested Time Series

site/docs/transform/flatten.md
```diff
@@ -0,0 +1,31 @@
...
+{% include table.html props="flatten,as" source="FlattenTransform" %}
+
+
+## Examples
+
+Below are some examples to demonstrate the usage of the flatten transform.
+
+### Basic Example
+
+In this example, `flatten` is used on two fields simultaneously. `null` values are added to the shorter array.
```
kanitw: Not sure where is the `null` value you mention?

site/docs/transform/flatten.md
```diff
@@ -0,0 +1,31 @@
...
+---
+
+The **flatten** transform maps array-valued _fields_ to a set of individual data objects, one per array entry. This transform generates a new data stream in which each data object consists of an extracted array value as well as all the original fields of the corresponding input data object.
+
+
+## Flatten Transform Definition
+
+{% include table.html props="flatten,as" source="FlattenTransform" %}
+
+
```
kanitw: Having a usage section without specs like Vega does is helpful.  

The visual example is good additions, but having explicit examples how data is reshaped is useful.

site/docs/transform/flatten.md
kanitw: NVM got it now.

examples/specs/circle_flatten.vl.json
```diff
@@ -0,0 +1,25 @@
...
+  {"values":
+   [{ "key": "alpha", "foo": [1, 2],    "bar": ["A", "B"] },
+    { "key": "beta",  "foo": [3, 4, 5], "bar": ["C", "D"] }
+   ]
+  },
+  "transform": [{"flatten": ["foo", "bar"]}],
+  "mark": "circle",
+  "encoding": {
+    "x": {
+
```
kanitw: 1) Why line breaks here.  

2) I think you can fit field and type in one online for each of x, y, and color

PR #3813, opened 2018-05-25
Add Fold Transform

invokesus:
TODO:
- [ ] Docs
- [ ] Example

### DIFF
--- a/src/compile/data/fold.ts
+++ b/src/compile/data/fold.ts
@@ -14,20 +14,15 @@ export class FoldTransformNode extends DataFlowNode {
 
   constructor(parent: DataFlowNode, private transform: FoldTransform) {
     super(parent);
-    this.transform.as = this.getNames();
+    const specifiedAs = this.transform.as || [undefined, undefined];
+    this.transform.as = [specifiedAs[0] || 'key', specifiedAs[1] || 'value'];
   }
 
   public producedFields() {
-    const names = this.getNames();
-
-    if (names === undefined) {
-      return {'key': true, 'value': true};
-    } else {
-      return names.reduce((result,item) => {
-        result[item] = true;
-        return result;
-      }, {});
-    }
+    return this.transform.as.reduce((result,item) => {
+      result[item] = true;
+      return result;
+    }, {});
   }
 
   private getNames(): [string, string] | undefined {
--- a/src/compile/data/fold.ts
+++ b/src/compile/data/fold.ts
@@ -25,17 +25,6 @@ export class FoldTransformNode extends DataFlowNode {
     }, {});
   }
 
-  private getNames(): [string, string] | undefined {
-    const as = this.transform.as;
-    if (as && as.length > 1) {
-      return [as[0], as[1]];
-    } else if (as && as[0]) {
-      return [as[0], 'value'];
-    } else {
-      return undefined;
-    }
-  }
-
   public assemble(): VgFoldTransform {
     const {fold, as} = this.transform;
     const result: VgFoldTransform = {
--- a/src/compile/data/fold.ts
+++ b/src/compile/data/fold.ts
@@ -1,0 +1,55 @@
+import {FoldTransform as VgFoldTransform} from 'vega-typings';
+import {isString} from 'vega-util';
+import {FoldTransform} from '../../transform';
+import {duplicate} from '../../util';
+import {DataFlowNode} from './dataflow';
+
+/**
+ * A class for flatten transform nodes
+ */
+export class FoldTransformNode extends DataFlowNode {
+
+  public clone() {
+    return new FoldTransformNode(this.parent, duplicate(this.transform));
+  }
+
+  constructor(parent: DataFlowNode, private transform: FoldTransform) {
+    super(parent);
+  }
+
+  public producedFields() {
+    const defaultNames = this.getDefaultName();
+
+    if (defaultNames === undefined) {
+      return {'key': true, 'value': true};
+    } else {
+      return defaultNames.reduce((result,item) => {
+        result[item] = true;
+        return result;
+      }, {});
+    }
+  }
+
+  private getDefaultName(): [string, string] | undefined {
+    const as = this.transform.as;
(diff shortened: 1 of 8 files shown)


src/compile/data/fold.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (as && isString(as[0]) && isString(as[1])) {
+      return [as[0], as[1]];
+    } else {
+      return undefined;
+    }
+  }
+
+  public assemble(): VgFoldTransform {
+    const fields: string[] = [];
+    for (const field of this.transform.fold) {
```
kanitw: Why do you have to convert undefined to null?  Can't you just do `fields: this.transform.fold`?

src/compile/data/fold.ts
```diff
@@ -0,0 +1,55 @@
...
+      return {'key': true, 'value': true};
+    } else {
+      return defaultNames.reduce((result,item) => {
+        result[item] = true;
+        return result;
+      }, {});
+    }
+  }
+
+  private getDefaultName(): [string, string] | undefined {
```
kanitw: This method do more than returning default names.   (It returns non-default specified value too, so the method name is a bit misleading.)

src/compile/data/parse.ts
```diff
@@ -92,6 +93,12 @@ export function parseTransformArray(head: DataFlowNode, model: Model, ancestorPa
       for (const field of keys(stack.producedFields())) {
         ancestorParse.set(field, 'derived', false);
       }
+    } else if (isFold(t)) {
+      const fold = head = new FoldTransformNode(head, t);
+
+      for (const field of keys(fold.producedFields())) {
+        ancestorParse.set(field, 'derived', false);
```
kanitw: Don't forget to refactor these once we merge all of fold and flatten PR.

src/vega.schema.ts
```diff
@@ -1,4 +1,5 @@
 import {AggregateOp} from 'vega';
+import {FoldTransform as VgFoldTransform} from 'vega-typings';
```
kanitw: Great!  This is exactly what we need.  

@domoritz  I wonder if we should import and reexport from this file so the rest of the codebase do not have to do `as` and thus will have consistent name?

src/compile/data/fold.ts
```diff
@@ -0,0 +1,55 @@
...
+    const as = this.transform.as;
+    if (as && isString(as[0]) && isString(as[1])) {
```
kanitw: 1) I don't think you need to check if each item is a string as that's already enforced by the schema.  Just need to check if they exist.

2) if as[0] is specified, but as[1] isn't.  Shouldn't we use `[as[0], 'value']` ?

test/compile/data/fold.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        as: ['A'] as any
+      };
+      const fold = new FoldTransformNode(null, transform);
+      assert.deepEqual(fold.assemble(), {
+        type: 'fold',
+        fields: ['a','b'],
+        as: undefined
+      });
+    });
+    it ('should return proper produced fields for no "as"', () => {
```
kanitw: Please add line breaks between tests.

test/compile/data/fold.test.ts
```diff
@@ -0,0 +1,64 @@
+/* tslint:disable:quotemark */
+
+import {assert} from 'chai';
+import {FoldTransformNode} from '../../../src/compile/data/fold';
+import {Transform} from '../../../src/transform';
+
+describe('compile/data/fold', () => {
+  describe('Testing FoldTransformNode', () => {
```
kanitw: 'Testing FoldTransformNode' => 'FoldTransformNode'

Everything under test is a test. :)

src/transform.ts
```diff
@@ -254,7 +254,18 @@ export interface LookupTransform {
...
 
+export interface FoldTransform {
+  /**
+   * An array of data fields indicating the properties to fold.
+   */
+  fold: string[];
 
+  /**
+   * The output field names for the key and value properties produced by the fold transform.
+   * The default is ["key", "value"].
```
kanitw: Always use `__Default value:__` and wrap the default value in back ticks

```
__Default value:__ `["key", "value"]`
```

src/compile/data/fold.ts
```diff
@@ -0,0 +1,53 @@
...
+    return new FoldTransformNode(this.parent, duplicate(this.transform));
+  }
+
+  constructor(parent: DataFlowNode, private transform: FoldTransform) {
+    super(parent);
+    this.transform.as = this.getNames();
+  }
+
+  public producedFields() {
+    const names = this.getNames();
```
kanitw: It's weird that you're calling `getNames()` again in producedFields(). 

The whole point of setting as in constructor is that you can just simply uses `as` to determine producedFields.

kanitw: Let me fix it :)

test/compile/data/fold.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    it ('should return proper produced fields for no "as"', () => {
+      const transform: Transform = {
+        fold: ['a', 'b']
+      };
+      const fold = new FoldTransformNode(null, transform);
+      assert.deepEqual(fold.producedFields(), {'key': true, 'value': true});
+    });
+
+    it ('should return proper produced fields for no "as"', () => {
```
invokesus: Forgot to change this. Should be `complete` instead of `no`.

kanitw: Can you fix it and make test pass

invokesus: Yup.

src/compile/data/fold.ts
```diff
@@ -25,17 +25,6 @@ export class FoldTransformNode extends DataFlowNode {
     }, {});
   }
 
-  private getNames(): [string, string] | undefined {
```
kanitw: ah sorry forgot to remove this.

Issue #3770, opened 2018-05-18
Error Band + ErrorBar Refactor


kanitw:
@chanwutk  When you do error band, please also do the following refactor 

1) Drop size in ErrorBar by `filterUnsupportedChannels`

Currently we drop size using `const {size: _s, ...sharedEncoding} = encodingWithoutContinuousAxis;`
but we already have `filterUnsupportedChannels` -- better use it.  

2) Better move the following code inside `errorBarParams`  as they are "params" of errorbar too -- this way we can just reuse the whole `errorBarParams` for errorBand.  (You may need to input "mark" between errorbar or errrorband to customize error message. 

```
spec = filterUnsupportedChannels(spec, ERRORBAR);

  // TODO: use selection
  const {mark, encoding, selection, projection: _p, ...outerSpec} = spec;
  const markDef: ErrorBarDef = isMarkDef(mark) ? mark : {type: mark};

  // TODO(https://github.com/vega/vega-lite/issues/3702): add selection support
  if (selection) {
    log.warn(log.message.selectionNotSupported('errorbar'));
  }

  const center: ErrorBarCenter = markDef.center || config.errorbar.center;
  const extent: ErrorBarExtent = markDef.extent || ((center === 'mean') ? 'stderr' : 'iqr');

  if ((center === 'median') !== (extent === 'iqr')) {
    log.warn(`${center} is not usually used with ${extent} for error bar.`);
  }
```

kanitw:
It's also ok for `errorband.ts` to improve things from `errorbar.ts` (no need to move to common)

chanwutk:
If we drop size in filterUnsupportedChannels, wouldn't it also drop the size of Boxplot in its encoding?

PR #3820, opened 2018-05-26
Implement Error Band

chanwutk:


chanwutk:
@kanitw At line 135 of errorbar.ts, I put `any` as a return type because it gives me an error if otherwise. What should be the correct type to put in as the return type?

kanitw:
`transform` is `Transform[]`.  The rest of them, you should be able to hover each of the variable and see the type.  

(ps. I look at it and I'm surprised TS isn't smart enough to figure this out too --weird!)

chanwutk:
I forgot to do the #3770, I will do it in the next commit.

src/compositemark/errorband.ts
```diff
@@ -0,0 +1,98 @@
...
+import {ErrorBarCenter, ErrorBarExtent, errorBarParams} from './errorbar';
+
+export const ERRORBAND: 'errorband' = 'errorband';
+export type ErrorBand = typeof ERRORBAND;
+
+export type ErrorBandPart = 'band' | 'rims';
+
+const ERRORBAND_PART_INDEX: Flag<ErrorBandPart> = {
+  band: 1,
+  rims: 1
```
kanitw: @domoritz any thoughts about the name `rims`? 

A few other alternatives: edges, borders

domoritz: outline, limits

I think border(s) is probably the most intuitive

chanwutk:
Since we have `borders` for `error band` and `ticks` for `error bar`, should we find a word that can substitute both `borders` and `ticks` but still makes sense, so that both `error bar` and `error band` have the same syntax?

domoritz:
I'm not too worried about the difference but I'm open to suggestions.

chanwutk:
How about edges or ends?

domoritz:
Edge doesn't work for error bar, ends doesn't work for error band, I think.

kanitw:
I like `borders` too

PR #3821, opened 2018-05-26
Update vega-typings version

invokesus:


kanitw:
You may want to rebase `feature/transform` and your own branches so we have this new typing included.

kanitw:
(I would have been easier if you just include this upgrade in your branch, so nested rebases needed.  But given that we already merged, then you have to rebase :p)

PR #3823, opened 2018-05-27
Add concat with point and line example. Fixes #3655

### LINKED ISSUE #3655: Add focus and context example with (not yet) nested data
Once we have flatten, we can use nested data

```json
    {"id": "001", "ra":"243.35", "dec":"+54.6", "lc":[{"time": "1", "mag":18.5}, {"time": "2", "mag": 19.0}]}, 
    {"id": "002", "ra":"210.35", "dec":"+14.6", "lc":[{"time": "1", "mag":19.5}, {"time": "2", "mag": 20.0}]}, 
    {"id": "003", "ra":"143.35", "dec":"+33.6", "lc":[{"time": "1", "mag":19.0}, {"time": "2", "mag": 18.0}]} 
```

```json
{
  "description": "This will be a plot of RA and Dec and light curves.",
  "data": {
    "values": [
      {"id": "001", "ra": "243.35", "dec": "+54.6", "time": "2", "mag": 19},
      {"id": "001", "ra": "243.35", "dec": "+54.6", "time": "1", "mag": 18.5},
      {"id": "002", "ra": "210.35", "dec": "+14.6", "time": "2", "mag": 20},
      {"id": "002", "ra": "210.35", "dec": "+14.6", "time": "1", "mag": 19.5},
      {"id": "003", "ra": "143.35", "dec": "+33.6", "time": "2", "mag": 18},
      {"id": "003", "ra": "143.35", "dec": "+33.6", "time": "1", "mag": 19}
    ]
  },
  "vconcat": [
    {
      "width": 300,
      "height": 200,
      "title": "sky position",
      "transform": [{"aggregate": [], "groupby": ["ra", "dec", "id"]}],
      "selection": {
        "pts": {"type": "single", "fields": ["id"], "empty": "none"}
      },
      "encoding": {
        "x": {"field": "ra", "type": "quantitative"},
        "y": {"field": "dec", "type": "quantitative"},
        "color": {
          "condition": {"selection": "pts", "value": "darkred"},
          "value": "grey"
        },
        "size": {"value": 60}
      },
      "mark": "circle"
    },
    {
      "width": 300,
      "height": 200,
      "title": "light curve",
      "transform": [{"filter": {"selection": "pts"}}],
      "encoding": {
        "x": {
          "field": "time",
          "type": "quantitative",
          "scale": {"zero": false}
        },
        "y": {"field": "mag", "type": "quantitative"},
        "color": {"value": "darkred"}
      },
      "mark": "line"
    }
  ]
}
```

domoritz:


examples/specs/interactive_concat.vl.json
```diff
@@ -0,0 +1,51 @@
...
+    ]
+  },
+  "vconcat": [
+    {
+      "width": 300,
+      "height": 200,
+      "title": "Sky position",
+      "transform": [{"aggregate": [], "groupby": ["ra", "dec", "id"]}],
+      "selection": {
+        "pts": {"type": "single", "fields": ["id"], "empty": "none"}
```
kanitw: Why empty none?

examples/specs/interactive_concat.vl.json
```diff
@@ -0,0 +1,51 @@
...
+      {"id": "003", "ra": "143.35", "dec": "+33.6", "time": "2", "mag": 18},
+      {"id": "003", "ra": "143.35", "dec": "+33.6", "time": "1", "mag": 19}
+    ]
+  },
+  "vconcat": [
+    {
+      "width": 300,
+      "height": 200,
+      "title": "Sky position",
+      "transform": [{"aggregate": [], "groupby": ["ra", "dec", "id"]}],
```
kanitw: I wonder if the concept you wanna show is more just this aggregate with groupby only.  (Maybe worth adding a simpler example to the aggregate page instead? Or at least add another one.)

examples/specs/interactive_concat.vl.json
domoritz: all is probably better

examples/specs/interactive_concat.vl.json
domoritz: I don't think aggregate without group by is not that magical.

domoritz: I don't think aggregate with only group by is that magical and not the main point of this example.

domoritz:
On slack @kanitw agreed that it makes sense to use flatten and nested data instead.

PR #3825, opened 2018-05-27
Support v4 Legends

domoritz:


src/scale.ts
```diff
@@ -51,6 +50,9 @@ const SCALE_CATEGORY_INDEX: {
   log: 'numeric',
   pow: 'numeric',
   sqrt: 'numeric',
+  quantize: 'numeric',
```
kanitw: linear/log/pow (numeric scales) map  Q => Q values

quantile, quantize, threshold map quantitative to discrete values and thus shouldn't be the same scale type. 

If you want to support these, I suggest that you separate adding them to another PR first since there are probably delicate details that requires thorough review.

site/docs/encoding/legend.md
```diff
@@ -89,19 +89,19 @@ To provide themes for all legends, the legends config (`config: {legend: {...}}`
 
 ### General
 
-{% include table.html props="cornerRadius,fillColor,orient,offset,strokeColor,strokeDash,strokeWidth,padding" source="LegendConfig" %}
+{% include table.html props="columns,columnPadding,cornerRadius,fillColor,gridAlign,orient,offset,rowPadding,strokeColor,strokeDash,strokeWidth,padding" source="LegendConfig" %}
```
kanitw: A lot of these are now repeated.  Better not repeat them but instead say config support all properties except a, b, c...  (and, if applicable, config can also have a few more properties).

site/docs/encoding/legend.md
```diff
@@ -61,7 +61,7 @@ To customize legends, a `legend` object in [an encoding channel's definition](en
 
 {:#properties}
 
-{% include table.html props="type,entryPadding,format,offset,orient,padding,tickCount,title,values,zindex" source="Legend" %}
```
kanitw: So entryPadding is removed?  (Sounds like a breaking change.)

src/compile/legend/properties.ts
```diff
@@ -15,15 +15,3 @@ export function values(legend: Legend) {
   }
   return vals;
 }
-
-export function type(t: Type, channel: Channel, scaleType: ScaleType): 'gradient' {
```
kanitw: I'm curious why this should be removed.  You probably have a good reason, but better explain too. :)

src/scale.ts
kanitw: quantile, quantize, threshold (and bin-linear, bin-ordinal) are called discretizing scales in Vega https://vega.github.io/vega/docs/scales/#discretizing

kanitw: I think we maybe want to leave if bin-linear and bin-ordinal using their keyword though as we probably don't want to allow merging them?

domoritz:
Most of these changes are in #3778. Let's get that in first.

Saba9:
Vega-Lite should be updated to support Vega's`gradientLength` and `gradientThickness` properties. The Vega-Lite docs currently mention analogous `gradientHeight` and `gradientWidth` properties but they don't currently work.

site/docs/encoding/legend.md
domoritz: Yes

src/compile/legend/properties.ts
domoritz: Now, the type is inferred from the scale type and we do not override it anymore.

src/legend.ts
```diff
@@ -1,15 +1,15 @@
...
 
 
-export interface LegendConfig extends VgLegendConfig, VlOnlyGuideConfig {}
+export interface LegendConfig extends LegendConfigNoSignals, VlOnlyGuideConfig {}
 
 /**
  * Properties of a legend or boolean flag for determining whether to show it.
  */
-export interface Legend extends VgLegendBase, Guide {
+export interface Legend extends BaseLegendConfig, Guide {
```
kanitw: Shouldn't everything down here now should all extend from vega-typings?

A few exception
- `type` -- we want to override "default") 
- `orient` -- (maybe not an exception if the default is not different from Vega)
- `encoding` - VL specific.

src/compile/legend/properties.ts
```diff
@@ -14,14 +11,9 @@ export function values(legend: Legend, fieldDef: FieldDef<string>) {
...
 
-export function type(t: Type, channel: Channel, scaleType: ScaleType): 'gradient' {
-  if (
-      isColorChannel(channel) && (
-        (t === 'quantitative' && !isBinScale(scaleType)) ||
-        (t === 'temporal' && contains<ScaleType>(['time', 'utc'], scaleType))
-      )
-    ) {
-    return 'gradient';
+export function clipHeight(scaleType: ScaleType) {
```
kanitw: Write a test for this?

src/compile/legend/encode.ts
```diff
@@ -1,34 +1,25 @@
...
+export function symbols(fieldDef: FieldDef<string>, symbolsSpec: any, model: UnitModel, channel: Channel, legendCmp: LegendComponent): SymbolEncodeEntry {
+  if (legendCmp.get('type') === 'gradient') {
     return undefined;
   }
 
   let out = {
     ...applyMarkConfig({}, model, FILL_STROKE_CONFIG),
     ...mixins.color(model)
-  };
+  } as SymbolEncodeEntry;
```
kanitw: Why do we have to cast here.  Can't we assign type to out? (Add comment if we can't)

site/docs/encoding/legend.md
```diff
@@ -85,25 +107,8 @@ To customize legends, a `legend` object in [an encoding channel's definition](en
...
-### Labels
-
-{% include table.html props="labelAlign,labelBaseline,labelColor,labelFont,labelFontSize,labelLimit,labelOffset,shortTimeLabels" source="LegendConfig" %}
-
-### Symbols
-
-{% include table.html props="entryPadding,symbolColor,symbolType,symbolSize,symbolStrokeWidth" source="LegendConfig" %}
-
-### Title
+To provide themes for all legends, the legends config (`config: {legend: {...}}`) can all [legend properties](#properties) except `direction`, `format`, `orient`, `tickCount`, `title`, `values`, and `zindex`.
```
kanitw: "can all" -> "support all"

- orient is supported in Vega legend.

- Perhaps worth mentioning that  `direction` has configs based on type (`gradientDirection` and `symbolDirection`).

site/docs/encoding/legend.md
```diff
@@ -61,7 +61,29 @@ To customize legends, a `legend` object in [an encoding channel's definition](en
 
 {:#properties}
 
-{% include table.html props="type,entryPadding,format,offset,orient,padding,tickCount,title,values,zindex" source="Legend" %}
+### General
+
```
kanitw: Missing `type`?

build/vega-lite-schema.json
```diff
@@ -4854,55 +5022,72 @@
...
+          "type": "number"
         },
-        "gradientHeight": {
-          "description": "The height of the gradient, in pixels.",
-          "minimum": 0,
+        "cornerRadius": {
+          "description": "Corner radius for the full legend.",
           "type": "number"
         },
-        "gradientLabelBaseline": {
```
kanitw: So this is gone?

build/vega-lite-schema.json
```diff
@@ -4854,55 +5022,72 @@
...
+          "description": "The color of the gradient stroke, can be in hex color code or regular color name.\n\n__Default value:__ `\"lightGray\"`.",
           "type": "string"
         },
         "gradientStrokeWidth": {
-          "description": "The width of the gradient stroke, in pixels.",
+          "description": "The width of the gradient stroke, in pixels.\n\n__Default value:__ `0`.",
           "minimum": 0,
           "type": "number"
         },
-        "gradientWidth": {
```
kanitw: So this is gone?

build/vega-lite-schema.json
```diff
@@ -4918,25 +5103,48 @@
...
+            }
+          ],
+          "description": "The strategy to use for resolving overlap of labels in gradient legends. If `false`, no overlap reduction is attempted. If set to `true` (default) or `\"parity\"`, a strategy of removing every other label is used. If set to `\"greedy\"`, a linear scan of the labels is performed, removing any label that overlaps with the last visible label (this often works better for log-scaled axes).\n\n__Default value:__ `true`."
+        },
         "offset": {
-          "description": "The offset, in pixels, by which to displace the legend from the edge of the enclosing group or data rectangle.\n\n__Default value:__  `0`",
+          "description": "The offset in pixels by which to displace the legend from the data rectangle and axes.\n\n__Default value:__ `18`.",
           "type": "number"
         },
-        "orient": {
```
kanitw: orient is supported in Vega Legend config: https://vega.github.io/vega/docs/config/#legends

src/legend.ts
domoritz: We cannot extend and change the type of some.

src/compile/legend/properties.ts
domoritz: This is not used yet.

domoritz: Done

src/compile/legend/encode.ts
domoritz: Added a fixme

site/docs/encoding/legend.md
domoritz: Done

site/docs/encoding/legend.md
domoritz: Good catch

build/vega-lite-schema.json
domoritz: Yes

build/vega-lite-schema.json
domoritz: It's been replaced. Yes

build/vega-lite-schema.json
domoritz: Oh, this is annoying. Orient is supported in config and legend but it does not support signals.

PR #3833, opened 2018-05-30
Add Example: Cumulative Frequency Distribution

kanitw:


domoritz:
Can you change the y axis title?

PR #3839, opened 2018-06-01
TS 2.9 follow up

domoritz:
Requires Microsoft/TypeScript/issues/24588 and https://github.com/Microsoft/TypeScript/issues/24546

Follow up for #3836


jasongrout:
I am also having problems compiling JupyterLab with vega-lite 2.5.1 because vega-lite is importing the json-stable-stringify library like this:

https://github.com/vega/vega-lite/blob/40bd312601dde2c2e23947196e5f55f59d10d04d/src/util.ts#L1

My error is: 

```
../../node_modules/vega-lite/build/src/util.d.ts:1:8 - error TS1192: Module '"[snip]/node_modules/@types/json-stable-stringify/index"' has no default export.

1 import stableStringify from 'json-stable-stringify';
         ~~~~~~~~~~~~~~~
```
I can suppress the error by setting `"allowSyntheticDefaultImports": true` in the tsconfig, but I think that's still probably not doing the right thing, given I'm not using Babel or something else that is converting that import for me. Reading https://github.com/Microsoft/TypeScript/pull/5577, it seems that vega-lite should be importing this as `import * as stableStringify from 'json-stable-stringify'` to make it compatible across different kinds of loaders, including loaders that don't magically make a default export when there isn't one.

domoritz:
This change simplifies the package.json and also improves our sourcemaps.

package.json
```diff
@@ -45,7 +45,7 @@
     "renameschema": "scripts/rename-schema.sh",
     "presite": "npm run prebuild && npm run data && npm run build:site && npm run build:toc && npm run build:versions && scripts/create-example-pages",
     "site": "bundle exec jekyll serve --incremental",
-    "lint": "tslint -p .",
+    "lint": "tslint -p . -e 'package.json'",
```
domoritz: This is to get around a bug in tslint.

domoritz:
Ready @kanitw

package.json
domoritz: https://github.com/palantir/tslint/issues/3957

PR #3846, opened 2018-06-04
Update Feature_request.md

kanitw:


domoritz:
I don't think we've had problems with feature requests before so I don't know whether we need more structure. It is super necessary for bugs, though.

kanitw:
Well, it's not "problematic" because we ask them more questions later.   This template helps them consider helping us even more.

kanitw:
Plus, I remember people did do duplicate issues for feature requests too.

PR #3847, opened 2018-06-04
Move Vega-Lite 1 down

domoritz:


domoritz:
This should also fix this:

screen shot 2018-06-04 at 11 29 09

PR #3848, opened 2018-06-04
Add compostion `align`, `bound`, `center` to Facet/Repeat/ConcatSpec

kanitw:
This allow users to customize align/bound/center properties of the underlying Vega's layout operator.

Note that we considered name like `layoutAlign/Bound/Center`.  However, we don't have a concept of a "layout" in VL and composition operators are already layout operator so adding prefix doesn't help much and only makes things more complicated.   Plus, I normally think of "layout" as general layout like force-directed layout/tree layout/graph layout more than just specific to grid layout.  

(I don't think `align` is useful for concat though, so I didn't add it.)


domoritz:
Since repeat is just a shorthand for concat, I think we should support all properties that work in repeat in concat as well.

kanitw:
> Since repeat is just a shorthand for concat, I think we should support all properties that work in repeat in concat as well.

It seems like it is, but it's not.  Repeat is grid operator, but concat is not.  Unless we have wrapped concat (which is a grid operator), I don't see why align is useful for concat.

domoritz:
I see. I guess we will never move repeat to the normalizer then. Instead, we will at some point add the ability to move the axis outside of the concat so that SPLOMs look great. Makes sense.

kanitw:
>  I guess we will never move repeat to the normalizer then.

Not if we think concat shouldn't contain wrapped concat.

kanitw:
@domoritz  you may want to add your marginal histogram example as a part of this PR as an example.

kanitw:
> Before we merge this, I'd like to get an okay form @jheer and @arvind.

Feel free to focus on the names of these properties.  I think whether we add `align` to concat can be dealt later. (Since we're not adding it, this means we can add it later.)

domoritz:
> @domoritz you may want to add your marginal histogram example as a part of this PR as an example.

Well, the example uses concat and you don't want to add align :-P

kanitw:
> Well, the example uses concat and you don't want to add align :-P

1) @domoritz you [asked for `bounds`](https://github.com/vega/vega-lite/pull/3840), not `align`. 
2) Maybe `align` is useful for this case if we use a `grid` concat to create the marginal histogram example.  But if you use concat twice, I'm not sure why is it useful here.  (I could be wrong.)

domoritz:
Added.

examples/specs/concat_marginal_histograms.vl.json
```diff
@@ -0,0 +1,77 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "data": {"url": "data/movies.json"},
+  "spacing": 10,
```
kanitw: @domoritz  Thanks for adding. 

Btw, I'm wondering if 15 is better since the 0 of the top plot is a bit too close to the lower plot right now.

domoritz: I like it being close but I see the argument for more space.

domoritz: I added more spacing and pushed.

PR #3850, opened 2018-06-04
Add Impute Transform

invokesus:
Will make separate PRs for 
- Decoupling from stack 
- Impute Transform docs

### DIFF
--- a/src/compile/data/impute.ts
+++ b/src/compile/data/impute.ts
@@ -1,0 +1,32 @@
+import {ImputeTransform} from '../../transform';
+import {duplicate} from '../../util';
+import {VgImputeTransform} from '../../vega.schema';
+import {DataFlowNode} from './dataflow';
+
+export class ImputeTransformNode extends DataFlowNode {
+  public clone() {
+    return new ImputeTransformNode(this.parent, duplicate(this.transform));
+  }
+
+  public producedFields() {
+    // typescript detects true as boolean type
+    return {[this.transform.impute]: true as true} ;
+  }
+
+  constructor(parent: DataFlowNode, private transform: ImputeTransform) {
+    super(parent);
+  }
+
+  public assemble(): VgImputeTransform {
+    const {impute, key, keyvals, method, groupby, value} = this.transform;
+    return {
+      type: 'impute',
+      field: impute,
+      key,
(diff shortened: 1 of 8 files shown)


src/compile/data/impute.ts
```diff
@@ -0,0 +1,32 @@
...
+import {DataFlowNode} from './dataflow';
+
+export class ImputeTransformNode extends DataFlowNode {
+  public clone() {
+    return new ImputeTransformNode(this.parent, duplicate(this.transform));
+  }
+
+  public producedFields() {
+    // typescript detects true as boolean type
+    return {[this.transform.impute]: true as true} ;
```
kanitw: This casting is weird.  What's wrong with boolean type?

src/transform.ts
```diff
@@ -212,6 +212,38 @@ export interface WindowTransform {
...
+   * __Default value:__:  `"value"`
+   */
+  method?: 'value' | 'mean' | 'median' | 'max' | 'min';
+  /**
+   * An optional array of fields by which to group the values.
+   * Imputation will then be performed on a per-group basis.
+   */
+  groupby?: string[];
+  /**
+   * The field value to use when the imputation method is `"value"`.
```
kanitw:
```
imputation `method`
```

src/compile/data/impute.ts
invokesus: Getting this error 
`Type '{ [x: string]: boolean; }' is not assignable to type 'Dict<true>'.`

invokesus:
I use a GUI so doesn't take that much time .🙂

src/compile/data/impute.ts
kanitw: Ok then.

kanitw:
> I get lost if I get big merge conflicts so I like to make smaller commits.

Ok. That's fine.  Let's say you can either do an "atomic" PR (like this one--which mean I'm gonna ignore your commit messages and just merge squash) or a PR with one or a few atomic commits (which makes is much easier for reviewer).
  
If you do multiple atomic things in a PR, please make your PR contains truly atomic commits, instead subatomic commits as the latter is equivalent to having one big commits for a reviewer's perspective.

kanitw:
Awesome :)

PR #3852, opened 2018-06-04
Allow bin.as to be an array of 2 strings

### LINKED ISSUE #3831: Allow bin's `as` to be an array of (two) strings?
Currently `as` is a string, which makes sense if we only want to used the binned output in aggregate.

However, if we plan to use the two output to do additional calculation, then it's unclear from the spec that the second field is `<field_name>_end`.  Given that we enforce all fields in `transform` to be named, this can violate our policy if `<field_name>_end` is directly used. 

cc: @domoritz @invokesus

invokesus:
Fixes #3831

### DIFF
--- a/src/compile/data/bin.ts
+++ b/src/compile/data/bin.ts
@@ -38,7 +38,6 @@ function getSignalsFromModel(model: Model, key: string) {
   };
 }
 
-// Bad Name?
 function isBinTransform(t: FieldDef<string> | BinTransform): t is BinTransform {
   return 'as' in t;
 }
@@ -46,10 +45,8 @@ function isBinTransform(t: FieldDef<string> | BinTransform): t is BinTransform {
 function createBinComponent(t: FieldDef<string> | BinTransform, model: Model) {
   let as: [string, string];
 
-  if (isBinTransform(t) && isString(t.as)) {
-    as = [t.as, `${t.as}_end`];
-  } else if (isBinTransform(t)) {
-    as = [t.as[0], t.as[1]];
+  if (isBinTransform(t)) {
+    as = isString(t.as)? [t.as, `${t.as}_end`] : [t.as[0],t.as[1]];
   } else {
     as = [vgField(t, {}), vgField(t, {binSuffix: 'end'})];
   }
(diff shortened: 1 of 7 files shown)


src/compile/data/bin.ts
```diff
@@ -38,15 +38,18 @@ function getSignalsFromModel(model: Model, key: string) {
   };
 }
 
+// Bad Name?
```
kanitw: What would be a better name?

src/compile/data/bin.ts
```diff
@@ -38,15 +38,18 @@ function getSignalsFromModel(model: Model, key: string) {
...
   return 'as' in t;
 }
 
 function createBinComponent(t: FieldDef<string> | BinTransform, model: Model) {
   let as: [string, string];
 
-  if (isBinTransform(t)) {
+  if (isBinTransform(t) && isString(t.as)) {
     as = [t.as, `${t.as}_end`];
+  } else if (isBinTransform(t)) {
```
kanitw: I think you better add `if (isString(t.as))` in side the original `isBinTransform` if branch

src/compile/data/bin.ts
invokesus: My bad. I didn't pay attention to the type of the argument.

kanitw:
Not sure why one of the example is outdated.

invokesus:
Can't reproduce locally 😕

kanitw:
Try rebasing to the latest master, upgrade deps, and re-run build:only and build:example?

src/compile/data/parse.ts
```diff
@@ -57,9 +57,12 @@ export function parseTransformArray(head: DataFlowNode, model: Model, ancestorPa
...
       head = new FilterNode(head, model, t.filter);
     } else if (isBin(t)) {
-      head = BinNode.makeFromTransform(head, t, model);
+      const bin = head = BinNode.makeFromTransform(head, t, model);
+
+      for (const field of keys(bin.producedFields())) {
+        ancestorParse.set(field, 'number', false);
+      }
 
-      ancestorParse.set(t.as, 'number', false);
```
invokesus: So, previously only the `as` `string` was being set. Now, both `as` elements(implicit as well as explicit) are set, which changed that example. The svg is unchanged.

PR #3854, opened 2018-06-05
Support Facet Sort 

### LINKED ISSUE #2176: Make facet support sort

kanitw:
Fix #2176 

~~Currently it works only if we have only row or column.  (Crossing case is way more complicated.)~~


kanitw:
> There is a problem if data is missing.

Nice catch.  We need to be able to impute.  However, facet doesn't have impute support. 
At this point, please help file a bug in both Vega-Lite and Vega. We'll have to ignore this case for the scope of this PR.

kanitw:
@domoritz -- this should have reasonable test coverage now.  For the case that requires impute, please file an issue to ask for support in Vega.

brucala:
Hi,
I don't know if this is known, but it looks like the facet sorting doesn't work properly when there is an aggregation in one of the encodings. For instance:

```
{
  "data": {
    "values": [
      {"a": "A","b": "X", "x": 1, "y": 4},
      {"a": "A","b": "Y", "x": 3, "y": 5},
      {"a": "A","b": "Z", "x": 5, "y": 6},
      {"a": "B","b": "X", "x": 2, "y": 7},
      {"a": "B","b": "Y", "x": 9, "y": 8},
      {"a": "B","b": "Z", "x": 8, "y": 9},
      {"a": "C","b": "X", "x": 7, "y": 1},
      {"a": "C","b": "Y", "x": 6, "y": 2},
      {"a": "C","b": "Z", "x": 4, "y": 3}
    ]
   },
  "facet": {
    "column": {"field": "a", "type": "nominal", "sort": ["B", "A", "C"]},
    "row": {"field": "b", "type": "ordinal", "sort": ["Y", "X", "Z"]}
  },
  "spec":{
  "mark": "point",
  "encoding": {
    "x": {
      "type": "quantitative",
      "field": "x",
      "aggregate": "sum",
      "scale": {"type": "linear"}
    },
    "y": {
      "type": "quantitative",
      "field": "y",
      "scale": {"type": "linear"}
    }
  }}
}
```

If `"aggregate": "sum"` is removed from `x` the sorting works.

loganpowell:
Oh man, thank you @domoritz  for your example! Just what the doctor ordered!

PR #3857, opened 2018-06-06
[To Discuss] Add header's `label/titlePadding`

kanitw:
Depends on https://github.com/vega/vega-lite/pull/3851


src/compile/layout/header.ts
```diff
@@ -1,3 +1,5 @@
+// TODO: refactor this into proper parse/assemble phase like other
```
domoritz: like other what?

src/facet.ts
```diff
@@ -17,7 +17,21 @@ export interface Header extends Guide {
    */
   labelAngle?: number;
 
-  // TODO: labelPadding
+  /**
+   * Padding spacing in pixels between header labels and axes.
```
domoritz: padding? spacing? Why both?

domoritz: See https://vega.github.io/vega-lite/docs/axis.html#title for how we name things in axes.

kanitw: I already saw that.  This is pretty consistent.

src/compile/layout/header.ts
kanitw: Like axis/legend.

kanitw:
Thinking more about this, let's not add this until someone asks for this, or until someone think clearly how `header` would work with wrapped facet so we can avoid breaking changes.  

This isn't as important as spacing anyway.

kanitw:
#4028 is probably better (simpler code)

PR #3858, opened 2018-06-06
[Example] Use parsing correctly in candlestick example + remove unnecessary timeUnit, scale and axis

kanitw:


examples/specs/layer_candlestick.vl.json
```diff
@@ -1,8 +1,13 @@
 {
   "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
   "width": 320,
-  "description":"A candlestick chart inspired from Protovis(http://mbostock.github.io/protovis/ex/candlestick.html)",
+  "description": "A candlestick chart inspired from Protovis(http://mbostock.github.io/protovis/ex/candlestick.html)",
```
domoritz: inpired by

examples/specs/layer_candlestick.vl.json
```diff
@@ -1,8 +1,13 @@
...
   "data": {
+    "format": {
+      "parse": {
+        "date": "date:'%Y-%m-%d'"
```
domoritz: Browsers should parse this automatically. Why do you need to add the format?

examples/specs/layer_candlestick.vl.json
kanitw: ?

examples/specs/layer_candlestick.vl.json
kanitw: Its parsing as utc

kanitw: Vega & Vega-Lite simply uses JS parsing.  The behavior is consistent with the screenshot you send.  But time scale is default local time.  So if you want to visualize local time, need to parse correctly.

domoritz: It's a bit weird that we `utc:FORMAT`, `utc`, and `date` parse as UTC but `date:FORMAT` parses as local, though.

examples/specs/layer_candlestick.vl.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
   "width": 320,
-  "description": "A candlestick chart inspired from Protovis(http://mbostock.github.io/protovis/ex/candlestick.html)",
+  "description": "A candlestick chart inspired by Protovis (http://mbostock.github.io/protovis/ex/candlestick.html)",
```
domoritz: maybe better say "inspired by an example in Protovis"

domoritz: :-D

PR #3863, opened 2018-06-07
Add TS 2.7 compatibility. Fixes #3862.

### LINKED ISSUE #3862: Typedoc can't deal with one line in the new vega-lite package
Typedoc is currently constrained to use TypeScript 2.7 (upgrade issue at https://github.com/TypeStrong/typedoc/issues/733). We use typedoc to generate important API docs, and we also depend on vega-lite (through vega-embed). There is only one line that TS 2.7 chokes on in the vega-lite compiled package, which in turn is causing our api doc build to fail: 

```
Error: [snip]/node_modules/vega-lite/build/src/compositemark/index.d.ts(12)
 ',' expected.
```

That line is:
```
export declare const COMPOSITE_MARK_STYLES: import("./boxplot").BoxPlotStyle[];
```
Presumably it is generated from https://github.com/vega/vega-lite/blob/e2280723a89df327575c011784d6c7371216f1ec/src/compositemark/index.ts#L30

Do you think there might be any way to somehow get the vega-lite package to not generate TS 2.9-specific syntax, that `import("./boxplot").BoxPlotStyle[]`, in the .d.ts file? I realize it's a bit of a short-term hack (presumably would go away when typedoc is updated, which may be weeks or months away), but it lets people that depend on vega-lite to continue using typedoc, even with the newest version.

domoritz:


domoritz:
Is it okay if we wait for the 2.6 release before releasing this?

jasongrout:
Of course it's your call. I'm curious how long before 2.6? This is the only thing blocking us in JupyterLab from moving to TS 2.9, which we'd like to do soon as well.

jasongrout:
(We'd like to do a prerelease of the next version of JupyterLab early next week, and it would be great if we could move to TS 2.9 by that time. But at the same time I realize that you may have different timeline constraints.)

domoritz:
Okay, then I will make a 2.5.2 release off a branch. We already merged things into master that require a minor version bump.

domoritz:
2.5.2 is out. I hope I didn't break anything as I had to make the tags and release manually.

jasongrout:
Thanks!

PR #3878, opened 2018-06-12
Remove default prefix

domoritz:


domoritz:
Will make a new commit.

PR #3880, opened 2018-06-13
More fixes to href cursor

melissatdiamond:
Updating documentation language, need to see if href is a mark config and should be considered and if baseEncodeEntry should be considered

### DIFF
--- a/site/docs/encoding.md
+++ b/site/docs/encoding.md
@@ -236,7 +236,7 @@ Similar to [`detail`](#detail), you can use an array of field definitions. Vega-
 {:#href}
 ## Hyperlink Channel
 
-By setting the `href` channel, a mark becomes a hyperlink. The specified URL is loaded upon a mouse click. When the `href` channel is specified, the [`cursor` mark property](mark.html#hyperlink) is set to `pointer` to serve as affordance for hyperlinks.
+By setting the `href` channel, a mark becomes a hyperlink. The specified URL is loaded upon a mouse click. When the `href` channel is specified, the [`cursor` mark property](mark.html#hyperlink) is set to "pointer" by default to serve as affordance for hyperlinks.
 
 {% include table.html props="href" source="Encoding" %}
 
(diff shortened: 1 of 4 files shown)


site/docs/encoding.md
```diff
@@ -236,7 +236,7 @@ Similar to [`detail`](#detail), you can use an array of field definitions. Vega-
 {:#href}
 ## Hyperlink Channel
 
-By setting the `href` channel, a mark becomes a hyperlink. The specified URL is loaded upon a mouse click. When the `href` channel is specified, the [`cursor` mark property](mark.html#hyperlink) is set to `pointer` to serve as affordance for hyperlinks.
```
domoritz: This should be `"pointer"`

melissatdiamond: Is it not "pointer" in the commit? Do I need a certain kind of quotation mark?

domoritz: Use \` and "

melissatdiamond: Got it

melissatdiamond:
As is the cursor does not change on this spec so I think it needs to account for config:

{
  "description": "A scatterplot showing horsepower and miles per gallons that opens a Google search for the car that you click on.",
  "config": {
    "mark": {
      "href": "http://www.google.com"
    }
  },
  "mark": {
    "type": "point"
  },
  "encoding": {
    "color": {"field": "Origin", "type": "nominal"},
    "tooltip": {"field": "Name", "type": "nominal"}
  }
}

domoritz:
Good catch. `getMarkConfig` is probably what we need to use.

melissatdiamond:
I just checked the config.mark.href in the same place as the other checks. I think getMarkConfig when passed in href returns the actual link. I could use this to check if the result is undefined if that works better than what I did here. Also, is there any place in the documentation I should change for this? I don't see anything specific about href properties related to mark configs.

PR #3881, opened 2018-06-13
Impute VL-specific features

invokesus:
- [X] `sequence` expression as `keyvals`
- [X] `frame` property over which values are imputed
- [X] `impute` property in Encoding.

### DIFF
--- a/src/compile/data/impute.ts
+++ b/src/compile/data/impute.ts
@@ -1,6 +1,6 @@
-import {ImputeTransform} from '../../transform';
+import {ImputeSequence, ImputeTransform, isImputeSequence} from '../../transform';
 import {duplicate} from '../../util';
-import {VgImputeTransform} from '../../vega.schema';
+import {VgImputeTransform, VgSignalRef} from '../../vega.schema';
 import {DataFlowNode} from './dataflow';
 
 export class ImputeTransformNode extends DataFlowNode {
@@ -10,11 +10,17 @@ export class ImputeTransformNode extends DataFlowNode {
 
   public producedFields() {
     // typescript detects true as boolean type
-    return {[this.transform.impute]: true as true} ;
+    return {[this.transform.impute]: true as true};
   }
 
   constructor(parent: DataFlowNode, private transform: ImputeTransform) {
     super(parent);
+
+  }
+
+  private processSequence(keyvals: ImputeSequence): VgSignalRef {
+    const {start, stop, step=1} = keyvals;
+    return {signal: `sequence(${start}, ${stop}, ${step})`};
   }
 
   public assemble(): VgImputeTransform {
@@ -23,10 +29,10 @@ export class ImputeTransformNode extends DataFlowNode {
       type: 'impute',
       field: impute,
       key,
-      ...(keyvals? {keyvals}: {}),
-      ...(method? {method}: {}),
-      ...(groupby? {groupby}: {}),
-      ...(value? {value}: {})
+      ...(keyvals ? {keyvals: isImputeSequence(keyvals) ? this.processSequence(keyvals) : keyvals}: {}),
+      ...(method ? {method}: {}),
+      ...(groupby ? {groupby}: {}),
+      ...(value ? {value}: {})
     };
   }
 }
--- a/src/compile/data/impute.ts
+++ b/src/compile/data/impute.ts
@@ -1,11 +1,14 @@
+import {isFieldDef} from '../../fielddef';
 import {ImputeSequence, ImputeTransform, isImputeSequence} from '../../transform';
 import {duplicate} from '../../util';
-import {VgImputeTransform, VgSignalRef} from '../../vega.schema';
+import {VgFormulaTransform, VgImputeTransform, VgSignalRef, VgWindowTransform} from '../../vega.schema';
+import {pathGroupingFields} from '../mark/mark';
+import {UnitModel} from '../unit';
 import {DataFlowNode} from './dataflow';
 
-export class ImputeTransformNode extends DataFlowNode {
+export class ImputeNode extends DataFlowNode {
   public clone() {
-    return new ImputeTransformNode(this.parent, duplicate(this.transform));
+    return new ImputeNode(this.parent, duplicate(this.transform));
   }
 
   public producedFields() {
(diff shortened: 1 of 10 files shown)


kanitw:
Can you create another PR for sequence so we can review them separately?

src/transform.ts
```diff
@@ -212,6 +212,26 @@ export interface WindowTransform {
   sort?: SortField[];
 }
 
+export interface ImputeSequence {
+  /**
+   * The starting value of the sequence.
+   */
+  start: number;
```
kanitw: I believe that Vega expression's sequence function supports a sequence with only `stop` and will start with `0` (or `1`). 

I wonder if we should support sequence with only `stop` (akin to python's `xrange()`). 

If you agree, please update the docs and modify the behavior. 
 (It's okay to always call sequence with 3 params.)

kanitw:
NVM I already start reviewing.

src/compile/data/impute.ts
```diff
@@ -23,16 +26,59 @@ export class ImputeTransformNode extends DataFlowNode {
...
+    if (transformRepresentation === undefined) {
+      return null;
+    }
+    return new ImputeTransformNode(parent, transformRepresentation);
+  }
+
+  public assemble(): [VgImputeTransform] | [VgImputeTransform, VgWindowTransform, VgFormulaTransform] {
+    const {impute, key, keyvals, method, groupby, value, frame} = this.transform;
+    if (method !== 'value' && frame !== undefined) {
+      const initialImpute: VgImputeTransform = {
```
kanitw: Please confirm if you have played with this in Vega to see if we really need additonal impute.

src/compile/data/impute.ts
```diff
@@ -1,8 +1,11 @@
+import {normalizeimputeProperties} from '../../impute';
 import {ImputeSequence, ImputeTransform, isImputeSequence} from '../../transform';
 import {duplicate} from '../../util';
-import {VgImputeTransform, VgSignalRef} from '../../vega.schema';
+import {VgFormulaTransform, VgImputeTransform, VgSignalRef, VgWindowTransform} from '../../vega.schema';
+import {UnitModel} from '../unit';
 import {DataFlowNode} from './dataflow';
 
+// TODO Rename to ImputeNode
```
kanitw: Please add another commit to just rename it in this PR so we won't forget to do so.

src/fielddef.ts
```diff
@@ -223,6 +224,12 @@ export interface PositionFieldDef<F> extends ScaleFieldDef<F> {
    * (3) At least one of non-position channels mapped to an unaggregated field that is different from x and y.  Otherwise, `null` by default.
    */
   stack?: StackOffset | null;
+  /**
```
kanitw: Add line breaks between properties.

src/fielddef.ts
```diff
@@ -223,6 +224,12 @@ export interface PositionFieldDef<F> extends ScaleFieldDef<F> {
...
+   * An object defining the properties of the Impute Operation to be applied.
+   * The field value of the other positional channel is taken as `key` of the `Impute` Operation.
+   * The field of the `color` channel if specified is used as `groupby` of the `Impute` Operation.
+   */
+  impute?: ImputeProperties;
```
kanitw: Please name this `ImputeParams`.

I think the docs here would look a bit out of place in the documentation.

src/impute.ts
```diff
@@ -0,0 +1,53 @@
...
+  // Should we add this?
+  // keyChannel?: string;
+
+  /**
+   * The imputation method to use for the field value of imputed data objects.
+   * One of `value`, `mean`, `median`, `max` or `min`.
+   *
+   * __Default value:__  `"value"`
+   */
+  method?: 'value' | 'mean' | 'median' | 'max' | 'min';
```
kanitw: Supported impute method shouldn't be defined separately.  Use defined type from Vega-Util?

src/impute.ts
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * The field value to use when the imputation `method` is `"value"`.
+   */
+  value?: any;
+
+  // Should we add this?
+  // groupby?: string;
+
+  /**
+   * The frame over which the `method` will be applied.
```
kanitw:
```
The frame over which the impute `method` will be applied.
```

We should have an example in the documentation to show this.

src/compile/data/impute.ts
```diff
@@ -1,32 +1,84 @@
...
-      ...(method? {method}: {}),
-      ...(groupby? {groupby}: {}),
-      ...(value? {value}: {})
-    };
+  public static makeFromTransform(parent: DataFlowNode, imputeTransform: ImputeTransform): ImputeTransformNode {
+    return new ImputeTransformNode (parent, imputeTransform);
+  }
+
+  public static makeFromEncoding(parent: DataFlowNode, model: UnitModel) {
+    const transformRepresentation = normalizeimputeProperties(model.encoding);
```
kanitw: `normalizeimputeProperties`  is only used here.

Since this internal logic of a compiler, let's remove the `normalizeimputeProperties` method as just move the logic here.

test/impute.test.ts
```diff
@@ -0,0 +1,69 @@
+/* tslint:disable:quotemark */
+import {assert} from 'chai';
+import {Encoding} from '../src/encoding';
+import {normalizeimputeProperties} from '../src/impute';
+
+describe('impute', () => {
```
kanitw: As I suggest that we remove `normalizeimputeProperties`, all of these tests should be moved to match the `test/compile/data/impute.test.ts`.

src/impute.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  if (isFieldDef(xDef) && isFieldDef(yDef)) {
+
+    const imputedChannel = xDef.impute ? xDef : (yDef.impute ? yDef: undefined);
+    if (imputedChannel === undefined) {
+      return undefined;
+    }
+    const keyChannel = xDef.impute ? yDef : (yDef.impute ? xDef: undefined);
+    const {method, value, frame} = imputedChannel.impute;
+    return {
```
kanitw: groupby should be automatically inferred.  See `pathGroupingFields` in `mark.ts` for the logic for line grouping.

src/transform.ts
invokesus: I think it makes sense. Should I check for the case when `stop` and `step` are specified but `start` is not or maybe infer a default?


Also, I have set a default for `step` in my code which I will remove, since that should be dealt with by Vega.

src/compile/data/impute.ts
invokesus: Do you mean having the same `VgImputeTransform` for both `frame` and regular imputes and then the 2 optional transforms if `frame` is present?

If so then I thought this was easier to follow, since sharing the `VgImputeTransform` leads to adding 3 or 4 more checks.(eg. when `method === 'value' && frame !== undefined`)

src/transform.ts
invokesus:
```
const {start, stop, step} = keyvals;
const result = [
    ...(start ? [start] : []),
    stop,
    ...(step ? [step] : [])
].join(',');
return {signal: `sequence(${result})`};
 ```
Is this okay?

Comment (kanitw) [reply src/transform.ts]: That would be wrong when stop and step are specified.  

I think it's actually better to to always call Vega expression's `sequence` with 3 params.

File: src/compile/data/impute.ts (hunk 2 above)
Comment (kanitw) [reply src/compile/data/impute.ts]: I mean, I don't know if we really need an impute at all if frame is presensted. (We might need, but you better try and tell me why.)

Comment (kanitw) [reply src/compile/data/impute.ts]: Thinking more about it I still don't like the fact that encoding impute with frame doesn't behave like encoding's impute without frame.  Perhaps we should use `undefined` as the placeholder instead of null?

Comment (invokesus) [reply src/compile/data/impute.ts]: If I understand correctly, you are not comfortable with `impute` altering the values of data objects that were present in the `source` with `null` values?
Other than using `undefined`, I can only think of using `Symbol` to generate a unique value to initially `impute` with.

Comment (kanitw) [reply src/compile/data/impute.ts]: Yes, if impute affect prior `null` values, it should always affect them (not only when `frame` is specified).

Symbol may work too.  Give it a try?

Comment (invokesus) [reply src/compile/data/impute.ts]: `Symbol` cannot be used in this case since the  `value` we are imputing with has to be converted to a `string` for use in the `formula` step.
Other than generating a random string, I think `undefined` is the only option.

File: src/compile/data/impute.ts [hunk 11]
```diff
@@ -1,14 +1,92 @@
...
 import {ImputeSequence, ImputeTransform, isImputeSequence} from '../../transform';
-import {duplicate} from '../../util';
+import {duplicate, keys} from '../../util';
 import {VgFormulaTransform, VgImputeTransform, VgSignalRef, VgWindowTransform} from '../../vega.schema';
 import {UnitModel} from '../unit';
 import {DataFlowNode} from './dataflow';
 
-// TODO Rename to ImputeNode
-export class ImputeTransformNode extends DataFlowNode {
+export function normalizeimputeProperties(encoding: Encoding<string>): ImputeTransform | undefined {
```
kanitw: normalizeImputeProperties

src/compile/data/impute.ts
```diff
@@ -1,14 +1,92 @@
...
+      ...(value ? {value} : {}),
+      ...(frame ? {frame} : {}),
+      ...(groupbyFields.length ? {groupby: groupbyFields} : {} )
+    };
+  }
+  return undefined;
+}
+
+// Behavior almost identical to pathGroupingFields but without the Mark argument
+export function deriveEncodingGroupby(encoding: Encoding<string>): string[] {
```
kanitw: Why without the mark argument? 

1) This is a part of impute in encoding, which means you have the mark in your context.
2) Even if you really shouldn't have mark parameter here, copying almost identical code like this will cause maintainability issue in the future.

src/compile/data/impute.ts
kanitw: Also, https://github.com/vega/vega-lite/pull/3881/files/08844a9f23a0eaa158969ebe538e7a841367de7c..ddd334dd544ca99792df7648c2537e08783604ba#r195305085

kanitw: Ah I get what do you now.

kanitw: Instead of having a separate function here why don't you copy all the code in this method to be in `makeFromEncoding`?  I don't see why we need an additional code jump here.

src/impute.ts
```diff
@@ -1,53 +1,22 @@
-import {Encoding} from './encoding';
-import {isFieldDef} from './fielddef';
-import {ImputeTransform} from './transform';
+import {ImputeMethods} from './vega.schema';
 
-export interface ImputeProperties {
-
-  // Should we add this?
-  // keyChannel?: string;
+export interface ImputeParams {
```
kanitw: Some of the properties here are shared with `ImputeTransform`, but are not shared in the code.  Please make sure they are shared by adding `ImputeBase` interface.  This way, any changes to the property docs would be always in sync for both the `ImputeParams` and `ImputeTransform`.

kanitw:
Also please improve test coverage

src/impute.ts
invokesus: Should I put `ImputeBase` in `src/transform.ts` or `src/impute.ts`?

kanitw: Let's say `impute.ts`

src/transform.ts
```diff
@@ -233,7 +233,7 @@ export function isImputeSequence(t: ImputeSequence | any[] | undefined): t is Im
   return t && t['start'] !== undefined && t['stop'] !== undefined;
 }
 
-export interface ImputeTransform {
+export interface ImputeTransform extends ImputeBase {
```
kanitw: This is wrong.  ImputeTransform doesn't have `frame`.

kanitw: Oh I see -- you actually want VL impute transform to support more than Vega.  That's a good idea.

Issue #3266, opened 2018-01-14
Add all missing mark properties that are already supported by Vega 


kanitw:
From https://vega.github.io/vega/docs/marks/rule/, I saw `strokeCap`, `strokeDash`, `strokeJoin`, `strokeMiterLimit`. 

We should systematically inspect Vega docs for all marks and add all of missing mark properties.

PR #3883, opened 2018-06-14
Add missing mark properties

melissatdiamond:
Issue #3266:

Questions:
1. Should `strokeJoin` and `strokeMiterLimit` be in the list `STROKE_CONFIG` on line 124 of mark.ts? `strokeDash` and `strokeDashOffset` are there but `strokeCap` is not.
2. I added `cornerRadius` to the view config in `config.ts`, but it looks strange when used because the axes do not curve while the inner rectangle does. Is there a way to fix this?

TODO: Update documentation

### DIFF
--- a/src/vega.schema.ts
+++ b/src/vega.schema.ts
@@ -267,7 +267,7 @@ export interface VgSignal {
   push?: string;
 }
 
-export type VgEncodeChannel = 'x'|'x2'|'xc'|'width'|'y'|'y2'|'yc'|'height'|'opacity'|'fill'|'fillOpacity'|'stroke'|'strokeWidth'|'strokeCap'|'strokeOpacity'|'strokeDash'|'strokeDashOffset'|'cursor'|'clip'|'size'|'shape'|'path'|'innerRadius'|'outerRadius'|'startAngle'|'endAngle'|'interpolate'|'tension'|'orient'|'url'|'align'|'baseline'|'text'|'dir'|'ellipsis'|'limit'|'dx'|'dy'|'radius'|'theta'|'angle'|'font'|'fontSize'|'fontWeight'|'fontStyle'|'tooltip'|'href'|'cursor'|'defined';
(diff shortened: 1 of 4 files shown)


domoritz:
> I added cornerRadius to the view config in config.ts, but it looks strange when used because the axes do not curve while the inner rectangle does. Is there a way to fix this?

Can you add a screenshot?

melissatdiamond:
screen shot 2018-06-14 at 1 49 25 pm
 spec:
```
{
  "config": {
      "view": {
          "cornerRadius": 15
      }
  },
  "mark": "line",
  "encoding": {
  }
}
```

domoritz:
Hmm, let's not support cornerRadius for the for now then.

melissatdiamond:
Sounds good. It still works for bar and tick, I'll just take it out of the view config.

src/vega.schema.ts
```diff
@@ -972,6 +972,8 @@ export type Cursor = 'auto' | 'default' | 'none' |
   'row-resize' | 'all-scroll' | 'zoom-in' |
   'zoom-out' | 'grab' | 'grabbing';
 export type StrokeCap = 'butt' | 'round' | 'square';
+export type StrokeJoin = 'miter' | 'round' | 'bevel';
```
kanitw: Shouldn't we use typings from Vega typing?

kanitw:
LGTM besides the minor comment.

src/vega.schema.ts
domoritz: We don't have it over there yet. Vega-typings just use string for now so we need to fix that first.

kanitw:
@melissatdiamond In the future, please say "fix #3266" in the PR description when your PR actually fixes the whole issue.

melissatdiamond:
I'm currently updating the documentation to account for these changes. Should I make another PR for this update?

domoritz:
Yes, separate PR.

Issue #2446, opened 2017-06-04
Facet customization


kanitw:
Facet operator still lacks ways for customizing many underlying properties.

- [x] labels / title configs like https://vega.github.io/vega-lite/docs/axis.html#labels-config and https://vega.github.io/vega-lite/docs/axis.html#title-config.
  - [x] Some properties such as labelBaseline are still missing

Steps:
1. Add `HeaderConfig` to `Config` like `AxisConfig`
2. Add the output
- For title properties, add the right output to the output `encode` block in `getTitleGroup()` in `header.ts`
- For label properties, add the right output to the output `encode` block in `getTitleGroup()` in `header.ts`
- (In both cases, writing a helper method to enumerate through all source VL properties (e.g., `labelLimit` and target VG properties (e.g., `limit`) could be helpful. 
4. Write unit tests for both cases. 

-----

- [ ] Header orient 

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A bar chart of the Berkeley Admissions data, showing Simpson's Paradox",
  "width": 100,
  "data": {
    "values": [
      {"Department": "A", "gender" : "Male", "value" : 62.06},
      {"Department": "A", "gender" : "Female", "value" : 82.41},
      {"Department": "B", "gender" : "Male", "value" : 5.90},
      {"Department": "B", "gender" : "Female", "value" : 7.04},
      {"Department": "C", "gender" : "Male", "value" : 33.09},
      {"Department": "C", "gender" : "Female", "value" : 34.93},
      {"Department": "D", "gender" : "Male", "value" : 63.04},
      {"Department": "D", "gender" : "Female", "value" : 68.00},
      {"Department": "E", "gender" : "Male", "value" : 36.92},
      {"Department": "E", "gender" : "Female", "value" : 34.06},
      {"Department": "C", "gender" : "Male", "value" : 27.75},
      {"Department": "C", "gender" : "Female", "value" : 23.92}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "gender", "type": "nominal", 
    "axis" : null},
    "y": {"field": "value", "type": "quantitative",
      "axis" : {"title": "Acceptance Rate  (%)"}},
    "color": {"field": "gender", "type" : "nominal", 
    "scale": {
      "range" : ["pink","steelblue"]
    }
    },
    "column" : { "field": "Department", "type": "nominal", "header":{"orient": "bottom"}}
  }
}
```

[...]

PR #3888, opened 2018-06-15
Add Label and Title Properties to Facet Header 

melissatdiamond:
Working on fixing #2446:

In this initial commit, I created a top level `header.ts` file and moved classes from the lower level header file (`src/compile/layout/header.ts`) to the new file. Will later add `HeaderConfig` to this file (like `AxisConfig` in axis.ts). This change causes a runtime error when running yarn test I'm not sure about.


src/header.ts
```diff
@@ -0,0 +1,46 @@
...
+export type HeaderChannel = 'row' | 'column';
+export const HEADER_CHANNELS: HeaderChannel[] = ['row', 'column'];
+
+export type HeaderType = 'header' | 'footer';
+export const HEADER_TYPES: HeaderType[] = ['header', 'footer'];
+
+/**
+ * A component that represents all header, footers and title of a Vega group with layout directive.
+ */
+export interface LayoutHeaderComponent {
```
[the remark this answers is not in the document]
melissatdiamond: Gotcha. So would `HeaderConfig` go in the new `src/compile/header/index` then?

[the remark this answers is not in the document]
kanitw: If something is a part of the Vega-Lite specs, then it's "external", and won't be in `src/compile`.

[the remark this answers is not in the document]
melissatdiamond: Ok, so then do we still need an external header file for `HeaderConfig` in addition to the renamed internal file?

src/compile/header/index.ts
```diff
@@ -0,0 +1,180 @@
+/**
```
kanitw: Shouldn't this be moved from another file instead of just adding them?

src/compile/header/index.ts
```diff
@@ -75,7 +75,7 @@ export function getTitleGroup(model: Model, channel: HeaderChannel) {
     // also make sure that guide-title config override these Vega-lite default
   };
 
-  getHeaderTitleProperties(config, update);
+  getHeaderConfigProperties(config, update);
```
kanitw: This method is causing side effect.  Perhaps it's better to not pass update in and use

test/compile/header/index.test.ts
```diff
@@ -11,7 +11,7 @@ describe('compile/layout/header', () => {
   });
 
   describe('label baseline adjusted according to angle', () => {
-    assert.deepEqual(labelBaseline(10), {});
+    assert.deepEqual(labelBaseline(10), {baseline: {value: 'middle'}});
```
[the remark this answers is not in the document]
melissatdiamond: no I forgot to undo this change, which I think caused the tests to fail.

src/header.ts
```diff
@@ -0,0 +1,122 @@
...
+   */
+  titleAlign?: string;
+
+  /**
+   * Angle in degrees of the header title.
+   */
+  titleAngle?: number;
+  /**
+   * Vertical text baseline for the header title.
+   */
```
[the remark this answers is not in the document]
melissatdiamond: I didn't see anything more specific in the Vega-Lite Axis Config documentation, which is what I have been going off of.

src/header.ts
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * The font size of the header label, in pixels.
+   *
+   * @minimum 0
+   */
+  labelFontSize?: number;
+
+  /**
+   * Maximum allowed pixel width of header labels.
+   */
```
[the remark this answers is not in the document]
melissatdiamond: That is what the Vega-Lite spec says for Axis Config, but there is this more specific description of limit in relation to text marks:

"The maximum length of the text mark in pixels (default 0, indicating no limit). The text value will be automatically truncated if the rendered size exceeds the limit."

Should I add something like this here?

src/header.ts
```diff
@@ -0,0 +1,122 @@
...
+  // ---------- Title ----------
+  /**
+   * Horizontal text alignment of the header title.
+   */
+  titleAlign?: string;
+
+  /**
+   * Angle in degrees of the header title.
+   */
+  titleAngle?: number;
```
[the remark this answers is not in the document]
melissatdiamond: Will add this.

src/header.ts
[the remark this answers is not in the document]
domoritz: Look at the types. If we can, we should make this type more specific.

src/header.ts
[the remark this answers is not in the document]
domoritz: Make the text consistent (maybe need to adjust the axis config as well).

src/header.ts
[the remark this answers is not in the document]
melissatdiamond: Found it.

src/header.ts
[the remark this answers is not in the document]
melissatdiamond: Will do.

src/compile/header/index.ts
```diff
@@ -138,6 +159,11 @@ function getHeaderGroup(model: Model, channel: HeaderChannel, headerType: Header
 
       };
 
+      const update = {
+        ...defaults,
+        ...getHeaderLabelProperties(config)
```
kanitw: Note that title has a lot of native properties supported (see https://vega.github.io/vega/docs/title/). 

This means that in many cases you may not need to use `encode.update`

kanitw: From a quick skim, it looks like all title and label properties you're adding except (`x` and `y`) are supported by `title`. Thus, I wonder if it is better to map them directly to Vega `title` directive and drop supports for `x` and `y`. (They aren't very useful anyway.)

src/header.ts
```diff
@@ -0,0 +1,122 @@
...
+  titleY: 'y',
+};
+
+export const HEADER_LABEL_PROPERTIES_MAP = {
+  labelAngle: 'angle',
+  labelColor: 'fill',
+  labelFont: 'font',
+  labelFontSize: 'fontSize',
+  labelLimit: 'limit',
+  labelOverlap: 'overlap',
```
kanitw: Can you really support labelOverlap here?  I suspect that you can't.

src/header.ts
```diff
@@ -0,0 +1,122 @@
+import {Flag, flagKeys, keys} from './util';
+import {FontWeight, TitleOrient} from './vega.schema';
+
+export const HEADER_TITLE_PROPERTIES_MAP = {
```
kanitw: Define type to be keyOf HeaderBase

src/header.ts
```diff
@@ -0,0 +1,122 @@
...
+  titleColor: 'fill',
+  titleFont: 'font',
+  titleFontSize: 'fontSize',
+  titleFontWeight: 'fontWeight',
+  titleLimit: 'limit',
+  titleX: 'x',
+  titleY: 'y',
+};
+
+export const HEADER_LABEL_PROPERTIES_MAP = {
```
kanitw: Define type to be map from a key of `HeaderBase` to a property of `VgTitle`

src/compile/header/index.ts
domoritz: Should we do this in a follow up or refactor this PR?

src/header.ts
domoritz: I don't think so. This will be a problem when you create a grouped bar chart with row or column.

melissatdiamond: So I should remove this?

domoritz: Yes. Make sure you only include properties that work in Vega.

src/header.ts
melissatdiamond: Where would `HeaderBase` be defined and what should go there?

src/compile/header/index.ts
kanitw: This PR -- This is all new and should be easy to do.

src/header.ts
kanitw: I mean `HeaderConfig`, sorry.

kanitw:
@melissatdiamond -- FYI, we modify `src/compile/layout/header.ts` and its test file a bit in https://github.com/vega/vega-lite/pull/3854.  

You may want to
1) Squash this PR into one commit (the commits aren't really atomic anyway)
2) After squashing, rebase to the latest master while preserving the changes from https://github.com/vega/vega-lite/pull/3854.

src/header.ts
melissatdiamond: Got it.

src/compile/header/index.ts
```diff
@@ -219,12 +193,17 @@ function getHeaderGroup(model: Model, channel: HeaderChannel, headerType: Header
   return null;
 }
 
-function getHeaderLabelProperties(config: Config) {
+export function getHeaderProperties(config: Config, facetFieldDef: FacetFieldDef<string>, properties: string[], propertiesMap: {[k in keyof HeaderConfig]: keyof VgTitleConfig}) {
```
kanitw: Should each property in `properties` be a key of HeaderConfig?

src/compile/header/index.ts
```diff
@@ -145,27 +135,15 @@ function getSort(facetFieldDef: FacetFieldDef<string>): VgComparator {
...
+export function getHeaderGroup(model: Model, channel: HeaderChannel, headerType: HeaderType, layoutHeader: LayoutHeaderComponent, headerCmpt: HeaderComponent) {
   if (headerCmpt) {
     let title = null;
     const {facetFieldDef} = layoutHeader;
     if (facetFieldDef && headerCmpt.labels) {
       const {header = {}} = facetFieldDef;
       const {format, labelAngle} = header;
       const config = model.config? model.config : undefined;
-      const defaults = {
+      const update = {
```
kanitw: Do we really need to keep these `update` block?  Can't we use the angle, anchor, and baseline of the `title` block?

src/compile/header/index.ts
melissatdiamond: The types only seem to work when `properties` is a `string[]` because the lists I pass in are created by calling `Object.keys` on a map, which seems to return a `string[]`. Is there a way around this?

src/compile/header/index.ts
```diff
@@ -98,15 +97,15 @@ export function getHeaderGroups(model: Model, channel: HeaderChannel): VgMarkGro
 
 // 0, (0,90), 90, (90, 180), 180, (180, 270), 270, (270, 0)
 
-export function labelAlign(angle: number) {
+export function labelAnchor(angle: number) {
```
kanitw: Do we still have labelAngle property? 

If so, I wonder if replacing labelAlign with labelAnchor rule still make them nice?

src/compile/header/index.ts
```diff
@@ -143,22 +142,18 @@ export function getHeaderGroup(model: Model, channel: HeaderChannel, headerType:
...
 
       title = {
         text: formatSignalRef(facetFieldDef, format, 'parent', model.config),
         offset: 10,
         orient: channel === 'row' ? 'left' : 'top',
         style: 'guide-label',
-        ...getHeaderProperties(config, facetFieldDef, HEADER_LABEL_PROPERTIES, HEADER_LABEL_PROPERTIES_MAP),
-        ...(keys(update).length > 0 ? {encode: {update}} : {})
+        ...(
+          labelAngle !== undefined ? {angle: {value: labelAngle}} : {}
```
kanitw: Why do we need to wrap it with `{value: ...}` here?

src/compile/header/index.ts
melissatdiamond: I'm playing with it and it looks like anchor does not do the same thing as align. Should I change it back?

kanitw: Maybe, let's look at it together at the scrum.

src/compile/header/index.ts
melissatdiamond: I changed the type using `util.keys` and it worked locally and passed header tests, but this change caused many of the other tests to fail because of type errors so I left it as is.

kanitw:
I'm merging this to `feature/header-label-title`.  Please rebase the doc branch and change its merge target to `feature/header-label-title` too.

PR #3894, opened 2018-06-19
added "Vega Preview" a Visual Studio Code extension to applications.md

8:
Hi,
I've written an Extension for Visual Studio Code that enables a live preview for Vega and Vega-Lite. I've seen that you encourage a pull requests against the documentation against `applications.md` if we have something to add, so that's what I've done.

The extension is available inside Visual Studio Code or at https://marketplace.visualstudio.com/items?itemName=mdk.vega-preview
The source code is available at: https://github.com/8/vega-preview

Take care,
Martin


domoritz:
I see. Instead of using an svg output, you'd need to have an HTML page. Then embed will help you a lot. 

Since the nehancement is unrelated to this issue, I'll merge it.

domoritz:
Before I do that, why do you say that the extension is for editing Vega and Vega-Lite visualizations? As far as I see it, it's only for viewing them. Maybe add a specifier that you can only view static visualizations.

Issue #3891, opened 2018-06-18
Row header label's default baseline should be middle


kanitw:
See the facet_bullet example in https://github.com/vega/vega-lite/pull/3717/files for example. 

Once we fix this, we should add the facet_bullet example to the gallery. :)

kanitw:
Fixed in #3895

PR #3895, opened 2018-06-19
Set row header label's default to "middle"

melissatdiamond:
Fix issue #3891


kanitw:
Rebase and compile examples again?

examples/compiled/bar_grouped.vg.json
```diff
@@ -144,7 +144,14 @@
                 },
                 "offset": 10,
                 "orient": "top",
-                "style": "guide-label"
+                "style": "guide-label",
+                "encode": {
```
kanitw: We can actually use `title` directive's `baseline` property.
However, it seems like this is not the problem introduced by this PR. Thus, let's refactor this in a separate PR.  (Perhaps as a part of https://github.com/vega/vega-lite/pull/3888?)

melissatdiamond:
It seems like the only example that is not working is the facet bullet example? I'm not sure why that would be the case.

kanitw:
Test still failing.

kanitw:
Please:

1) Fix this

image

2) squash the second rebuild examples into the commit the make changes.

3) As stated in #3891, add the facet_bullet example to the gallery. (Otherwise this issue won't really fix #3891)

domoritz:
Ready @kanitw

kanitw:
Hmm there is now a conflict.  Rebase, re-compile example, and squash my commit to be a part of the "Add facet_bullet to the gallery" commit?

domoritz:
@melissatdiamond Can you squash the last two commits?

domoritz:
@kanitw Please merge.

PR #3899, opened 2018-06-20
Add weather heatmap example.

Saba9:
Part of issue #1486. Heatmap is almost identical to the [one found in the Altair user guide](https://altair-viz.github.io/user_guide/transform.html#timeunit-transform) except it uses the San Francisco data rather than the Seattle data.

image


_data/examples.json
```diff
@@ -196,6 +196,10 @@
         "title": "Table Heatmap",
         "style": "background-size: auto"
       },
+      {
+        "name": "rect_heatmap_weather",
+        "title": "Table Heatmap of Weather"
```
kanitw: I wonder if we should call this "Annual Weather Heatmap"?

(We can keep file name the same.)

Saba9: That sounds good to me. I can go ahead and update the title.

g3o2:
maybe the gridlines could be made slightly thinner?

kanitw:
> maybe the gridlines could be made slightly thinner?

These are actuall gap between rect bands.

@saba9 Perhaps remove the top level width and height and set rangeStep of x and y scales to the same values so we get square rects with equal gap (more beautiful)

PR #3902, opened 2018-06-20
Create Error Bar Document

chanwutk:


site/docs/mark/errorbar.md
```diff
@@ -0,0 +1,154 @@
...
+
+Besides the properties listed above, `"rule"` and `"ticks"` can be used to specify the underlying [mark properties](mark.html#mark-def) for different [parts of the error bar](#parts) as well.
+
+
+## Using Error Bars to Aggregate Raw Data
+
+If the data is not aggregated yet, Vega-Lite will aggregate the data based on the `extent` and `center` properties in the mark definition.
+
+
+1) __Standard Error Error Bar__ is the default error bar in Vega-Lite, or cen be specified by setting `extent` to `stderr`. The size of lower and upper rules are standard error. As a default, the rules expand from the mean; however, `center` can be set to `median`, so that the rules expand from the median instead.
```
kanitw: Change all of "XXX Error Bar" to "Error bar showing XXX"?

kanitw:
1) I think it's weird that the whole errorbar page we never show error bar with main marks at all.  We should change them to the use the layered examples (and thus we can also remove most of the standalone errorbar examples.) 

This way we can also explain that users will likely use ci, stdev, stderr with mean and iqr with median points. 

2) If you think adding layer of point mark makes spec long, then remove the following part as we don't really need them:

```
      "scale": {"zero": false},
      "title": "Barley Yield"
```

3) median with stderr is a bad idea. (I'm sort of okay with supporting them as some experts may have exceptional case to use them but don't mention about them in the documentation to suggest novice users bad ideas.) 

4) For keys like `extent`, we can omit quotes (like in Javascript), but for values like `"stdev"`, please always include quotes (as omitting quotes in JS won't work.) 

5) Not sure why the following notes is under the "Error Bar showing Standard Error" seciton.

> Note: if the center is median, and the extent is not specified, the default extent is iqr instead of stderr

Better just describe it in "4) Error Bar showing Interquartile".

kanitw:
7) An example of a errorbar where the color encoding channel is specified.

- You specify color of mark definition, not encoding channel in the example
- When demonstrating a color, use color clear color (e.g., our default blue color -- see the config or just plain red/orange/blue or something). Using gray makes it looks like not customized (the default color is also gray, thought a slightly different shade of gray).

PR #3903, opened 2018-06-20
Sample Docs

invokesus:


site/docs/transform/sample.md
```diff
@@ -0,0 +1,26 @@
+---
+layout: docs
+menu: docs
+title: Sample
+permalink: /docs/sample.html
+---
+
+The **sample** transform randomly samples a data stream to create a smaller stream. As input data objects are added and removed, the sampled values may change in first-in, first-out manner. This transform uses [reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling) to maintain a representative sample of the stream.
```
domoritz: We don't call it stream in Vega-Lite.

site/docs/transform/sample.md
```diff
@@ -0,0 +1,26 @@
...
+{"sample": 500}
+```
+
+Filters a data stream to a random sample of at most 500 data objects.
+
+## Example
+
+Comparison between plots of the complete data and sampled data.
+
+<div class="vl-example" data-name="hconcat_bar_stacked_manchuria_sample"></div>
```
domoritz: This is an odd example. Let's use  a scatterplot with and without sampling.

site/docs/transform/sample.md
invokesus: > The **sample** transform randomly samples a data stream to create a smaller stream.

The sample transform randomly samples a data source to create a smaller source.
Is this ok?

kanitw:
@domoritz I'll let you merge this since you have comments earlier.

PR #3904, opened 2018-06-20
Add Fold Docs

invokesus:


site/docs/transform/fold.md
```diff
@@ -0,0 +1,48 @@
+---
+layout: docs
+menu: docs
+title: Fold
+permalink: /docs/fold.html
+---
+
+The **fold** transform collapses (or "folds") one or more data fields into two properties: a _key_ property (containing the original data field name) and a _value_ property (containing the data value).
```
kanitw: We don't bold the transform name in existing docs.  If you think it is important to bold, submit a PR to do the same for docs of existing transforms

invokesus: Removing for now. Will make a PR later to bold the transform names once the new ones have been merged into master.

PR #3908, opened 2018-06-21
Create Error Band Document

chanwutk:
screen shot 2018-06-20 at 17 30 00
@kanitw @domoritz Should we call this vertical error band or horizontal error band?

It looks horizontal, but if we follow the error bar, this is vertical.


domoritz:
Orient is vertical, right?

chanwutk:
Yes, the Orient is vertical.

chanwutk:
@domoritz I misunderstood your question. Error band does not add orient property to the area/line. The orient of the error band is to determine the axis that would be aggregated.

kanitw:
> @domoritz I misunderstood your question. Error band does not add orient property to the area/line. The orient of the error band is to determine the axis that would be aggregated.

The question is what is the underlying (implicit) area orientation. You can test this by editing the `orient` in the normalized spec.  The right one shouldn't modify the plot.

PR #3909, opened 2018-06-21
Refactor error bar and error band

chanwutk:
Refactor error bar and error band as in issue #3770 .


src/encoding.ts
```diff
@@ -230,7 +232,8 @@ export function extractTransformsFromEncoding(oldEncoding: Encoding<string>) {
       // now the field should refer to post-transformed field instead
       encoding[channel] = {
         field: vgField(channelDef),
-        type: channelDef.type
+        type: channelDef.type,
+        title: title(channelDef, config)
```
kanitw: Actually, if there is neither bin not timeUnit, we shouldn't have to add this custom title.

PR #3914, opened 2018-06-21
Minor fix to the window comment

domoritz:


src/transform.ts
```diff
@@ -202,7 +202,7 @@ export interface WindowTransform {
   ignorePeers?: boolean;
 
   /**
-   * The data fields for partitioning the data objects into separate windows. If unspecified, all data points will be a single group.
```
kanitw: be in a single group?

PR #3916, opened 2018-06-21
Refactor Composite Marks

chanwutk:


examples/specs/errorband_ci_line.vl.json
```diff
@@ -1,37 +1,21 @@
 {
```
kanitw: Can we name the file `layer_line_errorband_ci`? (same for others since it's more like layering line with errorband_ci than the other way round.)

chanwutk: Should I name the scatter plot graph with global mean and standard deviation `scatter` or `point`?

test/compositemark/errorbar.test.ts
```diff
@@ -381,28 +381,26 @@ describe('normalizeErrorBar with raw data input', () => {
...
-      [false, false], [false, false], [false, true], [true, true], [false, false]
-    ],
-    [
-      [true, false], [true, false], [true, true], [false, true], [false, false]
-    ],
-    [
-      [false, false], [false, false], [false, false], [false, false], [false, false]
-    ]
+   //    stderr           stdev           ci            iqr         undefined
+    [[false, false], [false, false], [false, true], [true, true], [false, false]],   // mean
```
kanitw: Whoa -- this test seems very unreadable.

kanitw: Can you instead refactor it to use dictionary / map?

chanwutk: Yes, it is now using dictionary.

PR #3918, opened 2018-06-22
Include example coverage in total coverage

kanitw:
I start feeling like this will make us write less redundant tests?


kanitw:
Lets not do this

PR #3926, opened 2018-06-24
Sh/prebin

sirahd:
- I already finished implementing logic for `bar`, `circle`, `point`, `rect`, `square`
- For `line` and `tick`, there's a little bit of issue because it cannot automatically determine the orientation given for example `x (bin_start)` `x2 (bin_end)` and `y (count)` (warning will be displayed, but it will output something)
- `area` is not implemented yet

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -6,8 +6,10 @@
 import {RangeType} from './compile/scale/type';
 import {Encoding} from './encoding';
 import {FacetMapping} from './facet';
+import {isFieldDef} from './fielddef';
+import {CIRCLE, POINT} from './mark';
 import {Mark} from './mark';
-import {Flag, flagKeys} from './util';
+import {contains, Flag, flagKeys} from './util';
 
 export namespace Channel {
   // Facet
@@ -218,8 +220,16 @@ export type SupportedMark = {
  * @param mark the mark type
  * @return whether the mark supports the channel
  */
-export function supportMark(channel: Channel, mark: Mark) {
-  return mark in getSupportedMark(channel);
+export function supportMark(encoding: Encoding<string>, channel: Channel, mark: Mark) {
+  if (contains([CIRCLE, POINT], mark) && channel === X2) {
+    if (isFieldDef(encoding.x) && encoding.x.scale && encoding.x.scale.binned) {
+      return true;
+    } else {
+      return false;
+    }
+  } else {
+    return mark in getSupportedMark(channel);
+  }
 }
 
 /**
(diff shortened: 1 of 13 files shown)


src/channel.ts
```diff
@@ -218,8 +220,16 @@ export type SupportedMark = {
  * @param mark the mark type
  * @return whether the mark supports the channel
  */
-export function supportMark(channel: Channel, mark: Mark) {
-  return mark in getSupportedMark(channel);
+export function supportMark(encoding: Encoding<string>, channel: Channel, mark: Mark) {
+  if (contains([CIRCLE, POINT, SQUARE, TICK, LINE], mark) && channel === X2) {
```
kanitw: Why only X2?  (What about Y2?)

src/compile/axis/parse.ts
```diff
@@ -242,7 +242,11 @@ function getProperty<K extends keyof AxisComponentProps>(property: K, specifiedA
       return numberFormat(fieldDef, specifiedAxis.format, model.config);
     case 'grid': {
       const scaleType = model.getScaleComponent(channel).get('type');
-      return getSpecifiedOrDefaultValue(specifiedAxis.grid, properties.grid(scaleType, fieldDef));
+      if (model.specifiedScales[channel].binned) {
```
kanitw: Hmm if binned is specified via `"bin"`, should `binned` really still be a part of the specified scale?  (Esp. that we still have access to the fieldDef here.)

src/compile/mark/mixins.ts
```diff
@@ -283,7 +284,27 @@ export function centeredBandPosition(channel: 'x' | 'y', model: UnitModel, defau
...
+ * Return mixins for binned scale with position channel
+ */
+export function binnedPosition(
+  fieldDef: FieldDef<string>, fieldDef2: FieldDef<string>, channel: 'x'|'y',
+  scaleName: string, spacing: number, reverse: boolean
+) {
+  if (channel === 'x') {
+    return {
+      x2: ref.fieldRef(fieldDef2, scaleName, {}),
+      x: ref.fieldRef(fieldDef, scaleName, {}, {offset: reverse ? -spacing : spacing})
```
kanitw: 1) Why is the offset reverse different from the `binPosition` below?

src/compile/axis/parse.ts
kanitw: Ah just notice that you're implementing `scale.binned = true`, not `bin: 'binned'` that we finally agreed on.

PR #3928, opened 2018-06-24
Revise overlay mark defs

### LINKED ISSUE #3927: Clipping line of area mark
```javascript
{
  "mark": {
   "type": "area",
   "line": true, //{"clip": true}--> [Warning] Validation: /mark should match some schema in anyOf
   "clip": true, 
   "interpolate": "step-after" // <--Now works as expected :) thanks!
  },
  "encoding": {
    "x": {"field": "date", "type": "temporal",
    "scale": {
        "domain": [
          {"year": 2006},
          {"year": 2009}
        ]}
    },
  }
}
```
export 12
I expected that clip will be inherited from the area description like "interpolate" does now. At least `"line": {"clip": true}` should not violate schema.

By the way `"line": {"interpolate": "basis"}` would not have any effects or errors that is backward incompatible.

domoritz:
Support complete mark defs in the line and point overlays. Also correctly
inherit all properties from the main mark def. Fixes #3927

### DIFF
--- a/src/mark.ts
+++ b/src/mark.ts
@@ -205,6 +205,8 @@ export interface BarConfig extends BarBinSpacingMixins, MarkConfig {
   discreteBandSize?: number;
 }
 
+export type OverlayMarkDef = MarkConfig & MarkDefMixins;
+
 export interface PointOverlayMixins {
   /**
    * A flag for overlaying points on top of line or area marks, or an object defining the properties of the overlayed points.
@@ -217,7 +219,7 @@ export interface PointOverlayMixins {
    *
    * __Default value:__ `false`.
    */
-  point?: boolean | MarkConfig | 'transparent';
+  point?: boolean | OverlayMarkDef | 'transparent';
 }
 
 export interface LineConfig extends MarkConfig, PointOverlayMixins {}
@@ -232,7 +234,7 @@ export interface LineOverlayMixins {
    *
    * __Default value:__ `false`.
    */
-  line?: boolean | MarkConfig;
+  line?: boolean | OverlayMarkDef;
 }
 
 export interface AreaConfig extends MarkConfig, PointOverlayMixins, LineOverlayMixins {}
@@ -248,15 +250,7 @@ export interface TickThicknessMixins {
   thickness?: number;
 }
 
-// Point/Line OverlayMixins are only for area, line, and trail but we don't want to declare multiple types of MarkDef
-export interface MarkDef extends BarBinSpacingMixins, MarkConfig, PointOverlayMixins, LineOverlayMixins, TickThicknessMixins {
-  /**
-   * The mark type.
-   * One of `"bar"`, `"circle"`, `"square"`, `"tick"`, `"line"`,
-   * `"area"`, `"point"`, `"geoshape"`, `"rule"`, and `"text"`.
-   */
-  type: Mark;
-
+export interface MarkDefMixins {
   /**
    * A string or array of strings indicating the name of custom styles to apply to the mark. A style is a named collection of mark property defaults defined within the [style configuration](https://vega.github.io/vega-lite/docs/mark.html#style-config). If style is an array, later styles will override earlier styles. Any [mark properties](https://vega.github.io/vega-lite/docs/encoding.html#mark-prop) explicitly defined within the `encoding` will override a style default.
    *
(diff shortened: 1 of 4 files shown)


src/mark.ts
```diff
@@ -248,15 +250,7 @@ export interface TickThicknessMixins {
...
-// Point/Line OverlayMixins are only for area, line, and trail but we don't want to declare multiple types of MarkDef
-export interface MarkDef extends BarBinSpacingMixins, MarkConfig, PointOverlayMixins, LineOverlayMixins, TickThicknessMixins {
-  /**
-   * The mark type.
-   * One of `"bar"`, `"circle"`, `"square"`, `"tick"`, `"line"`,
-   * `"area"`, `"point"`, `"geoshape"`, `"rule"`, and `"text"`.
-   */
-  type: Mark;
-
+export interface MarkDefMixings {
```
kanitw: Mixins*

src/spec.ts
```diff
@@ -591,24 +590,26 @@ function normalizePathOverlay(spec: NormalizedUnitSpec, config: Config = {}): No
   }
 
   if (lineOverlay) {
-    const {interpolate} = markDef;
+    const lineMarkDef = omit(markDef, ['line']);
```
kanitw: I don't think we should really inherit everything from the main mark.  For example, if you set the main area mark's opacity, it should not affect the overlay line's opacity.  Similar if you set the main line mark's opacity, it should not affect overlay point's opacity.  

Plus, separate them mean you can customize both of them. If you make the overlay derive all properties, then we require user to specify properties of both the main layer and the overlay layer in order to set *only* the main layer.  I think that's an undesired behavior.

Instead you could have just inherited `clip`.

domoritz: Okay, I will inherit clip and interpolate.

domoritz:
Reviesed.

src/mark.ts
```diff
@@ -205,6 +205,8 @@ export interface BarConfig extends BarBinSpacingMixins, MarkConfig {
   discreteBandSize?: number;
 }
 
+export type OverlayMarkDef = MarkConfig & MarkDefMixins;
```
domoritz: Fixed typo

src/spec.ts
```diff
@@ -591,13 +590,12 @@ function normalizePathOverlay(spec: NormalizedUnitSpec, config: Config = {}): No
   }
 
   if (lineOverlay) {
-    const {interpolate} = markDef;
     layer.push({
       ...(projection ? {projection} : {}),
       mark: {
+        ...pick(markDef, ['clip', 'interpolate']),
```
kanitw: Not sure why you have to change the order here. I think putting mark type at the top is more desirable.

src/spec.ts
```diff
@@ -606,9 +604,9 @@ function normalizePathOverlay(spec: NormalizedUnitSpec, config: Config = {}): No
     layer.push({
       ...(projection ? {projection} : {}),
       mark: {
-        type: 'point',
         opacity: 1,
         filled: true,
+        type: 'point',
```
kanitw: Not sure why you have to change the order here. I think putting mark type at the top is more desirable.

domoritz: Consistency with the other one. If we want to copy some properties from the main mark def, we want to do it after opacity and filled.

src/spec.ts
domoritz: Only if we never copy the whole properties object.

domoritz: Changed the order and rebased.

src/spec.ts
```diff
@@ -591,13 +590,12 @@ function normalizePathOverlay(spec: NormalizedUnitSpec, config: Config = {}): No
...
   if (lineOverlay) {
-    const {interpolate} = markDef;
     layer.push({
       ...(projection ? {projection} : {}),
       mark: {
         type: 'line',
-        ...lineOverlay,
-        ...(interpolate ? {interpolate} : {})
+        ...pick(markDef, ['clip', 'interpolate']),
+        ...lineOverlay
```
domoritz: I think the order was wrong here before. @kanitw

kanitw: Ok -- another is don't you have to clip the point overlay below too?

domoritz: True, will change and merge.

Issue #3779, opened 2018-05-19
Update Legend, Axis, and Title properties for Vega 4


domoritz:
Vega 4 adds some properties that were previously only in the config as properties of Axes, Legends, and Titles. We can add those to Vega-Lite. 

See https://github.com/vega/vega/pull/1269

kanitw:
Note:  

1) Implementing this would involve moving these properties from `VgAxis/LegendConfig` to `VgAxis/LegendBase` and update the `COMMON_AXIS/LEGEND_PROPERTIES_INDEX` accordingly.

2) One  part to be careful is to make sure that a lot of default rules in `compile/axis/encode.ts` (e.g., `labelAlign/Baseline`) are only applied if the labelAlign/Baseline properties are not specified by users. 

3) Also remove manual implementation of labelAngle in VL.  (Also, check if there is any other manual implementation.)

kanitw:
I'm close since this is pretty much done if #3999 is merged.

PR #3936, opened 2018-06-26
Add V4 Axis Properties

melissatdiamond:
Addressing the Axis portion of issue #3779. Initial commit moves properties that are now in Axis in Vega 4 from`VgAxisConfig` to `VgAxisBase`.

### DIFF
--- a/src/vega.schema.ts
+++ b/src/vega.schema.ts
@@ -1,5 +1,6 @@
 import {AggregateOp} from 'vega';
 import {isArray} from 'vega-util';
+import {Color} from '../node_modules/@types/d3';
 import {BaseBin} from './bin';
 import {NiceTime, ScaleType} from './scale';
 import {SortOrder} from './sort';
@@ -283,17 +284,29 @@ export type AxisOrient = 'top' | 'right' | 'left' | 'bottom';
 
 export interface VgAxis {
   scale: string;
+  bandPosition?: number;
   domain?: boolean;
+  domainColor?: Color;
+  domainWidth?: number;
   format?: string;
   grid?: boolean;
   gridScale?: string;
-
+  gridColor?: Color;
+  gridDash?: number[];
+  gridOpacity?: number;
+  gridWidth?: number;
   labels?: boolean;
-
+  labelBaseline?: string;
+  labelAlign?: string;
   labelBound?: boolean | number;
   labelFlush?: boolean | number;
   labelPadding?: number;
   labelOverlap?: boolean | 'parity' | 'greedy';
+  labelAngle?: number;
+  labelColor?: number;
+  labelFont?: string;
+  labelFontSize?: number;
+  labelLimit?: number;
   maxExtent?: number;
   minExtent?: number;
   offset?: number;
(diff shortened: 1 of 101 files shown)


domoritz:
Why do the examples change?

src/vega.schema.ts
```diff
@@ -503,187 +525,187 @@ export type VgBinding = VgCheckboxBinding | VgRadioBinding |
...
    */
-  labels?: boolean;
+  domain?: boolean;
 
   /**
-   * Indicates if labels should be hidden if they exceed the axis range. If `false `(the default) no bounds overlap analysis is performed. If `true`, labels will be hidden if they exceed the axis range by more than 1 pixel. If this property is a number, it specifies the pixel tolerance: the maximum amount by which a label bounding box may exceed the axis range.
+   * Stroke width of axis domain line
    *
-   * __Default value:__ `false`.
+   * __Default value:__  (none, using Vega default).
```
kanitw: I think saying "(none, using Vega default)" is almost equivalent to not saying anything as users don't really the information they should know. 

You can look at https://github.com/vega/vega-parser/blob/master/src/config.js to see some defaults.  (`domainWidth` is `1`, for example.)

Same for all other properties. :)

src/vega.schema.ts
```diff
@@ -503,187 +525,187 @@ export type VgBinding = VgCheckboxBinding | VgRadioBinding |
...
-   *
-   * __Default value:__ `true`
-   */
-  domain?: boolean;
-
-  /**
-   * A boolean flag indicating if grid lines should be included as part of the axis
-   *
-   * __Default value:__ `true` for [continuous scales](https://vega.github.io/vega-lite/docs/scale.html#continuous) that are not binned; otherwise, `false`.
+   * An interpolation fraction indicating where, for `band` scales, axis ticks should be positioned. A value of `0` places ticks at the left edge of their bands. A value of `0.5` places ticks in the middle of their bands.
```
kanitw: 0.5 is the default according to https://github.com/vega/vega-parser/blob/master/src/config.js.

src/vega.schema.ts
```diff
@@ -739,6 +761,8 @@ export interface VgAxisConfig extends VgAxisBase {
   titleY?: number;
 }
 
+export interface VgAxisConfig extends VgAxisBase {}
```
kanitw: If all properties of `VgAxisBase` are in `VgAxisConfig`, then we don't need to declare a separate `VgAxisBase`.

melissatdiamond: I think `Axis` in `src/axis.ts` also extends `VgAxisBase` but has additional properties as well. Should I change `VgAxisBase` to `VgAxisConfig` and have `Axis` extend that?

kanitw: Yep

src/vega.schema.ts
```diff
@@ -501,145 +525,118 @@ export type VgBinding = VgCheckboxBinding | VgRadioBinding |
...
    */
-  labels?: boolean;
+  domainWidth?: number;
 
   /**
-   * Indicates if labels should be hidden if they exceed the axis range. If `false `(the default) no bounds overlap analysis is performed. If `true`, labels will be hidden if they exceed the axis range by more than 1 pixel. If this property is a number, it specifies the pixel tolerance: the maximum amount by which a label bounding box may exceed the axis range.
+   * Color of axis domain line.
    *
-   * __Default value:__ `false`.
+   * __Default value:__  `gray`.
```
kanitw: We need to include quotes for strings (e.g., `"gray"`).  Please do the same for other properties.

PR #3938, opened 2018-06-27
Unhide Stack Transform

invokesus:


examples/specs/rect_mosaic_labelled_with_offset.vl.json
```diff
@@ -236,7 +236,7 @@
   },
   "config": {
     "view": {
-      "stroke": null
+      "stroke": ""
```
kanitw: Why empty string here?  I think

kanitw: Same for all other examples

invokesus: Tests were failing if `stroke` was not a string.

kanitw:
LGTM - I'm closing this so we can merge it together with the docs PR though.

PR #3939, opened 2018-06-27
Stack Docs

invokesus:
~~Merge after #3938 .~~


site/docs/transform/stack.md
```diff
@@ -4,10 +4,15 @@ title: Stack
 permalink: /docs/stack.html
 ---
```
kanitw: Missing intro paragraph

site/docs/transform/stack.md
```diff
@@ -27,43 +32,40 @@ determines type of stacking offset if the field should be stacked.
...
+### Normalized Stacked Bar and Area Charts
 
 You can set `stack` to `"normalize"` to create normalized (or percentage) stacked bar and area charts.
 
 <div class="vl-example" data-name="stacked_bar_normalize"></div>
 
 <div class="vl-example" data-name="stacked_area_normalize"></div>
-
-## Streamgraph
+### Streamgraph
```
kanitw: Why removing line breaks here?

site/docs/transform/stack.md
```diff
@@ -94,8 +96,47 @@ If you want to define custom sort order, you can derive a new field using the [`
...
+     ...
+  ],
+  ...
+}
+```
+
+The `stack` transform in the `transform` array has the following properties:
+
+{% include table.html props="stack,groupby,offset,sort,as" source="StackTransform" %}
+
```
kanitw: To make readers understand how it works, you should pick one of the simplest example from above and produce an equivalent spec with stack transfrom so users understand the idea.

You can then explain that this is useful for composing stack with other transforms such as diverging bar chart and mosaic chart.

site/docs/transform/stack.md
```diff
@@ -94,8 +96,47 @@ If you want to define custom sort order, you can derive a new field using the [`
...
+    {"stack": ..., "groupby": ..., "offset": ..., "sort": ..., "as" ...} // Stack Transform
+     ...
+  ],
+  ...
+}
+```
+
+The `stack` transform in the `transform` array has the following properties:
+
+{% include table.html props="stack,groupby,offset,sort,as" source="StackTransform" %}
```
kanitw: Somehow the table is still empty

image

site/docs/transform/stack.md
```diff
@@ -94,8 +96,47 @@ If you want to define custom sort order, you can derive a new field using the [`
...
+{:#transform}
+## Stack Transform
+
+{: .suppress-error}
+```json
+// A View Specification
+{
+  ...
+  "transform": [
+    {"stack": ..., "groupby": ..., "offset": ..., "sort": ..., "as" ...} // Stack Transform
```
kanitw: Please expand this object to be on multiple lines (one property per line)

site/docs/transform/stack.md
```diff
@@ -94,8 +96,47 @@ If you want to define custom sort order, you can derive a new field using the [`
...
+The `stack` transform in the `transform` array has the following properties:
+
+{% include table.html props="stack,groupby,offset,sort,as" source="StackTransform" %}
+
+### Diverging Bar Chart
+
+Here we initially `stack` by `"question"` and then use the `window` transform to offset each stack.
+<div class="vl-example" data-name="bar_diverging_stack_transform"></div>
+
+### Basic Mosaic Chart
```
kanitw: 1) I don't know if we need to show all of these variant of mosaic chart examples.

site/docs/transform/stack.md
```diff
@@ -94,8 +96,47 @@ If you want to define custom sort order, you can derive a new field using the [`
...
+```
+
+The `stack` transform in the `transform` array has the following properties:
+
+{% include table.html props="stack,groupby,offset,sort,as" source="StackTransform" %}
+
+### Diverging Bar Chart
+
+Here we initially `stack` by `"question"` and then use the `window` transform to offset each stack.
+<div class="vl-example" data-name="bar_diverging_stack_transform"></div>
```
kanitw: Is `bar_diverging_stack_transform` the same example as in the gallery?

site/docs/transform/stack.md
kanitw: For 2., I'm asking @domoritz  to add `json-stringify-pretty-compact` to  bit.ly/js2json -- so we can use it.

invokesus: I kept only the simple mosaic example. Will make a separate PR for adding `rect_mosaic_labelled_with_offset`to the gallery and then add a link to it.

site/docs/transform/stack.md
invokesus: I'll make a separate PR for replacing the gallery example.

kanitw: I think it should be done within the scope of this PR.

site/docs/transform/stack.md
kanitw: Let's just do it here.

_data/examples.json
```diff
@@ -46,7 +46,7 @@
         "title": "Layered Bar Chart"
       },
       {
-        "name": "bar_diverging_stacked",
```
kanitw: Please remove the actual example too as it's not needed anymore

kanitw:
Please also just do this here:
I kept only the simple mosaic example. Will make a separate PR for adding rect_mosaic_labelled_with_offsetto the gallery and then add a link to it.

PR #3943, opened 2018-06-27
[Feature Branch] Add Label and Title Properties to Facet Header

kanitw:


examples/specs/facet_custom_header.vl.json
```diff
@@ -0,0 +1,29 @@
+{
```
pratapvardhan: Correct me if I'm wrong -- isn't this example an exact duplicate of `examples/specs/facet_custom.vl.json`?

kanitw: They are equivalent, but one is a shorthand.

PR #3953, opened 2018-06-29
An attempt to make "binned"  show up in the schema only for x and y

kanitw:
cc: @sirahd I found a way to enforce "binned" to show up only for x and y

We can't override property with more properties, but overriding with less is fine.
Not sure if this makes the type too complicated though!


kanitw:
Note: test is failing because the original PR is failing.  But feel free to merge this and use this code so you don't have to write the code to drop bin: "binned". :)

PR #3954, opened 2018-06-29
[WIP] Make scale domain support date strings & relax their type

kanitw:
I'm including bump in this PR since I think we should release after this PR


jakevdp:
Are you pushing to have a 2.6 release soon?

Issue #3797, opened 2018-05-23
Adding shape and color to line chart renders two legends


d-werner:
Hi guys,

I would like to setup a line chart with predefined color and shape scales. But I struggle already with the default scales, because vega-lite renders two different legends. One for each of the two encoding channels. 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A scatterplot showing horsepower and miles per gallons for various cars.",
  "data": {"url": "data/cars.json"},
  "mark": {"type": "line"},
  "encoding": {
    "x": {"field": "Year", "type": "temporal"},
    "y": {
      "aggregate": "mean",
      "field": "Miles_per_Gallon",
      "type": "quantitative"
    },
    "shape": {"field": "Origin", "type": "nominal"},
    "color": {"field": "Origin", "type": "nominal"}
  }
}
```
![bildschirmfoto 2018-05-23 um 11 26 28](https://user-images.githubusercontent.com/2510850/40415977-78ea7026-5e7c-11e8-9082-32e666cb41ce.png)

It would expect the chart to have a single, combined legend, which is the case for mark type "point".

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "A scatterplot showing horsepower and miles per gallons for various cars.",
  "data": {"url": "data/cars.json"},
  "mark": {"type": "point"},
  "encoding": {
    "x": {"field": "Year", "type": "temporal"},
    "y": {
      "aggregate": "mean",
      "field": "Miles_per_Gallon",
      "type": "quantitative"
    },
    "shape": {"field": "Origin", "type": "nominal"},
    "color": {"field": "Origin", "type": "nominal"}
  }
}
```
![bildschirmfoto 2018-05-23 um 11 34 37](https://user-images.githubusercontent.com/2510850/40416338-4df7be18-5e7d-11e8-97f6-2aa398c929cc.png)

This issue might be related to [issue F](https://github.com/vega/vega-lite/issues/3208) with custom shapes.

BTW, the vega editor shows the warning `[Warning] shape dropped as it is incompatible with "line".`, although it seems to be valid since [(Macro) adding shape to line should automatically add another point layer](https://github.com/vega/vega-lite/issues/3532) has been resolved. ;-)

d-werner:
@jakevdp's [solution for an interactive legend](https://github.com/vega/vega-lite/issues/1657#issuecomment-388438849) would also be a workaround for this issue! :-)

[...]

Issue #3152, opened 2017-10-20
Ordering of connected scatter plot incorrect


lawrlee:
I'm trying to create a connected scatter plot using the following data

```
{
    "values": [
      {
        "value": 1,
        "label": "blue"
      },
      {
        "value": 2,
        "label": "green"
      },
      {
        "value": 3,
        "label": "yellow"
      },
      {
        "value": 4,
        "label": "red"
      }
    ]
}
```

and I want to order the x-axis values by `value` while showing the labels `label`. The following spec works correctly:

https://vega.github.io/editor/#/gist/vega-lite/lawrlee/448f05a10b9107dfcc529f6a6e211f23

with this `layer`:

```
  "layer": [
    {
        "encoding": {
            "x": {"field": "label", "scale": {"zero": false}, "type": "ordinal", "sort": {"field": "value", "op": "values"}},
            "y": {"field": "value", "scale": {"zero": false}, "type": "quantitative"},
            "order": {"field": "value", "type": "quantitative"}
        },
        "mark": "line"
    }
  ]
```

![image](https://user-images.githubusercontent.com/3534824/31799561-a509b5c2-b4ef-11e7-9cde-e7b141c410d3.png)

However, when I add the `point` layer, the ordering of the x-axis resets to alphabetical based on the `label` values.

https://vega.github.io/editor/#/gist/vega-lite/lawrlee/1403c3b87e712e905234e971fae12f39

```
  "layer": [
    {
      "encoding": {
        "x": {
          "field": "label",
          "scale": {
            "zero": false
          },
          "type": "ordinal",
          "sort": {
            "field": "value",
            "op": "values"
          }
        },
        "y": {
          "field": "value",
          "scale": {
            "zero": false
          },
          "type": "quantitative"
        },
        "order": {
          "field": "value",
          "type": "quantitative"
        }
      },
      "mark": "line"
    },
    {
      "encoding": {
        "x": {
          "field": "label",
          "scale": {
            "zero": false
          },
          "type": "ordinal"
        },
        "y": {
          "field": "value",
          "scale": {
            "zero": false
          },
          "type": "quantitative"
        }
      },
      "mark": "point"
    }
  ]
```

![image](https://user-images.githubusercontent.com/3534824/31799632-1536346a-b4f0-11e7-9988-3aa0d25f4bce.png)

How do I achieve the ordering sorted by value with both line and point layers? I have tried multiple combinations of `order` and `sort` with little success.

kanitw:
[...]

Issue #2964, opened 2017-09-06
Layering error bar with sort field does not work


kanitw:
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {"url": "data/barley.json"},
  "layer": [{
    "mark": {"type": "point", "filled": true},
    "encoding": {
      "x": {
        "aggregate": "mean", "field": "yield", "type": "quantitative",
        "scale": {"zero": false},
        "axis": {"title": "Barley Yield"}
      },
      "y": {
        "field": "variety", "type": "ordinal",
        "sort": {"field": "yield","op": "mean", "order": "descending"}
      },
      "color": {"value": "black"}
    }
  }, {
    "mark": "rule",
    "encoding": {
      "x": {
        "aggregate": "ci0", "field": "yield", "type": "quantitative",
        "scale": {"zero": false}
      },
      "x2": {
        "aggregate": "ci1", "field": "yield", "type": "quantitative"
      },
      "y": {
        "field": "variety", "type": "ordinal",
        "sort": {"field": "yield","op": "mean", "order": "descending"}
      }
    }
  }]
}
```

gets
```
WARN Dropping sort property "{"field":"yield","op":"mean","order":"descending"}" as unioned domains only support boolean or op 'count'.
```

The output also does not sort correctly:

![image](https://user-images.githubusercontent.com/111269/30133210-e0a39c92-9307-11e7-9c76-2a8124e4c133.png)

kanitw:
I'll remove `sort` from the `layer_error_bars` example for now.
- [ ] Once we fix this, let add the `sort` back.

domoritz:
This is not possible with Vega. We have to optimize the dataflow to merge redundant dataflows. So this is on the agenda anyway and this specific issue doesn't need to have a new issue.

kanitw:
At first glance I agree with you, but thinking a bit more, this case should not require dataflow optimization.

Basically, you have two layers with their own aggregate data. However, both layers sort `variety` by another field, which means that the domain for `variety` (y-scale) should be drawn from the pre-aggregate data (`data_0`) anyway.  Thus, this is definitely possible in Vega.  There must be some bug in domain that makes this wrong.

domoritz:
I think I disagree. data_0 and data_2 are the non-aggregated datasets but they are not merged.

kanitw:
Oh true, I'll add this as a test case for #2177 then.

PR #3959, opened 2018-07-02
[WIP] Merge Parse Nodes at a Fork

invokesus:
Haven't been able to figure out why multiple `data_*` are being created in the failing test.

Correct output:
```
      [ { name: 'source_0',
          values:
           [ { date: 'Sun, 01 Jan 2012 23:00:01', price: 150 },
             { date: 'Sun, 02 Jan 2012 00:10:02', price: 100 },
             { date: 'Sun, 02 Jan 2012 01:20:03', price: 170 },
             { date: 'Sun, 02 Jan 2012 02:30:04', price: 165 },
             { date: 'Sun, 02 Jan 2012 03:40:05', price: 200 } ] },
        { name: 'data_1',
          source: 'source_0',
          transform:
           [ { type: 'formula', expr: 'toDate(datum["date"])', as: 'date' },
             { type: 'formula',
               expr: 'toNumber(datum["price"])',
               as: 'price' },
             { type: 'formula',
               as: 'seconds_date',
               expr: 'datetime(0, 0, 1, 0, 0, seconds(datum["date"]), 0)' },
             { type: 'filter',
               expr:
                'datum["seconds_date"] !== null && !isNaN(datum["seconds_date"]) && datum["price"] !== null && !isNaN(datum["price"])' } ] },
        { name: 'data_2',
          source: 'source_0',
          transform:
           [ { type: 'formula', expr: 'toDate(datum["date"])', as: 'date' },
             { type: 'formula',
               expr: 'toNumber(datum["price"])',
               as: 'price' },
             { type: 'formula',
               as: 'seconds_date',
               expr: 'datetime(0, 0, 1, 0, 0, seconds(datum["date"]), 0)' },
             { type: 'filter',
               expr:
                '!(length(data("two_store"))) || (vlSingle("two_store", datum))' },
             { type: 'formula',
               as: 'minutes_date',
               expr: 'datetime(0, 0, 1, 0, minutes(datum["date"]), 0, 0)' },
             { type: 'filter',
               expr:
                'datum["minutes_date"] !== null && !isNaN(datum["minutes_date"]) && datum["price"] !== null && !isNaN(datum["price"])' } ] } ]
```

Failing output:

```
      [ { name: 'source_0',
          values:
           [ { date: 'Sun, 01 Jan 2012 23:00:01', price: 150 },
             { date: 'Sun, 02 Jan 2012 00:10:02', price: 100 },
             { date: 'Sun, 02 Jan 2012 01:20:03', price: 170 },
             { date: 'Sun, 02 Jan 2012 02:30:04', price: 165 },
             { date: 'Sun, 02 Jan 2012 03:40:05', price: 200 } ] },
        { name: 'data_1',
          source: 'source_0',
          transform:
           [ { type: 'formula', expr: 'toDate(datum["date"])', as: 'date' },
             { type: 'formula',
               expr: 'toNumber(datum["price"])',
               as: 'price' } ] },
        { name: 'data_2',
          source: 'data_1',
          transform:
           [ { type: 'formula',
               as: 'seconds_date',
               expr: 'datetime(0, 0, 1, 0, 0, seconds(datum["date"]), 0)' },
             { type: 'filter',
               expr:
                'datum["seconds_date"] !== null && !isNaN(datum["seconds_date"]) && datum["price"] !== null && !isNaN(datum["price"])' } ] },
        { name: 'data_3',
          source: 'data_1',
          transform:
           [ { type: 'formula',
               as: 'seconds_date',
               expr: 'datetime(0, 0, 1, 0, 0, seconds(datum["date"]), 0)' },
             { type: 'filter',
               expr:
                '!(length(data("two_store"))) || (vlSingle("two_store", datum))' },
             { type: 'formula',
               as: 'minutes_date',
               expr: 'datetime(0, 0, 1, 0, minutes(datum["date"]), 0, 0)' },
             { type: 'filter',
               expr:
                'datum["minutes_date"] !== null && !isNaN(datum["minutes_date"]) && datum["price"] !== null && !isNaN(datum["price"])' } ] },
        { name: 'data_4',
          source: 'data_1',
          transform:
           [ { type: 'formula',
               as: 'seconds_date',
               expr: 'datetime(0, 0, 1, 0, 0, seconds(datum["date"]), 0)' },
             { type: 'filter',
               expr:
                '!(length(data("two_store"))) || (vlSingle("two_store", datum))' },
             { type: 'formula',
               as: 'minutes_date',
               expr: 'datetime(0, 0, 1, 0, minutes(datum["date"]), 0, 0)' },
             { type: 'filter',
               expr:
                'datum["minutes_date"] !== null && !isNaN(datum["minutes_date"]) && datum["price"] !== null && !isNaN(datum["price"])' } ] },
        { name: 'data_5',
          source: 'source_0',
          transform:
           [ { type: 'formula', expr: 'toDate(datum["date"])', as: 'date' },
             { type: 'formula',
               expr: 'toNumber(datum["price"])',
               as: 'price' },
             { type: 'formula',
               as: 'seconds_date',
               expr: 'datetime(0, 0, 1, 0, 0, seconds(datum["date"]), 0)' },
             { type: 'filter',
               expr:
                '!(length(data("two_store"))) || (vlSingle("two_store", datum))' },
             { type: 'formula',
               as: 'minutes_date',
               expr: 'datetime(0, 0, 1, 0, minutes(datum["date"]), 0, 0)' },
             { type: 'filter',
               expr:
                'datum["minutes_date"] !== null && !isNaN(datum["minutes_date"]) && datum["price"] !== null && !isNaN(datum["price"])' } ] } ]
```
@domoritz Any ideas why?
I've add the built examples to compare the generated vega specs after the parse merging.


domoritz:
Can you list what examples (broken ones from the different issues) work now?

invokesus:
1. https://github.com/vega/vega-lite/issues/2177#issuecomment-369970525
g302 example

2. [Layering error bar with sort field](https://github.com/vega/vega-lite/issues/2964#issue-255733431)
error_bar_with_sort

#3797 doesn't work since it requires identical transform merging.

I'm not sure why #3152 doesn't work. Is the example spec at the bottom correct?
I'm getting these warnings.
```
WARN x-scale's "zero" is dropped as it does not work with point scale.
WARN Dropping sort property {"field":"value","op":"values"} as unioned domains only support boolean or op 'count'.
```

src/compile/data/optimize.ts
```diff
@@ -114,6 +115,33 @@ function getLeaves(roots: DataFlowNode[]) {
   return leaves;
 }
 
+export function isParseMergeable(x: ParseNode, y: ParseNode) {
```
kanitw: Add comment explaining what's the criteria for something to be mergeable?  It seems like that the intersection of the two should have identical values.

src/compile/data/optimize.ts
```diff
@@ -114,6 +115,33 @@ function getLeaves(roots: DataFlowNode[]) {
...
+  const xParse = x.parse;
+  const yParse = y.parse;
+  for (const i of keys(xParse)) {
```
domoritz: use k instead of i for keys. i is iterator

src/compile/data/optimize.ts
```diff
@@ -114,6 +115,33 @@ function getLeaves(roots: DataFlowNode[]) {
...
+    if (i in yParse && yParse[i] !== xParse[i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+export function mergeParse(node: DataFlowNode) {
+  const parseChildren = node.children.filter((x): x is ParseNode => x instanceof ParseNode);
```
domoritz: what is this is syntax?

`.filter((x:ParseNode) => x instanceof ParseNode);`

test/compile/data/optimize.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import {assert} from 'chai';
+import {DataFlowNode} from '../../../src/compile/data/dataflow';
+import {ParseNode} from '../../../src/compile/data/formatparse';
+import {mergeParse} from '../../../src/compile/data/optimize';
+
+describe('compile/data/optimize', () => {
+  describe('mergeParse', () => {
+    it('should merge non-conflicting ParseNodes', () => {
+      const root = new DataFlowNode(null, 'root');
+      const parse1 = new ParseNode(root, {a: 'number', b: 'string'});
```
domoritz: add a property where the keys and values are the same

src/compile/data/optimize.ts
invokesus: That's the return `type`. Equivalent to

invokesus: Unless the _return_ type is coerced, type narrowing doesn't happen.

domoritz: Ohh, I see. Makes sense.

test/compile/data/optimize.test.ts
```diff
@@ -17,6 +17,16 @@ describe('compile/data/optimize', () => {
       assert.deepEqual(mergedParseNode.parse, {a: 'number', b: 'string', c: 'boolean', d: 'date'});
     });
 
+    it('should merge identical ParseNodes', () => {
+      const root = new DataFlowNode(null, 'root');
+      const parse1 = new ParseNode(root, {a: 'number', b: 'string'});
+      const parse2 = new ParseNode(root, {a: 'number', b: 'string'});
```
domoritz: use (a,b) and (b,c) and merge the two tests into one.

domoritz:
Can you rebase?

src/compile/data/optimize.ts
```diff
@@ -114,6 +115,35 @@ function getLeaves(roots: DataFlowNode[]) {
...
+
+  if (parseChildren.length > 1) {
```
kanitw: What if there are more than 2 children and only two of them are mergeable.  
Do we abort the merge somehow?

domoritz: I think we should walk over the children and extract all the parse that is consistent and move it down and only leave what cannot be merged.

kanitw: That's better but might be way more complicated.  

If it's too hard, I think it's okay to do the abort version for now. 
FWIW, our scale, axis, and legend logic abort when one of them can't be merged. (If easy, we don't have to do the same here, but I'm just saying it's better than before already.)

domoritz: It's pretty easy. Create a new node and add to it all the parses from the nodes you are looking at. Then remove empty parse nodes.

examples/compiled/area_overlay.vg.json
```diff
@@ -11,28 +11,8 @@
...
-      "transform": [
-        {"type": "formula", "expr": "toDate(datum[\"date\"])", "as": "date"},
-        {"type": "formula", "expr": "toNumber(datum[\"price\"])", "as": "price"}
-      ]
-    },
-    {
-      "name": "data_2",
-      "source": "source_0",
       "transform": [
+        {"type": "filter", "expr": "datum.symbol==='GOOG'"},
```
kanitw: @domoritz  Why is filter before parsing here?

examples/compiled/concat_bar_layer_circle.vg.json
```diff
@@ -32,11 +32,17 @@
...
+    },
+    {
+      "name": "data_2",
+      "source": "data_1",
+      "transform": [
         {
           "type": "extent",
           "field": "IMDB_Rating",
-          "signal": "concat_0_layer_0_bin_maxbins_10_IMDB_Rating_extent"
+          "signal": "concat_0_layer_1_bin_maxbins_10_IMDB_Rating_extent"
```
kanitw: @domoritz this is somehow still re-ordered.

src/compile/data/optimize.ts
kanitw: So it looks like the current logic already merges all mergeable children and simply ignores unmergable ones, but it doesn't introduce a new node to apply the shared parse. 

The question is whether this is correct if we don't introduce an intermediate node.  (I think it's probably wrong -- as something will get parse in the parent, which now merged conflicting parse, if we try to parse the field into another format later, then it won't get the original input?) 

If this is wrong, we may have to revise our approach.

domoritz: Yeah, it is wrong to parse the same field again unless you parse a string as a number or a number as a date later.

examples/compiled/concat_bar_layer_circle.vg.json
domoritz: This may be due to swapping nodes. We do not guarantee order preservation there.

examples/compiled/area_overlay.vg.json
domoritz: Well, the symbol does not need to be parsed to it's fine.

src/compile/data/optimize.ts
kanitw: Hmm, then shouldn't we try to do this parse merging in a bottom-up fashion -- while collecting potential conflicting information along the way?

kanitw: If we only use this information during the parse-merge logic, I think we don't have to make it a part of the dataflow. We can actually collect and store them as a part of this parse-merge logic.

kanitw: From slack conversation, sounds like we can pretty much always merge. Please correct this.

src/compile/data/optimize.ts
```diff
@@ -108,6 +109,35 @@ function getLeaves(roots: DataFlowNode[]) {
...
+    const mergedParse = parseChildren.pop();
+    const mergeableParseChildren = parseChildren.filter(x => isParseMergeable(x, mergedParse));
+    mergeableParseChildren.forEach(x => {
+      node.removeChild(x);
```
kanitw: This is still removing and adding later, which means it won't preserve the order. Better call `x.remove()` so we can leverage Dom's change in #4038

src/compile/data/optimize.ts
```diff
@@ -111,30 +110,35 @@ function getLeaves(roots: DataFlowNode[]) {
...
+        }
+      }
+    }
+    if (keys(commonParse).length !== 0) {
+      const mergedParseNode = new ParseNode(node, commonParse);
+      for (const parseNode of parseChildren) {
+        for (const key of keys(commonParse)) {
+          delete parseNode.parse[key];
+        }
+        node.removeChild(parseNode);
```
kanitw: Looks like the goal here is to change parent to a new parent. 
So the link to the old parent (if exist) should be removed.

Instead of calling removeChild, maybe consider modifying `parent` setter in a separate PR first?

kanitw: Actually let's do that in a separate PR later. I've assigned https://github.com/vega/vega-lite/issues/4052 to you.

This should be good enough for now.

kanitw:
LGTM, let's handle refactoring later.  I'll @domoritz merge this though as he should take another look. :)

PR #3960, opened 2018-07-02
Impute with null instead of undefined

invokesus:
- Always impute with `null` 
- Always add `window` with default `frame` of `[null,null]`

### DIFF
--- a/src/compile/data/impute.ts
+++ b/src/compile/data/impute.ts
@@ -63,9 +63,9 @@ export class ImputeNode extends DataFlowNode {
     return null;
   }
 
-  public assemble(): [VgImputeTransform] | [VgImputeTransform, VgWindowTransform, VgFormulaTransform] {
-    const {impute, key, keyvals, method, groupby, value, frame} = this.transform;
-    if (method !== 'value' && frame !== undefined) {
+  public assemble(): [VgImputeTransform, VgWindowTransform | VgFormulaTransform, VgFormulaTransform] {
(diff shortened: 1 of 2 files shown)


src/compile/data/impute.ts
```diff
@@ -63,47 +63,44 @@ export class ImputeNode extends DataFlowNode {
...
+          type: 'window',
+          as: ['derived_field'],
+          ops: [method],
+          fields: [impute],
+          frame,
+          ignorePeers: false,
+          ...(groupby ? {groupby}: {})
+        };
+      } else {
+        deriveNewField = {
```
kanitw: For the value case, we shouldn't have to derive a new field just for the value, which uses more memory.  Better use the value directly in the final transform.
Basically, in that cases, only two transforms are needed.

src/compile/data/impute.ts
```diff
@@ -63,47 +63,44 @@ export class ImputeNode extends DataFlowNode {
...
-        ignorePeers: false,
-        ...(groupby ? {groupby}: {})
+        value: null
       };
 
+    let deriveNewField: VgWindowTransform | VgFormulaTransform;
+      if (method && method !== 'value') {
+        deriveNewField = {
+          type: 'window',
+          as: ['derived_field'],
```
kanitw: Instead of naming this as a fixed "derived field", can we name it with something more specific to impute?

e.g., `imputed_${field}_value`

domoritz:
Why don't we use the value field of the impute transform?

PR #3962, opened 2018-07-02
Make `build:examples` and `build:examples-only` run `build:only` and add `post` commands for when we want to skip `build:only`

kanitw:


domoritz:
This makes me sad but oh well.

PR #3965, opened 2018-07-02
Support extended nested specs

domoritz:
Fixes altair-viz/altair#991.


scripts/rename-schema.sh
```diff
@@ -7,6 +7,8 @@ perl -pi -e s,'FacetedCompositeUnitSpec','FacetedUnitSpec',g build/vega-lite-sch
 perl -pi -e s,'ExtendedLayerSpec','LayerSpec',g build/vega-lite-schema.json
 perl -pi -e s,'GenericLayerSpec<CompositeUnitSpec>','LayerSpec',g build/vega-lite-schema.json
 perl -pi -e s,'Generic(.*)<CompositeUnitSpec\,LayerSpec>','\1',g build/vega-lite-schema.json
+perl -pi -e s,'Generic(.*)<FacetedUnitSpec\,LayerSpec>','Faceted\1',g build/vega-lite-schema.json
```
kanitw: I don't think this line should be added.  VConcatSpec should remain its name, for example.

domoritz: Done

PR #3972, opened 2018-07-03
Impute Docs

invokesus:
Merge after #3971 .
I don't know why the `docs_toc.md` has such a terrible diff.


kanitw:
I had to resolve conflicts in Docs TOC earlier and it seems like it changes the type of line returns in the files. 

Try rebasing and re-generating the docs_toc and see if it's still problematic?

kanitw:
NVM - I already rebased -- in order to read the site.

site/docs/transform/impute.md
```diff
@@ -0,0 +1,92 @@
...
+
+The `impute` property of [a field definition](encoding.html#field-def) can be used to generate new data objects in place of the missing data.
+
+{% include table.html props="value,method,frame" source="ImputeParams" %}
+
+In this example, the data object where `"x"` is `3` and `"c"` is `1` is missing. So we can impute the missing data and give the imputed field a value of `0`.
+
+<div class="vl-example" data-name="line_impute_value"></div>
+
+We can also use a `method` on existing data points to generate the missing data.
```
kanitw: Need to explain more how the mean are being calculated.

site/docs/transform/impute.md
```diff
@@ -0,0 +1,92 @@
...
+
+<div class="vl-example" data-name="line_impute_method"></div>
+
+The `frame` property of `impute` can be used to control the window over which the specified `method` is applied.
```
kanitw: With just this short description, it is unclear what the frame means.


The frame JSDocs in the `Impute` declaration is also insufficiently described.

examples/specs/line_impute_transform_value.vl.json
```diff
@@ -0,0 +1,23 @@
...
+      {"x": 1, "y": 43, "c": 0},
+      {"x": 1, "y": 55, "c": 1},
+      {"x": 2, "y": 81, "c": 0},
+      {"x": 2, "y": 53, "c": 1},
+      {"x": 3, "y": 19, "c": 0}]},
+  "transform": [
+    {"impute": "y",
+     "key": "x",
+     "value": 0,
+     "groupby": ["c"]}],
```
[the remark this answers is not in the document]
kanitw: (I suspect that the formatter might not keep `[{` and `}]` on the same lines though.)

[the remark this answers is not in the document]
kanitw: I mean

examples/specs/line_impute_transform_frame.vl.json
```diff
@@ -8,18 +8,34 @@
...
+    ]
+  },
   "transform": [
-    {"impute": "y",
-     "key": "x",
-     "groupby": ["c"],
-     "method": "mean",
-     "frame": [-2, 2]
-    }],
+    {
```
kanitw: As suggested in an earlier comment, you don't need a line break between `[` and `{` as well as between `}` and `]`.

examples/specs/line_impute_transform_frame.vl.json
```diff
@@ -8,18 +8,34 @@
...
+      "method": "mean",
+      "frame": [-2, 2]
+    }
+  ],
   "mark": "line",
   "encoding": {
-    "x": {"field": "x", "type": "quantitative", "scale": {"nice": 1}},
-    "y": {"field": "y", "type": "quantitative"},
-    "color": {"field": "c", "type": "nominal"}
+    "x": {
```
kanitw: The old encoding here was already good and concise.  I just don't like that you don't do new lines for closing brackets.

site/docs/transform/impute.md
```diff
@@ -0,0 +1,110 @@
...
+The `impute` property of [a field definition](encoding.html#field-def) can be used to generate new data objects in place of the missing data.
+
+{% include table.html props="value,method,frame" source="ImputeParams" %}
+
+In this example, the data object where `"x"` is `3` and `"c"` is `1` is missing. So we can impute the missing data and give the imputed field a value of `0`.
+
+<div class="vl-example" data-name="line_impute_value"></div>
+
+We can also use a `method` on existing data points to generate the missing data.
+Values are automatically grouped by the fields of some of the Encoding Channels if specified viz.: `color`, `opacity`, `fill`, `stroke`, `opacity`, `size`, `key`, `detail`.
```
kanitw: Never use the term "viz." as it's an _informal_ abbreviation of visualization.  Better just use "visualization".

kanitw: > Encoding Channels

src/impute.ts
```diff
@@ -16,7 +16,9 @@ export interface ImputeParams {
   value?: any;
 
   /**
-   * The frame over which the `method` will be applied.
+   * A frame specification as a two-element array indicating how the sliding window should proceed. The array entries should either be a number indicating the offset from the current data object, or null to indicate unbounded rows preceding or following the current data object. The default value is `[null, null]`, indicating that the sliding window includes all objects. The value `[-5, 5]` indicates that the window should include five objects preceding and five objects following the current object. Finally, `[null, null]` indicates that the frame should always include all data objects.
```
kanitw: I guess you simply copied this from window.  However, in the context of impute,  "the sliding window" here comes out of nowhere.

src/transform.ts
```diff
@@ -240,8 +240,8 @@ export interface ImputeTransform extends ImputeParams {
    */
   key: string;
   /**
-   * An optional array of key values that should be considered for imputation.
-   * If provided, this array will be used in addition to the key values observed within the input data.
+   * An optional array or `ImputeSequence` of key values that should be considered for imputation.
```
kanitw: `ImputeSequence` is an interface in TypeScript, but is not a part of our syntax.  
Plus, just giving the interface name doesn't explain to the users what it is. 

Useful earlier note: 

> Better explain that this "can be an object that define a sequence" and describe how such object looks like.

site/docs/transform/impute.md
invokesus: "viz." as in [namely](https://dictionary.cambridge.org/dictionary/english/viz)

kanitw: > the fields of some of the Encoding Channels

How are these "some" determined?  Given the current text, users will perceive this as a black box. 

I guess you group by all non-position channels or something? 
Better read the path grouping logic and summarize the important details here. 

In a separate PR, you should take this chance to help revise `line.md` as well as I suspect if the following text there is still accurate (it was written in the old days)

> Adding a field to a mark property channel such as color groups data points into different series, producing a multi-series colored line chart.

site/docs/transform/impute.md
```diff
@@ -0,0 +1,110 @@
...
+Here the `method` used is `mean`. The values are grouped by the `field` of the `color` encoding and the imputation is performed group by group.
+
+<div class="vl-example" data-name="line_impute_method"></div>
+
+The `frame` property of `impute` can be used to control the window over which the specified `method` is applied.
+Here the `frame` is `[-2, 2]` which indicates that the window should include two objects preceding and two objects following the current object.
+
+<div class="vl-example" data-name="line_impute_frame"></div>
+
+If `frame` is unspecified, then it is equivalent to `[null, null]` i.e., the window includes all objects in each group.
```
kanitw: If `frame` is unspecified, then it is there is no frame.  This is equivalent to `[null, null]` i.e., the window includes all objects in each group.


(Also have to add a clause explaining that null represent -infinity and infinity)

site/docs/transform/impute.md
```diff
@@ -0,0 +1,110 @@
...
+}
+```
+
+
+{% include table.html props="impute,key,keyvals,groupby" source="ImputeTransform" %}
+
+The same charts can be created using the `impute` transform as well.
+Here we have to manually specify the `key` and `groupby` fields which were inferred automatically in the `impute` property of the `encoding` field definition.
+
+The inferred `key`, in case of `encoding` `impute`, is the `field` of the other Position Encoding Channel i.e., `field` of `y` if `x` is being `imputed` or `field` of `x` if `y` is being imputed.
```
kanitw: > Position Encoding Channel

Why capitalize?  Please revise the whole file if you accidentally capitalize things else where.

site/docs/transform/impute.md
```diff
@@ -0,0 +1,110 @@
+---
+layout: docs
+title: Impute
+permalink: /docs/impute.html
+---
+
+To impute data in Vega-Lite, users can either use use an `impute` property of an [encoding field definition](#encoding) or an `impute` transform inside the [`transform`](#transform) array.
```
kanitw: You may need to explain establish the concept of `key` and `groupby` up here so you can use it to explain under encoding.

site/docs/transform/impute.md
kanitw:
```
The inferred `key`, in case of `encoding` `impute`**,** is the `field` of the other
```

Wrong position for highlighted comma? 

This is also the explanation for the `encoding` so it's weird that it's explain down here. 
This means that by reading the docs for the encoding section alone, users still won't get it.

You should briefly explain in the intro paragraph what the impute transform do and move this inference rule to the "encoding" part.

site/docs/transform/impute.md
kanitw: > "viz." as in namely

Ah I see. The sentence still reads pretty confusing though. 
Instead of listing all the channels and say "some".  Give them an intuition what do these channels have in common.  Moreover, listing all channels isn't a sustainable strategy as we will have more channels in the future. Your exhaustive list here will then be outdated.

src/transform.ts
kanitw:
```
Defines the key values that should be considered for imputation.
One: an array of key values or an object defining a [number sequence](https://vega.github.io/vega-lite/docs/impute.html#sequence-def).

If provided, this will be used in addition to the key values observed within the input data.  If not provided, the values will be derived from all unique values of the `key` field. For `impute` in `encoding`, the key field is `x` if `y` field is imputed, or vice versa. 

If there is no impute grouping, this property _must_ be specified. 
```

site/docs/transform/impute.md
```diff
@@ -0,0 +1,105 @@
...
+}
+```
+
+The `impute` property of [a field definition](encoding.html#field-def) can be used to generate new data objects in place of the missing data.
+
+{% include table.html props="value,method,frame" source="ImputeParams" %}
+
+For `impute` in encoding, grouping fields and the key field (for identifying missing values) are automatically determined. Values are automatically grouped by the specified fields of [mark property channels](encoding.html#mark-prop), [key channel](encoding.html#key) and [detail channel](encoding.html#detail). If x-field is `impute`d, y-field is the key field. Basically, any missing y-value in each group will lead to a new tuple imputed, and vice versa.
+
+In this example, the data object where `"x"` is `3` and `"c"` is `1` is missing. So we can impute the missing data and give the imputed field a value of `0`.
```
kanitw: This is the first example in the page.  Better explain here c is the groupby field and then x is the key field.  Then say that for group c=1, x=3 is missing.

site/docs/transform/impute.md
```diff
@@ -0,0 +1,105 @@
...
+
+<div class="vl-example" data-name="line_impute_value"></div>
+
+We can also use a `method` on existing data points to generate the missing data.
+Here, we set  `method` = `mean`. The values are grouped by the `field` of the `color` encoding and the imputation is performed group by group.
```
kanitw: > by the `field` of the `color` encoding

kanitw: > The values are grouped by the `field` of the `color` encoding and the imputation is performed group by group.

The grouping here is the same as the first example.  After you explain in the first one, no need to repeat.

site/docs/transform/impute.md
```diff
@@ -0,0 +1,105 @@
...
+
+<div class="vl-example" data-name="line_impute_method"></div>
+
+The `frame` property of `impute` can be used to control the window over which the specified `method` is applied.
+Here, the `frame` is `[-2, 2]` which indicates that the window should include two objects preceding and two objects following the current object.
```
kanitw: > the window should include ...

the window for calculating mean would include

site/docs/transform/impute.md
```diff
@@ -0,0 +1,105 @@
...
+    }
+    ...
+  ],
+  ...
+}
+```
+
+{% include table.html props="impute,key,keyvals,groupby" source="ImputeTransform" %}
+
+The same charts can be created using the `impute` transform as well.
```
kanitw: > The same charts 

The same chart above 

Also link "above" to the above section.

site/docs/transform/impute.md
```diff
@@ -0,0 +1,105 @@
...
+Here, we have to manually specify the `key` and `groupby` fields which were inferred automatically in the `impute` property of the `encoding` field definition.
+
+<div class="vl-example" data-name="line_impute_transform_value"></div>
+
+<div class="vl-example" data-name="line_impute_transform_frame"></div>
+
+`keyvals` may be provided which will be used in addition to the key values observed within the input data. If not provided, the values will be derived from all unique values of the `key` field.
```
kanitw: keyvals is also applicable for encoding.  If there is no grouping field (e.g., no color in the example on the top), then keyvals must be specified.

src/transform.ts
```diff
@@ -238,8 +238,12 @@ export interface ImputeTransform extends ImputeParams {
    */
   key: string;
   /**
-   * An optional array of key values that should be considered for imputation.
-   * If provided, this array will be used in addition to the key values observed within the input data.
+   * Defines the key values that should be considered for imputation.
+   * One: an array of key values or an object defining a [number sequence](https://vega.github.io/vega-lite/docs/impute.html#sequence-def).
```
kanitw: What is "One:" here?

kanitw: "One of:"

site/docs/transform/impute.md
```diff
@@ -0,0 +1,123 @@
...
+
+<div class="vl-example" data-name="line_impute_transform_value"></div>
+
+<div class="vl-example" data-name="line_impute_transform_frame"></div>
+
+Similarly if the `groupby` property is not specified, `keyvals` _must_ be specified.
+
+<div class="vl-example" data-name="line_impute_keyvals"></div>
+
+<div class="vl-example" data-name="line_impute_keyvals_sequence"></div>
```
kanitw: This second example is unnecessary, please remove it from the doc and the example folder.

site/docs/transform/impute.md
```diff
@@ -0,0 +1,123 @@
...
+
+An [encoding field definition](encoding.html#field-def) can include an `impute` definition object to generate new data objects in place of the missing data.
+
+The `impute` definition object can contain the following properties:
+
+{% include table.html props="frame,keyvals,method,value" source="ImputeParams" %}
+
+For `impute` in encoding, the grouping fields and the key field (for identifying missing values) are automatically determined. Values are automatically grouped by the specified fields of [mark property channels](encoding.html#mark-prop), [key channel](encoding.html#key) and [detail channel](encoding.html#detail). If x-field is `impute`d, y-field is the key field. Basically, any missing y-value in each group will lead to a new tuple imputed, and vice versa.
+
+In this example, we `impute` the `y` channel's field (`"y"`), so the `x`'s field (`"x"`) will be used as the `"key"`.
```
kanitw: having x-channel's field named `"x"` makes the text read a bit confusing. I think you should rename the field in all examples to be a and b, so x = a, y = b.  Thus, the text will make it clear that `x` is a channel, while `"a"` is a field name.

kanitw:
@invokesus  Can you address these two comments and rebase? (You can drop the travis schema commit and let it regenerates another one) 

I think we can then merge. :)

PR #3976, opened 2018-07-04
Correct channelIncompatibility for geojson type

kanitw:


src/fielddef.ts
```diff
@@ -576,6 +576,13 @@ const COMPATIBLE = {compatible: true};
 export function channelCompatibility(fieldDef: FieldDef<Field>, channel: Channel): {compatible: boolean; warning?: string;} {
   const type = fieldDef.type;
 
+  if (type === 'geojson' && channel !== 'shape') {
```
domoritz: Why does color not work?

kanitw: Why does it work?  Isn't geojson only for path shape?

domoritz: I thought geojson can have properties that can be encoded with the color of the shape for instance.

kanitw: For those properties, shouldn't the `type` be other typical type (not geojson)?

domoritz: Oh, I see. Yes.

PR #3986, opened 2018-07-04
Make deprecated lat/long type obsolete

kanitw:


domoritz:
Rename: make obsolete -> remove. Obsolete means that it is no longer in use but that's not what you're trying to say.

PR #3990, opened 2018-07-06
Enable code lens in workspace settings

domoritz:


domoritz:
I don't see implementation very often but find it almost always useful when vscode shows it.

kanitw:
> I don't see implementation very often but find it almost always useful when vscode shows it.

You can turn it on your computer, not forcing everyone. :)

PR #3991, opened 2018-07-06
 Support adding tooltip based on all fields in encoding (default) or in the data

kanitw:


src/compile/mark/valueref.ts
```diff
@@ -201,6 +202,20 @@ export function midPoint(
   return isFunction(defaultRef) ? defaultRef() : defaultRef;
 }
 
+export function tooltipForChannelDefs(channelDefs: FieldDef<string>[], config: Config) {
+  const keyValues: string[] = [];
+  const usedKey = {};
+  channelDefs.forEach(fieldDef => {
```
domoritz: I prefer `for (const fieldDef of channelDef) {...}` because it's a real for loop that you can use break and return with.

src/mark.ts
```diff
@@ -99,6 +103,14 @@ export interface MarkConfig extends ColorMixins, VgMarkConfig {
    *
    */
   filled?: boolean;
+
+  /**
+   * The tooltip text string to show upon mouse hover or an object defining which fields should the tooltip be derived from.
+   *
+   * - If `tooltip` is `{"fields": "encodings"}`, then all fields in encodings will be used.
```
domoritz: "all fields from encodings will be used"

src/mark.ts
```diff
@@ -99,6 +103,14 @@ export interface MarkConfig extends ColorMixins, VgMarkConfig {
...
+   * - If `tooltip` is `{"fields": "data"}`, then all fields in the underlying data will be used.
```
domoritz: "all fields that appear in the data will be used"

domoritz:
@jheer @arvind @jakevdp Please let us know what you think about this change.

src/compile/mark/mixins.ts
```diff
@@ -232,31 +229,38 @@ export function wrapCondition(
...
+      // If tooltipDef does not exist, then use value from markDef or config
+      const markTooltip = markDef.tooltip !== undefined ? markDef.tooltip : getMarkConfig('tooltip', markDef, config);
+      if (isString(markTooltip)) {
+        return {value: markTooltip};
+      } else if (isObject(markTooltip)) {
+        // `tooltip` is `{fields: 'encodings' | 'fields'}`
+        if (markTooltip.fields === 'encodings') {
+          return ref.tooltipForChannelDefs(fieldDefs(encoding), config);
+        } else {
+          return {signal: 'datum'};
```
domoritz: Just to note, we won't use assign here to format the data fields nicer.

src/mark.ts
kanitw: Meh.

src/compile/mark/mixins.ts
kanitw: Yeah, we can consider that later.  Not a priority for now.

kanitw:
I think tooltip should be added by default since Vega-Lite's primary goal is to support exploration. I think we can experiment with our pre-release version to see if this leads to some other unexpected issues. (From testing with existing interactive examples, I didn't see any problem.)

src/mark.ts
kanitw: It's actually all fields from `encoding`.  This makes me if it should be `{"fields": "encodings"}`, suppose we don't change to `{"encoding": true}`

arvind:
> Please let us know what you think about this change.

Happy to weigh in but having trouble figuring out what the language/API changes are proposed in this PR (or the linked issue). Could someone provide examples? Thanks :)

jakevdp:
> see if this leads to some other unexpected issues.

Before the tooltip refactoring, I experimented with having tooltips enabled by default in the Colab frontend (back when the behavior was to show all fields). The spec below is an example of one that had some very weird tooltip behavior – can you check how it looks with the new proposed all-tooltip default?
```json
{
  "mark": "point",
  "encoding": {
    "x": {"type": "quantitative", "field": "Displacement"},
    "y": {"type": "quantitative", "field": "Horsepower"}
  },
  "selection": {
    "selector": {
      "type": "single",
      "on": "mouseover",
      "nearest": true
    }
  }
}
```

kanitw:
@arvind TLDR - we're trying to do two things here 

1) Support adding tooltip based on either all fields in encoding or all fields in the data point (equivalent to `signal: 'datum'` in Vega)

2) Make tooltip use all fields in encoding by default. 



1) should definitely be done.  The question is whether we're happy with this syntax.
2) is the one we have to experiment a bit more 

-----

@jakevdp   I tried 

```
{
  "mark": "point",
  "encoding": {
    "x": {"type": "quantitative", "field": "Displacement"},
    "y": {"type": "quantitative", "field": "Horsepower"},
    "color": {
      "condition": {
        "selection": "selector",
        "value": "red"
      }
    }
  },
  "selection": {
    "selector": {
      "type": "single",
      "on": "mouseover",
      "nearest": true
    }
  }
}
````
The selection behavior works correctly, but I don't see tooltip showing at all, that's probably because the mouseover event is redirect to selection and thus doesn't trigger tooltip handler.  I think that's a reasonable behavior though. 

Here is the output Vega spec for the same spec: https://pastebin.com/aSZ1801Z.

domoritz:
LGTM

PR #3993, opened 2018-07-07
Use axis typings from Vega typings and update legend typings.

domoritz:
- [x] Update docs if necessary


examples/compiled/area.vg.json
```diff
@@ -83,7 +83,6 @@
       "scale": "x",
       "orient": "bottom",
       "grid": true,
-      "tickCount": {"signal": "ceil(width/40)"},
```
kanitw: Why should this affect the output Vega at all?

build/vega-lite-schema.json
```diff
@@ -354,12 +366,12 @@
           "type": "string"
         },
         "grid": {
-          "description": "A boolean flag indicating if grid lines should be included as part of the axis\n\n__Default value:__ `true` for [continuous scales](https://vega.github.io/vega-lite/docs/scale.html#continuous) that are not binned; otherwise, `false`.",
+          "description": "A boolean flag indicating if grid lines should be included as part of the axis.",
```
kanitw: Missing default value.   Review the schema diff by yourself first?

domoritz: Hmm, I added the right comment already. Let me see what's going wrong here.

examples/compiled/area.vg.json
domoritz: Ahh, it looks like the difference is that I didn't add tickCount to the grid axis.

domoritz: Fixed.

domoritz:
I looked through the schema and it all looks good to me. I also updated the axis docs.

build/vega-lite-schema.json
domoritz: Fixed

build/vega-lite-schema.json
```diff
@@ -1006,6 +1064,276 @@
       },
       "type": "object"
     },
+    "BaseAxis<number,number,boolean,(number|boolean),string,Color,FontWeight,Align,TextBaseline,VgLayoutAlign,SymbolShape,LabelOverlap,number[]>": {
```
kanitw: This should be renamed lol

build/vega-lite-schema.json
```diff
@@ -3720,26 +4082,32 @@
...
-      "maximum": 900,
-      "minimum": 100,
-      "type": "number"
-    },
-    "FontWeightString": {
       "enum": [
         "normal",
-        "bold"
+        "bold",
+        "100",
```
kanitw: Why "100", "200" and so on?

build/vega-lite-schema.json
```diff
@@ -7478,6 +7889,44 @@
       ],
       "type": "string"
     },
+    "ScaledValueRef<SymbolShape>": {
```
kanitw: This shouldn't be supported in VL.

build/vega-lite-schema.json
```diff
@@ -5154,54 +5542,58 @@
...
+          "description": "Stroke color for legend symbols."
         },
         "symbolStrokeWidth": {
           "description": "The width of the symbol's stroke.\n\n__Default value:__ `1.5`.",
           "minimum": 0,
           "type": "number"
         },
         "symbolType": {
-          "$ref": "#/definitions/SymbolShape",
+          "$ref": "#/definitions/SymbolShapeValue",
```
kanitw: This shouldn't support ValueRef.

src/compile/axis/component.ts
```diff
@@ -1,10 +1,10 @@
+import {Axis as VgAxis} from 'vega';
 import {Axis, AxisPart} from '../../axis';
 import {FieldDefBase} from '../../fielddef';
 import {duplicate, Omit} from '../../util';
-import {VgAxis} from '../../vega.schema';
 import {Split} from '../split';
 
-function isFalseOrNull(v: boolean | null) {
+function isFalseOrNull(v: any) {
```
kanitw: Why do we need to change the type here?

src/axis.ts
```diff
@@ -177,6 +175,8 @@ export const AXIS_PROPERTY_TYPE: {
   titleX: 'main',
   titleY: 'main',
 
+  tickWidth: 'both',
+  tickCount: 'both',
```
kanitw: 1) This index should have alphabetical order, so the output sorts axis properties nicely.  Plus, doing so would remove all the unnecessary example diff. 

2) Can you explain why tickWidth should be in the grid axis?

src/axis.ts
```diff
@@ -44,10 +44,12 @@ export interface AxisConfig extends VgAxisConfigNoSignals, VlOnlyGuideConfig {
    *
    * __Default value:__ `true` for non-nominal fields with non-log scales; `"greedy"` for log scales; otherwise `false`.
    */
-  labelOverlap?: boolean | 'parity' | 'greedy';
+  labelOverlap?: LabelOverlap;
```
kanitw: Please check if this leads to weird "LabelOverlap" in the documentation table.  If so, we better keep `boolean | 'parity' | 'greedy'`.

site/docs/encoding/axis.md
```diff
@@ -46,33 +44,38 @@ Besides `axis` property of a field definition, the configuration object ([`confi
...
+__See also:__  [`guide-label` style config](mark.html#style-config) (common styles for axis, [legend](legend.html), and [header](facet.html#header) labels).
 
 {:#ticks}
 ### Ticks
 
-{% include table.html props="ticks,tickColor,tickCount,tickRound,tickSize,tickStep,tickWidth,values" source="Axis" %}
+{% include table.html props="tickColor,tickCount,tickExtra,tickOffset,tickOpacity,tickRound,ticks,tickSize,tickStep,tickWidth,values" source="Axis" %}
+
+{:#domain}
+### Domain
```
kanitw: Let's add this section above labels instead.  (Alphabetical order.)

site/docs/encoding/axis.md
```diff
@@ -46,33 +44,38 @@ Besides `axis` property of a field definition, the configuration object ([`confi
...
 
-__See also:__ [Axis Ticks Config](#ticks-config).
+{% include table.html props="domain,domainColor,domainOpacity,domainWidth" source="Axis" %}
 
 {:#title}
 ### Title
 
-{% include table.html props="title,titleAlign,titleAngle,titleBaseline,titleColor,titleFont,titleFontSize,titleFontWeight,titleMaxLength,titleLimit,titlePadding,titleX,titleY" source="Axis" %}
```
kanitw: What happens to titleMaxLength?

site/docs/types.md
```diff
@@ -29,13 +29,25 @@ Reference documentation for common parameter **types** expected by Vega-Lite spe
 
 Accepts any literal value, including a string, number, boolean, or `null`.
 
+<a name="align" href="#align">#</a>
+**Align**
```
kanitw: I don't think string enum like Align should be a special type in the docs. Feel like an overkill to jump to another page for simple string enum.

site/docs/types.md
```diff
@@ -29,13 +29,25 @@ Reference documentation for common parameter **types** expected by Vega-Lite spe
...
 Accepts array values. For example: `[]`, `[1, 2, 3]`, `["foo", "bar"]`. If individual array items must adhere to a specific type, bracket notation &ndash; such as `Number[]` or `String[]` &ndash; is used to indicate the item type.
 
 In most cases, arrays may also have [signal references](#Signal) as items. For example: `[{"signal": "width"}, {"signal": "height"}]`.
 
+<a name="baseline" href="#baseline">#</a>
+**Baseline**
+
+How text such as legends, labels, or titles should be aligned vertically.
+
+Possible values: `"top"`, `"middle"`, `"bottom"`, or `"alphabetic"`.
```
kanitw: I don't think string enum like Align should be a special type in the docs. Feel like an overkill to jump to another page for simple string enum.

site/docs/types.md
```diff
@@ -73,6 +84,12 @@ Accepts number values. For example: `1`, `3.14`, `1e5`.
 
 Accepts object literals. For example: `{"left":5, "right":30, "top":5, "bottom":50}`. The valid object property names and types will vary across parameters; read the individual parameter descriptions for more information.
 
+<br/><a name="overlap" href="#overlap">#</a>
+**Overlap**
```
kanitw: This is only for axis and isn't used elsewhere. Better make the type for it clear in the table and don't require jumping to the type page.

build/vega-lite-schema.json
domoritz: vega-typings use this. It actually makes sense since 211 is not a supported font weight.

domoritz: See https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight

domoritz: But yeah, the strings are weird and I'll fix that.

domoritz: https://github.com/vega/vega-parser/blob/master/schema/encode.js#L5

build/vega-lite-schema.json
domoritz: Done

build/vega-lite-schema.json
domoritz: Fixed

build/vega-lite-schema.json
domoritz: Fixed with the issue above.

src/compile/axis/component.ts
domoritz: This can be a BooleanValue now because axis config is a VgAxis (which has potentially signals).

site/docs/encoding/axis.md
domoritz: It's now called `titleLimit`.

site/docs/encoding/axis.md
domoritz: doen

site/docs/types.md
domoritz: Done

site/docs/types.md
domoritz: Done

site/docs/types.md
domoritz: Done

domoritz: That's not true. Legends support it as well.

domoritz: I removed these anyway.

domoritz:
Addressed comments and rebased to master.

build/vega-lite-schema.json
kanitw: NVM then.

src/compile/axis/component.ts
kanitw: Hmm, shouldn't BooleanValue be not allowed for Vega-Lite?

build/vega-lite-schema.json
domoritz: It will be fixed the next time we update vega typings

src/compile/axis/component.ts
domoritz: It is allowed to be generated in a component, though.

domoritz: An axis component is a generic that is typed with VgAxis. An axis component could use signals in it's definition (and it does in fact for `tickCount`)

src/axis.ts
```diff
@@ -134,37 +222,44 @@ const COMMON_AXIS_PROPERTIES_INDEX: Flag<keyof (VgAxis | Axis)> = {
...
   domain: 1,
   domainColor: 1,
+  domainOpacity: 1,
   domainWidth: 1,
   format: 1,
   grid: 1,
   gridColor: 1,
   gridDash: 1,
   gridOpacity: 1,
+  gridScale: 1,
```
kanitw: gridScale shouldn't be supported in Vega-Lite

domoritz: Good point. How do we remove it, though?

src/axis.ts
```diff
@@ -212,35 +307,35 @@ export interface AxisConfigMixins {
   /**
    * X-axis specific config.
    */
-  axisX?: VgAxisConfig;
+  axisX?: VgAxisConfigNoSignals;
```
kanitw: If you use `VgAxisConfigNoSignals` here, then the comment won't be Vega-LIte's?

kanitw: I wonder if we should just make all of these just using AxisConfig, and file a bug and shortTimeLabels doesn't work for these yet.  This way, our schema will be way smaller (no two versions of axis config).

domoritz: Yes. I thought that was the idea here. We don't do anything smart with the values anyway, right?

domoritz: Hmm, we should probably use AxisMixins as well. I'll fix it.

kanitw: Hmm, I'm confused what you do mean "I thought that was the idea here."

kanitw: Oh I see. you mean the comment should be different.   The idea is just the axis-specific config does not support shortTimeLabels yet.  But that's partly why our schema is gigantic.  So I'm voting for consolidating them to use the same definition.

domoritz: Actually, we don't support `grid` here, right?

PR #3994, opened 2018-07-07
Fix bug in derived fields for aggregate transforms

domoritz:
Required for https://github.com/vega/vega-lite/pull/3992


domoritz:
@kanitw @chanwutk This was an oversight when I first wrote this code. We should not add format parse for derived fields.

PR #3995, opened 2018-07-07
add histogram example with binned bin

sirahd:


domoritz:
Should we delete histogram_bin_transform then?

kanitw:
I think this should replace the existing transform.

examples/specs/histogram_binned_transform.vl.json
```diff
@@ -10,7 +10,16 @@
   "encoding": {
```
kanitw: I don't think we should rename the file?  
It's a bin transform, not a binned transform.

domoritz:
Let's squash this.

PR #3999, opened 2018-07-09
Vega title typings

domoritz:
Depends on #3993


build/vega-lite-schema.json
```diff
@@ -1265,6 +1257,58 @@
       },
       "type": "object"
     },
+    "BaseTitle<number,string,Color,FontWeight,TextBaseline,TitleFrame,TitleAnchor>": {
```
kanitw: Let's rename this. I'm sure Altair won't like this crazy generic.

site/docs/view/title.md
```diff
@@ -38,6 +35,7 @@ For example, we can customize the `anchor` of the title of a bar chart.
 }
 ```
 
-To provide themes for all titles, the title config `config: {title: {...}}` can contain the following properties:
+{:#config}
+## Title Config
 
-{% include table.html props="anchor,angle,baseline,color,font,fontSize,fontWeight,limit,offset,orient" source="VgTitleConfig" %}
+To provide themes for all titles, the title configuration `config: {title: {...}}` supports all [title parameters](#params).
```
kanitw: add parentheses to wrap the inline code block?

kanitw:
Hmm test still failing.

src/title.ts
```diff
@@ -28,7 +23,27 @@ export interface TitleBase {
    */
   style?: string | string[];
 
-  // TODO: name, encode, interactive, zindex
+  /**
+   * A boolean flag indicating if the title element should respond to input events such as mouse hover.
+   */
+  interactive?: boolean;
```
kanitw: I don't think we expose `interactive` elsewhere in the codebase?

domoritz:
Rebased and fixed tests.

src/title.ts
domoritz: Done

PR #4000, opened 2018-07-10
Sh/quantizescale

sirahd:
I haven't edited documentation yet and will do it after the code change settles. Can @kanit check if I miss any cases here?


domoritz:
Congratulations for issue/pr #4000!

kanitw:
:tada: Haha, how cool.  I'll add this PR to my talk today!

examples/specs/concat_bar_scales_discretize.vl.json
```diff
@@ -0,0 +1,122 @@
...
+          "legend": {
+            "title": "Threshold"
+          }
+        }
+      }
+    }
+  ],
+  "resolve": {
+    "scale": {
+      "color": "independent",
```
kanitw: Do we really need to enforce color to be independent to avoid merging?

src/compile/scale/properties.ts
```diff
@@ -160,7 +160,8 @@ export function nice(scaleType: ScaleType, channel: Channel, fieldDef: FieldDef<
   if (fieldDef.bin || util.contains([ScaleType.TIME, ScaleType.UTC], scaleType)) {
     return undefined;
   }
-  return util.contains([X, Y], channel); // return true for quantitative X/Y unless binned
+  // return true for quantitative X/Y unless binned or quantize scale
+  return util.contains([X, Y], channel) || scaleType === ScaleType.QUANTIZE;
```
kanitw: Can you add comment in the code to explain why quantize should have "nice" by default?

src/scale.ts
```diff
@@ -65,7 +67,10 @@ const SCALE_CATEGORY_INDEX: {
   ordinal: 'ordinal',
   'bin-ordinal': 'bin-ordinal', // TODO: should bin-ordinal support merging with other
   point: 'ordinal-position',
-  band: 'ordinal-position'
+  band: 'ordinal-position',
+  quantile: 'numeric',
```
kanitw: I don't think these types should be numeric. (They are "discretizing" scale.)

examples/specs/concat_bar_scales_discretize.vl.json
kanitw: Ok I look at the logic for merging.  It seems a bit wrong.  We should fix this.  Maybe create an issue and paste this spec so we can fix later?

site/docs/encoding/scale.md
```diff
@@ -318,6 +318,36 @@ For example, the following plot has a binned field on the `color` channel.
...
+Quantize scales are particularly useful for creating color or size encodings with a fixed number of output values. Using a discrete set of encoding levels (typically between 5-9 colors or sizes) sometimes supports more accurate perceptual comparison than a continuous range.
+For related functionality see [quantile scales](https://vega.github.io/vega-lite/docs/scale.html#quantile), which partition the domain into groups with equal element counts, rather than uniform domain extents.
+
+{:#threshold}
+### Threshold Scales
+
+Threshold scales (`"threshold"`) are similar to [quantize scales](https://vega.github.io/vega-lite/docs/scale.html#quantize), except they allow mapping of arbitrary subsets of the domain (not uniform segments) to discrete values in the range.
+The input domain is still continuous, and divided into slices based on a set of threshold values provided to the domain property. The range property must have N+1 elements, where N is the number of threshold boundaries provided in the domain.
+
+<span class="vl-example" data-name="concat_bar_scales_discretize"></span>
```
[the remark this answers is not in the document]
sirahd: I think showing comparison between different types of discretizing scales in one example might be more helpful. I can break it down if you want.

src/compile/scale/properties.ts
sirahd: I was wrong here. We probably shouldn't make `nice` true by default

kanitw:
> @kanitw I couldn't find fieldDefMatchScaleType. Do you mean channelSupportScaleType and scaleTypeSupportDataType?

I guess.  The issue was so old and we probably have renamed them.

site/docs/encoding/scale.md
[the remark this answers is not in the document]
kanitw: Yeah, this one might be a good one for the gallery?  (So we don't have to add 3 different example for "discretizing scales" in the gallery.) 

But for the docs, having giant example with concat makes the code block a bit overwhelming.

domoritz:
- [ ] Also consider https://github.com/vega/vega/issues/1302

src/compile/scale/range.ts
```diff
@@ -221,7 +226,12 @@ export function defaultRange(
         // Only nominal data uses ordinal scale by default
         return type === 'nominal' ? 'category' : 'ordinal';
       } else if (isContinuousToDiscrete(scaleType)) {
-        return 'ordinal';
+        if (config.range && config.range.ordinal) {
+          return 'ordinal';
+        } else {
+          // for now 4 is the default value for range cardinality. we might change it later
+          return {scheme: 'blues', count: 4};
```
kanitw: Should it be 4 for all of the quantile, quantize, threshold?

kanitw: (It makes sense for quantile, but I doubt if it is for quantize.)

src/compile/scale/range.ts
```diff
@@ -233,6 +243,21 @@ export function defaultRange(
...
+ * Returns the linear interpolation of the range according to the cardinality
+ *
+ * @param rangeMin start of the range
+ * @param rangeMax end of the range
+ * @param cardinality number of values in the output range
+ */
+function interpolateRange(rangeMin: number, rangeMax: number, cardinality: number) {
+  const ranges: number[] = [];
+  for (let i = 1; i <= cardinality; i++) {
+    ranges.push(rangeMin + (i * (rangeMax - rangeMin)) / cardinality);
```
kanitw: 1) This formula won't include rangeMin

2) You can calculate step once instead of every time.

3) Step should be `(rangeMax - rangeMin)) / (cardinality - 1)`, so you can start from i = 0, 1, 2, ..., cardinality - 1

src/compile/scale/range.ts
```diff
@@ -221,7 +226,12 @@ export function defaultRange(
         // Only nominal data uses ordinal scale by default
         return type === 'nominal' ? 'category' : 'ordinal';
       } else if (isContinuousToDiscrete(scaleType)) {
-        return 'ordinal';
+        if (config.range && config.range.ordinal) {
```
kanitw: config.range is [provided by Vega](https://github.com/vega/vega-parser/blob/master/src/config.js#L201) so it will be falsy in VL code.

 Adding this clause mean the default won't ever be used.

src/compile/scale/range.ts
kanitw: This 4 should be in the config too.  Maybe `config.scale.quantileCount`

kanitw: FWIW, the [default `ordinal` scheme in Vega](https://github.com/vega/vega-parser/blob/master/src/config.js#L201) uses `extent: [0.2, 1]` too

src/compile/scale/range.ts
sirahd: I tried including `rangeMin`, but it turns out to be pretty small for default range of 1 in the case of `circle`

domoritz:
Don’t forget to set the right label overlap by default.

sirahd:
@domoritz So `greedy` is the correct default label overlap?

domoritz:
Not always. It's the right one for threshold and quantile legends but not quantize.

src/compile/scale/range.ts
```diff
@@ -219,8 +225,16 @@ export function defaultRange(
       if (scaleType === 'ordinal') {
         // Only nominal data uses ordinal scale by default
         return type === 'nominal' ? 'category' : 'ordinal';
+      } else if (isContinuousToDiscrete(scaleType)) {
+        if (config.range && config.range.ordinal) {
```
kanitw: Don't forget to exclude this condition as the property could be specified in Vega config and be undefined in VL.

src/compile/scale/range.ts
kanitw: minSize is `9` by default.  

> minSize: 9, // Point size is area. For square point, 9 = 3 pixel ^ 2, not too small!

Looks like your discretizing scales include `zero=true` by default.  I think you should change that (and update the docs).

kanitw: In any case, given the name interpolateRange, it's weird that the method skip the first point.  (Then it's not a typical interpolation function anymore.)

src/compile/scale/range.ts
```diff
@@ -210,7 +211,12 @@ export function defaultRange(
       // TODO: support custom rangeMin, rangeMax
       const rangeMin = sizeRangeMin(mark, zero, config);
       const rangeMax = sizeRangeMax(mark, xyRangeSteps, config);
-      return [rangeMin, rangeMax];
+      if (isContinuousToDiscrete(scaleType)) {
+        // for now 4 is the default value for range cardinality. we might change it later
+        return interpolateRange(rangeMin, rangeMax, 4);
```
kanitw: 4 here should be draw from config :)

src/compile/scale/range.ts
```diff
@@ -213,7 +213,7 @@ export function defaultRange(
       const rangeMax = sizeRangeMax(mark, xyRangeSteps, config);
       if (isContinuousToDiscrete(scaleType)) {
         // for now 4 is the default value for range cardinality. we might change it later
-        return interpolateRange(rangeMin, rangeMax, 4);
+        return interpolateRange(rangeMin, rangeMax, config.scale.quantileCount);
```
kanitw: 1) For quantize, there should be a separate quantizeCount too. 
2) For threshold, you need to read domain length

> If the number of values in the scale’s domain is N, the number of values in the scale’s range must be N+1.

(from https://github.com/d3/d3-scale/blob/master/README.md#threshold_range)

src/compile/scale/range.ts
```diff
@@ -226,11 +226,10 @@ export function defaultRange(
         // Only nominal data uses ordinal scale by default
         return type === 'nominal' ? 'category' : 'ordinal';
       } else if (isContinuousToDiscrete(scaleType)) {
-        if (config.range && config.range.ordinal) {
-          return 'ordinal';
+        if (config.range && isVgScheme(config.range.ordinal)) {
```
kanitw: I still don't understand why do we need to read config.range at all?  

Unless you have convincing reason, maybe we should just return 

```
{scheme: 'blues', count};
```

Provide that count use the same rule as size above. (Write a function for `defaultContinuousToDiscreteCount()`)

src/scale.ts
```diff
@@ -347,6 +347,15 @@ export interface ScaleConfig {
    * @minimum 0
    */
   maxStrokeWidth?: number;
+
+  /**
+   * Default range cardinality for discretizing scales, e.g. [`quantile`](https://vega.github.io/vega-lite/docs/scale.html#quantile), [`quantize`](https://vega.github.io/vega-lite/docs/scale.html#quantize), [`threshold`](https://vega.github.io/vega-lite/docs/scale.html#threshold)
```
kanitw: 1) The default count of threshold must depend on threshold
2) If the name is quantile, we shouldn't use it for all of discretizing scales.  We can't name it discretizing either as this number won't work for threshold. So we better separate quantile/quantize.

src/compile/scale/range.ts
sirahd: If the users specify different scheme in `config.range.ordinal`, should we use the specified scheme instead of `blues`?

src/scale.ts
sirahd: Done

src/compile/scale/range.ts
sirahd: Done

src/compile/scale/range.ts
```diff
@@ -229,6 +256,39 @@ export function defaultRange(
...
+  switch (scaleType) {
+    case 'quantile':
+      return config.scale.quantileCount;
+    case 'quantize':
+      return config.scale.quantizeCount;
+    case 'threshold':
+      if (domain !== undefined && isArray(domain)) {
+        return domain.length + 1;
+      } else {
+        return 4;
```
kanitw: I wonder if returning 4 makes sense at all.  
(Shouldn't we throw warning that an array domain must be specified?)

site/docs/encoding/scale.md
```diff
@@ -321,6 +321,38 @@ For example, the following plot has a binned field on the `color` channel.
...
+Quantize scales are particularly useful for creating color or size encodings with a fixed number of output values. Using a discrete set of encoding levels (typically between 5-9 colors or sizes) sometimes supports more accurate perceptual comparison than a continuous range.
+For related functionality see [quantile scales](https://vega.github.io/vega-lite/docs/scale.html#quantile), which partition the domain into groups with equal element counts, rather than uniform domain extents.
+
+<span class="vl-example" data-name="bar_scale_quantize"></span>
+
+{:#threshold}
+### Threshold Scales
+
+Threshold scales (`"threshold"`) are similar to [quantize scales](https://vega.github.io/vega-lite/docs/scale.html#quantize), except they allow mapping of arbitrary subsets of the domain (not uniform segments) to discrete values in the range.
+The input domain is still continuous, and divided into slices based on a set of threshold values provided to the domain property. The range property must have N+1 elements, where N is the number of threshold boundaries provided in the domain.
```
kanitw: "The range property, if specified, ..."

Perhaps, it's worth discussing the default range for all of these scales?

src/compile/scale/range.ts
```diff
@@ -219,8 +239,15 @@ export function defaultRange(
       if (scaleType === 'ordinal') {
         // Only nominal data uses ordinal scale by default
         return type === 'nominal' ? 'category' : 'ordinal';
+      } else if (isContinuousToDiscrete(scaleType)) {
+        if (config.range && isVgScheme(config.range.ordinal)) {
+          return config.range.ordinal;
```
kanitw: I get what you try to do now, but I think you should return

```
{
   ...config.range.ordinal,
   count
}
```

instead :)

sirahd: Yup!

site/docs/encoding/scale.md
sirahd: I think the default range should be a uniform threshold boundaries with certain number, e.g. 4

src/compile/scale/range.ts
sirahd: We can also throw a warning and still returns 4?

kanitw: I think you have to test a bit.  What's the default number of threshold level there is no specified domain?  The default number here should match it.

site/docs/encoding/scale.md
kanitw: I mean discuss / "describe" on the docs so readers understand the scale's default behavior. :)

src/compile/scale/range.ts
sirahd: As far as I know, there's no default range threshold in vega. If no domain and range is specified for threshold scale, vega will display one empty legend with `>=0.5` label

kanitw:
re: https://github.com/vega/vega-lite/pull/4000#discussion_r204231911

> As far as I know, there's no default range threshold in vega. If no domain and range is specified for threshold scale, vega will display one empty legend with >=0.5 label

Yes that's in the D3 documentation but that's not the case we care about since Vega-Lite always input domain to Vega. 

The question is what does Vega do with this domain.  For continuous scales, Vega would calculate the domain's extent and input an array of a tuple of min and max to the scales (as that's what continuous scales expect). But I'm not sure what Vega does for discretizing scales.  If it does the same with continuous scales, then n = 2, so n+1 = 3, not 4.  

Another alternative to consider is to aggregate Q1, median (Q2), and Q3 by default and use that these 3 fields as the default domain for threshold. This way the default `count=4` is reasonable too and we no longer have to throw the domain error.  If we want to do this Q1-Q3 thing we should do it in a separate PR though as it's quite a complicated workflow.  (I think it's a good default, but I'm not sure it's worth the engineering time / added complexity to the codebase since people would normally wanna use threshold when they do really have customize threshold.  Implementing custom code for the default might be a bit unnecessary.)

kanitw:
Btw, I try to build the branch, but fail to build.  Perhaps, rebase and squash all code changes into one commit. (Drop Travis's example changes, don't merge them into the code change commit.)

site/docs/encoding/scale.md
```diff
@@ -321,6 +321,38 @@ For example, the following plot has a binned field on the `color` channel.
...
+For related functionality see [quantize scales](https://vega.github.io/vega-lite/docs/scale.html#quantize), which partition the domain into uniform domain extents, rather than groups with equal element counts.
+Quantile scales have the benefit of evenly distributing data points to encoded values. In contrast, quantize scales uniformly segment the input domain and provide no guarantee on how data points will be distributed among the output visual values.
+
+<span class="vl-example" data-name="bar_scale_quantile"></span>
+
+{:#quantize}
+### Quantize Scales
+
+Quantize scales (`"quantize"`) are similar to [linear scales](https://vega.github.io/vega-lite/docs/scale.html#linear), except they use a discrete rather than continuous range. The continuous input domain is divided into uniform segments based on the number of values in (i.e., the cardinality of) the output range.
+Each range value y can be expressed as a quantized linear function of the domain value x: _y_ = _m round(x) + b_. The `range` property, if not specified, is default to 4 uniform segments between minimum and maximum range value determined by Vega-lite based on mark types.
```
kanitw: 1) Since quantize supports `nice` and `zero`, add a table for it too?

2)

> Each range value y can be expressed as a quantized linear function of the domain value x: _y_ = _m round(x) + b_. 
> The `range` property, if not specified, is default to 4 uniform segments between minimum and maximum range value determined by Vega-lite based on mark types.

This reads a bit confusing.  If I understand correctly, the range value here is like the first step of the scale. (A quantize scale maps continuous to quantized values, then maps the quantized values to the specified range.)


3) 

>  default to 4 uniform segments between minimum and maximum range value determined 

If there are 4 points, there are 3 uniform segments. 


4)

> determined by Vega-lite based on mark types.

Readers will be confused "how is it based on mark types"?  (Even I don't know what this means.

site/docs/encoding/scale.md
```diff
@@ -321,6 +321,38 @@ For example, the following plot has a binned field on the `color` channel.
...
+For related functionality see [quantile scales](https://vega.github.io/vega-lite/docs/scale.html#quantile), which partition the domain into groups with equal element counts, rather than uniform domain extents.
+
+<span class="vl-example" data-name="bar_scale_quantize"></span>
+
+{:#threshold}
+### Threshold Scales
+
+Threshold scales (`"threshold"`) are similar to [quantize scales](https://vega.github.io/vega-lite/docs/scale.html#quantize), except they allow mapping of arbitrary subsets of the domain (not uniform segments) to discrete values in the range.
+The input domain is still continuous, and divided into slices based on a set of threshold values provided to the domain property. The `range` property must have N+1 elements, where N is the number of threshold boundaries provided in the domain. If the `range` property is not specified, is default to 4 uniform segments between minimum and maximum range value determined by Vega-lite based on mark types.
```
kanitw: > is default to 4 uniform segments between minimum and maximum range value 

It's not 4 anymore right?  Looks like you do N+1 


> determined by Vega-lite based on mark types.

Reading this is still a bit confusing how is it based on mark type.  Perhaps, just say that by default Vega-Lite provides N+1 values that provide N uniform segment as the range?

src/compile/scale/properties.ts
```diff
@@ -257,8 +264,8 @@ export function zero(channel: Channel, fieldDef: FieldDef<string>, specifiedScal
 
   // 1) using quantitative field with size
   // While this can be either ratio or interval fields, our assumption is that
-  // ratio are more common.
-  if (channel === 'size' && fieldDef.type === 'quantitative') {
+  // ratio are more common. However, if the scaleType is
```
kanitw: Incomplete comment

> However, if the scaleType is

src/compile/scale/range.ts
```diff
@@ -210,7 +225,16 @@ export function defaultRange(
       // TODO: support custom rangeMin, rangeMax
       const rangeMin = sizeRangeMin(mark, zero, config);
       const rangeMax = sizeRangeMax(mark, xyRangeSteps, config);
-      return [rangeMin, rangeMax];
+      if (isContinuousToDiscrete(scaleType)) {
+        // for now 4 is the default value for range cardinality. we might change it later
```
kanitw: Will "we really change it later"?  (Why we should we?)

In any case, the comment shouldn't be here?

src/compile/scale/range.ts
```diff
@@ -229,6 +264,46 @@ export function defaultRange(
...
+      return config.scale.quantizeCount;
+    case 'threshold':
+      if (domain !== undefined && isArray(domain)) {
+        return domain.length + 1;
+      } else {
+        log.warn(log.message.domainRequiredForThresholdScale(channel));
+        // default threshold boundaries for threshold scale since domain has cardinality of 2
+        return 3;
+      }
+    default:
```
kanitw: I think we don't need default here since there are only 3  continuousToDiscrete scale types. (So this case is impossible.) 

Better modify the type `scaleType: ScaleType` to include only these 3 types and remove this case.

src/compile/scale/range.ts
sirahd: Done

src/compile/scale/range.ts
sirahd: Done

src/compile/scale/properties.ts
sirahd: Oops!

site/docs/encoding/scale.md
sirahd: 3) Actually, there are 4 segments for 4 points (example: https://github.com/vega/vega/blob/22cd624cec5815d1263ca6c1e8effe58cee4f2d1/spec/scales-discretize.vg.json)

4) I meant the minimum and maximum range is determined based on type, but this seems confusing so I'll remove it.

site/docs/encoding/scale.md
sirahd: Done.

kanitw:
:tada: Let's merge when the test passes.  Great job @sirahd!

examples/specs/concat_bar_scales_discretize.vl.json
```diff
@@ -0,0 +1,123 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
+  "description": "A simple bar chart with embedded data.",
```
pratapvardhan: Probably meant to have the description as "Horizontally concatenated charts that show different types of discretizing scales" instead?

pratapvardhan: Also, noticed other examples description too, seem to be not updated.

domoritz: Good catch. Can you send a PR to fix these? Don't worry about updating the Vega and the images; they will update automatically.

PR #4012, opened 2018-07-12
Make Travis rebuild all svgs if the commit message include "[SVG]"

kanitw:
(So the SVGs are always consistently built in Travis)


domoritz:
Why is this necessary? Isn't it enough to rebuild the SVGs when the Vega spec or version changes?

kanitw:
It is necessary because if you build your examples offline, then the output SVGs might be slightly different from what Travis would create (floating point snafu).  Then it will cause unexpected diff in subsequence PRs.  Ideally we shouldn't have to use this often, but we may need it in rare cases.

PR #4024, opened 2018-07-15
Make referenced String enum types show as String + refactor template

kanitw:


domoritz:
Is there a way to see the options? Does this change  any pages for the users now?

kanitw:
This changes all the confusing type name (eg “AutoSizeType”) to strings.

Options should be included manually in the typedocs so they show up in autocomplete.

kanitw:
One alternative would be showing all values on a tooltip (showing all of them by default can be a disaster for type like aggregate) but thats beyond the scope of this PR

domoritz:
Since they are in the types, they already show up as tooltips. I wonder whether we should show the options automatically in the docs instead of manually adding them to the comments.

kanitw:
1) I thought about it but showing options automatically won't explain what the values mean like we do manually

2) Even if there is a better way, it's beyond the scope of this PR.  This is already an improvement.

domoritz:
It's beyond the scope of the PR for sure. I still think it would be useful since we often just list the options after a description of what the different options mean.

kanitw:
>  after a description of what the different options mean.

The problem is we often list them before, which makes it harder to auto-list.

PR #4026, opened 2018-07-15
better handling for invalid values

kanitw:
This removes the need for invalid filter, which makes dataflow optimization harder.


kanitw:
Github is broken when I force push .. will create another branch.

PR #4028, opened 2018-07-16
Add header `titlePadding` and `labelPadding`

domoritz:


src/header.ts
```diff
@@ -82,6 +85,20 @@ export interface HeaderConfig {
    */
   titleLimit?: number;
 
+  /**
+   * The orthogonal offset in pixels by which to displace the title from its position along the edge of the chart.
+   *
+   * __Default value:__ `10`, indicating no limit
+   */
+  titleOffset?: number;
```
kanitw: A similar thing is called titlePadding in Axis.

src/header.ts
```diff
@@ -116,6 +133,20 @@ export interface HeaderConfig {
    * __Default value:__ `0`, indicating no limit
    */
   labelLimit?: number;
+
+  /**
+   * The orthogonal offset in pixels by which to displace the title from its position along the edge of the chart.
+   *
+   * __Default value:__ `10`, indicating no limit
+   */
+  labelOffset?: number;
```
kanitw: A similar thing is called labelPadding in Axis.

src/header.ts
```diff
@@ -82,6 +85,20 @@ export interface HeaderConfig {
...
+   * __Default value:__ `10`, indicating no limit
+   */
+  titleOffset?: number;
+
+  /**
+   * Title orientation (`"top"`, `"bottom"`, `"left"`, or `"right"`).
+   *
+   * __Default value:__ `"left"` for rows and `"top"` for columns.
+   */
+  titleOrient?: TitleOrient;
```
kanitw: Does mapping title orient to orient of underlying title work?

Don't you have to do much more (correctly applying to header/footer, etc.)?

src/header.ts
```diff
@@ -116,6 +133,20 @@ export interface HeaderConfig {
...
+   * __Default value:__ `10`, indicating no limit
+   */
+  labelOffset?: number;
+
+  /**
+   * Label orientation (`"top"`, `"bottom"`, `"left"`, or `"right"`).
+   *
+   * __Default value:__ `"left"` for rows and `"top"` for columns.
+   */
+  labelOrient?: TitleOrient;
```
kanitw: Does mapping label orient to orient of underlying title work?

I thin you have to do much more (correctly applying to header/footer, etc.)?

src/header.ts
domoritz: Done

src/header.ts
domoritz: Done

src/header.ts
```diff
@@ -82,6 +84,13 @@ export interface HeaderConfig {
    */
   titleLimit?: number;
 
+  /**
+   * The orthogonal distance in pixels by which to displace the title from its position along the edge of the chart.
+   *
+   * __Default value:__ `10`, indicating no limit
```
kanitw: Why is 10 "no limit"

src/header.ts
```diff
@@ -116,6 +125,13 @@ export interface HeaderConfig {
    * __Default value:__ `0`, indicating no limit
    */
   labelLimit?: number;
+
+  /**
+   * The orthogonal distance in pixels by which to displace the title from its position along the edge of the chart.
+   *
+   * __Default value:__ `10`, indicating no limit
```
kanitw: Why is 10 "no limit"

PR #4029, opened 2018-07-16
Merge identical transforms at forks

invokesus:


invokesus:
Fixes #2177 and one of the problems @kanitw mentioned in [#3797](https://github.com/vega/vega-lite/issues/3797#issuecomment-391603239).
```
[
  {"data": "data_0", "field": "Origin"},
  {"data": "data_1", "field": "Origin"}
]
```
now is 
```
{"data": "source_0", "field": "mean_Miles_per_Gallon"}
```

examples/compiled/concat_layer_voyager_result_future.vg.json
```diff
@@ -70,16 +70,6 @@
       "name": "data_3",
       "source": "data_1",
       "transform": [{"type": "filter", "expr": "datum.label !== 'PoleStar'"}]
-    },
```
kanitw: In the future PRs, we should try to see how we can make data_2 and data_3 merge for this example.  (No need for this PR.)

src/compile/data/filter.ts
```diff
@@ -22,4 +22,9 @@ export class FilterNode extends DataFlowNode {
       expr: this.expr
     };
   }
+
+  public hash() {
+    // FIX include something from model in hash
```
kanitw: Hmm I thought we agreed to decouple SelectionFilter from ExpressionFilter. 
So we only merge ExpressionFilter, not SelectionFilter.  

That said, the current code means that it will always be merged.  The question is whether this is the right behavior. Maybe it's fine. 

To test it, create an spec where the same selection is applied two layers and see if merging the results still produce correct output.  (Maybe concat one unit spec with a layered spec then apply selection from the unit spec to both layers in the other spec.)

src/compile/data/optimize.ts
```diff
@@ -161,6 +161,7 @@ export function optimizeDataflow(dataComponent: DataComponent) {
 
   roots.forEach(moveFacetDown);
   roots.forEach(mergeParse);
+  roots.forEach(optimizers.mergeIdenticalTransforms);
```
kanitw: Hmm, why is the call pattern different from others (mergeParse isn't in optimizers?)

cc: @domoritz

domoritz: mergeParse should be in optimizers. It's not super clear because in a way it's necessary for correctness as well.

src/compile/data/filter.ts
kanitw: Actually, modify interactive_stocks_nearest_index to use timeUnit would demonstrate if this works well.

kanitw: The sad news is that adding `timeUnit` to `interactive_stocks_nearest_index` doesn't even work, but the diff that this PR is causing to that version seems reasonable.

Thus, I wonder if we should ignore this case for now.  

@domoritz  any thoughts?


In any case, @invokesus  we better remove `// FIX include something from model in hash`.  I don't think we should add stuff from the model to the hash, but rather get rid of model from Filter https://github.com/vega/vega-lite/issues/4084

invokesus:
The coverage test is failing but I don't think it makes sense to write new tests for the concerned files.

src/compile/data/optimizers.ts
```diff
@@ -57,6 +57,35 @@ export function moveParseUp(node: DataFlowNode) {
...
+
+/**
+ * Merge Identical Transforms at forks by comparing hashes.
+ */
+export function mergeIdenticalTransforms(node: DataFlowNode) {
+  const transforms = node.children.filter((x): x is TransformNode => isTransformNode(x));
+  const hashes = transforms.map(x => x.hash());
+  const buckets = {};
+  for (let i = 0; i < hashes.length; i++) {
+    if (buckets[hashes[i]] === undefined) {
```
domoritz: `buckets[hashes[i]] = (buckets[hashes[i]] || []).push(transforms[i])`?

src/compile/data/optimize.ts
domoritz: Wouldn't it be better to start from the leaves?

src/compile/data/bin.ts
```diff
@@ -144,6 +144,9 @@ export class BinNode extends DataFlowNode {
     return out;
   }
 
+  public hash() {
+    return hash(this.bins);
```
domoritz: Could there be problems since the node type is not part of the hash?

src/compile/data/window.ts
```diff
@@ -59,6 +59,10 @@ export class WindowTransformNode extends DataFlowNode {
     return windowFieldDef.as || vgField(windowFieldDef);
   }
 
+  public hash() {
+    return hash(this.transform);
```
domoritz: Do you think it makes sense to have tests for these hash functions so we can read them in the diff and see whether they look good?

src/compile/data/bin.ts
invokesus: Good point, I'll include it in the hash. 
What's the best way though? Adding a `nodeType` key?

Maybe something like
`return hash({nodeType: "BinNode", ...this.bins});`

src/compile/data/optimize.ts
invokesus: Yeah I think so. But I thought it would be better to have the top-down version working with the `mergeParse`(which is also top-down) and refactor them both to bottom-up when I've got the multiple pass thing working.

src/compile/data/window.ts
invokesus: Are you suggesting tests like
i) `hash(node) === hash(node.clone())`
or 
ii) `hash(node)` === `'__hash_value__'`
I see the benefit that *test ii* might provide, but the value will have to be updated each time we change things.
However, I think I should write some tests for the hash function itself with probable inputs in `util.test.ts`.

src/compile/data/bin.ts
domoritz: Or just type + hash

src/compile/data/optimize.ts
domoritz: Okay, just don't forget

src/compile/data/window.ts
domoritz: Latter

src/compile/data/stack.ts
```diff
@@ -191,6 +192,10 @@ export class StackNode extends DataFlowNode {
     }, {});
   }
 
+  public hash() {
+    return 'StackNode' + hash(this._stack);
```
domoritz: `\`Stack ${hash(this._stack)}\``

invokesus: colon?

domoritz: I don't care but @kanitw suggested that we already use colons in the hash. It doesn't really matter so I'd say space is fine.

kanitw: It looks nicer in the test without colon lol -- not that it's a big deal.

invokesus:
Let me know if I missed anything.

domoritz:
Remove "Node"  from hash

src/compile/data/optimizers.ts
domoritz: I guess we could use concat but whatever.

domoritz:
🎉

PR #4032, opened 2018-07-16
Handle invalid values with production rule instead of filter

kanitw:


domoritz:
Why does the weather svg change in such a weird way?

domoritz:
Also, how come the count values have changes here:

screen shot 2018-07-16 at 20 14 10

kanitw:
> The count doesn't change, but the range now includes buckets with of the value as null as they are no longer filtered.

Note that for legend of raw data, including the whole data range is quite a useful information.

domoritz:
I'm not sure I understand. How can the counts change? We run an aggregation and group by x and y. The invalid values would be in a different category, no?

kanitw:
>I'm not sure I understand. How can the counts change? We run an aggregation and group by x and y. The invalid values would be in a different category, no?

The count __does not__ change, but the size scale previously didn't include the count of buckets with invalid values in their domain.  Now they do.  I don't think it hurts to do so.

In fact for raw data (that this example shows aggregate), having the scale represents the actual data range is the right thing to do for a few reasons. 

1) Users wouldn't make wrong assumption about the range due to some rows being filtered
2) It makes the legend consistent across chart no matter which field you visualize.
(Previously the legend of the same field may change depending on what other fields you visualize. If you happen to visualize a field on x with so many null values, you can screw up the y, color, and size.

(Now I wonder though, if we should keep filterinvalid node as an option, but don't make them the default.  Hiding invalid points like we do here provides a better default.)

kanitw:
>(Now I wonder though, if we should keep filterinvalid node as an option, but don't make them the default. Hiding invalid points like we do here provides a better default.)

Thinking more, I think there is no point keeping it.

PR #4033, opened 2018-07-16
Use nice for all scales, not just x and y

domoritz:
See https://github.com/vega/vega/issues/1364


domoritz:
I don't think this is an improvement and we should close this. @kanitw ?

PR #4035, opened 2018-07-16
Refactor examples

chanwutk:
# CO2 Plot
- Change color type from `nominal` to `ordinal` because it seems like there is no connection between line when using `nominal` color.
- Refactor:
	- use `window` instead of `aggregate` to get first and last values.
	- refactor common code

screen shot 2018-07-16 at 12 13 39

# Waterfall Chart
- refactor common code


domoritz:
I ❤️ this. Thanks for the improvements.

examples/specs/normalized/layer_line_co2_concentration_normalized.vl.json
```diff
@@ -0,0 +1,102 @@
...
+    },
+    {
+      "window": [
+        {"op": "first_value", "field": "scaled_date", "as": "first_date"}
+      ],
+      "sort": [{"field": "scaled_date", "order": "ascending"}],
+      "groupby": ["decade"]
+    }
+  ],
+  "config": {"text": {"align": "left", "dx": 3, "dy": 1}},
```
kanitw: Let's put config at the end of the file.

PR #4036, opened 2018-07-16
nice:false is the default.

domoritz:


kanitw:
lol FYI, you can just edit the commit message when you "squash and merge", no need to force push.

kanitw:
> Hmm, I would actually not squash this since it's nice to be able to revert a commit without having to deal with changed examples.

Well if you revert, you should revert the change examples too.  So I don't see why we shouldn't squash.  Otherwise we shouldn't squash any PR anyway.

domoritz:
I think we should squash the code changes but not all PRs in general.

domoritz:
Nonw of the generated SVG changes. I think this is good.

kanitw:
> I think we should squash the code changes but not all PRs in general.

When should things be squashed and when should they not then?

kanitw:
I'd argue that most PRs won't get reverted so the shorter commit logs the better when we go through and write the release notes.

domoritz:
What if I want to look at what a commit changed? I have to scroll through a bunch of example changes as well. Le'ts discuss at the standup.

kanitw:
> What if I want to look at what a commit changed? I have to scroll through a bunch of example changes as well.

`fn + arrow right = end` -- pretty easy to go to the end.

PR #4038, opened 2018-07-16
Do not reorder children when we remove a dataflow node

domoritz:


test/compile/data/dataflow.test.ts
```diff
@@ -70,15 +70,41 @@ describe('compile/data/dataflow', () => {
...
+
+        const rootChild1 = new DataFlowNode(root, 'rootChild1');
+        const node = new DataFlowNode(root, 'node');
+        const rootChild2 = new DataFlowNode(root, 'rootChild2');
+
+        const child1 = new DataFlowNode(node, 'child1');
+        const child2 = new DataFlowNode(node, 'child2');
+
+        expect(root.children).toEqual([rootChild1, node, rootChild2]);
+        assert.equal(rootChild1.parent, root);
```
kanitw: No need to change for this PR, but let's always use `expect` in new code. 
So it will be less work to remove chai later. 

Note that the auto-migrating tool doesn't work for TS yet. TT

domoritz: I changed it because I wanted to get a good diff for this PR :-P

kanitw: I mean you add `assert.equal` below, which could have been `expect(...).toEqual()`. :p

domoritz: Oh yeah, I copied a test.

PR #4040, opened 2018-07-16
Define a default interpolated color range for non-sequential qualitat…

### LINKED ISSUE #4037: Log colour scales don't seem to choose sensible default domains/ranges
Summary: If I use a log colour scale, there is very little variation across the colour range. If I log-transform the variable myself, or, oddly, set the colour range for the scale, I get a much better result.

I assume it is about the range being chosen by default in this case. I don't have this issue with linear colour scales, and don't see any analogous issue with using a log scale on a spatial axis.

I originally raised this issue for Altair at https://github.com/altair-viz/altair/issues/1014 . You can see the Python code for the example there - I've extracted the JSON below from that as I don't usually use vega-lite directly. I hope it's what's needed. 

@jakevdp (https://github.com/altair-viz/altair/issues/1014) has pointed out that this behaviour seems to occur in Vega 4 but not Vega 3.

Here's a toy example:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.4.3.json",
  "config": {"view": {"height": 300, "width": 400}},
  "vconcat": [
    {
      "data": {
        "values": [
          {"c": 1, "x": 1, "y": 1},
          {"c": 10, "x": 2, "y": 2},
          {"c": 100, "x": 3, "y": 3},
          {"c": 1000, "x": 4, "y": 4}
        ]
      },
      "encoding": {
        "color": {
          "field": "c",
          "scale": {"type": "log"},
          "type": "quantitative"
        },
        "x": {"field": "x", "type": "quantitative"},
        "y": {"field": "y", "type": "quantitative"}
      },
      "mark": {"size": 90, "type": "circle"},
      "title": "A: log scale"
    },
    {
      "data": {
        "values": [
          {"c": 1, "x": 1, "y": 1},
          {"c": 10, "x": 2, "y": 2},
          {"c": 100, "x": 3, "y": 3},
          {"c": 1000, "x": 4, "y": 4}
        ]
      },
      "encoding": {
        "color": {"field": "logc", "type": "quantitative"},
        "x": {"field": "x", "type": "quantitative"},
        "y": {"field": "y", "type": "quantitative"}
      },
      "mark": {"size": 90, "type": "circle"},
      "title": "B: log transform, linear scale",
      "transform": [{"as": "logc", "calculate": "log(datum.c)"}]
    },
    {
      "data": {
        "values": [
          {"c": 1, "x": 1, "y": 1},
          {"c": 10, "x": 2, "y": 2},
          {"c": 100, "x": 3, "y": 3},
          {"c": 1000, "x": 4, "y": 4}
        ]
      },
      "encoding": {
        "color": {
          "field": "c",
          "scale": {"range": ["red", "blue"], "type": "log"},
          "type": "quantitative"
        },
        "x": {"field": "x", "type": "quantitative"},
        "y": {"field": "y", "type": "quantitative"}
      },
      "mark": {"size": 90, "type": "circle"},
      "title": "C: log scale with range set"
    },
    {
      "data": {
        "values": [
          {"c": 1, "x": 1, "y": 1},
          {"c": 10, "x": 2, "y": 2},
          {"c": 100, "x": 3, "y": 3},
          {"c": 1000, "x": 4, "y": 4}
        ]
      },
      "encoding": {
        "color": {
          "field": "logc",
          "scale": {"range": ["red", "blue"], "type": "linear"},
          "type": "quantitative"
        },
        "x": {"field": "x", "type": "quantitative"},
        "y": {"field": "y", "type": "quantitative"}
      },
      "mark": {"size": 90, "type": "circle"},
      "title": "D: log transform, linear scale, range set",
      "transform": [{"as": "logc", "calculate": "log(datum.c)"}]
    }
  ]
}
```

altair_logcolourbug

The problem occurs in Plot A, where I use a log colour scale and do not set the range. I would expect A to look very similar to B, where I use a log data transform instead of a scale transform. As you can see though the colour bar covers a pretty small range - Plot B, which uses a transform instead of a scale, is much better. Conversely, Plot C does look very similar to D. The only difference is in these two plots I have set the scale range explicitly.

domoritz:
…ive color scales. Fixes #4037

Looks like this: https://github.com/vega/vega-lite/pull/4040#issuecomment-405700269

```json
{
  "mark": "rect",
  "width": 300,
  "height": 200,
  "encoding": {
    "x": {
      "bin": {"maxbins":60},
      "field": "IMDB_Rating",
      "type": "quantitative"
    },
    "y": {
      "bin": {"maxbins": 40},
      "type": "quantitative"
    },
    "color": {
      "aggregate": "count",
      "type": "quantitative",
      "scale": {
        "type": "log"
      }
    }
  },
  "config": {
    "view": {
      "stroke": "transparent"
    }
  }
}

```


domoritz:
cc @claresloggett

src/compile/scale/range.ts
```diff
@@ -220,6 +220,9 @@ export function defaultRange(
         // Only nominal data uses ordinal scale by default
         return type === 'nominal' ? 'category' : 'ordinal';
       }
+      if (isContinuousToContinuous(scaleType)) {
+        return ['#f7fbff', '#0e427f']; // also update interpolation if you change this
```
kanitw: Wonder if it's worth adding to Vega config.

domoritz: @jheer?

domoritz:
Btw, cubehelix interpolation would look like this

image

src/compile/scale/range.ts
kanitw: If we want to be consistent with [the `ramp` / `ordinal` range](https://github.com/vega/vega-parser/blob/master/src/config.js#L205), perhaps we could use

```
{
  scheme: 'blues',
  extent: [0.2, 1],
  count: 2
}
```

domoritz: screen shot 2018-07-17 at 09 16 15

kanitw: Haha, ok.  Bad suggestion then.

domoritz:
I also optimized the default label overlap for log legends.

image

domoritz:
With hcl and manual colors

image

domoritz:
Ready

src/config.ts
```diff
@@ -297,8 +297,16 @@ export const defaultConfig: Config = {
     borders: false
   },
 
+  range: {
```
kanitw: No you shouldn't change the ramp config as this will change all examples and also affect the sequential scale's default.  

```
{
      scheme: 'blues',
      extent: [0, 1]
    }
```

is good though if you return it directly.

domoritz: Well, if we do that, we can use better colors.

domoritz:
Okay, reverted the last change.

src/compile/legend/properties.ts
```diff
@@ -17,3 +18,10 @@ export function clipHeight(scaleType: ScaleType) {
   }
   return undefined;
 }
+
+export function labelOverlap(scaleType: ScaleType): LabelOverlap {
+  if (scaleType === 'log') {
```
kanitw: Document the changed default?

src/compile/scale/properties.ts
```diff
@@ -156,6 +159,13 @@ export function parseNonUnitScaleProperty(model: Model, property: keyof (Scale |
   });
 }
 
+export function interpolate(channel: Channel, scaleType: ScaleType) {
```
kanitw: Document the default?

src/compile/scale/range.ts
```diff
@@ -220,6 +220,10 @@ export function defaultRange(
         // Only nominal data uses ordinal scale by default
         return type === 'nominal' ? 'category' : 'ordinal';
       }
+      if (isContinuousToContinuous(scaleType)) {
+        // Manually set colors for now. We will revise this after https://github.com/vega/vega/issues/1369
+        return ['#f7fbff', '#0e427f'];
```
kanitw: Document the default?

domoritz:
Revised and rebased.

src/scale.ts
```diff
@@ -608,7 +608,9 @@ export interface Scale {
...
+   * The interpolation method for range values. Except for continuous color scales, by default, a general interpolator for numbers, dates, strings and colors (in RGB space) is used. For color ranges, this property allows interpolation in alternative color spaces. Legal values include `rgb`, `hsl`, `hsl-long`, `lab`, `hcl`, `hcl-long`, `cubehelix` and `cubehelix-long` ('-long' variants use longer paths in polar coordinate spaces). If object-valued, this property accepts an object with a string-valued _type_ property and an optional numeric _gamma_ property applicable to rgb and cubehelix interpolators. For more, see the [d3-interpolate documentation](https://github.com/d3/d3-interpolate).
+   *
+   * * __Default value:__ `hcl` for continuous color scales. Otherwise not set.
```
kanitw: "Otherwise not set" is confusing. 

Does interpolate really work for scales that are not "continuous"? 
If not, better just say the default is `"hcl"`.

domoritz: Fixed

kanitw:
Test coverage ?

domoritz:
Ready. The rest of the test coverage doesn't make much sense.

src/compile/legend/parse.ts
```diff
@@ -134,12 +133,13 @@ function getProperty(
...
     //   return getFirstDefined(specifiedLegend.clipHeight, properties.clipHeight(model.getScaleComponent(channel).get('type')));
+    case 'labelOverlap':
+      return getFirstDefined(
+        specifiedLegend.labelOverlap,
+        properties.labelOverlap(model.getScaleComponent(channel).get('type'))
+      );
     case 'values':
       return properties.values(specifiedLegend, fieldDef);
-    case 'labelOverlap':
-      if (contains(['quantile', 'threshold'], scale.type)) {
```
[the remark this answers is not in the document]
sirahd: Yup! Thanks for double checking

PR #4041, opened 2018-07-17
Use some more vega typings

domoritz:
The commits are separate and don't need to be squashed. Examples and schema should not change.


src/vega.schema.ts
```diff
@@ -90,13 +68,13 @@ export interface VgValueRef {
...
   fields: (any[] | VgDataRef | SignalRef)[];
-  sort?: VgUnionSortField;
+  sort?: UnionSortField;
 }
 
 export interface VgFieldRefUnionDomain {
   data: string;
-  fields: VgFieldRef[];
-  sort?: VgUnionSortField;
+  fields: FieldValue[];
```
kanitw: The name FieldValue is pretty confusing.

kanitw: I think it should be `Field` in Vega typings, and we probably should call it `VgField` here.

src/vega.schema.ts
```diff
@@ -90,13 +68,13 @@ export interface VgValueRef {
...
+  sort?: UnionSortField;
```
kanitw: This is a Vega only thing, so it's confusing that you remove Vg prefix.

src/axis.ts
```diff
@@ -286,6 +285,7 @@ const AXIS_PROPERTIES_INDEX: Flag<keyof Axis> = {
 
 const VG_AXIS_PROPERTIES_INDEX: Flag<keyof VgAxis> = {
   scale: 1,
+  gridScale: 1,
```
kanitw: Not sure why you have to move this down. -- It's causing so many diffs

domoritz: It had to move from AXIS_PROPERTIES_INDEX to VG_AXIS_PROPERTIES_INDEX

domoritz:
I'm not going to add the vg prefix because it causes a lot of work when we want to import something from Vega (we have to repeat this for every import) and it easily gets out of date.

domoritz:
No be wise the auto import will pull from Vega directly. We decided that for utils for example we do not re export. Hence, we should not re export typings either.

kanitw:
So, wise guy -- if you re-export in `vega.schema.ts`, and then type `VgSortField` when you import, the auto import will correctly import from `vega.schema.ts`.  Easy enough, no confusion.  

Util is different as it doesn't lead to confusion.

domoritz:
I didn't add prefix for BaseAxis, though and so I wonder what the rule should be. If there is the same property in Vega and in Vega-Lite but with different defintions, then we add the prefix?

kanitw:
>  If there is the same property in Vega and in Vega-Lite but with different definitions, then we add the prefix?

I think anything that's Vega-specific and not for Vega-Lite, we should add prefix. Otherwise, it's quite hard to read the method signature and understand them. 

For example, `FormulaTransform` should have a Vg prefix even though it's called `CalculateTransform` in Vega-Lite -- having the prefix just make it clear that it's Vega only. `FieldValue` also sounds confusing so we should make it clear that it's Vega. 

I think in general, we should add prefix to avoid confusion -- except for things that are identical in Vg and Vl. For BaseAxis, it is okay to not have prefix because it's the base Axis for both Vg and Vl.

> (we have to repeat this for every import) and it easily gets out of date.

Also, having a single file to do aliasing means we only need to fix it in one place.  (Importing them separately = fix in every file it is imported.)

domoritz:
What about all of these things?

```
import {
  Align,
  BaseLegend,
  FontWeight,
  LabelOverlap,
  Legend as VgLegend,
  LegendConfig as VgLegendConfig,
  LegendOrient,
  Orientation,
  SymbolShape,
  TextBaseline
} from 'vega';
```

kanitw:
Align, FontWeight, LabelOverlap, LegendOrient, Orientation, SymbolShape, TextBase are identical in VL and VG.  (All of these will appear in VL json schema, so definitely should not have Vg prefix. -- Perhaps, it's worth checking for no `Vg` in `check-schema.sh`) 

These two also make sense to have `Vg-` prefix.
```
Legend as VgLegend,
LegendConfig as VgLegendConfig,
```

kanitw:
Hmm I'm trying to look at how long does it take to build example.  It seems like the second build for build example is somehow incorrect.

image

If we won't parallel VM, we actually should not have to re-build again for building examples.

domoritz:
Did you run yarn? The schema builds correctly on my machine and on travis.

kanitw:
> Did you run yarn? The schema builds correctly on my machine and on travis.

The SS is from Travis log of this PR.

domoritz:
Idk, the last build works fine. 

screen shot 2018-07-17 at 10 22 48

I blame the github outage foo from yesterday.

kanitw:
Oh it's actually only problematic for Node 8 in `travis-ci/pr`.  (It shows the error in my SS despite passing the check. -- I think the checkout we do at the top of the check and fix could be wrong.)

domoritz:
Revised

src/compile/data/sample.ts
```diff
@@ -1,4 +1,4 @@
-import {SampleTransform as VgSampleTransform} from 'vega-typings';
+import {SampleTransform as VgSampleTransform} from 'vega';
```
kanitw: It might be worth doing

kanitw: It's probably worth centralizing aliasing for transform. 

Basically do 

`export {BinTransform as VgBinTransform, ...} from 'vega';`

so we don't have to do alias in other files.


This can be done later though (e.g., when @invokesus refactor to use vega typings for all transforms).

kanitw:
You merge it in the wrong order ...

PR #4043, opened 2018-07-17
Run tests on parallel VMs

domoritz:


scripts/ci.sh
```diff
@@ -0,0 +1,28 @@
...
+    ;;
+  "test")
+    yarn jest test/ --collectCoverage=true
+    yarn codecov
+    yarn jest examples/
+    yarn test:runtime
+    bash <(curl -s https://codecov.io/bash)
+    ;;
+  "fix")
+    ./scripts/check-and-fix.sh
```
kanitw: check and fix also check for schema diff, so this doesn't work.

scripts/ci.sh
```diff
@@ -0,0 +1,29 @@
...
+case "$1" in
+  "build")
+    yarn lint
+    yarn build
+    ./scripts/check-schema.sh
+    ;;
+  "test")
+    yarn jest test/ --collectCoverage=true
+    yarn codecov
+    yarn jest examples/
```
kanitw: Don't you have to run schema before example test?

scripts/ci.sh
```diff
@@ -0,0 +1,29 @@
...
+    yarn test:runtime
```
kanitw: Does run time depend on build? (Maybe not, just asking)

.travis.yml
```diff
@@ -7,25 +7,21 @@ node_js:
...
-  - cp parallel sem
-  - sudo mv parallel sem /usr/local/bin
+env:
+  global:
+    secure: I9sReagJiE7B0154AnU3t8WcdPtmZm86vMp+6umKegvzYr46jbypPAmlWW50A09fw4AmTaGOtwpYtvglf0JikkR6saXY0rRquDnO70DRmbdrO2o2WMjjCzkYID58SkGvqsqVJ4HXcu0HdJp6o5L7v7JMdzji6abYKxtsnw2ouk0=
+
+  matrix:
+    - TEST_SCRIPT=build
+    - TEST_SCRIPT=test
+    - TEST_SCRIPT=fix
```
kanitw: Perhaps, update is a better name -- now that we rely on Travis to update example (and optionally schema)?   (Though the prettifier part is technically still a fix.)

scripts/ci.sh
domoritz: Don't we have the schema in the repo?

scripts/ci.sh
domoritz: No, I don't think so.

scripts/check-and-fix.sh
```diff
@@ -4,17 +4,16 @@ git config --global user.email "travis@travis-ci.org"
...
 git checkout $TRAVIS_BRANCH
 
-node_version=$(node --version | cut -d. -f1 | cut -c2-)
-
 echo ""
 echo "------- Checking Schema -------"
 echo ""
 
 # Commit the schema if outdated
+yarn schema
```
kanitw: Ah I got an idea!

Since schema takes a minute, I think it's better to split this file into  

1) Update schema
2) Fix Lint
3) Update Example

1 and 2 could go into the "build" VM (since build VM has to run lint and fix schema anyway so we better run lint fix before linting and leverage the schema built in the VM.

The "fix" VN could then only become "update example", which already takes time on its own especially when SVG changes.

domoritz:
Let's try that!

scripts/check-and-fix.sh
domoritz: Isn't the problem that we might push a change from the update schema VM and the update example VM that both trigger a build then?

kanitw: Fair.   Then it's actually more time to restart the Travis Builds. (not worth the schema building time anymore)

domoritz: Somehow the builds are really slow in this PR anyway. Not sure why.

domoritz:
Why is it so slow? 40 minutes and counting? https://travis-ci.org/vega/vega-lite/builds/404977749

domoritz:
I don't have the cycles to debug this right now.

domoritz:
https://github.com/vega/vega-lite/pull/4193 is better.

PR #4048, opened 2018-07-17
Make removeChild private -- people should use `node.remove()`

kanitw:


domoritz:
The message is a bit confusing. What do you want people to use?

kanitw:
corrected*

invokesus:
`removeChild` is useful for inserting a new node in between two existing nodes.
E.g. in the Parse Merging case I want to insert a new node and only mutate(not remove) the existing nodes to keep only the conflicting parse fields.
ie. I want do 
```
A.removeChild(B);
B.parent(D);
```
and same for C
```
  A                          A -D
 / \              ->           /\
B   C                         B' C'
                                    
 ```
Maybe writing an `insert` method in `DataFlowNode` makes sense.

**Comment by kanitw** [comment]
I see.

Suppose we have the follow structure where B and C are parse nodes.

```
     A
   /   \
   B   C
   |   |
   D   E
```

One possible solution is to merge C into B (so we get B' and C', supposes some part of C isn't mergeable.) since the first one is always mergeable anyway. 


```
     A
    /   
   B'   
  /  \   
  D   C'
      |
      E
```

kanitw:
I'd say try to avoid using `removeChild` if possible, but if you really need it, that's fine.

PR #4051, opened 2018-07-17
Make scale and orient come first in axis

kanitw:
(We have to merge this to master as it was previously merged in the wrong order.)


kanitw:
Hmm, somehow example doesn't change

kanitw:
NVM, it was already like that.

kanitw:
So my test is useless.

PR #4054, opened 2018-07-18
Only parse aggregate as number when op is min or max and field quantitative

Saba9:


domoritz:
Depends on https://github.com/vega/vega-transforms/pull/9

domoritz:
Th main point here is that if we have an aggregate node then a parse node should move through the aggregate node but switch the field from the output to the input.

Saba9:
This is already in master.

PR #4061, opened 2018-07-20
Fix size config bugs

kanitw:


src/util.ts
```diff
@@ -333,3 +333,15 @@ export function accessPathDepth(path: string) {
   }
   return splitAccessPath(path).length;
 }
+
+/**
+ * This is a replacement for chained || for numeric properties or properties that respect null so that 0 will be included.
+ */
+export function getFirstDefined<T>(...args: T[]): T {
```
domoritz: ❤️

kanitw: lol so do you approve the PR.

PR #4064, opened 2018-07-20
Refactor `isExplicit` for axis/legend and remove unused axis `title` method

kanitw:


src/compile/axis/parse.ts
```diff
@@ -188,6 +188,24 @@ function getFieldDefTitle(model: UnitModel, channel: 'x' | 'y') {
   return undefined;
 }
 
+function isExplicit(value: any, property: keyof VgAxis, axis: Axis, model: UnitModel, channel: PositionScaleChannel) {
```
domoritz: `function isExplicit<T>(value: T, ...)`?

src/compile/legend/parse.ts
```diff
@@ -51,6 +51,21 @@ function getLegendDefWithScale(model: UnitModel, channel: NonPositionScaleChanne
   }
 }
 
+function isExplicit(value: any, property: keyof VgLegend, legend: Legend, fieldDef: FieldDef<string>) {
```
domoritz: same as above

PR #4065, opened 2018-07-20
Add interpolate as one of the errorband properties.

chanwutk:


src/compositemark/errorband.ts
```diff
@@ -43,6 +45,11 @@ export interface ErrorBandConfig extends ErrorBandPartsMixins {
    * __Default value:__ `"stderr"`.
    */
   extent?: ErrorBarExtent;
+
+  /**
+   * The interpolate options for error band parts.
```
kanitw: Let's copy docs from normal `interpolate` so it includes all the interpolation property and add this property the errorband.md. :)

src/compositemark/errorband.ts
```diff
@@ -82,8 +89,18 @@ export function normalizeErrorBand(
...
   const is2D = spec.encoding.x !== undefined && spec.encoding.y !== undefined;
-  const bandMark = is2D ? 'area' : 'rect';
-  const bordersMark = is2D ? 'line' : 'rule';
+
+  const bandMark: MarkDef = {type: is2D ? 'area' : 'rect'};
+  const bordersMark: MarkDef = {type: is2D ? 'line' : 'rule'};
+
+  if (markDef.interpolate) {
+    if (is2D) {
+      bandMark.interpolate = markDef.interpolate;
```
kanitw: Actually let's add [`tension`](https://vega.github.io/vega-lite/docs/line.html#properties) and apply it too.

chanwutk:
@kanitw Is `area` not working with some value of `interpolate` and `tenstion`?
I tested this spec and it does not work when `tension` is 0, but works fine when `tension` is 1:
```{
  "mark": {
    "type": "area",
    "opacity": 0.3,
    "interpolate": "bundle",
    "tension": 1
  },
  "encoding": {
    "y": {
      "field": "Miles_per_Gallon",
      "type": "quantitative",
      "title": "Mean of Miles per Gallon (95% CIs)",
      "aggregate": "ci0"
    },
    "y2": {
      "field": "Miles_per_Gallon",
      "type": "quantitative",
      "aggregate": "ci1"
    },
    "x": {"field": "Year", "type": "temporal", "timeUnit": "year"}
  }
}```

kanitw:
> I tested this spec and it does not work when tension is 0, but works fine when tension is 1:

I guess tension should be non-zero?

PR #4067, opened 2018-07-21
Apply get first defined where appropriate

kanitw:


src/compile/axis/assemble.ts
```diff
@@ -91,7 +91,7 @@ export function assembleAxis(
       grid: false,
       ...(titleString ? {title: titleString} : {}),
       ...axis,
-      zindex: zindex !== undefined ? zindex : 1 // put axis line above marks by default
+      zindex: zindex || 1 // put axis line above marks by default
```
domoritz: What if the user sets zindex to 0?

src/compile/common.ts
```diff
@@ -125,13 +125,6 @@ export function formatSignalRef(
   }
 }
 
-export function getSpecifiedOrDefaultValue<T>(specifiedValue: T, defaultValue: T | {signal: string}) {
```
domoritz: 🎉

src/compile/axis/assemble.ts
kanitw: Good catch!

Issue #1830, opened 2017-01-16
APIs to interact with Selection's Data and Signals


kanitw:

kanitw:
This [question in the vega-js group](https://groups.google.com/forum/?fromgroups=#!topic/vega-js/Hbc72LqP9hA) is very relevant to this issue.

domoritz:
@arvind Can you post an example of how one might access a selection? I'm trying to build an application where you can crossfilter between different visualization. Since I want to re-query the data, I cannot use a single spec for the different charts.

arvind:
Sure. The selection states are stored in datasets named `selectionName_store` (e.g., if you had a selection named `brush`, the dataset would be named `brush_store`). Accessing the dataset via the view api (`view.data('brush_store')`) gives you the constituent queries for each of the selection instances (i.e., no resolution will be performed). In the case of point (single/multi) selections, this will be an array of values; for interval selections it will be the data extents. You can similarly set the selection state via the view API provided the tuples you insert follow the same structure. Note: for interval selections, inserting new selection instances via the view API may not always correctly update the brush mark state.

domoritz:
> Note: for interval selections, inserting new selection instances via the view API may not always correctly update the brush mark state.

Why is that? Is there a way to correct the brush?

arvind:
The brush mark is currently driven by signals within each unit. It's difficult to update these signals based on updates to the backing dataset (we need to identify the matching tuple, and extract information from it). @jheer and I decided that this would be a limitation with 2.0 that we would address in subsequent releases once we better understood how users wanted to update selections via the API.

arvind:
To make forward progress on this, I would like to decouple an API for writing to selections (which involves hairier design/implementation issues) from reading selections (which should hopefully be more straightforward). Here're some ideas sketching out the latter.

### New API Methods

* `vl.selection(view, selectionName)` -- returns an array of tuples that define a selection's predicate, respecting any resolution rules. For example, `[{Origin: 'Japan', Year: 1981}, {Origin: 'USA', Year: 1982}]` for a multi selection or `[{Horsepower: [40, 150], Miles_per_Gallon: [40, 15]}]` for an interval selection. 

* `vl.addSelectionListener(view, selectionName, handler)` and `view.removeSelectionListener(view, selectionName, handler)`.

### Notes

[...]

PR #4068, opened 2018-07-21
Refactor + Standardize Selection State Representation

arvind:
A major refactoring of how a selection state is stored within its dataset. This state uses tuples of the form:

```
{unit: string, fields: array<fielddef>, values: array<*>}.
```

where `fielddef` is an object of the form:

```
{field: string, channel: string, type: 'E' | 'R' | ... }
```
A `fielddef`'s `type` identifies whether the values being stored are simple enumerations (`E`) or represent ranges (`R`) of different kinds (e.g., a right-exclusive range would be `R-RE`). 

This new state requires corresponding new Vega expression functions (vega/vega-parser#89). Besides simplifying several code paths, it's also allowed us to much more easily add a top-level signal named (partially addressing #1830 with a much simpler strategy than I described here: https://github.com/vega/vega-lite/issues/1830#issuecomment-379000974).

I've marked this PR as WIP until I've had a chance to test the changes more thoroughly, but it's ready for high-level feedback. 

Note: While this PR does not introduce any breaking changes to the Vega-Lite language, it is tied to whatever Vega version vega/vega-parser#89 lands in. So, in that sense, it is a "breaking change" that might warrant us bumping the major version number of Vega-Lite.


arvind:
*Siiiigh*, the issue with `intersect` failing tests was due to a stale local build of Vega. 😭 Those tests have been re-enabled and everything passes locally for me now.

domoritz:
I added a blocked tag. We can remove it once the necessary changes land in Vega.

domoritz:
I guess we are waiting for a Vega release before we merge this. Is there anything else left?

domoritz:
@arvind @jheer This PR is blocked by a Vega release only, right?

arvind:
Yes, I think so.

arvind:
Hooray, with Vega 4.3 this PR now passes CI and is officially ready for review/merge @domoritz, @kanitw!

domoritz:
Not sure why, buy a bunch of SVGs are gone in this PR.

domoritz:
@arvind do you know what's up?

arvind:
Huh, that's a great question. Looks like all of those deletions seem to have happened via the `[Travis]` commits? I'm not familiar with how those are run and, off the top of my head, nothing in my commits for this PR should cause example deletions so I'm as puzzled as you...

arvind:
One thought: is it actually deletions or an error in generating the files? It looks like they're all 1 byte sized, and viewing them in the rich diff view produces an "file is invalid" error. So perhaps Travis is having difficulty producing the SVGs for some reason?

arvind:
Wow, I'm really confused about what is going on. After the runtime tests execute, Travis seems to go [completely off the rails](https://travis-ci.org/vega/vega-lite/jobs/440766451#L1124) and no longer gets to the example generation stage. Hmm.

arvind:
@domoritz Looks like the error with SVGs has gone away with the fixes to the Travis process. So this PR should now be actually ready for review.

arvind:
Thanks for the LGTM and the code cleanup @domoritz. I squashed #4216 into this PR but do you feel strongly about squashing this PR into master? My preference would be to rebase it to preserve the granularity of code commits in what is a pretty major refactoring. It'll help (me, in particular) remember design decisions more easily. I worry that squashing it will produce a single, extremely complex commit and it will be hard to untangle all the various components that went into the refactoring.

domoritz:
The advantage of squashing is that we can read the commits to write the changelog. Moreover, you get a reference to the PR in the commit message. If we need to look into what went into a commit, we can go to the PR, read the comments and also see the individual commits. This only works if you go to Github. We have squashed complex PRs in the past. For this PR, I'm also okay with merging if you use `--no-ff`. Go ahead with whichever you prefer. I approve the PR.

arvind:
Thanks for describing the rationale. The alternate workflow you propose makes sense to me, so I'll go ahead and squash this PR.

PR #4070, opened 2018-07-21
Make Value Def Generic 

kanitw:
so schema is more accurate

Depends on #4426 and #4430.


scripts/rename-schema.sh
```diff
@@ -11,6 +11,14 @@ perl -pi -e s,'Generic(.*)<FacetedUnitSpec\,LayerSpec>','\1',g build/vega-lite-s
 perl -pi -e s,'GenericUnitSpec<EncodingWithFacet\,AnyMark>','FacetedCompositeUnitSpecAlias',g build/vega-lite-schema.json
 perl -pi -e s,'GenericUnitSpec<Encoding\,AnyMark>','CompositeUnitSpecAlias',g build/vega-lite-schema.json
 
+perl -pi -e s,'FieldDefWithoutScale','FieldDef',g build/vega-lite-schema.json
```
domoritz: Do you think that we might ever accidentally rename two distinct properties that then become the same? Any idea how we could test it automatically?

kanitw: If that's the case, we will have redundant keys in `schema.definitions` and the schema will no longer be a valid json.

domoritz:
still wip?

kanitw:
> still wip?

Yes -- look a the schema diff.  still terrible.

domoritz:
@kanitw What do you think about this PR? If we want it, we should try to get it in before 3.0 to not break the schema later.

PR #4072, opened 2018-07-21
NPM package with es2015 imports instead of commonjs

domoritz:


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "outDir": "./build/",
     "target": "es5",
-    "module": "ES2015",
+    "module": "es2015",
```
kanitw: lol

PR #4073, opened 2018-07-21
Use transform typings from vega-typings

invokesus:


domoritz:
Next time try not to update everything in yarn.lock to isolate the change.

kanitw:
I still see a lot of TODOs.  Why removing the WIP label?

domoritz:
I think the todos are copies from the Vega typings?

domoritz:
I rejected the typings PR but suggested a solution here. Ping me when this is ready.

domoritz:
@invokesus Do you want to finish this PR?

invokesus:
Yup!

domoritz:
This probably needs to be redone.

PR #4076, opened 2018-07-23
Auto generated tool tips now check the format and title fields of a channe's axis and legend when appropriate

Saba9:
We used to not check the axis or legend for a specified title or format when generating the key value pairs used in tooltips. This PR adds checks to see if these fields are defined, if so, it'll use them to populate the tooltip keys and values.

### DIFF
--- a/src/compile/mark/valueref.ts
+++ b/src/compile/mark/valueref.ts
@@ -12,6 +12,8 @@ import {
   FieldDef,
   FieldRefOption,
   isFieldDef,
+  isMarkPropFieldDef,
+  isPositionFieldDef,
   isValueDef,
   TextFieldDef,
   title,
@@ -206,7 +208,14 @@ export function tooltipForChannelDefs(channelDefs: FieldDef<string>[], config: C
   const keyValues: string[] = [];
   const usedKey = {};
   for (const fieldDef of channelDefs) {
-    const key = getFirstDefined(fieldDef.title, title(fieldDef, config));
+    let key;
+    if (isPositionFieldDef(fieldDef) && fieldDef.axis && fieldDef.axis.title) {
+      key = fieldDef.axis.title;
+    } else if (isMarkPropFieldDef(fieldDef) && fieldDef.legend && fieldDef.legend.title) {
+      key = fieldDef.legend.title;
+    } else {
+      key = getFirstDefined(fieldDef.title, title(fieldDef, config));
+    }
     const value = text(fieldDef, config).signal;
     if (!usedKey[key]) {
       keyValues.push(`"${key}": ${value}`);
@@ -216,13 +225,22 @@ export function tooltipForChannelDefs(channelDefs: FieldDef<string>[], config: C
   return keyValues.length ? {signal: `{${keyValues.join(', ')}}`} : undefined;
 }
 
-export function text(textDef: ChannelDefWithCondition<TextFieldDef<string>>, config: Config): VgValueRef {
+export function text(channelDef: ChannelDefWithCondition<TextFieldDef<string>>, config: Config): VgValueRef {
   // text
-  if (textDef) {
-    if (isFieldDef(textDef)) {
-      return formatSignalRef(textDef, textDef.format, 'datum', config);
-    } else if (isValueDef(textDef)) {
-      return {value: textDef.value};
+  if (channelDef) {
+    if (isValueDef(channelDef)) {
+      return {value: channelDef.value};
+    }
+    let format;
+    if (isFieldDef(channelDef) && channelDef.format) {
+      format = channelDef.format;
+    } else if (isPositionFieldDef(channelDef) && channelDef.axis && channelDef.axis.format) {
+      format = channelDef.axis.format;
+    } else if (isMarkPropFieldDef(channelDef) && channelDef.legend && channelDef.legend.format) {
+      format = channelDef.legend.format;
+    }
+    if (isFieldDef(channelDef)) {
+      return formatSignalRef(channelDef, format, 'datum', config);
     }
   }
   return undefined;
--- a/src/fielddef.ts
+++ b/src/fielddef.ts
@@ -338,6 +338,14 @@ export function isScaleFieldDef<F>(channelDef: ChannelDef<F>): channelDef is Sca
   return !!channelDef && (!!channelDef['scale'] || !!channelDef['sort']);
 }
 
+export function isPositionFieldDef<F>(channelDef: ChannelDef<F>): channelDef is PositionFieldDef<F> {
+  return !!channelDef && (!!channelDef['axis'] || !!channelDef['stack'] || !!channelDef['impute']);
+}
+
+export function isMarkPropFieldDef<F>(channelDef: ChannelDef<F>): channelDef is MarkPropFieldDef<F> {
+  return !!channelDef && !!channelDef['legend'];
+}
+
 export interface FieldRefOption {
   /** Exclude bin, aggregate, timeUnit */
   nofn?: boolean;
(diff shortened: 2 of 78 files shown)


src/compile/mark/valueref.ts
```diff
@@ -206,7 +208,14 @@ export function tooltipForChannelDefs(channelDefs: FieldDef<string>[], config: C
   const keyValues: string[] = [];
   const usedKey = {};
   for (const fieldDef of channelDefs) {
-    const key = getFirstDefined(fieldDef.title, title(fieldDef, config));
+    let key;
+    if (isPositionFieldDef(fieldDef) && fieldDef.axis && fieldDef.axis.title) {
```
kanitw: Maybe better include this logic in the `title()` function that you call here, so it's reusable in many other places. 

Also, check if it is FacetFieldDef and access `header.title`.

This is a good opportunity to clean this up.
(Try find all references of `title?:` in `TitleMixins` and all references of the `title()` method.)

src/compile/mark/valueref.ts
```diff
@@ -216,13 +225,22 @@ export function tooltipForChannelDefs(channelDefs: FieldDef<string>[], config: C
...
-  if (textDef) {
-    if (isFieldDef(textDef)) {
-      return formatSignalRef(textDef, textDef.format, 'datum', config);
-    } else if (isValueDef(textDef)) {
-      return {value: textDef.value};
+  if (channelDef) {
+    if (isValueDef(channelDef)) {
+      return {value: channelDef.value};
+    }
+    let format;
```
kanitw: Extract a format function? 
(And check all references of `format`, so we can reuse it and fix similar problem elsewhere?)

src/compile/mark/valueref.ts
```diff
@@ -216,13 +217,15 @@ export function tooltipForChannelDefs(channelDefs: FieldDef<string>[], config: C
...
-  if (textDef) {
-    if (isFieldDef(textDef)) {
-      return formatSignalRef(textDef, textDef.format, 'datum', config);
-    } else if (isValueDef(textDef)) {
-      return {value: textDef.value};
+  if (channelDef) {
+    if (isValueDef(channelDef)) {
+      return {value: channelDef.value};
+    }
+    const formatString = isFieldDef(channelDef) ? format(channelDef) : undefined;
```
kanitw: Why checking isFieldDef multiple time & swapping the condition order? 

I think the code should be pretty much the same as the original code except replacing `textDef.format` with ` format(channelDef)`.

There is also no need to rename the variable `textDef`.

src/fielddef.ts
```diff
@@ -338,6 +339,18 @@ export function isScaleFieldDef<F>(channelDef: ChannelDef<F>): channelDef is Sca
...
 
+export function isPositionFieldDef<F>(channelDef: ChannelDef<F>): channelDef is PositionFieldDef<F> {
+  return !!channelDef && (!!channelDef['axis'] || !!channelDef['stack'] || !!channelDef['impute']);
+}
+
+export function isMarkPropFieldDef<F>(channelDef: ChannelDef<F>): channelDef is MarkPropFieldDef<F> {
+  return !!channelDef && !!channelDef['legend'];
+}
+
+export function isTextFieldDef<F>(channelDef: ChannelDef<F>): channelDef is TextFieldDef<F> {
```
domoritz: Wouldn't it make more sense to check the channel and then get the field def (which should be a text field def). Otherwise, this test isn't very safe.

src/fielddef.ts
```diff
@@ -485,9 +498,34 @@ export function resetTitleFormatter() {
 }
 
 export function title(fieldDef: FieldDefBase<string>, config: Config) {
+  if (isFieldDef(fieldDef)) {
+    if (isPositionFieldDef(fieldDef) && fieldDef.axis && fieldDef.axis.title) {
```
kanitw: This logic will ignore if user intentionally specify title to be '' or false.

For axis/legend/header, we want to respect if users want to disable the title.
However, for tooltip, we should ignore the disabling. 

Thus, let's one more object params:   `, {allowDisabling}: {allowDisabling: boolean}`.  
Note that this is more readable than boolean param `allowDisabling: boolean` as the method will be called like `title(fieldDef, config, {allowDisabling: true}` instead having a vague boolean param  `title(fieldDef, config, true)`

src/compile/legend/parse.ts
```diff
@@ -126,7 +126,6 @@ function getProperty(
       return (
         getFirstDefined(
           specifiedLegend.title, // legend title has higher precedence than fieldDef title
-          fieldDef.title,
```
kanitw: Why not simply using `fieldDefTitle(fieldDef, model.config) || undefined`?

(No need to use `getFirstDefined` anymore)

src/fielddef.ts
```diff
@@ -315,7 +316,7 @@ export function hasConditionalValueDef<F>(
 }
 
 export function isFieldDef<F>(
-  channelDef: ChannelDef<F>
+  channelDef: ChannelDef<F> | FieldDefBase<F>
```
kanitw: Why would FieldDefBase needs to be rechecked if it is a field def? This seems wrong.

src/fielddef.ts
```diff
@@ -485,9 +498,34 @@ export function resetTitleFormatter() {
 }
 
 export function title(fieldDef: FieldDefBase<string>, config: Config) {
+  if (isFieldDef(fieldDef)) {
```
kanitw: Why would FieldDefBase needs to be rechecked if it is a field def? This seems wrong.
Also, if this is not a fieldDef, the last line outside this `if` would be wrong too.

src/compile/mark/valueref.ts
kanitw: Actually renaming textDef to channelDef makes sense, but you have to change the type to be general FieldDef too.

src/fielddef.ts
kanitw: That would require the whole encoding and would require a lot of refactor, so I think this is better.

For text fieldDef without format, it is behaving like a normal fielddef anyway.

src/compile/legend/parse.ts
Saba9: I believe making that change would modify the function's behavior as it would no longer consider `specifiedLegend.title`. We could probably change it to `specifiedLegend.title || fieldDefTitle(fieldDef.title, model.config)`

src/fielddef.ts
Saba9: The reason for this was because all the `is*FieldDef` functions require a `ChannelDef` as their input and as a result would not accept a `FieldDefBase`, therefore, I couldn't pass `fieldDef` to them without first running it through `isFieldDef`.

src/fielddef.ts
Saba9: I added this so that I could use the `is*FieldDef` (e.g. `isTextFieldDef`, `isPositionFieldDef`) methods with a `FieldDefBase`. This is why I make the call to `isFieldDef` which you mention below.

src/fielddef.ts
kanitw: Hmm still doesn’t sound like we are solving the problem at its cause here  — let’s discuss more in person

src/compile/legend/parse.ts
kanitw: specifiedLegend is fieldDef.legend

So what I suggested is right

src/fielddef.ts
kanitw: (We better try to remove these two weird changes and solve the problem in other places with a less hacky solution)

kanitw: For example, why don'tee make is*FieldDef require FieldDefBase instead of ChannelDef

kanitw:
@Saba9  I extracted the refactoring part to #4079 and added one more refactor.

PR #4077, opened 2018-07-24
Rename getFirstDefined => coalesce

domoritz:
It’s the name that SQL uses. See https://docs.microsoft.com/en-us/sql/t-sql/language-elements/coalesce-transact-sql?view=sql-server-2017/


kanitw:
Hmm. It's quite confusing though:

image

domoritz:
Coalesce is really hard to type and probably harder to understand for newcomers.

PR #4080, opened 2018-07-27
Make config.timeFormat affect only text, header, and labels but not axis + Bring the default back

kanitw:
Rationale: we previously remove this since default axis format is smarter in Vega.
However, removing this caused text and header labels to have no default time format at all and thus text can disappear.

For legend, since we use sequential scale by default, the text would disappear too. (Legend default will be smart if we use time scale, but the color would be inferior.)


domoritz:
> Legend default will be smart if we use time scale, but the color would be inferior

Shouldn't there be a sequential time scale then?

kanitw:
> Shouldn't there be a sequential time scale then?

This is why I say sequential scale is quite confusing and unnecessary.  

Better redirect apply sequential to any continuous scales when scheme is specified for range (and remove the concept of "sequential" altogether.) 

Anyway, that's beyond the scope of this PR.

PR #4094, opened 2018-07-27
Extend docs about UTC

domoritz:


jakevdp:
Looks good. I might add something about this being the default behavior in Javascript, rather than a deliberate decision within vega.

PR #3890, opened 2018-06-18
Add parents of parse field to ParseNode.dependentFields

invokesus:
Fixes #3832 .

Issue #3832, opened 2018-05-29
Make parse for nested field come after the transform that produces it


invokesus:
- [  ] Fix the issue
- [  ] update the concat flatten example

```
{  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "values": [
      { "id": "001",
        "ra": 243.351,
        "dec": "+54.6",
        "lc": [{"time": 1, "mag": 18.5}, {"time": 2, "mag": 19}]
      },
      { "id": "002",
        "ra": 210.35,
        "dec": "+14.6",
        "lc": [{"time": 1, "mag": 19.5}, {"time": 2, "mag": 20}]
      },
      { "id": "003",
        "ra": 143.35,
        "dec": "+33.6",
        "lc": [{"time": 1, "mag": 19}, {"time": 2, "mag": 18}]
      }
    ]
  },
      "transform": [{"flatten": ["lc"]}],
      "mark": "line",
      "encoding": {
        "x": {
          "field": "lc.time",
          "type": "quantitative",
          "scale": {"zero": false}
        },
        "y": {"field": "lc.mag", "type": "quantitative"}
      }
}
```

domoritz:
Parse will be moved before transforms unless the transform specifies that it produces a field that needs to be parsed.

kanitw:
I guess existing logic doesn't know that `lc.time` is a part of `lc` that `flatten` produces.   

Point @invokesus to the place he should look?

domoritz:
https://github.com/vega/vega-lite/blob/master/src/compile/data/optimizers.ts#L32

kanitw:
Will be fixed in https://github.com/vega/vega-lite/pull/3890

PR #4097, opened 2018-07-30
Run optimizeDataflow multiple times

invokesus:


src/compile/data/optimizers.ts
```diff
@@ -28,48 +31,51 @@ export function iterateFromLeaves(f: (node: DataFlowNode) => boolean) {
...
   if (node instanceof OutputNode || node.numChildren() > 0 || node instanceof FacetNode) {
     // no need to continue with parent because it is output node or will have children (there was a fork)
-    return false;
+    return [false, false];
   } else {
+    flag = true;
     node.remove();
   }
-  return true;
+  return [true, flag];
```
domoritz: Let's return an object instead.

src/compile/data/optimize.ts
```diff
@@ -140,31 +148,51 @@ export function mergeParse(node: DataFlowNode) {
...
-  roots.forEach(mergeParse);
-
-  keys(dataComponent.sources).forEach(s => {
-    if (dataComponent.sources[s].numChildren() === 0) {
-      delete dataComponent.sources[s];
+export function optimizeDataflow(data: DataComponent) {
+  function helper(dataComponent: DataComponent) {
+    let roots: SourceNode[] = vals(dataComponent.sources);
+    let mutatedFlag = false;
+    // PITFALL Short Circuit logic
```
domoritz: This is not a very meaningful comment.

src/compile/data/optimizers.ts
invokesus: `{continue: true, mutated: flag}` ?

domoritz: Yep. It's much easier to interpret.

src/compile/data/optimize.ts
```diff
@@ -140,31 +148,51 @@ export function mergeParse(node: DataFlowNode) {
...
-  getLeaves(roots).forEach(optimizers.removeDuplicateTimeUnits);
-
-  roots.forEach(moveFacetDown);
-  roots.forEach(mergeParse);
-
-  keys(dataComponent.sources).forEach(s => {
-    if (dataComponent.sources[s].numChildren() === 0) {
-      delete dataComponent.sources[s];
+export function optimizeDataflow(data: DataComponent) {
+  function helper(dataComponent: DataComponent) {
```
kanitw: The method name helper is a bit confusing.  Basically this is a method that run one pass.  Give it a better name?

kanitw: Also, look likes it's unnecessary to nest this method inside `optimizeDataflow`.  So don't nest it then.

examples/compiled/vconcat_flatten.vg.json
```diff
@@ -45,7 +44,8 @@
           "type": "formula",
           "expr": "toNumber(datum[\"lc\"] && datum[\"lc\"][\"mag\"])",
           "as": "lc.mag"
-        }
+        },
+        {"type": "flatten", "fields": ["lc"], "as": ["lc"]}
```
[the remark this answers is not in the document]
invokesus: Yeah... some of the facet examples are broken.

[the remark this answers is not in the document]
invokesus: This is because of the #3832 .
`hasIntersection` returns `{}` for `{ lc: true }` & `{ ra: true, dec: true, 'lc.time': true, 'lc.mag': true }`.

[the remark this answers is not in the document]
domoritz: Shouldn't `lc` be in the list then?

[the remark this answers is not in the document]
invokesus: I had a go at that in #3890, but I think we decided to replace with `hasIntersection` with something that generated `lc` and *then* calculated the intersection instead of changing `dependentFields`.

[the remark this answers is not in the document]
domoritz: Ahh, I remember. @kanitw Do you know why we decided that? 

I think we either should use a different function to compute dependencies then or go the route that you initially went and add all fields in nested fields.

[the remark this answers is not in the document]
invokesus: @kanitw Any preferences?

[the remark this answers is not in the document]
domoritz: I'd say since you ran into this multiple times, let's add all fields that we depend on and deconstruct nested fields. I did something similar in https://github.com/vega/vega-lite/pull/4099

domoritz:
What's up with this?

screen shot 2018-08-05 at 19 29 15

domoritz:
Other than the issue above, this looks good to me.

invokesus:
Todo:
- [ ] Nested Field dependent bug fix
- [ ] Debug facet_independent_scale_layer_broken

src/compile/data/optimizers.ts
```diff
@@ -5,21 +5,34 @@ import {ParseNode} from './formatparse';
 import {SourceNode} from './source';
 import {TimeUnitNode} from './timeunit';
 
+export interface OptimizerFlags {
+  /**
+   * If true iteration continues
+   */
+  continueFlag: boolean;
```
domoritz: Let's rename these to just `continue` and `mutated` in a follow up PR. Also, add a comma to the comment: "If true, iteration continues."

invokesus: `continue` is a reserved keyword in javascript.

domoritz: Ups... then just fix the comment.

PR #4099, opened 2018-07-31
Calculate dependent fields for expressions.

domoritz:


kanitw:
Improve test cov would be nice

PR #4100, opened 2018-07-31
Remove the note for argmin/max issue in CO2 example

chanwutk:
Now, we are using `window` instead of `aggregate`; this example does not show a solution to the `argmin/max` issue.

### DIFF
--- a/_data/examples.json
+++ b/_data/examples.json
@@ -165,7 +165,8 @@
       },
       {
         "name": "layer_line_co2_concentration",
-        "title": "Carbon Dioxide in the Atmosphere"
+        "title": "Carbon Dioxide in the Atmosphere",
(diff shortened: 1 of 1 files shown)


_data/examples.json
```diff
@@ -165,7 +165,8 @@
       },
       {
         "name": "layer_line_co2_concentration",
-        "title": "Carbon Dioxide in the Atmosphere"
+        "title": "Carbon Dioxide in the Atmosphere",
+        "description": "This example was inspired by [Carbon Dioxide Concentration By Decade](https://www.datawrapper.de/_/OHgEm/) from Datawrapper"
```
domoritz: It's Gregor Aisch (who is at datawrapper)

PR #4109, opened 2018-08-02
Add filter for valid fields. Fixes #4098

### LINKED ISSUE #4098: Low contrast in binned heatmap
See https://vega.github.io/vega-lite/examples/rect_binned_heatmap.html

domoritz:


src/predicate.ts
```diff
@@ -139,12 +141,20 @@ export interface FieldOneOfPredicate extends FieldPredicateBase {
   oneOf: string[] | number[] | boolean[] | DateTime[];
 }
 
+export interface FieldValidPredicate extends FieldPredicateBase {
+  valid: true;
```
kanitw: This should support false too.  Users may want to analyze the pattern of data that have some fields missing.

site/docs/transform/filter.md
```diff
@@ -112,7 +112,16 @@ For example, to check if the `height` field's value is greater than or equals to
...
-For example, `{"filter": {"field": "car_color", "oneOf":["red", "yellow"]}}` checks if the `car_color` field's value is `"red"` or `"yellow"`.
+For example, `{"filter": {"field": "car_color", "oneOf": ["red", "yellow"]}}` checks if the `car_color` field's value is `"red"` or `"yellow"`.
+
+
+{:#valid-predicate}
+### Field Valid Predicate
+
+{% include table.html props="valid" source="FieldValidPredicate" %}
+
+For example, `{"filter": {"field": "car_color", "valid": true}}` checks if the `car_color` field's value is not `null` and whether it is not `NaN`.
```
kanitw: Adjust the language to make it clear which data items would be included.

src/predicate.ts
domoritz: You can always use `not`. I'd rather support fewer ways to do the same thing for now. We can change it later.

domoritz:
Done

src/predicate.ts
kanitw: I disagree.  Users wouldn't expect that it can't be `false`, so better fix it now since it's very easy and provides a way nicer syntax that nesting it under "not".

domoritz:
I added some tests

src/predicate.ts
```diff
@@ -216,7 +229,9 @@ export function fieldFilterExpression(predicate: FieldPredicate, useInRange = tr
     const lower = predicate.gte;
     return `${fieldExpr}>=${predicateValueExpr(lower, timeUnit)}`;
   } else if (isFieldOneOfPredicate(predicate)) {
-    return 'indexof([' + predicateValuesExpr(predicate.oneOf, timeUnit).join(',') + '], ' + fieldExpr + ') !== -1';
+    return `indexof([${predicateValuesExpr(predicate.oneOf, timeUnit).join(',')}], ${fieldExpr}) !== -1`;
+  } else if (isFieldValidPredicate(predicate)) {
+    return predicate.valid ? `${fieldExpr}!==null&&!isNaN(${fieldExpr})` : `${fieldExpr}===null||isNaN(${fieldExpr})`;
```
kanitw: I'd prefer having spaces for readability.  

It might be worth submitting another PR to standardize other the filter expressions too.  
(Currently, some do have spaces between operators, some do not.)

domoritz: Let's do this in a follow up. Issue: https://github.com/vega/vega-lite/issues/4112

PR #4115, opened 2018-08-03
Support strict typings with declarations

domoritz:


domoritz:
This blocks a release of Vega-Embed right now.

PR #4118, opened 2018-08-05
Make ParseNode Merging Logic Bottom-Up

invokesus:
Merge after #4097 is merged and #3832 is resolved.

### DIFF
--- a/src/compile/data/optimize.ts
+++ b/src/compile/data/optimize.ts
@@ -1,4 +1,5 @@
 import {MAIN} from '../../data';
+import * as log from '../../log';
 import {flatten, keys, vals} from '../../util';
 import {AggregateNode} from './aggregate';
 import {DataFlowNode, OutputNode} from './dataflow';
@@ -206,19 +207,25 @@ function optimizationDataflowHelper(dataComponent: DataComponent) {
 export function optimizeDataflow(data: DataComponent) {
   // check before optimizations
   checkLinks(vals(data.sources));
-
-  for (let i = 0; i < 5; i++) {
+  let warningFlag = false;
+  for (let i = 0; i < MAX_OPTIMIZATION_RUNS; i++) {
     if (!optimizationDataflowHelper(data)) {
       break;
+    }
+    if (i === MAX_OPTIMIZATION_RUNS - 1) {
+      warningFlag = true;
     }
   }
 
   // move facets down and make a copy of the subtree so that we can have scales at the top level
   vals(data.sources).map(moveFacetDown);
 
-  for (let i = 0; i < 5; i++) {
+  for (let i = 0; i < MAX_OPTIMIZATION_RUNS; i++) {
     if (!optimizationDataflowHelper(data)) {
       break;
+    }
+    if (warningFlag || i === MAX_OPTIMIZATION_RUNS - 1) {
+      log.warn(`Maximum optimization runs(${MAX_OPTIMIZATION_RUNS}) reached.`);
     }
   }
 
--- a/src/compile/data/optimize.ts
+++ b/src/compile/data/optimize.ts
@@ -124,10 +124,8 @@ function getLeaves(roots: DataFlowNode[]) {
 export function mergeParse(node: DataFlowNode): optimizers.OptimizerFlags {
   let mutatedFlag = false;
   const parent = node.parent;
-  if (parent === undefined) {
-    return {continueFlag: false, mutatedFlag: false};
-  }
   const parseChildren = parent.children.filter((x): x is ParseNode => x instanceof ParseNode);
+
   if (parseChildren.length > 1) {
     const commonParse = {};
     for (const parseNode of parseChildren) {
@@ -207,14 +205,14 @@ function optimizationDataflowHelper(dataComponent: DataComponent) {
 export function optimizeDataflow(data: DataComponent) {
   // check before optimizations
   checkLinks(vals(data.sources));
-  let warningFlag = false;
+  let firstPassCounter = 0;
+  let secondPassCounter = 0;
+
   for (let i = 0; i < MAX_OPTIMIZATION_RUNS; i++) {
     if (!optimizationDataflowHelper(data)) {
       break;
     }
-    if (i === MAX_OPTIMIZATION_RUNS - 1) {
-      warningFlag = true;
-    }
+    firstPassCounter++;
   }
 
   // move facets down and make a copy of the subtree so that we can have scales at the top level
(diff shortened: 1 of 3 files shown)


invokesus:
As @Saba9 pointed out in #4126, `moveParseUp` has to be performed on all leaves before merging `ParseNode`s, so keeping separate functions is better.

domoritz:
> so keeping separate functions is better

I think it makes sense to have separate functions to run a particular optimization but it still makes sense to have a single function that walks over the tree and calls the optimizers, no?

invokesus:
> a single function that walks over the tree and calls the optimizers

Yes. I was talking about having a single function for both `mergeParse` and `moveParseUp` is not a good idea.

src/compile/data/optimize.ts
```diff
@@ -41,11 +41,12 @@ function cloneSubtree(facet: FacetNode) {
  * Move facet nodes down to the next fork or output node. Also pull the main output with the facet node.
  * After moving down the facet node, make a copy of the subtree and make it a child of the main output.
  */
-function moveFacetDown(node: DataFlowNode) {
+function moveFacetDown(node: DataFlowNode): boolean {
+  let flag = false;
```
Saba9: Might want to rename `flag` to something like `mutatedFlag` to make it clear what we're tracking

src/compile/data/optimize.ts
```diff
@@ -112,8 +120,13 @@ function getLeaves(roots: DataFlowNode[]) {
 /**
  * Inserts an Intermediate ParseNode containing all non-conflicting Parse fields and removes the empty ParseNodes
  */
-export function mergeParse(node: DataFlowNode) {
-  const parseChildren = node.children.filter((x): x is ParseNode => x instanceof ParseNode);
+export function mergeParse(node: DataFlowNode): optimizers.OptimizerFlags {
```
Saba9: I wonder if it would make more sense to have this function merge the children of `node` rather than having it access `node`'s parent and then merge the parent's children. I think making this change might simplify the function as you'd no longer have check if a parent exists.

src/compile/data/optimize.ts
```diff
@@ -127,45 +140,71 @@ export function mergeParse(node: DataFlowNode) {
...
     }
   });
+  return mutatedFlag;
+}
+
+/**
+ * Optimizes the dataflow of the passed in data component.
+ */
+export function optimizeDataflow(data: DataComponent) {
+  for (let i = 0; i < 5; i++) {
```
Saba9: Having a constant like `MAX_OPTIMIZATION_RUNS` would be more readable than having the hard-coded 5.

src/compile/data/optimize.ts
invokesus: I had thought of doing that but `iterateFromLeaves` exits when it encounters a `SourceNode`. To implement your suggestion that behavior would have to be changed, so I took the less invasive way.

The check actually isn't necessary since it will encounter a `SourceNode` before getting an `undefined` parent and exit but I thought adding it wouldn't hurt.

domoritz:
Please rebase and address https://github.com/vega/vega-lite/pull/4097/files/b3f487f3aa508cd698aa7d22d82639004dee8507#diff-63289f48fb1db9affdddce376d826054. Thanks!

src/compile/data/optimize.ts
invokesus: Done.

src/compile/data/optimize.ts
invokesus: Done.

src/compile/data/optimize.ts
```diff
@@ -195,7 +202,7 @@ function optimizationDataflowHelper(dataComponent: DataComponent) {
  * Optimizes the dataflow of the passed in data component.
  */
 export function optimizeDataflow(data: DataComponent) {
-  for (let i = 0; i < 5; i++) {
+  for (let i = 0; i < MAX_OPTIMIZATION_RUNS; i++) {
```
domoritz: Let's add a warning if we reach the `MAX_OPTIMIZATION_RUNS`. I'd like to know when this happens.

domoritz:
So, what happens in this case:

```
1  2  3
|  |__|
|   |  
|___|
  |
```

In one pass the parse node from 1 is moved to the bottom fork and from 2 and 3 to the top fork. Then the top fork is merged but then the bottom fork is also merged and we may have a conflict, no? How do we prevent not merging the top first?

src/compile/data/optimize.ts
```diff
@@ -120,9 +121,13 @@ function getLeaves(roots: DataFlowNode[]) {
 /**
  * Inserts an Intermediate ParseNode containing all non-conflicting Parse fields and removes the empty ParseNodes
  */
-export function mergeParse(node: DataFlowNode): boolean {
-  let mutated = false;
-  const parseChildren = node.children.filter((x): x is ParseNode => x instanceof ParseNode);
+export function mergeParse(node: DataFlowNode): optimizers.OptimizerFlags {
+  let mutatedFlag = false;
+  const parent = node.parent;
+  if (parent === undefined) {
```
domoritz: Didn't you say this case never happens? Delete this then.

src/compile/data/optimize.ts
```diff
@@ -199,20 +207,26 @@ function optimizationDataflowHelper(dataComponent: DataComponent) {
...
 
   // move facets down and make a copy of the subtree so that we can have scales at the top level
   vals(data.sources).map(moveFacetDown);
 
-  for (let i = 0; i < 5; i++) {
+  for (let i = 0; i < MAX_OPTIMIZATION_RUNS; i++) {
     if (!optimizationDataflowHelper(data)) {
       break;
     }
+    if (warningFlag || i === MAX_OPTIMIZATION_RUNS - 1) {
```
domoritz: This is weird. Let's just add one counter variable and don't initialize it in the for loop. Okay?

test/compile/data/optimize.test.ts
```diff
@@ -35,4 +36,58 @@ describe('compile/data/optimize', () => {
       assert.deepEqual(children[1].parse, {a: 'boolean'});
     });
   });
+  describe('optimizeDataFlow', () => {
+    it('should push up common parse', () => {
```
domoritz: push -> move

src/compile/data/optimize.ts
```diff
@@ -224,11 +222,13 @@ export function optimizeDataflow(data: DataComponent) {
...
-    if (warningFlag || i === MAX_OPTIMIZATION_RUNS - 1) {
-      log.warn(`Maximum optimization runs(${MAX_OPTIMIZATION_RUNS}) reached.`);
-    }
+    secondPassCounter++;
   }
 
   // check after optimizations
   checkLinks(vals(data.sources));
+
+  if (Math.max(firstPassCounter, secondPassCounter) === MAX_OPTIMIZATION_RUNS) {
```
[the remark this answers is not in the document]
invokesus: The counters are incremented only if things are changed otherwise it breaks out of the loop.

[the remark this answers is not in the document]
domoritz: Now i see it.

PR #4119, opened 2018-08-05
Refactor optimizers to classes

invokesus:
Merge after #4097, #4118 are merged and #3832 is resolved.

### DIFF
--- a/src/compile/data/optimize.ts
+++ b/src/compile/data/optimize.ts
@@ -128,7 +128,7 @@ function getLeaves(roots: DataFlowNode[]) {
  * Inserts an Intermediate ParseNode containing all non-conflicting Parse fields and removes the empty ParseNodes
  */
 export class MergeParse extends BottomUpOptimizer {
-  public optimize(node: DataFlowNode): optimizers.OptimizerFlags {
+  public run(node: DataFlowNode): optimizers.OptimizerFlags {
     const parent = node.parent;
     const parseChildren = parent.children.filter((x): x is ParseNode => x instanceof ParseNode);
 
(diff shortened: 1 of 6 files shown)


src/compile/data/optimize.ts
```diff
@@ -121,76 +127,80 @@ function getLeaves(roots: DataFlowNode[]) {
...
+    return this.flags;
   }
-  return {continueFlag: true, mutatedFlag};
 }
 
 export function isTrue(x: boolean) {
   return x;
 }
 
+// any type to avoid error of Cannot use 'new' with an expression whose type lacks a call or construct signature.
```
domoritz: I don't get this comment

domoritz: Wouldn't it be better if this was a method of the class? Something like

const opt = new RemoveNodeOptimizer(...).
Opt.run()

domoritz: It's a bit uglier, though. 

Before you do this, let's talk a bit more high level. What's the advantage of using classes? I feel one is type safety and with the any you loose that, no?

domoritz: To make code more readable, the first argument should be the class!

src/compile/data/optimizer.ts
```diff
@@ -0,0 +1,69 @@
+import {DataFlowNode} from './dataflow';
+import {OptimizerFlags} from './optimizers';
+import {SourceNode} from './source';
+
+abstract class OptimizerBase {
```
domoritz: All of these classes in this file need comments.

test/compile/data/optimizers.test.ts
```diff
@@ -1,7 +1,7 @@
 /* tslint:disable:quotemark */
 import {DataFlowNode} from '../../../src/compile/data/dataflow';
 import {ImputeNode} from '../../../src/compile/data/impute';
-import {mergeIdenticalNodes, mergeNodes} from '../../../src/compile/data/optimizers';
```
domoritz: In the tests, let's check the flags as well.

PR #4125, opened 2018-08-09
Added Image Pixel Render Example

amitkaps:


domoritz:
Do you think it's better if we set the padding to 0 with `scale: {padding: 0}`? 

Maybe you could add a range input (slider) to change the padding but set it to 0 by default? 

screen shot 2018-08-09 at 09 07 14

amitkaps:
Makes sense. Removed the padding. I don't think we need a slider.

For the example in the docs, will replicate the same but with a smaller inline image array.

PR #4126, opened 2018-08-10
Add tests for optimizeDataFlow

Saba9:
Added a couple of tests for optimizeDataFlow. One simply checks to see if a common parse node is being generated in a simple case where to adjacent ParseNodes have one common parse and one conflicting parse. This test passes. 

The second case showcases what I believe to be a bug with the current implementation of moveParse. This test demonstrates how in some cases a common parse node is added above two adjacent ParseNodes even though this parse node conflicts with another ParseNode that exists at a lower level of the DataFlow tree.

I believe the solution to this involves changing the order in which we move up ParseNodes such that the nodes at lower levels are moved up first before merging nodes that exist further up in the tree.

### DIFF
--- a/test/compile/data/optimize.test.ts
+++ b/test/compile/data/optimize.test.ts
@@ -67,7 +67,7 @@ describe('compile/data/optimize', () => {
     });
     it('should push parse up from lowest level first to avoid conflicting common parse', () => {
       const source = new SourceNode(null);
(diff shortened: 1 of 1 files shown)


domoritz:
@invokesus Please look at this when you get the chance. The parse move logic should move nodes from all branches to the fork before attempting to merge at that fork.

Issue #4135, opened 2018-08-17
Example missing for ranged area


kanitw:
![image](https://user-images.githubusercontent.com/111269/44291550-e3949f00-a233-11e8-968d-7805ee91ab62.png)

domoritz:
It renders for me :-/

kanitw:
@chanwutk this might be related to some old example that got renamed?

chanwutk:
Now, the example is not explicitly using `x2` and `y2` anymore. Should I create another example for this or should I just use the normalized version of the `errorband`?

domoritz:
An explicit example would be good. For example minimum and maximum temperature would be an easy one.

kanitw:
If the normalized version is clear (and explicit), then it's fine too.

chanwutk:
Should we close this issue?

PR #4136, opened 2018-08-19
Fix missing area example

chanwutk:
Change the example to normalized ci error band.
issue #4135


domoritz:
I think it would be nice to have a chart that shows the minimum and maximum temperature over time. Ci is another thing users have to learn to understand the example.

chanwutk:
Would that be the data from this [Weekly Temperature Example](https://vega.github.io/vega/examples/weekly-temperature/)?

domoritz:
Love it. Thanks.

PR #4137, opened 2018-08-19
Add only a single "facet" signal per FacetModel (fixes #3688).

### LINKED ISSUE #3688: Facet with multiple selections fails
The following spec fails, and I'm not certain why. If one of the selections is removed, then it works:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.4.1.json",
  "data": {
    "url": "https://vega.github.io/vega-datasets/data/iris.json"
  },
  "facet": {
    "column": {
      "field": "species",
      "type": "nominal"
    }
  },
  "spec": {
    "layer": [
      {
        "encoding": {
          "color": {
            "condition": {
              "field": "species",
              "selection": "selection1",
              "type": "nominal"
            },
            "value": "lightgray"
          },
          "x": {
            "field": "petalLength",
            "type": "quantitative"
          },
          "y": {
            "field": "petalWidth",
            "type": "quantitative"
          }
        },
        "height": 200,
        "mark": "point",
        "selection": {
          "selection1": {
            "empty": "none",
            "nearest": true,
            "on": "mouseover",
            "type": "single"
          }
        },
        "width": 200
      },
      {
        "encoding": {
          "opacity": {
            "condition": {
              "selection": "selection2",
              "value": 1
            },
            "value": 0
          },
          "text": {
            "field": "species",
            "type": "nominal"
          },
          "x": {
            "field": "petalLength",
            "type": "quantitative"
          },
          "y": {
            "field": "petalWidth",
            "type": "quantitative"
          }
        },
        "height": 200,
        "mark": "text",
        "selection": {
          "selection2": {
            "empty": "none",
            "nearest": true,
            "on": "mouseover",
            "type": "single"
          }
        },
        "width": 200
      }
    ]
  }
}
```
reported in https://github.com/altair-viz/altair/issues/798

arvind:
Add only a single "facet" signal per FacetModel (fixes #3688).

### DIFF
--- a/src/compile/selection/selection.ts
+++ b/src/compile/selection/selection.ts
@@ -5,7 +5,7 @@ import {Channel, ScaleChannel, X, Y} from '../../channel';
 import {warn} from '../../log';
 import {LogicalOperand} from '../../logical';
 import {BrushConfig, SELECTION_ID, SelectionDef, SelectionResolution, SelectionType} from '../../selection';
-import {accessPathWithDatum, Dict, logicalExpr, varName} from '../../util';
+import {accessPathWithDatum, Dict, keys, logicalExpr, varName} from '../../util';
 import {VgBinding, VgData, VgEventStream} from '../../vega.schema';
 import {DataFlowNode} from '../data/dataflow';
 import {TimeUnitNode} from '../data/timeunit';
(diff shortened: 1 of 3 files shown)


arvind:
Caught+fixed a bug, and compile-time and runtime tests now both pass!

src/compile/selection/selection.ts
```diff
@@ -137,9 +137,12 @@ export function assembleUnitSelectionSignals(model: UnitModel, signals: any[]) {
     });
   });
 
-  const facetModel = getFacetModel(model);
-  if (signals.length && facetModel) {
-    const name = stringValue(facetModel.getName('cell'));
+  return signals;
+}
+
+export function assembleFacetSignals(model: FacetModel, signals: any[]) {
```
domoritz: Not necessary for this PR, but now that we have typings for Vega signals, we should try to replace the `any`s. https://github.com/vega/vega-typings/blob/master/types/spec/signal.d.ts

arvind: Agreed! I've replaced a bunch of them in my refactoring PR (#4068, there may be more though). I branched this one off master to keep it unblocked.

Issue #4140, opened 2018-08-22
boxplot with min-max extent overrides transform


meetamit:
Hi there. It appears that using a boxplot with `extent: 'min-max'` causes any existing `transform`s to be ignored (or overwritten or perhaps processed later than the boxplot's transforms). However, if `extent` is a number, `transform` is *correctly* processed.

To demonstrate, in the following spec we `calculate` variable `b2` and use it as `field` in a 1D boxplot with `extent: min-max`. This does **not** render correctly (Infinite domain). However, changing to `extent: 100000`, which (for this dataset) produces the same result as `min-max`, calculates and renders properly. 

````json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v3.0.0-rc3.json",
  "data": {
    "values": [
      {"a": "A", "b": 28},
      {"a": "B", "b": 55},
      {"a": "C", "b": 43},
      {"a": "G", "b": 19},
      {"a": "H", "b": 87},
      {"a": "I", "b": 52},
      {"a": "D", "b": 91},
      {"a": "E", "b": 81},
      {"a": "F", "b": 53}
    ]
  },
  "transform": [
    {"calculate": "2*datum.b", "as": "b2"}
  ],
  "mark": {
    "type": "boxplot",
    "extent": "min-max"
  }, 
  "encoding": {
    "y": {"field": "b2", "type": "quantitative"}
  }
}
````
Thanks!

domoritz:
Thank you for reporting the bug. It looks like a bug in the normalizer code. We will look into it.

domoritz:
Fixed in #4141

PR #4141, opened 2018-08-22
Fix overwritten transform in composite mark

chanwutk:
combine transform from outer spec with transform from composite mark before return to prevent overwriting.
issue #4140

### DIFF
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -186,7 +186,7 @@ export function normalizeBoxPlot(
   }
   return {
     ...outerSpec,
-    transform,
+    transform: (outerSpec.transform || []).concat(transform),
     layer: boxLayer
   };
 }
--- a/src/compositemark/errorbar.ts
+++ b/src/compositemark/errorbar.ts
@@ -211,6 +211,7 @@ export function errorBarParams<
 
   return {
     transform: [
+      ...(outerSpec.transform || []),
       ...bins,
       ...timeUnits,
       ...(!aggregate.length ? [] : [{aggregate, groupby}]),
--- a/test/compositemark/boxplot.test.ts
+++ b/test/compositemark/boxplot.test.ts
@@ -3,6 +3,7 @@ import {assert} from 'chai';
 
 import * as log from '../../src/log';
 import {normalize} from '../../src/spec';
+import {Transform} from '../../src/transform';
 import {defaultConfig} from '.././../src/config';
 
 describe('normalizeBoxMinMax', () => {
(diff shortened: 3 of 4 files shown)


src/compositemark/errorbar.ts
```diff
@@ -211,6 +211,7 @@ export function errorBarParams<
 
   return {
     transform: [
+      ...(outerSpec.transform ? outerSpec.transform : []),
```
domoritz: `outerSpec.transform || []`?

src/compositemark/boxplot.ts
```diff
@@ -186,7 +186,7 @@ export function normalizeBoxPlot(
   }
   return {
     ...outerSpec,
-    transform,
+    transform: [...(outerSpec.transform ? outerSpec.transform : []), ...transform],
```
domoritz: `[...(outerSpec.transform || []), ...transform]`

src/compositemark/boxplot.ts
```diff
@@ -186,7 +186,7 @@ export function normalizeBoxPlot(
   }
   return {
     ...outerSpec,
-    transform,
+    transform: [...(outerSpec.transform || []), ...transform],
```
domoritz: Hmm, maybe it's cleaner to write `(outerSpec.transform || []).concat(transform)` and use concat below as well.

chanwutk: What do you mean below?

domoritz: The other line you changed

chanwutk: transform in error bar has more than 1 elements to be combined. Would it be better to use this instead of `concat` in errorbar.ts?

domoritz: You're right. It's good as it is.

test/compositemark/boxplot.test.ts
```diff
@@ -1555,6 +1556,29 @@ describe('normalizeBoxMinMax', () => {
...
+          extent: 'min-max'
+        },
+        transform: [{calculate: 'age * 2', as: 'age2'}],
+        encoding: {x: {field: 'age', type: 'ordinal'}, y: {field: 'people', type: 'quantitative', title: 'population'}}
+      },
+      defaultConfig
+    );
+
+    const transforms: Transform[] = outputSpec.transform;
+    if (transforms.length > 0) {
```
domoritz: Why add an if...else? Can you just add ad expect(transform).to.haveLengthAtLeast(1) or something?

domoritz: A failure in one assert should abort the test case.

chanwutk: Ok, that seems a lot cleaner.

domoritz: We want to eventually change all assert statements but there is no point in doing that right now.

PR #4160, opened 2018-09-14
Change the intersection function to determine dependencies

### LINKED ISSUE #3832: Make parse for nested field come after the transform that produces it

invokesus:
Fixes #3832

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -185,30 +185,21 @@ export function hasIntersection(a: StringSet, b: StringSet) {
   return false;
 }
 
-export function fieldIntersection(a: StringSet, b: StringSet): boolean {
+export function prefixGenerator(a: StringSet): StringSet {
+  const prefixes = {};
   // for ... of used to avoid for ... in must be filtered linter warning
   for (const x of Object.keys(a)) {
-    for (const y of Object.keys(b)) {
-      if (stringIntersection(x, y)) {
-        return true;
-      }
-    }
+    const splitField = splitAccessPath(x);
+    // Wrap every element other than the first in `[]`
+    const wrappedWithAccessors = splitField.map((y, i) => (i === 0 ? y : `[${y}]`));
+    const computedPrefixes = wrappedWithAccessors.map((_, i) => wrappedWithAccessors.slice(0, i + 1).join(''));
+    computedPrefixes.forEach(y => (prefixes[y] = true));
   }
-  return false;
+  return prefixes;
 }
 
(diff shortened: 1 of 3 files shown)


domoritz:
@saba9 can you help review this?

domoritz:
Thank you. Let's make this function more efficient. 

Instead of doing n^2 many comparisons, let's iterate over the two lists and expand them into their prefixes. Then compare the sets for an intersection afterwards. 

Can you do this?

src/util.ts
```diff
@@ -185,6 +185,23 @@ export function hasIntersection(a: StringSet, b: StringSet) {
   return false;
 }
 
+export function prefixGenerator(a: StringSet): StringSet {
+  const prefixes = {};
+  // for ... of used to avoid for ... in must be filtered linter warning
+  for (const x of Object.keys(a)) {
```
domoritz: Use utils.keys and delete comment

test/util.test.ts
```diff
@@ -153,4 +153,28 @@ describe('util', () => {
       assert.equal(replacePathInField('foo\\.bar'), 'foo\\.bar');
     });
   });
+
+  describe('prefixGenerator', () => {
+    it('should return the correct value for simple nested field', () => {
+      assert.deepEqual(prefixGenerator({'a.b': true}), {a: true, 'a[b]': true});
```
domoritz: For these we should switch to jest's expects instead of chai's assert.

PR #4169, opened 2018-09-25
Set default title frame to group for unit and layered plots. Fixes #4159

### LINKED ISSUE #4159: Center titles by the chart area, not the whole chart
Right now: 

screen shot 2018-09-13 at 18 51 46

domoritz:


domoritz:
Rebased and ready for review.

domoritz:
Thanks. I just want to check whether @kanitw also thinks that this is a good change.

domoritz:
Rebased

kanitw:
Why is there no SVG changes at all?  Is it because you change the examples?  But I wanna understand why those examples shouldn't use the default?

domoritz:
Hmm, they should change. Maybe the script didn't work.

kanitw:
Rebased and  force-pushed -- let's see if this fixes the problems.

PR #4170, opened 2018-09-25
Update to v3 schema

domoritz:


domoritz:
Merged since other's are blocking on this. Please send me follow-up feedback and I can make follow-up PRs.

PR #4171, opened 2018-09-25
make CI commits come from travis user

domoritz:


domoritz:
nm, we are already doing this.

PR #4176, opened 2018-09-25
Produced and dependent field fixes

domoritz:
See individual commits.


domoritz:
@invokesus Do you think window and aggregates should have the group by fields in their produced fields?

invokesus:
Having the same field in both `producedFields` and `dependentFields` is a bit confusing, no?

domoritz:
I'm trying to think through what we use the dependent and produced field for. I think you're right. For example when we push a filter through aggregates, we should be able to do it. Okay, I'll revise.

domoritz:
Rebased onto master

PR #4178, opened 2018-09-26
Removing excess changes from boxplot iqr pr

mattwchun:
Removing excess changes from #2485. Tested before and after removing excess chunk and all tests pass.


domoritz:
Thank you! Let's see whether the examples change.

PR #4185, opened 2018-09-29
Don’t hash datasets

### LINKED ISSUE #4146: Caching a Dataset for Multiple Specs
I have a use case that a user can create different visualizations with a single dataset in a trial-and-error fashion. I tried inlining the parsed dataset into the specs but it didn't seem to help. From the Chrome profiling tool, Vega-lite always "parse" the dataset even if it's inline javascript objects. Relevant screenshots are provided below.

This report is created according to the stackoverflow thread: [https://stackoverflow.com/questions/51455030/cache-a-dataset-in-vega-or-vega-embed](https://stackoverflow.com/questions/51455030/cache-a-dataset-in-vega-or-vega-embed)

The spec as shown by `console.log`, which shows the data is indeed inline.
vega_lite_spec_inline_data

The screenshot of Chrome's profiling tool shows that the `parse` function takes most of the time.
vega_lite_stack

As this is my first time creating a bug report, I hope I am not breaking any rules. Let me know if I can provide more materials.

domoritz:
Fixes #4146.

### DIFF
--- a/src/compile/data/index.ts
+++ b/src/compile/data/index.ts
@@ -7,9 +7,9 @@ import {SourceNode} from './source';
 
 export interface DataComponent {
   /**
-   * A dictionary of sources indexed by a hash.
+   * A list on unique sources.
    */
-  sources: Dict<SourceNode>;
+  sources: SourceNode[];
 
   /**
    * Registry of output nodes.
--- a/src/compile/data/index.ts
+++ b/src/compile/data/index.ts
@@ -7,7 +7,7 @@ import {SourceNode} from './source';
 
 export interface DataComponent {
   /**
-   * A list on unique sources.
+   * A list of unique sources.
    */
   sources: SourceNode[];
 
--- a/src/compile/data/optimize.ts
+++ b/src/compile/data/optimize.ts
@@ -1,6 +1,6 @@
 import {MAIN} from '../../data';
 import * as log from '../../log';
-import {flatten, keys, vals} from '../../util';
+import {flatten, keys} from '../../util';
 import {AggregateNode} from './aggregate';
 import {DataFlowNode, OutputNode} from './dataflow';
 import {checkLinks} from './debug';
@@ -8,9 +8,8 @@ import {FacetNode} from './facet';
 import {ParseNode} from './formatparse';
 import {DataComponent} from './index';
 import {BottomUpOptimizer, TopDownOptimizer} from './optimizer';
-import {MergeIdenticalNodes} from './optimizers';
 import * as optimizers from './optimizers';
-import {SourceNode} from './source';
+import {MergeIdenticalNodes} from './optimizers';
 import {StackNode} from './stack';
 import {WindowTransformNode} from './window';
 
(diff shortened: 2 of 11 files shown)


src/compile/data/index.ts
```diff
@@ -7,9 +7,9 @@ import {SourceNode} from './source';
 
 export interface DataComponent {
   /**
-   * A dictionary of sources indexed by a hash.
+   * A list on unique sources.
```
invokesus: *of

src/compile/data/parse.ts
```diff
@@ -39,17 +39,41 @@ import {TimeUnitNode} from './timeunit';
...
+export function findSource(data: Data, sources: SourceNode[]) {
+  for (const other of sources) {
+    const otherData = other.data;
+    if (isInlineData(data) && isInlineData(otherData)) {
+      const srcVals = data.values;
+      const otherVals = otherData.values;
+      if (deepEqual(srcVals, otherVals)) {
+        return other;
+      }
+    } else if (isUrlData(data) && isUrlData(otherData)) {
```
invokesus: How about `else if (isUrlData(data) && isUrlData(otherData) && data.url === otherData.url) {` ?

src/compile/data/optimize.ts
```diff
@@ -159,8 +159,9 @@ export function isTrue(x: boolean) {
 }
 
 function optimizationDataflowHelper(dataComponent: DataComponent) {
-  let roots: SourceNode[] = vals(dataComponent.sources);
+  let roots: SourceNode[] = dataComponent.sources;
```
invokesus: This is a copy by reference.

src/compile/data/optimize.ts
```diff
@@ -185,11 +186,9 @@ function optimizationDataflowHelper(dataComponent: DataComponent) {
 
   mutatedFlag = roots.map(mergeParse).some(isTrue) || mutatedFlag;
   mutatedFlag = roots.map(optimizers.mergeIdenticalNodes).some(isTrue) || mutatedFlag;
-  keys(dataComponent.sources).forEach(s => {
-    if (dataComponent.sources[s].numChildren() === 0) {
-      delete dataComponent.sources[s];
-    }
-  });
+
+  dataComponent.sources = roots;
```
invokesus: Both are the same array.

src/compile/data/parse.ts
invokesus: Oh, the type narrowing won't work with this I guess.

src/compile/data/optimize.ts
domoritz: No, we use filter to remove things from the array and that is not in place.

src/compile/data/optimize.ts
domoritz: See comment below

domoritz:
Rebased

src/compile/data/index.ts
domoritz: Fixed

Issue #4164, opened 2018-09-17
Boxplot: temporal -> ordinal axis labels do not format correctly


ijlyttle:
The formatting for the labels does not seem to work - I have concatted a chart using points for comparison.

![image](https://user-images.githubusercontent.com/2800934/45605017-60679400-b9ff-11e8-9f36-147210d376d5.png)

Please see https://bl.ocks.org/ijlyttle/8cce9bf5458d4e10b4de1e828e4eade1 for a live example.

As well - I the boxplots themselves don't seem correct, and there are no outliers plotted. Please advise if you would like me to file separate issues for these.

Per usual, I keep open the possibility I am doing something wrong...

---

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v3.json",
  "data": {
    "url": "https://vega.github.io/vega-datasets/data/seattle-weather.csv",
    "format": {
      "type": "csv"
    }
  },
  "hconcat": [
    {
      "mark": {
        "type": "boxplot",
        "extent": 1.5
      },
      "encoding": {
        "x": {
          "field": "date",
          "timeUnit": "month",
          "type": "ordinal",
          "axis": {
            "format": "%b"
          }
        },
        "y": {
          "field": "temp_max",
          "type": "quantitative"
        }
      }
    },
    {
      "mark": {
        "type": "point"
      },
      "encoding": {
        "x": {
          "field": "date",
          "timeUnit": "month",
          "type": "ordinal",
          "axis": {
            "format": "%b"
          }
        },
        "y": {
          "field": "temp_max",
          "type": "quantitative"
        }
      }
    }
  ]
}
```

domoritz:
Thank you for reporting the issue. Note that boxplots are part of the release candidate so there are definitely a few bugs to iron out. We will fix this before the final release.

artfulgarfunk:
I would just like to add that Boxplot also does not allow axis title to be set on the ordinal axis. 
![screen shot 2018-09-21 at 11 36 25](https://user-images.githubusercontent.com/22254546/45876480-a7081600-bd92-11e8-826d-bd86fe99db10.png)

domoritz:
Thank you! We're looking into this issue.

domoritz:
@artfulgarfunk Your issue has been fixed in master. We are working on also supporting formatting when there is no time unit.

PR #4190, opened 2018-09-30
add support to axis' time format in composite mark

chanwutk:
from issue #4164, format axis when `timeUnit` is specified in encoding channels.

### DIFF
--- a/examples/specs/normalized/layer_line_errorband_pre_aggregated_normalized.vl.json
+++ b/examples/specs/normalized/layer_line_errorband_pre_aggregated_normalized.vl.json
@@ -30,7 +30,7 @@
               "title": "Year (year)",
               "type": "temporal",
               "field": "year_Year",
-              "axis": {"format": "%Y"}
+              "axis": {"format": "%Y", "formatType": "time"}
             },
             "tooltip": [
               {"field": "upper_ci1", "type": "quantitative", "title": "ci0"},
@@ -39,7 +39,7 @@
                 "title": "Year (year)",
                 "type": "temporal",
                 "field": "year_Year",
-                "axis": {"format": "%Y"}
+                "axis": {"format": "%Y", "formatType": "time"}
               }
             ]
           }
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -101,6 +101,11 @@ export interface Axis extends AxisOrientMixins, BaseAxisNoSignals, Guide {
   tickMinStep?: number;
 
   /**
+   * A type of axis format to be forced to.
+   */
+  formatType?: 'number' | 'time';
+
+  /**
    * Explicitly set the visible axis tick values.
    */
   values?: number[] | string[] | boolean[] | DateTime[];
@@ -303,7 +308,8 @@ const COMMON_AXIS_PROPERTIES_INDEX: Flag<keyof (VgAxis | Axis)> = {
 
 const AXIS_PROPERTIES_INDEX: Flag<keyof Axis> = {
   ...COMMON_AXIS_PROPERTIES_INDEX,
-  encoding: 1
+  encoding: 1,
+  formatType: 1
 };
 
 const VG_AXIS_PROPERTIES_INDEX: Flag<keyof VgAxis> = {
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -101,11 +101,6 @@ export interface Axis extends AxisOrientMixins, BaseAxisNoSignals, Guide {
   tickMinStep?: number;
 
   /**
-   * A type of axis format to be forced to.
-   */
-  formatType?: 'number' | 'time';
-
-  /**
    * Explicitly set the visible axis tick values.
    */
   values?: number[] | string[] | boolean[] | DateTime[];
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -2,8 +2,10 @@ import {isArray} from 'vega-util';
 import {isBinning} from '../bin';
 import {Config, StyleConfigIndex} from '../config';
 import {
+  FieldDef,
   FieldDefBase,
   FieldRefOption,
+  isPositionFieldDef,
   isScaleFieldDef,
   isTimeFieldDef,
   OrderFieldDef,
@@ -87,11 +89,13 @@ export function formatSignalRef(
     return {
       signal: `${formatExpr(vgField(fieldDef, {expr, binSuffix: 'range'}), format)}`
     };
-  } else if (isTimeFieldDef(fieldDef)) {
+  } else if (isTimeFieldDef(fieldDef) || getFormatType(fieldDef) === 'time') {
     const isUTCScale = isScaleFieldDef(fieldDef) && fieldDef['scale'] && fieldDef['scale'].type === ScaleType.UTC;
     return {
       signal: timeFormatExpression(
-        vgField(fieldDef, {expr}),
+        vgField(fieldDef, {
+          expr
+        }),
         fieldDef.timeUnit,
         specifiedFormat,
         config.text.shortTimeLabels,
@@ -101,9 +105,7 @@ export function formatSignalRef(
       )
     };
   } else {
-    return {
-      signal: `''+${vgField(fieldDef, {expr})}`
-    };
+    return {signal: `''+${vgField(fieldDef, {expr})}`};
   }
 }
 
@@ -111,7 +113,7 @@ export function formatSignalRef(
  * Returns number format for a fieldDef
  */
 export function numberFormat(fieldDef: TypedFieldDef<string>, specifiedFormat: string, config: Config) {
-  if (isTimeFieldDef(fieldDef)) {
+  if (isTimeFieldDef(fieldDef) || getFormatType(fieldDef) === 'time') {
     return undefined;
   }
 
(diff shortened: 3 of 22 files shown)


examples/specs/normalized/layer_line_errorband_pre_aggregated_normalized.vl.json
```diff
@@ -29,7 +29,8 @@
             "x": {
               "field": "year_Year",
               "type": "temporal",
-              "title": "Year (year)"
+              "title": "Year (year)",
+              "timeUnit": "year"
```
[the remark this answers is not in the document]
chanwutk: We cannot only have `timeUnit` in the encoding because `timeUnit` is required for `groupby` in aggregate for parts of composite mark like this:
```json
"transform": [
  {
    "timeUnit": "month",
    "field": "date",
    "as": "month_date"
  },
  {
    "aggregate": [
      {
        "op": "max",
        "field": "temp_max",
        "as": "max_temp_max"
      }
    ],
    "groupby": [
      "month_date"
    ]
  }
]
```

So, I think our only choice is to format the field for different type of timeUnit?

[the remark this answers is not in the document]
domoritz: Right. But you can still specify a format. Is the problem that for ordinal we don't have a way to specity time formats?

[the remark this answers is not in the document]
chanwutk: Yes, I think we can have another calculate transform to format the field, then use the new field instead?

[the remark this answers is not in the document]
domoritz: I don't understand the suggestion. Can you explain it?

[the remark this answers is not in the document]
chanwutk: In the normalized spec, instead of only transform `timeUnit` like this:
```json
"transform": [
  {
    "timeUnit": "month",
    "field": "date",
    "as": "month_date"
  },
  {
    "aggregate": [
      {
        "op": "max",
        "field": "temp_max",
        "as": "max_temp_max"
      }
    ],
    "groupby": [
      "month_date"
    ]
  }
]
```
should we format them first like this?
```json
"transform": [
  {
    "timeUnit": "month",
    "field": "date",
    "as": "tmp_month_date"
  },
  {
    "calculate": "['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'][month(datum.tmp_month_date)]",
    "as": "month_date"
  },
  {
    "aggregate": [
      {
        "op": "max",
        "field": "temp_max",
        "as": "max_temp_max"
      }
    ],
    "groupby": [
      "month_date"
    ]
  }
]
```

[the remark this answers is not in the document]
domoritz: No, we should set the format on the endcoding. I think at least that is the easiest way forward. Cc @kanitw what do you think?

domoritz:
@chanwutk please rebase

examples/compiled/a_test.vg.json
```diff
@@ -0,0 +1,128 @@
+{
+  "$schema": "https://vega.github.io/schema/vega/v4.json",
+  "autosize": "pad",
+  "padding": 5,
+  "height": 200,
+  "data": [
+    {
+      "name": "source_0",
+      "url": "https://vega.github.io/vega-datasets/data/seattle-weather.csv",
+      "format": {"type": "csv"}
```
chanwutk: @domoritz @kanitw Is there a way to add `"parse": {"date": "date"}` to `format` when there is `timeFormat` specified in `axis` like in examples/specs/a_test.vl.json?

chanwutk: That's right.

kanitw:
Now I'm quite confused.  Are you adding support for "timeUnit" or "timeFormat" in this PR?

kanitw:
From  #4164

> We are working on also supporting formatting when there is no time unit.

It seems like the title of this PR is misleading?

kanitw:
> Should we add support to time format in legend in this PR?

Nope.  Better get it ready for merging.

kanitw:
Is it ready?

src/compositemark/common.ts
```diff
@@ -121,6 +127,34 @@ export function makeCompositeAggregatePartFactory<P extends PartsMixins<any>>(
   };
 }
 
+export function compositeMarkExtractTransformsFromEncoding(
```
kanitw: Why do we need a new method?

Why shouldn't this format replacement be down in `extractTransformsFromEncoding`?

src/compile/common.ts
```diff
@@ -101,17 +106,16 @@ export function formatSignalRef(
...
   }
 }
 
 /**
  * Returns number format for a fieldDef
  */
 export function numberFormat(fieldDef: TypedFieldDef<string>, specifiedFormat: string, config: Config) {
-  if (isTimeFieldDef(fieldDef)) {
+  const formatType = getFormatType(fieldDef);
+  if (formatType === 'time' || (!formatType && isTimeFieldDef(fieldDef))) {
```
kanitw: It feel wrong that a method named `numberFormat` can return a time format.  Perhaps, a bit refactor needs to be done here.

src/axis.ts
```diff
@@ -86,6 +86,11 @@ export interface Axis extends BaseAxisNoSignals, Guide {
    */
   tickStep?: number;
 
+  /**
+   * A type of axis format to be forced to.
+   */
+  formatType?: 'number' | 'time';
```
kanitw: Better add this to `Guide`, which both `Axis` and `Legend` extend from.  
Please also add it to `TextFieldDef`.

src/compile/common.ts
```diff
@@ -127,6 +131,10 @@ export function numberFormat(fieldDef: TypedFieldDef<string>, specifiedFormat: s
   return undefined;
 }
 
+function getFormatType(fieldDef: FieldDef<string>): 'number' | 'time' {
+  return isPositionFieldDef(fieldDef) && fieldDef.axis ? fieldDef.axis.formatType : undefined;
```
kanitw: What about format in `legend` and in text encoding?

I'd:

1) Check if the fieldDef is a TextFieldDef   (Add a type guard `isTextFieldDef()`, which checks for the containment of `format` or `formatType` )
2)  call the `getGuide()` method and read `formatType` from it.

src/encoding.ts
```diff
@@ -322,11 +322,11 @@ export function extractTransformsFromEncoding(oldEncoding: Encoding<Field>, conf
...
           const format = getDateTimeComponents(timeUnit, config.axis.shortTimeLabels).join(' ');
           if (isNonPositionScaleChannel(channel)) {
-            newChannelDef['legend'] = {format, ...newChannelDef['legend']};
+            newChannelDef['legend'] = {format, formatType: 'time', ...newChannelDef['legend']};
           } else if (isPositionChannel) {
-            newChannelDef['axis'] = {format, ...newChannelDef['axis']};
+            newChannelDef['axis'] = {format, formatType: 'time', ...newChannelDef['axis']};
           } else if (channel === 'text' || channel === 'tooltip') {
-            newChannelDef['format'] = newChannelDef['format'] || format;
+            newChannelDef['format'] = {...(newChannelDef['format'] || format), ...{fotmatType: 'time'}};
```
chanwutk: @kanit Would this else if statement be useful? When the channel is `text` or `tooltip`, wouldn't it go to the first if where `isNonPositionScaleChannel(channel)` is true?

src/encoding.ts
```diff
@@ -322,11 +322,9 @@ export function extractTransformsFromEncoding(oldEncoding: Encoding<Field>, conf
           // Add formatting to appropriate property based on the type of channel we're processing
           const format = getDateTimeComponents(timeUnit, config.axis.shortTimeLabels).join(' ');
           if (isNonPositionScaleChannel(channel)) {
```
domoritz: `isLegendScaleChannel` does not exist but we should use something equivalent here or move the check for text and tooltip up.

test/transformextract.test.ts
```diff
@@ -41,9 +40,8 @@ describe('extractTransforms()', () => {
     'interactive_layered_crossfilter.vl.json',
     'interactive_layered_crossfilter_discrete.vl.json',
     'interactive_seattle_weather.vl.json',
-    'layer_bar_dual_axis.vl.json',
     'layer_bar_dual_axis_minmax.vl.json',
-    'layer_bar_month.vl.json',
+    'layer_bar_month.vl.json', // data transform switchs the order
```
domoritz:
```suggestion
    'layer_bar_month.vl.json', // data transform switches the order
```

domoritz:
works

```
{
  "data": {"url": "data/stocks.csv", "format": {"parse": {"date": "date"}}},
  "mark": "line",
  "encoding": {
    "x": {"field": "date", "type": "quantitative", "axis": {"format": "%Y:%m", "formatType": "time"}},
  }
}
```

but this does not

```
{
  "mark": "line",
  "encoding": {
    "x": {"field": "date", "type": "quantitative", "axis": {"format": "%Y:%m", "formatType": "time"}},
  }
}
```

domoritz:
Please finish this and merge master.

domoritz:
I meant merge master into this, not merge this into master.

kanitw:
I've reverted the merge please create a different PR for this to redo.

PR #4191, opened 2018-09-30
Improve legend docs and example with legends

### LINKED ISSUE #3853: Revise default legend
1) It seems liked the default legend orient is changed when we [upgraded to Vega 4 rc](https://github.com/vega/vega-lite/pull/3743) -- not sure if this is simply due to Vega 4's changed default?

Since we didn't have a check to always re-compile examples until recently: https://github.com/vega/vega-lite/commit/cba6f359128792b90ca66fe04d86f9b36bfaa5d0. 
This change slipped through our releases. 

2) This leads to some better orient

image

But this also leads to so many examples that has very imbalanced layout.

image

image

I think horizontal is still a better default since it looks less worse in many general cases.   For example, even when the plot is tall enough but the legend's field has long title it still looks imbalanced.  

Plus, this will be more inline with previous behavior prior to 2.5.  

image

domoritz:
Fixes #3853


domoritz:
I decided to only update the one example that is bad.

domoritz:
ping @kanitw

kanitw:
FYI, I played a bit more with legend and think that vertical direction isn't a better default, especially when there are multiple legends or when the variable name is a bit long. 

I think default should minimize bad situation, and this current default isn't every ideal.

domoritz:
I think vertical is much better in most situations. It's also the default in matplotlib, ggplot, and plotly.

kanitw:
Let me clarify: vertical is good for discrete, but not for continuous.

> It's also the default in matplotlib, ggplot, and plotly.

1) AFAIK, ggplot2 doesn't have continuous legends.
2) matplotlib and plotly do *not* have variable title, so it doesn't look as awkward. 
3) Horizontal is the default in Tableau, which is the closest to our design (has title) -- so if we were to take an inspiration, Tableau is probably the right one.

kanitw:
Ok, im wrong about that then.  (Weird that [the manual](https://ggplot2.tidyverse.org/reference/guide_legend.html) doesn't show an example of it.) 

Still, all the examples they have in the page has short and unexplanatory variable names.

domoritz:
Let's discuss legend directions tomorrow at the meeting. I think vertical legends are better and we would be consistent with Vega and the symbol legends.

PR #4192, opened 2018-10-01
Set up CI with Azure Pipelines

domoritz:


domoritz:
It looks like builds are a bit faster on Azure but the configuration is a bit of a nightmare. See https://dev.azure.com/vega-vis/vega-lite/_build for details of the pipeline.

Problems so far
* No build caches.
* Integration with Github not as great as Travis. For example, emails are sent to everyone, I can't add Github users to azure.
* The config file is a mess.

domoritz:
Azure builds take 6 minutes. Travis more like 15.

domoritz:
Maybe @chrisrpatterson or others have suggestions for how we can improve the integration. For example, I see that Typescript uses separate definitions for different node versions (https://typescript.visualstudio.com/TypeScript/_build?definitionId=7) but I'm not sure how they set things up.

domoritz:
Not worth the effort to switch from Travis for now.

PR #4194, opened 2018-10-02
Setup Azure Pipelines

chrisrpatterson:
Basic Azure Pipelines configuration for multiple versions of node.  
Missing a couple of scripts because I do not fully understand what they do.
Included test publishing with Junit reporter.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,18 @@
+# Node.js
+# Build a general Node.js project with npm.
+# Add steps that analyze code, save build artifacts, deploy, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/languages/javascript
+
+pool:
+  vmImage: 'Ubuntu 16.04'
+
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '8.x'
+  displayName: 'Install Node.js'
+
+- script: |
+    npm install
+    npm run build
+  displayName: 'npm install and build'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,8 +32,6 @@ steps:
     yarn codecov
     yarn jest examples/
     yarn test:runtime
-#    ./scripts/check-schema.sh
-#    ./scripts/check-and-fix.sh
     yarn lint
   displayName: 'Build, Test and Lint'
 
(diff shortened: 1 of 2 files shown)


domoritz:
Oh wow, thank you! I'll take a look when the tests pass.

domoritz:
The scripts 

```
./scripts/check-schema.sh
./scripts/check-and-fix.sh
```

are used to check whether generated files are up to date and if not, will automatically commit them. They currently expect to run on Travis (they depend on travis variables) ([for example](https://github.com/vega/vega-lite/blob/master/scripts/check-and-fix.sh)).

I appreciate the PR but after revising the travis script based on some things I learned while trying azure script, I was able to reduce the build times significantly and that was the main reason I looked into azure pipelines. The PR with the updates is at https://github.com/vega/vega-lite/pull/4193. 

The main reasons we will probably stick with Travis are that we already have all the scripts up and running and Travis uses package caching, which has some performance benefits. 

Having said that, after playing with azure pipelines today, I really liked it and will use it when I start my next project from scratch. The startup time is great!

chrisrpatterson:
No problem, I saw the earlier attempt.  This is a bit cleaner but could still be better.  Might want to break up the scripts to provide more focused logs.  
I wasn't sure how to change these two scripts so I have not included them yet
```
  - ./scripts/check-schema.sh
  - ./scripts/check-and-fix.sh
```
Also right now the two legs of the matrix do not run in parallel on your scale unit but a fix for that is coming soon.

chrisrpatterson:
ok, if you have any questions feel free to ping me.

domoritz:
> Also right now the two legs of the matrix do not run in parallel on your scale unit but a fix for that is coming soon.

Yeah, I noticed that as well. I was able to fix it by setting `maxParallel: 2` but I guess it was a hack and I'm glad to hear that a fix is coming soon.

domoritz:
@chrisrpatterson I'm setting up azure pipelines for another repo in https://github.com/vega/ts-json-schema-generator/pull/48 and I would like to automatically publish tagged commits to NPM and only from node 10. 

On travis I can just use `on`. I looked into conditions but couldn't find anything about filtering tagged commits. 

```
language: node_js
node_js:
- '6'
- '10'
script:
- yarn lint
- yarn test
- yarn build
cache: yarn
deploy:
  provider: npm
  skip_cleanup: true
  email: domoritz@gmail.com
  api_key:
    secure: CODE
  on:
    tags: true
    repo: vega/ts-json-schema-generator
    node: '10'
```

How would you extend the npm publish task?

```
- task: Npm@1
  inputs:
    command: publish
    publishEndpoint: 'npmjs'
```

PR #4197, opened 2018-10-02
Add new Saite and Hanami entries.

jsa-aerial:
Add Saite exploration system under "Bindings ... " heading

Add new "Frameworks ..." heading and place Hanami under that

### DIFF
--- a/site/usage/applications.md
+++ b/site/usage/applications.md
@@ -42,7 +42,7 @@ We mark featured plugins and tools with a <span class="octicon octicon-star"></s
 
 ## Frameworks and Libraries utilizing Vega-Lite
 
--   [Hanami]((https://github.com/jsa-aerial/hanami)  A Clojure(Script) library and framework for creating domain and task specific interactive visualization applications using Vega-Lite and Vega. Built on [reagent](http://reagent-project.github.io/) (react) and [re-com](https://github.com/Day8/re-com) enabled.
+-   [Hanami](https://github.com/jsa-aerial/hanami)  A Clojure(Script) library and framework for creating domain and task specific interactive visualization applications using Vega-Lite and Vega. Built on [reagent](http://reagent-project.github.io/) (react) and [re-com](https://github.com/Day8/re-com) enabled.
 
 
 ## Programming / Data Science Environment that supports Vega-Lite
--- a/site/usage/applications.md
+++ b/site/usage/applications.md
@@ -23,6 +23,7 @@ We mark featured plugins and tools with a <span class="octicon octicon-star"></s
 -   <span class="octicon octicon-star"></span> [Tooltips for Vega and Vega-Lite](https://github.com/vega/vega-lite-tooltip)
 -   <span class="octicon octicon-star"></span> [Leaflet Tile Map integration for Vega and Vega-Lite](https://github.com/nyurik/leaflet-vega)
 
+
 ## Bindings for Programming Languages
 
 -   <span class="octicon octicon-star"></span> [Altair](https://altair-viz.github.io) exposes a Python API for building statistical visualizations that follows Vega-Lite syntax.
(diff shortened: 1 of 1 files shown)


site/usage/applications.md
```diff
@@ -35,9 +36,15 @@ We mark featured plugins and tools with a <span class="octicon octicon-star"></s
 -   [Vegas](https://github.com/aishfenton/Vegas) brings visualizations to Scala and Spark using Vega-Lite.
 -   [Vizard](https://github.com/yieldbot/vizard) tiny REPL client to visualize Clojure data in browser w/ Vega-Lite.
 -   [oz](https://github.com/metasoarous/oz) Vega & Vega-Lite based visualization toolkit for Clojure & ClojureScript (fork of Vizard).
+-   [Saite](https://github.com/jsa-aerial/saite) Exploratory graphics and ad hoc visualization system. Built in Clojure(Script) on top of [Hanami](https://github.com/jsa-aerial/hanami).
 -   [to-vega](https://github.com/gjmcn/to-vega) is a simple JavaScript library for creating Vega-Lite specifications.
 
 
+## Frameworks and Libraries utilizing Vega-Lite
+
+-   [Hanami](https://github.com/jsa-aerial/hanami)  A Clojure(Script) library and framework for creating domain and task specific interactive visualization applications using Vega-Lite and Vega. Built with [reagent](http://reagent-project.github.io/) (react) and [re-com](https://github.com/Day8/re-com) enabled.
```
domoritz: Do you think this may fit under "Tools for Embedding Vega-Lite Visualizations"? If not, please put this under "Tools that use Vega-Lite". I don't see that this needs a separate heading.

site/usage/applications.md
```diff
@@ -35,9 +36,15 @@ We mark featured plugins and tools with a <span class="octicon octicon-star"></s
 -   [Vegas](https://github.com/aishfenton/Vegas) brings visualizations to Scala and Spark using Vega-Lite.
 -   [Vizard](https://github.com/yieldbot/vizard) tiny REPL client to visualize Clojure data in browser w/ Vega-Lite.
 -   [oz](https://github.com/metasoarous/oz) Vega & Vega-Lite based visualization toolkit for Clojure & ClojureScript (fork of Vizard).
+-   [Saite](https://github.com/jsa-aerial/saite) Exploratory graphics and ad hoc visualization system. Built in Clojure(Script) on top of [Hanami](https://github.com/jsa-aerial/hanami).
```
domoritz: I might misunderstand, but I think of as a more high level tool so it should go under "Tools that use Vega-Lite". We put low level libraries in this group.

domoritz:
Thank you for the PR! I left a few comments.

site/usage/applications.md
jsa-aerial: I don't think it fits under "Tooils for Embedding ...", but as I mentioned in the thread,  it could fit under "Tools that use ..." as the closest existing heading.

The reason for having a heading like that is that most people don't think of libraries (or frameworks) as 'tools'.

domoritz: Let's put it under there and we can maybe adjust the heading as well.

site/usage/applications.md
jsa-aerial: Well, it is unclear to me. Altair is certainly rather higher level than say Oz (and so Vizard, which is really almost nothing), but both of those are also 'things' that make visualizations as opposed to 'bindings'.

Hmmmm, the more I think about it, Hanami may go better under this 'Bindings ... ' heading as it is really at a level which 1.) you create VG/VGL specs using higher level parameterized templates (which are transformed to finished specs) coupled with various capabilities for creating client only or client/server applications based on the templates.

Saite is actually a complete mini client/server application built on top of these Hanami capabilities so it probabably would fit better under "Tools that use..."

What do you think of that placement??

domoritz: I trust you to fit it under existing categories.

jsa-aerial:
Have a look at this one - I think it is better

domoritz:
Thanks!

jsa-aerial:
Thank you guys for these great libs!!

PR #4198, opened 2018-10-03
Use meta + top level props fix

### LINKED ISSUE #3610: add `usermeta` to the schemas, similarly to vega
- [x] Check for duplicate issues.

domoritz:
Move datasets from top level props to top level spec since it cannot be in config. Add support for user meta. Fixes #3610

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -2045,10 +2045,6 @@
           "description": "Default axis and legend title for count fields.\n\n__Default value:__ `'Number of Records'`.",
           "type": "string"
         },
-        "datasets": {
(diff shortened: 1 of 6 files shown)


arvind:
One thought I had while reviewing this PR: I wonder if we're able to modify our schema generator library to [separate definitions and references](https://cswr.github.io/JsonSchema/spec/definitions_references/). That is what we've done in Vega since the beginning, but that schema is hand constructed. I imagine this would significantly reduce the size of the Vega-Lite schema by cutting the amount of duplication. But, I have no idea how complex a change this would be.

domoritz:
We already use definitions and reference them or am I missing something?

PR #4201, opened 2018-10-03
Start datum encoding. 

domoritz:
Ref #1601

Starting point for someone to pick up.

### DIFF
--- a/src/fielddef.ts
+++ b/src/fielddef.ts
@@ -30,23 +30,37 @@ import {AggregatedFieldDef, WindowFieldDef} from './transform';
 import {getFullName, QUANTITATIVE, Type} from './type';
 import {contains, flatAccessWithDatum, getFirstDefined, replacePathInField, titlecase} from './util';
 
-type Value = number | string | boolean | null;
+type VisualValue = number | string;
+type DatumValue = number | string | boolean | DateTime | null;
 
 /**
- * Definition object for a constant value of an encoding channel.
+ * Definition object for a constant visual value of an encoding channel.
  */
 export interface ValueDef {
   /**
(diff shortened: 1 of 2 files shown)


src/fielddef.ts
```diff
@@ -30,23 +30,37 @@ import {AggregatedFieldDef, WindowFieldDef} from './transform';
...
+}
+
+/**
+ * Definition object for a constant datum of an encoding channel.
+ */
+export interface DatumDef {
+  /**
+   * A constant value in data domain.
+   */
+  datum: DatumValue;
```
kanitw: Don't you have to support scale (and possibly guide)?

kanitw:
https://stackoverflow.com/questions/55015304/how-do-i-add-a-layer-in-a-shape-of-a-box-to-an-altair-plot suggests that having `datum` would be useful here.

kanitw:
Once this is done, it's worth modifying the example in https://github.com/vega/vega-lite/pull/5012 too.

src/fielddef.ts
```diff
@@ -30,23 +30,37 @@ import {AggregatedFieldDef, WindowFieldDef} from './transform';
 import {getFullName, QUANTITATIVE, Type} from './type';
 import {contains, flatAccessWithDatum, getFirstDefined, replacePathInField, titlecase} from './util';
 
-type Value = number | string | boolean | null;
+type VisualValue = number | string;
```
kanitw: 1) value can be null too, why are you removing it? 

2) Better not rename and keep it `Value`

kanitw:
will be replaced by https://github.com/vega/vega-lite/pull/6150

Issue #4188, opened 2018-09-29
Composite Mark does not allow axis customization


chanwutk:
From @artfulgarfunk in issue #4164 , axis cannot be customized in composite mark.

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v3.0.0-rc5.json",
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"dose": 0.5, "response": 32659},
      {"dose": 0.5, "response": 40659},
      {"dose": 0.5, "response": 29000},
      {"dose": 1, "response": 31781},
      {"dose": 1, "response": 30781},
      {"dose": 1, "response": 35781},
      {"dose": 2, "response": 30054},
      {"dose": 4, "response": 29398},
      {"dose": 5, "response": 27779},
      {"dose": 10, "response": 27915},
      {"dose": 15, "response": 27410},
      {"dose": 20, "response": 25819},
      {"dose": 50, "response": 23999},
      {"dose": 50, "response": 25999},
      {"dose": 50, "response": 20999}
    ]
  },
  "mark": "boxplot",
  "encoding": {
    "x": {
      "field": "dose",
      "type": "ordinal",
      "axis": {"title": "WhyNoAngle", "labelAngle": 15}
    },
    "y": {
      "field": "response",
      "type": "quantitative",
      "axis": {"labelAngle": 15}
    }
  }
}
```

domoritz:
Fixed in #4202

PR #4202, opened 2018-10-03
Fix axis customization issue of composite mark

chanwutk:
From this issue #4188 

Problem: extractTransformFromEncoding drops axis after extracting encoding.


domoritz:
What is left to do in the PR?

chanwutk:
Sorry, I forget to take this off.

PR #4214, opened 2018-10-12
Travis should fail when tests fail

domoritz:
This makes sure travis fails if the tests fail and also fixes the dataflow tests.


domoritz:
I'm just going to merge this.

PR #4215, opened 2018-10-12
Update check-and-fix.sh to run on node 10 (#4193).

arvind:
Looks like we missed this with #4193 which made changes to run schema/example checking/generation only on Node v10 but forgot to update the associated bash script.


domoritz:
Good catch!

arvind:
Interesting, so it looks like `check-and-fix.sh` is now correctly rebuilding the schema/examples/etc. but isn't able to push its commits because [`remote: Anonymous access to vega/vega-lite.git denied.`](https://travis-ci.org/vega/vega-lite/jobs/440774144#L2154). 

@domoritz/@kanitw: Has something changed with access controls to Travis?

arvind:
Hmm, it seems like the `${GH_TOKEN}` variable is now blank.

arvind:
Grr, and now Travis no longer seems to be building on my pushes.

arvind:
Looks like the updated schema caught a errant find/replace, and the updated examples account for minor Vega 4.3 changes that the upgrade PR didn’t catch because of the bugs fixed by this PR. 

Going to go ahead and merge.

PR #4216, opened 2018-10-13
Replace forEach with for, remove as, remove push.apply, add test, and initialize fill array

domoritz:


arvind:
Thanks for improving the code style @domoritz, and for catching the missing unit test for `fill`!

PR #4223, opened 2018-10-17
Drop falsy title when merge

### LINKED ISSUE #4195: {"title" : null} prints out the string null for axis labelling
need to set title as null otherwise it automatically uses the Data name ("Response" in this case as the axis label), however this then prints out null as part of the axis title.

https://tinyurl.com/ya5ersw7

chanwutk:
Fixes #4195
 
Drop `{"title" : null}` when merging titles


domoritz:
Thank you!

PR #4225, opened 2018-10-18
Remove mocha ui opt from wdio config as we use chai.

### LINKED ISSUE #4166: Update wdio
We should update wdio to the latest versions but it breaks the tests right now

```
 wdio-dot-reporter       0.0.9  →  0.0.10
 wdio-mocha-framework  ^0.5.13  →  ^0.6.3
```

arvind:
This allows us to upgrade our wdio dependencies (fixes #4166).


domoritz:
👍

PR #4228, opened 2018-10-22
Use the Color type for fill and stroke

marcprux:
This is cosmetic for vega-lite itself, but other projects that use the exported schema can customize the handling of the Color type.


domoritz:
Thnk you!

PR #4243, opened 2018-11-05
Move optimizer functions to correct file

invokesus:


domoritz:
Thank you. Looks good.

PR #4245, opened 2018-11-05
Add correct deep partial type for mergeDeep

domoritz:


arvind:
I don't know enough TypeScript to understand that signature so I'm removing myself as a reviewer :)

domoritz:
I simplified the type since deep merge does not merge arrays. Now it's much easier.

src/util.ts
```diff
@@ -118,10 +118,15 @@ export function fill<T>(val: T, len: number) {
   return arr;
 }
 
+/**
+ * Like TS Partial but applies recursively to all properies.
```
invokesus:
```suggestion
 * Like TS Partial but applies recursively to all properties.
```

PR #4250, opened 2018-11-07
Fix some typos.

waywardmonkeys:


src/compile/scale/domain.ts
```diff
@@ -370,7 +370,7 @@ export function domainSort(
 
 /**
  * Determine if a scale can use unaggregated domain.
- * @return {Boolean} Returns true if all of the following conditons applies:
+ * @return {Boolean} Returns true if all of the following conditions applies:
  * 1. `scale.domain` is `unaggregated`
```
domoritz: Apply, not applies

domoritz:
Thank you for the PR!

waywardmonkeys:
Will correct that and update the PR in my morning (about 6 hours).

PR #4253, opened 2018-11-13
fix opacity discrepancy

Saba9:
Fix issue where the opacity of channels from which aggregate transformations were extracted have their opacity set to 0.7 because normalizeMarkDef() doesn't know that the channel is displaying an aggregated field after the extraction.


src/compile/mark/init.ts
```diff
@@ -25,6 +25,10 @@ export function normalizeMarkDef(mark: Mark | MarkDef, encoding: Encoding<string
     markDef.opacity = opacity(markDef.type, encoding);
```
domoritz: Does this also work if we just return undefined from `opacity(..)` when the value is 1? If not directly, we could implement the same approach as for filled below and only set the opacity when it's defined. 

The problem with your code is that undefined does not get set in json but it does get set in objects.

src/compile/mark/init.ts
```diff
@@ -25,6 +25,10 @@ export function normalizeMarkDef(mark: Mark | MarkDef, encoding: Encoding<string
...
   }
 
+  if (specifiedOpacity === 1.0) {
```
kanitw: This seems like a pretty dangerous condition to add.  Can you explain why?

src/transformextract.ts
```diff
@@ -108,3 +117,16 @@ function extractTransformsHConcat(
     })
   };
 }
+
+/* Set mark's opacity to 1.0 if its opacity is not set elsewhere */
```
kanitw: The comment should explain more why we have to do this.

src/compile/mark/init.ts
domoritz: In the transform extraction, we need to set the opacity to avoid auto opacity of 0.7. However, in this case the Vega specs look different from the original that had a default of no opacity defined. The Vega default is 1 so this is safe to do.

src/compile/mark/init.ts
Saba9: I updated the code to delete the opacity property in the case where `specifiedOpacity` equals one. I believe this addresses the concern regarding the difference between undefined in objects and JSON.

I tried an implementation where `opacity(...)` returned undefined when the mark's opacity is one but it seemed wrong since the purpose of `opacity(...)` is for the cases where the opacity property isn't explicitly defined.

Saba9:
Added a more in-depth comment regarding the need to explicitly set the mark's opacity to 1.0 during transformation extraction and changed `normalizeMarkDef(...)` to delete the opacity instead of setting it to `undefined` in the case where `specifiedOpacity === 1.0`.

src/compile/mark/init.ts
```diff
@@ -22,7 +22,10 @@ export function normalizeMarkDef(mark: Mark | MarkDef, encoding: Encoding<string
   // set opacity and filled if not specified in mark config
   const specifiedOpacity = getFirstDefined(markDef.opacity, getMarkConfig('opacity', markDef, config));
   if (specifiedOpacity === undefined) {
-    markDef.opacity = opacity(markDef.type, encoding);
+    markDef.opacity = opacity(markDef, encoding);
+  } else if (specifiedOpacity === 1.0) {
+    // Vega's default opacity is 1.0 so it's safe to delete the opacity in this case
+    delete markDef.opacity;
```
domoritz: What if the user sets `markDef.opacity` to 1 and the config is 0.5? Wouldn't this break it?

Saba9: Yeah, you're right. I didn't think of that case.

domoritz: That's why they pay me the big bucks 💰 (not really, actually...)

Saba9: Hmm, the first fix that comes to my mind would be to set the mark's opacity to the first defined of config opacity or 1.0 during transform extract and then delete the mark's opacity in `normalizeMarkDef(...)` if the mark's opacity is equal to the first defined of config opacity or 1.0.

This solution does seem pretty dirty though.

domoritz: Actually, I think I was wrong. The mark config we read from here is a Vega-Lite specific config, I think. Can you try the case I described above and delete the opacity from the Vega spec manually? I believe it should not change anything.

domoritz: Remind me why we need to apply the config property to the mark then? Isn't that counter the purpose of a config? @kanitw

kanitw: If I remember correctly, we set default config for opacity to be `undefined`, so that by default we use our smart rules.  When user set the config.mark.opacity, then we use that exact opacity number.

domoritz: 👍 makes sense now.

domoritz:
This needs some more work. @Saba9 I marked it as WIP

kanitw:
@domoritz  This seems super stale at this point.  Should we close this?

PR #4258, opened 2018-11-14
Add monthdatehours. Fixes #4256

### LINKED ISSUE #4256: monthdatehours
In vega-lite documentation [Time Unit ](https://vega.github.io/vega-lite/docs/timeunit.html) monthdatehours is not mentioned although this works and is useful. The other option yearmonthdatehours takes up too much space on the graph x-axis to be useful in certain situations.

domoritz:
We already have `MONTHDATE`.


kanitw:
I wonder if we should check if monthdate is problematic with leap year, and if so remove it instead?  (If it's not problematic, then this PR is a fair game.)

domoritz:
How would it be problematic with leap years? We still should set the correct month, date, and hour but just in the year 1900 instead of whatever year it was originally.

kanitw:
> We still should set the correct month, date, and hour but just in the year 1900 instead of whatever year it was originally.

But 1900 isn't a leap year, so if you enter 29 Feb (and whatever optional hours) of 1900, then wouldn't it become 1 Mar?

domoritz:
Oh, the this comment is wrong: `// MONTHDATE always include 29 February since we use year 0th (which is a leap year);`

Anyway, in JS setting the year to 0 actually means 1900. The first leap year after that is 1972 so I propose that we use that instead of 0.

kanitw:
> The first leap year after that is 1972 

1972 is fine but the first leap year after 1900 is actually 1904.  (Leap pretty much happens every four years, except when the year number is divisible by 100. See https://kalender-365.de/leap-years.php)

domoritz:
Sorry, I meant the first leap year after unix time (https://en.wikipedia.org/wiki/Unix_time) starts is 1972.

domoritz:
Let me know if there is anything left to do.

test/timeunit.test.ts
```diff
@@ -1,200 +1,201 @@
-import {assert} from 'chai';
 import {containsTimeUnit, convert, fieldExpr, formatExpression, TimeUnit} from '../src/timeunit';
 
 describe('timeUnit', () => {
   describe('containsTimeUnit', () => {
     it('should return true for quarter given quarter', () => {
       const fullTimeUnit = TimeUnit.QUARTER;
       const timeUnit = TimeUnit.QUARTER;
-      assert.equal(containsTimeUnit(fullTimeUnit, timeUnit), true);
+      expect(containsTimeUnit(fullTimeUnit, timeUnit)).toBeTruthy();
```
kanitw: Why truthy, not true? (Shouldn't this method return boolean?)

domoritz: Yeah, you're right. I fixed it.

Issue #4155, opened 2018-09-11
Interactive views should use cursors to provide affordance


domoritz:
For example, when the user can create a brush, it would be nice to change the cursor accordingly.

kanitw:
When fixing this issue, please also update [this example](
https://github.com/vega/vega-lite/pull/4259#discussion_r233605024).

kanitw:
Single/Multi -- Use `pointer`

Brush use `crosshair`

Pan uses `grab` -- if there is a smart way to learn about this

domoritz:
Falcon uses pretty reasonable cursors.

kanitw:
To make it easy to follow, what’re the difference between suggested set above?

domoritz:
<img width="1052" alt="screen shot 2018-12-05 at 09 39 22" src="https://user-images.githubusercontent.com/589034/49532570-aa7b4b80-f871-11e8-9d03-03e75b53f897.png">

And you can give it a try at https://uwdata.github.io/falcon/flights/. 

The difference is that resize has a cursor as well (which we don't support yet) and I use `move` instead of `grab`.

ajbouh:
Can this be done today with the cursor signal?

domoritz:
Vega-Lite does not yet have support for signals. You can modify the Vega, yes.

domoritz:
We now have move cursors over brushes for interval selections thanks to @alliefeldman in #9164.

joelostblom:
[...]

PR #4259, opened 2018-11-14
Add fill/strokeOpacity and strokeWidth channels to support interaction design

kanitw:
Also added an example of an interactive bar chart with highlighting on hover and selecting on click. (Inspired by Tableau's interaction style.)

tableau

kapture 2018-11-14 at 11 08 57

Note that I also added cursor: pointer after capturing the video.


examples/specs/interactive_bar_select_highlight.vl.json
```diff
@@ -0,0 +1,49 @@
...
+      "condition": {"selection": "select", "value": 1},
+      "value": 0.3
+    },
+    "strokeWidth": {
+      "condition": [
+        {
+          "test": {
+            "and": [
+              {"selection": "select"},
+              "length(data(\"select_store\"))"
```
kanitw: Note that this is a cheat by directly accessing the underlying data store to check if the selection is not empty. I think we might wanna expose this as a type of predicate in the future.

examples/specs/interactive_bar_select_highlight.vl.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "selection": {
+    "highlight": {"type": "single", "empty": "none", "on": "mouseover"},
+    "select": {"type": "multi"}
+  },
+  "mark": {
+    "type": "bar",
+    "fill": "#4C78A8",
+    "stroke": "black",
+    "cursor": "pointer"
```
domoritz: I think this should be the default. Let's not add it manually just here, okay?

examples/specs/interactive_bar_select_highlight.vl.json
domoritz: Don't we have a way to change the selection to be empty or everything when nothing is selected by the user?

examples/specs/interactive_bar_select_highlight.vl.json
kanitw: I disagree. I don't know when that default gonna land. You can undo that in a separate PR.

examples/specs/interactive_bar_select_highlight.vl.json
kanitw: We do, but it doesn't work for this case since we need the `opacity` to be `1` when the `"select"` selection is empty but the `strokeWidth` to be `0` when the `"select"` selection is empty.

examples/specs/interactive_bar_select_highlight.vl.json
domoritz: Okay, please create an issue and link to this example then.

examples/specs/interactive_bar_select_highlight.vl.json
domoritz: Can you make two selections?

kanitw: Nope, that totally doesn't make sense. There should be only one selection for clicking that gets applied to production rule differently.

examples/specs/interactive_bar_select_highlight.vl.json
kanitw: lol you already created it a few months ago. https://github.com/vega/vega-lite/issues/4155

kanitw:
> Another simple design is to change the opacity/darkness on hover. See https://schasins.com/blog/when-do-cs-researchers-peak.html

That doesn't work if you want to handle both hovering to highlight AND clicking to do (hard) selection.

examples/specs/interactive_bar_select_highlight.vl.json
domoritz: Haha, shows how much I want it.

PR #4263, opened 2018-11-16
Add support to pre-aggregated-error error bar and error band

### LINKED ISSUE #4131: Error bar macro for value + error
Commonly users come with data of the form `value, error`. Our errorbar macro should support this use case and not require users to create a transform. 

```
mark: "errorbar",
encoding: {
  y: {field: value, type: "quantitative"},
  yError: {field: error, type: "quantitative"}
}
```

```json
{
  "data": {
    "values": [
      {
        "yield_error": 3,
        "yield": 32.4,
        "variety": "Glabron"
      },
      {
        "yield_error": 3.2,
        "yield": 30.96667,
        "variety": "Manchuria"
      },
      {
        "yield_error": 1.2,
        "center": 33.966665,
        "variety": "No. 457"
      },
      {
        "yield_error": 2,
        "yield": 30.45,
        "variety": "No. 462"
      }
    ]
  },
  "layer": [
    {
      "mark": "errorbar",
      "encoding": {
        "x": {
          "field": "yield",
          "type": "quantitative",
          "scale": {"zero": false},
          "title": "yield"
        },
        "xError": {"field": "yield_error", "type": "quantitative"},
      }
    },
    {
      "mark": {
        "type": "point",
        "filled": true,
        "color": "black"
      },
      "encoding": {
        "x": {
          "field": "yield",
          "type": "quantitative"
        },
      }
    }
  ]
}
```

chanwutk:
resolve #4131


domoritz:
Can you add an example?

chanwutk:
Yes.

src/encoding.ts
```diff
@@ -59,6 +59,26 @@ export interface Encoding<F> {
   // TODO: Ham need to add default behavior
   y2?: FieldDefWithoutScale<F> | ValueDef;
 
+  /**
+   * Error value of x coordinates for error specified `"errorbar"` and `"errorband"`.
+   */
+  xError?: FieldDefWithoutScale<F> | ValueDef;
```
kanitw: Are we okay with this being suggested as possible channel for other marks?

kanitw:
> Can you add an example?

and update docs?

src/encoding.ts
chanwutk: I and @domoritz agreed that we would want to get this work first, then make `x/yError` only supported by error bar and error band later because it seems to be a big typing refactor in Vega-Lite.

chanwutk:
Sure

src/encoding.ts
domoritz: It might make sense to already have a new spec type for composite marks but I stand with my decision not to make new types for all mark types in the same PR as supporting errors.

kanitw:
From #4131

> How is a user supposed to know which error (xError or xError2) is positive or negative?

I don't see any of more discussion about this so I'm quite surprised that this is merged already.

domoritz:
We decided that it does not matter. You have to use a positive and a negative value.

domoritz:
👍

PR #4267, opened 2018-11-20
Allow titles of type FieldDefBase to be used with TitleMixins [WIP]

Saba9:
As mentioned by @kanitw in #4254 a cleaner approach to fixing the problem of differing title behavior between specs which have had extracted transformations and those without is to allow the `title` field in `TitleMixins` to be of type `string | FieldDefBase<string>[]` where the latter type is used to define the correct implicit title.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "deploy:gh": "scripts/deploy-gh.sh",
     "deploy:schema": "scripts/deploy-schema.sh",
     "preschema": "npm run prebuild",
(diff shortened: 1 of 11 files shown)


Saba9:
For some reason the schema build process fails and throws an exception stating that "The call stack size exceeded" after making the type change in `guide.ts`. Not sure what's going on with that.

domoritz:
As discussed, increase the stack size.

kanitw:
Is this still needed? If not, please close.

Saba9:
There are a certain specs which I believe are impossible to extract transforms from while maintaining the same output without this. I stopped working on this for a period of time because of the talk of extracting transformations from Vega rather than VL, however, since it was concluded that we need transformation extract for composite marks it is still relevant.

I'll work on merging master into this branch and getting it ready for merging.

Saba9:
Replaced by #4967

PR #4269, opened 2018-11-20
Optimize dataflow by merging AggregateNodes with the same dimensions

Saba9:
Added an optimization which merges AggregateNodes with the same dimensions. Also modified AggregateNode to take a StringSet of aliases rather than a single string in order to assist with the merging process. 

It might be better to add a CalculateNode in this case but the StringSet solution seems simpler. Another possibility would be to add an optimization to generate CalculateNodes after the aggregate merging occurred.


src/compile/data/aggregate.ts
```diff
@@ -155,11 +156,15 @@ export class AggregateNode extends DataFlowNode {
...
-    const out = {};
+    let out = {};
 
     for (const field of keys(this.measures)) {
       for (const op of keys(this.measures[field])) {
-        out[this.measures[field][op] || `${op}_${field}`] = true;
+        if (keys(this.measures[field][op]).length === 0) {
+          out[`${op}_${field}`] = true;
+        } else {
+          out = {...out, ...this.measures[field][op]};
```
domoritz: `Object.assign(out, ...this.measures[][])`  or just `assign`

src/compile/data/optimizers.ts
```diff
@@ -294,3 +294,30 @@ export class MergeParse extends BottomUpOptimizer {
     return this.flags;
   }
 }
+
+export class MergeAggregateNodes extends BottomUpOptimizer {
+  public run(node: DataFlowNode): optimizers.OptimizerFlags {
+    const parent = node.parent;
+    const aggChildren = parent.children.filter((x): x is AggregateNode => x instanceof AggregateNode);
+    for (let i = 0; i < aggChildren.length; i++) {
```
domoritz: Convert to linear complexity as discussed.

src/compile/data/aggregate.ts
Saba9: I believe `Object.assign` is exclusive to ES6. The only reference to assign I could find was `__assign(...)` from tslib which we probably shouldn't be using (?).

src/util.ts
```diff
@@ -181,15 +181,21 @@ export type StringSet = Dict<true>;
...
-        return true;
-      }
+export function isEqual<T>(dict: Dict<T>, other: Dict<T>) {
+  const dictKeys = keys(dict);
+  const otherKeys = keys(other);
+  if (dictKeys.length !== otherKeys.length) {
+    return false;
+  }
+  for (const key of dictKeys) {
+    if (!contains(otherKeys, key)) {
```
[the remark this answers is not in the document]
Saba9: Yup. Fixed in the latest commit

domoritz:
@kanitw Can you merge this since I was involved in the design process?

kanitw:
Awesome!

kanitw:
Btw, note that we only need to be super careful for new syntax addition (like the error bar case).

Feel free to merge simple bug fixes and optimization. (If you have complicated bug fix / optimization, feel free to loop me in.)

PR #4278, opened 2018-11-26
Fix duplicate data issue

### LINKED ISSUE #4274: Duplicate Dataset error in Vega-Lite 3
Recently an Altair user discovered this error in the Vega editor, which is running Vega 3. Example here:

[Vega Editor link](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEDGD2B2AzAlgc0gLjMSA3ZApgO6baRHIAmALgBakAsADEwDRS0Fq3WkDMLAL6DW4KABsAhgE8CAJ1IBtMRFAR1USpOqTSOeJIC2BUpC06AtNAIB2ABwAjSXbsNoAThsBGPpMl9EBwIGG18vREoCADYAVkkTERV1SENJOQBrUwAHWGR4XlENCEgCeDhKPPQsNSLigA89SGppLJMsSABHAFdJfOQdamRcE3ZIFAJxSlMGxNri6Ubm1tNu3sGBoZGoccnTBcEkiAONWfUajTNtXWrIA2NTc0krW0dnVw9vX39A4NDJcMisXikFOF1SGVM4jyI0OUFK5Uqelh9UWLTaUFWfQ2w0gox2U3aM0Kc0gCxuS3RnR6WO0m1x20Iu3a+1hx3UbIAusTIAASADO0E4qVMPGoWT5GAA9JLhqhJAA6VD9WhdBzy5CwSUCoWSGUEOUWKHUAgygBM8qi8qY8oAVnyELixJcdHyCNQ+UiLo9nvYnC43J4fH4AkEQmEItE4ujlLUcA0sGwoGSwExQWdIPGwF5RsmvGnVBnSKac6RU0kOWIDoIgA)

The Altair chart that generated this is as follows:
```python
import altair as alt
import pandas as pd

data = pd.DataFrame({'x': [0, 1, 2], 'y': [0, 1, 0]})
chart = alt.Chart(data).encode(x='x', y='y')
chart.mark_point() + chart.mark_line()
```

Is this intended, that a named dataset can only be referred to once within any compound chart in Vega-Lite 3? If so, we'll need to rework how Altair handles data internally.

domoritz:
Fixes #4274


Saba9:
It's not clear to me why`isNamedData(otherData)`was returning false and resulting in the buggy behavior. Would it be more appropriate to change it to return true in these cases or is `otherData` not truly NamedData?

domoritz:
`otherData` is not necessarily a data source since we extract the name when we parse a dataset. My reasoning was that I wanted to have the name only in one place.

domoritz:
I added some tests

domoritz:
@kanitw @arvind please merge when you think this is ready

src/compile/data/source.ts
```diff
@@ -77,6 +77,11 @@ export class SourceNode extends DataFlowNode {
   }
 
   public assemble(): VgData {
+    // remove empty format
+    if (this._data.format && keys(this._data.format).length === 0) {
```
kanitw: Would this cause side effect?

domoritz: On the original data? Hmm, let me check that.

domoritz: No problem. We assign a new object to `this._data`.

PR #4281, opened 2018-11-26
Run tests on node 11

domoritz:


kanitw:
Can you give more context i.e., explain the rationale behind the change?

kanitw:
But didn't node only have long-term-support (LTS) for even number versions?  

If so, shouldn't the main test be 10 until 12 is out?

domoritz:
Merge @kanitw?

kanitw:
I don't have strong opinion here. So if you strongly think we should do this, go ahead.

PR #4284, opened 2018-11-26
Add ChannelDefWithCondition with stricter types. Fixes #4236

### LINKED ISSUE #4236: Empty color encoding passes JSON schema validation
In vega-lite 3.0.0rc8, the following JSON passes schema validation (in the online vega-lite editor, as well as elsewhere) with an empty "color" encoding object:

```json
{
  "data": {
    "values": [ {"a": "A","b": 28} ]
  },
  "mark": "bar",
  "encoding": {
    "color": { }
  }
}
```

This isn't the case with other encodings like "x" or "y", since they will complain that the "type" field is required. The issue seems to be that "color" is `anyOf` either `MarkPropFieldDefWithCondition` or `MarkPropValueDefWithCondition`, but since `MarkPropValueDefWithCondition` consists of only non-required properties `condition` and `value`, an empty object passes validation.

The underlying issue seems to be that in the `ValueDefWithCondition` type, `condition` and `value` are both optional. It would be nice to be able to express that either one or the other be optional, but since that isn't possible in TypeScript, one solution might be make `value` a required member of `ValueDefWithCondition` and add a new type `ConditionWithoutAnyValue`, and change `ChannelDefWithCondition` to be like:

```typescript
export type ChannelDefWithCondition<F extends FieldDef<any>> = FieldDefWithCondition<F> | ValueDefWithCondition<F> | ConditionWithoutAnyValue<F>;
```

Note that there is a runtime warning about the property ("Dropping {} from channel "color" since it does not contain data field or value."), but since the schema can require that there be a valid value, this enables external tools to be stricter with the vega-lite specs that can be generated.

domoritz:
@marcprux could you check this?

### DIFF
--- a/src/encoding.ts
+++ b/src/encoding.ts
@@ -7,6 +7,7 @@ import {Config} from './config';
 import {FacetMapping} from './facet';
 import {
   ChannelDef,
+  ChannelDefWithCondition,
   Field,
   FieldDef,
   FieldDefWithCondition,
@@ -90,7 +91,7 @@ export interface Encoding<F> {
    * 1) For fine-grained control over both fill and stroke colors of the marks, please use the `fill` and `stroke` channels.  If either `fill` or `stroke` channel is specified, `color` channel will be ignored.
    * 2) See the scale documentation for more information about customizing [color scheme](https://vega.github.io/vega-lite/docs/scale.html#scheme).
    */
-  color?: FieldDefWithCondition<MarkPropFieldDef<F>> | ValueDefWithCondition<MarkPropFieldDef<F>>;
+  color?: ChannelDefWithCondition<MarkPropFieldDef<F>>;
 
   /**
    * Fill color of the marks.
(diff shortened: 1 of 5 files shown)


marcprux:
Looks great!

src/encoding.ts
```diff
@@ -90,50 +91,50 @@ export interface Encoding<F> {
    * 1) For fine-grained control over both fill and stroke colors of the marks, please use the `fill` and `stroke` channels.  If either `fill` or `stroke` channel is specified, `color` channel will be ignored.
    * 2) See the scale documentation for more information about customizing [color scheme](https://vega.github.io/vega-lite/docs/scale.html#scheme).
    */
-  color?: FieldDefWithCondition<MarkPropFieldDef<F>> | ValueDefWithCondition<MarkPropFieldDef<F>>;
+  color?: ChannelDefWithCondition<MarkPropFieldDef<F>>;
```
kanitw: Please do NOT use `ChannelDefWithCondition` in Encoding's declaration because the docs should make it clear whether certain channel support valueDef or not. 

image

image

src/fielddef.ts
```diff
@@ -102,7 +105,20 @@ export interface ValueDefWithCondition<F extends FieldDef<any>> {
...
+  value: Value;
+}
+
+/**
+ * A Condition<ValueDef | FieldDef>
+ * {
+ *   condition: {field: ...} | {value: ...}
+ * }
+ */
+export interface ConditionWithoutAnyValue<F extends FieldDef<any>> {
```
kanitw: `ConditionOnlyDef`?   It's without *neither* value nor field.

domoritz:
Replaced by https://github.com/vega/vega-lite/pull/4288

PR #4285, opened 2018-11-26
Switch from uglifyjs to terser as it support es6.

domoritz:


kanitw:
We shouldn't merge this until RStudio migrate?

PR #4288, opened 2018-11-26
Make ValueDefWithCondition type more strict. Fixes #4236

### LINKED ISSUE #4236: Empty color encoding passes JSON schema validation

domoritz:
Replaces https://github.com/vega/vega-lite/pull/4284

### DIFF
--- a/src/fielddef.ts
+++ b/src/fielddef.ts
@@ -48,6 +48,15 @@ export interface ValueDef {
  */
 export type ChannelDefWithCondition<F extends FieldDef<any>> = FieldDefWithCondition<F> | ValueDefWithCondition<F>;
 
+/**
+ * A ValueDef with Condition<ValueDef | FieldDef> where either the conition or the value are optional.
+ * {
+ *   condition: {field: ...} | {value: ...},
+ *   value: ...,
+ * }
+ */
+export type ValueDefWithCondition<F extends FieldDef<any>> = ValueDefWithOptionalCondition<F> | ConditionOnlyDef<F>;
+
 export type Conditional<T> = ConditionalPredicate<T> | ConditionalSelection<T>;
 
 export type ConditionalPredicate<T> = {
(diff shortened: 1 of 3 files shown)


marcprux:
LGTM

src/fielddef.ts
```diff
@@ -48,6 +48,15 @@ export interface ValueDef {
...
 export type ChannelDefWithCondition<F extends FieldDef<any>> = FieldDefWithCondition<F> | ValueDefWithCondition<F>;
 
+/**
+ * A ValueDef with Condition<ValueDef | FieldDef> where either the conition or the value are optional.
+ * {
+ *   condition: {field: ...} | {value: ...},
+ *   value: ...,
+ * }
+ */
+export type ValueDefWithCondition<F extends FieldDef<any>> = ValueDefWithOptionalCondition<F> | ConditionOnlyDef<F>;
```
kanitw: Technically, `ConditionOnlyDef` isn't a ValueDef, but I think this is good for the docs since the docs won't include `ConditionOnlyDef`, which is a bit too complicated to include.

PR #4290, opened 2018-11-26
Wrap overflowing code

domoritz:
Before

screen shot 2018-11-26 at 14 50 20

After

screen shot 2018-11-26 at 14 50 14

### DIFF
--- a/site/static/main.css
+++ b/site/static/main.css
@@ -172,7 +172,7 @@ code, tt {
 pre code {
   margin: 0;
   padding: 0;
-  white-space: pre-wrap;
+  white-space: pre;
   border: none;
   background: transparent;
   font-family: monospace;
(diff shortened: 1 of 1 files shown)


arvind:
Is this preferable to `overflow: auto` or `scroll`? I wonder if it'll get confusing if non-description lines of the spec every wrap?

domoritz:
Updated

domoritz:
Hmm, we actually use wrapping in our docs: https://vega.github.io/vega-lite/docs/flatten.html. The advantage of wrapping is that you never have to scroll horizontally (which many people can't do).

domoritz:
I reverted to wrapping.

domoritz:
@kanitw @arvind Is this good?

kanitw:
wrapping is good.  :)

domoritz:
@kanitw approve the pr / merge it then?

PR #4139, opened 2018-08-20
Initialize Selections

arvind:
**This PR relies on the refactoring introduced by #4068, so please merge that one first.**

This PR adds an `init` property to selection definitions to initialize them using the fields or encoding channels a selection is projected over. For example, here is an updated version of our `interactive_query_widgets` example that initializes the selection to select cars with 8 cylinders manufactured in 1971. 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "Drag the sliders to highlight points.",
  "data": {"url": "data/cars.json"},
  "transform": [{"calculate": "year(datum.Year)", "as": "Year"}],
  "layer": [{
    "selection": {
      "CylYr": {
        "type": "single", "fields": ["Cylinders", "Year"],
        "init": {"Cylinders": 4, "Year": 1977},
        "bind": {
          "Cylinders": {"input": "range", "min": 3, "max": 8, "step": 1},
          "Year": {"input": "range", "min": 1969, "max": 1981, "step": 1}
        }
      }
    },
    "mark": "circle",
    "encoding": {
      "x": {"field": "Horsepower", "type": "quantitative"},
      "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
      "color": {
        "condition": {"selection": "CylYr", "field": "Origin", "type": "nominal"},
        "value": "grey"
      }
    }
  }, {
    "transform": [{"filter": {"selection": "CylYr"}}],
    "mark": "circle",
    "encoding": {
      "x": {"field": "Horsepower", "type": "quantitative"},
      "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
      "color": {"field": "Origin", "type": "nominal"},
      "size": {"value": 100}
    }
  }]
}
```

Similarly, an updated `interactive_brush` example to initialize the brush selection is as follows:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "description": "Drag out a rectangular brush to highlight points.",
  "data": {"url": "data/cars.json"},
  "selection": {
    "brush": {
      "type": "interval",
      "init": {"x": [55, 160], "y": [13, 37]}
    }
  },
  "mark": "point",
  "encoding": {
    "x": {"field": "Horsepower", "type": "quantitative"},
    "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
    "color": {
      "condition": {"selection": "brush", "field": "Cylinders", "type": "ordinal"},
      "value": "grey"
    }
  }
}
```

[...]

Issue #1966, opened 2017-02-18
Follow up Minor Fixes for `as/selections`


kanitw:
We'll merge https://github.com/vega/vega-lite/pull/1795 first so the code doesn't linger for long and we can build production rules.  However, at some point it's probably worth addressing comment left in the PR (including ones that got collapsed due to file  renaming / other changes on the same lines).

A few style comments:
- Let's keep `index.ts` -- I think it's a better style. We should restyle the rest of the code base to use index.  (Sorry about that!)

- If you do not want to declare full interface for a thing, giving it a name and do
`type XXX = any` can make the code is a bit more understandable. 

- When you add more code later, let's try avoid using default export ([why?](https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html)).  
If we have time (definitely not important for now), we might wanna turn on a tslint flag for this for consistency, so not adding more reduce refactor efforts required if we wanna do so in the future.

arvind:
* [ ] Add unit tests for code paths introduced by multiview operators.

kanitw:
I think we can revise these later ...

PR #4291, opened 2018-11-27
Refactor selection code paths

arvind:
Please merge after #4139. 

This PR cleans up the selection code paths into more manageable `parse` and `assemble` modules that mimic the structure elsewhere. It also reintroduces a `compile/selection/index.ts` as per #1966 to hold the definitions needed on the compiler side

### DIFF
--- a/src/compile/selection/index.ts
+++ b/src/compile/selection/index.ts
@@ -1,0 +1,113 @@
+import {Binding, NewSignal, SignalRef} from 'vega';
+import {stringValue} from 'vega-util';
+import {FACET_CHANNELS} from '../../channel';
+import {
+  BrushConfig,
+  SELECTION_ID,
+  SelectionInit,
+  SelectionInitArray,
+  SelectionResolution,
+  SelectionType
+} from '../../selection';
+import {accessPathWithDatum, Dict} from '../../util';
+import {EventStream} from '../../vega.schema';
+import {FacetModel} from '../facet';
+import {isFacetModel, Model} from '../model';
+import {UnitModel} from '../unit';
+import interval from './interval';
+import multi from './multi';
+import single from './single';
+import {SelectionProjection, SelectionProjectionComponent} from './transforms/project';
+
+export const STORE = '_store';
+export const TUPLE = '_tuple';
+export const MODIFY = '_modify';
+export const SELECTION_DOMAIN = '_selection_domain_';
+export const VL_SELECTION_RESOLVE = 'vlSelectionResolve';
+
+export interface SelectionComponent<T extends SelectionType = SelectionType> {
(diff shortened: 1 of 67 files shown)


domoritz:
@arvind Is this still relevant?

kanitw:
I think the things that this PR (the last 2 commits) do is useful, but it's been quite stale.  I've tried rebasing this but it leads to so many conflicts.  

I think the best solution is to redo the refactor manually, so we don't introduce regressions due to this refactor.

arvind:
Yes, still very relevant and I just rebased. It looks like all tests are passing except one:

```
 FAIL  test/transformextract.test.ts (11.87s)
  ● extractTransforms() › should compile ./examples/specs/interactive_area_brush.vl.json to the same spec
```

Not quite sure why only this one test specifically would be failing...

arvind:
What's weird is when I copy the expected/received values and manually diff them, the diff comes up empty...

kanitw:
I'd add `interactive_area_brush` to the `failsList` in `transformextract.test.ts` for now.

For context, the `extractTransform()` should preserve the Vega output for all specs.  However, there are some cases that it's still failing, and this might be one of them.   We should fix them, but it's not a priority for 3.0.

arvind:
Thanks @kanitw! Updated and the tests are passing except for Node 11 (perhaps we should only run Travis against Node LTS versions?).

arvind:
I'm not entirely sure what's going with our tests. After rebasing on master, skipping `interactive_area_brush.vl.json` in `transformextract` failed locally but [passed on Travis](https://travis-ci.org/vega/vega-lite/builds/503624846). I force pushed to no longer skip it and it [still passes on Travis](https://travis-ci.org/vega/vega-lite/builds/503638725)!

arvind:
Ok, @kanitw and I resolved to merge this before it runs stale again and investigate the testing issue later.

src/compile/selection/index.ts
```diff
@@ -0,0 +1,113 @@
...
+export const SELECTION_DOMAIN = '_selection_domain_';
+export const VL_SELECTION_RESOLVE = 'vlSelectionResolve';
+
+export interface SelectionComponent<T extends SelectionType = SelectionType> {
+  name: string;
+  type: T;
+
+  // Use conditional typing (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html)
+  // so we have stricter type of init (as the type of init depends on selection type)
+  init?: (T extends 'interval'
```
domoritz: Why is `init` an array here? It seems like it should only be a `SelectionInitArray`, `SelectionInit`, or `SelectionInit | SelectionInit[]`, no?

PR #4296, opened 2018-11-28
Remove all side effects from the compiler

kanitw:


domoritz:
👍 Thanks for adding the test!

examples/examples.test.ts
```diff
@@ -57,12 +57,17 @@ describe('Examples', () => {
...
     const jsonSpec = JSON.parse(fs.readFileSync('examples/specs/' + example));
+    const stringifiedSpec = JSON.stringify(jsonSpec);
 
     describe(
       example,
       log.wrap(localLogger => {
         const vegaSpec: VgSpec = compile(jsonSpec).spec;
 
+        it('should not cause any side effects', () => {
+          expect(JSON.stringify(jsonSpec)).toEqual(stringifiedSpec);
```
domoritz: Shouldn't `const jsonSpec = duplicate(josnSpec); expect(jsonSpec).toEqual(originalJsonSpec)` work as well? It should show the diff in the tests so that would be good.

Issue #4237, opened 2018-10-30
standardize format for internal fields


domoritz:
We should have a reserved naming scheme for fields that are generated by us.

domoritz:
Let's use `__name` for now. Also, let's create a single function that does this so it's easy to change/track.

kanitw:
What u gonna donwith tooltip? This will make tooltip names ugly by default and will make it buggy with __ prefix.

I wanna see a more convincing argument why we need this change. (There is no argument in this issue yet.)

domoritz:
The argument is that field names like `count_*` could cause problems in databases and that users may define their own `foo_mean`. By cleanly separating user defined and internal fields, we avoid possible collisions. I agree that we have to support clean tooltips so this may be a good opportunity to address this issue.

kanitw:
`__name__` ?   like `__data__` in D3.

domoritz:
Not sure why we need the additional _ at the end but the particular format does not matter that much to me (although the shorter is better).

PR #4300, opened 2018-11-29
Use standardized field prefixes for generated fields [WIP]

Saba9:
Addresses #4237. Currently only implemented for the count field.


Saba9:
A few of our example specs such as `layer_text_heatmap` make reference to the `count_*` field. They'll have to be changed in order to work entirely. This also means that these changes will break any user's specs referencing internally generated fields.

domoritz:
> This also means that these changes will break any user's specs referencing internally generated fields.

Excatly for this reason I want to make the names more explicit so that we make that expectation clear. 

cc @kanitw

kanitw:
Are there that many specs like that?   I think for `layer_text_heatmap` we should just manually extract the transform so we don't have to "hack" like this.

Saba9:
It's only in `layer_text_heatmap` and `layer_text_heatmap_normalized`.

domoritz:
`__foo` is easier to check (you just check the first two characters). I don't care about symmetry.

kanitw:
ok whatever. It's not a big deal.

domoritz:
Wer're now doing this for count. @Saba9 please redo this for other internal fields.

PR #4305, opened 2018-12-01
[Docs] Improve line marker examples + add stroked version

kanitw:


kanitw:
Note that  examples/compiled/line_overlay_stroked.svg has a bug for legend, which should be fixed in https://github.com/vega/vega-lite/pull/4304, but we could merge squash this first to keep commits atomic.

Issue #4308, opened 2018-12-04
`Datasets` is not well documented


kanitw:
https://vega.github.io/vega-lite/docs/data.html#datasets

- Menu bar doesn't list `dataset`  -- we should change data -> data/datasets
- The intro paragraph in the data page doesn't talk about `dataset` at all. So `dataset` is simply hidden at the end of the page.

kanitw:
Fixed in #4309

PR #4309, opened 2018-12-04
Fix data docs

kanitw:
Fix #4038


domoritz:
You meant that this fixes #4308, not #4038

PR #4311, opened 2018-12-04
Update documents and example name for #4263

### LINKED ISSUE #4299: Follow up fix for #4263
See comments in https://github.com/vega/vega-lite/pull/4263

chanwutk:
Fix #4299


site/docs/mark/errorband.md
```diff
@@ -22,12 +23,14 @@ Error band in Vega-Lite can either be used to aggregate raw data or directly vis
 To create an error band, set `mark` to `"errorband"`.
 
 ## Documentation Overview
+
 {:.no_toc}
 
 - TOC
-{:toc}
+  {:toc}
```
domoritz: why?

site/docs/mark/errorbar.md
```diff
@@ -22,12 +23,14 @@ Error bars in Vega-Lite can either be used to aggregate raw data or directly vis
 To create an error bar, set `mark` to `"errorbar"`.
 
 ## Documentation Overview
+
 {:.no_toc}
 
 - TOC
-{:toc}
+  {:toc}
```
domoritz: Why

site/docs/mark/errorband.md
chanwutk: This is what prettier does for me. Should I revert back?

kanitw:
> Hmm, this PR does not seem to delete the existing SVGs/compiled specs, does it?

I think the script do not delete them by default.  So you have to manually delete them.  Or even better, update the script to do so.

domoritz:
Yep, let's fix the script. I'm running an experiment in https://github.com/vega/vega-lite/pull/4312 right now.

domoritz:
@chanwutk Please update the script to generate specs/svgs to either delete them or create an issue to fix this later.

kanitw:
The tricky part is that we only re-run SVGs when Vega changes.  

So we might want to
1) Always delete all Vega
2) Check if any remaining SVGs do not Vega source and delete them.

site/docs/mark/errorband.md
domoritz: Please rebase so you have the right format.

site/docs/mark/errorbar.md
domoritz: Still needs to be reverted

scripts/check-and-fix.sh
```diff
@@ -46,17 +46,17 @@ fi
 # Commit examples if outdated
 
 # Note: we need to add all files first so that new files are included in `git diff --cached` too.
-git add ./examples/compiled/vega_version ./examples/compiled/*.vg.json ./examples/compiled/*.svg ./examples/specs/normalized/*.vl.json
+git add ./examples/compiled/vega_version './examples/compiled/*.vg.json' './examples/compiled/*.svg' './examples/specs/normalized/*.vl.json'
```
kanitw: Why are these script fix a part of this follow fix?  

If they're not the same issue, I don't think we should merge them in the same PR?

kanitw: NVM -- it's good to combine them given they're a good test bed for this issue.

kanitw:
Thank you :)

PR #4314, opened 2018-12-04
[WIP] Change default legend orient to horizontal

kanitw:


kanitw:
I'm not more convinced that vertical is probably a better default.  Let's close this.

Issue #1657, opened 2016-11-02
Possible to use legend to filter graph data?


brylie:
NVD#, and possibly other charting libraries, allow for charts to be filtered by interacting with the chart legend:

![peek 2016-11-02 12-16](https://cloud.githubusercontent.com/assets/17307/19924835/4ea5a276-a0f6-11e6-8232-5458e09b6e7e.gif)

This is useful for exploratory analysis, and makes toggling categories very easy.

# Feature
Is it planned/possible for vega-lite charts to have interactive legends for filtering chart data?

kanitw:
Thanks for asking. This will likely be a part of v2.x release.

domoritz:
Interactions with legends was a bit tricky with vega (at least v2). Has that changed? cc @arvind

arvind:
Yes, with Vega 3, axes and legends can now be interacted with. So this should be supported with Vega-Lite 2.0. The open question with interactions, however, is what defaults should be. For example, if a legend is added to the visualization, should it be interactive by default? Should the interaction it supports also extend to interacting with the plotted data points as well? Etc.

jheer:
We should definitely try to support these kinds of interaction! Not sure if interactive-by-default is the right way to go, but we would need to be able to provide explicit control in either case.

This relates to the selection `project` transform. As I understand it, our default design right now is no projection for single/multi point selections, and [x, y] projection for interval selections. For axes and continuous (color gradient) legends, I imagine one could define an interval selection projected to the appropriate field/channel/guide. For discrete legends, we might need to default a single-point or multi-point selection to project just the field(s) backing the legend. 

There is also the potential issue that users might just want to be able to (more or less) say: "I want to enable selections on this legend (or axis)" without having to think about the appropriate selection type (interval, etc).

domoritz:
I don't think we want any interactions by default. The problem is that I don't know what channel we would use (e.g. for highlighting) and how we resolve conflicts if the user uses that channel to encode data.

kanitw:
RE: default interaction, let's create a new issue (#1658) to discuss that 
since it's going a little off topic here.

brylie:
For what it's worth, NVD3 chart legends are interactive by default.

domoritz:
We have this on the roadmap for 2.1

kanitw:
Given that we will follow semantic versioning, it is probably more accurate to say "very early versions of 2.x minor release".

domoritz:
[...]

PR #4317, opened 2018-12-04
[WIP] Interactive Legends

djbarnwal:
To be merged after PR #4291 

This PR attempts to solve Issue #1657. Legends can now respond to user clicks when `legend: true` is present within a selection containing a project transform.


src/selection.ts
```diff
@@ -50,6 +50,12 @@ export interface BaseSelectionDef {
    * When set to `none`, empty selections contain no data values.
    */
   empty?: 'all' | 'none';
+
+  /**
+   * A boolean determining whether the legend if present should be interactive
+   * or not
+   */
+  legend?: boolean;
```
kanitw: Have you consider other alternative syntax for expressing this?

For example, if a selection is projected on X (`encodings: ["x"]`), what would `legend: true` do?

Also,  if a selection is projected on Color (`encodings: ["color"]`), should legend be interactive by default?

src/compile/selection/assemble.ts
```diff
@@ -138,6 +138,34 @@ export function assembleLayerSelectionMarks(model: LayerModel, marks: any[]): an
...
+      const vname = varName(selCmpt.name);
+      store = stringValue(vname + STORE);
+    }
+  });
+
+  if (hasLegend) {
+    const newValue = value ? value : {opacity: {value: 0.9}};
+    // To do : Add test case for legends and symbols
+    newValue.opacity = [
+      {test: `!(length(data(${store}))) || vlSelectionTest(${store}, {${field}: datum.value})`, ...newValue.opacity},
```
arvind: I recommend using the `assembleSelectionPredicate` method to generate this test string, so that we're not duplicating important logic. That method may require some minor modification to support the custom second argument to `vlSelectionTest`.

src/compile/legend/parse.ts
```diff
@@ -14,16 +14,20 @@ import {
...
 import {parseGuideResolve} from '../resolve';
+import {assembleLegendSelection} from '../selection/assemble';
 import {defaultTieBreaker, Explicit, makeImplicit, mergeValuesWithExplicit} from '../split';
 import {UnitModel} from '../unit';
 import {LegendComponent, LegendComponentIndex} from './component';
 import * as encode from './encode';
 import * as properties from './properties';
 
+// For MVP, later remove global variable and create new types
+export let selectionOnChannel: any = [];
```
djbarnwal: @arvind What would be the best way to send this information to a selection transform function. We can create this list using the `model` param but `has` method of a transform does not have a `model` argument to it. In theory, we can add that argument to the function. Moreover, we can also get this list when the model is being parsed by creating a new method at `model.ts`. But, both of these methods look like a long stretch to me. I think I am missing something of a very simple solution here. Any thoughts?

PR #4321, opened 2018-12-04
Don't import util from vega directly.  Otherwise our users can no longer use `vega-lib`.

kanitw:
cc: @Saba9 @domoritz @arvind 

We should do #4322 to prevent this.


Saba9:
Oops, I'll keep an eye on that from now on. Thanks for fixing it.

kanitw:
@Saba9  -- yeah I actually didn't notice we have to worry about this until this release, so it's not really your fault. (I'm just cc-ing so we're all aware of this. :) )

PR #4325, opened 2018-12-04
Default gradient legend length

kanitw:


src/compile/legend/properties.ts
```diff
@@ -20,6 +21,34 @@ export function clipHeight(scaleType: ScaleType) {
...
+      return gradientLengthSignal(model, 'width', gradientHorizontalMinLength, gradientHorizontalMaxLength);
+    }
+  } else {
+    return gradientLengthSignal(model, 'height', gradientVerticalMinLength, gradientVerticalMaxLength);
+  }
+}
+
+function gradientLengthSignal(model: Model, sizeType: 'width' | 'height', min: number, max: number) {
+  const sizeSignal = model.getSizeSignalRef(sizeType).signal;
+  return {signal: `min(max(${sizeSignal}, ${min}), ${max})`};
```
domoritz: Use `clamp` instead.

build/vega-lite-schema.json
```diff
@@ -5554,6 +5554,14 @@
           "$ref": "#/definitions/Orientation",
           "description": "The default direction (`\"horizontal\"` or `\"vertical\"`) for gradient legends.\n\n__Default value:__ `\"vertical\"`."
         },
+        "gradientHorizontalMaxLength": {
+          "description": "Max legend length for a horizontal gradient.\n\n__Default value:__ `200`",
```
domoritz: This is a bit misleading since the user can set a gradient length that is outside these bounds, no?

Issue #4326, opened 2018-12-04
Revise color precedence


kanitw:
Previously I defined that color would be ignored when `fill` or `stroke` is specified. 

However, this has lead to so many confusions in my own use cases.  

We should revise if this is really the right model.

kanitw:
[...]

PR #4328, opened 2018-12-05
Only ignore `config.color` when `config.mark.fill` or `config.mark.stroke` is specified.

kanitw:
Ignoring it when markDef.fill/stroke  or encoding.fill/stroke are specified makes it very confusing. 

Part of #4326


kanitw:
I think this solves one of the problem, but not all of them. 

That said, it's small, non-disruptive change that can be reviewed first.

PR #4331, opened 2018-12-05
Default legend direction

kanitw:
## Top/Bottom

image

## Left/Right 

image

## Inner (e.g., top-right)

Compact layout (like Tableau)

image

image


kanitw:
cc: @jheer -- not sure if this logic should go into Vega instead. But we could argue that Vega doesn't do this sort of "smart default" things.

Issue #3761, opened 2018-05-18
Using chai with jest produces TERRIBLE diff for deepEqual


kanitw:
(It doesn't compare the diff.) 

~~This might be one solution. https://www.npmjs.com/package/chai-jest-diff~~  Try and it doesn't work

kanitw:
This basically makes the diff no longer debuggable...

domoritz:
The solution is to use `expect(...).toEqual(...)` and we started doing the conversion step by step. I think we can close this issue unless you want us to do this in one big refactor step.

kanitw:
Let's close.  Refactoring in one big step is a pain and I don't want to volunteer anyone for it. lol

PR #4332, opened 2018-12-05
Use expect

domoritz:
Towards #3761


domoritz:
This does not fix everything but a good part.

PR #4334, opened 2018-12-05
Make default tooltip respect the order of channels in `encoding`

kanitw:
By simply using `for..of` in `fieldDefs()` that default tooltip generator depends on.


kanitw:
Motivation: it's weird that default tooltip for pre-binned histogram have count in the middle 

image

kanitw:
With this PR: 

image

domoritz:
Lgtm. Why do none of the examples change?

kanitw:
Weird

kanitw:
Rebased -- now it's generating.

PR #4340, opened 2018-12-06
Add rect/barBandPaddingInner/Outer

### LINKED ISSUE #4333: Separate default bandPaddingInner/Outer for rect and bar
Currently, band scale's paddingInner/Outer are determined by `config.scale.bandPaddingInner/Outer` for all marks. 

However, this leads to poor default for `rect` heatmap since we have non-zero padding inner for bars, but heatmap looks way better without paddingInner. 

Thus, I think we should separate `barBandPaddingInner/Outer` from `rectBandPaddingInner/Outer` (and use the current value for the former and zero for the latter.)

The question is where should this config be:

a) `config.scale.bar/rectBandPaddingInner/Outer` 

b) `config.bar/rect.bandPaddingInner/Outer`

I think a) is better since this ultimately affects scale (though b) is a bit more concise). 
However, we also need to retain `config.scale.bandPaddingInner/Outer`, which will be both `undefined` by default and will use value based on its bar/rect variants. 

@domoritz  any thoughts?

kanitw:
Fix #4333


kanitw:
image


I don't know why the example script doesn't work ...

domoritz:
So muchy better!

PR #4341, opened 2018-12-06
Try updating latest node version to see if it would fix the example outdated problem

kanitw:
cc: @domoritz


kanitw:
I'm merging this for now since this is breaking many subsequent example generations.

PR #4351, opened 2018-12-07
Change default scale `rangeStep` to `20` 

### LINKED ISSUE #4348: Revise default rangeStep?
In the old days, we used point scale for bars and somehow found that 21 (odd number) make the bar looks nicer. 

However, now that we use band scale for bars (a more sensible default). Right now we still use rangeStep = 21, but need to use the fractional padding. Thus, currently the bars boundary do not fall on exact pixel.  

We could revise the rangeStep to be 20, so that applying paddingInner = 0.1 makes the bar boundaries fall on the exact pixel.

On a related note:

1) If we want to use stroke to highlight selection like in the [interactive_bar_select_highlight](https://vega.github.io/editor/#/examples/vega-lite/interactive_bar_select_highlight).  
Changing the `barBandPaddingInner` to 0.15 or 0.2 gives more breathing room for the strokes by default. 

2) (minor) we may want to change the default barBandWidth for point scale to match the band scales to make it consistent for layering.  

Any thoughts?

kanitw:
and default tickSize to be `3/4` of `rangeStep` (so it's nicely `15` by default)

Fix #4348


domoritz:
Just wondering, what do you think about going to 24 pixels instead? The reason is that 24 can be divided by 2,3,4,6,8,12 while 20 can only be divided by 2,4,5,10. Moreover, I think that most people look at single charts and 20 is pretty small for marks that do not bar.

kanitw:
Nah. Range step does not affect mark size except bar and rect. Plus 20
means you can fit more on a single screen for high cardinaility axis.

kanitw:
20 is also a less drastic change from 21.

domoritz:
20

screen shot 2018-12-06 at 19 49 13


24

screen shot 2018-12-06 at 19 49 04

kanitw:
Actually, 24 is not great for decimal fraction padding. Better use number
divisible by 10 such as 20.

domoritz:
> Actually, 24 is not great for decimal fraction padding. Better use number
divisible by 10 such as 20.

I think a) you can specify 1/6 as a fraction instead and b) 24 allows more possible divisions that fall on full pixels so I think 24 is preferred over 20.

domoritz:
Unfortunately, humanity is not ready to get rid of the decimal system.

kanitw:
For context

> we can’t specify a fraction unless we wanna do a `{signal: ‘1/6’}`

Also I think people often design UI components size to be divisible by 10 (rather than 24).

PR #4359, opened 2018-12-11
Use jest expect instead of chai assert for all tests

domoritz:
I could not convert the runtime tests as they don't support jest.


kanitw:
Great efforts.  I'll just merge since it's gonna be a pain to check line-by-line :)

PR #4360, opened 2018-12-11
Use signal type from Vega typings

domoritz:


src/compile/selection/transforms/transforms.ts
```diff
@@ -1,26 +1,27 @@
...
+import scales from './scales';
+import toggle from './toggle';
+import translate from './translate';
+import zoom from './zoom';
 
 export interface TransformCompiler {
   has: (selCmpt: SelectionComponent | SelectionDef) => boolean;
   parse?: (model: UnitModel, def: SelectionDef, selCmpt: SelectionComponent) => void;
-  signals?: (model: UnitModel, selCmpt: SelectionComponent, signals: VgSignal[]) => VgSignal[];
+  signals?: (model: UnitModel, selCmpt: SelectionComponent, signals: VgSignal[]) => Signal[]; // the output can be a new or a push signal
```
kanitw: Why is this one the only one that uses Signal, rather than VgSignal?

domoritz: Because VgSignal is only a `NewSignal` type but it generates a `PushSignal` type.

domoritz: I think we are causing more confusion with the Vg prefix than necessary. Instead, we could use `NewSignal` instead of `Signal`. Vega-Lite does not have signals and so we don't need a separate type. Moreover, if we add signals, they will be the same as Vega's.

kanitw: Sounds good, esp. I'm becoming more and more convinced that we should add Signal in VL, thought not 100% yet.

domoritz:
Done. We now use NewSignal from Vega.

PR #4362, opened 2018-12-11
Improve typings

domoritz:


src/sort.ts
```diff
@@ -1,9 +1,17 @@
 import {AggregateOp} from 'vega';
 import {isArray} from 'vega-util';
 import {DateTime} from './datetime';
-import {VgComparatorOrder} from './vega.schema';
 
-export type SortOrder = VgComparatorOrder | null;
+export type SortOrder = 'ascending' | 'descending' | null;
+export type Compare =
```
kanitw: The first one is a `SortField`.  

Should `Compare = SortField | SortFields`? (Instead of redeclaring them?)

src/vega.schema.ts
```diff
@@ -1,17 +1,17 @@
 import {
   AggregateOp,
-  Align,
+  Align as VgAlign,
```
kanitw: Are these all different in Vega? (If they are the same, we don't need to add Vg prefix?)

domoritz: No, they are the same. I will change this then.

src/sort.ts
domoritz: Why is `ascending` and `descending` a sort field? It's a way to order.

domoritz:
Fixed and rebased

src/sort.ts
kanitw: This is not related to SortOrder at all.  Here you have: 

```
Compare =
  | {
      field: string;
      order?: SortOrder;
    }
  | {
      field: string[];
      order?: (SortOrder)[];
    };
```

kanitw: Ok. Let's change this so we can merge.

domoritz: On it

domoritz:
I removed the Compare type.

domoritz:
Ready @kanitw

PR #4365, opened 2018-12-12
Format markdown and css with prettier

domoritz:
Depends on https://github.com/prettier/prettier/issues/5631


.prettierrc.json
```diff
@@ -1,15 +1,25 @@
...
         "parser": "typescript",
         "singleQuote": true,
         "bracketSpacing": false,
         "arrowParens": "avoid",
         "trailingComma": "none"
       }
+    }, {
+      "files": "*.md",
+      "options": {
+        "proseWrap": "never"
```
kanitw: "bracketSpacing": false. ?

domoritz: I fixed this by setting the same format for everything. Does this look good to you?

PR #4377, opened 2018-12-18
Change default title for Count to "Count of Records" instead of "Number of Records"

kanitw:
Originally we follow Tableau for this.  However, users directly type "count" in Vega-Lite while users directly drag the "Number of Records" pill around in Tableau.

So having "Number of Records" as the default can confuse users. 

~~Plus, count is shorter, which works better for smaller plots.~~


kanitw:
Sure.  Note that since this is configurable, people who prefer the old default can override the configs.

jheer:
Other aggregate measures note both the operation and the input (e.g., "Sum of Population"). In comparison, "Count" all by itself might be misleading to some chart readers -- who may not be the same as chart creators! Count of what? "Count of Records" or "Number of Records" is safer in this regard. We didn't just follow Tableau blindly, we did it because the reasoning for the current default is sensible and consistent with other operations.

I'm also generally against changes that don't have strong motivation to avoid unnecessary churn. This change could potentially break downstream test outputs, for example. Moreover, if people really find changing the config to be as easy as you seem to imply, they can always change the axis title to "Count"...

kanitw:
Note that I have seen multiple people get confused for a bit why the "count" operation produces the label "Number of Records" as a title rather than something that says "count" for a few reasons:

- Unlike other operations, the default titles always include the operation names.

- "Number of Records" as the default requires additional thinking that count = number of records)

- Given Vega-Lite doesn't use the terminology "records" anywhere else.  People also get confused for a bit what we mean "records".  On the other hand, Tableau uses the terminology more consistently. Plus, in Tableau and Voyager, this isn't a problem since users directly drag pills with the label "Number of Records".

kanitw:
I refined the PR to change to "Count of Records" instead. That said, this still doesn't solve the following issue: 

> - Given Vega-Lite doesn't use the terminology "records" anywhere else. People also get confused for a bit what we mean "records".

kanitw:
From a quick survey on the web I see:
- ggplot2 uses "count"
- A lot of paper figures use "Frequency".  (More of them don't even label count!)
- Newsrooms often use "Number of X" where X is the item name such as players. 

image

(From https://www.google.com/url?sa=i&source=images&cd=&cad=rja&uact=8&ved=2ahUKEwiZvq7K8ajfAhVmslQKHQCbCe8Qjxx6BAgBEAI&url=https%3A%2F%2Ffivethirtyeight.com%2Ffeatures%2Fwhat-our-nba-projections-got-right-and-wrong-last-season%2F&psig=AOvVaw3bcYscqlxR85Rx4MKJlglk&ust=1545204624540559)


- Newsroom also often show histogram label in %
image

jheer:
I agree that "Count of Records" is more straightforward / consistent than "Number of Records"!

domoritz:
Does this work with transform extraction from @Saba9?

kanitw:
> Does this work with transform extraction from @Saba9?

Yes, I did fix the unit test for it. 

https://github.com/vega/vega-lite/pull/4377/files#diff-4fecb77be8d42e352cbe3327b4b49aae

PR #4379, opened 2018-12-18
Format HTML, layouts, and includes

chanwutk:


domoritz:
@kanitw I suggested these changes so can you merge?

domoritz:
Let's not do this for now since prettier does not work with templates.

PR #4383, opened 2018-12-20
Better error bar and error band tooltip

### LINKED ISSUE #4261: Tooltip for box plots is weird
It currently only shows only a few variables.

image

1) the "population" title is a bit wrong
2) only two fields are shown on hover although box plots are basically visual version of a 5-number summary.  

Ideally, it should explicitly says median_people, q3_people, etc.  (or even better, show "Q3 of people", "Median of people".)

chanwutk:
fix #4261

### DIFF
--- a/examples/compiled/errorband_2d_horizontal_color_encoding.vg.json
+++ b/examples/compiled/errorband_2d_horizontal_color_encoding.vg.json
@@ -19,9 +19,17 @@
         {
           "type": "aggregate",
           "groupby": ["year_Year"],
-          "ops": ["ci0", "ci1"],
-          "fields": ["Miles_per_Gallon", "Miles_per_Gallon"],
-          "as": ["lower_Miles_per_Gallon", "upper_Miles_per_Gallon"]
+          "ops": ["ci0", "ci1", "mean"],
+          "fields": [
+            "Miles_per_Gallon",
+            "Miles_per_Gallon",
+            "Miles_per_Gallon"
+          ],
+          "as": [
+            "lower_Miles_per_Gallon",
+            "upper_Miles_per_Gallon",
+            "center_Miles_per_Gallon"
+          ]
         }
       ]
     }
@@ -39,7 +47,7 @@
           "orient": {"value": "vertical"},
           "fill": {"value": "black"},
           "tooltip": {
-            "signal": "{\"Ci1 of Miles_per_Gallon\": format(datum[\"upper_Miles_per_Gallon\"], \"\"), \"Ci0 of Miles_per_Gallon\": format(datum[\"lower_Miles_per_Gallon\"], \"\"), \"Year (year)\": timeFormat(datum[\"year_Year\"], '%Y')}"
+            "signal": "{\"Ci1 of Miles_per_Gallon\": format(datum[\"upper_Miles_per_Gallon\"], \"\"), \"Ci0 of Miles_per_Gallon\": format(datum[\"lower_Miles_per_Gallon\"], \"\"), \"Mean of Miles_per_Gallon\": format(datum[\"center_Miles_per_Gallon\"], \"\"), \"Year (year)\": timeFormat(datum[\"year_Year\"], '%Y')}"
           },
           "x": {"scale": "x", "field": "year_Year"},
           "y": {"scale": "y", "field": "lower_Miles_per_Gallon"},
@@ -60,7 +68,7 @@
         "update": {
           "stroke": {"value": "black"},
           "tooltip": {
-            "signal": "{\"Ci1 of Miles_per_Gallon\": format(datum[\"upper_Miles_per_Gallon\"], \"\"), \"Ci0 of Miles_per_Gallon\": format(datum[\"lower_Miles_per_Gallon\"], \"\"), \"Year (year)\": timeFormat(datum[\"year_Year\"], '%Y')}"
+            "signal": "{\"Ci1 of Miles_per_Gallon\": format(datum[\"upper_Miles_per_Gallon\"], \"\"), \"Ci0 of Miles_per_Gallon\": format(datum[\"lower_Miles_per_Gallon\"], \"\"), \"Mean of Miles_per_Gallon\": format(datum[\"center_Miles_per_Gallon\"], \"\"), \"Year (year)\": timeFormat(datum[\"year_Year\"], '%Y')}"
           },
           "x": {"scale": "x", "field": "year_Year"},
           "y": {"scale": "y", "field": "lower_Miles_per_Gallon"},
@@ -80,7 +88,7 @@
         "update": {
           "stroke": {"value": "black"},
           "tooltip": {
-            "signal": "{\"Ci1 of Miles_per_Gallon\": format(datum[\"upper_Miles_per_Gallon\"], \"\"), \"Ci0 of Miles_per_Gallon\": format(datum[\"lower_Miles_per_Gallon\"], \"\"), \"Year (year)\": timeFormat(datum[\"year_Year\"], '%Y')}"
+            "signal": "{\"Ci1 of Miles_per_Gallon\": format(datum[\"upper_Miles_per_Gallon\"], \"\"), \"Ci0 of Miles_per_Gallon\": format(datum[\"lower_Miles_per_Gallon\"], \"\"), \"Mean of Miles_per_Gallon\": format(datum[\"center_Miles_per_Gallon\"], \"\"), \"Year (year)\": timeFormat(datum[\"year_Year\"], '%Y')}"
           },
           "x": {"scale": "x", "field": "year_Year"},
           "y": {"scale": "y", "field": "upper_Miles_per_Gallon"},
--- a/examples/specs/normalized/layer_point_errorbar_ci_normalized.vl.json
+++ b/examples/specs/normalized/layer_point_errorbar_ci_normalized.vl.json
@@ -21,7 +21,8 @@
         {
           "aggregate": [
             {"op": "ci0", "field": "yield", "as": "lower_yield"},
-            {"op": "ci1", "field": "yield", "as": "upper_yield"}
+            {"op": "ci1", "field": "yield", "as": "upper_yield"},
+            {"op": "mean", "field": "yield", "as": "center_yield"}
           ],
           "groupby": ["variety"]
         }
@@ -47,6 +48,11 @@
                 "field": "lower_yield",
                 "type": "quantitative",
                 "title": "Ci0 of yield"
+              },
+              {
+                "field": "center_yield",
+                "type": "quantitative",
+                "title": "Mean of yield"
               },
               {"field": "variety", "type": "ordinal"}
             ]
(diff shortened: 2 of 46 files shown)


src/compositemark/common.ts
```diff
@@ -23,6 +31,31 @@ export type GenericCompositeMarkDef<T> = GenericMarkDef<T> &
     clip?: boolean;
   };
 
+export interface CompositeMarkTooltipSummary {
+  fieldPrefix: string;
+  titlePrefix: string;
+}
+
+export function getCompositeMarkTooltip(
```
kanitw: What is `tooltipSummaryMap` btw?  Why `-Map` suffix?   Why not `tooltipSummaries`?

src/compositemark/errorbar.ts
```diff
@@ -107,9 +110,9 @@ export function normalizeErrorBar(
     ...outerSpec,
     transform,
     layer: [
-      ...makeErrorBarPart('ticks', tick, 'lower'),
-      ...makeErrorBarPart('ticks', tick, 'upper'),
-      ...makeErrorBarPart('rule', 'rule', 'lower', 'upper')
+      ...makeErrorBarPart('ticks', tick, 'lower', null, tooltipEncoding),
```
kanitw: I think it's worth making `makeErrorBarPart` take parameters as an object since this object this method is getting a bit hard at this point. 

(I have wanted to change this earlier, but haven't got the chance yet.)

src/compositemark/errorbar.ts
```diff
@@ -340,10 +349,29 @@ export function errorBarParams<
...
+export function errorBarTooltipEncoding(
+  continuousAxisChannelDef: PositionFieldDef<string>,
+  encodingWithoutContinuousAxis: Encoding<string>
+): Encoding<string> {
+  const tooltopSummaryMap: CompositeMarkTooltipSummary[] = [
+    {fieldPrefix: 'upper', titlePrefix: 'Upper error'},
+    {fieldPrefix: 'lower', titlePrefix: 'Lower error'}
+  ];
+
+  const tooltip: TextFieldDef<string>[] = getCompositeMarkTooltip(
```
kanitw: Why shouldn't `getCompositeMarkTooltip` return the output wrapped with `{tooltip: ...}` already? 

Also, I wonder if `errorBarTooltipEncoding` needs to be a separate method at all?   (Same for box)

It seems like you only need to define `errorBarTooltipSummaries()` (I think the name `-Map` is confusing) and just call `getCompositeMarkTooltip` directly.

src/compositemark/errorbar.ts
```diff
@@ -340,10 +349,29 @@ export function errorBarParams<
...
+    tooltipEncoding
   };
 }
 
+export function errorBarTooltipEncoding(
+  continuousAxisChannelDef: PositionFieldDef<string>,
+  encodingWithoutContinuousAxis: Encoding<string>
+): Encoding<string> {
+  const tooltopSummaryMap: CompositeMarkTooltipSummary[] = [
+    {fieldPrefix: 'upper', titlePrefix: 'Upper error'},
```
kanitw: Title "Upper/lower error" is confusing -readers won't get whether the error bar encodes Stdev, Stderr, CI. (It's basically confusing.) 

For pre-aggregated case, better not provide special title and use original field name. 

For aggregating case, we should provide information about what error bar presents.  For example, we could  say `mean(x)+stdev(x) = ...`, `mean(x) - stdev(x)` = ...Alternatively, we could just directly set `titlePrefix: titlecase(aggregate)` in the tooltip and show the error instead of upper/lower (so we get `Stdev of X`) and reduce the complex +/- thing. 

cc: @domoritz  Any thoughts

kanitw: It might be worth looking at how other systems display tooltips for error bars too.  

I suspect that some might even says  "10 ± 5".  So another alternative is 
`Mean of x ± Stdev of X:   10 ± 5`.   

Right now I think this might be the best, but would require some custom signal that we couldn't do without a signal / expression support yet.

test/compositemark/errorbar.test.ts
```diff
@@ -62,8 +62,8 @@ describe('normalizeErrorBar with raw data input', () => {
             y2: {field: 'upper_people', type: 'quantitative'},
             x: {field: 'age', type: 'ordinal'},
             tooltip: [
-              {field: 'upper_people', title: 'Upper error of people', type: 'quantitative'},
-              {field: 'lower_people', title: 'Lower error of people', type: 'quantitative'},
+              {field: 'upper_people', title: 'mean + stderr of people', type: 'quantitative'},
```
kanitw: We use Titlecase for standard title names.  (e.g., `"Mean of x"`)

test/compositemark/errorbar.test.ts
```diff
@@ -668,8 +668,8 @@ describe('normalizeErrorBar with aggregated upper and lower bound input', () =>
             y2: {field: 'upper_people', type: 'quantitative'},
             x: {field: 'age', type: 'ordinal'},
             tooltip: [
-              {field: 'upper_people', title: 'Upper error of people', type: 'quantitative'},
-              {field: 'lower_people', title: 'Lower error of people', type: 'quantitative'},
+              {field: 'lower_people', title: 'people', type: 'quantitative'},
```
kanitw: People and people2 is worse than upper/lower error

examples/specs/normalized/layer_point_errorbar_ci_normalized.vl.json
```diff
@@ -41,12 +41,12 @@
               {
                 "field": "upper_yield",
                 "type": "quantitative",
-                "title": "Upper error of yield"
+                "title": "ci1 of yield"
```
kanitw: ci0 / ci1 are not naturally understood operations (very specific to Vega [for "The lower boundary of the bootstrapped 95% confidence interval of the mean field value."](https://vega.github.io/vega/docs/transforms/aggregate/#ops))

Perhaps, "Mean of <fieldname> (Upper/Lower bound of 95% CI)". 
(Though I think it's still a bit too long.)  

@domoritz  Any suggestion?

kanitw: Actually, if users directly uses ci, they gets this title as well. So perhaps, we can keep it like this in this PR.  

(But make sure that this title thing is generated from a common code (by calling [`title()` in `fielddef.ts`](https://github.com/vega/vega-lite/blob/master/src/fielddef.ts#L524)).  

We can then consider if we should revise default title for `ci0`/`ci1`.

test/compositemark/errorbar.test.ts
chanwutk: This is ranged error bar case, should we use the original field name?

kanitw: The axis title should just be the field name (e.g., `people`). 

However, the title in the tooltip should be quite descriptive. 
(What does the error exactly represent? -- ci0/ci1, mean+stdev/err?)

test/compositemark/errorbar.test.ts
chanwutk: Would you prefer `Mean(people) + Stderr(people)` or `Mean + Stderr of people`?

kanitw:
Mean + stderr of people, with small s — consistent with other titles

kanitw:
Use original field names then.

src/compositemark/errorbar.ts
chanwutk: Can you clarify this a bit more?

kanitw: See 86eae67bb905be5b6582bf27d2e60f6635918862 for example.  I did the same thing to `tickCount()` there so it will enforce us to include parameter names.

Right now this method is extremely hard to read.  It's hard to remember what the first 'rule' and the second 'rule' are for. By forcing them to be an object, we can understand the parameter name right away.

Just like in Python `f(x=3,y=4,z=5,opt={})` is clearer than `f(3,4,5,{})` as we know that the first parameter is `x`, second is `y` and so on.  For method with 1-2 parameters, this isn't so important. But once we have several, it's becoming quite hard to read.

kanitw: https://basarat.gitbooks.io/typescript/content/docs/tips/functionParameters.html

kanitw:
Rebase?

domoritz:
This seems unintended, no? 

screen shot 2018-12-26 at 13 33 02

chanwutk:
I will fix it.

examples/compiled/errorband_2d_horizontal_color_encoding.vg.json
```diff
@@ -39,7 +39,7 @@
           "orient": {"value": "vertical"},
           "fill": {"value": "black"},
           "tooltip": {
-            "signal": "{\"Miles per Gallon (95% CIs)\": format(datum[\"lower_Miles_per_Gallon\"], \"\"), \"upper_Miles_per_Gallon\": format(datum[\"upper_Miles_per_Gallon\"], \"\"), \"Year (year)\": timeFormat(datum[\"year_Year\"], '%Y')}"
+            "signal": "{\"Ci1 of Miles_per_Gallon\": format(datum[\"upper_Miles_per_Gallon\"], \"\"), \"Ci0 of Miles_per_Gallon\": format(datum[\"lower_Miles_per_Gallon\"], \"\"), \"Year (year)\": timeFormat(datum[\"year_Year\"], '%Y')}"
```
domoritz: Why "Ci" instead of "CI"? It stands for Confidence Interval.

kanitw: As commented above, we should systematically revise title of ci0/ci1 for both composite mark and general operation in a separate PR.

domoritz: Okay, I'll let you make the call on this or then.

test/compositemark/errorbar.test.ts
```diff
@@ -1121,4 +1272,61 @@ describe('normalizeErrorBar with aggregated error input', () => {
...
+    );
+
+    const layer = isLayerSpec(outputSpec) && outputSpec.layer;
+    expect(layer).toBeTruthy();
+    for (const unitSpec of layer) {
+      const encoding = isUnitSpec(unitSpec) && unitSpec.encoding;
+      expect(encoding).toBeTruthy();
+      const tooltip = encoding.tooltip;
+      expect(tooltip).toEqual([
+        {field: 'upper_people', title: 'people + people_error1', type: 'quantitative'},
```
kanitw: Why is there no tooltip for `y` if `y`, `yError`, and `yError2` are all provided?

Also, why is extent specified when users directly specify yError/yError2?

kanitw:
When we know the center, I think the tooltip should show the center as well. 

image

(Otherwise, the audience has to manually calculate the center in their head.)

chanwutk:
Since we have to aggregate the center value for error bar with `stdev` and `stderr`, would it be useful to show the center value of the bar in tooltip as well.
And for the errorbar with `ci` and `iqr`, should we also aggregate for mean and median and include them in tooltip?

PR #4385, opened 2018-12-20
Kw/init selections

kanitw:
A copy of #4139, but without the refactoring in https://github.com/vega/vega-lite/pull/4139/commits/027708f11a71a90a76d6dc7d41e4c2dc474a2a58.

(I wanna see the real diffs that do not involve refactoring)


kanitw:
We no longer need this as I fix the original branch instead.

PR #4390, opened 2018-12-22
Refactor Axis

kanitw:


kanitw:
Many branches on the plane depend on this simple refactor. Feel free to send follow up comments and I'll fix.

PR #4392, opened 2018-12-22
Add docs to spec.ts and move TopLevelSpec to the top

kanitw:
Also fix incorrect casting in model.ts


kanitw:
I'll combine this into #4398.

PR #4394, opened 2018-12-22
 Revise default tickCount rules for binned field

kanitw:
~~Merge https://github.com/vega/vega-lite/pull/4390 first~~


kanitw:
I wonder if I should bound maximum default tickCount to width / c where c is some constant such as `4`.  (At most 1 tick every 4 pixel)

examples/compiled/bar_aggregate_count.vg.json
```diff
@@ -101,7 +101,9 @@
       "title": "precipitation (binned)",
       "labelFlush": true,
       "labelOverlap": true,
-      "tickCount": {"signal": "ceil(width/20)"},
+      "tickCount": {
+        "signal": "(domain('x')[1] - domain('x')[0]) / bin_maxbins_10_precipitation_bins.step + 1"
```
domoritz: Why not `(bin_maxbins_10_precipitation_bins.stop - bin_maxbins_10_precipitation_bins.stop) / bin_maxbins_10_precipitation_bins.step`?

kanitw: I think you mean stop-start. Sure we can do that. 

In any case, we need plus 1 for sure.

kanitw: Actually, that makes the code a bit more complicated because for prebinned, we might have the bin signal and only have the step.  

See axis/properties.ts

image

domoritz:
What's the goal of the PR? We previouly limited the tick count to clean up charts. 

Before:
image

After:
image

kanitw:
The number of ticks now align with number of bins.

As said above, we can limit the number of ticks per pixel. But in general case where u dont have crazy number of bins, this clearly facilitate chart reading by having ticks matching the bins.

kanitw:
image

This is clearly better, for example.  The old limit that we impose was arbitary.

domoritz:
> The number of ticks now align with number of bins.

Wouldn't you get that by removing the tick count?

I feel what you want is a higher tick count.

kanitw:
> Wouldn't you get that by removing the tick count?

No — how would the axis know the bin count if you dont provide it.

kanitw:
> I feel what you want is a higher tick count.

No i want five+1 ticks for five bins

domoritz:
Close in favor of #4420?

PR #4395, opened 2018-12-22
Change default continuousPadding

kanitw:


kanitw:
This is probably a bad idea. Let's not do it.

PR #4399, opened 2018-12-24
Remove unused methods in `spec/`

kanitw:
Please merge #4530 first.


src/spec/index.ts
```diff
@@ -109,39 +107,3 @@ export function isStacked(spec: TopLevel<FacetedCompositeUnitSpec>, config?: Con
   }
   return false;
 }
-
-/**
- * Takes a spec and returns a list of fields used in encoding
- */
-export function usedFields(spec: NormalizedSpec): string[] {
```
domoritz: @Saba9 uses this

kanitw: Where?  In a separate repo?  (I don't see any in VL)

domoritz: Hmm, I can't find it. @Saba9?

kanitw: so?

domoritz: We are not going to use this anymore if we decide to drop the transform extraction in Vega-Lite (more on this on Friday).

kanitw: 1) As said in multiple places, we can't drop `extractTransformsFromEncoding`. 

2) `extractTransformsFromEncoding` doesn't call this, so why is it relevant?

domoritz: @saba calls it in the new version of https://github.com/vega/vega-lite-transforms2sql if I understood him correctly today.

Saba9: Right. We could choose to remove this method down the line if we do decide to abandon extracting transforms from Vega-Lite for use with alternative backends while keeping `extractTransformsFromEncoding` for composite marks. But as @domoritz said it will be used in vega-lite-transforms2sql until that point.

domoritz: Let's leave it in until we make the call.

domoritz: Can we move this function to `vega-lite-transforms2sql`?

kanitw: Moving it there is probably a good idea.

domoritz: @Saba9 Please move it.

Saba9: Sounds good. Will do

domoritz:
LGTM

kanitw:
Manually merged into #4530

PR #4405, opened 2018-12-25
Improve default point size for binned scatterplot + Introduce SignalRefComponent

kanitw:
This is a more complete version of https://github.com/vega/vega-lite/pull/4391.

We can later introduce a helper method that allows us to generate SignalRefComponent with manually listing signal names. However, I think it's beyond the scope of this PR.


kanitw:
The output that you see is clearly better -- the maximum size in the old output can overlap like crazy whereas the maximum size in the new output would never cause overlap.  

However, the new output highlights one problem that occurs when we no longer filter null values.  Basically, null values are not rendered on the bin scale, but the size scale domain still accounts for the maximum size.  Thus, the new output is not great.

I don't think fixing the null issue is the concern for this PR.

Meanwhile, I think the size can be slightly bigger than the one in this PR, so I have a follow up PR https://github.com/vega/vega-lite/pull/4405 to change the formula a little bit.

domoritz:
Lol, you reference #4405 from this PR.

src/compile/scale/range.ts
```diff
@@ -371,9 +417,13 @@ function sizeRangeMax(mark: Mark, xyRangeSteps: number[], config: Config) {
         return config.scale.maxSize;
       }
 
-      // FIXME this case totally should be refactored
+      // TODO: leverage Vega-Expression to write this formula only once
       const pointStep = minXYRangeStep(xyRangeSteps, scaleConfig);
-      return (pointStep - 2) * (pointStep - 2);
+      if (pointStep instanceof SignalRefComponent) {
+        return pointStep.map(expr => `pow(${expr} - 2, 2) `);
```
domoritz: Why -2? I think it's fine for marks to touch in the most extreme case. For circles it may even be okay to overlap (although I think you don't like it).

kanitw: This was an old decision — a follow up PR changes - 2 to 95% which is -1 for range step = 20

PR #4407, opened 2018-12-25
Make `build-example` throw `exit 1` when there is an empty SVG

kanitw:
by modifying `build-examples.sh`


kanitw:
This doesn't work ...

PR #4408, opened 2018-12-26
Make `build-example` throw `exit 1` when there is an empty SVG

kanitw:


kanitw:
Hmm this approach doesn't work well as it will stop other files from compiling.

PR #4412, opened 2018-12-26
Clean up set

### LINKED ISSUE #4406: Infinitely long output in the Travis script
I don't know why there are so many lines of RVM nonsense in the Travis log after the `lint` command.  

(Our real log is like ~ 2k lines, but there are >8k lines of gibberish output!)

domoritz:
Fixes #4406


domoritz:
This seems to fix it.

PR #4413, opened 2018-12-26
Rename applications to ecosystem

domoritz:


kanitw:
I think the term application is confusing, but I'm not sure if ecosystem is better.

domoritz:
It's definitely less confusing.

domoritz:
When I see applications, I think of "apps" or "tools" rather than general "usages".

domoritz:
https://parquet.apache.org/ uses "adopters".

domoritz:
Another option is to have top level "tools" (for users) and "adopters" (to show how awesome we are).

PR #4420, opened 2018-12-29
Adjust formula for bin tick count

kanitw:
Merge https://github.com/vega/vega-lite/pull/4418 first.

This might be a better alternative to https://github.com/vega/vega-lite/pull/4394.

(I think we should later extract these `10` and `40` as constants.)


examples/compiled/interactive_concat_layer.vg.json
```diff
@@ -255,7 +255,7 @@
           "grid": false,
           "title": "Rotten_Tomatoes_Rating (binned)",
           "labelOverlap": true,
-          "tickCount": {"signal": "ceil(concat_0_height/20)"},
+          "tickCount": {"signal": "ceil(concat_0_height/10)"},
```
domoritz: I don't think this is better

screen shot 2018-12-28 at 22 09 36

examples/compiled/layer_circle_independent_color.vg.json
```diff
@@ -212,7 +212,7 @@
       "title": "Acceleration (binned)",
       "labelFlush": true,
       "labelOverlap": true,
-      "tickCount": {"signal": "ceil(width/20)"},
+      "tickCount": {"signal": "ceil(width/10)"},
```
domoritz: I think this is just busier

screen shot 2018-12-28 at 22 10 29

domoritz:
This is definitely the clean approach but I'm not yet convinced that every 10 pixels is better than every 20 pixels. Axes get quite busy after the change.

kanitw:
It’s clearly better man. The new one facilitates readers to count number of
bins easily and see the bin boundary. Basically 20 is too high esp when
there are 12-14 bins.

domoritz:
If your chart is larger, the 10px between ticks can be quite small. I feel it adds clutter. I can usually interpolate without problems even if I don't have e ticks every 10px.

kanitw:
>  I can usually interpolate without problems even if I don't have e ticks every 10px.

You can, but it takes quite more effort to read it. Plus, my colleagues have mentioned that they get confused why we leave so much space between ticks instead of adding ticks between them. 

I think clutter is less important than having the ability the read the value right away.  I think it's too dense if we have more than 1 tick per 10 pixel though.

kanitw:
@domoritz Any idea why Travis fails? I don't see any error in the log.

kanitw:
@domoritz Any idea why Travis fails? I don't see any error in the log.

domoritz:
Hmm, it's only one build: https://travis-ci.org/vega/vega-lite/jobs/473331462

kanitw:
From conversation on slack, we agreed to merge this.

domoritz:
To repeat the conversation. I still feel that 10px is tight but many users said they would like more ticks so we are changing this.

PR #4421, opened 2018-12-29
Define formula that can take either signals or primitive values only once

kanitw:


kanitw:
This still has a mysterious failure on Node 11. ><"

kanitw:
Oh actually no really mysterious :)

PR #4424, opened 2018-12-29
Update versions and toc, fix markdown

domoritz:


domoritz:
Not sure why the examples updated now. But can you merge this @kanitw?

Issue #4423, opened 2018-12-29
Tooltip for Range 


kanitw:
Tooltip for ranged mark / binned field would benefit a lot from an ability to combine multiple fields in one slot. 

We could consider extending `TooltipFieldDef` to have `field2` and `delimiter` properties.

For instance, for a prebinned field we could say:

```
tooltip: {
  field: "binned_x_start",
  field2: "binned_x_end",
  type: "quantitative",
  title: "X (binned)",
  "delimiter": " - " // This should be the default value
}
```

Or more generally, we could consider supporting expression via [`signal`](#3338). 

cc: @domoritz @chanwutk

domoritz:
I think for advanced use cases, signals would be good.

kanitw:
Thinking more the problem with the `field2` approach is that it's a bit inconsistent with how `x2` and `y2` work. Basically, `x` and `y` do not have `field2` as we have `x2` and `y2`.  

We originally makes this decision as we may want to apply different aggregate / timeUnit to `x2` and `y2` and having `timeUnit2` and `aggregate2` seem a bit too confusing (though requiring `type` in both `x`/`x2` and `x2`/`y2` is a bit unnecessary and potentially confusing as `x2`/`y2` must have the same type as `x` and `y`. 

Due to this consistency, I agree with @domoritz that `signal` is perhaps a better choice.  That said, I don't think we can simply add `signal` to `tooltip` without deciding if we will support `signal` throughout the language at all (#3338).

kanitw:
FWIW, tooltip for prebinned field would work correctly as of #5129 

![image](https://user-images.githubusercontent.com/111269/60379358-a0c05d80-99e6-11e9-8cd2-8348acfaea57.png)

I think we probably can avoid the unnecessary complexity here. (We can reopen this again if we found a better motivating example.)

PR #4426, opened 2018-12-30
Make `x2/y2/latitude2/longitude2` no longer have a `type` and `bin`

kanitw:
(as the `type` should be exactly the same as `x/y/latitude/longitude`)

- Remove `normalizeRangedUnit`, which is no longer necessary
- Simplify code / test for `errorbar.ts`
- Change `tooltipForChannelDefs` to `tooltipForEncoding`


kanitw:
I'm also adding @chanwutk to review as removing type actually simply many logic for error bars.

build/vega-lite-schema.json
```diff
@@ -3888,6 +3888,55 @@
...
       "type": "object"
     },
+    "FieldDefWithoutType": {
+      "additionalProperties": false,
+      "properties": {
+        "aggregate": {
+          "$ref": "#/definitions/Aggregate",
+          "description": "Aggregation function for the field\n(e.g., `mean`, `sum`, `median`, `min`, `max`, `count`).\n\n__Default value:__ `undefined` (None)"
+        },
+        "bin": {
```
domoritz: Could `x2` ever be binned differently from `x`?

build/vega-lite-schema.json
```diff
@@ -3888,6 +3888,55 @@
...
+            {
+              "type": "string"
+            },
+            {
+              "$ref": "#/definitions/RepeatRef"
+            }
+          ],
+          "description": "__Required.__ A string defining the name of the field from which to pull a data value\nor an object defining iterated values from the [`repeat`](https://vega.github.io/vega-lite/docs/repeat.html) operator.\n\n__Note:__ Dots (`.`) and brackets (`[` and `]`) can be used to access nested objects (e.g., `\"field\": \"foo.bar\"` and `\"field\": \"foo['bar']\"`).\nIf field names contain dots or brackets but are not nested, you can use `\\\\` to escape dots and brackets (e.g., `\"a\\\\.b\"` and `\"a\\\\[0\\\\]\"`).\nSee more details about escaping in the [field documentation](https://vega.github.io/vega-lite/docs/field.html).\n\n__Note:__ `field` is not required if `aggregate` is `count`."
+        },
+        "timeUnit": {
```
domoritz: Same question as for bin.

build/vega-lite-schema.json
kanitw: Good point. Actually, we shouldn't even support `bin` for `x2`.  (If `x` is binned, `x2` is inherently the bin end.)

build/vega-lite-schema.json
kanitw: In practice, I don't see why it should be different.  But It's probably feasible to create a weird plot that have them differ.

build/vega-lite-schema.json
domoritz: Then the name of the field def needs to be something else but I think this will make things much cleaner. So 👍

kanitw:
> Just a question about binning and time units but otherwise LGTM.

At this point, given that `aggregate` is the only property that can differ, I wonder if the alternative approach to have `field2` and `aggregate2` for `x` / `y` as suggested in #4423 is more sensible? 

That said, we still have to support the old syntax otherwise it will be quite a massive breaking change.

build/vega-lite-schema.json
domoritz: Hmm, timeUnit is usually like bin so I'd think it's the same as bin. Should we just disallow this until someone complains?

domoritz:
What about a chart where you make the start of a bar a fixed value and the end data dependent. That wouldn't be possible with `aggregate2`.

kanitw:
> What about a chart where you make the start of a bar a fixed value and the end data dependent. That wouldn't be possible with aggregate2.

Good point.  We may consider `aggregate2` and `field2` as shorthands in the future then.  But we definitely need to keep `x2` and `y2` as the primary way to specify this.

examples/compiled/area_temperature_range.vg.json
```diff
@@ -12,7 +12,7 @@
       "url": "data/seattle-weather.csv",
       "format": {
         "type": "csv",
-        "parse": {"date": "date", "temp_max": "number", "temp_min": "number"}
+        "parse": {"date": "date", "temp_max": "number"}
```
kanitw: Need to still include `temp_min`

examples/compiled/interactive_brush.vg.json
```diff
@@ -295,7 +295,7 @@
             {"value": "grey"}
           ],
           "tooltip": {
-            "signal": "{\"Horsepower\": format(datum[\"Horsepower\"], \"\"), \"Miles_per_Gallon\": format(datum[\"Miles_per_Gallon\"], \"\"), \"Cylinders\": ''+datum[\"Cylinders\"]}"
+            "signal": "{\"Horsepower\": format(datum[\"Horsepower\"], \"\"), \"Miles_per_Gallon\": format(datum[\"Miles_per_Gallon\"], \"\")}"
```
kanitw: Somehow Cylinders is now missing

domoritz:
> Good point. We may consider aggregate2 and field2 as shorthands in the future then. But we definitely need to keep x2 and y2 as the primary way to specify this.

Maybe later. I'm not yet convinced we want it until we have thoroughly considered the design.

build/vega-lite-schema.json
domoritz: Agreed

src/encoding.ts
```diff
@@ -67,7 +73,8 @@ export interface Encoding<F> {
   /**
    * Secondary error value of x coordinates for error specified `"errorbar"` and `"errorband"`.
    */
-  xError2?: FieldDefWithoutScale<F> | ValueDef;
+  // `xError2` cannot have type as it should have the same type as `xError`
```
domoritz: `xError` should have the same type as `x`. Therefore `xError` should also be a `SecondaryRangeFieldDef`

domoritz: Please fix in a follow up PR.

PR #4427, opened 2018-12-30
Adjust formual for tickCount to use `floor()`

kanitw:


kanitw:
Since this isn't strictly better, let's not do this.

PR #4428, opened 2018-12-30
Improve `type` of different channels

kanitw:
- Make only `shape` accept nominal/ordinal/geojson
- Make lat/long accepts quantitative only
- Make key/href accepts nominal only

TODO: @domoritz make JSON schema supports default generic type?

Currently `yarn schema` throws:

```
LogicError: Could not find type parameter "T"
```


domoritz:
Schema generator works with 0.38.1. Please update.

PR #4437, opened 2019-01-01
Make `type` optional for LatLongFieldDef 

kanitw:
Note: I still keep bin property for lat long, but make it support `null` only for simplicity of the codebase

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1826,6 +1826,9 @@
                 "binned"
               ],
               "type": "string"
+            },
+            {
+              "type": "null"
             }
           ],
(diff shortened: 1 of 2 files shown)


build/vega-lite-schema.json
```diff
@@ -6071,7 +6059,63 @@
...
     },
     "LatLongFieldDef": {
-      "$ref": "#/definitions/FieldDef<\"quantitative\">"
+      "additionalProperties": false,
+      "properties": {
+        "aggregate": {
+          "$ref": "#/definitions/Aggregate",
+          "description": "Aggregation function for the field\n(e.g., `mean`, `sum`, `median`, `min`, `max`, `count`).\n\n__Default value:__ `undefined` (None)"
+        },
+        "bin": {
```
kanitw: Bin should support `null` only here since I declare

```ts
export type LatLongFieldDef<F extends Field> = FieldDefBase<F> &
  TitleMixins &
  Partial<TypeMixins<'quantitative'>> &
  GenericBinMixins<null>;
```

kanitw: https://github.com/vega/ts-json-schema-generator/issues/58

kanitw:
Btw, we can merge this since this is already an improvement.  The remaining problem will be fixed once we fix the schema generator.

build/vega-lite-schema.json
domoritz: This is the case now.

PR #4453, opened 2019-01-12
Eliminate Vg prefixes from the JSON schema

### LINKED ISSUE #4071: Eliminate Vg prefix from  JSON schema
(If they're in the schema, they are a part of Vega-Lite too, so we shouldn't need a Vg prefix.)

kanitw:
Fix #4071

Rationale: Users of Altair shouldn't need to care which part of Vega-Lite spec is identical to Vega. (The Vg- prefix is weird an API.) 

Merge https://github.com/vega/vega-lite/pull/4452 first.


domoritz:
Do you think we should add a test to ensure that no type in the schema has a Vg prefix?

kanitw:
Probably good, feel free to do it.

PR #4462, opened 2019-01-19
Automatically delete merged branches

domoritz:
From https://github.com/lannonbr/issue-label-manager-action


domoritz:
Oh "During the limited public beta, you can only configure workflows in private repositories."

I guess we have to wait until GitHub activates actions in public repos as well. We can just leave the file in the repo so we don't miss when they do.

PR #4465, opened 2019-01-19
Switch from Webdriverio to Puppeteer

domoritz:
Advantages

* wdio v5 does not support the static server that we need
* We could not use jest, now we can
* fewer dependencies


domoritz:
Tests are failing because we use different ids to clip: https://www.diffchecker.com/1lSThiFF.

domoritz:
@arvind @jheer @kanitw How is it possible that tests pass in node 11 but not before with the failures being due to different SVGs? Shouldn't the SVG generation be predictable/repeatable in different node versions? 

screen shot 2019-01-23 at 22 18 49

kanitw:
That’s why I disabled it lol

domoritz:
Not a fan :-P I changed it so we run runtime tests only on node 11.

kanitw:
> Not a fan :-P I changed it so we run runtime tests only on node 11.

I believe we didn't run Node 11 back then so it wasn't an option.

domoritz:
It must have been correct on some version of node, no?

kanitw:
> It must have been correct on some version of node, no?

Right now just hope for node 11.   :) 


LGTM, but I'll let @arvind merge this.

test-runtime/resolve.test.ts
```diff
@@ -15,37 +14,45 @@ import {
...
+        it('should have one global selection instance', async () => {
           const selection = {
             type,
             resolve: 'global',
             ...(specType === 'facet' ? {encodings: ['y']} : {})
           };
 
           for (let i = 0; i < hits[specType].length; i++) {
-            embed(spec(specType, i, selection));
+            await await embed(spec(specType, i, selection));
```
arvind: `await await`?

domoritz: Ups

domoritz: Fixed in https://github.com/vega/vega-lite/pull/4506

Issue #4431, opened 2018-12-31
Cannot run `yarn site`


kanitw:
```
node_modules/vega-lite/build/src/util.d.ts:1:29 - error TS7016: Could not find a declaration file for module 'json-stable-stringify'. '/Users/kanitw/Documents/code/vega-lite/node_modules/json-stable-stringify/index.js' implicitly has an 'any' type.
  Try `npm install @types/json-stable-stringify` if it exists or add a new declaration (.d.ts) file containing `declare module 'json-stable-stringify';`

1 import stableStringify from 'json-stable-stringify';
                              ~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```

domoritz:
Hmm, the problem is that the site uses an old version of Vega-Lite. This will be fixed with the next release, I think. 

I don't know who's pulling in vega-lite in the site.

domoritz:
So I'd say the fix is to make a new release.

kanitw:
We already released a new version, but I still run into this problem.

domoritz:
I just sent a PR to fix it.

kanitw:
Fixed in $#4469

PR #4469, opened 2019-01-19
Real fix for #4431

kanitw:
It's asking for a type declaration, so the right fix is the add type declaration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
   },
   "dependencies": {
     "@types/clone": "^0.1.30",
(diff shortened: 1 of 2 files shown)


domoritz:
No, we don't use `json-stable-stringify` anymore. We should not merge this. 

The problem was that the old Vega-Lite depended on typings that were not in the dependencies.

kanitw:
Why does that stay in the site folder? Then something in the site folder is wrong.

domoritz:
I'm not sure.

PR #4474, opened 2019-01-19
[Refactor] Correctly move interfaces to `spec/base`

kanitw:


src/spec/toplevel.ts
```diff
@@ -66,6 +65,7 @@ export interface TopLevelProperties {
   autosize?: AutosizeType | AutoSizeParams;
 }
 
+<<<<<<< HEAD
```
domoritz: Leftover from merge

Issue #1489, opened 2016-08-04
Make `sort.op` optional


kanitw:
Maybe with `min` or `mean` as default op.

kanitw:
I think it's unclear what's the best default so maybe let's not do it.

kanitw:
People are asking for this in https://github.com/altair-viz/altair/issues/884. 

Any of min, median, mean work ok for the simple case.  However, determining a good default requires thinking about more general case.  For example,  if sort is specified for y-encoding of [a strip plot](https://vega.github.io/editor/#/examples/vega-lite/tick_strip), it is unclear what is the reasonable default `op`.

domoritz:
I think @palewire is right. If the chart is not aggregated, the user shouldn't have to specify an aggregation. Unfortunately, Vega requires it so we have to set something and we definitely don't want to set a default when the chart is aggregated.

kanitw:
>  Unfortunately, Vega requires it

Well it's not entirely Vega's fault.  Semantically, it only _always_ work if there is an aggregate operation applied.  (You never know if the field you sort by always have 1:1 correspondence with the axis's field.) 

> we definitely don't want to set a default when the chart is aggregated.

If we have good rationale for the default, I'm open to set it.  But we haven't arrive at a conclusion for a good rationale yet.

domoritz:
Could we set a default (anything really, let's use `max`) for non-aggregated charts?

kanitw:
Given that you prefer max, I'd guess you'd prefer min to mean/median?  **

palewire:
Am I crazy to think that a sensible default would be that if no aggregation function is provided Vega should assume there in a 1:1 relationship between the axis and the sort, raising an error if there isn't?

domoritz:
That would be an enhancement to Vega. Unfortunately, we can't do anything about this in Vega-Lite as we don't have access to the data. Adding @jheer to see what he thinks.

iliatimofeev:
I think `mean` is the best solution. In most of the time you just try to sort bar chart, reordering other types of chart will need very special reason. 

As new user I felt a enlightenment when I come up with understanding that sorting field in fact do not match ordering axis and that is reason behind so complex definition of sorting.

iliatimofeev:
More than, I think would be helpful to have some simple syntax to sort items over second axis values without defining which field to use. I believe that is really default use case that cover most of all. 


```javascript
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
[...]

PR #4475, opened 2019-01-19
Make sort op optional and make `"mean"` the default op

kanitw:
Fix https://github.com/vega/vega-lite/issues/1489

I think this will resolve a common error that many novice users run into.  While this may introduce a slightly different problem (namely that people may not specify op: 'mean' even when they want to), I think the benefits outweight the cons in this case. 

FWIW, the lack of default sort op has confused so many users already.

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -3566,7 +3566,7 @@
         },
         "op": {
           "$ref": "#/definitions/AggregateOp",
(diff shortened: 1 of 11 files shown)


domoritz:
On slack, we decided to make the default min so it's blatantly obvious that aggregation is happening.

domoritz:
jk, we decided to stick with mean.

PR #4479, opened 2019-01-21
Update dependencies to fix the site. Fixes #4451.

### LINKED ISSUE #4451: Documentation examples broken
Documentation viz examples aren't rendering for some reason. For example, here

https://vega.github.io/vega-lite/docs/scale.html#continuous

I see this:

image

I observed this in both Chromium and Firefox.

Wondering if there's any relation to #3535.

Thanks

domoritz:
I just ran `rm yarn.lock && yarn`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ addons:
 
 language: node_js
 node_js:
-  - "6"
   - "8"
   - "10"
 env: LATEST=false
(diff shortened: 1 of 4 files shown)


kanitw:
PR test failing

domoritz:
Weird. I rebased to master already.

domoritz:
No idea why we see this error. This PR switches from `--list-different` to `--check` so the command is wrong.

PR #4484, opened 2019-01-23
Support nested row of row facet. We previously returned a signal `unk…

domoritz:
…nown` for columns.

Now this works

```
{
  "facet": {
    "column": {"field": "Cylinders","type": "ordinal"}
  },
  "spec": {
    "facet": {
      "column": {"field": "Origin","type": "ordinal"}
    },
    "spec": {
      "mark": "point",
      "encoding": {    
        "x": {"field": "Displacement","type": "quantitative"},
        "y": {"field": "Horsepower","type": "quantitative"}
      }
    }
  }
}
```

### DIFF
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -224,7 +224,7 @@ export class FacetModel extends ModelWithField {
     return {
       ...this.getHeaderLayoutMixins(),
 
-      columns,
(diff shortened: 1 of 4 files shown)


domoritz:
Instead, I made the tests stricter.

PR #4485, opened 2019-01-23
Do not align when scales are independent

domoritz:
```json
{
  "facet": {
    "row": {"field": "Cylinders","type": "ordinal"}
  },
  "spec": {
    "mark": "bar",
    "encoding": {    
      "y": {"field": "Origin","type": "ordinal"},
      "x": {"aggregate": "count","type": "quantitative"}
    }
  },
  "resolve": {
    "scale": {
      "y": "independent"
    }
  }
}
```

```json
{
  "facet": {
    "column": {"field": "Cylinders","type": "ordinal"}
  },
  "spec": {
    "mark": "bar",
    "encoding": {    
      "x": {"field": "Origin","type": "ordinal"},
      "y": {"aggregate": "count","type": "quantitative"}
    }
  },
  "resolve": {
    "scale": {
      "x": "independent"
    }
  }
}
```

@jheer this should make the layout in your example from today better by default.

### DIFF
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -221,8 +221,8 @@ export class FacetModel extends ModelWithField {
 
     let align: LayoutAlign = 'all';
 
-    // Do not align the cells if the scale corresponding to teh directin is indepent.
-    // We always align when we facet into row and column.
+    // Do not align the cells if the scale corresponding to the directin is indepent.
+    // We always align when we facet into both row and column.
     if (!this.channelHasField('row') && this.component.resolve.scale.x === 'independent') {
       align = 'none';
     } else if (!this.channelHasField('column') && this.component.resolve.scale.y === 'independent') {
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -1,4 +1,4 @@
-import {AggregateOp, NewSignal} from 'vega';
+import {AggregateOp, LayoutAlign, NewSignal} from 'vega';
 import {isArray} from 'vega-util';
 import {Channel, COLUMN, ROW, ScaleChannel} from '../channel';
 import {Config} from '../config';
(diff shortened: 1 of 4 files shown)


src/compile/facet.ts
```diff
@@ -219,14 +219,22 @@ export class FacetModel extends ModelWithField {
   protected assembleDefaultLayout(): VgLayout {
     const columns = this.channelHasField('column') ? this.columnDistinctSignal() : 1;
 
-    // TODO: determine default align based on shared / independent scales
+    let align: LayoutAlign = 'all';
+
+    // Do not align the cells if the scale corresponding to teh directin is indepent.
```
kanitw: typos

src/compile/facet.ts
```diff
@@ -219,14 +219,22 @@ export class FacetModel extends ModelWithField {
...
+    // We always align when we facet into row and column.
```
kanitw: The comment "We always align when we facet into row and column." doesn't match what you do below.

test/compile/facet.test.ts
```diff
@@ -246,6 +246,34 @@ describe('FacetModel', () => {
       });
     });
 
+    it('should not align independent scales', () => {
```
kanitw: Add another row case?

kanitw:
Feel free to merge after fixing

src/compile/facet.ts
domoritz: Better? "We always align when we facet into both row and column."

domoritz:
I changed a number of tests (use ordinal instead of quantitative). Can you confirm and merge?

test/compile/facet.test.ts
domoritz: DOne

src/compile/facet.ts
domoritz: Done

domoritz:
I just fixed master so you can ignore the pr build.

PR #4487, opened 2019-01-23
BUG: Selection dropdown html is broken

pratapvardhan:
https://vega.github.io/vega-lite/docs/selection.html select dropdown is broken due to newline between html content in markdown. This patch removes it.


domoritz:
Thanks

PR #4490, opened 2019-01-24
Use html instead of markdown to add class. Fixes #4489

### LINKED ISSUE #4489: Website front page has a broken directive
The rendered text of the front page of the Vega-Lite Jekyll site is showing the following:

`{:.lead} Vega-Lite`

This looks like an error in which a directive is passing through into the HTML output.

domoritz:


kanitw:
Oops, just saw that Jeff is the reviewer...

PR #4497, opened 2019-01-28
Bin optimizer

domoritz:
@Saba9 Please continue


domoritz:
When this is ready, please update the title and labels.

domoritz:
Please rebase to master. this should be easy now that we sue Vega 5 and the new SignalRefWrapper.

src/compile/data/assemble.ts
```diff
@@ -203,8 +203,8 @@ export function assembleFacetData(root: FacetNode): VgData[] {
 export function assembleRootData(dataComponent: DataComponent, datasets: Dict<InlineDataset>): VgData[] {
   const data: VgData[] = [];
 
-  // roots.forEach(debug);
-  // draw(roots);
+  // dataComponent.sources.forEach(debug);
+  // draw(dataComponent.sources);
```
domoritz: Thanks

test/compile/scale/domain.test.ts
```diff
@@ -125,12 +126,8 @@ describe('compile/scale', () => {
               y: fieldDef
             }
           });
-
-          expect(testParseDomainForChannel(model, 'y')).toEqual([
-            {
-              signal: '[bin_maxbins_15_origin_bins.start, bin_maxbins_15_origin_bins.stop]'
-            }
-          ]);
+          const domain = testParseDomainForChannel(model, 'y')[0] as SignalRefWrapper;
```
domoritz:
```suggestion
          const domain = testParseDomainForChannel(model, 'y')[0] as SignalRef;
```

src/compile/data/optimizers.ts
```diff
@@ -343,3 +350,51 @@ export class MergeAggregateNodes extends BottomUpOptimizer {
     return this.flags;
   }
 }
+
+/**
+ * Merge bin nodes and move bin nodes up through forks but stop at filters and parse.
```
domoritz:
```suggestion
 * Merge bin nodes and move bin nodes up through forks but stop at filters and parse as we want them to stay before the bin node.
```

src/compile/scale/range.ts
```diff
@@ -116,10 +116,13 @@ function getRangeStep(model: UnitModel, channel: 'x' | 'y'): number | SignalRef
...
       const sizeType = getSizeType(channel);
       const sizeSignal = model.getName(sizeType);
-      return new SignalRefWrapper(() => `${model.getSignalName(sizeSignal)} / (${model.getSignalName(binCount)})`);
+      return new SignalRefWrapper(() => {
+        const signalName = model.getSignalName(binSignal);
+        const binCount = `(${signalName}.stop - ${signalName}.start) / ${signalName}.step`;
+        return `${model.getSignalName(sizeSignal)} / (${binCount})`;
+      });
     }
     // TODO: handle binned case
```
domoritz: Don't we do this above?

src/compile/scale/range.ts
```diff
@@ -116,10 +116,13 @@ function getRangeStep(model: UnitModel, channel: 'x' | 'y'): number | SignalRef
       const binSignal = model.getName(vgField(fieldDef, {suffix: 'bins'}));
 
       // TODO: extract this to be range step signal
-      const binCount = `(${binSignal}.stop - ${binSignal}.start) / ${binSignal}.step`;
       const sizeType = getSizeType(channel);
       const sizeSignal = model.getName(sizeType);
```
domoritz: Just wondering, why do we need this at all? What happens if we delete all of it and never generate a rangeStep for binned fields?

domoritz: We may need it for ordinal data so try that.

domoritz:
Probably needs to merge master again.

src/compile/scale/range.ts
Saba9: Ahh, looks like we do need the rangeStep signal if you look at the updated examples.

domoritz:
Is this ready?

domoritz:
There is some untested code that we should cover in tests: https://codecov.io/gh/vega/vega-lite/compare/2307cc3ee83e4f5bf80a9106941964ffaa58d5b2...3d2fb4ed0b9a300bb1319adb5250ad2189f76a1d/src/src/compile/scale/range.ts#L317

domoritz:
Ping @Saba9

Saba9:
yeah I believe this is ready. Just need to write some tests for coverage to pass

Saba9:
actually I need to revert a change

Saba9:
Alright, should be ready for review. I can squash the commits before we merge.

domoritz:
Thank you!

domoritz:
I think something is missing here. We are merging only whole bin nodes but not individual bin operations. Am I right @Saba9?

PR #4499, opened 2019-01-28
Allow scheme config with only count or only extent.

domoritz:
We now support

```json
{
  "transform": [{
    "filter": {"and": [
      {"field": "IMDB_Rating", "valid": true},
      {"field": "Rotten_Tomatoes_Rating", "valid": true}
    ]}
  }],
  "mark": "rect",
  "width": 300,
  "height": 200,
  "encoding": {
    "x": {
      "bin": {"maxbins":60},
      "field": "IMDB_Rating",
      "type": "quantitative"
    },
    "y": {
      "bin": {"maxbins": 40},
      "type": "quantitative"
    },
    "color": {
      "aggregate": "count",
      "type": "quantitative",
      "scale": {
        "type": "quantize",
        "scheme": {
          "count": 5
        }
      }
    }
  },
  "config": {
    "range": {
      "heatmap": {
        "scheme": "greenblue"
      }
    },
    "view": {
      "stroke": "transparent"
    }
  }
}
```

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -9064,9 +9064,6 @@
           "type": "string"
         }
       },
-      "required": [
-        "name"
-      ],
       "type": "object"
     },
     "SecondaryFieldDef": {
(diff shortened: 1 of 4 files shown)


domoritz:
Actually, this does not work.

PR #4504, opened 2019-01-29
Suppress all errors in JSON syntax highlighting

domoritz:
I saw an error on the config page and decided to just hide all errors. 

screen shot 2019-01-28 at 17 26 12

### DIFF
--- a/site/docs/composition/concat.md
+++ b/site/docs/composition/concat.md
@@ -21,8 +21,6 @@ If you concatenate similar views where the only difference is the field that is 
 
 Put multiple views into a row by putting the specs for each view into an array and assign it to the `hconcat` property.
 
-{: .suppress-error}
-
 ```json
 {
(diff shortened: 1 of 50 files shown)


domoritz:
Same as before we broke it (black text with no background).

kanitw:
What about using js instead of JSON?

domoritz:
We can experiment with js instead of json. I just wanted to not see errors everywhere anymore.

kanitw:
I think JS is better as we'll still get syntax highlighting.

domoritz:
Why don't we get syntax highlighting with json?

domoritz:
@kanitw Let's merge this, okay?

kanitw:
Try JS?

domoritz:
Yeah, this is not relevant with #4587

PR #4510, opened 2019-01-29
Add example application using Vega (Olmsted)

metasoarous:


metasoarous:
@domoritz Awesome! Thanks again!

PR #4523, opened 2019-02-08
Fix Customize your Visualization section of tutorial

melissachang:
The default title for average aggregate was `MEAN(b)`. Now it's `Average of b`. With the new title, the [Customize your Visualization section](https://vega.github.io/vega-lite/tutorials/getting_started.html#customize-your-visualization) needs to be updated.

Before:
before

After:
after

With `yarn test`, `toggle.test.ts` fails. But it also fails at head, so I assume I can ignore.


domoritz:
Thank you for the fix!

PR #4526, opened 2019-02-09
Allow disabling tooltips from the config and document how to disable tooltips.

### LINKED ISSUE #4512: Document how to disable tooltips
https://vega.github.io/vega-lite/docs/tooltip.html does not mention how to disable tooltips.

domoritz:
Fixes #4512


kanitw:
Note that this PR doesn't "disabling tooltips from the config" as it already works.  The PR only properly provides typing and docs. :)

PR #4440, opened 2019-01-04
Separate `NormalizedUnitSpec` from `CompositeUnitSpec`

kanitw:
and introduce `ExtendedUnitSpec = NormalizedUnitSpec | CompositeUnitSpec` 

This makes the typing clearer between what's normalized (supported internally) vs what's composite unit spec only. 

cc: @jakevdp 

Merge #4493 First

PR #4528, opened 2019-02-09
Don't include extra encoding for composite mark in the main `Encoding`

kanitw:
This PR simplifies #4440 and keep the number of `Encoding` declarations small. 

In the schema, there are still simply `FacetedEncoding` and `Encoding`.

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -85,6 +85,22 @@
       ],
       "type": "string"
     },
+    "AnyMark": {
+      "anyOf": [
+        {
+          "$ref": "#/definitions/CompositeMark"
+        },
+        {
+          "$ref": "#/definitions/CompositeMarkDef"
+        },
+        {
+          "$ref": "#/definitions/Mark"
+        },
+        {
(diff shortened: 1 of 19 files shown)


build/vega-lite-schema.json
```diff
@@ -4020,6 +3678,56 @@
       ],
       "type": "string"
     },
+    "ErrorExtraEncoding": {
```
kanitw: This is actually never referred anywhere. Probably an example of the bug in the generator.

kanitw: Ok. I re-declare this in a way that this problem would no longer surface.

domoritz: Yeah, we sometimes output unnecessary stuff but it should not hurt (except for larger schemas).

kanitw: lol. Isn't Vega-Lite schema large enough?

domoritz: Oh, I'm not saying it's not a bug but it's not a blocking bug.

PR #4532, opened 2019-02-10
[Docs] Add `vscode-jest-runner` to the list of recommended tools

kanitw:


domoritz:
The link isn't very useful as it provides to installation instructions.

PR #4534, opened 2019-02-10
Restructure spec name for repeat

kanitw:
This will make it easier to understand spec name (for interacting with selection API) 

Merge #4535 first

### DIFF
--- a/src/compile/repeat.ts
+++ b/src/compile/repeat.ts
@@ -40,7 +40,7 @@ export class RepeatModel extends BaseConcatModel {
     // cross product
     for (const rowField of row) {
       for (const columnField of column) {
-        const name = (rowField ? '_' + rowField : '') + (columnField ? '_' + columnField : '');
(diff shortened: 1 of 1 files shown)


src/compile/repeat.ts
```diff
@@ -40,7 +40,7 @@ export class RepeatModel extends BaseConcatModel {
     // cross product
     for (const rowField of row) {
       for (const columnField of column) {
-        const name = (rowField ? '_' + rowField : '') + (columnField ? '_' + columnField : '');
+        const name = (rowField ? `__row_${rowField}` : '') + (columnField ? `__column_${columnField}` : '');
```
domoritz: Please use the `internalField` method so it's easier to track where internal fields are being created.

kanitw: I'm not sure what you're talking about.  This is setting a spec "name".

domoritz: Ignore me.

PR #4536, opened 2019-02-10
Add `repeat` without nested mapping (for wrapped repeat)

kanitw:
Merge #4534 first


domoritz:
These two are equivalent, right? 

```
{
  data: ...
  repeat: {
    row: {field: 'category', type: 'nominal'}
  },
  columns: 1,
  spec: ...
}
```

```
{
  data: ...
  repeat: {
    column: {field: 'category', type: 'nominal'}
  },
  spec: ...
}
```

kanitw:
Good point. We should make `columns` only respected when it's 1D repeat.

domoritz:
Yeah, ideally, we could have two different types for this to make sure Altair users don't get confused.

kanitw:
> Yeah, ideally, we could have two different types for this to make sure Altair users don't get confused.

That will explode the schema, make the internal code way more complex, and perhaps make it even harder to deal with auto-generated code in Altair, so I don't think it's a good idea.

domoritz:
Let's make sure to throw a warning then.

kanitw:
Note that we can't do that without introducing a lot of hacks until we fix https://github.com/vega/ts-json-schema-generator/issues/62 as well. (Otherwise, we can't declare `|` for these two types of repeat.  The rest will be nightmarish to read.)

kanitw:
Columns is now dropped and warned for facet with row/column mapping.

PR #4530, opened 2019-02-10
Introduce a common SpecMapper

kanitw:
~~Merge #4528 first.~~

Part of #4376

PR #4537, opened 2019-02-10
Make Facet & Repeat's unit names more decodable

kanitw:
Basically, we will append

- `__facet_${channel}_${fieldValue}` for facet
- `__repeat_${channel}_${repeatValue}` for repeat

This will make it easy to decode unit name and enable us to implement a more accessible selection API. (https://github.com/vega/vega-lite/issues/1830)

Merge #4530


src/compile/repeat.ts
```diff
@@ -40,7 +40,8 @@ export class RepeatModel extends BaseConcatModel {
     // cross product
     for (const rowField of row) {
       for (const columnField of column) {
-        const name = (rowField ? '_' + rowField : '') + (columnField ? '_' + columnField : '');
+        const name =
+          (rowField ? `__repeat-row_${rowField}` : '') + (columnField ? `__repeat-column_${columnField}` : '');
```
kanitw: @domoritz @arvind 

Do we have a method to escape rowField so that it's a valid CSS name? (Our spec name must be a valid CSS name.)

src/compile/selection/selection.ts
```diff
@@ -364,7 +364,7 @@ export function unitName(model: Model) {
     const {facet} = facetModel;
     for (const channel of FACET_CHANNELS) {
       if (facet[channel]) {
-        name += ` + '_' + (${accessPathWithDatum(facetModel.vgField(channel), 'facet')})`;
+        name += ` + '__facet-${channel}_' + (${accessPathWithDatum(facetModel.vgField(channel), 'facet')})`;
```
kanitw: @arvind  -- not sure if `unitName` has to be a valid CSS name.  (If so, do we have a Vega expression method to escape it?)

src/compile/repeat.ts
domoritz: I don't know.

kanitw:
@arvind -- do you have any objection for this? (If not, we'll merge this.)

src/compile/repeat.ts
arvind: We have `varName` which I think should produce valid CSS names.

src/compile/selection/selection.ts
arvind: Not sure why we need the method exposed in Vega expression? The strings could be escaped prior to passing it in? Perhaps I'm misunderstanding?

arvind:
This looks great to me, no objections! Thanks @kanitw!

PR #4538, opened 2019-02-10
add top-k plot WITH others

manzt:
Following the example in Vega (https://vega.github.io/vega/examples/top-k-plot-with-others/), I implemented this same example in VL performing subsequent aggregations.


manzt:
#1486

domoritz:
This is great. At some point we might want to add these examples to Vega-Lite itself. Is that okay with you?

manzt:
Yeah, absolutely.

PR #4554, opened 2019-02-15
Add joinaggregate. Fixes #4121

### LINKED ISSUE #4121: Join aggregate transform
We already support window but joinaggregates are probably the most common and are supported natively in Vega. 

See #3912

domoritz:


src/transform.ts
```diff
@@ -207,6 +208,35 @@ export interface WindowTransform {
...
 }
 
+export interface JoinAggregateDef {
+  /**
+   * The aggregation operation to apply (e.g., sum, average or count). See the list of all supported operations [here](https://vega.github.io/vega-lite/docs/aggregate.html#ops).
+   */
+  op: AggregateOp;
+
+  /**
+   * The data field for which to compute the aggregate function. This can be omitted for window functions that do not operate over a field such as `count`.
```
kanitw: >  for window functions

This no longer "window functions" right?

.prettierignore
```diff
@@ -0,0 +1 @@
+examples/specs/*.vl.json
```
kanitw: Can we put this rule in `.prettierrc.json`?

If we can, probably better to consolidate them in one place.

src/transform.ts
domoritz: Fixed

.prettierignore
domoritz: I don't think it works yet. https://github.com/prettier/prettier/issues/3460

examples/compiled/area_cumulative_freq.vg.json
```diff
@@ -12,13 +12,10 @@
       "format": {"type": "json"},
       "transform": [
         {
-          "type": "window",
-          "params": [null],
+          "type": "joinaggregate",
```
kanitw: Cumulative frequency should not use join aggregate.  

Window with frame = `[null, 0]` makes more sense. (See SVG)

examples/compiled/layer_cumulative_histogram.vg.json
```diff
@@ -33,13 +33,10 @@
...
         {
-          "type": "window",
-          "params": [null],
+          "type": "joinaggregate",
           "as": ["cumulative_count"],
           "ops": ["sum"],
-          "fields": ["count"],
-          "sort": {"field": ["bin_IMDB_Rating"], "order": ["ascending"]},
-          "frame": [null, 0]
+          "fields": ["count"]
```
kanitw: Cumulative frequency should not use join aggregate.  

Window with frame = `[null, 0]` makes more sense. (See SVG)

examples/compiled/waterfall_chart.vg.json
```diff
@@ -30,12 +30,10 @@
       "source": "source_0",
       "transform": [
         {
-          "type": "window",
-          "params": [null],
+          "type": "joinaggregate",
```
kanitw: 1) Don't rename this file

2) waterfall should use window (with default frame = `[null, 0]`)

examples/compiled/window_cumulative_running_average.vg.json
```diff
@@ -19,14 +19,10 @@
           "expr": "datetime(year(datum[\"Year\"]), 0, 1, 0, 0, 0, 0)"
         },
         {
-          "type": "window",
-          "params": [null],
+          "type": "joinaggregate",
```
kanitw: wrong

examples/specs/bar_diverging_stack_transform.vl.json
```diff
@@ -63,8 +63,8 @@
           "as": "offset"
         }
       ],
-      "groupby": ["question"],
-      "frame": [null, null]
+      "frame": [null, null],
```
kanitw: This one can be joinaggregate

examples/compiled/area_cumulative_freq.vg.json
domoritz: Ups, my mistake.

domoritz: The code should only do it when the frame is `[null, null]`

domoritz: This was a bug in the code. Fixed.

examples/specs/waterfall_chart.vl.json
```diff
@@ -21,7 +21,7 @@
   "width": 800,
   "height": 450,
   "transform": [
-    {"window": [{"op": "sum", "field": "amount", "as": "sum"}]},
+    {"joinaggregate": [{"op": "sum", "field": "amount", "as": "sum"}]},
```
domoritz: @kanitw We use joinaggregate for the window where it makes sense. The name became a bit unwieldy and I don't think we need to list every feature in the filename for this complex example. I was thinking that we maybe want a separate section for complex/custom designs at the bottom of the examples page.

kanitw: You can keep the name if you have strong preference, but window doesn't make sense here.   See SVG.

kanitw:
Instead of just doing it in the code, I think better update all the examples (except one of them) to use joinaggregate in VL examples too.

domoritz:
> Instead of just doing it in the code, I think better update all the examples (except one of them) to use joinaggregate in VL examples too.

That's what I did. Which ones did I miss?

domoritz:
Fixed bar_diverging as well

domoritz:
@kanitw This is ready

examples/compiled/waterfall_chart.vg.json
kanitw: Waterfall chart is still wrong.  It should not use `joinaggregate`.

kanitw: image

site/docs/transform/joinaggregate.md
```diff
@@ -0,0 +1,78 @@
+---
+layout: docs
+menu: docs
+title: Join Aggregate
+permalink: /docs/joinaggregate.html
+---
+
+The joinaggregate transform extends the input data objects with aggregate values. Aggregation is performed and the results are then joined with the input data. This transform can be helpful for creating derived values that combine both raw data and aggregate calculations, such as percentages of group totals. This transform is a special case of the [window](window.html) transform where the `frame` is always `[null, null]`.
```
kanitw: > The joinaggregate transform extends the input data objects with aggregate values.

1) Explicitly mention that the extension is a new column?  (Right now it's still a bit ambiguous.)


> This transform is a special case of the [window](window.html) transform where the `frame` is always `[null, null]`.

It's good to compare this to `window`, but perhaps it's more important to highlight the difference between this and normal aggregate.  (Similarly, add docs to aggregate to mention that "use joinaggregate if you want to preserve the table structure and instead add a new column for the aggregated values.

src/transform.ts
```diff
@@ -207,6 +208,35 @@ export interface WindowTransform {
   sort?: SortField[];
 }
 
+export interface JoinAggregateDef {
```
kanitw: FWIW, the similar thing for Aggregate is called `AggregateFieldDef`, so maybe `JoinAggregateFieldDef` for consistency?

site/docs/transform/joinaggregate.md
domoritz: Done

src/transform.ts
domoritz: Done

PR #4556, opened 2019-02-15
Use gnu parallel from apt

domoritz:


domoritz:
No idea why travis is not running yarn anymore. I sent a support ticket to Travis.

Issue #4508, opened 2019-01-29
Symbol legend has gradient length


domoritz:
See https://vega.github.io/editor/#/examples/vega-lite/point_color_with_shape

`gradientLength` is not needed here. 

```json
"legends": [
    {
      "stroke": "color",
      "gradientLength": {"signal": "clamp(height, 64, 200)"},
      "title": "Origin",
      "encode": {
        "symbols": {
          "update": {
            "fill": {"value": "transparent"},
            "opacity": {"value": 0.7}
          }
        }
      },
      "shape": "shape"
    }
  ],
```

jshreyans551:
Hi @domoritz! 
Considering that this issue was a trivial one, I have directly submitted a PR for this. You can find it here:
[#4560](https://github.com/vega/vega-lite/pull/4560)

PR #4560, opened 2019-02-17
Removed gradient length for #4508

jshreyans551:
The gradient length has been removed from the symbol type


domoritz:
Hmm, this just removed it from the automatically compiled Vega but doesn't actually fix the Vega-Lite compiler. Unfortunately, that's not enough to actually fix the issue.

jshreyans551:
I will try to look for a better solution

domoritz:
Thanks. I'll close this for now.

PR #4566, opened 2019-02-21
Upgrade to Vega 5

domoritz:


domoritz:
Blocked by https://github.com/vega/vega/issues/1607

domoritz:
Blocked by a new Vega release.

src/scale.ts
```diff
@@ -10,41 +9,39 @@ import {ScaleInterpolate, ScaleInterpolateParams} from './vega.schema';
...
   export const UTC: 'utc' = 'utc';
-  // sequential
-  export const SEQUENTIAL: 'sequential' = 'sequential';
 
-  // Quantile, Quantize, threshold
+  // Discretizing scales
   export const QUANTILE: 'quantile' = 'quantile';
   export const QUANTIZE: 'quantize' = 'quantize';
   export const THRESHOLD: 'threshold' = 'threshold';
+  export const BIN_ORDINAL: 'bin-ordinal' = 'bin-ordinal';
```
kanitw: I think bin ordinal isn't really a discretizing scales because its input values are already discrete. (But let's see what Jeff says on Slack.)

src/scale.ts
```diff
@@ -123,22 +118,14 @@ export function scaleTypePrecedence(scaleType: ScaleType): number {
...
-  'log',
-  'pow',
-  'sqrt',
-  'time',
-  'utc'
-];
+export const CONTINUOUS_TO_CONTINUOUS_SCALES: ScaleType[] = ['linear', 'log', 'pow', 'sqrt', 'symlog', 'time', 'utc'];
 const CONTINUOUS_TO_CONTINUOUS_INDEX = toSet(CONTINUOUS_TO_CONTINUOUS_SCALES);
 
+// FIXME: bin-ordinal is also continuous to discrete
```
kanitw: Not really, the input is already discrete.  In any case, we probably shouldn't leave this FIXME.

src/scale.ts
```diff
@@ -433,14 +416,14 @@ export const defaultScaleConfig: ScaleConfig = {
...
-   * A color scheme name for sequential/ordinal scales (e.g., `"category10"` or `"viridis"`).
+   * A color scheme name for ordinal scales (e.g., `"category10"` or `"blues"`).
    *
    * For the full list of supported schemes, please refer to the [Vega Scheme](https://vega.github.io/vega/docs/schemes/#reference) reference.
    */
   name: string;
 
   /**
-   * For sequential and diverging schemes only, determines the extent of the color range to use. For example `[0.2, 1]` will rescale the color scheme such that color values in the range _[0, 0.2)_ are excluded from the scheme.
+   * For schemes only, determines the extent of the color range to use. For example `[0.2, 1]` will rescale the color scheme such that color values in the range _[0, 0.2)_ are excluded from the scheme.
```
kanitw: This is a property of a scheme param, so I'm not sure if the text "For schemes only," makes sense here.

src/scale.ts
```diff
@@ -489,11 +472,11 @@ export interface Scale {
...
    *
-   * 3) [**Discretizing Scales**](https://vega.github.io/vega-lite/docs/scale.html#discretizing) -- mapping continuous domains to discrete output ranges ([`"bin-linear"`](https://vega.github.io/vega-lite/docs/scale.html#bin-linear), [`"bin-ordinal"`](https://vega.github.io/vega-lite/docs/scale.html#bin-ordinal), [`"quantile"`](https://vega.github.io/vega-lite/docs/scale.html#quantile), [`"quantize"`](https://vega.github.io/vega-lite/docs/scale.html#quantize) and [`"threshold"`](https://vega.github.io/vega-lite/docs/scale.html#threshold).
+   * 3) [**Discretizing Scales**](https://vega.github.io/vega-lite/docs/scale.html#discretizing) -- mapping continuous domains to discrete output ranges [`"bin-ordinal"`](https://vega.github.io/vega-lite/docs/scale.html#bin-ordinal), [`"quantile"`](https://vega.github.io/vega-lite/docs/scale.html#quantile), [`"quantize"`](https://vega.github.io/vega-lite/docs/scale.html#quantize) and [`"threshold"`](https://vega.github.io/vega-lite/docs/scale.html#threshold).
```
kanitw: I think bin ordinal isn't really a discretizing scales because its input values are already discrete. (But let's see what Jeff says on Slack.)

src/axis.ts
```diff
@@ -277,9 +282,7 @@ const COMMON_AXIS_PROPERTIES_INDEX: Flag<keyof (VgAxis | Axis)> = {
 
 const AXIS_PROPERTIES_INDEX: Flag<keyof Axis> = {
   ...COMMON_AXIS_PROPERTIES_INDEX,
-  encoding: 1,
-  labelAngle: 1,
```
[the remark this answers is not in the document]
domoritz: Why was it here in the first place?

src/scale.ts
domoritz: bin-ordinal takes as input quantitative values, I thought.

src/axis.ts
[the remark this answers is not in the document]
kanitw: Ah, I think I was here because we added it because Vega had it, and forgot to remove from here.

src/scale.ts
kanitw: I think you're right.  Why don't you update it so we don't have this FIXME left.

src/config.ts
```diff
@@ -52,7 +52,7 @@ export const defaultViewConfig: ViewConfig = {
 
 export type RangeConfigValue = (number | string)[] | SchemeConfig | {step: number};
 
-export type RangeConfig = RangeConfigProps & {[name: string]: RangeConfigValue};
+export type RangeConfig = RangeConfigProps & {[name in keyof RangeConfigProps]: RangeConfigValue};
```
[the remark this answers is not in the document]
domoritz: Fixed

src/scale.ts
domoritz: I removed the fixme since the domain you have to assign to the scale is still discrete.

src/scale.ts
domoritz: Fixed

domoritz:
Blocked by https://github.com/vega/vega/issues/1641

PR #4569, opened 2019-02-22
Dom/signal ref wrapper

domoritz:


src/compile/signal.ts
```diff
@@ -1,56 +1,33 @@
...
-        // This expression contains a function that needs to be evaluated at runtime, so let's keep it as a SignalRefComponent.
-        return new SignalRefComponent(expr, []);
-      }
-      throw error;
-    }
-  }
+export function evalExpression(expr: string) {
+  const ast = parse(expr);
+  const {code} = generate(ast);
+  const f = new Function('global', `return ${code};`); // tslint:disable-line:function-constructor
```
kanitw: There are some expression that need to be run at runtime, and that will fail this `new Function` thing, so we better keep the try-catch here.

domoritz: Okay. I wonder whether we should just keep the expressions. How much do we gain by compiling them into values?

domoritz: Btw, I think you are wrong. The typings say that we will only ever have numbers as unknown parameters when we eval (for Signals we always generate the expression).

domoritz:
Tests are passing here now.

src/compile/scale/range.ts
```diff
@@ -329,17 +334,18 @@ export function defaultContinuousToDiscreteCount(
...
-    cardinality
-  });
+export function interpolateRange(rangeMin: number, rangeMax: number | SignalRef, cardinality: number): SignalRef {
+  // always return a signal since it's better to compute the sequence in Vega later
+  const f = () => {
+    const rMax = isSignalRef(rangeMax) ? rangeMax.signal : rangeMax;
+    const step = `(${rMax} - ${rangeMin}) / (${cardinality} - 1)`;
+    return `sequence(${rangeMin}, ${rangeMax} + ${step}, ${step})`;
+  };
+  if (isSignalRef(rangeMax)) {
```
kanitw: You should keep `evalOrMakeSignalRefWrapper` (renamed) so you don't need to do this check manually. 

The reason I structure it this way (in the old code) is so that when we support signal for more properties, you don't have to change too much code (as the check for rangeMin / cardinality should come for free too).

src/compile/scale/range.ts
```diff
@@ -382,7 +384,11 @@ function sizeRangeMax(
       }
       const min = minXYRangeStep(xyRangeSteps, config.scale);
 
-      return evalOrMakeSignalRefComponent(`min - 1`, {min});
+      if (isSignalRef(min)) {
```
kanitw: Don't do this manual check.

src/compile/scale/range.ts
```diff
@@ -398,9 +404,11 @@ function sizeRangeMax(
       }
 
       const pointStep = minXYRangeStep(xyRangeSteps, scaleConfig);
-      return evalOrMakeSignalRefComponent(`pow(${MAX_SIZE_RANGE_STEP_RATIO} * pointStep, 2)`, {
-        pointStep
-      });
+      if (isSignalRef(pointStep)) {
```
kanitw: Don't do this manual check.

src/compile/scale/range.ts
domoritz: Why not?

src/compile/scale/range.ts
domoritz: How would this work now that we don't use a regex to replace things in expressions anymore? Can I do this as a follow up? I'll create an issue.

domoritz:
I removed the eval for now. If we think we need it later, we can bring it back.

PR #4571, opened 2019-02-23
[Refactor] Use `joinaggregate` instead of `window` in Boxplots

kanitw:


domoritz:
This should not affect the Vega output.

kanitw:
Well, it doesn't.

Issue #4568, opened 2019-02-22
Wrong whiskers of box plot


akngs:
Box plots (the schematic plot, a.k.a. *Tukey box plot*) rendered by vega-lite have the wrong length of whiskers. This is different issue from already closed one: #3163

It seems like vega-lite draws whiskers using the following formulas:

> Q1 - 1.5 IQR for the lower one and Q2 + 1.5 IQR for the upper one,

while the correct ones are:

> **the smallest data within** Q1 - 1.5 IQR for the lower one and **the largest data within** Q2 + 1.5 IQR for the upper one. (J. Tukey, p44-47, Exploratory Data Analysis)

Here's a detailed explanation and demonstration:

https://beta.observablehq.com/@akngs/vega-lite-renders-box-plots-differently

domoritz:
If this is wrong, we should fix it for the 3.0 release.

kanitw:
Will be fixed in https://github.com/vega/vega-lite/pull/4573

PR #4572, opened 2019-02-23
Move outlier unit to be below the boxplot

kanitw:
This is a step toward https://github.com/vega/vega-lite/issues/4568 as whiskers need to be moved to pair with the "raw" joinaggregate for filtering, and thus they should be below the "box"


domoritz:
Shouldn't outliers never be directly under the box? So this does not affect what the charts look like.

kanitw:
> Shouldn't outliers never be directly under the box? So this does not affect what the charts look like.

Yes, it doesn't affect how chart looks like.  That's why I extract this refactor first. They need to be grouped so we can calculate joinaggregate only once for this outliers group and whiskers group.

PR #4573, opened 2019-02-23
Create correct Tukey boxplot

### LINKED ISSUE #4568: Wrong whiskers of box plot

kanitw:
Fix #4568


kanitw:
For min-max boxplot,  we show 5 summary (and groupbys) for the tooltip of both whisker and box.

image

However, for Tukey boxplot, we can’t access “upper/lower whisker” values for the box as they require a different dataflow for filtering, and it’s not worth joining just for the tooltip.  I currently show upper/lower whisker for the whisker 

image

and 5 summaries for the box

image



The question is should we show min/max in the tooltip for the box of the Tukey box plot?  

(Unlike min-max boxplot that visualize min/max with just the whiskers, the Tukey boxplot represents the min/max using either the outlier points or the whisker if there is no outlier on each end.)

domoritz:
The behavior you describe sounds good to me.

PR #4576, opened 2019-02-23
Change default composition spacing

### LINKED ISSUE #4545: Revise default composition layout spacing
image

image

I think we don't have enough spacing between plots, especially when the axes are independent.

kanitw:
Fix #4545

The original spacing is simply too tight.  See #4545

Ideally, this should be in a config, but we haven't created a config for this yet. (I'll do that later.)


domoritz:
Can we do this after the Vega 5 transition?

kanitw:
Sure, please rebase and re-run the SVG generator. :)

domoritz:
Please rebase so we can see the changes.

kanitw:
From a quick skim -- Most examples clearly look better, except group bar (which shouldn't use facet anyway -- we should do the offset instead).

domoritz:
> From a quick skim -- Most examples clearly look better, except group bar (which shouldn't use facet anyway -- we should do the offset instead).

That's an important example, though. What can we do now until we have offset?

kanitw:
I've a follow-up PR to fix some examples including the group bar chart.

There are a few examples that should be fixed, but let's merge this first.

My main argument for doing this is that with the old spacing, it's sometimes not even clear which plot axis titles belong to:

image

kanitw:
Plus, our default rangeStep is 20 and that's a "within plot" spacing.  It doesn't make sense that between plot spacing is 10 (half of rangeStep).

PR #4579, opened 2019-02-24
[BugFix] Ignore tooltip in `stackBy`

### LINKED ISSUE #4001: Area charts are rendered strangely if tooltip contains string field that contains numbers
Please:
- [x] Provide an example spec in JSON, wrapped by triple backticks like this:

when generating an area chart based on an array of these objects

```json
{
        "date": "2002-02-05",
        "Timepoint": "Exam 3",
        "lesion": "T02",
        "longaxis": 12.5,
        "longaxisDisplayValue": "12.5 mm",
        "shortaxis": 11,
        "status": "Present",
        "valid": true
      }
```
and using either "Timepoint" or "longaxisDisplayValue" as a field in the tooltip, the areas are rendered incorrectly. When not using these fields or explicitly setting type = quantitative (which will result in _NaN_, the areas are rendered correctly. type = nominal does not help.

area rendering should not depend on tooltype content. Displaying strings that contain numbers (and letters) in a tooltip should be possible without affecting the chart

Observable that demonstrates this:
https://beta.observablehq.com/@codetalker78/vega-lite-demo

- [x] If applicable, include error messages and screenshots, GIF videos (e.g. using https://www.cockos.com/licecap/), or working examples (e.g. https://bit.ly/vega-lite-blocks)

kanitw:
Fix #4001

### DIFF
--- a/src/stack.ts
+++ b/src/stack.ts
@@ -106,7 +106,8 @@ export function stack(m: Mark | MarkDef, encoding: Encoding<Field>, stackConfig:
 
   // Should have grouping level of detail that is different from the dimension field
   const stackBy = NONPOSITION_CHANNELS.reduce((sc, channel) => {
-    if (channelHasField(encoding, channel)) {
+    // Ignore tooltip in stackBy (https://github.com/vega/vega-lite/issues/4001)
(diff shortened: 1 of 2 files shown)


domoritz:
Great!

PR #4581, opened 2019-02-24
[Docs] Add missing docs for the header's `title/labelPadding`

kanitw:


src/header.ts
```diff
@@ -85,7 +85,7 @@ export interface HeaderConfig {
   titleLimit?: number;
 
   /**
-   * The orthogonal distance in pixels by which to displace the title from its position along the edge of the chart.
+   * The padding, in pixel, between facet header's title and the label.
```
domoritz: I think orthogonal distance is more meaningful to most people than padding, don't you agree?

kanitw: The new text is more consistent with axis label/titlePadding description. 

If you wanna fix, better fix all of them in another PR.  :)

domoritz: Okay

PR #4583, opened 2019-02-24
Improve colors in some examples

domoritz:


examples/specs/normalized/boxplot_2D_horizontal_color_size_normalized.vl.json
```diff
@@ -129,7 +129,7 @@
             "x2": {"field": "upper_box_people", "type": "quantitative"},
             "y": {"field": "age", "type": "ordinal"},
             "size": {"value": 5},
-            "color": {"value": "blue"},
+            "color": {"value": "steelblue"},
```
kanitw: Teal?  At least make it a bit more different from the default color

domoritz: Just this one?

PR #4584, opened 2019-02-24
Don't generate type for secondary field defs. Test overlay specs. Rem…

domoritz:


kanitw:
Is the explicit aggregate spec listed in the docs anywhere? — wanna make sure we dont remove by mistake.

domoritz:
We never reference the file.

PR #4591, opened 2019-02-24
Fix examples spacing 

kanitw:


domoritz:
Builds are failing because the examples are not valid.

PR #4593, opened 2019-02-24
Fix examples

domoritz:


kanitw:
I guess these are simply outdated due to rebasing?

domoritz:
I guess so. This should fix master. Please squash the commits.

domoritz:
We'll need another commit fro travis to update the SVGs as well.

PR #4595, opened 2019-02-25
Wrap Docs

kanitw:
Docs

- [x] encoding.html -- macro
- [x] unit.html -- add layout
- [x] facet.html
- [x] concat.html
- [x] repeat.html


kanitw:
This should be quite a good start.  I'll be busy during the week. To speed up, feel free to edit this directly.

examples/specs/trellis_barley.vl.json
```diff
@@ -4,8 +4,9 @@
   "description": "The Trellis display by Becker et al. helped establish small multiples as a “powerful mechanism for understanding interactions in studies of how a response depends on explanatory variables”. Here we reproduce a trellis of Barley yields from the 1930s, complete with main-effects ordering to facilitate comparison.",
   "data": {"url": "data/barley.json"},
   "mark": "point",
+  "columns": 2,
```
domoritz: I don't know whether I like this version more as it makes it much harder to compare the different sites. 6 is not that bad to show in a column.

Issue #2544, opened 2017-06-21
Better handling of invalid values


kanitw:
- [ ] Show points with invalid values with grey color instead of filtering them.  (This can be a new propery value for config.invalidValues https://github.com/vega/vega-lite/pull/2539.)

- [ ] We can even extract a special region to show invalid values along side continuous scales (as discussed [here](https://github.com/vega/vega-lite/issues/1548#issuecomment-275315485)).

- [ ] Similar rules can be apply for log scales that we need to filter non positive values.  (In fact we should integrate `NullFilter` and `NonPositiveFilter` as their purpose are the same, it's just that log scale require different treatment.

kanitw:
We need to consider when a specific channel is invalid vs when any of the channels is invalid.

kanitw:
- [ ] When the invalid values are not filtered, they are interpret by the scales as `0`.  This can be problematic when the scale has `zero=false`.

kanitw:
[The spec in this comment](https://github.com/vega/vega-lite/pull/4405#issuecomment-450028617) highlights that not filtering null can be problematic.

kanitw:
Some challenges of the top of my head:

### Positions

- For example, while point is straightforward, line marks should use either point on the margins or have the default `strokeCap` to be `"round"`.  
  - It's also unclear what to do for area and range marks that have only one end missing (e.g., valid x but invalid x2).  (I guess we should probably still filter for range mark case.)   

  - A good implementation for VL should also dynamically hide the margins when there is no null value (so we can have the marginal null on by default without being a breaking change.)  

### Continuous Non-Position 

- If there is color encoding (e.g., for field `"c"`), do we make all points with null values grey or just the points that has invalid `"c"`? (Probably the latter.)  
- What should we do for size channel?

- What do we do for opacity? 


### Discrete Non-Position 

- For categorical color, how do we make sure the categorical palette do not include grey by default and so we can use grey to encode null?  (Do we add tableau10/20withoutGrey to default palettes?)

- For shape and the future stroke dash, should we assign null to certain shape / stroke dash by default to consistently encode nulls.

PR #4605, opened 2019-02-25
Handling broken layout due to hidden null value appears beyond the view group

### LINKED ISSUE #4601: Apply conditional encoding to make sure null values are not off the view group
Null values are no longer filtered, and thus can be off-chart when the axis does not contain zero.  

The trellis histogram is one example that demonstrates this problem. 

image

kanitw:
Merge #4604 first.  

Fix #4601


kanitw:
@domoritz I actually just realized that another solution is to turn "clamp" on by default.

kanitw:
This is a bit more complicated for the case needed, but will affect fewer examples.  Changing clamp will affect all examples.    Any preference?

domoritz:
If you zoom, clamp won't fix this issue, right?

kanitw:
> If you zoom, clamp won't fix this issue, right?

It will, but basically, you cannot turn on clamp without facing this problem anymore.

So I think it's better to do this predicate things so people still have freedom to use clamp or not.  

Plus, some code here would be a good infrastructure for even better null handling https://github.com/vega/vega-lite/issues/2544.

kanitw:
image

src/compile/scale/component.ts
```diff
@@ -24,6 +25,20 @@ export class ScaleComponent extends Split<ScaleComponentProps> {
     );
     this.setWithExplicit('type', typeWithExplicit);
   }
+
+  /**
+   * Whether the scale definitely includes zero in the domain
+   */
+  public get domainDefinitelyIncludeZero() {
```
domoritz:
```suggestion
  public get domainDefinitelyIncludesZero() {
```

src/vega.schema.ts
```diff
@@ -68,6 +68,7 @@ export interface VgValueRef {
   mult?: number;
   offset?: number | VgValueRef;
   band?: boolean | number | VgValueRef;
+  test?: string;
```
domoritz: Can we use the type from Vega instead to avoid problems like this in the future?

src/compile/mark/valueref.ts
```diff
@@ -88,13 +145,22 @@ export function position2({
...
+    defaultRef
+  });
 }
 
 export function getOffset(channel: 'x' | 'y' | 'x2' | 'y2', markDef: MarkDef) {
-  const offsetChannel = channel + 'Offset';
-  // TODO: in the future read from encoding channel too
+  const offsetChannel = (channel + 'Offset') as 'xOffset' | 'yOffset' | 'x2Offset' | 'y2Offset'; // Need to cast as the type can't be inferred automatically
 
+  // TODO: in the future read from encoding channel too
```
domoritz: Why not now?

kanitw: There is no offset channel

src/vega.schema.ts
kanitw: Yes, but it’s beyond the scope of this PR.  Please send a follow up fix.

domoritz:
Cool, lgtm

PR #4609, opened 2019-02-27
Update for Vega 5 RC 4

domoritz:


build/vega-lite-schema.json
```diff
@@ -4321,7 +4641,16 @@
           "type": "string"
         },
         {
-          "$ref": "#/definitions/RepeatRef"
+          "$ref": "#/definitions/SignalRef"
```
kanitw: Why is this changing here? 


cc: @domoritz

domoritz: Hmm, idk. Have a look at the types.

kanitw: It's probably due to a bug in the schema generator that often include types that we don't really need and it may accidentally override Field. The type for Field in Vega-Lite has not changed. It's just overrided by Vega's in the schema.

domoritz: I think producing unused refs and getting the wrong type are different.

kanitw: Yes, but they might be related.  I think what happened is the generator produces unused refs AND have no mechanism for dealing with name collision.

domoritz: Hmm, name collisions could but should not be a problem but I would have to look more into it.

PR #4613, opened 2019-02-27
Fix incorrect labelAlign in spec with top orient

### LINKED ISSUE #4491: Labels in top oriented axis is aligned incorrectly
```json
{
  "data": {"values": [
    {"source": 0, "destination": 0},
    {"source": 1, "destination": 1},
    {"source": 2, "destination": 2}
  ]},
  "mark": "rect",
  "encoding": {
    "y": {
      "field": "source",
      "title": "Source Server",
      "type": "ordinal"
    },
    "x": {
      "field": "destination",
      "title": "Destination Server",
      "type": "ordinal",
      "axis": {
        "orient": "top"
      }
    },
    "color": {
      "value": "indigo"
    }
  }
}
```

image

sirahd:
- fix #4491

### DIFF
--- a/examples/compiled/bar_layered_weather.vg.json
+++ b/examples/compiled/bar_layered_weather.vg.json
@@ -284,7 +284,7 @@
       "labels": false,
       "ticks": false,
       "titlePadding": 25,
-      "labelAlign": "right",
+      "labelAlign": "left",
       "labelAngle": 270,
       "labelBaseline": "middle",
       "labelOverlap": true,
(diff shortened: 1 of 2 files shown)


domoritz:
Can you add a few tests?

jheer:
FYI, with the new Vega v5 `titleOrient` (legends) and `titleAnchor` (legends/axes) options, I've made changes to the Vega config and also have more logic for automatic determination of alignment and baseline when they are not explicitly specified. I don't know if that is relevant here, but at least wanted to call your attention to it, as the solution to some layout problems may actually be "do nothing at all".

domoritz:
Oh, so we should see whether this new Vega logic is as smart as the logic we have currently implemented in Vega-Lite and maybe drop it entirely. @sirahd could you look into this? 

I also saw that this PR already has 100% test coverage so don't worry too much about tests for now.

domoritz:
Let's revise this after this PR.

domoritz:
See https://github.com/vega/vega-lite/issues/4632

PR #4619, opened 2019-03-01
[Refactors] for bar/rect/binPosition

kanitw:


domoritz:
I see code deletions. Nice!

PR #4621, opened 2019-03-02
Make `x`, `y`, `x2`, `y` work in mark definition block and config

kanitw:
Why?: Even I can't remember that we don't support this. 

Merge https://github.com/vega/vega-lite/pull/4605 first


kanitw:
Thanks, will merge as part of #4605

PR #4626, opened 2019-03-02
Redirect `titleAnchor` to the underlying `layout.titleBand` + Automatically adjusting titleAlign

### LINKED ISSUE #4624: Redirect header's titleAnchor to layout titleBand
As title block doesn't have explicit width, setting the "anchor" has no effect.  
We should instead redirect this property to layout's `titleBand` instead.

kanitw:
Fix #4624

### DIFF
--- a/src/compile/header/index.ts
+++ b/src/compile/header/index.ts
@@ -101,7 +101,7 @@ export function assembleTitleGroup(model: Model, channel: FacetChannel) {
   };
 }
 
-function titleAlign(titleAnchor: TitleAnchor) {
+export function titleAlign(titleAnchor: TitleAnchor) {
   switch (titleAnchor) {
     case 'start':
       return {align: 'left'};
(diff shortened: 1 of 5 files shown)


src/compile/header/index.ts
```diff
@@ -87,11 +95,23 @@ export function assembleTitleGroup(model: Model, channel: FacetChannel) {
...
 }
 
+export function titleAlign(titleAnchor: TitleAnchor) {
+  switch (titleAnchor) {
+    case 'start':
+      return {align: 'left'};
+    case 'end':
+      return {align: 'right'};
+  }
+  // TODO: take TitleAngle into account for the "middle" case
```
domoritz: Why not do this?

kanitw: Much more testing, beyond the scope of this PR.

domoritz: Okay

PR #4636, opened 2019-03-03
Facet wrap with shared axes

kanitw:


examples/specs/trellis_barley_shared.vl.json
```diff
@@ -0,0 +1,29 @@
...
+    },
+    "y": {
+      "field": "variety",
+      "type": "ordinal",
+      "sort": {"encoding": "x", "order": "descending"},
+      "scale": {"rangeStep": 12}
+    },
+    "color": {"field": "year", "type": "nominal"}
+  },
+  "resolve": {"axis": {"x": "shared", "y": "shared"}}
```
domoritz: Should this be the default?

src/compile/data/facet.ts
```diff
@@ -182,12 +184,65 @@ export class FacetNode extends DataFlowNode {
...
+            : {signal: cardinality};
+
+        data.push({
+          name: `${this.facet.name}_${headerChannel}`,
+          transform: [
+            {
+              type: 'sequence',
+              start: 0,
+              stop
+            } as any // TODO: typing is wrong
```
domoritz: How so? It should be any in the vega-typings.

src/compile/header/assemble.ts
```diff
@@ -29,6 +29,7 @@ import {
   LayoutHeaderComponentIndex
 } from './component';
 
+// TODO: rename to assembleHeaderTitleGroup
```
domoritz: 👍

src/vega.schema.ts
```diff
@@ -392,7 +392,15 @@ export type VgTransform =
   | VgWindowTransform
   | VgJoinAggregateTransform
   | VgFoldTransform
-  | VgSampleTransform;
+  | VgSampleTransform
+  | VgSequenceTransform;
+
+export interface VgSequenceTransform {
```
domoritz: Lol, you added the typings here. Why make them wrong?

PR #4639, opened 2019-03-03
Promoting @manzt's top-K with others to be in the main gallery

kanitw:
cc: @manzt


domoritz:
Mean is a bit odd here. I added a title to make it at least a bit less confusing.

PR #4642, opened 2019-03-04
Fix #4586: header smart align

### LINKED ISSUE #4586: Smart alignment based on angle for header
(We already do this for axis, but not for header.)

EvsChen:
Fix #4586


domoritz:
Do you have an example? Can we make this simpler by not having cases for the different channels?

kanitw:
I'm wondering if there is a general rule that we can use across header / axis  (and basically if we know orient and angle, we can do some math like rotating orient left by 90 degree to reuse the same logic for orient:top)

domoritz:
I'm working on this in a separate branch.

domoritz:
Replaced by https://github.com/vega/vega-lite/pull/4658

PR #4644, opened 2019-03-05
Use test environment without jsdom.

domoritz:


domoritz:
Doesn't seem to make a difference so whatever.

PR #4651, opened 2019-03-05
Update to Vega 5

domoritz:


domoritz:
Needs https://github.com/vega/vega/pull/1669

PR #4653, opened 2019-03-05
Update author and collaborators in package.json.

jheer:
It was missing Arvind!


domoritz:
Good catch!

arvind:
Thanks for catching that :)

PR #4656, opened 2019-03-05
Don't set the minExtent for all examples and instead document how and when to use it. 

domoritz:


jheer:
Why not? IIRC the issue is that selections can result in changes to axis label lengths which in turn leads to "jumpy" distracting axis titles. Maybe it makes sense to not set minExtent by default, but include it when selections that filter data or modify scale domains are used? (And sorry if I'm missing other context here!)

domoritz:
This PR was just an experiment to see what it changes. We have not made a decision here. 

There are two reasons. First, we want more title offset than the default and second jumping titles. I don't think min extent is a good solution for either of these issues. For the first, we should instead set the title offset explicitly. For the second issue, a min extent does not help either when titles are longer than the minimum.

jheer:
Thanks! I totally agree with reason 1: if you want more title offset you can set that directly. Reason 2 is more murky. Using `minExtent` is an imperfect solution, but was added because it helped improve the situation for a reasonable range of labels. Moreover, it can be infeasible to predict how long labels may become due to interaction, so I doubt a "perfect" solution is even  achievable. Again, perhaps a good improvement is for minExtent to be set dynamically, and only when a selection is involved.

kanitw:
> Again, perhaps a good improvement is for minExtent to be set dynamically, and only when a selection is involved.

Is it better to set `minExtent` when selection filter is used or better to use do nothing and document that “users” might find “minExtent”  handy for jumping axes?

kanitw:
(In any case, I agree that setting minExtent to _all_ plots is a bit overkill.)

domoritz:
As discussed on slack, if we set it based on whether there is a selection or not, then the chart looks different if you have a selection, which is odd. I am for documenting it and eventually finding a solution in Vega. 

One such solution could be to only grow the extent and never shrink, for example.

kanitw:
> One such solution could be to only grow the extent and never shrink, for example.

That would still be jumpy when you first interact with the plot, which isn't great either? 

> As discussed on slack, if we set it based on whether there is a selection or not, then the chart looks different if you have a selection, which is odd.

I think inconsistency when transition between specs is bad, but I think that jumpy plots whenever there is a selection filter is worse.

kanitw:
Thinking more, the problem with dynamically setting `minExtent` when a selection filter is used is that it involves quite a complex code for propagating the information that a selection filter is used.  So perhaps, documenting it for now might be better.

PR #4659, opened 2019-03-05
Remove outdated prebuild

domoritz:


kanitw:
What if you just clone the repo, does the build still work?

domoritz:
Yes, it does. Now travis tests also pass again.

PR #4663, opened 2019-03-06
Set minExtent for jumpy examples

kanitw:
There might be others, but I checked all examples with selection bind and filter.


domoritz:
I think we should instead set it for the encodings where it matters as discussed in the docs.

domoritz:
Replaced by https://github.com/vega/vega-lite/pull/4667

PR #4666, opened 2019-03-06
Make Vega a peer dependency

domoritz:
This way users will know which version of vega we expect but we don't force anyone to install Vega (you can ignore the warning).


jheer:
Sounds reasonable to me.

domoritz:
Note that I also changed Vega-Embed 4 to use peer dependencies instead of depending directly on Vega or Vega-Lite.

domoritz:
@tafsiri mentioned that it would be good to specify compatible versions with peer dependencies.

domoritz:
@kanitw @arvind I would like to merge this. Any concerns?

package.json
```diff
@@ -117,6 +117,9 @@
     "vega-util": "^1.9.0",
     "yargs": "^13.2.2"
   },
+  "peerDependencies": {
+    "vega": "^5.1.0"
```
kanitw: This version of Vega-Lite doesn't work with 5.2?  If so, remove ^?

domoritz: Good point.

domoritz:
I switched to ~ for all dependencies. I think that's a good idea.

domoritz:
We're pretty good about making new releases and updating dependencies so that using ~ instead of ^ will lead to fewer headaches.

PR #4669, opened 2019-03-06
[BugFix] Include necessary transform needed for join-aggregation for filtered layers

### LINKED ISSUE #4524: Unexpected offset of boxplots when filtering aggregated temporal data
Using the Boxplot mark when aggregating and filtering temporal data produces an unexpected  chart offset that varies depending on how far after 1970 values are filtered out. In the example below try changing `1985` to a later or earlier year.

The fact this only occurs beyond 1970 suggests there may be a time parsing issue somewhere. If the mark is changed to a primitive such as `circle`, there is no offset. 

```javascript
{
  "width": 600,
  "data": {
    "url": "https://gicentre.github.io/data/temperatureAnomalies.json",
    "format": {"parse": {"Anomaly": "number"}}
  },
  "transform": [{"filter": "year(datum.Date) >= 1985"}],
  "encoding": {
    "x": {
      "field": "Date",
      "type": "temporal",
      "timeUnit": "year"
    },
    "y": {"field": "Anomaly", "type": "quantitative"}
  },
  "mark": "boxplot"
}
```

chanwutk:
This fixes the offset the problem when filtering boxplot with `timeUnit`
Fix #4524

### DIFF
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -1,5 +1,5 @@
 import {Orientation} from 'vega';
-import {isNumber, isObject} from 'vega-util';
+import {array, isNumber, isObject} from 'vega-util';
 import {Config} from '../config';
 import {Encoding, extractTransformsFromEncoding} from '../encoding';
 import {PositionFieldDef} from '../fielddef';
@@ -106,7 +106,8 @@ export function normalizeBoxPlot(
     continuousAxis,
     groupby,
     encodingWithoutContinuousAxis,
-    ticksOrient
+    ticksOrient,
+    preJoinAggregate
   } = boxParams(spec, extent, config);
 
   const {color, size, ...encodingWithoutSizeColorAndContinuousAxis} = encodingWithoutContinuousAxis;
@@ -289,11 +290,13 @@ export function normalizeBoxPlot(
   }
 
   if (filteredLayersMixins) {
+    filteredLayersMixins.transform.unshift(...preJoinAggregate);
+
     // tukey box plot with outliers included
     return {
       ...outerSpec,
       layer: [
-        ...(filteredLayersMixins ? [filteredLayersMixins] : []),
+        ...[filteredLayersMixins],
         {
           // boxplot
           transform,
@@ -335,6 +338,7 @@ function boxParams(
   continuousAxis: 'x' | 'y';
   encodingWithoutContinuousAxis: Encoding<string>;
   ticksOrient: Orientation;
+  preJoinAggregate: Transform[];
 } {
   const orient = compositeMarkOrient(spec, BOXPLOT);
   const {continuousAxisChannelDef, continuousAxis} = compositeMarkContinuousAxis(spec, orient, BOXPLOT);
@@ -389,6 +393,14 @@ function boxParams(
 
   const ticksOrient: Orientation = orient === 'vertical' ? 'horizontal' : 'vertical';
 
+  const preJoinAggregate: Transform[] = [];
+  for (const transform of [...bins, ...timeUnits, ...postAggregateCalculates]) {
+    const transformOutputs: string[] = array(transform.as);
+    if (transformOutputs.some(transformOutput => groupby.includes(transformOutput))) {
+      preJoinAggregate.push(transform);
+    }
+  }
+
   return {
     transform: [
       ...bins,
@@ -403,6 +415,7 @@ function boxParams(
     continuousAxisChannelDef,
     continuousAxis,
     encodingWithoutContinuousAxis,
-    ticksOrient
+    ticksOrient,
+    preJoinAggregate
   };
 }
--- a/test/compositemark/boxplot.test.ts
+++ b/test/compositemark/boxplot.test.ts
@@ -915,4 +915,35 @@ describe('normalizeBoxIQR', () => {
       }
     });
   });
+
+  it('should include necessary transform for the groupby fields of joinaggregate in filtered layers', () => {
+    const normalizedSpec = normalize(
+      {
+        $schema: 'https://vega.github.io/schema/vega-lite/v3.json',
+        data: {
+          url: 'data/cars.json'
+        },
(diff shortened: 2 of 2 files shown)


src/compositemark/boxplot.ts
```diff
@@ -309,6 +313,16 @@ export function normalizeBoxPlot(
   };
 }
 
+function extractDependencyTransformsOfGroupby(transforms: Transform[], groupby: string[]): Transform[] {
```
domoritz: What is a dependency transform?

chanwutk: With this spec
```
{
  "data": {
    "url": "data/cars.json"
  },
  "mark": "boxplot",
  "encoding": {
    "y": {
      "field": "Miles_per_Gallon",
      "type": "quantitative"
    },
    "x": {
      "field": "Year",
      "type": "temporal",
      "timeUnit": "year"
    }
  }
}
```
Vega-Lite does not add timeUnit transform `{"timeUnit": "year", "field": "Year", "as": "year_Year"}` in to the `transform` for `filteredLayers`, which includes outliers and whiskers or boxplot. However, the joinaggregate transform in the `filteredLayers` has groupby `['year_Year']`. So the joinaggregate is depending on the timeUnit transform.

This is the normalized spec before including the timeUnit transform to the `filteredLayers`
```
{
  "layer": [
    {
      "transform": [
        {
          "joinaggregate": [
            {
              "op": "q1",
              "field": "Miles_per_Gallon",
              "as": "lower_box_Miles_per_Gallon"
            },
            {
              "op": "q3",
              "field": "Miles_per_Gallon",
              "as": "upper_box_Miles_per_Gallon"
            }
          ],
          "groupby": ["year_Year"]
        }
      ],
      "layer": [
        {
          "transform": [
            {
              "filter": "(datum.Miles_per_Gallon < datum.lower_box_Miles_per_Gallon - 1.5 * (datum.upper_box_Miles_per_Gallon - datum.lower_box_Miles_per_Gallon)) || (datum.Miles_per_Gallon > datum.upper_box_Miles_per_Gallon + 1.5 * (datum.upper_box_Miles_per_Gallon - datum.lower_box_Miles_per_Gallon))"
            }
          ],
          "mark": {"type": "point", "style": "boxplot-outliers"},
          "encoding": {
            "x": {
              "title": "Year (year)",
              "type": "temporal",
              "field": "year_Year",
              "axis": {"format": "%Y"}
            }
          }
        },
        {
          ... whiskers of boxplot
        }
      ]
    },
    {
      ... box and mid tick of the boxplot
    }
  ]
}
```

kanitw:
Please don't include feature name in brackets in the title.  We only use brackets for [Refactor], [BugFix].

src/compositemark/boxplot.ts
```diff
@@ -286,6 +286,10 @@ export function normalizeBoxPlot(
       filteredLayersMixins = filteredWhiskerSpec;
       filteredLayersMixins.transform.unshift(joinaggregateTransform);
     }
+    filteredLayersMixins.transform = [
```
kanitw: `filteredLayersMixins` can be undefined here --  better streamline the logic a bit more.   Plus, you can just call `unshift` like above.

src/compositemark/boxplot.ts
kanitw: Instead of extracting it here, which is a hacky solution, better refactor `boxParams` to provide this kind of information from the beginning.

test/compositemark/boxplot.test.ts
```diff
@@ -915,4 +915,35 @@ describe('normalizeBoxIQR', () => {
       }
     });
   });
+
+  it('should include necessary transform for joinaggregating in filtered layers', () => {
```
kanitw: should include necessary transform for the groupby fields of joinaggregate in filtered layers

src/compositemark/boxplot.ts
kanitw: In fact, better just include this as a part of all joinAggregateTransform above -- something like: 

`unshift(preJoinAggregate, joinAggregate)`

chanwutk: I think `unshift(preJoinAggregate, joinAggregate)` might not work in this case because `preJoinAggregate` should be an array since `joinAggregate` might depend on more than one transform.

kanitw: `unshift(...preJoinAgg, joinAgg)`

chanwutk: Ok, thank you.

src/compositemark/boxplot.ts
```diff
@@ -389,6 +393,14 @@ function boxParams(
 
   const ticksOrient: Orientation = orient === 'vertical' ? 'horizontal' : 'vertical';
 
+  const preJoinAggregate: Transform[] = [];
+  for (const transform of [...bins, ...timeUnits, ...postAggregateCalculates]) {
+    const transformOutputs: string[] = isArray(transform.as) ? transform.as : [transform.as];
```
domoritz: Vega has `utils.array` for this. Let's use it.

src/compositemark/boxplot.ts
```diff
@@ -289,11 +290,13 @@ export function normalizeBoxPlot(
   }
 
   if (filteredLayersMixins) {
+    filteredLayersMixins.transform.unshift(...preJoinAggregate);
```
kanitw: Instead of adding duplicates of all of these transform to `filteredLayersMixins` 

Can we restructure this such that we only do timeUnit/ binning computation only once?

chanwutk: Do you mean putting the transform outside layer or combining the filtered layer and the boxplot together?

kanitw: Yes, that should be possible?

Otherwise, we're always running the computation multiple times, which seems suboptimal.

kanitw:
Since we should change the approach, let's close this for now.

PR #4682, opened 2019-03-07
Fix double invocation of interval selections initialized on x&y.

arvind:
See vega/vega#1672 for an explanation on what was causing the double invocation. Also added a comment to explain why we can't use a straightforward `value` + `update` as @jheer suggests -- initial values may be (or may contain) datetime objects that we need to parse via datetime expression functions.


arvind:
Not sure why Node v11 tests are failing with the following error message. As far as I can tell, these messages are unrelated to the change introduced in this PR:

```
FAIL test/compositemark/errorbar.test.ts
  ● Test suite failed to run
    TypeError: Cannot assign to read only property 'Symbol(Symbol.toStringTag)' of object '#<process>'
      at _default (node_modules/jest-util/build/createProcessObject.js:85:34)
      at new PuppeteerEnvironment (node_modules/jest-environment-puppeteer/lib/PuppeteerEnvironment.js:28:1)
```

jheer:
FWIW Vega is not testing on Node 11 at the moment. IIRC, some of the Vega tests also break under Node 11 due to changes that result in subtly different floating-point coordinate values (perhaps in Math functions? That's happened before...). I was going to wait until Node 12 to investigate further; see https://nodejs.org/en/about/releases/

arvind:
Rebased on master and tests on all node versions are now passing!

src/compile/selection/interval.ts
```diff
@@ -85,7 +85,7 @@ const interval: SelectionCompiler<'interval'> = {
       ...(init ? {init: `{${update}: ${assembleInit(init)}}`} : {}),
       on: [
         {
-          events: [{signal: dataSignals.join(' || ')}], // Prevents double invocation, see vega/vega#1672.
+          events: [{signal: dataSignals.join(' || ')}], // Prevents double invocation, see https://github.com/vega/vega#1672.
```
arvind: 👍

PR #4688, opened 2019-03-08
New example: `layer_line_mean_point_raw`

kanitw:
(My colleague was asking how to do this earlier today and I realize there isn't an example yet)


examples/specs/layer_line_mean_point_raw.vl.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v3.json",
+  "description": "Plot showing average data with raw values in the background.",
+  "data": {"url": "data/stocks.csv"},
+  "transform": [{"filter": "datum.symbol==='GOOG'"}],
+  "layer": [{
+    "mark": {"type": "point", "opacity": 0.3},
```
domoritz: It seems a bit low, no?

kanitw: I want it to be more like a background thing so I make it a bit low.

Issue #4575, opened 2019-02-23
Make selection init work without projection


kanitw:
(Copied from Slack)

From https://github.com/vega/vega-lite/pull/4139, it seems like init only works when there is a selection projection (`fields` or `encodings`).
(At least I see that the `init` is only initialized in the project transform compiler.)

I wonder if `init` should actually work as the `datum` object when there is no projection.  ~~(Basically, default to “field” when none of the projection type is defined.)~~

This is not blocking for 3.0 though.  I also haven’t looked deeply enough if this is even feasible in Vega, but since it’s not pressing for 3.0, I’ve stopped looking. 

cc: @arvind

arvind:
When not projected over and fields or encodings, discrete selections works by using an internal `__vgsid__` field that is calculated via an identifier transform. User's don't have access to their field apriori, and the identifier transform maintains a global counter so consistent IDs are not guaranteed between re-parsing.

I'm not sure what you mean by "default to 'field'".

kanitw:
Oh right, I forgot that we use the internal `__vgsid__` field for syncing discrete selection.  (Thus, please ignore the "default to 'field'" part.)

Sounds like we can't really support this then. 

Perhaps, it's worth throwing a warning that `init` must be used with a selection projection? 

[...]

PR #4691, opened 2019-03-08
Infer selection projection from initial value if none given.

arvind:
Addresses #4575. If no explicit projection is provided for a selection, it first looks to the initial value and _then_ uses the default configuration values. As before, the initial value first attempts to interpret initial value keys as encoding channels, and then as field names.

### DIFF
--- a/src/compile/selection/transforms/project.ts
+++ b/src/compile/selection/transforms/project.ts
@@ -1,6 +1,6 @@
 import {isArray} from 'vega-util';
 import {SelectionComponent} from '..';
-import {ScaleChannel, SingleDefChannel} from '../../../channel';
+import {isSingleDefUnitChannel, ScaleChannel, SingleDefUnitChannel} from '../../../channel';
 import * as log from '../../../log';
 import {hasContinuousDomain} from '../../../scale';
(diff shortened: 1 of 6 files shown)


src/compile/selection/transforms/project.ts
```diff
@@ -56,6 +55,31 @@ const project: TransformCompiler = {
       return {[range]: signals[sg] = sg};
     };
 
+    // If no explicit encodings are specified, set some defaults.
```
kanitw: if not explicit "projection (either fields or encoding)"?

src/compile/selection/transforms/project.ts
```diff
@@ -56,6 +55,31 @@ const project: TransformCompiler = {
...
+    // If an initial value is set, try to infer projections for more concise specification.
```
kanitw: > for more concise specification.

Is it just more concise?  I thought we need this to make it init work at all?

src/compile/selection/transforms/project.ts
arvind: Fixed.

src/compile/selection/transforms/project.ts
arvind: Depends how you parse the sentence. But I remove the clause to avoid confusion and further wordsmithing.

Issue #4683, opened 2019-03-07
Vega module not found in webpack build


erikjchan:
This is a similar issue to #4339. In fact, it is almost identical. Webpack builds using vega-lite are failing.

The errors I get are:

ERROR in ./node_modules/vega-lite/build/src/compile/repeat.js
Module not found: Error: Can't resolve 'vega' in '/Users/ndb071/Desktop/ve-test/node_modules/vega-lite/build/src/compile'
 @ ./node_modules/vega-lite/build/src/compile/repeat.js 1:0-31 17:22-29 18:25-32 19:30-37 43:24-31
 @ ./node_modules/vega-lite/build/src/compile/buildmodel.js
 @ ./node_modules/vega-lite/build/src/compile/compile.js
 @ ./node_modules/vega-lite/build/src/index.js
 @ ./index.js

ERROR in ./node_modules/vega-lite/build/src/compile/scale/component.js
Module not found: Error: Can't resolve 'vega' in '/Users/ndb071/Desktop/ve-test/node_modules/vega-lite/build/src/compile/scale'
 @ ./node_modules/vega-lite/build/src/compile/scale/component.js 1:0-31 21:12-19 22:38-45
 @ ./node_modules/vega-lite/build/src/compile/scale/parse.js
 @ ./node_modules/vega-lite/build/src/compile/model.js
 @ ./node_modules/vega-lite/build/src/compile/layer.js
 @ ./node_modules/vega-lite/build/src/compile/buildmodel.js
 @ ./node_modules/vega-lite/build/src/compile/compile.js
 @ ./node_modules/vega-lite/build/src/index.js
 @ ./index.js

ERROR in ./node_modules/vega-lite/build/src/spec/base.js
Module not found: Error: Can't resolve 'vega' in '/Users/ndb071/Desktop/ve-test/node_modules/vega-lite/build/src/spec'
 @ ./node_modules/vega-lite/build/src/spec/base.js 1:0-31 26:35-42
 @ ./node_modules/vega-lite/build/src/config.js
 @ ./node_modules/vega-lite/build/src/compile/compile.js
 @ ./node_modules/vega-lite/build/src/index.js
 @ ./index.js

Perhaps this is related to vega no longer being in devDependencies?

domoritz:
Vega-Lite depends on Vega CLI which depends on Vega, no?

domoritz:
Actually, only for Dev. Vega-Lite should not use Vega except for the typings. Is this the bug?

[...]

Issue #4679, opened 2019-03-07
Vega is not in devDependencies anymore


nohanna:
Hello,

It is not really a bug but there is no Vega dependency in the devDependencies anymore.
In this commit 596bb5544c8d40969d9f64618e07e76d23cf21b4, dependency was removed.

So I've got a compilation error:
`ERROR in node_modules/vega-embed/build/src/embed.d.ts(1,29): error TS2307: Cannot find module 'vega'.`

Should I add Vega dependency myself now?

domoritz:
We depend on vega-cli, which depends on Vega.

nohanna:
Indeed, vega-cli depends on Vega.
This is weird, I'm getting this compilation error since 2-3 days so I had to depend on Vega.

domoritz:
There is probably a place in Vega-Lite where we wrongly depend on Vega.

PR #4692, opened 2019-03-09
 Throw an error when we import node util or vega

domoritz:
This makes sure #4679 and #4683 never happen again.


kanitw:
:tada:

domoritz:
The nice thing is that you can still important types from Vega but as soon as you important something else, it will fail.

PR #4698, opened 2019-03-09
add support to axis' time format in composite mark (use Vega 5.1's formatType)

chanwutk:


domoritz:
Tests are failing.

kanitw:
Note: timeFormat for Vega doesn't really support quarter

domoritz:
> Note: timeFormat for Vega doesn't really support quarter

:-(

@jheer do you think it's worth supporting it in Vega? If yes, we probably want to do it in D3, right?

jheer:
Yes, we use d3-time-format directly so that would be the ideal place to do it.

kanitw:
FYI, I created https://github.com/d3/d3-time-format/issues/56 and Mike has some thoughts about deprecating the library.

jheer:
Thanks @kanitw! Maybe Vega 6 can move to browser built-in’s for date formatting and parsing. It would be wonderful to have international locales out-of-the-box!!

In the meantime, however, we might consider forking d3-time-format if a PR for quarter suppport will not be accepted?

jheer:
FYI it also looks like Date.toLocaleString does not support quarters... ☹️

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

domoritz:
@chanwutk What's left here?

chanwutk:
I am trying to use pass format and formatType from Vega-Lite to Vega directly. But since Vega hasn't supported quarter yet, aren't we going to use the old implementation (format time with `encode`) #4190 ?

domoritz:
Yes

kanitw:
I guess we don't need this anymore.

PR #4701, opened 2019-03-09
Add config.axisX/Y.orient

### LINKED ISSUE #4648: Support axis.orient in the config
In Vega, config.axisX/Y.orient doesn't make sense as we determine the axis properties from the orient.  

However, in Vega-Lite, we can read from channels and apply the orient accordingly.

kanitw:
Note that I don't distinguish types for config.axis from config.axisX/Y even though orient doesn't make sense for the general axis because I don't want to make the schema too complicated, which in turn will make Altair APIs too complicated.

Also refactor getAxisConfig to use `...`. 

Fix #4648

cc: @evschen -- I was fixing similar things in header and found that this is quite easy for me to fix so I went ahead and fix it to save your time.

image

### DIFF
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -59,16 +59,18 @@ interface AxisMixins {
   labelOverlap?: LabelOverlap;
 }
 
-export type AxisConfig = VgAxisConfigNoSignals & VlOnlyGuideConfig;
-
-export interface Axis extends BaseAxisNoSignals, Guide {
+export interface AxisOrientMixins {
   /**
-   * The orientation of the axis. One of `"top"`, `"bottom"`, `"left"` or `"right"`. The orientation can be used to further specialize the axis type (e.g., a y axis oriented for the right edge of the chart).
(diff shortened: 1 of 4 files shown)


EvsChen:
Thanks @kanitw :)

src/axis.ts
```diff
@@ -48,16 +48,18 @@ interface AxisMixins {
   labelOverlap?: LabelOverlap;
 }
 
-export type AxisConfig = VgAxisConfigNoSignals & VlOnlyGuideConfig;
-
-export interface Axis extends BaseAxisNoSignals, Guide {
+export interface AxisOrientMixins {
   /**
-   * The orientation of the axis. One of `"top"`, `"bottom"`, `"left"` or `"right"`. The orientation can be used to further specialize the axis type (e.g., a y axis oriented for the right edge of the chart).
+   * The orientation of the axis. One of `"top"`, `"bottom"`, `"left"` or `"right"`. The orientation can be used to further specialize the axis type (e.g., a y-axis oriented for the right edge of the chart).
```
```suggestion
   * The orientation of the axis. One of `"top"`, `"bottom"`, `"left"` or `"right"`. The orientation can be used to further specialize the axis type (e.g., a y-axis oriented towards the right edge of the chart).
```

Issue #4664, opened 2019-03-06
Missing color in legend with redundant size + color encoding 


haldenl:
Not sure if this has already been addressed, I couldn't find anything with a search. Redundantly encoding a transformed field with a both size and color doesn't result in the legend reflecting both encodings. I'm not sure if this is a bug or just an artifact of the VL model.

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v3.json",
  "description": "A scatterplot showing horsepower and miles per gallons for various cars.",
  "data": {
    "url": "data/cars.json"
  },
  "mark": "circle",
  "encoding": {
    "x": {
      "field": "Miles_per_Gallon",
      "type": "quantitative",
      "bin": true
    },
    "size": {
      "field": "Horsepower",
      "type": "quantitative",
      "aggregate": "average"
    },
    "color": {
      "field": "Horsepower",
      "type": "quantitative",
      "aggregate": "average"
    },
    "y": {
      "field": "Origin",
      "type": "nominal"
    }
  }
}
```

![image](https://user-images.githubusercontent.com/11954298/53853973-8c121280-3f7c-11e9-81ad-72b6f251f6a7.png)

[Link to editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoBmOgCtY3AHaQANOCgATZAgBOjQnh4qckAIJgE6Asl2Ex3PJcTcA7oyXUwL3bGSE3WzB0JU0wVEYxHTBCINoxByVYMAAzbl0wCnR9blZk+GzYOjUNSE1rTBxQCAhIVl0xbC0K0gLfeUUVDQBfdVqMXQBrJsh4Rl14KJLa5CV4bk0PaibqmsgADxWNGqgUxmQxTRGAWUidAH1Y3XOAcXQE5WmdqDwAT1iRgEdWEKM8a0YVCeO0g7A8TTwulYyG2YF6sMgsEYAC9kFtnrt9ocRgAJdJ+AKuWzAtZvD6mb6-ZgAoF9Z6QdDUai6YR4NGmdBUXSMtGw+Eg+YOXTo+l7A5HUx43z+QLCukgsnsqCUpR-Glo+VrRnM1lKhlcnmQPmayCvEUgsXY0wAeX0TBUmtqipGSm4ESU9yNO26PRA3SAA)

[...]

PR #4708, opened 2019-03-10
Fix missing color in legend for size + color legends.

haldenl:
There is currently a bug wherein redundantly encoding size + color creates a legend w/ no color.

#4664

image

[editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoBmOgCtY3AHaQANOCgATZAgBOjQnh4qckAIJgE6Asl2Ex3PJcTcA7oyXUwL3bGSE3WzB0JU0wVEYxHTBCINoxByVYMAAzbl0wCnR9blZk+GzYOjUNSE1rTBxQCAhIVl0xbC0K0gLfeUUVDQBfdVqMXQBrJsh4Rl14KJLa5CV4bk0PaibqmsgADxWNGqgUxmQxTRGAWUidAH1Y3XOAcXQE5WmdqDwAT1iRgEdWEKM8a0YVCeO0g7A8TTwulYyG2YF6sMgsEYAC9kFtnrt9ocRgAJdJ+AKuWzAtZvD6mb6-ZgAoF9Z6QdDUai6YR4NGmdBUXSMtGw+Eg+YOXTo+l7A5HUx43z+QLCukgsnsqCUpR-Glo+VrRnM1lKhlcnmQPmayCvEUgsXY0wAeX0TBUmtqipGSm4ESU9yNO26PRA3SAA)

Culprit is the `fill` attribute in `update` created by the `size` legend:

```
{
  ...
  "legends": [
    {
      "fill": "color",
      "gradientLength": {"signal": "clamp(height, 64, 200)"},
      "title": "Average of Horsepower",
      "size": "size",
      "encode": {
        "symbols": {
          "update": {
            "fill": {"value": "black"},
            "shape": {"value": "circle"},
            "fillOpacity": {"value": 1},
            "stroke": {"value": "transparent"}
          }
        }
      }
    }
  ]
}
```

Proposed fix is to remove that `update.fill` property if top-level `fill` is already defined.


src/compile/legend/assemble.ts
```diff
@@ -26,13 +26,18 @@ export function assembleLegends(model: Model): VgLegend[] {
...
+
     if (legend.encode && legend.encode.symbols) {
       const out = legend.encode.symbols.update;
       if (out.fill && out.fill['value'] !== 'transparent' && !out.stroke && !legend.stroke) {
         // For non color channel's legend, we need to override symbol stroke config from Vega config if stroke channel is not used.
         out.stroke = {value: 'transparent'};
       }
+
+      if (legend.fill && legend.size) {
+        // If fill and size are both used, for non color channel's legend, we need remove fill, as it is already specified.
```
domoritz: Shouldn't we always delete this when fill is defined?

haldenl: Wasn't sure if under other conditions you may want an override, but I guess that does not make much sense. Fixed.

examples/compiled/interactive_seattle_weather.vg.json
```diff
@@ -562,14 +562,7 @@
...
       "symbolType": "circle",
-      "encode": {
-        "symbols": {
-          "update": {
-            "fill": {"value": "transparent"},
-            "opacity": {"value": 0.7}
-          }
-        }
-      },
+      "encode": {"symbols": {"update": {"opacity": {"value": 0.7}}}},
```
haldenl: @domoritz Aha point symbol needs the override.

haldenl: or is this correct? The bar chart layer triggers the fill. Should the legend have circles (to match bar chart) or points (to match scatterplot)?

[example](https://vega.github.io/editor/#/examples/vega-lite/interactive_seattle_weather)

domoritz: We have to decide. I don't remember what the default resolution is (first or last legend when merging). @kanitw?

kanitw: What do you mean override here?

src/compile/legend/assemble.ts
```diff
@@ -26,13 +26,18 @@ export function assembleLegends(model: Model): VgLegend[] {
...
-    // For non color channel's legend, we need to override symbol stroke config from Vega config
+
     if (legend.encode && legend.encode.symbols) {
       const out = legend.encode.symbols.update;
       if (out.fill && out.fill['value'] !== 'transparent' && !out.stroke && !legend.stroke) {
         // For non color channel's legend, we need to override symbol stroke config from Vega config if stroke channel is not used.
         out.stroke = {value: 'transparent'};
       }
+
+      if (legend.fill) {
```
kanitw: This solves the problem, but I'm wonder if fill shouldn't even be in the component in the first place. (Should we fix this in the parse stage, not here?)

examples/compiled/interactive_seattle_weather.vg.json
haldenl: The `fill` attribute in `... encode.symbols.update` overriding the `fill` attribute at the top level `legend`.

src/compile/legend/assemble.ts
haldenl: I think the issue is only with multiple channels sharing the same domain, and merging these occurs in this stage.

domoritz:
This should have high priority, I think.

examples/compiled/interactive_seattle_weather.vg.json
haldenl: btw if we decide to remove this override, we probably want to make sure the opacity of the fill matches the opacity of the stroke.

domoritz:
@haldenl What's left here or is this ready?

kanitw:
I’m not convinced if this is the right fix. Please don’t merge yet.

domoritz:
@kanitw @haldenl We should find a solution for this issue. What's the proposed approach? Can we get a simple fix in soon?

kanitw:
Ok I take a closer look.  There is no way to get rid of that "black" entirely given Vega default symbolBaseFill is transparent.   So this PR is fine.

haldenl:
@domoritz @kanitw 

Follow up question: just wanted to make sure we made a decision on the fill of the legend. 

image

The stroke and fill don't quite match here—any thoughts?

domoritz:
Is that coming from opacity on the fill?

haldenl:
yep

domoritz:
We have a default mark opacity that might be applied here. I have to see the full spec.

haldenl:
It's the weather example

[https://vega.github.io/editor/#/examples/vega-lite/interactive_seattle_weather](https://vega.github.io/editor/#/examples/vega-lite/interactive_seattle_weather)

PR #4709, opened 2019-03-10
Update to Vega 5.2

domoritz:


domoritz:
There are weird runtime test failures. 

In `yarn jest test-runtime/resolve.test.ts` the first few tests pass and then fail.

domoritz:
Ahh, the tests fail after the page crashes.

test-runtime/index.html
```diff
@@ -8,7 +8,14 @@
     <div id="vis"></div>
 
     <script type="text/javascript">
+      // global
+      var view = null;
```
domoritz: I just added this to make it explicit that the view is global and so that the `if (view)` below does not crash when the variable is undefined.

Issue #4306, opened 2018-12-03
Add support for clearing selection


kanitw:
1) Cannot deselect marks in tabular heatmaps

As deselection is only triggered on the plot area, there is no area to click to deselect 

```json
{
  "data": {
    "values": [
      {
        "Actual": "Class A",
        "Predicted": "Class A",
        "Count": 13
      },
      {
        "Actual": "Class A",
        "Predicted": "Class B",
        "Count": 0
      },
      {
        "Actual": "Class A",
        "Predicted": "Class C",
        "Count": 0
      },
      {
        "Actual": "Class B",
        "Predicted": "Class A",
        "Count": 0
      },
      {
        "Actual": "Class B",
        "Predicted": "Class B",
        "Count": 10
      },
      {
        "Actual": "Class B",
        "Predicted": "Class C",
        "Count": 6
      },
      {
        "Actual": "Class C",
        "Predicted": "Class A",
        "Count": 0
      },
      {
        "Actual": "Class C",
        "Predicted": "Class B",
        "Count": 0
      },
      {
        "Actual": "Class C",
        "Predicted": "Class C",
        "Count": 9
      }
    ]
  },
  "encoding": {
    "y": {
      "field": "Actual",
      "type": "nominal"
    },
    "x": {
      "field": "Predicted",
      "type": "nominal"
    },
    "color": {
      "field": "Count",
      "type": "quantitative",
      "scale": {
        "zero": true
      }
    }
  },
  "layer": [
    {
      "mark": {
        "type": "rect",
        "cursor": "pointer",
        "filled": true
      },
      "selection": {
        "selected": {
          "type": "multi"
        },
        "highlighted": {
          "type": "single",
          "empty": "none",
          "on": "mouseover"
        }
      },
      "encoding": {
        "fillOpacity": {
          "condition": {
            "test": {
              "and": [
                "length(data(\"selected_store\"))",
                {
                  "not": {
                    "selection": "selected"
                  }
                }
              ]
            },
            "value": 0.3
          }
        }
      }
    },
    {
      "transform": [
        {
          "filter": {
            "or": [
              {
                "selection": "highlighted"
              },
              {
                "and": [
                  "length(data(\"selected_store\"))",
                  {
                    "selection": "selected"
                  }
                ]
              }
            ]
          }
        }
      ],
      "mark": {
        "type": "rect",
        "cursor": "pointer",
        "opacity": 1,
        "filled": true,
        "stroke": "black"
      },
      "encoding": {
[...]

PR #4720, opened 2019-03-13
Add support for clearing selection

allenjlee:
This PR adds support clearing selections in Vega-Lite (#4306) for single, multi, and interval types.

In summary: there is now a `clear` selection transform that will clear all selected points as well as clear all panning and zooming in the visualization. The `clear` `EventStream` is `mouseout` for `on: mouseover`, and `dblclick` for everything else by default. Users can also specify what `EventStream` they want `clear` to be.

Here is a table of the behavior:

| Interaction  | Clear Behavior |
| ------------- | ------------- |
| Single select  | Deselects Everything  |
| Pan + Zoom w/o Data Selection  | Changes back to original scale |
| Shift to Multi-Select | Deselects Everything |
| Query Widgets | Deselects everything including query sliders |
| Hover Line Chart | Deselects point and hides line |
| Brush | Deselects everything and hides brush window |
| Interactive Scatterplot Matrix | Deselects points and hides brush window and changes back to original scale |

Files created:
- clear.ts
- clear.test.ts

Files modified:
- selection.ts
- transforms.ts
- single.test.ts
- toggle.test.ts
- inputs.test.ts
- interval.test.ts
- multi.test.ts

TESTING: 
- clear.test.ts added (single, multi, interval tests - default, input, false).
- yarn test passes
- Checked against all interactive examples with local Vega editor


kanitw:
Nice to see this coming along!

Some suggestions:

1) Once this PR is ready, it's better to squashing and merging this into a feature branch (e.g., `clear`) rather than master, given that we will release 3.0 soon. 
It's better to have all of them ready at ship them together all at once.  

2) Also, I know this is still a WIP, but once you request a review, you may want to make your commits a bit more atomic and make your commit message meaningful.  (We want to avoid looking at git log later and see "testing editor linking".)

(Learn `git commit --amend` and `git rebase -i` if you haven't)

I hope this is helpful.  I'm excited about your contribution. :)

allenjlee:
@kanitw How do I merge into a feature branch?

src/selection.ts
```diff
@@ -57,6 +57,16 @@ export interface BaseSelectionDef {
...
+  /**
+   * Controls clearing selections. Can be `true`, `false`, or a `string` that is
+   * one of the [Supported Event Types](https://vega.github.io/vega/docs/event-streams/).
+   *
+   * __Default value:__ `true`, which corresponds to `dblclick` (i.e.
+   * user doubleclick clears current selection).
+   *
+   * See the [TODO: clear] documentation for more information.
+   */
+  clear?: string | boolean;
```
arvind: Let's set `clear` to be an `EventStream | boolean`. Even though `EventStream` is aliased to `any`, we prefer more specific typings wherever possible both as a self-documenting mechanism and for when we make the types more specific in the future.

src/selection.ts
```diff
@@ -247,13 +278,15 @@ export const defaultConfig: SelectionConfig = {
   multi: {
     on: 'click',
     fields: [SELECTION_ID],
+    clear: 'dblclick',
```
arvind: Not sure why `multi` and `interval` selections have a `clear` set by default but `single` doesn't? Recall that in #4306, the second example motivated the need for clearing selections using a `single` selection. Let's have `clear: 'dblclick'` on all selection types.

src/compile/selection/transforms/clear.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const clear: TransformCompiler = {
+  has: selCmpt => {
+    return selCmpt.clear;
+  },
+
+  signals: (model, selCmpt, signals) => {
+    const tupleIdx = signals.findIndex(i => i.name === selCmpt.name + TUPLE);
+    signals[tupleIdx].on.push({
+      events: [{source: 'scope', type: selCmpt.clear}],
```
arvind: Wrap `selCmpt.clear` with a call to `parseSelector` (like in `translate.ts` or `zoom.ts`). This is important to making sure that, in multi-view visualizations, clearing only fires if the user double-clicks within the same view as the selection was defined in (`scope`).

src/selection.ts
domoritz: Should `EventStream` be `string` or `any`?

src/selection.ts
allenjlee: I had a quick question related to this that I wanted to follow up about. What should the `signals` array contain when I call it from the transform? I had thought it would contain all `signals` in the given `selCmpt`, but when I try to iterate through it appears some are missing. I haven't been able to figure out why this is happening.

Are transforms compiled further upstream before other parts?

src/selection.ts
```diff
@@ -272,6 +269,7 @@ export const defaultConfig: SelectionConfig = {
   single: {
     on: 'click',
     fields: [SELECTION_ID],
+    clear: 'dblclick',
```
kanitw: The default clear should be `mouseout` if `on` is `mouseover`.  

(But `dblclick` is good otherwise. Thus, I don't think `dblclick` should be defined here in config.) .

kanitw: Also, I don't think `"interval"` needs default "clear" given that click would be the default clear without an additional signal anyway. (Click triggers brush, and if you don't drag,  that's equivalent to clearing.)

arvind: > But dblclick is good otherwise. Thus, I don't think dblclick should be defined here in config.

Hmm, this is an interesting point. The default `"on": "click"` is defined in the configuration, so it seems to make sense to define clear consistently as well. But I see your point that if the user changed to `"on": "mouseover"` we might want to detect that and reflect it. If that's the only exception we can think of, I'd rather handle it separately, when we parse defaults, rather than not have `clear` defined in our `SelectionConfig` whatsoever.

> Also, I don't think "interval" needs default "clear" given that click would be the default clear without an additional signal anyway.

I disagree. If we had brush-to-zoom functionality (#4742) a `clear` would be necessary to reset the view to the original scale domains. Moreover, as much as possible, I would like to keep all selection types working consistently (that is, after all, our original vision of a _grammar_ of interaction rather than a typology of selections).

kanitw: > Hmm, this is an interesting point. The default "on": "click" is defined in the configuration, so it seems to make sense to define clear consistently as well. But I see your point that if the user changed to "on": "mouseover" we might want to detect that and reflect it. If that's the only exception we can think of, I'd rather handle it separately, when we parse defaults, rather than not have clear defined in our SelectionConfig whatsoever.

In the rest of the language, we make the config `undefined` if the default value has condition rule.  For example, we don't set `opacity` in mark config. Same for axis `labelAngle`.

When we want users to be able to customize config for each case, we then introduce a more specific version of config.  For example, `bandPaddingInner` scale config is `undefined` (automatic) by default, but the automatic rule will draw value from `barBandPaddingInner` and `rectBandPaddingInner` scale configs.  If users set `bandPaddingInner` scale config to a specific value, then the default value is fixed and `barBandPaddingInner` and `rectBandPaddingInner` are ignored.  

So I think we should have consistent config design here. 


>  When an interval selection is bound to scales, it would be helpful to clear it to return to the original scale domains

Good point.  I agree.  

Note that as I mentioned previously, I see the benefit of interaction typologies as a macro level on top of selection grammar and have started prototyping it a bit, but that's a different issue separate from this. :)

allenjlee: Sorry for the late reply, I just got back to the US.

Is the intention to have clear non-conditionally defined, i.e. set default configs to `dblclick` and in cases where user changed to `"on: "mouseover"` they also need to change `"clear": "mouseout"`. Or have it conditionally defined, making the config `undefined` and conditionally choosing the EventStream trigger based off of the `on` config?

If the latter, should I make these changes in this PR or in a follow-up?

kanitw: The latter.  

Since it's a simple change, let's do in this PR?

kanitw: Also, don't forget to add `clear` to the docs. :)

site/docs/selection/clear.md
```diff
@@ -0,0 +1,32 @@
...
+layout: docs
+menu: docs
+title: Clearing a selection
+permalink: /docs/clear.html
+---
+
+The `clear` selection transformation resets the visualization to its initial configuration:
+
+- For `single` and `multi` selections, it will clear all selected values.
+- For `interval` selections, it will reset the grid position to its initial configuration.
```
kanitw: 1) Does this mean there  is no way to "empty" the interval selection if `init` is set? 

2) If 1 is a desired behavior, interval seems a bit inconsistent with single/multi? What if single and multi has "init", shouldn't it reset to the init too?

site/docs/selection/clear.md
```diff
@@ -0,0 +1,32 @@
...
+
+It can take one of the following values:
+
+- `false` -- disables clear behavior; there will be no trigger that resets the visualization to its initial configuration.
+- A [Vega event stream definition](https://vega.github.io/vega/docs/event-streams/) to indicate which events should trigger clearing of the visualization.
+
+Vega-Lite automatically adds a clear transform to all selections by default. It will check your Vega-Lite configuration and conditionally pick an appropriate `clear` trigger based off of the `on` trigger (e.g. `click` to `dblclick`, `mouseover` to `mouseout`).
```
kanitw: The default behavior should be in the TS Docs (so it shows up in tools like Altair too.) 

> It will check your Vega-Lite configuration and conditionally pick an appropriate `clear` trigger based off of the `on` trigger (e.g. `click` to `dblclick`, `mouseover` to `mouseout`).

This description is a bit unclear what is the default if `on` is not click/mouseover.  (Missing "else" case.)

site/docs/selection/clear.md
kanitw: Given the property name is "clear", I think it should really clear the selection rather than "resetting". 

If resetting is a good behavior to support, we could consider a separate `"reset"` property.  

(For now, I think clear is more critical.)

site/docs/selection/clear.md
```diff
@@ -0,0 +1,32 @@
+---
```
kanitw: Just like `on` that is listed  here: https://vega.github.io/vega-lite/docs/selection.html#selection-properties, I think clear is a basic selection property so should be listed in the same section instead of having its own page.

kanitw: To be fair, I think our  line between what is selection transform and what is not isn't very clear. 

However, given that "clear" should be paired with "on", it's  better to put them in the same place for now.

kanitw:
- [ ] FYI, I've also checked out this branch and try and the following spec doesn't work yet:

```
{
  "selection": {
    "brush": {
      "type": "interval",
      "init": {"x": [55, 160], "y": [13, 37]},
      "clear": "mouseup"
    }
  },
  "mark": "point",
  "encoding": {
    "color": {
      "condition": {"selection": "brush", "field": "Cylinders", "type": "ordinal"},
      "value": "grey"
    }
  }
}
```

allenjlee:
What is the intended behavior? I see you specified `clear: mouseup`, so when you release after clicking it will clear the selection? Or am I misunderstanding the `mouseup` EventStream?

allenjlee:
For interval types, I'm having trouble clearing both the brush window and the panning/zooming of the scales, specifically for the scatterplot matrix. I'm able to do either of these independently, but together there are problems.

Right now, my strategy is to check which selection signal is being bound to scales, then iterating through the projection fields to get the appropriate signals corresponding to the x- and y-scale, and setting them to null. This clears the panning/zooming of the scales. However, when I try to do this in conjunction with clearing the brush window I get errors (screenshot attached below). 

I get these series of errors:
Screenshot 2019-04-03 00 34 55

kanitw:
> @kanitw What is the intended behavior of the snippet you posted? I see you specified clear: mouseup, so when you release after clicking it will clear the selection? Or am I misunderstanding the mouseup EventStream?

Yes, basically only include the brush while user is still dragging. You can see this  behavior in Apple's and [Google's Stocks](https://www.google.com/search?q=aapl&rlz=1C5CHFA_enUS819US819&oq=aapl&aqs=chrome..69i57j0j69i60l2j0l2.740j0j7&sourceid=chrome&ie=UTF-8).

site/docs/selection/clear.md
allenjlee: What are the TS Docs?

kanitw: https://github.com/vega/vega-lite/pull/4720/files#diff-a02c9ec474a7c7ff01691b050f063691R60  -- You already added some, but didn't include default behavior. 


https://github.com/Microsoft/tsdoc

site/docs/selection/clear.md
arvind: Agreed, the line between a selection property and transformation is unclear. Nevertheless, my preference would be to keep clear on a separate page for a couple of reasons:

(1) It would mirror `toggle` that lives on its own separate page.

(2) The main selection documentation page is already pretty long. `clear` has some non-trivial explanation associated with it plus 1-2 examples. We could shoe-horn it under the `Selection Properties` heading, but seems unnecessary given that it is well-scoped to be standalone (whereas `on` is a little more core to a selection's definition). 

(Fwiw, I had planned on moving `init` out to a separate page too, because of how complex its definition can be. But I ran out of time when I was developing that feature.)

We can water down the "transformation" language and just call it a "property" if you prefer. This is what we do with "bind" (but not for any principled reason). In fact, I'm not really sure that we need to explicitly distinguish between property/transformation in the documentation -- I think that distinction is more useful in the research context.

kanitw: Sounds good. :) 

If we say they are all selection properties, I agree that we should have clear on a separate page as the main page is already too long. 

In fact, in a separate  PR later, we should just make a page for each selection "property".  I also wonder if on should be on the same page as "clear", just like the width/height page.   But I think we can handle that later in a separate PR too.

site/docs/selection/clear.md
allenjlee: I updated the TSDocs and Vega-lite docs, let me know if things are still worded poorly.

arvind:
Hi @allenjlee, nice job refactoring the code and making a first pass at the documentation! Things looked good, and I made a few updates to:

* Clean up the code to follow our style guide lines (e.g., we prefer `camelCase` variable names rather than underscore-delimited ones).
* Minimize the amount of signals that were appended. As we'd discussed, for interval selections, they only need to be added on the `vname` signals (e.g., `brush_x` or `brush_y`) and the rest of the logic propagates that throughout to produce the right behavior. Similarly, we only need to append to the tuple for `single` or `multi` selections.
* Your table at the top of this PR correctly identifies the behavior that should occur on query widgets but I wasn't able to reproduce it. This is because the `clear` transform was not firing for `topLevelSignals`. I recall discussing this with you as well. 
* Remember to double check that any new examples you add as part of documentation are included in the PR as well. Also, to reference the examples, you need to specify them as part of the `data-name` HTML property (not `id`). 


Finally, in the future, I recommend you use `git rebase master` to keep your branches in sync with the master branch. This strategy will avoid the merge commits polluting your history, and makes reviewing a little bit easier :) 

Things look good from me here. I'll hand off to @kanitw to merge, in case he has any final comments.

allenjlee:
Thanks for the comments, will remember in the future!

allenjlee:
Is it ok to squash and merge this in now?

kanitw:
Yeah, let's do it. I found some problem when I tried to replicate the stock example, but it might be a different problem somewhere else.  Given this is working well for basic functionality already, let's get this in.

kanitw:
@allenjlee  Congrats on your PR. 🎉 


ps. I'll create the example and file an issue later.

domoritz:
🎉 Congratulations on this PR @allenjlee. This is a great improvement.

PR #4725, opened 2019-03-14
Fix interval scale_triggers to only register dependencies on scales.

arvind:
This should unblock bumping to Vega v5.2 (#4709) which was blocked because selection runtime tests would crash due to an infinite loop in the dataflow graph.

Scale trigger signals were introduced to break infinite loops in the dataflow graph by proxing scale reactions. Unfortunately, by using a straight `update` function, the same dependency links were being
registered (i.e., the infinite loop in the dataflow graph persisted) but went undetected due to dataflow bugs in Vega pre-5.1 (see vega/vega#1678). Once these bugs were fixed, the infinite loop resurfaced (see vega/vega#1689). By using on-event-triggers, this commit should actually break the loop
by making scale_triggers dependent on only the scales, rather than the interval signals as well.

/cc @jheer

### DIFF
--- a/examples/compiled/brush_table.vg.json
+++ b/examples/compiled/brush_table.vg.json
@@ -195,7 +195,13 @@
         },
         {
           "name": "brush_scale_trigger",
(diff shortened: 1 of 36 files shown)


kanitw:
LGTM

Do we have to bump the Vega version as a part of this PR?  If not, feel free to merge.

PR #4730, opened 2019-03-15
Make aggregate use the right group-by field for lat/long

### LINKED ISSUE #4368: Aggregate doesn't use the right group-by field for projections
```
{
  "width": 500,
  "height": 300,
  "layer": [
    {
      "data": {
        "url": "data/us-10m.json",
        "format": {"type": "topojson", "feature": "states"}
      },
      "projection": {"type": "albersUsa"},
      "mark": {"type": "geoshape", "color": "#ccc"}
    },
    {
      "data": {"url": "data/airports.csv"},
      "projection": {"type": "albersUsa"},
      "transform": [
        {
          "lookup": "origin",
          "from": {
            "data": {"url": "data/flights-airport.csv"},
            "key": "origin"
          },
          "as": ["flights"]
        }
      ],
      "mark": "point",
      "encoding": {
        "latitude": {"field": "latitude", "type": "quantitative"},
        "longitude": {"field": "longitude", "type": "quantitative"},
        "size": {"aggregate": "count", "type": "quantitative"}
      }
    }
  ]
}
```

It should `"groupby": ["layer_1_x", "layer_1_y"]`, not `["longitude", "latitude"]`

kanitw:
Fix #4368

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -93,6 +93,30 @@ export type PositionChannel = 'x' | 'y' | 'x2' | 'y2';
 
 export type GeoPositionChannel = 'longitude' | 'latitude' | 'longitude2' | 'latitude2';
 
+export function isGeoPositionChannel(c: Channel): c is GeoPositionChannel {
+  switch (c) {
+    case LATITUDE:
+    case LATITUDE2:
+    case LONGITUDE:
+    case LONGITUDE2:
+      return true;
+  }
+  return false;
+}
+
+export function getPositionChannelFromLatLong(channel: GeoPositionChannel): PositionChannel {
+  switch (channel) {
+    case LATITUDE:
+      return 'y';
+    case LATITUDE2:
+      return 'y2';
+    case LONGITUDE:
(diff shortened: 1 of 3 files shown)


src/channel.ts
```diff
@@ -93,6 +93,30 @@ export type PositionChannel = 'x' | 'y' | 'x2' | 'y2';
 
 export type GeoPositionChannel = 'longitude' | 'latitude' | 'longitude2' | 'latitude2';
 
+export function isGeoPositionChannel(c: Channel): c is GeoPositionChannel {
+  switch (c) {
```
domoritz: `contains(c, [LAT,LAT2,...])`?

kanitw: The current code is faster.

domoritz: I believe you but how do you know?

kanitw: switch is look up, contains is for loop?  It probably doesn't matter. Let's say I won't change because it doesn't matter.   But let me add tests.

PR #4732, opened 2019-03-15
Interactive Legend for Unit Specs

djbarnwal:
A step towards Issue #1657. Continuation of PR #4317 

Legends can be interacted with when the mark encoding channels completely populate the selection fields.  

vega-lite

### DIFF
--- a/src/compile/legend/parse.ts
+++ b/src/compile/legend/parse.ts
@@ -125,6 +125,9 @@ export function parseLegendForChannel(model: UnitModel, channel: NonPositionScal
 }
 
 export function interactiveLegendExists(model: UnitModel) {
+  if (model.parent) {
+    return [];
+  }
   const selections: InteractiveSelections[] = [];
   // Look over all selections
   forEachSelection(model, selCmpt => {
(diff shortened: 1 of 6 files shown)


src/compile/legend/parse.ts
```diff
@@ -111,13 +113,122 @@ export function parseLegendForChannel(model: UnitModel, channel: NonPositionScal
...
+    }, {${field}: datum.value})`;
+    if (maxProjSelection.fields.length > 1) {
+      test = `!${maxProjSelection.name}_${field}_legend || datum.value === ${maxProjSelection.name}_${field}_legend`;
+    }
+    if (part === 'symbols' && channel === OPACITY) {
+      let strokeValue = '#000000';
+      if (updateValue.stroke) {
+        strokeValue = updateValue.stroke.value;
+      }
+      updateValue.stroke = [{test, value: strokeValue}, {value: '#aaaaaa'}];
```
domoritz: How does this work on a black background? I think we need to use a config option for these.

djbarnwal: I was planning to remove these hard-coded values. @domoritz what would be the best possible place in the spec to have config options such as color, opacity values or disabling interactive legends. Should this be incorporated as an optional field for every mark encoding channel?

domoritz: We should use the Vega-Lite config for this. It might make sense to have a legend property as well that allows users to override the value from the config.

domoritz:
It would be great to get this in. @djbarnwal are you still working on this?

djbarnwal:
> It would be great to get this in. @djbarnwal are you still working on this?

@domoritz I am waiting for a review on this.

arvind:
Superseded by #5305.

PR #4750, opened 2019-03-22
Support time format in composite mark axes

### LINKED ISSUE #4164: Boxplot: temporal -> ordinal axis labels do not format correctly

chanwutk:
Redo of the branch #4190. Fixes #4164.

### DIFF
--- a/examples/specs/normalized/errorband_2d_horizontal_color_encoding_normalized.vl.json
+++ b/examples/specs/normalized/errorband_2d_horizontal_color_encoding_normalized.vl.json
@@ -41,7 +41,7 @@
               "title": "Year (year)",
               "type": "temporal",
               "field": "year_Year",
-              "axis": {"format": "%Y", "formatType": "time"}
+              "axis": {"format": "%Y"}
             },
             "color": {"value": "black"},
             "tooltip": [
@@ -64,7 +64,7 @@
                 "title": "Year (year)",
                 "type": "temporal",
                 "field": "year_Year",
-                "axis": {"format": "%Y", "formatType": "time"}
+                "axis": {"format": "%Y"}
               }
             ]
           }
@@ -82,7 +82,7 @@
               "title": "Year (year)",
               "type": "temporal",
               "field": "year_Year",
-              "axis": {"format": "%Y", "formatType": "time"}
+              "axis": {"format": "%Y"}
             },
             "color": {"value": "black"},
             "tooltip": [
@@ -105,7 +105,7 @@
                 "title": "Year (year)",
                 "type": "temporal",
                 "field": "year_Year",
-                "axis": {"format": "%Y", "formatType": "time"}
+                "axis": {"format": "%Y"}
               }
             ]
           }
@@ -123,7 +123,7 @@
               "title": "Year (year)",
               "type": "temporal",
               "field": "year_Year",
-              "axis": {"format": "%Y", "formatType": "time"}
+              "axis": {"format": "%Y"}
             },
             "color": {"value": "black"},
             "tooltip": [
@@ -146,7 +146,7 @@
                 "title": "Year (year)",
                 "type": "temporal",
                 "field": "year_Year",
-                "axis": {"format": "%Y", "formatType": "time"}
+                "axis": {"format": "%Y"}
               }
             ]
           }
--- a/examples/specs/normalized/errorband_2d_horizontal_color_encoding_normalized.vl.json
+++ b/examples/specs/normalized/errorband_2d_horizontal_color_encoding_normalized.vl.json
@@ -41,7 +41,7 @@
               "title": "Year (year)",
               "type": "temporal",
               "field": "year_Year",
-              "axis": {"format": "%Y"}
+              "axis": {"format": "%Y", "formatType": "time"}
             },
             "color": {"value": "black"},
             "tooltip": [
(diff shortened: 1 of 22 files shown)


src/compile/axis/encode.ts
```diff
@@ -15,8 +15,8 @@ export function labels(model: UnitModel, channel: PositionScaleChannel, specifie
 
   let labelsSpec: any = {};
 
-  // Text
-  if (isTimeFieldDef(fieldDef)) {
+  // We use a label encoding instead of setting the `format` property because Vega does not let us determine how the format should be interpreted.
```
domoritz: I thought this is not true anymore? Please clarify.

src/compile/axis/encode.ts
```diff
@@ -15,8 +15,8 @@ export function labels(model: UnitModel, channel: PositionScaleChannel, specifie
...
+  if (isTimeFormatFieldDef(fieldDef)) {
     const isUTCScale = model.getScaleComponent(channel).get('type') === ScaleType.UTC;
 
     const expr = timeFormatExpression(
```
domoritz: My hope was that we could completely delete this function actually. But I guess it's an unrelated change so let's do this in a follow up.

examples/specs/normalized/errorband_2d_horizontal_color_encoding_normalized.vl.json
```diff
@@ -41,7 +41,7 @@
               "title": "Year (year)",
               "type": "temporal",
               "field": "year_Year",
-              "axis": {"format": "%Y"}
+              "axis": {"format": "%Y", "formatType": "time"}
```
domoritz: It doesn't seem necessary to set the timeFormat when the type is already `temporal`. Can you change this?

src/compile/axis/encode.ts
domoritz: Never mind.

src/compile/axis/encode.ts
domoritz: Ignore this.

src/compile/common.ts
```diff
@@ -101,20 +92,27 @@ export function formatSignalRef(
...
+      const endField = vgField(fieldDef, {expr, binSuffix: 'end'});
+      return {
+        signal: binFormatExpression(startField, endField, format, config)
+      };
+    } else if (fieldDef.type === 'quantitative') {
+      return {
+        signal: `${formatExpr(vgField(fieldDef, {expr, binSuffix: 'range'}), format)}`
+      };
+    } else {
+      return {signal: `''+${vgField(fieldDef, {expr})}`};
```
domoritz: Why the `''+` here?

chanwutk: To convert to string first?

domoritz: But why?

chanwutk: I just tested `return {signal: `${vgField(fieldDef, {expr})}`};`, and use number field, and it works fine. So, I'm not sure why do we need `''+`. Do you want me to remove it?

domoritz: Please think through it but remove it if it's not needed.

chanwutk: Screen Shot 2019-03-24 at 21 25 14

It will convert `null` to string. The left one is with `''+` and the right one is without. So, I think we should keep it.

domoritz: Yep 👍. Thanks for checking.

domoritz:
Ping @kanitw

src/axis.ts
```diff
@@ -303,7 +303,8 @@ const COMMON_AXIS_PROPERTIES_INDEX: Flag<keyof (VgAxis | Axis)> = {
 
 const AXIS_PROPERTIES_INDEX: Flag<keyof Axis> = {
   ...COMMON_AXIS_PROPERTIES_INDEX,
-  encoding: 1
+  encoding: 1,
+  formatType: 1
```
kanitw: Doesn't Vega now have `formatType`?

domoritz: Yes, it does. This should not be here. Good catch!

PR #4760, opened 2019-03-23
Add wheat and wages example to example gallery.

domoritz:
It's a complex example that showcases how customizable Vega-Lite is.


examples/specs/wheat_wages.vl.json
```diff
@@ -0,0 +1,167 @@
...
+      "encoding": {
+        "x": {
+          "field": "year",
+          "type": "quantitative"
+        },
+        "y": {
+          "field": "wages",
+          "type": "quantitative"
+        },
+        "color": { "value": "#EE8182" }
```
kanitw: Make the spec more concise by replacing value defs with mark properties?

domoritz: Done

Issue #4766, opened 2019-03-24
Support columnar initialization for multi selections


mattijn:
See:
<img width="1046" alt="Screenshot 2019-03-24 at 20 47 22" src="https://user-images.githubusercontent.com/5186265/54884891-6fc80e00-4e76-11e9-87a0-e381c1f322e9.png">

Validation shows that `"init": {"x": ["E", "I"]}` can be one of many options, but `array` is not one of them, which should be included as well I guess.

[...]

PR #4771, opened 2019-03-25
Fix bug with inferring projection from selection initialization

arvind:
Fixes the bug identified in https://github.com/vega/vega-lite/issues/4766#issuecomment-476023981.


domoritz:
But it does not fix https://github.com/vega/vega-lite/issues/4766 itself, right?

arvind:
Not the overall question of types for multi selections, no not yet. This bug exists regardless.

domoritz:
Master is broken until #4772 gets merged.

PR #4772, opened 2019-03-25
Rebuild example to fix master

domoritz:
cc @arvind


domoritz:
Hmm, I don't see why the PR build is failing.

PR #4774, opened 2019-03-25
Require Vega ^5.3.2 as peer since 5.3.2 fixes major config bug that m…

kanitw:
…akes Vega-Lite config broken


domoritz:
Please update `vega-cli` itself as well.

kanitw:
> Please update vega-cli itself as well.

Done, but I highly wonder why we even need to update  both of them.

Issue #4786, opened 2019-03-27
Hover selection can't sync across view


kanitw:
[Applying single selection on hover doesn't sync across cell](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoBmOgCtY3AHaQANOCgATZAgBOjQnh4qckACK701MN1Z4w6MLuTw86JdVZj0usO12ssIhgeNxgiIzUiBJR9oTcjEp4sHRqGpCa6G7YYMCQrLpiORlZ6KTwPikKypAAvuoQkIjwyhV4OQDaoBCNsMhiLkY1ON09jYjcVLo5o2ONeACehMjFsInUA2lzPZDDUKi2fZPI0w3bUGiGC8VKyisaY7UPYPXPkBi6ANbF8YntZ41kEoWpp1jNno0AB4zSAAM0Y-U0xQAEtxdH14gB3E5qKCLZbFACOrHcRjcRiodQBO2uIzhCLESNMAFlGANYAB9Za6DkAcXQYjENVUeKWK1MxNJzCyjEpr22kBaQumIwhOxaSlBQxMuUgfQGrmMxQmU1x9MRxQAwgsJJqTrAzfjxVA0aClAKqWrGhQBaxnZBqM5rmqno8QBp6rMoPrBkbVQqTTj4+dIE7VutNtSxrsde9DshjqcvRdUFcbndICHnvKdh9vqZfkktjsgSCwcns9C6fCLaZUejkFicSLU2KiSSktKKSsa9nabqe4zigBBeDwfonGXC0UEiUTskyuVZqBKtHg84n5RauO6mOGvZNQtmxdMqDW23adGOsemV2JD2znMkA+mIfrFIGyDBtsoY9E8EYALogE8QA):

![image](https://user-images.githubusercontent.com/111269/55116303-0a596480-50a4-11e9-9847-17c836174164.png)

The root cause is that the generated identifier transform the produces the underlying `_vgsid_` key does not get merged. 

[...]

PR #4787, opened 2019-03-27
Set hash for identifier transform node to allow merging

kanitw:
Fix https://github.com/vega/vega-lite/issues/4786


src/compile/data/identifier.ts
```diff
@@ -15,6 +15,10 @@ export class IdentifierNode extends DataFlowNode {
     return new Set([SELECTION_ID]);
   }
 
+  public hash() {
+    return 'Identifier';
```
domoritz: No `SELECTION_ID`?

kanitw: `export const SELECTION_ID = '_vgsid_';` --> why would a constant matter?

domoritz: Right.

kanitw: Merge?

PR #4794, opened 2019-03-29
Support inline argmin/argmax

### LINKED ISSUE #3520: A way to use argmin/argmax inside encoding
Argmin/argmax won't work in `encoding` as it involves two fields.  We need better syntax for this.

kanitw:
Let's agree on default title format first 

Fix #3520


examples/compiled/bar_argmax.vg.json
```diff
@@ -0,0 +1,110 @@
...
+      "paddingInner": 0.1,
+      "paddingOuter": 0.05
+    }
+  ],
+  "axes": [
+    {
+      "scale": "x",
+      "orient": "bottom",
+      "grid": false,
+      "title": "Production_Budget for max US_Gross",
```
kanitw: Production_Budget for max US_Gross -- is the title good enough?

domoritz: It's not quite correct (because we are looking at the record, not the value) but it's the shortest and most reasonable title. So yes, it's good.

test/fielddef.test.ts
```diff
@@ -22,6 +22,14 @@ describe('fieldDef', () => {
     it('should access flattened fields in expression', () => {
       expect(vgField({field: 'foo.bar\\.baz'}, {expr: 'datum'})).toEqual('datum["foo.bar.baz"]');
     });
+
+    it('should access argmin field in expression', () => {
+      expect(vgField({aggregate: {argmin: 'b'}, field: 'a'})).toEqual('argmin_b.a');
```
domoritz: toBe

kanitw: why toBe?

domoritz: Because it's just a string. `toBe` is a stricter comparison (like `===` instead of `==`) and stricter is better for tests.

kanitw: We better fix that in a separate PR.   (See all tests around this file.)

PR #4803, opened 2019-03-30
[BugFix] Filter null data by default

### LINKED ISSUE #4798: VL3 breaks null filtering for scale domains
This chart creates a y-axis that is much too large:

```json
{"mark":{"type":"circle"},"data":{"url":"data/movies.json"},"encoding":{"x":{"field":"Rotten_Tomatoes_Rating","type":"quantitative","bin":{"maxbins":20}},"y":{"type":"quantitative","aggregate":"count"}}}
```

If I add an explicit null filter, I get the correct scale domain:

```json
{"transform":[{"filter":"datum.Rotten_Tomatoes_Rating!==null"}],"mark":{"type":"circle"},"data":{"url":"data/movies.json"},"encoding":{"x":{"field":"Rotten_Tomatoes_Rating","type":"quantitative","bin":{"maxbins":20}},"y":{"type":"quantitative","aggregate":"count"}}}
```

This appears to be a regression that violates the expected behavior from VL2. This breaks our curriculum examples and I'd like to request that it be an immediate priority for a fix and patch release.

kanitw:
Fix #4798


domoritz:
This is a breaking change, no?

kanitw:
I consider this a bug fix

jheer:
If it is un-breaking an unintended / undocumented breaking change I think that classifies as a fix?

domoritz:
`null` is now interpreted as `0` in some examples

Screen Shot 2019-03-30 at 11 17 58

kanitw:
> null is now interpreted as 0 in some examples



Sounds like we still have a bug. (But that's why I make this "draft" as I wanna see example diff first.)

jheer:
Looking at the generated Vega specs for the test cases, it looks like the mark logic for hiding invalid values was removed, but the invalid value filters were not added. I'm not familiar with the VL internals, but based on the first commit on this PR, perhaps this is just a remaining configuration bug (i.e., the filter generation is not getting invoked?).

domoritz:
I looked through a third of the examples and they look good. 

I strongly think that this change warrants a minor version bump at least. It's not just a bugfix (i.e. minor version bump) since we had the change in Vega-Lite for many months.

PR #4804, opened 2019-03-30
Add search to docs overview.

domoritz:
CC @leibatt

### DIFF
--- a/site/docs/overview.md
+++ b/site/docs/overview.md
@@ -9,6 +9,14 @@ permalink: /docs/
 
 This documentation describes the [JSON specification language](#spec) and how to [use Vega-Lite visualizations]({{site.baseurl}}/usage/embed.html) in a web application.
 
+<form action="https://google.com/search" method="get" class="search">
+  <input type="hidden" name="q" value="site:https://vega.github.io/vega-lite/docs/">
+  <input class="search" type="text" name="q" results="0" placeholder="Search Docs">
+  <button type="submit">
+    Search
+  </button>
+</form>
+
 {:#toc}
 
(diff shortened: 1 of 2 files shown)


kanitw:
Screenshot?

domoritz:
Forgot to make one and now I'm on mobile. It's a plain search input and a button next to it.

kanitw:
This totally looks out  of nowhere. 

image

I agree that it's useful, but shouldn't the search box be on the menu bar instead? 

We can make room for it  by renaming Usage to "Using Vega-Lite" and move ecosystem to become another tab(s) inside "Using Vega-Lite".  The "Ecosystem" tab  is weird any way.

domoritz:
> We can make room for it by renaming Usage to "Using Vega-Lite" and move ecosystem to become another tab(s) inside "Using Vega-Lite".

I'm worried that it makes the applications much less visible. I agree that it would be good to have the search more prominent but it's also not the best search since it just calls google (it's not well integrated into our page and it also only searches the docs right now).

I'd say this is better than nothing and we can iterate from here in follow up PRs.

g3o2:
hmm, once you consider Search for documentation, I wonder whether you rather want a dedicated vl schema property search? i.e. search through keywords instead of a generic Google search ...

domoritz:
> hmm, once you consider Search for documentation, I wonder whether you rather want a dedicated vl schema property search? i.e. search through keywords instead of a generic Google search ...

Yes, I think better search would be nice. I think we get some of that for free with https://github.com/vega/vega-lite/issues/4872. 

I'm merging this PR for now and we can iterate on it.

PR #4808, opened 2019-03-31
Simplify tsconfig.json

kanitw:
- I don't know why we have `"forceConsistentCasingInFileNames": false`
- The  other two are also false by default.
- Reorder to make it easier to compare


domoritz:
Looks good to me.

PR #4811, opened 2019-03-31
Improve Docs

kanitw:
Note that the reference if currently quite nightmarish, but I think fixing https://github.com/vega/ts-json-schema-generator/issues/84 should alleviate this problem

(Still need to check the site if this really fixes the problem, but I can't run `yarn site`


domoritz:
> but I think fixing vega/ts-json-schema-generator#84 should alleviate this problem

I'm working on it.

kanitw:
Since everyone is busy, I'll just merge this for now. Please feel free to send follow up edits.

PR #4816, opened 2019-03-31
Do not set interpolation for nominal color scales

### LINKED ISSUE #4775: Do not set interpolation for nominal color scales
For example in https://vega.github.io/editor/#/examples/vega-lite/stacked_bar_weather, we should not set the `interpolate` property on the color scale.

kanitw:
Fix #4775


src/compile/scale/properties.ts
```diff
@@ -192,8 +192,8 @@ export function bins(model: Model, fieldDef: TypedFieldDef<string>, channel: Cha
   return undefined;
 }
 
-export function interpolate(channel: Channel) {
-  if (contains([COLOR, FILL, STROKE], channel)) {
+export function interpolate(channel: Channel, scaleType: ScaleType) {
+  if (contains([COLOR, FILL, STROKE], channel) && scaleType !== 'ordinal') {
```
domoritz: We might want interpolation for Vega-Lite `ordinal` but not `nominal` encodings. Let's see what changes.

kanitw: oh right

kanitw: Changed.

domoritz:
Thanks. This is great.

PR #4820, opened 2019-03-31
[Docs] Fix binned description

kanitw:


kanitw:
lol this makes me realize how redundant our schema is.

PR #4825, opened 2019-03-31
Rename FieldDef.ts => ChannelDef.ts

kanitw:
Why?:  It includes ValueDef, which is not a field def.  So this is a clearer. name.


kristw:
@kanitw You broke my repo build. lol. But I also thought about renaming the file the same way so can forgive u for this.

jheer:
☝️ brotherly love!

kanitw:
Yeah, note that we use semantic versioning for our JSON syntax, but not for internal files.  

(We will bump the major version if the JSON syntax has breaking change, but we won't bump the major version if we rename our source code files.) 

Thus, if you import our code files directly, better not use `^` in the dependencies.

PR #4837, opened 2019-04-02
Add bin output names to keys used for merging to prevent erroneous merge

### LINKED ISSUE #4810: Bin transforms are merged even though they have different as
```
{
  "concat": [{
    "mark": "rect",
    "encoding": {
      "x": {"bin": true, "field": "Horsepower","type": "quantitative"},
      "y": {"bin": true, "field": "Miles_per_Gallon","type": "quantitative"}
    }
  }, {
    "transform": [{
      "bin": true, "field": "Miles_per_Gallon", "as": "bin_mpg"
    }],
    "mark": "rect",
    "encoding": {
      "x": {"bin": true, "field": "Horsepower","type": "quantitative"},
      "y": {"field": "bin_mpg","type": "quantitative"}
    }
  }]
}
```

image

The first plot shouldn't be empty, but it's empty because the transforms got merged.

Saba9:
Fixes #4810

### DIFF
--- a/examples/compiled/histogram_bin_transform.vg.json
+++ b/examples/compiled/histogram_bin_transform.vg.json
@@ -14,15 +14,17 @@
         {
           "type": "extent",
           "field": "IMDB_Rating",
-          "signal": "bin_maxbins_10_IMDB_Rating_extent"
+          "signal": "bin_maxbins_10_IMDB_Rating_bin_IMDB_Rating_bin_IMDB_Rating_end_extent"
         },
         {
           "type": "bin",
           "field": "IMDB_Rating",
           "as": ["bin_IMDB_Rating", "bin_IMDB_Rating_end"],
-          "signal": "bin_maxbins_10_IMDB_Rating_bins",
+          "signal": "bin_maxbins_10_IMDB_Rating_bin_IMDB_Rating_bin_IMDB_Rating_end_bins",
           "maxbins": 10,
-          "extent": {"signal": "bin_maxbins_10_IMDB_Rating_extent"}
+          "extent": {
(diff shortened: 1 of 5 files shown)


examples/compiled/histogram_bin_transform.vg.json
```diff
@@ -14,15 +14,17 @@
         {
           "type": "extent",
           "field": "IMDB_Rating",
-          "signal": "bin_maxbins_10_IMDB_Rating_extent"
+          "signal": "bin_maxbins_10_IMDB_Rating_bin_IMDB_Rating_bin_IMDB_Rating_end_extent"
```
kanitw: This signal looks nightmarish, so I'm not convinced that the output name should be in the signal.  

In fact, if there are two similar bin transforms (with different `as`), wouldn't it be better to create only one extent transform and one bin transform to avoid redundant computation. We then instead derive new fields (with Vega formula) according to multiple "as".

kanitw: I imagine refactoring `as` in the bin node to be `[string, string][]` instead of `[string, string]` and allow merging non-unique "as" tuple.

Saba9: That seems like a sensible solution to me. I'll try getting that working.

kanitw: Thanks — sorry I didn’t see Dom’s comment earlier.

kanitw: Thanks — sorry I didn’t see Dom’s comment earlier.

kanitw: Let’s close  this PR and open another one then since it will be a totally different solution

PR #4844, opened 2019-04-04
Allow latitude/longitude channels to accept constant values.

### LINKED ISSUE #4836: Add latitude/longitude value ref.
In some cases one may have only longitude or latitude values, and wish to use a constant value for the other geo-coordinate. An example includes plotting samples along a specific meridian (e.g., for labeling). Right now one must generate *both* lat/lon field values and bind them to channels, even though one of those fields is constant.

jheer:
Changes:
- Allow latitude/longitude channels to accept a constant value instead of a field.
- Update GeoJSON/GeoPoint Vega schema types to allow `expr` type in addition to string field.
- Add corresponding tests and example.

Close #4836.


src/vega.schema.ts
```diff
@@ -43,6 +43,10 @@ export interface VgData {
   transform?: VgTransform[];
 }
 
+export interface VgExpr {
```
domoritz: Do we have a type like this in Vega already? If so, we should use it.

domoritz: @willium see https://github.com/vega/vega-lite/pull/4847

domoritz:
@willium Next time you merge, please always squash PRs.

PR #4847, opened 2019-04-05
Use ExprRef from Vega.

domoritz:


jheer:
LGTM, assuming all tests pass!

PR #4850, opened 2019-04-05
Add graticule data generator.

### LINKED ISSUE #3337: Support Graticule transform
https://vega.github.io/vega/docs/transforms/graticule/

jheer:
Changes:
- Update data types to include both data sources and data generators.
- Add generator for graticule GeoJSON data.
- Add tests / examples for graticule generation.

Close #3337.


jheer:
@domoritz When you get the chance, please feel free to check my TypeScript. For example, there is some redundancy between `GraticuleParams` in the `Data` definitions, and the vega-schema `VgGraticuleTransform`. Not sure if there is a preferred way to model these bits.

src/data.ts
```diff
@@ -139,7 +141,41 @@ export function isNamedData(data: Partial<Data> | Partial<VgData>): data is Name
...
+}
+
+export interface GraticuleGenerator extends GeneratorBase {
+  /**
+   * Generate graticule GeoJSON data for geographic reference lines.
+   */
+  graticule: true | GraticuleParams;
+}
+
+export interface GraticuleParams {
```
kanitw: If we want the TSDocs for each of the property to be exactly the same as ones in `vega-typing`'s `VgGraticuleTransform` (which has no docs right now), we can do. 

```
export type GraticuleParams = Omit<VgGraticuleTransform, 'type'>;
```

(We can also redefine docs for jus some properties too.) 

```
export type GraticuleParams = Omit<VgGraticuleTransform, 'type'> & {
  /** different docs text*/
  extent?: number[][];
};
```

jheer: I tried this, and while it worked within VL proper, ts-json-schema-generator crashed as a result:

`UnknownNodeError: Unknown node "Exclude<keyof T, K>" (ts.SyntaxKind = 175) at /work/vega/vega-lite/src/util.ts(326,49)`

domoritz: https://github.com/vega/ts-json-schema-generator/issues/93

jheer:
@domoritz or @kanitw, can I get a review on this PR? Thanks!

kanitw:
Yes, but FYI I'll traveling a lot this and next week though. 
(I mostly get time to hack on the plane, but I can't do code review without access to Github.)

jheer:
Totally understand, but I was planning to write curriculum materials that would build on these and related geo updates _within the next week_. So that would require both a review as well as a 3.2 release. I'm hoping this is possible.

domoritz:
If I have internet on the bus to Vancouver, I will do the review today.

kanitw:
If we want a release, we should have a documentation, so we can point the release note to the documentation?

jheer:
Yes, I plan to add docs very soon, along with a few additional generators (sphere and sequence).

PR #4852, opened 2019-04-07
Vega 5.3.5 and other updates

domoritz:


domoritz:
For some reason, the runtime tests are passing locally but break on Travis. How is that possible since we both use chromium. 

Screen Shot 2019-04-07 at 16 21 57

domoritz:
@jheer @kanitw @arvind any idea why this could happen?

jheer:
v5.3.5 includes three changes. Two of them should have no effect here (one is just test case updates, the other involves GeoJSON issues that are not within the gamut of current VL tests). The one change that could modify output here is "Fix mark bounds for removed label overlaps. Add tests. (#1747)". This change updates mark bounds calculation so that axis labels removed by the overlap check are no longer included in the total bounds for a set of sibling marks.

That said, I have no idea why you would see different outputs between local and Travis... unless there was a discrepancy in terms of the actual Vega versions being used? I suppose you could temporarily modify your .travis.yml file to use MacOS instead and see if you can rule out a platform difference (hopefully so, as I have no insight into why one would see a platform difference here).

domoritz:
Hmm, looks like I had something wrong in my local setup. It works now.

PR #4853, opened 2019-04-08
[Fix Schema] 

kanitw:
- Remove non-null `bin` from `SecondaryFieldDef` / `LatLongFieldDef`
- Remove "binned" from `FacetFieldDef`
- Consolidate `ColorField/ValueDef` as `StringField/ValueDef`
- Consolidate `TextValueDef` as `ValueDef`
- `FieldDef` in the schema => `FieldDefWithoutScale`

### DIFF
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -763,7 +763,7 @@ export function normalizeFieldDef(fieldDef: FieldDef<string>, channel: Channel) 
     fieldDef = {
       ...fieldDef,
       bin: normalizeBin(bin, channel)
-    };
+    } as FieldDef<string>;
   }
 
   if (isBinned(bin) && !contains(POSITION_SCALE_CHANNELS, channel)) {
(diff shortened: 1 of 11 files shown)


kanitw:
Hmm, weird that this is not changing the schema.

domoritz:
Odd indeed. Does it change the schema locally for you? If not, could this be a bug in the schema generator?

kanitw:
There is definitely a bug, but here I'm introducing a workaround and simplify the declarations a bit.

src/channeldef.ts
```diff
@@ -275,9 +251,14 @@ export interface TypeMixins<T extends Type> {
 /**
  *  Definition object for a data field, its type and transformation of an encoding channel.
  */
-export type TypedFieldDef<F extends Field, T extends Type = Type> = FieldDefBase<F> & TitleMixins & TypeMixins<T>;
-
-export interface SortableFieldDef<F extends Field, T extends Type = StandardType> extends TypedFieldDef<F, T> {
+export type TypedFieldDef<
+  F extends Field,
+  T extends Type = Type,
+  B extends Bin = boolean | BinParams | 'binned' | null
```
domoritz: Should this use `Bin` from `src/bin.ts`?

PR #4854, opened 2019-04-08
[Fix Schema] Make `href` channel use `TextFieldDefWithCondition` 

kanitw:
instead of `StringFieldDefWithCondition`

(StringFieldDefWithCondition includes scale, legend -- which do not make sense  for `href`.


kanitw:
It's unbelievable that we can make the schema 360 lines shorter with just a change in  one line ><"

PR #4873, opened 2019-04-12
Use gray for unfiltered, blue for filtered instead of blue/gold for crossfilter (similar to Falcon)

domoritz:


jheer:
Much better! FYI I made the same change in my interaction notebook, while also changing the example to use wider charts and `vconcat`.

arvind:
LGTM. Thanks! Do we also want to switch to `vconcat` as @jheer suggested? I'm fine with it either way, but could see the vertical layout playing nicer within the editor, for instance.

domoritz:
I prefer the horizontal layout for now as it works better on the example page on the website (https://vega.github.io/vega-lite/examples/interactive_layered_crossfilter.html).

PR #4878, opened 2019-04-15
Fix documentation comment for sequence parameters.

jheer:
Previously the `SequenceParams` type mistakenly had the documentation comment for `GraticuleParams`, which propagates out to the JSON schema. This PR adds a proper description for the sequence generator.


domoritz:
LGTM

PR #4880, opened 2019-04-15
Add default translate for scaled projection.

jheer:
Changes:
- If a custom projection is specified with `scale` but not `translate`, emit a default `translate` property in the output Vega spec, set to one-half the `width` and `height` signals.
- Update unit tests.


jheer:
FWIW, I would tend to consider this a bug fix (patch release) as this is the envisioned albeit heretofore unimplemented default behavior I had in mind for the new custom projection facilities.

PR #4887, opened 2019-04-17
Include continuousPadding in scale config

WillMLumere:
The `continuousPadding` scale config option referenced in the [definition of the `padding` option for continuous scales](https://vega.github.io/vega-lite/docs/scale.html#continuous) is not listed in the table of scale config padding options. I'm not sure this change will resolve the issue as I'm not familiar enough with this repo to trace down the `ScaleConfig` referenced in the include template directive.


domoritz:
We automatically pull properties from the schema so if this property exists, this is enough. Thanks for the pr!

domoritz:
Thanks!

Issue #4859, opened 2019-04-10
Selection nearest invalidates tooltip


jheer:
When using a selection with `nearest: true`, the `tooltip` encoding becomes incorrect, showing `undefined` instead of the tooltip value. [Here is a spec demonstrating the issue.](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEC2CGBOBrSAuMoIUgFwJ4AcBTVKAYwEt5SAbY8MAXwBp7IATWbWEjTSAV3jUS7TrAD00APYA3coQDOAOgBWCqQDtI9ZqwWFapbOU096WABazC8M5j54iIheQ0BzWpBb2shaPjwRDU1ibx9IUzQYKX59a1sw+0gNQgRFbBJseH5Cc0YdMMhCDVIpNlc3Oz4ADyqkgDN5ajYRACUpbGxigH0AFSk4bClFbtbOCq88rEdiKIBHflgNYy5jGTp7XSTcOr5GgxaogEkAWQARACFR8fdJnygZkQWllfH17U3Eh6kpamN8XZYfbNES9cjYTxfaYEWZQYLQVywYR5LZ8UrUKS2NC8JIyJE5ZxdAwAI2oBKhZE05WMkXQUz4+kMNK0USs6wS9KweLJsMgjXghGJ8HIpGQ9IYKIKrAA7uQ2NgLCQACwABhVhQshHIbgsGTQAGY1SAGEA)

It appears that the error is due to the Voronoi cells intercepting the mouse event but attempting to produce a tooltip for the wrong data field. As the Voronoi cells use reactive geometry, to reference an underlying data field one should use `datum.datum.field`, not the normal `datum.field`.

jheer:
I just ran into this bug again in another context, so I'm upping the priority and assigning.

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/4888

PR #4888, opened 2019-04-17
Fix tooltip propagation for voronoi marks.

arvind:
#4726 introduced a `reactiveGeom` flag to lookup tooltip values on `datum.datum` rather than `datum` for voronoi marks which use reactive geometry. This PR fixes one location that was missed in the original PR that caused the bug to recur in #4859.


domoritz:
Does this fix #4859?

arvind:
It should, yes!

PR #4890, opened 2019-04-18
Don't group box plots by tooltip + Apply tooltip fields to box or outliers based on whether they are aggregated

### LINKED ISSUE #4882: tooltip breaks boxplot mark
Tooltip channel will brake example https://vega.github.io/vega-lite/examples/boxplot_minmax_2D_vertical.html .
I guess it's similar to #3220

In fact, my intention was to investigate outliners  
 
visualization - 2019-04-16T235726 767

```json
{
  "description": "A vertical 2D box plot showing median, min, and max in the US population distribution of age groups in 2000.",
  "data": {"url": "data/population.json"},
  "mark": {"type": "boxplot", "extent": "min-max"},
  "encoding": {
    "x": {"field": "age", "type": "ordinal"},
    "y": {
      "field": "people",
      "type": "quantitative",
      "axis": {"title": "population"}
    },
    "tooltip": {"field": "year", "type": "quantitative"}
  }
}
```
[Vega-Editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoBmOgCtY3AHaQANOCgATZAgBOjQnh4qckAIJgquo-HRiwAJgAiYdtwAeYQmO54wsRG4Ad0YlajBUZE1GdCVVCND42M0I9E9QsBZkMABVAGUvbkJWMXQjZTBo2Dx9dlZypTBuADMwdGps6l1uVmIwDIcABmG6NQ1ITTLMHGBIVl0xbC0p0kIikrLjeUUVAF91CEgMXQBrJdm8AE9CZCXIN3dvXzUoZHc8ZCU8O9RQkQx3JB9uNPvBuNEwucNIdATNIM1GMgxJo7u1bvFIFcbnduLoIXYgQcIFBLlDicT4YjkXcbkUxOjoRSsbdTABHVixIx4TZUMbkw5pRiwc6Y5j0mnrUoNIGM4HkzHcbhiIyEEUIpEo0yXZDoXQvTHXFlQdmc5g8267DSW3ZAA)

chanwutk:
Fixes #4882


domoritz:
Should some examples change?

chanwutk:
I think we do not have any `boxplot` example with custom `tooltip`.
Should we add one?

domoritz:
Can you make this one work as well?

```json
{
  "data": {"url": "data/population.json"},
  "mark": {
    "type": "boxplot",
    "extent": 1.5
  },
  "encoding": {
    "x": {"field": "age","type": "ordinal"},
    "y": {
      "field": "people",
      "type": "quantitative",
      "axis": {"title": "population"}
    },
    "size": {"value": 5},
    "tooltip": {
      "field": "people",
      "type": "quantitative",
      "aggregate": "max"
    }
  }
}
```

domoritz:
this one works now:

```json
{
  "data": {"url": "data/population.json"},
  "mark": {
    "type": "boxplot",
    "extent": 1.5
  },
  "encoding": {
    "x": {"field": "age","type": "ordinal"},
    "y": {
      "field": "people",
      "type": "quantitative",
      "axis": {"title": "population"}
    },
    "size": {"value": 5},
    "tooltip": {
      "field": "people",
      "type": "quantitative"
    }
  }
}
```

kanitw:
From the title

> Add tooltip to only outliers for Boxplot 

This sounds pretty confusing.  Does this mean that tooltip no longer works for the main part of the boxplot?  If so, why should that be the case?  

Do you mean "Apply tooltip encoding to only the  outliers  points, but retain the default tooltip for the main box part?  (This sounds more reasonable, but still I'm curious to hear why the outliers should have stronger precedence.)

domoritz:
Maybe "Don't group boxplots by tooltip" is a better title here. See #4882 for a description of the issue.

chanwutk:
Custom tooltips can be encoded in either outliers or the main box depending on whether they are an aggregating channel. If a tooltip has aggregation, it will be encoded into the main box and whiskers, but if there is no aggregation, it will be encoded into outliers. For example:
```json
{
  "data": {"url": "data/population.json"},
  "mark": {
    "type": "boxplot",
    "extent": 1.5
  },
  "encoding": {
    "x": {"field": "age","type": "ordinal"},
    "y": {
      "field": "people",
      "type": "quantitative",
      "axis": {"title": "population"}
    },
    "size": {"value": 5},
    "tooltip": [
      {
        "field": "people",
        "type": "quantitative",
        "aggregate": "mean"
      },
      {
        "field": "people",
        "type": "quantitative"
      }
    ]
  }
}
```

The first tooltip with `aggregate: 'mean'` will go into the main box and whiskers, and the second tooltip will go into outliers

kanitw:
I see.  Please make sure to document this with an example that has unaggregated field in tooltip.  (Maybe another with aggregated field in tooltip, but I don't see why users would wanna do this in practice, so maybe a short note after the main example is probably fine). 

In any case, we shouldn't try to include both aggregated and unaggregated fields in the same example because it's too complicated.

domoritz:
> In any case, we shouldn't try to include both aggregated and unaggregated fields in the same example because it's too complicated.

Agreed. Let's not make an example with that.

PR #4901, opened 2019-04-20
Use ESlint instead of TSlint

domoritz:
TSLint is deprecated (https://medium.com/palantir/tslint-in-2019-1a144c2317a9) and we already use ESLint in Vega. This makes things consistent. 

Once this is in, we can also start using recommended rule sets such as the one from airbnb (https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb).


domoritz:
Eslint with typescript has terrible performance so let's not use it until they fix it https://github.com/typescript-eslint/typescript-eslint/issues/389.

src/compile/selection/transforms/project.ts
```diff
@@ -45,14 +44,15 @@ const project: TransformCompiler = {
     const parsed: Dict<SelectionProjection> = {};
     const timeUnits: Dict<TimeUnitComponent> = {};
 
-    const signals = {};
+    const signals = new Set<string>();
```
domoritz: I changed the code here so we don't do an assignment in the return.

src/compile/unit.ts
```diff
@@ -249,27 +249,6 @@ export class UnitModel extends ModelWithField {
     return this.encoding;
   }
 
-  public toSpec(excludeConfig?: any, excludeData?: any) {
```
domoritz: Removed unused method.

src/log/message.ts
```diff
@@ -0,0 +1,321 @@
+import {AggregateOp} from 'vega';
```
domoritz: This makes the namespace cleaner.

test/compile/scale/type.test.ts
```diff
@@ -88,8 +89,8 @@ describe('compile/scale', () => {
         it('should return band scale for ordinal X,Y when mark is rect, rule, bar', () => {
           [ORDINAL, NOMINAL].forEach(t => {
             [X, Y].forEach(channel => {
-              ['bar', 'rule', 'rect'].forEach(mark => {
-                expect(scaleType({}, channel, {type: t}, 'rect', defaultScaleConfig)).toEqual(ScaleType.BAND);
+              [BAR, RULE, RECT].forEach(mark => {
+                expect(scaleType({}, channel, {type: t}, mark, defaultScaleConfig)).toEqual(ScaleType.BAND);
```
domoritz: This was a bug I found through eslint.

kanitw:
This migration makes sense at some point. The question is whether the TS support in ESLint is good enough yet?

domoritz:
Support seems pretty good. You can see https://github.com/typescript-eslint/typescript-eslint and in particular the list of rules at https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#supported-rules

domoritz:
Not sure why the PR build fails. It says it can't find tslint but we don't need it anymore. @travis is behaving oddly here, I think.

domoritz:
I moved Embed over to Eslint.

kanitw:
Btw, if you have time to mess with this, please revert tsc project between src and test first.

domoritz:
@kanitw I reverted the ts project references. Once it's in, can we merge this (I can merge master into this then)?

domoritz:
Merged master

src/compile/unit.ts
kanitw: Are you sure it is unused across projects?

domoritz: I think compassql might use it but I think then it would make sense to move it there. https://github.com/search?q=org%3Avega+toSpec&type=Code

PR #4902, opened 2019-04-20
Use local data path instead of absolute URL

domoritz:
Replaces https://github.com/vega/vega-lite/pull/4900


domoritz:
@marcprux see how we automatically compile for you. I guess travis needs to have permission to do it.

PR #4908, opened 2019-04-21
Use template strings everywhere

domoritz:


domoritz:
Haha, me neither. That's why I made it a new PR. I'm happy not to do it.

kanitw:
The question is whether this gets auto-fixed (say via prettier). If so, I'm ok with having this option on.  If it's not auto-fixed though, I'll vote against it.

domoritz:
Yep, it gets auto fixed. That's how I made this pr. Took me two minutes. We can do this later as well. I'm not too happy with the results. 

I'd close this unless you like it.

Issue #4910, opened 2019-04-23
Data merging logic merges datasets with same URL but different format parameters


jheer:
Vega-Lite's data merging logic is over-aggressive, leading to major errors.

To replicate:
- Load a TopoJSON file twice, once using the `feature` parameter and once using the `mesh` parameter.
- Examine the compiled Vega spec to discover that the two have been erroneously merged, like so:

```json
{
  "name": "source_0",
  "url": "https://cdn.jsdelivr.net/npm/us-atlas@2/us/10m.json",
  "format": {"type": "topojson", "mesh": "states", "feature": "counties"}
}
```

Expected behavior:
- Different format parameters should result in different dataset definitions.

domoritz:
This is bad. We should be able to resolve the error by fixing the logic in https://github.com/vega/vega-lite/blob/875d772bf7dcd2cf9fc0616c998d32d7013d42df/src/compile/data/parse.ts#L58.

domoritz:
This is bad. As far as I see it, we should be able to use the same source but since the format is different, we have to have a fork in the dataflow. 

@Saba9 do you have a few cycles to look at this?

Saba9:
@domoritz I'll take a look at it. I'm pretty busy this week so I can't guarantee I'll be able to get a fix in until the weekend but I'll definitely give it a shot.

domoritz:
In Vega, is it possible to load data only once but use `feature` and `mesh` in separate data sources? Or does Vega internally avoid loading the same JSON file twice?

jheer:
It is not possible to load data only once but use separate formats in different data sources. Derived data sources do not accept format properties. Vega does not internally avoid double loading. We could add internal caching, but I wonder how much browser caching already helps us.

domoritz:
I suspect that even though we do browser caching, we still get the overhead of parsing the JSON document twice. I don't think it's the most pressing issue but eventually it would be great if Vega supported different format properties for the same source file.

kanitw:
Will be fixed in https://github.com/vega/vega-lite/pull/4914

PR #4912, opened 2019-04-23
Stop merging conflicting data sources

Saba9:
Addresses #4910. Check to see if data sources conflict with each other and don't merge them if they do. A conflict can occur if the sources have properties with different values in their format or if they have mutually exclusive properties e.g. 'mesh' and 'feature'.

### DIFF
--- a/src/compile/data/parse.ts
+++ b/src/compile/data/parse.ts
@@ -26,7 +26,7 @@ import {
   isTimeUnit,
   isWindow
 } from '../../transform';
-import {deepEqual, mergeDeep} from '../../util';
+import {deepEqual, keys, mergeDeep} from '../../util';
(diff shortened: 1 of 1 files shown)


Saba9:
A possible refactor would be to modify `findSource` to take these conflicts into consideration when searching. Might be better depending on `findSource`'s intended behavior.

domoritz:
Yeah, I think we should rather modify `findSource`.

Having said that, the real solution is not to keep the source separate but use separate parses in separate data sources that are depending on the same source dataset. However, I think this is a current limitation in Vega so we should fix this in Vega-Lite for now.

kanitw:
Replaced by #4914. If I'm wrong, feel free to re-open.

PR #4915, opened 2019-04-25
docs: Add timelines chart by @RomZeroX

RomZeroX:


domoritz:
Thanks for the PR. Could you update the colors so that the text is white (easier to read)?

This example is similar to https://vega.github.io/vega-lite/examples/bar_gantt.html

RomZeroX:
Hi,

The color is updated.

You are right, it is really similar to the Gantt bar chart (I have not seen it !!!) but splitted by category and with color coding.



----- Mail original -----
De: "Dominik Moritz" <notifications@github.com>
À: "vega/vega-lite" <vega-lite@noreply.github.com>
Cc: "RomZeroX" <bulgroz07@free.fr>, "Author" <author@noreply.github.com>
Envoyé: Jeudi 25 Avril 2019 23:54:47
Objet: Re: [vega/vega-lite] Update index.md : vega-lite sample addition (timelines) (#4915)



Thanks for the PR. Could you update the colors so that the text is white (easier to read)? 

This example is similar to https://vega.github.io/vega-lite/examples/bar_gantt.html

kanitw:
Instead of adding a similar example, do you want to improve existing the gantt example?

domoritz:
I don't think the existing example is necessarily bad. It's the simplest possible gantt chart. https://vega.github.io/vega-lite/examples/wheat_wages.html is the most complex example we have with timelines. This example is somewhat in between.

domoritz:
Actually, this example is already very similar to https://vega.github.io/vega-lite/examples/bar_gantt.html so we decided not to add a redundant example for now.

PR #4919, opened 2019-04-26
Remove default tooltip + add support for `(config.)mark.tooltip = true`

kanitw:
@arvind @domoritz  and I chatted and agree to do it now.  This will be a minor version bump since we change the default but doesn't remove any existing feature. 

(It's debatable, but we also don't want to bump to 4.0 right now.) 

- [x] Add docs for the new `(config.)mark.tooltip = true`
- [x] Revise examples that we have to disable tooltip

cc: @jheer @jakevdp


kanitw:
I'm wondering if we should make this 4.0 together with the regression/loess that Jeff is adding. (That's a huge enhancement anyway.) 

cc: @arvind @domoritz @jheer

PR #4930, opened 2019-05-02
Change axis label's angle of Waterfall Chart of Monthly Profit and Loss

chanwutk:
The axis being 90 degrees rotated is hard to read, so I change it to be 0.

visualization
This is the output chart before changing.


kanitw:
Please change PR title to be more accurate.


The title is currently confusing.  You’re just modifying one example but the title reads like u do it for all vega outputs.

PR #4940, opened 2019-05-06
Add FieldName alias to string

marcprux:
This PR externalizes strings that semantically refer to field names as a new "`FieldName`" type. In the typescript, this is just an alias to `string`, but since it is a separate `definitions` type in the JSON schema, this can help third-party tools to identify when a field reference is being sought. For example, the editor (https://github.com/vega/editor) could use this field to try to auto-complete field names from the data set associated with the spec.

This is similar to how color names are externalized to a "Color" type (https://github.com/vega/vega-lite/pull/4228).

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -55,7 +55,7 @@
         "groupby": {
           "description": "The data fields to group by. If not specified, a single group containing all data objects will be used.",
           "items": {
-            "type": "string"
+            "$ref": "#/definitions/FieldName"
           },
           "type": "array"
         }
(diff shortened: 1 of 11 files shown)


kanitw:
Hmm, did you incorrectly rebase? The diff now contains so many irrelevant changes.  Could you please double check? Thank you!

marcprux:
Sorry about that … I think something went haywire when I had to resolve a conflict. I've re-rebased and the commits looks OK to me know.

kanitw:
No problem at all.  It happens to me all the time too. :)

Issue #4886, opened 2019-04-17
Clamp scale extents when zooming


zufanka:
Originally [opened an issue for this on altair,](https://github.com/altair-viz/altair/issues/1419) got referred to vega-lite

I am making an interactive barchart with `altair` : `Chart.interactive()`. I would like it not to be able to go beyond the 0 and above 1000 on the X axis when zooming in and moving the bars around.

I am trying to set boundaries to the X axis with `alt.X(scale = alt.scale(domain=(0,1000)))` but that does not prevent the visualization going into out of these bounds on the X axis.

![image](https://user-images.githubusercontent.com/2411066/56299247-ef729100-6133-11e9-9b0c-02f10491deeb.png)

Could there be some way to set the limits to scales on interactive visualizations in vega-lite?

jheer:
In Altair, try adding a `clip=True` argument to your barMark method?

jakevdp:
I think @zufanka is asking for something different: the ability to constrain the interactivity of the axes to a particular range. So that, e.g. you could zoom freely within the range 0...1000, but the axis extent would be prevented from being larger than [0, 1000].

jheer:
Thanks @jakevdp, it appears I pattern-matched too quickly! I agree that this would be a great feature to include.

zufanka:
@jakevdp yes indeed! In the case of barcharts it's namely not that intuitive (for me at least) that I will be zooming in where the mouse pointer is. That is why constraining the axis would be a very welcome feature

Benjamin-Lee:
To add another use case, I'm trying to make a Gantt chart using relative times (_i.e._ we don't care the time a step happens, only its time relative to the other ones). To make it work, I'm offsetting from 0 Unix time, which works perfectly unless the user scrolls to the left:

<img width="803" alt="image" src="https://user-images.githubusercontent.com/13618205/61743115-959ceb00-ad62-11e9-9500-7a2a64a25821.png">

I want to limit scroll to only time range of the Gantt chart to prevent confusion with an "overflow".

rl2999:
I agree that this would be very useful. The unbounded interactions can be a bit distracting when you are exploring a dataset with integer quantities but looking between gaps of decimal point axes.

aldanor:
Agreed, this would be very nice to have.

(In fact if you think about it this might have even been a reasonable default option since there's always a limited amount of data and it doesn't make much sense to navigate and scroll beyond it.)

PBrockmann:
[...]

PR #4948, opened 2019-05-13
Add clamps for zooming and panning

djbarnwal:
Fix towards #4886 

Schema is of the form 
```json
"selection": {
    "zoom-signal": {
      "type": "interval", "bind": {"scales":true, "clamp": {"x": [0,250], "y": [0,50]}}
    }
```

WIP. Some parts do not work as expected yet. (There is a bug with panning and then zooming when both extents are mentioned and there is slight distortion in shape for singly clamped zoom.)

Currently, we are scaling down the clamp extent margins (when both clamps are mentioned) such that the initial ratio of the visualization is maintained throughout the zoom-in/zoom-out process.

1. The current system is designed such that if you keep zooming out, after a certain point there is some translation into effect to centre out the visualization relative to the given clamp extents. This way we have the same result at the end of the zoom out irrespective of the anchor point. Is this the way we want it to be or are we look at something else here?

2. Since we are clamping the zoom out feature, zoom is weird when the clamp range is "inside" the initialized domain range. I am not sure if there is a use case for this.  

3. We should probably add a warning when clamps are mentioned but the encoding domain is not. This leads to a more consistent workflow in my opinion (or should we default the domains to the clamp extents in absence of an explicit encoding domain)

4. The X axis right side of the visualization is slighlty padded for a given domain (250 extends to something around 255). I am not sure whether it was originally intended to have it this way or it's a bug. This causes a slight glitch when the clamp range is the same as the encoding domain range.

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1689,6 +1689,35 @@
       ],
       "type": "object"
     },
+    "BindClamp": {
+      "additionalProperties": false,
+      "properties": {
+        "clamp": {
+          "additionalProperties": {
+            "items": [
+              {
+                "type": "number"
+              },
+              {
+                "type": "number"
+              }
+            ],
+            "maxItems": 2,
+            "minItems": 2,
+            "type": "array"
+          },
+          "type": "object"
+        },
+        "scales": {
+          "type": "boolean"
+        }
+      },
+      "required": [
+        "scales",
+        "clamp"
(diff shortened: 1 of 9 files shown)


domoritz:
@djbarnwal @arvind What's the status of this change?

domoritz:
Closing due to inactivity. Please open a new pull request.

PR #4945, opened 2019-05-11
Add density, loess, and regression transforms.

jheer:
Changes:
- Add `density`, `loess`, and `regression` transforms.
- Add unit tests for new transforms.
- Add documentation and examples.

PR #4950, opened 2019-05-13
Update to Vega 5.4.0

kanitw:
Let's separate the upgrade from new VL feature in #4945 so the diff is smaller.


domoritz:
Sounds good. Tests are failing right now but looks good otherwise.

jheer:
FWIW, tests were passing on my branch. Looks like `tsc` is failing right away on this branch.

domoritz:
This branch is missing the updates to add `legendX` and `legendY` to legend properties.

domoritz:
I noticed in #4945 that a lot of gradient legends are black. Is that a Vega bug or do we need to generate something different from Vega-Lite?

Screen Shot 2019-05-14 at 17 24 07

kanitw:
Yeah, forgot to make this WIP.  I was adding this PR on a plane with very bad internet that can't D/L dependencies.

kanitw:
> I noticed in #4945 that a lot of gradient legends are black. Is that a Vega bug or do we need to generate something different from Vega-Lite?


Yeah, it's happening here too. 

@domoritz @jheer

jheer:
How does the VL scaffolding generate the SVGs? I looked at the output and see that the SVG `<defs>` block is missing and that the `fill` values (which should be `url(...)` gradient references) instead show up as `[Object object]` strings... hence the black gradients.

I tried to replicate this over in the Vega repository and failed, including testing using vega-cli generation at the command line. As a sanity check, is VL using the most up-to-date cli? (Not just in the package.json, but in terms of which binary and imports are being used?)

I'll try to investigate more if I have time.

jheer:
I found the issue in my local copy: Vega-Lite was pulling in vega-scenegraph 4.1.0, even though the latest required by vega/vega-cli 5.4 is vega-scenegraph 4.2.0. Maybe the problem is that the Vega peer dependency hadn't been updated appropriately? I neglected to do that in my regression branch, and pushed an update there.

jheer:
I believe I isolated the problem to the yarn.lock file. After I deleting the file and re-running `yarn`, the vega-scenegraph entries are corrected and the gradient examples no longer show up as changed files in the PR here: https://github.com/vega/vega-lite/pull/4945/files

domoritz:
I pushed an update with updates package.json and a new yarn.lock.

domoritz:
Looks good now.

Issue #3264, opened 2018-01-14
Include other useful transforms from Vega


kanitw:
## Definitely useful

- [ ] fold
- [ ] sample
- [ ] sequence
- [ ] stack

## Probably useful? 

- [ ] project

domoritz:
Sequence would be amazing for creating examples of mathematical functions. However, we have to decide whether it should go in data or transforms.

kanitw:
To me, sequence feels more natural as a special type of data source.

domoritz:
Yes, I think so too. It's just that Vega has it as a transform and I wanted to point out that it's not the most natural place and we can change that in Vega-Lite. Yet another reason to not just support all Vega transforms.

domoritz:
Sequence is covered in https://github.com/vega/vega-lite/issues/3864

domoritz:
Graticule is in https://github.com/vega/vega-lite/issues/3337

domoritz:
I don't think we need pivot in Vega-Lite, do we?

kanitw:
It's actually very useful.   For example, with `pivot`, we can create ["Multi Series Line Chart with Tooltip"](https://vega.github.io/vega-lite/examples/interactive_multi_line_tooltip.html) with the stocks  data (as the stocks data has a  long format, but we  need a pivoted format for "Multi Series Line Chart with Tooltip".

pjsier:
@kanitw I'm interested in submitting a PR to add `pivot`, but from a quick look at the code I'm not sure how to implement `producedFields` since they're generated from the data values. Should it just return an empty set, or would that cause issues?

domoritz:
@pjsier Thanks for looking into this. Figuring out the correct `producedFields` is super tricky here. I think we can only return `null` or `{}` in this case but we need to update the checks that use this field so that when we check for intersections and we have a `pivot` transform, we need to always say that there is an intersection. This will make it difficult to optimize the dataflow when there is a pivot transform.

pjsier:
@domoritz sounds good, thanks! I’ll try to put together a PR with that setup

kanitw:
"null" or "{}" would seems like "nothing".   I wonder if we should use a specific string literal "unknown" to make it explicit that we don't know at compile time.

PR #4958, opened 2019-05-15
Add `pivot` transform

pjsier:
Fix part of #3264 by adding `pivot` transform.

Not sure about how I'm handling `producedFields` now, but it looks like I'm handling the relevant check of intersection. This should be ready for an initial review


site/docs/transform/pivot.md
```diff
@@ -0,0 +1,57 @@
+---
+layout: docs
+menu: docs
+title: Pivot
+permalink: /docs/pivot.html
+---
+
+The **pivot** transform maps unique values from a field to new aggregated fields (columns) in the output stream. The transform requires both a field to pivot on (providing new field names) and a field of values to aggregate to populate the new cells. In addition, any number of groupby fields can be provided to further subdivide the data into output data objects (rows).
+
+Pivot transforms are useful for creating matrix or cross-tabulation data, acting as an inverse to the [fold](./fold) transform.
```
domoritz:
```suggestion
Pivot transforms are useful for creating matrix or cross-tabulation data, acting as an inverse to the [fold](fold.html) transform.
```

test/compile/data/pivot.test.ts
```diff
@@ -0,0 +1,85 @@
+/* tslint:disable:quotemark */
```
domoritz:
```suggestion
```

We use eslint now so this is not needed anymore

src/compile/data/optimizers.ts
```diff
@@ -73,7 +74,10 @@ export class MoveParseUp extends BottomUpOptimizer {
         parent.merge(node);
       } else {
         // don't swap with nodes that produce something that the parse node depends on (e.g. lookup)
-        if (fieldIntersection(parent.producedFields(), node.dependentFields())) {
+        if (
+          parent instanceof PivotTransformNode ||
+          fieldIntersection(parent.producedFields(), node.dependentFields())
```
domoritz: How about we add a check to `fieldIntersection` to not allow `PivotTransformNode` as a parent. We might even be able to enforce this with the typing system. 

I think we are missing a check in `run` below, don't we?

pjsier: It looks like `fieldIntersection` only accepts two `Set<string>` arguments, so I'm not sure the best way of doing that.

For `run` below, the only other ones I saw that called `hasIntersection` or `fieldIntersection` checked checked whether the node was a certain type (`TimeUnitNode` in `RemoveDuplicateTimeUnits` and `BinNode` in `MergeBins`). Let me know if I'm missing a check somewhere though

domoritz: Ahh, good point. I guess we could change the function to take two nodes so that we can ensure our assertion. I think that would be a good idea. We can still keep the existing function since we use it in our tests. You can just wrap it with another function. 

In `MergeBins.run` we only check the child, but not the parent. Is that sufficient?

pjsier: Updated with that function `nodeFieldIntersection`, and I also added it to `MergeBins.run` because I wasn't sure if it would be necessary to check

domoritz:
Thanks. I made a new PR to automatically update the examples on travis: https://github.com/vega/vega-lite/pull/4996

PR #4959, opened 2019-05-15
[Build] Don't use project references. Fixes #4856

### LINKED ISSUE #4856: Project references are not compiled automatically
We are using project references as described in https://www.typescriptlang.org/docs/handbook/project-references.html. That means that you have to compile after checking out Vega-Lite. To best resolve this issue, we recommend that you compile in the background with `yarn tsc -b -w`.

We should fix our setup so this is not necessary. 

See https://github.com/Microsoft/TypeScript/issues/26913

domoritz:
We don't use project references in the main tsconfig.json file anymore but we still only compile what's necessary when you build.


package.json
```diff
@@ -44,15 +44,18 @@
...
     "prettierbase": "prettier '{src,test,test-runtime,site,typings}/**/*.{ts,md,css}'",
     "format": "tslint -p . --fix && npm run prettierbase -- --write",
     "lint": "tslint -p . && npm run prettierbase -- --check",
     "test": "jest test/ && npm run lint && npm run schema && jest examples/ && npm run test:runtime",
     "test:inspect": "node --inspect-brk ./node_modules/.bin/jest --runInBand test",
     "test:runtime": "TZ=America/Los_Angeles jest test-runtime/",
     "test:runtime:generate": "rm -Rf test-runtime/resources && VL_GENERATE_TESTS=true npm run test:runtime",
-    "watch:build": "npm run build:only && concurrently --kill-others -n Typescript,Rollup 'tsc -b src -w' 'rollup -c -w'",
-    "watch:site": "concurrently --kill-others -n Typescript,Rollup 'tsc -b site --watch' 'rollup -c site/rollup.config.js --watch'",
+    "watch:src": "npm run tsc:src -- -w",
```
kanitw: Why do we need both `watch:src` and `watch:build`?  

I'm a bit worried that we're adding too many commands that no one know how to use except you.

package.json
```diff
@@ -44,15 +44,18 @@
     "renameschema": "scripts/rename-schema.sh",
     "presite": "npm run data && npm run build:site && npm run build:toc && npm run build:versions && scripts/create-example-pages",
     "site": "bundle exec jekyll serve --incremental",
+    "tsc:src": "tsc -b src/tsconfig.src.json",
```
kanitw: Do we really need tsc:src?

package.json
domoritz: I deleted it and updated the contribution docs.

package.json
domoritz: We use this in  bunch of commands and also in the contributing guide.

PR #4961, opened 2019-05-15
Add docs for usermeta

domoritz:


kanitw:
Is usermeta top-level only or spec-level?

domoritz:
For now only top level but we should probably move it to the spec level.

domoritz:
Vega has it at the scope level.

PR #4967, opened 2019-05-16
Expand type of TitleMixins title field to support array of FieldDefBase

Saba9:
Replaces #4267. Allows for five more example specs to pass transform extract tests. This is because title merging behavior differs between implicit and explicitly defined titles.

### DIFF
--- a/test/transformextract.test.ts
+++ b/test/transformextract.test.ts
@@ -14,7 +14,6 @@ describe('extractTransforms()', () => {
   const failsList = new Set([
     'area_temperature_range.vl.json',
     'bar_argmax.vl.json',
-    'bar_argmax_transform.vl.json',
     'bar_aggregate_count.vl.json',
     'bar_aggregate_sort_by_encoding.vl.json',
     'bar_aggregate_sort_mean.vl.json',
@@ -45,7 +44,6 @@ describe('extractTransforms()', () => {
     'interactive_layered_crossfilter_discrete.vl.json',
     'interactive_layered_crossfilter.vl.json',
     'interactive_seattle_weather.vl.json',
-    'joinaggregate_mean_difference.vl.json',
     'layer_bar_dual_axis_minmax.vl.json',
     'layer_bar_month.vl.json', // data transform switches the order
     'layer_dual_axis.vl.json', // tooltip does not use the correct field name
@@ -72,7 +70,6 @@ describe('extractTransforms()', () => {
     'layer_point_errorbar_stdev.vl.json',
     'layer_precipitation_mean.vl.json',
     'layer_rect_extent.vl.json',
-    'layer_scatter_errorband_1D_stdev_global_mean.vl.json',
     'line_calculate.vl.json',
     'line_color_binned.vl.json',
     'line_max_year.vl.json',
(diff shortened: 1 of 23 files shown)


test/transformextract.test.ts
```diff
@@ -106,16 +103,16 @@ describe('extractTransforms()', () => {
...
     'vconcat_weather.vl.json',
     'window_top_k_others.vl.json'
   ]);
   for (const file of fs.readdirSync(specsDir)) {
+    // If true make sure that specs in the failsList do indeed fail
+    const TEST_EXPECTED_FAILS = false;
     const filepath = specsDir + file;
-    if (filepath.slice(-5) === '.json' && !failsList.has(file)) {
-      it(`should compile ${filepath} to the same spec`, () => {
+    if (filepath.slice(-5) === '.json' && (!failsList.has(file) || TEST_EXPECTED_FAILS)) {
```
Saba9: I added back the ability for these tests to also check to make sure that items in the failing list do actually fail because it makes it a lot easier to see what specs are positively impacted by changes. This feature can be turned on or off via the TEST_EXPECTED_FAILS boolean.

I'd be fine with removing this if it's seen as unnecessary.

domoritz: When did we remove this? I think it makes a lot of sense to fail when new specs are passing so that we keep the list of failing examples up to date. I'd even remove the boolean.

Saba9: Looks like it was removed by this commit **https://github.com/vega/vega-lite/commit/4bebfe9c3c62cc60c3d1bda7290e2067f25f2f37#diff-4fecb77be8d42e352cbe3327b4b49aaeR116**

domoritz:
Hmm, reading this I am not actually a huge fan. It seems kind of dirty to expand the type of title just to support better merging. I can understand if we change the title component that we use internally but changing what title can be, seems a bit overkill to me. @kanitw what do you think?

kanitw:
> I can understand if we change the title component that we use internally but changing what title can be, seems a bit overkill to me.

I agree that if changing internal representation is sufficient we should aim for that. 

However, it seems like there is some problems with transform extraction if we don't do this.
That said, reading the old PRs, it isn't clear what is the case that's problematic anymore.

@Saba9 -- could you provide a specification of a problematic case?  So that would make it clearer why we need to do this.

Saba9:
The issue which we're able to solve by allowing for titles of type `FieldDefBase[]` is the following. Consider the following layer spec:

```
{
  "layer": [
    {
      "mark": "rule",
      "encoding": {
        "x": {
          "aggregate": "mean", "field": "a",  "type": "quantitative"
        }
      }
    },
    {
      "mark": "rule",
      "encoding": {
        "x": {
          "title": "foo", "field": "b", "type": "quantitative"
        }
      }
    }
  ]
}
```
The resulting visualization will have an x-axis title of "foo". This is because in our second layer we explicitly specify an x-axis title which results in Vega-Lite ignoring all implicit x-axis titles such as the one implied by the first layer. Before this PR, if we were to extract the transforms from this specification we'd get the following spec:
```
{
  "transform": [
    {"aggregate": [{"op": "mean", "as": "mean_a", "field": "a"}], "groupby": ["b"]}
  ],
  "layer": [
    {
      "mark": "rule",
      "encoding": {
        "x": {
          "title": "Mean of a", "field": "mean_a", "type": "quantitative"
        }
      }
    },
    {
      "mark": "rule",
      "encoding": {
        "x": {
          "title": "foo", "field": "b", "type": "quantitative"
        }
      }
    }
  ]
}
```
In this case the resulting x-axis title would be "Mean of a, foo" because when we extracted the transformations from the first layer we set the x-axis title explicitly. This PR solves this issue by using setting titles of type `FieldDefBase[]` which are treated as implicit.

kanitw:
@Saba9  That makes sense.  

@domoritz: If you worry about exposing `FieldDefBase[]` to title, perhaps we could use `@hide` to hide the `FieldDefBase` part from the external schema?

domoritz:
@Saba9 and I decided that since this will go away when we have tracking, to look into tracking a short while. Once we know how difficult tracking is, we can decide to merge this or not.

kanitw:
FYI this won't really go away with tracking.

domoritz:
Closing due to inactivity. Please open a new pull request.

kanitw:
replaced with https://github.com/vega/vega-lite/pull/6643

Issue #4611, opened 2019-02-27
Add gradient definition


kanitw:
As added in Vega in https://github.com/vega/vega/pull/1648/commits/4ca11a254e3e4937a192e81cd3f5f219c687059e

domoritz:
We might get this for free if we use the Vega typings.

kanitw:
Not in encoding though, at least we need to modify the type of valueDef for color/fill channel.

PUNITBATRA:
@kanitw  Can I work on this issue as a beginner? In this issue, I have to replicate it like vega or can you provide any more information about it.

kanitw:
The tasks is to

1) Expose the gradient definition in the typings of
- [ ] value definition of color/fill/stroke encodings 
- [ ] color/fill/stroke properties of mark definition and config

2) Create new line chart example with area gradient.

PUNITBATRA:
Hi @kanitw can you please tell me the specific folder where I have to do the task 1 as specified by you above or I have to do it by creating a new file(Add gradient) here:- [https://github.com/PUNITBATRA/vega-lite/tree/master/site/docs].

kanitw:
You should modify existing typings for `Encoding`, `MarkDef` and `MarkConfig`.

You may need to update `vega-typings` for the new gradient too, if they haven't been updated yet.

PUNITBATRA:
@kanitw As mentioned above (task 2) by you I have created the line chart with area gradient by reading the docs as shown below in the screenshot. In this, I have taken data values randomly. Is it fine or I have to take the data from some specific JSON file?
Please suggest me the changes which you want in the task 2...
![Screenshot from 2019-03-12 19-58-28](https://user-images.githubusercontent.com/34002809/54230496-85f7d500-452c-11e9-9374-dec012760975.png)

kanitw:
I'd modify this example: https://vega.github.io/editor/#/examples/vega-lite/area_overlay and check if it is referenced in any docs in `site/docs`.  
If not, just modify the example directly in examples/specs.  If it is referenced, create a new area_overlay_gradient.

You spec won't pass the schema validation when you have to do 1. first before submitting a PR.

kanitw:
@PUNITBATRA any update on this?

PUNITBATRA:
[...]

PR #4970, opened 2019-05-16
Add gradient definition, gradient mark and encoding

hungngo97:
Check point to see if I am heading in the right direction for adding gradient definition


kanitw:
We should first agree on whether we want `color: {value: Gradient}` or just `color: Gradient`, provided that `GradientDef` is just [similar to Vega](https://vega.github.io/vega/docs/types/#Gradient).

(In other words, `Gradient = GradientLinear | GradientRadial` from [vega-typings](https://github.com/vega/vega/blob/master/packages/vega-typings/types/spec/encode.d.ts#L86)).  


The latter is more concise.  However, the former will
- make it more consistent across mark and encoding (`mark: {color: XXX}` is equilvalent to `{encoding: {color: {value: XXX}}`). (I think consistency beats concision here.)
- be easier to implement as we just need to add `Gradient` to the `ColorValueDef` type. (Minor point)

kanitw:
In any case, please fork from master to avoid conflicts. :)

kanitw:
Please also refer to relevant issue.  (e.g., "Fix: https://github.com/vega/vega-lite/issues/4611")  :)

kanitw:
Some check lists to help guide you: 

- [ ] Add to encoding
- [ ] Add to marks
- [ ] Add an example of line + area with gradient

hungngo97:
Hi Ham I followed your directions and work with Dominik the other day but it seems to still have some conflict with the compile phase :(

src/encoding.ts
```diff
@@ -130,21 +131,21 @@ export interface Encoding<F extends Field> {
    * _Note:_ When using `stroke` channel, `color ` channel will be ignored. To customize both stroke and fill, please use `stroke` and `fill` channels (not `stroke` and `color`).
    */
 
-  stroke?: ColorFieldDefWithCondition<F> | ColorValueDefWithCondition<F> | GradientValueDef<F>;
+  stroke?: ColorFieldDefWithCondition<F> | ColorValueDefWithCondition<F> | Gradient;
```
kanitw: Not sure why you need `| Gradient` here given you already added it to colorvaluedef  already

src/encoding.ts
```diff
@@ -130,21 +131,21 @@ export interface Encoding<F extends Field> {
...
 
   /**
    * Opacity of the marks.
    *
    * __Default value:__ If undefined, the default opacity depends on [mark config](https://vega.github.io/vega-lite/docs/config.html#mark)'s `opacity` property.
    */
-  opacity?: NumericFieldDefWithCondition<F> | NumericValueDefWithCondition<F>;
+  opacity?: NumericFieldDefWithCondition<F> | NumericValueDefWithCondition<F> | Gradient;
```
kanitw: I don't think opacity makes sense with gradient?

src/encoding.ts
```diff
@@ -130,21 +131,21 @@ export interface Encoding<F extends Field> {
...
 
   /**
    * Fill opacity of the marks.
    *
    * __Default value:__ If undefined, the default opacity depends on [mark config](https://vega.github.io/vega-lite/docs/config.html#mark)'s `fillOpacity` property.
    */
-  fillOpacity?: NumericFieldDefWithCondition<F> | NumericValueDefWithCondition<F>;
+  fillOpacity?: NumericFieldDefWithCondition<F> | NumericValueDefWithCondition<F> | Gradient;
```
kanitw: I don't think opacity makes sense with gradient?

kanitw:
Umm, looks like this is still broken.  Please let me know when you're done.

src/encoding.ts
hungngo97: Yeah that was me rebasing to master yesterday haha :D I fixed following your instructions but it still fail to compile during the parsing phase since SignalRef expects signal is String not Gradient. I think it was where I declare the Gradient from vega-typings, should it be directly GradientLinear or should it be a string?

hungngo97:
Hey Ham, I followed your instructions but currently my build still giving me error at the compile phase when VgValueRef is checking the type is a string or not. I suspect that my declarations for Gradient is wrong by importing directly from Vega-typings, should it be strings of some sort?

kanitw:
Umm, so:

1) the code that you have here looks like you either haven't followed the instruction or haven't pushed the latest code.   (It's hard to help you if you don't push the code.) 

2) I don't know where your error is, but try to follow the logic and see if you want to modify it.  Ultimately, we want to output gradient definition object when the input is gradient object.   I suggest you give it another try.  :)

hungngo97:
Haha yeah thank you Ham, seeing many errors come up during compile phase make me panic 👍 I got it working now in the latest push, please take a look. I will get some gradient examples on the doc by next week

domoritz:
Travis is still failing. Can you fix it?

domoritz:
Let us know when this PR is ready for review.

kanitw:
@hungngo97  -- any updates?

hungngovinh:
Yeah sorry everyone. It is finals week so I will get back to this issue next week :D

domoritz:
@hungngovinh Good luck with exams. Let me know if you have any questions. I will be around.

domoritz:
Please use https://github.com/vega/vega-lite/pull/5077

PR #4973, opened 2019-05-17
Merge TimeUnit nodes

Saba9:
Add optimization to merge adjacent TimeUnit nodes in DataFlow tree.

### DIFF
--- a/test/compile/data/timeunit.test.ts
+++ b/test/compile/data/timeunit.test.ts
@@ -55,7 +55,9 @@ describe('compile/data/timeunit', () => {
         }
       });
       const timeUnitNode = TimeUnitNode.makeFromEncoding(null, model);
-      expect(timeUnitNode.hash()).toBe('TimeUnit {"month_a":{"as":"month_a","field":"a","timeUnit":"month"}}');
+      expect(timeUnitNode.hash()).toBe(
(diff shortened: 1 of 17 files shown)


test/compile/data/timeunit.test.ts
```diff
@@ -55,7 +55,9 @@ describe('compile/data/timeunit', () => {
         }
       });
       const timeUnitNode = TimeUnitNode.makeFromEncoding(null, model);
-      expect(timeUnitNode.hash()).toBe('TimeUnit {"month_a":{"as":"month_a","field":"a","timeUnit":"month"}}');
+      expect(timeUnitNode.hash()).toBe(
+        'TimeUnit {"{\\"as\\":\\"month_a\\",\\"field\\":\\"a\\",\\"timeUnit\\":\\"month\\"}":{"as":"month_a","field":"a","timeUnit":"month"}}'
```
domoritz: Use ` to avoid escaping.

Saba9: I think it remains the same with string literals since backslash still needs escaping

domoritz: Good point. I think we should switch to using ' for escaping strings in all of Vega-Lite but that's a separate issue.

PR #4983, opened 2019-05-21
Remove outdated comment

domoritz:


kanitw:
Why is this outdated?  (We should do it)

domoritz:
The return type is already a VG spec. I think we've done it but wanted to check with you.

kanitw:
Oh ok got it now. We did it down there.

PR #4987, opened 2019-05-24
Remove "type" field from "x2" and "y2" 

marcprux:
Remove "type" field from "x2" and "y2" since they are disallowed by the schema, so this example schema is technically invalid.


domoritz:
Thank you

PR #4993, opened 2019-05-27
fix: Parse the dimension field of path marks

### LINKED ISSUE #4988: Quantitative data from CSV is sorted alphabetically rather than numerically
Reported in https://github.com/altair-viz/altair/issues/1531

Consider the following spec ([editor link](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAutIC4yghSBXATgGxSgAt54AHAZ2QHpqdYB3AOgHMBLeIrAIywoFMcAYwD2AO3j8JTUQFtqATxHiWReXETVZsCpJzUyIhoID6Q3ADd+MihcjgwAXwA0DyNpwBrQsEjwFZPyEkHhsYkHOUIZh8ITwOFj8Lm5SotBhLD4OmAAePn4BQaiQAI5YsBIcCGxWkJGQAGZs-HjQwQDqYdBgFIH83QAU8hQAlJDJGFAK+f6BwWUV8FVLtfVNLW3FAPJY5LtghsY4YAOe7WOODpeOQA)):
```json
{
  "data": {
    "url": "https://raw.githubusercontent.com/yoonghm/data/master/power_curve.csv"
  },
  "mark": {"type": "line", "point": true},
  "encoding": {
    "x": {"type": "quantitative", "field": "Wind speed (m/s)"},
    "y": {"type": "quantitative", "field": "Output power (kW)"}
  }
}
```
visualization (94)

I would expect based on the quantitative encoding that the x-axis would be sorted numerically, but it is sorted alphabetically by the string representation of the number. This of course can be fixed by specifying an appropriate mapping for ``data.format.parse``, but I think this default behavior should be considered a bug.

domoritz:
Fixes #4988

### DIFF
--- a/src/compile/data/formatparse.ts
+++ b/src/compile/data/formatparse.ts
@@ -188,10 +188,14 @@ export class ParseNode extends DataFlowNode {
       });
     }
 
-    // also parse the dimension field of path marks
+    // Parse quantitative dimension fields of path marks as numbers so that we sort them correctly.
     if (isUnitModel(model)) {
       const {mark, markDef, encoding} = model;
-      if (isPathMark(mark)) {
+      if (
+        isPathMark(mark) &&
+        // No need to sort by dimension if we have a connected scatterplot (order channel is present)
+        !model.encoding.order
+      ) {
         const dimensionChannel = markDef.orient === 'horizontal' ? 'y' : 'x';
         const dimensionChannelDef = encoding[dimensionChannel];
         if (
--- a/src/compile/data/formatparse.ts
+++ b/src/compile/data/formatparse.ts
@@ -194,7 +194,11 @@ export class ParseNode extends DataFlowNode {
       if (isPathMark(mark)) {
         const dimensionChannel = markDef.orient === 'horizontal' ? 'y' : 'x';
         const dimensionChannelDef = encoding[dimensionChannel];
-        if (isFieldDef(dimensionChannelDef) && dimensionChannelDef.type === 'quantitative') {
+        if (
+          isFieldDef(dimensionChannelDef) &&
+          dimensionChannelDef.type === 'quantitative' &&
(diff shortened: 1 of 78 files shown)


domoritz:
This is tricky, we have to add parse also when we normalize or extract transforms. This takes a lot of time to get right.

domoritz:
image

src/compile/data/formatparse.ts
```diff
@@ -177,13 +182,28 @@ export class ParseNode extends DataFlowNode {
...
-            throw new Error(
-              `Non-secondary channel ${channel} must have type in its field definition ${JSON.stringify(fieldDef)}`
-            );
+            throw new Error(log.message.nonSecondaryMustHaveType(channel, fieldDef));
           }
         }
       });
     }
+
+    // also parse the dimension field of path marks
```
kanitw: Describe why dimension field should have its special case?

kanitw:
> This is tricky, we have to add parse also when we normalize or extract transforms. This takes a lot of time to get right.

Can you elaborate why such cases would be tricky?

domoritz:
>> This is tricky, we have to add parse also when we normalize or extract transforms. This takes a lot of time to get right.

> Can you elaborate why such cases would be tricky?

The tricky bit is that the transform extraction actually. The problem is that we don't create parse nodes for derived fields and when we extract e.g. an aggregate, we don't parse the value used in the aggregate first. 

It could be nice to just parse all quantitative fields (then we definitely always parse when we need to as I'm still not sure we cover all the cases) but I don't see how we get there unless we have tracking of fields or correct transform extraction that adds a parse node before extracted aggregates. cc @Saba9

domoritz:
> Also test coverage?

Everything relevant in this PR has been fully covered. Just the log messaged I moved hasn't (yes, I shouldn't have done this here).

PR #4996, opened 2019-05-28
feat: Add `pivot` transform

domoritz:
Copy of https://github.com/vega/vega-lite/pull/4958 by @pjsier

Merge after https://github.com/vega/vega-lite/pull/5257


domoritz:
@pjsier Any idea why histogram_log.svg is changing in this PR?

pjsier:
@domoritz I'm not sure. I don't see anything that should have caused a change to any of the other examples

pjsier:
@domoritz realized that there's a typo in `nodeFieldIntersection` that's likely causing this. Right now the call to `fieldIntersection` is using `child.producedFields()` but it should be using `child.dependentFields()`

src/compile/data/optimizers.ts
```diff
@@ -398,3 +399,11 @@ export class MergeBins extends BottomUpOptimizer {
     return this.flags;
   }
 }
+
+/**
+ * Wraps fieldIntersection so that it can be called with parent and child nodes directly
+ * and check whether or not the parent node is a pivot transform
+ */
+export function nodeFieldIntersection(parent: DataFlowNode, child: DataFlowNode): boolean {
+  return parent instanceof PivotTransformNode || fieldIntersection(parent.producedFields(), child.producedFields());
```
pjsier: This should fix the issue where histogram examples are changing

```suggestion
  return parent instanceof PivotTransformNode || fieldIntersection(parent.producedFields(), child.dependentFields());
```

Saba9:
Is there a need for keeping fields produced by the pivot which aren't explicitly mentioned elsewhere in the spec? 

If not, a possible solution could be to have all of the produced fields fall under a user defined namespace. This would mean that if a user pivoted on a column with values of 'a', 'b', and 'c' and specified the namespace of 'foo' they would refer to the produced fields as 'foo.a', 'foo.b', etc. We could then determine what fields are produced by the transformation by scanning over referenced fields.

We do run into the issue of already existing fields potentially being overwritten by the output of the pivot transform despite the lack of reference to the outputted field. One solution to this could be the addition of a feature to Vega's pivot transform which would allow the user to provide a list of the values they want promoted to columns. Another way of solving this might be to filter the pivot field to only those referenced values prior to the transformation.

A downside to this solution is that I believe we'd have to process field names during assembly to remove namespace prefixes so that the resulting Vega would actually work. Also, I'm not sure if the described Vega feature is all that useful except for this particular case.

domoritz:
> Is there a need for keeping fields produced by the pivot which aren't explicitly mentioned elsewhere in the spec?

I don't understand the question. Where do we keep fields? I thought the pivot only produces the fields which are in the groupby as well as one field for each distinct value in the pivoted column. But I think you are onto something here that we could use.

> If not, a possible solution to the problem of determining produced fields could be to have all of the produced fields fall under a user defined namespace. 

So would users have to use that namespace when they reference one of the pivoted fields? I don't quite see why this is necessary. We can already check what fields are being used in the spec after the pivot and not in the groupby or otherwise produced. With proper tracking this shouldn't be too hard, I think.

domoritz:
Ahh, the PR is much cleaner now. Thanks for the fix @pjsier.

pjsier:
@domoritz Should I open a PR against this branch to resolve the merge conflict, or is there a better way of handling it? Thanks!

domoritz:
Fixed

pjsier:
@domoritz is there anything else I can help with to finish up this PR? It would be great for this to get merged, thanks!

domoritz:
Let's get this into VL 4.0. Can you merge master?

pjsier:
@domoritz Sure, want me to rebase and open a separate PR?

domoritz:
That would be great. Thank you!

pjsier:
Great! That's done in #5248

domoritz:
Thank you. I moved it here.

src/compile/data/optimizers.ts
pjsier: Missed this fix in the rebase somehow, but this suggestion should fix it

```suggestion
  return parent instanceof PivotTransformNode || fieldIntersection(parent.producedFields(), child.dependentFields());
```

examples/compiled/histogram_log.vg.json
```diff
@@ -41,6 +40,7 @@
           "maxbins": 10,
           "extent": {"signal": "bin_maxbins_10_log_x_extent"}
         },
+        {"type": "formula", "expr": "log(datum.x)/log(10)", "as": "log_x"},
```
domoritz: Oh, this is not good.

examples/compiled/interactive_layered_crossfilter.vg.json
```diff
@@ -13,17 +13,17 @@
       "transform": [
         {
           "type": "extent",
```
domoritz: this also isn't good

examples/compiled/histogram_log.vg.json
pjsier: @domoritz I think this is because of the bug that I fixed earlier in the field intersection function, but it looks like it got dropped in the rebase. I re added the suggestion inline that fixed it before

src/compile/data/optimizers.ts
domoritz: Ahh, okay. I was wondering. I think I will rewrite the logic a bit. 

I'm thinking that the intersectin check should return true if the produced or dependent fields is undefined since this means we don't have information to be confident that there is no intersection.

domoritz:
I refactored dataflow nodes to be abstract and implemented dependent and produced fields for all node types. I also rewrote the intersection check to return true if any input is `undefined`.

domoritz:
As a side effect, we are now merging FilterInvalidNodes better (since I implemented a better hash function). This is great but it means we are doing more than one thing in this PR. Let me pull things out.

domoritz:
Done in https://github.com/vega/vega-lite/pull/5257

src/vega.schema.ts
```diff
@@ -392,6 +392,15 @@ export interface VgIdentifierTransform {
   as: string;
 }
 
+export interface VgPivotTransform {
```
domoritz: It would be nice to move this (and al other transforms) to vega-typings.

PR #4998, opened 2019-05-28
Fix dependentFields calculation for JoinAggregate, Stack, and Window nodes

Saba9:
As is, the dependentFields() method in JoinAggregateNode, StackNode, and WindowNode attempt to iterate over optional fields resulting in exceptions when these fields aren't defined. This PR fixes this by checking to see if these fields are defined before iterating over them, and adds relevant tests.

### DIFF
--- a/src/compile/data/joinaggregate.ts
+++ b/src/compile/data/joinaggregate.ts
@@ -26,7 +26,9 @@ export class JoinAggregateTransformNode extends DataFlowNode {
   public dependentFields() {
     const out = new Set<string>();
 
-    this.transform.groupby.forEach(f => out.add(f));
+    if (this.transform.groupby) {
+      this.transform.groupby.forEach(f => out.add(f));
+    }
     this.transform.joinaggregate
       .map(w => w.field)
       .filter(f => f !== undefined)
--- a/src/compile/data/stack.ts
+++ b/src/compile/data/stack.ts
@@ -179,10 +179,15 @@ export class StackNode extends DataFlowNode {
 
     out.add(this._stack.stackField);
 
-    this.getGroupbyFields().forEach(out.add);
-    this._stack.facetby.forEach(out.add);
-    const field = this._stack.sort.field as string;
-    isArray(field) ? field.forEach(out.add) : out.add(field);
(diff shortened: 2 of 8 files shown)


src/compile/data/joinaggregate.ts
```diff
@@ -26,7 +26,9 @@ export class JoinAggregateTransformNode extends DataFlowNode {
   public dependentFields() {
     const out = new Set<string>();
 
-    this.transform.groupby.forEach(f => out.add(f));
```
domoritz: Instead you can write (foo || []).forEach or even better also use a for of loop.

src/compile/data/stack.ts
```diff
@@ -179,10 +179,15 @@ export class StackNode extends DataFlowNode {
...
-    this.getGroupbyFields().forEach(out.add);
-    this._stack.facetby.forEach(out.add);
-    const field = this._stack.sort.field as string;
-    isArray(field) ? field.forEach(out.add) : out.add(field);
+    this.getGroupbyFields().forEach(f => out.add(f));
+    this._stack.facetby.forEach(f => out.add(f));
+
+    const field = this._stack.sort.field;
+    if (isArray(field)) {
+      (field as [string]).forEach(f => out.add(f));
```
domoritz: Why do you need to force it to be a string array?

src/compile/data/window.ts
```diff
@@ -26,8 +26,12 @@ export class WindowTransformNode extends DataFlowNode {
   public dependentFields() {
     const out = new Set<string>();
 
-    this.transform.groupby.forEach(f => out.add(f));
-    this.transform.sort.forEach(m => out.add(m.field));
+    if (this.transform.groupby) {
```
domoritz: Same as above

src/compile/data/stack.ts
Saba9: If it's not forced as a string array I receive a type error because it could also be an array of `FieldRef` or `ExprRef` and out is a set of strings. Would it be better to not force the type and call `toString()` on `f` in the `forEach`?

domoritz: I thought we agreed in the meeting to change the types so that the field is always a string.

PR #5001, opened 2019-05-29
[WIP] Make EncodingSortField.field required

marcprux:
If `EncodingSortField.field` is not required, then `EncodingSortField` has no required fields at all, meaning it can match empty JSON. Having the JSON "sort": { } be interpreted to mean that the scale should be sorted by that field doesn't seem intuitive.

This PR makes `field` a required property of `EncodingSortField`. This could lead to redundant field specifications in the case where the user wants to sort by the same field as the scale, but otherwise is harmless.

An alternative solution could be to have a discriminator property (e.g. a required "sortType": "field" prop), but it seems easier to just make `field` non-optional.

### DIFF
--- a/src/sort.ts
+++ b/src/sort.ts
@@ -35,10 +35,8 @@ export const DEFAULT_SORT_OP = 'mean';
 export interface EncodingSortField<F> {
   /**
    * The data [field](https://vega.github.io/vega-lite/docs/field.html) to sort by.
-   *
-   * __Default value:__ If unspecified, defaults to the field specified in the outer data reference.
    */
-  field?: F;
+  field: F;
   /**
    * An [aggregate operation](https://vega.github.io/vega-lite/docs/aggregate.html#ops) to perform on the field prior to sorting (e.g., `"count"`, `"mean"` and `"median"`).
    * An aggregation is required when there are multiple values of the sort field for each encoded data field.


src/sort.ts
```diff
@@ -35,10 +35,8 @@ export const DEFAULT_SORT_OP = 'mean';
 export interface EncodingSortField<F> {
   /**
    * The data [field](https://vega.github.io/vega-lite/docs/field.html) to sort by.
-   *
-   * __Default value:__ If unspecified, defaults to the field specified in the outer data reference.
    */
-  field?: F;
+  field: F;
```
kanitw: Field is not required when the "op" is count.  So this would be also incorrect. 

To make the typing perfectly correct, we need to separate count from other aggregate in sort field. However, that will cause a problem downstream in a library such as Atlair as they rely on a class called EncodingSortField.  

Thus, I think the current syntax as is seems to be the least problematic solution.

kanitw:
I'll close this issue for now and add a PR to add a comment about this https://github.com/vega/vega-lite/pull/5015.

If you disagree, please feel free to comment on the new issue / reopen this issue if needed.

marcprux:
That makes sense. I still do have an issue with the fact that `sort` can legally be `{}`, which complicates my own usage of the schema (due to unrelated issues). Ideally we'd be able to say that *either* `sort` or `op` is required, but there's no simple way to implement that AFAIK.

kanitw:
Yeah, I wish there is a way too.

PR #5004, opened 2019-05-30
DOC: Add rolling mean example

eitanlees:
I recently added a rolling mean example to Altair (https://github.com/altair-viz/altair/pull/1540) and thought it would be a nice addition to the vega-lite examples. 

visualization (4)


domoritz:
Replaced by https://github.com/vega/vega-lite/pull/5006 so we can travis to compile things for us.

PR #5005, opened 2019-05-30
refactor: upgrade to Typescript 3.5, update code

domoritz:
I had to disable a test that was failing: https://github.com/vega/vega-lite/issues/5285. If you tell me what the expected behavior is, I can fix it.


domoritz:
Rebased onto master. @kanitw @arvind Can you take a look?

src/compile/axis/parse.ts
```diff
@@ -296,51 +296,51 @@ function getProperty<K extends keyof AxisComponentProps>(
 
   switch (property) {
     case 'scale':
-      return model.scaleName(channel);
+      return model.scaleName(channel) as any;
```
kanitw: Why so many `as any` here?  -- Can we do something like direction in the legend?

src/compile/legend/parse.ts
```diff
@@ -143,21 +143,27 @@ function getProperty<K extends keyof LegendComponentProps>(
...
+        scaleType
+      }) as LegendComponentProps[K];
 
     case 'format':
       // We don't include temporal field here as we apply format in encode block
       if (isTimeFormatFieldDef(fieldDef)) {
         return undefined;
       }
-      return numberFormat(fieldDef, legend.format, model.config);
+      return numberFormat(fieldDef, legend.format, model.config) as any;
```
kanitw: Why is `direction` the only one that can use `LegendComponentProps[K]`?

src/encoding.ts
```diff
@@ -422,7 +422,7 @@ export function normalizeEncoding(encoding: Encoding<string>, markDef: MarkDef):
     ) {
       if (channelDef) {
         // Array of fieldDefs for detail channel (or production rule)
-        normalizedEncoding[channel] = (isArray(channelDef) ? channelDef : [channelDef]).reduce(
+        (normalizedEncoding[channel] as any) = (isArray(channelDef) ? channelDef : [channelDef]).reduce(
```
kanitw: Why casting the LHS?

test/compile/facet.test.ts
```diff
@@ -74,7 +74,7 @@ describe('FacetModel', () => {
     it('applies number format for fieldref of a quantitative field', () => {
       const model = parseFacetModelWithScale({
         facet: {
-          column: {field: 'a', type: 'quantitative', format: 'd'}
```
kanitw: Why dropping format here?

kanitw:
For  #5285, I wonder why the behavior is changing in this PR?  (I gotta board the plane, so haven't got the time to figure out.)

src/compile/legend/parse.ts
domoritz: We can do as any or as key of props. I don't think tye latter gives us any benefits so I'd just use any everywhere.

src/encoding.ts
domoritz: Otherwise, we cannot assign to it. You can try to remove it but there is no type safe way to assign here. Note that previous typescript versions did this implicitly so we don't lose any type safety here.

test/compile/facet.test.ts
domoritz: I think it was invalid. I have to check again.

domoritz:
The behavior is not changing but we used a property that didn't exist and now TypeScript notices. Basically, the test was testing a single case (and a boring one) only.

test/compile/facet.test.ts
domoritz: Yep, we need `header: {format: 'f'}`

domoritz: I fixed it.

src/compile/legend/parse.ts
kanitw: Yep, in general we better avoid `any` since it normally means "I'm giving up on TS for this part.

PR #5006, opened 2019-05-30
DOC: Add rolling mean example

domoritz:
@eitanlees recently added a rolling mean example to Altair (https://github.com/altair-viz/altair/pull/1540) and thought it would be a nice addition to the vega-lite examples. 

visualization (4)


eitanlees:
Thanks for making this happen! I am still a little confused about the contribution process. I thought adding an example to `examples/specs` and adding an entry in `_data/examples.json` would do the trick. 

I think there was a problem with formatting. Is there something I can run to check the formatting before submitting a PR?

I am just thinking how to make this process easier in the future. 

I read over `CONTRIBUTING.md` but I still messed things up. Maybe a more details could be provided about adding examples to the main gallery.

domoritz:
Travis compiles the examples for us for it then tries to push them to your branch, which it fails to do. @kanitw might have an insight here.

PR #5009, opened 2019-06-01
Fix reference problem in `log/message.ts`

kanitw:


domoritz:
LGTM

kanitw:
lol -- your eslint option doesn't allow this.   
```
322:10  error  Move function declaration to program root  no-inner-declarations
```


So should we relax the lint or ignore the problem?

domoritz:
I do think this is bad style so we should keep the warning on in general, add an exception, file an issue with typescript, and add a link to the issue in our code.

PR #5012, opened 2019-06-01
[Example] Add confusion matrix example from Jane's paper

kanitw:
Still blocked by https://github.com/vega/vega-lite/issues/5011


domoritz:
Can we pre-aggregate the data? Right now, the spec is pretty large.

Issue #4981, opened 2019-05-19
Extract build config as reusable modules across projects


kanitw:
See https://github.com/apache-superset/build-config for example. 

If we can share build config across Vega-Lite, Vega-Editor, CompassQL, etc. (Start with the TS-driven parts of Vega first.)

domoritz:
Not sure whether it's worth the trouble of figuring out a common set of configs. How do we even keep this in sync?

PR #5037, opened 2019-06-05
Migrate prettier / eslint dependencies to `vega-lite-dev-config`

kanitw:
This is the first step for #4981

We should do the same for TSConfig and part of the Jest config too, but the current tsconfig setup is a bit hairy.


package.json
```diff
@@ -103,10 +98,11 @@
     "terser": "^4.0.0",
     "ts-jest": "^24.0.2",
     "ts-json-schema-generator": "^0.44.0",
-    "typescript": "^3.4.5",
```
domoritz: Why do you fix the ts version?

domoritz:
Beemo isn't used very much and I'm not yet convinced it will stick around. Are there more popular alternatives? How easy is it to roll back these changes?

kanitw:
> Beemo isn't used very much and I'm not yet convinced it will stick around.

It's developed by AirBnB and used a lot there internally and is used by a powerhouse like AirBnB.  Given it's a very small library that does a very simple job (copying / updating dev config), I don't think this will be a concern. (Unlike when we use a react component.) 

> How easy is it to roll back these changes?

As you can see, the actual diff is very small.  

Moving commonly-used plugins to the `dev-config` is also independent of beemo.  


> Are there more popular alternatives? 

The other alternative is having all of this config in a module, but that would be quite annoying for mono-repo.  

Given this is quite a simple task, I think beemo seems fine.  If you find a better tool, you can replace this.  However, given this is already an improvement (that we can share across CompassQL, Vega-Editor, Voyager, vega-lite-to-api), I think it's better to merge it.

package.json
kanitw: Because 3.5 does break the code, so it shouldn't be ^, I guess ~ is fine.

domoritz: Let's do ~ then

domoritz:
LGTM, we should add this to 

* [ ] Vega-Embed
* [ ] Vega tooltip
* [ ] Vega themes
* [ ] Voyager
* [ ] Vega
* [ ] CompassQL
* [ ] Vega Editor
* [ ] Vega-Lite to API

I wonder whether it makes sense to at least add this to another project (e.g. embed and the editor) to make sure the config is general enough.

kanitw:
>I wonder whether it makes sense to at least add this to another project (e.g. embed and the editor) to make sure the config is general enough.

Sounds good.

package.json
```diff
@@ -24,6 +24,7 @@
     "test": "test"
   },
   "scripts": {
+    "prepare": "beemo create-config --silent",
```
domoritz: What is this for?

package.json
```diff
@@ -124,6 +120,13 @@
     "vega-util": "~1.10.0",
     "yargs": "~13.2.4"
   },
+  "beemo": {
```
domoritz: move this below peer dependencies

domoritz:
For the ts-json-schema-generator, I would like to use this as well but I don't want to switch to prettier, yet. I think it would be nice if we can easily switch off prettier. I thought limiting the driver would be sufficient but it is not. 

```
"beemo": {
    "module": "vega-lite-dev-config",
    "drivers": [
      "eslint"
    ]
  },
```

package.json
kanitw: For making beemo init config when you haven't run any of the script.  So you have all the config for vscode right when you run yarn (`prepare` = `postinstall`)

kanitw:
> For the ts-json-schema-generator, I would like to use this as well but I don't want to switch to prettier, yet. I think it would be nice if we can easily switch off prettier. I thought limiting the driver would be sufficient but it is not.


This is beyond the scope of this PR. The problem is the eslint config triggers pretteir.  You can send a PR to make eslint config exclude all prettier in the dev-config repo.

package.json
domoritz: We should actually just remove this since it should be part of the dev config deps.

domoritz:
I think we can merge this. I am starting to use the config in other projects as well. Can you rebase and address the remaining issue?

domoritz:
Closing since this is pretty outdated and we better start over.

PR #5040, opened 2019-06-05
Use the Color type more consistently

domoritz:
Enables better autocomplete in the editor. See https://github.com/microsoft/monaco-json

### DIFF
--- a/src/vega.schema.ts
+++ b/src/vega.schema.ts
@@ -28,6 +28,9 @@ import {SortOrder} from './sort';
 
 export {VgSortField, VgUnionSortField, VgCompare, VgTitle, LayoutAlign, ProjectionType, VgExprRef};
 
+/**
+ * @format color-hex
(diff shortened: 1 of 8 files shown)


kanitw:
LGTM, after making the test pass.

src/vega.schema.ts
```diff
@@ -28,6 +28,9 @@ import {SortOrder} from './sort';
 
 export {VgSortField, VgUnionSortField, VgCompare, VgTitle, LayoutAlign, ProjectionType, VgExprRef};
 
+/**
+ * @format color-hex
```
kanitw: @domoritz color-hex is not a valid format and breaks the editor! https://json-schema.org/understanding-json-schema/reference/string.html

domoritz: I was going off of https://github.com/microsoft/monaco-json

PR #5049, opened 2019-06-07
fix: do not merge unioned domains with fixed data

### LINKED ISSUE #3153: Cannot specify custom domain (with order) for a layered plot
I was trying to create a workaround for https://github.com/vega/vega-lite/issues/3152
and realize that we cannot specify custom domain (with order) for a layered plot yet. 

```
{
  "data": {
    "values": [
      {
        "value": 1,
        "label": "blue"
      },
      {
        "value": 2,
        "label": "green"
      },
      {
        "value": 3,
        "label": "yellow"
      },
      {
        "value": 4,
        "label": "red"
      }
    ]
  },
  "layer": [
    {
      "encoding": {
        "x": {
          "field": "label",
          "scale": {
            "domain": [
              "blue",
              "green",
              "yellow",
              "red"
            ]
          },
          "type": "ordinal"
        },
        "y": {
          "field": "value",
          "scale": {
            "zero": false
          },
          "type": "quantitative"
        },
        "order": {
          "field": "value",
          "type": "quantitative"
        }
      },
      "mark": "line"
    },
    {
      "encoding": {
        "x": {
          "field": "label",
          "type": "ordinal"
        },
        "y": {
          "field": "value",
          "scale": {
            "zero": false
          },
          "type": "quantitative"
        }
      },
      "mark": "point"
    }
  ]
}
```

The fact that the output Vega x-scale has `sort: true` make it no longer respect the custom order

```
"domain": {
  "fields": [
     ["blue","green","yellow","red"],
     {"data": "data_2","field": "label"}
  ],
  "sort": true
}
```

domoritz:
We don't track whether the domain or sort is implicit or explicit right now so we don't have a better way to union domains. 

Going forward, we can either implement the explicit/implicit split and have special rules for sorting by value and by domain and prefer to sort by domain. These two things are orthogonal.  

Part of #3153

### DIFF
--- a/src/compile/scale/domain.ts
+++ b/src/compile/scale/domain.ts
@@ -1,5 +1,5 @@
 import {AggregateOp} from 'vega';
-import {isObject, isString} from 'vega-util';
+import {isArray, isObject, isString} from 'vega-util';
 import {SHARED_DOMAIN_OP_INDEX} from '../../aggregate';
 import {isBinning} from '../../bin';
 import {isScaleChannel, ScaleChannel} from '../../channel';
@@ -433,7 +433,7 @@ export function mergeDomains(domains: VgNonUnionDomain[]): VgDomain {
     util.hash
   );
 
-  const sorts: VgSortField[] = util.unique(
+  const sorts: (VgSortField | 'fixed')[] = util.unique(
     domains
       .map(d => {
         if (isDataRefDomain(d)) {
@@ -449,6 +449,8 @@ export function mergeDomains(domains: VgNonUnionDomain[]): VgDomain {
             }
           }
           return s;
+        } else if (isArray(d)) {
+          return 'fixed'; // if the domain is an array, it defines a sort order
         }
         return undefined;
       })
@@ -466,9 +468,12 @@ export function mergeDomains(domains: VgNonUnionDomain[]): VgDomain {
         log.warn(log.message.MORE_THAN_ONE_SORT);
         sort = true;
       }
+      if (sort === 'fixed') {
+        sort = false;
+      }
       return {
         ...domain,
-        sort
+        ...(sort ? {sort} : {})
       };
     }
     return domain;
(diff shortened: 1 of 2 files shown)


examples/compiled/interactive_brush.vg.json
```diff
@@ -7,7 +7,19 @@
   "height": 200,
   "style": "cell",
   "data": [
-    {"name": "brush_store"},
+    {
```
kanitw: Why does this change selection?

kanitw:
The code looks reasonable to me, but it would be nice to add tests.

kanitw:
Test is failing though.

examples/compiled/interactive_brush.vg.json
domoritz: Because the last PR broke master.

domoritz: I just fixed it so we can rebase.

kanitw:
I think the solution is to consider `sort` as one of the component (or part of the scale component) that gets respected when we later derive the CalculateNode (for sort index).

domoritz:
Yes, but now we get a warning. At least Vega-Lite is behaving how it says it is.

kanitw:
Then it's a not a fix then?

domoritz:
We used to not show a warning when we dropped the sort order. Now we do.

kanitw:
Yeah, but look at #3153, we're supposed to use the specified domain as the domain.  

I'm not sure why we wanna drop the sort order at all.  (We shouldn't merge, but then treat the domain as "explicit" for the merging.)

domoritz:
Yes, but we don't distinguish explicit and implicit right now, no?

kanitw:
> Yes, but we don't distinguish explicit and implicit right now, no?

Ah yes, but to fix the issue, you _may_ need to do so. In any case, to say that this PR fixes #3153 is inaccurate.

domoritz:
Updated to say part of instead of fixes

domoritz:
Replaced by https://github.com/vega/vega-lite/pull/5164

PR #5053, opened 2019-06-08
fix: remove format = color-hex as it's not a a valid format

kanitw:
It's breaking the monaco in the editor! https://json-schema.org/understanding-json-schema/reference/string.html

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -1993,7 +1993,6 @@
       "type": "object"
     },
     "Color": {
-      "format": "color-hex",
       "type": "string"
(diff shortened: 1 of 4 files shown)


domoritz:
Weird, monaco-json specifically mentions that they will give you a nice color picker if you set the format: https://github.com/microsoft/monaco-json

domoritz:
Do you mean the Vega Editor? It works for me.

kanitw:
image


Actually it's breaking ajv in the editor.

domoritz:
I already pushed the fix for that.

PR #5058, opened 2019-06-09
docs: simplify sin and cos example with fold

domoritz:


domoritz:
Actually, let me clean the docs up as well.

Issue #1274, opened 2016-04-01
Support layer as a channel in repeat with color encoding


kanitw:
We haven't implemented this but this seems useful for statistical plot.  

From a conversation this morning, @jheer has example statistical plot this would benefit from this.  

``` js
{
  "layers": [
{
  "data": {"url": "data/cars.json"},
  "mark": "point",
  "select": {
    "sub": {"type": "interval"}
  },
  "encoding": {
    "x": {"field": "Horsepower", "type": "quantitative"},
    "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
    "size": {"field": "Acceleration", "type": "quantitative"}
  }
},
{
  "data": {"url": "data/cars.json"},
  "transform": {
    "filterWith": "sub"
  },
  "mark": "rule",
  "encoding": {
    "y": {"field": "Miles_per_Gallon",  "type": "quantitative",
          "aggregate": "mean"},
    "color": {"value": "red"}
  }
},
{
  "data": {"url": "data/cars.json"},
  "transform": {
    "filterWith": "sub"
  },
  "mark": "rule",
  "encoding": {
    "y": {"field": "Miles_per_Gallon",  "type": "quantitative",
          "aggregate": "median"},
    "color": {"value": "green"}
  }
},
{
  "data": {"url": "data/cars.json"},
  "transform": {
    "filterWith": "sub"
  },
  "mark": "rule",
  "encoding": {
    "y": {"field": "Miles_per_Gallon",  "type": "quantitative",
          "aggregate": "q1"},
    "color": {"value": "black"}
  }
},
{
  "data": {"url": "data/cars.json"},
  "transform": {
    "filterWith": "sub"
  },
  "mark": "rule",
  "encoding": {
    "y": {"field": "Miles_per_Gallon",  "type": "quantitative",
          "aggregate": "q3"},
    "color": {"value": "black"}
  }
}
 ],
 "config": {"cell": {"width":500, "height":500}}
}

```

kanitw:
Another more compelling use case would be to create multiple series line chart of different fields without the need to unpivot the data into a long format.

For example, given a table from https://github.com/altair-viz/altair/issues/292,


![](https://cloud.githubusercontent.com/assets/7890526/22400015/dcd07518-e56f-11e6-8bc8-2278253bb4dc.png)


one should be able to create a spec like

```js
{
   repeat: {
     layer: ['Total', 'Tilt', 'Timing'] 
   },
   spec: {
[...]

PR #5059, opened 2019-06-09
docs: simplify sin and cos example with fold, use a simpler example in the docs

domoritz:


examples/specs/sequence_line_fold.json
```diff
@@ -0,0 +1,43 @@
...
+    {
+      "calculate": "sin(datum.x)",
+      "as": "sin(x)"
+    },
+    {
+      "calculate": "cos(datum.x)",
+      "as": "cos(x)"
+    },
+    {
+      "fold": ["sin(x)", "cos(x)"]
```
kanitw: It's worth noting that folding means we can't produce tooltip for both lines at the same time.  We should think about repeat layer with scale for each layer so we can make a legend.  
(Just noting, I think this PR is fine.)

domoritz: I'm okay with that limitation.

kanitw: FWIW, I can tell that in practice that is too limiting for adoption. 
People wanna compare different lines all the time. Let's discuss more in https://github.com/vega/vega-lite/issues/1274.

Issue #1631, opened 2016-10-29
Consider if we will support width and height channel for ranged mark


kanitw:
Currently we only support `x` and `x2` channels.

g3o2:
As is, adding `height` and ` width` makes vega-lite more flexible without breaking its interface. Naturally, vega-lite should at least throw a warning when both styles are used simultaneously.

Another consideration is whether vega-lite should further abstract this. However, usage of any style highly depends on the data at hand and at the flexibility of data transforms.

kanitw:
Sure, but we have `calculate` object which means we don't really need to add `width` and `height` channel, but it might be convenient for users to add (at the expense of quite more complicated internal code).

g3o2:
`Calculate` works fine indeed if the data provided are in a "tidy" shape, the argument of convenience being applicable to other data shapes.

kanitw:
Adding `xc` / `yc` may also be useful for violin plot too.

kanitw:
I think we should consider NOT adding xc/yc, but instead use `align` and `baseline` as a way to mapping x/y channels to xc/yc instead:

1) Marks like image marks already have align/baseline.  Introducing xc/yc would be redundant and unnecessarily expand the language surface.  It's just more consistent to apply align/baseline to other ranged marks like rect/rule/area. 

I also feel like it's confusing that x and xc are also the same thing for non-ranged marks.  Plus, xc is basically x that's center aligned.  With `align` / `baseline` as the modifier, I think this better fit people's mental model (or at least mine).  

2) Adding xc/yc means users can choose either x/y or xc/yc. Thus, our internal rules (e.g., for looking up scales and axes) have to check both x/y and xc/yc everywhere and will make the codebase very complicated. (That said, this is a minor point and we can only choose do this for our internal implementation and make xc/yc a macro of x/y with align/baseline.)

cc: @jheer @domoritz

kanitw:
We also need to decide whether they should have their own scales or simply use x/y-scales. 

All currently examples in Vega simply use width / height with x/y band scales.  But we don't need explicit width/height channels in Vega-Lite for that as we automatically produce width/height for x/y band scales.   


- x is discrete, width is continuous -- consider a horizontal violin plot, we would need a separate width scale that should be linear. 

- x is discrete, width is discrete -- not sure when we should need this? Maybe for mapping an ordinal field to width?  If so, also separate scale?


[...]

PR #5060, opened 2019-06-09
feat: add image mark 

kanitw:
~~Merge https://github.com/vega/vega-lite/pull/5130 first~~

~~Currently still blocked as we don't have width/height channels yet (#1631)~~


domoritz:
Can we use x and x2 instead of width? Do you have an example?

kanitw:
Yes, if you always want users to calculate x2 = x + width in all specs.

But that’s terrible.

kanitw:
See the example in the changelist

domoritz:
The example uses width and hieght in the mark def. That's different from adding channels, no?

kanitw:
> The example uses width and hieght in the mark def. That's different from adding channels, no?

Fundamentally, we need to rewrite the logic for ranged position (when to include x2 or width in the output). Once we have that, then we can easily get either mark property or channels. So I'd rather have add width as both mark properties and channels in the same PR.  (Same for height.)

kanitw:
Ok, thinking more we can add width/height channel later, but I added width/height as mark properties in #5066.  So let's merge this after #5066.

kanitw:
Oh still need to add docs.

domoritz:
This looks a bit sad

Screen Shot 2019-06-10 at 22 00 37

examples/compiled/scatter_image.vg.json
```diff
@@ -34,8 +34,8 @@
       "from": {"data": "data_0"},
       "encode": {
         "update": {
-          "width": 50,
-          "height": 50,
+          "width": {"value": 50},
```
kanitw: Hmm, I now wonder if there a bug in the SVG preview / generator given that this Vega spec works properly fine in the vega editor.

Still, given we currently output xc/yc, this is still weird.

kanitw: With this PR and #5132, the editor is producing correct output:

image

The SVG generator is probably buggy.

domoritz: > The SVG generator is probably buggy.

Maybe it's just because the paths are not correct in the folder where we generate the images.

domoritz: Without #5132 images work but are behind the axes. Is that expected? 

image

kanitw: Yes, expected, though maybe not desireable. (Need a separate issue to discuss this)

PR #5064, opened 2019-06-10
fix: support formatting of discrete headers.

### LINKED ISSUE #5063: Implement formatType for headers
See https://github.com/vega/vega-lite/issues/1763. 

We still don't support formatting of headers unless the type of the encoding is set to `quantitative`.

```json
{
  "data": {
    "values": [
      {"a": 12.4,"b": 28}, {"a": 32.2,"b": 55}
    ]
  },
  "mark": "bar",
  "encoding": {
    "column": {
      "field": "a",
      "type": "ordinal",
      "header": {
        "format": ".2f",
        "formatType": "number"
      }
    },
  }
}
```

domoritz:
Fixes #5063.

This means that we will automatically treat ordinal and nominal as numbers when a format is present. I think that's a reasonable thing to do.

### DIFF
--- a/test/compile/common.test.ts
+++ b/test/compile/common.test.ts
@@ -131,7 +131,7 @@ describe('Common', () => {
   });
 
   describe('formatSignalRef()', () => {
-    it('should format headers that are ordinal', () => {
+    it('should format ordinal field defs if format is present', () => {
       expect(formatSignalRef({field: 'foo', type: 'ordinal'}, '.2f', 'parent', {})).toEqual({
         signal: 'format(parent["foo"], ".2f")'
       });
--- a/test/compile/common.test.ts
+++ b/test/compile/common.test.ts
@@ -1,5 +1,5 @@
 import {vgField} from '../../src/channeldef';
-import {mergeTitle, numberFormat, timeFormatExpression} from '../../src/compile/common';
+import {mergeTitle, numberFormat, timeFormatExpression, formatSignalRef} from '../../src/compile/common';
 import {defaultConfig} from '../../src/config';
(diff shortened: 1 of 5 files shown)


test/compile/mark/mixins.test.ts
```diff
@@ -321,7 +321,7 @@ describe('compile/mark/mixins', () => {
       const model = parseUnitModelWithScaleAndLayoutSize({
         mark: {type: 'point'},
         encoding: {
-          color: {field: 'Foobar', type: 'nominal', legend: {title: 'baz', format: 's'}}
+          color: {field: 'Foobar', type: 'nominal', legend: {title: 'baz'}}
```
kanitw: Why does this have to be removed?

test/compile/common.test.ts
```diff
@@ -129,4 +129,12 @@ describe('Common', () => {
       expect(mergeTitle('title1', 'title2')).toBe('title1, title2');
     });
   });
+
+  describe('formatSignalRef()', () => {
+    it('should format headers that are ordinal', () => {
```
kanitw: (For the test description) is this only header?  seems like it's for any general field def.

test/compile/mark/mixins.test.ts
domoritz: Because otherwise we will format the field Foobar and the test will change. I have covered this case in the new test. I'm happy either way but this seemed to make more sense.

test/compile/common.test.ts
domoritz: Good point. I will update the description.

Issue #3442, opened 2018-03-03
Design mark for violin plot


kanitw:
From https://vega.github.io/vega/examples/violin-plot/

> A [violin plot](https://en.wikipedia.org/wiki/Violin_plot) visualizes a distribution of quantitative values as a continuous approximation of the probability density function, computed using kernel density estimation (KDE). The densities are additionally annotated with the median value and interquartile range, shown as black lines. Violin plots can be more informative than classical box plots.

- [ ] Understand https://vega.github.io/vega/examples/violin-plot/ example throughly, search online to understand other violin variants, and define the scope that we want to support. 

- [ ] Design `density` transform in Vega-Lite

- Design composite mark syntax
  - First we can focus on just the violin area part 
    - [ ] Design MarkDefinition block for Violin so that we can define property of the underlying `density` transform and other related properties
  - [ ] Second, we need to decide if we want to include interquartile range and median as a part of the violin mark (which is sort of like the "box" overlay on top of violin plot).  The syntax here should be very consistent with [box-plot](https://github.com/vega/vega-lite/blob/master/src/compositemark/boxplot.ts).

kanitw:
The tricky part about this is that Vega's `Violin plot` depends on the Vega facet operator to split data into subgroups between passing it to density transform.  (Density happens inside nested facet.) 

1) Consider the solution above that suggests implementing density transform first.  
Given VL's facet also always applies layout, we can't reproduce the violin example **with axis** using  implement density as a transform unless we do _one of_ the following:

  a) Make Vega density supports `groupby` (which is basically in place faceting)
  b) Support a variant of facet without layout  (pure `facet` in the data transformation sense)

__Note:__ we can reproduce violin plot using  VL `facet` operator, but we will then rely on row instead of y position for each violin.  

2) Alternatively, we could consider implementing violin as its own special mark that produce underlying density transform.  However, this approach will be less composable.  (For example, density plots https://vega.github.io/vega/examples/probability-density/ shouldn't be its own mark but rather using area plotting output from density transforms.)

kanitw:
We meet today to talk about this and conclude that we should make Vega density supports groupby.

HarvsG:
> We meet today to talk about this and conclude that we should make Vega density transform supports groupby.

@kanitw Any progress on implementation?

romainmartinez:
Thanks for Vega-lite.

I often use violin plots and I am looking forward to use them in Vega-lite/Altair.

[...]

PR #5066, opened 2019-06-10
feat: add `width` and `height` as mark properties

kanitw:


kanitw:
Question is should x+width be xc+width by default?  It kinda make more sense from visual encoding perspective that we map marks (e.g.,) to its center?

(I'll think about it later when I have time.)

src/compile/mark/mixins.ts
```diff
@@ -551,7 +551,7 @@ function position2orSize(channel: 'x2' | 'y2', markDef: MarkConfig) {
   if (markDef[channel]) {
     return {[channel]: ref.vgValueRef(channel, markDef[channel])};
   } else if (markDef[sizeChannel]) {
-    return {[sizeChannel]: markDef[sizeChannel]};
+    return {[sizeChannel]: {value: markDef[sizeChannel]}};
```
domoritz: I'm glad I asked you to write a test :-P

domoritz:
I think xc+width makes a bit more sense because in the other case (x+width) you could easily use x and x2 but this is just a gut feeling.

jheer:
Sanity check question: If we use `xc` by default but I then want to treat `x` and `width` as `[x1, x1 + width]`, is there still a way to indicate that?

src/compile/mark/mixins.ts
kanitw: lol the editor broken the other day and I was rushing to sleep.  This is also the cause of bug in the depending image PR.

kanitw:
Note that one big question is whether we should introduce xc as a channel (same for yc) at all.  (See https://github.com/vega/vega-lite/issues/1631#issuecomment-500649594).  I'm leaning towards no.  


Assuming that we don't, here are the answers to your questions: 

> Sanity check question: If we use xc by default but I then want to treat x and width as [x1, x1 + width], is there still a way to indicate that?

Suppose we don't add xc/yc as a channel at all, we can use align/baseline like image mark in Vega (but consistently apply the property to all ranged marks: bar, rect, area, rule, image).  

- VL: x+width+align=center  =>  xc + width in Vega
- VL: x+width+align=left  =>  x + width in Vega
- VL: x+width+align=right  =>  x2 + width in Vega


> Question is should x+width in VL be xc+width in VG by default? It kinda make more sense from visual encoding perspective that we map marks (e.g., image / rect) to its center?


Thinking more, how to apply x+width in VL to x/xc+width in Vega also depend on scale types and mark types:

- For discrete scales, x clearly should always be xc for all the cases I can think of. 
  - For example, [a bar with specified x+width](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoBmOgCtY3AHaQANOCgATZAgBOjQnh4qckAIJhYjVITHIw7dLrBIneMAHdmiMGnbJNbU0wTXQ8ejUNSFDw7DBQCAhICnQxVh04gG0NRPjITFMzNUh2OIAmAA4AX1U8gqgAIWLSnABWVpq6uMgAYWa4gBZpGpzE4HzugBF+nABOAEZO8frIAFEZsArF2uXugDEN1uH1XIhd0wBxDfnZpYnTAAkNioB2O5WASUOyqtGAXQ0IySGF0AGs4uM8ABPQjIbqOXRqKBeTQsOIdE5QZBKeDcTSMJTUCGjSAADwhkAAZoxkGJNN1MLVINDYd1uLp8UpUpAgYlIFCKdTafTTKUmSy4aYAI6sdBKIzhIxUHmAkC-IA), we already make x in VL = xc in Vega. 
  - Rect with specified x + width should behave similarly, but we haven't implemented this case consistently yet.  Same for image (which is in a pending PR).
  - For a vertical violin plot, x should be xc by default too, so we can make violin plot by default, but still can use align (either as encoding or by layering) to create a [split violin plot](https://github.com/vega/vega-lite/issues/3442#issuecomment-497481504). 


- For quantitative scales, let's consider by mark type. 
  - Image: For a scatterplot with image mark use case, xc makes more sense.  But for simple image layout, this might not be sufficient. 
  - Bar: 
    - For a Gantt chart with bar chart where width represents duration, then it is better to apply x as x_start.  
    - [Funnel charts with bar](
image) may wanna use center, but that's a bit unconventional case, I think.  
  - Area/Rule: The only rare case that would involve x+width I can think of is using area/rule for error band/bar with already calculated "width/height" (though we usually get error = width/2). In such case, center makes more sense. 

So if you agree with the analysis above, it seems like we wanna apply x+width as xc+width in most cases except the bar.  If you agree, do we make an exception for bar, or we make the rule consistent for all of them? 

Note that I ignore cases where it makes more sense to use x and x2 as a natural input.

domoritz:
👍 for not introducing xc/yc as channels and instead using align.

domoritz:
If we introduce width and height, do you think people might get confused about how they can specify a bar chart?

Are these two specs equivalent then? 

```json
{
  "data": ...,
  "mark": "bar",
  "encoding": {
  }
}
```

```json
{
  "data": ...,
  "mark": "bar",
  "encoding": {
    "height": {"field": "b", "type": "quantitative"}
  }
}
```

kanitw:
Yes, the two should be equivalent.  (Though that's more for #1631 since we're not adding the channel in this PR.)

domoritz:
I guess so is

```json
{
  "data": ...,
  "mark": "bar",
  "encoding": {
    "y": {"value": 0, "type": "quantitative"},
    "y2": {"field": "b", "type": "quantitative"}
  }
}
```

In a way `y2` and `height` behave similarly in that they augment the `x` encoding. Using `y2` means you get a mark from `[y,y2]` whereas using `height` means you get `[y,y+height]` or `[y-height/2,y+height/2]` depending on the scale and mark types as you explained above.

kanitw:
I'll deal with default and align/baseline in separate PRs.

PR #5072, opened 2019-06-12
fix: rename example file

ganeshpatro321:
This PR renames the vega-lite json file.


ganeshpatro321:
@domoritz 

This PR just renames the file which was causing an error during the setup.

domoritz:
Please clean up the pr message and fix the title to follow semantic commit titles. 

Thanks

kanitw:
Thanks. (Minor comment, let's not use past tense for commit message.  A common convention is that they should read like this commit "does X".)

PR #5084, opened 2019-06-16
feat!: replace `config.invalidValues` with `mark.invalid` for mark definition and config

### LINKED ISSUE #4802: Rename config.invalidValues to config.mark.invalid
I just finally realized that this is a config for how to deal with marks for invalid values.

kanitw:
This would allow us to customize how we handle null values differently for different marks

Fix #4802

BREAKING CHANGE: `config.invalidValues` no longer works

### DIFF
--- a/test/transformextract.test.ts
+++ b/test/transformextract.test.ts
@@ -116,8 +116,8 @@ describe('extractTransforms()', () => {
 
         const spec = JSON.parse(specString);
 
-        // Use invalidValues =  "hide" so we don't include filterInvalid in the comparison
-        const config = initConfig({...spec.config, invalidValues: 'hide'});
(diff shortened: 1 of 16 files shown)


test/transformextract.test.ts
```diff
@@ -116,8 +116,8 @@ describe('extractTransforms()', () => {
 
         const spec = JSON.parse(specString);
 
-        // Use invalidValues =  "hide" so we don't include filterInvalid in the comparison
-        const config = initConfig({...spec.config, invalidValues: 'hide'});
+        // Use invalid =  "hide" so we don't include filterInvalid in the comparison
```
domoritz:
```suggestion
        // Use invalid = "hide" so we don't include filterInvalid in the comparison
```

PR #5088, opened 2019-06-16
refactor: don't include unnecessary "descending" in sort output

kanitw:


domoritz:
I think you meant unnecessary

PR #5096, opened 2019-06-18
feat: make timeunit for bar/rect behaves more like bin 

### LINKED ISSUE #4868: Improve timeUnit positioning
(Copied from slack) 

@jheer: 

> I have a temporal domain with timeUnit `yearmonth`. I want to show a histogram with bars for each month. If I use a `bar` mark this works but the bars have a fixed size. I want them to scale to fit the known timeUnit, just like numerical bins do. Do we have any support for this? (Note that I don’t want to switch to an `ordinal` domain, as then empty/missing time units get dropped.)

@kanitw: 

> The way to do this in Vega is probably offset the `x2` by the smallest unit (a month for `yearmonth`)
> In fact,  I think this should even be the default behavior.
> The tricky part for this is that months have uneven number of days

@jheer 
> Yeah, I think the solution is to treat time units as bins. If no time unit is given (just type `T`) I think the current behavior is reasonable — it’s just like plotting bars over a `Q` domain. For a given time unit, we could calculate start/stop boundaries that are specific to the time unit, just like binning.

> Happy to look at Vega modifications to achieve this. I also added a Vega issue to perform imputation for time units (https://github.com/vega/vega/issues/1764), so that ordinal scales over proper time domains could also work.

kanitw:
Fix #4868 

In this PR, we only make timeUnit like binning for bar and rect. 
We do not center other marks (e.g., point/line/area) in the middle of the timeUnit band by default for a few reasons:

1) Area will be ugly, leaving empty spaces on the side

2) Centering all marks so will cause breaking changes to all plots with timeUnit. 
By only making timeUnit behave like bin for just bars and rects, we only change the behaviors for plots that currently look terrible. 

3) Doing so will cause the plots with and without timeUnit look quite different.
For example, if there is a date field with only year information.  Using no timeUnit and timeUnit=year will produce dramatically different results. 

When needed, users can easily set the `band` or mark config's `timeUnitBand` to `0.5` to center the data point in the middle of a time unit band.  See `line_month_center_band` for example. 

FWIW, Tableau also does NOT center the data point in the middle of a time unit band by default. 

- [ ] ~~default alignment~~ -- from an experiment the current one is already good

- [x] distinguish between band's width and position
  - pointBandSize, bandSize, bandPosition?

- [x] stack with impute case 

- [x] document band

- [ ] revise timeUnit docs

- [ ] ~~make band work for bar with temporal timeUnit~~ will do in https://github.com/vega/vega-lite/issues/5243

Note that we don't include support for a field that has already been "prebinned" with a time unit transform.  Let's handle that later in https://github.com/vega/vega-lite/issues/5242.


src/channeldef.ts
```diff
@@ -361,16 +363,50 @@ export interface PositionFieldDef<F extends Field>
   impute?: ImputeParams;
 
   /**
-   * For rect-based marks (`rect`, `bar`, and `image`), mark size relative to bandwidth of [band scales](https://vega.github.io/vega-lite/docs/scale.html#band). If set to `1`, the mark size is set to the bandwidth. If set to `0.5`, the mark size is half of the bandwidth.
+   * For rect-based marks (`rect`, `bar`, and `image`), mark size relative to bandwidth of [band scales](https://vega.github.io/vega-lite/docs/scale.html#band) or time units. If set to `1`, the mark size is set to the bandwidth or the time unit interval. If set to `0.5`, the mark size is half of the bandwidth or the time unit interval.
```
kanitw: rect/bar can be band position too, if size property is explicitly specified.

willium:
Can we center point for contour plots somehow?

kanitw:
> Can we center point for contour plots somehow?

Not sure why is that related to this PR, given we don't even support contour plot yet.

src/compile/mark/rect.ts
```diff
@@ -55,15 +55,16 @@ export function rectPosition(model: UnitModel, channel: 'x' | 'y', mark: 'bar' |
   // x, x2, and width -- we must specify two of these in all conditions
   if (
     isFieldDef(fieldDef) &&
-    (isBinning(fieldDef.bin) || isBinned(fieldDef.bin)) &&
+    // FIXME use band to determine condition here
```
kanitw: FYI, this will be fixed later in https://github.com/vega/vega-lite/pull/5244

willium:
> > Can we center point for contour plots somehow?
> 
> Not sure why is that related to this PR, given we don't even support contour plot yet. If you have specific feature request, file a separate issue?

https://github.com/vega/vega-lite/issues/4868#issuecomment-513629468

src/mark.ts
```diff
@@ -135,6 +139,11 @@ export interface MarkConfig extends ColorMixins, BaseMarkConfig {
    * - If `null`, all data items are included. In this case, invalid values will be interpreted as zeroes.
    */
   invalid?: 'filter' | Hide | null;
+
+  /**
+   * Default band position for a time unit.
+   */
+  timeUnitBandPosition?: 0 | 0.5;
```
domoritz: Why can this only be two values? Please document.

examples/compiled/bar_yearmonth_custom_format.vg.json
```diff
@@ -17,9 +17,14 @@
           "as": "yearmonth_date",
```
domoritz: Looking at the chart, it's really hard to know what the values for axis ticks are. Please revise this example. 

Screen Shot 2019-07-29 at 9 30 21 PM

kanitw:
> Should we remove the grid so that it looks like in bins?

No. Grid in time actually usually represent year / month boundaries, which are helpful to readers.  (Grid in bins do not mean much.)  

In a follow up PR, I'll even make the default band different between bin and timeUnit.

examples/compiled/bar_yearmonth_custom_format.vg.json
kanitw: This PR doesn't change anything about the label.  

> it's really hard to know what the values for axis ticks are

This example is only there for explaining that you can customize the format. (See `format.md`.)  So if you strongly think this should be change, please submit a separate PR for it.

src/mark.ts
kanitw: Technically it doesn't have to be either 0 or 0.5.  But in practice, these numbers won't make sense if they are not 0 (beginning of the timeUnit interval or 0.5 (middle of the timeUnit interval). 

I can relax it to number.

hoclun-rigsep:
New user, trying to follow the discussion here. With a known domain resolution of e.g. yearmonth, wouldn't you want the tick mark of the x-axis to be centered under the bar?

PR #5104, opened 2019-06-23
refactor: Narrow type for interval init to two-element array

domoritz:


arvind:
Nice! Thanks @domoritz!

PR #5108, opened 2019-06-23
fix: Merge output nodes. Fixes #5105.

### LINKED ISSUE #5105: Merge output nodes
If a dataflow node has output nodes as children, we should merge them. In the example below, this would enable us to also merge the aggregate nodes. 

Screen Shot 2019-06-22 at 18 10 49

domoritz:


domoritz:
Changed the title.

PR #5117, opened 2019-06-26
fix!: only drop color encoding/mark property when there is a conflicting fill/stroke

### LINKED ISSUE #4326: Revise color precedence

kanitw:
Fix #4326

Why?

1) With current behavior, it's quite hard to add stroke to highlight selection like in this example (as fill gets dropped automatically):

```
{
  "data": {
    "values": [
    ]
  },
  "selection": {"select": {"type": "multi", "empty": "none"}},
  "mark": "bar",
  "encoding": {
    "stroke": {"condition": {"selection": "select", "value": "black"}}
  }
}
```

2) Dropping color when adding fill/stroke means we have less smooth transition between different specifications. 

----

It seems like doing so barely changes existing examples.  ~~The only problematic case is graticule, which means we may consider tracking if a plot is deriving data from a graticule.~~


domoritz:
> The only problematic case is graticule

Can you explain why it's a problematic case?

kanitw:
> The only problematic case is graticule

Because geoshape is filled by default, but in our previous graticule examples, we have to manually set stroke, which will in turn disables default fill. However, now it no longer disables default fill, we get a filled graticule:

image


That said, the right fix is to change the behavior for geoshape with graticule data source to be "unfilled" by default, which will fix this problem and eliminate the need for explicit "stroke" in all graticule geoshapes.  The fix is in #5118 (a follow-up PR).

PR #5118, opened 2019-06-27
 fix: make geoshape with graticule data source unfilled (stroked) by default

kanitw:
Merge #5117 first. 

Note that this is a separate fix that solves the root cause of the problem in #5177. 

You can see that now graticule + geoshape works out of the box without the need for explicit "stroke".


src/encoding.ts
```diff
@@ -399,7 +401,9 @@ export function normalizeEncoding(encoding: Encoding<string>, mark: Mark): Encod
     }
 
     // Drop color if either fill or stroke is specified
-    if (channel === 'color' && ('fill' in encoding || 'stroke' in encoding)) {
+
+    // FIXME
```
domoritz: Why is this fixme?

PR #5132, opened 2019-06-29
 fix: use `x` / `xc` / `x2` based on `align` when width is specified

kanitw:
Merge #5060 first


domoritz:
Replaced by https://github.com/vega/vega-lite/pull/5149

PR #5135, opened 2019-06-29
docs: revise & refactor selection docs structure

### LINKED ISSUE #4841: Revise / Clarify if there are differences between selection properties and selection transforms
As I look through https://github.com/vega/vega-lite/pull/4720, I realize that we don't have a clear distinction (at least in the docs) between what is a selection property and selection transform.  (Should `clear` be a property or a transform?) 

From the page here is the explanation for selection property: 

> While selection types provide useful defaults, it can often be useful to override these properties to customize the interaction design. 

and selection transform: 

> Vega-Lite provides a number of selection transformations to further customize the behaviour of a selection. These include:

However, it is unclear what is the difference between them. (Both of them are for customizing the selection.  The only distinction I can see is that selection transform are all disabled by default.) 

What makes `toggle` or `fields` / `encodings` projection a transform and `on` not a transform? Basically, we should clarify this among ourselves and make it clear on the docs or revise to consolidate all of them to be all selection properties.

Unless we have a clear distinction, I'm leaning towards suggesting the latter as it will reduce another jargon that users will have to learn.  

Plus, we can do format the document to have:

- (1) General Selection Properties
- (2-4) Selection Properties specific for Single/Multi/Interval transforms. 

which should be a bit simpler for users to understand.  

Let me know what you think

kanitw:
- consolidate selection properties and transforms (as there is no clear distinction between them)
- extract details for certain properies into their own pages
- provide consistent side menu

fix #4841

Note that we still miss pages for on / empty, but that's beyond the scope of this PR, so I'm filing #5136.


kanitw:
@arvind  -- Given there are a lot of pending PRs, I'll merge this first.  Feel free to send a follow-up issue for commenting / PR for revision.

PR #5144, opened 2019-07-04
docs: Add Lens.org to "Tools that use Vega-Lite" section.

simon-lang:


domoritz:
I updated the PR a bit. Thanks for sending it.

PR #5146, opened 2019-07-04
fix: correct comments. Fixes #5141

### LINKED ISSUE #5141: Doc issues in the 3.3.0 schema
I noticed several documentation issues in the 3.3.0 version of the vega-lite schema. Most looked like cut-n-paste issues. Unfortunately I didn't note them all down, and I wasn't sure whether there should be separate issues filed. If I see any more I can add them here (or is it okay to submit a PR for simple changes like the `labelLimit` case below?)

- `zindex` is confusing - I assume that `"non-positive integer"` should be something like "integer, 0 or greater" or "an integer, 0 or 1".

```
"zindex": {
          "description": "A non-positive integer indicating z-index of the axis.\nIf zindex is 0, axes should be drawn behind all chart elements.\nTo put them in front, use `\"zindex = 1\"`.\n\n__Default value:__ `1` (in front of the marks) for actual axis and `0` (behind the marks) for grids.",
          "minimum": 0,
          "type": "number"
        }
```

- In `Legend` and `LegendConfig` I assume `axis tick labels/titles` should read `legend labels/titles`

```
 "labelLimit": {
          "description": "Maximum allowed pixel width of axis tick labels.\n\n__Default value:__ `160`.",
          "type": "number"
        },

"titleLimit": {
          "description": "Maximum allowed pixel width of axis titles.\n\n__Default value:__ `180`.",
          "minimum": 0,
          "type": "number"
        },
```

domoritz:


src/axis.ts
```diff
@@ -106,7 +106,7 @@ export interface Axis extends AxisOrientMixins, BaseAxisNoSignals, Guide {
   values?: number[] | string[] | boolean[] | DateTime[];
 
   /**
-   * A non-positive integer indicating z-index of the axis.
+   * A natural number indicating the z-index of the axis.
```
kanitw: Natural number doesn't have a consistent interpretation whether it includes zero (ISO standard includes 0 but layman's dictionary sometimes d o not), making it ambiguous.  

"non-negative integer" is clearer?

domoritz: Done.

PR #5147, opened 2019-07-04
fix: Don't force signal expressions to return a string.

domoritz:
Fixes vega/vega-tooltip#251

### DIFF
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -79,7 +79,7 @@ export function formatSignalRef(
   specifiedFormat: string,
   expr: 'datum' | 'parent' | 'datum.datum',
   config: Config,
-  coerce = true
+  castToString = true
 ) {
   if (isTimeFormatFieldDef(fieldDef)) {
     const isUTCScale = isScaleFieldDef(fieldDef) && fieldDef['scale'] && fieldDef['scale'].type === ScaleType.UTC;
@@ -109,7 +109,7 @@ export function formatSignalRef(
         signal: `${formatExpr(vgField(fieldDef, {expr, binSuffix: 'range'}), format)}`
       };
     } else {
-      return {signal: (coerce ? `''+` : '') + vgField(fieldDef, {expr})};
+      return {signal: (castToString ? `''+` : '') + vgField(fieldDef, {expr})};
     }
   }
 }
--- a/src/compile/common.ts
+++ b/src/compile/common.ts
@@ -78,7 +78,8 @@ export function formatSignalRef(
   fieldDef: TypedFieldDef<string>,
   specifiedFormat: string,
   expr: 'datum' | 'parent' | 'datum.datum',
-  config: Config
+  config: Config,
+  coerce = true
 ) {
   if (isTimeFormatFieldDef(fieldDef)) {
     const isUTCScale = isScaleFieldDef(fieldDef) && fieldDef['scale'] && fieldDef['scale'].type === ScaleType.UTC;
@@ -108,7 +109,7 @@ export function formatSignalRef(
         signal: `${formatExpr(vgField(fieldDef, {expr, binSuffix: 'range'}), format)}`
       };
     } else {
-      return {signal: `${vgField(fieldDef, {expr})}`};
+      return {signal: (coerce ? `''+` : '') + vgField(fieldDef, {expr})};
     }
   }
 }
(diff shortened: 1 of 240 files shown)


domoritz:
I don't remember why we converted to strings beforehand. This seems like an overall good change to me.

kanitw:
I’d dig into git blame and find an answer first to avoid regression

domoritz:
We did this in https://github.com/vega/vega-lite/pull/2768.

domoritz:
I decided to make coercion configurable and only force strings if the expression is not used in tooltips. This way we get `'null'` in headers and texts but `null` in tooltips and the handler can then decide what to do with it.

domoritz:
NO SVG changes anymore.

kanitw:
This will make tooltip show `null` / `NaN`?  If so, I think this is a wrong move. 

For `undefined`, it's more debatable.  While I can see that hiding them make may sense in some cases, I'm actually not convinced that this is the right default yet as it's causing inconsistencies across different encoding.  For example, if you have undefined in your x-field, your axis would show "undefined" as the value.  For consistency, `undefined` shouldn't show up as hidden for the tooltip of the same plot.


It's probably better to have a config to choose to display undefined or not in all places? Doing so could be tricky though given there are places we just use the field directly.

domoritz:
What this does it leave it up to the handler to decide how to display `null` and `undefined`. 

Vega-Tooltip shows `null` but hides `undefined` properties.

Screen Shot 2019-07-05 at 8 13 53 PM

src/compile/common.ts
```diff
@@ -78,7 +78,8 @@ export function formatSignalRef(
   fieldDef: TypedFieldDef<string>,
   specifiedFormat: string,
   expr: 'datum' | 'parent' | 'datum.datum',
-  config: Config
+  config: Config,
+  coerce = true
```
kanitw: boolean param is bad.   Reading the call like `text(fieldDef, config, expr, false)` gives us no idea what is false for at all. 

Also, coerce is also vague. `castToString`? or something similar?


Like `text(fieldDef, config, expr, {castToString: true/false})` ?

kanitw:
(Still, this may mean that your change in https://github.com/vega/vega-tooltip/commit/f76f0eb48ce2e0c6f931a282daa131aca2bb9817 still leads to inconsistency described above?)

src/compile/common.ts
domoritz: I'm happy to rename the property but expecting an object instead of a boolean is weird. In e.g. `timeFormatExpression`, we also don't use this pattern.

domoritz:
> (Still, this may mean that your change in vega/vega-tooltip@f76f0eb still leads to inconsistency described above?)

What's the inconsistency? We hide undefined but show null and NaN.

src/compile/common.ts
domoritz: I guess we started doing this in a few other places so I will update this.

kanitw:
We display "undefined" in the axis, if the value is undefined.

image

[Spec](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAutIC4yghSA3WAbArgKYDOKYA2uBhOtRpABpkBMAHADRV1QBCZkAIwCMkLhgC+nbrW6QAgv1gip3KE1QAWDStl9Ug5qO7ixAXSqSqkALawATgGsyMqPACeAB0L8PAewCWAHbwRmCWmISBAMa+0EEA5s5ikAAeSbIAZv6EuND8TDr07l78AI74sMH+iPD+WN5i4fRu6XSQWTl5+gqFmMXe+oG+1kF4oRBNmDG4vnat1O3Zufx8va6eA1BDI4FjjYWQvh6wUdUtqC5TvoFxtdfzbfAkIRdiskMvaG+qkMQ5hFE7oF+H9cACQt9qCZVGE1pgcARNgAGAB0AGZvtCJBYVJAYoEsokLk1fv9Af57q8FqDwQ8+ht+NZ8LhapA1otOqRUJQYfI2ZDeOMMKZ2UFqnT6ApuQLMLBDHD6HL+TDZSIBSKBZA9BQZVBhMqVQYDbyBIZ1ZjGhYQOIgA)

domoritz:
Yep, I might actually prefer to not show the blue dot. What do you think?

domoritz:
I see. I think we need to do this 4.0 then where we removed default tooltips. On slack we decided that we want to hide undefined values for default tooltips but not for manually defined tooltips.

domoritz:
It looks like we already have the behavior we want in VL 4.

PR #5149, opened 2019-07-05
fix: use `x` / `xc` / `x2` based on `align` when width is specified

domoritz:
replaces https://github.com/vega/vega-lite/pull/5132/


src/compile/mark/rect.ts
```diff
@@ -17,7 +17,13 @@ export const rect: MarkCompiler = {
   vgMark: 'rect',
   encodeEntry: (model: UnitModel) => {
     return {
-      ...mixins.baseEncodeEntry(model, {color: 'include', size: 'ignore', orient: 'ignore'}),
+      ...mixins.baseEncodeEntry(model, {
+        align: 'ignore',
+        baseline: 'ignore',
+        color: 'ignore',
```
domoritz: this should probably be include.

kanitw:
Why do u need a new pr lol

domoritz:
I tried to merge 4.0 into your branch and everything exploded. I guess I could have force pushed afterwards into your branch but oh well.

PR #5151, opened 2019-07-06
fix!: make axes appear below marks by default

kanitw:
better for `scatter_image`: 

image

better for `point_invalid_color`:

image

worse for tabular heatmap:

image


kanitw:
I think the good cases outweigh the bad cases though.  That said, I'm wondering if we should make this conditional.

kanitw:
I'm experimenting it in https://github.com/vega/vega-lite/pull/5153

PR #5153, opened 2019-07-06
fix: mark discrete axes of rect marks use axis's zindex = 1 by default

kanitw:
~~Merge https://github.com/vega/vega-lite/pull/5151 first.~~


domoritz:
Wouldn't it be cleaner/simpler to make rect marks not overlap with the axes? Note how the rectangle doesn't overlap with x but with y. 

image

kanitw:
> Wouldn't it be cleaner/simpler to make rect marks not overlap with the axes? Note how the rectangle doesn't overlap with x but with y.

Maybe, if you have a good idea how do to it while maintaining width for all rect go ahead. I just don't see how.

domoritz:
I guess it only works if the axes are rendered outside the chart area. 

Do we have to special case x and y axes?

kanitw:
The overlap is happening only for left/top axes (See https://github.com/vega/vega/issues/1867). 

However, putting discrete axes above marks isn't problematic for rect marks, so it might be better to keep the rule a bit more simplified.

PR #5159, opened 2019-07-07
Always parse quantitative as numbers

domoritz:
Alternative to https://github.com/vega/vega-lite/pull/4993. See https://github.com/vega/vega-lite/pull/4993#issuecomment-508992255 for details.


domoritz:
Need to pick up later.

PR #5160, opened 2019-07-07
refactor: optimize calculate nodes.

domoritz:
In this PR, I am adding a new optimizer that (similar to the bin optimizer) merges calculate nodes at forks and moves them up in the dataflow (towards the source). I extended the calculate node to support multiple calculates in a single node.

### DIFF
--- a/examples/compiled/circle_github_punchcard.vg.json
+++ b/examples/compiled/circle_github_punchcard.vg.json
@@ -12,6 +12,11 @@
       "transform": [
         {
           "type": "formula",
+          "expr": "time(datetime(2006, 0, day(datum[\"time\"])+1, 0, 0, 0, 0))===time(datetime(2006, 0, 1+1, 0, 0, 0, 0)) ? 0 : time(datetime(2006, 0, day(datum[\"time\"])+1, 0, 0, 0, 0))===time(datetime(2006, 0, 2+1, 0, 0, 0, 0)) ? 1 : time(datetime(2006, 0, day(datum[\"time\"])+1, 0, 0, 0, 0))===time(datetime(2006, 0, 3+1, 0, 0, 0, 0)) ? 2 : time(datetime(2006, 0, day(datum[\"time\"])+1, 0, 0, 0, 0))===time(datetime(2006, 0, 4+1, 0, 0, 0, 0)) ? 3 : time(datetime(2006, 0, day(datum[\"time\"])+1, 0, 0, 0, 0))===time(datetime(2006, 0, 5+1, 0, 0, 0, 0)) ? 4 : time(datetime(2006, 0, day(datum[\"time\"])+1, 0, 0, 0, 0))===time(datetime(2006, 0, 6+1, 0, 0, 0, 0)) ? 5 : time(datetime(2006, 0, day(datum[\"time\"])+1, 0, 0, 0, 0))===time(datetime(2006, 0, 0+1, 0, 0, 0, 0)) ? 6 : 7",
+          "as": "y_day_time_sort_index"
+        },
+        {
+          "type": "formula",
           "as": "day_time",
           "expr": "datetime(2006, 0, day(datum[\"time\"])+1, 0, 0, 0, 0)"
         },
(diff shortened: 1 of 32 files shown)


examples/compiled/circle_github_punchcard.vg.json
```diff
@@ -10,6 +10,11 @@
       "url": "data/github.csv",
       "format": {"type": "csv", "parse": {"time": "date"}},
       "transform": [
+        {
```
kanitw: This formula shouldn't get reordered?

src/compile/data/calculate.ts
```diff
@@ -53,24 +51,64 @@ export class CalculateNode extends DataFlowNode {
...
+  /**
+   * Merge calculate nodes. This method either integrates the bin config from the other node
+   * or if this node already has a bin config, renames the corresponding signal in the model.
+   */
+  public merge(other: CalculateNode) {
+    for (const key of keys(other.calculates)) {
+      if (key in this.calculates) {
+        // make sure we are not merging something incompatible
+        if (this.calculates[key].calculate !== other.calculates[key].calculate) {
+          /* istanbul ignore next: This should never happen. */
```
kanitw: What if you have two layers that calculate two different things "as" the same variable?  

If it shouldn't happen, why add it in the code?

src/compile/data/optimizers.ts
```diff
@@ -425,6 +403,60 @@ export class MergeBins extends BottomUpOptimizer {
...
 
+/**
+ * Merge calculate nodes and move them up through forks. Stop at filters and parse as we want them to stay before the calculate node.
+ *
+ * Similar to MergeBins but for Calculate Nodes.
+ */
+export class MergeCalculate extends BottomUpOptimizer {
+  public run(node: DataFlowNode): OptimizerFlags {
+    const parent = node.parent;
+    const moveCalcssUp = !(
```
kanitw: `moveCalcsUp`  (Too many `s`'s)

src/compile/data/calculate.ts
```diff
@@ -53,24 +51,64 @@ export class CalculateNode extends DataFlowNode {
...
-      type: 'formula',
-      expr: this.transform.calculate,
-      as: this.transform.as
-    };
+  public assemble(): VgFormulaTransform[] {
+    const transforms: VgFormulaTransform[] = [];
+
+    for (const {key, value} of entries(this.calculates)) {
+      if (key !== value.as) {
+        /* istanbul ignore next: This should never happen. */
```
kanitw: Why checking this if it shouldn't ever happen?  

(If you worry about bug, can you check this in a unit test instead? )

examples/compiled/waterfall_chart.vg.json
```diff
@@ -88,7 +88,7 @@
       "transform": [
         {
           "type": "filter",
-          "expr": "datum[\"previous_sum\"] !== null && !isNaN(datum[\"previous_sum\"])"
+          "expr": "datum[\"sum_inc\"] !== null && !isNaN(datum[\"sum_inc\"])"
```
kanitw: This example's dataflow got swapped like crazy!

examples/compiled/rect_mosaic_labelled_with_offset.vg.json
```diff
@@ -85,11 +85,8 @@
       "source": "source_0",
       "transform": [
         {
-          "type": "aggregate",
-          "groupby": ["Origin"],
-          "ops": ["min"],
-          "fields": ["xc"],
-          "as": ["min_xc"]
+          "type": "filter",
```
kanitw: filter/aggregate shouldn't get swapped here.

src/compile/data/calculate.ts
domoritz: This is an assertion that I added. It won't happen right now but I am worried that at some point I might forget to satisfy the assumptions.

domoritz:
I think the logic is flawed. I should merge calculates like I merge parses, not like I merge bins. For bins we know that the keys are sufficient to differentiate what the bin calcualtes.

PR #5164, opened 2019-07-07
fix: split scale domains into explicit and implicit.

### LINKED ISSUE #3153: Cannot specify custom domain (with order) for a layered plot

domoritz:
fixes #3153

image

### DIFF
--- a/src/compile/scale/component.ts
+++ b/src/compile/scale/component.ts
@@ -9,14 +9,12 @@ import {Explicit, Split} from '../split';
  * All VgDomain property except domain.
  * (We exclude domain as we have a special "domains" array that allow us merge them all at once in assemble.)
  */
-// TODO: also exclude domainRaw and property implement the right scaleComponent for selection domain
+// TODO: also exclude domainRaw and properly implement the right scaleComponent for selection domain
 
-export type ScaleComponentProps = Omit<VgScale, 'domain'>;
+export type ScaleComponentProps = Omit<VgScale, 'domain'> & {domains: VgNonUnionDomain[]};
 
 export class ScaleComponent extends Split<ScaleComponentProps> {
   public merged = false;
-
-  public domains: VgNonUnionDomain[] = [];
 
   constructor(name: string, typeWithExplicit: Explicit<ScaleType>) {
     super(
@@ -29,15 +27,11 @@ export class ScaleComponent extends Split<ScaleComponentProps> {
   /**
    * Whether the scale definitely includes zero in the domain
    */
-  public get domainDefinitelyIncludesZero() {
+  public domainDefinitelyIncludesZero() {
     if (this.get('zero') !== false) {
       return true;
     }
-    const domains = this.domains;
-    if (isArray(domains)) {
-      return some(domains, d => isArray(d) && d.length === 2 && d[0] <= 0 && d[1] >= 0);
-    }
-    return false;
+    return some(this.get('domains'), d => isArray(d) && d.length === 2 && d[0] <= 0 && d[1] >= 0);
   }
 }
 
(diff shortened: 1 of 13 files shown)


src/compile/scale/component.ts
```diff
@@ -29,15 +27,11 @@ export class ScaleComponent extends Split<ScaleComponentProps> {
   /**
    * Whether the scale definitely includes zero in the domain
    */
-  public get domainDefinitelyIncludesZero() {
+  public domainDefinitelyIncludesZero() {
```
kanitw: lol why do we have to make it no longer a getter?

src/compile/scale/component.ts
```diff
@@ -9,15 +9,13 @@ import {Explicit, Split} from '../split';
  * All VgDomain property except domain.
  * (We exclude domain as we have a special "domains" array that allow us merge them all at once in assemble.)
  */
-// TODO: also exclude domainRaw and property implement the right scaleComponent for selection domain
+// TODO: also exclude domainRaw and properly implement the right scaleComponent for selection domain
 
-export type ScaleComponentProps = Omit<VgScale, 'domain'>;
+export type ScaleComponentProps = Omit<VgScale, 'domain'> & {domains: VgNonUnionDomain[]};
```
kanitw: minor: technically, it's a component for scale domain. Given it's a domain of a single scale, then `domain` (singular) is more appropriate?

src/compile/split.ts
```diff
@@ -96,13 +96,10 @@ export function makeImplicit<T>(value: T): Explicit<T> {
   };
 }
 
+export type PropertyOf = 'scale' | 'axis' | 'legend' | '';
```
kanitw: propertyOf as a variable makes sense, but a bit confusing as a type. 

How about `parentProperty: SplitParentProperty` or something similar?

src/compile/scale/component.ts
domoritz: We actually compute something pretty complicated here. Using a getter feels like an antipattern to me.

src/compile/scale/component.ts
domoritz: Before parse, this is not the domain yet but an array of domains that gets merged unioned into one vega domain. The types are also incompatible so it's better to have a different name, imho.

PR #5170, opened 2019-07-08
chore: refactor optimizers

domoritz:


kanitw:
FYI, you can use `refactor:`

PR #5171, opened 2019-07-08
fix: implement isDataSourceNode instead of just checking for SourceNode

domoritz:


src/compile/data/optimizers.ts
```diff
@@ -9,7 +9,7 @@ import {FilterNode} from './filter';
 import {ParseNode} from './formatparse';
 import {JoinAggregateTransformNode} from './joinaggregate';
 import {FACET_SCALE_PREFIX} from './optimize';
-import {BottomUpOptimizer, TopDownOptimizer} from './optimizer';
+import {BottomUpOptimizer, isDataSourceNode, TopDownOptimizer} from './optimizer';
 import * as optimizers from './optimizers';
 import {SourceNode} from './source';
```
invokesus: Unused import 🙃

domoritz: Thanks. Fixed.

domoritz: Fixed. Thanks!

PR #5172, opened 2019-07-08
fix: parse optimizer should keep track of conflicting fields

domoritz:
We had a potential bug where we may ignore a field because it has a conflict but then add it back again.


domoritz:
Add a separate commit with a test that fails before the fix.

PR #5174, opened 2019-07-08
fix: refactor selection initialization for TS 3.5

domoritz:
Needed for https://github.com/vega/vega-lite/pull/5005


src/compile/selection/transforms/project.ts
```diff
@@ -68,7 +68,7 @@ const project: TransformCompiler = {
             if (isSingleDefUnitChannel(key)) {
               (selDef.encodings || (selDef.encodings = [])).push(key as SingleDefUnitChannel);
             } else {
-              if (isIntervalSelection(selDef)) {
+              if (selDef.type === 'interval') {
```
kanitw: This seems weird?

src/compile/selection/transforms/project.ts
```diff
@@ -139,15 +139,20 @@ const project: TransformCompiler = {
       if (scales.has(selCmpt)) {
         log.warn(log.message.NO_INIT_SCALE_BINDINGS);
       } else {
-        const parseInit = <T extends SelectionInitMapping | SelectionInitIntervalMapping>(i: T): T['a'][] => {
+        function parseInit(i: SelectionInitMapping): SelectionInitMapping['a'];
+        function parseInit(i: SelectionInitIntervalMapping): SelectionInitIntervalMapping['a'];
+        // eslint-disable-next-line no-inner-declarations
+        function parseInit(i: SelectionInitMapping | SelectionInitIntervalMapping): any {
```
kanitw: Better remove `SelectionInitIntervalMapping` and make `SelectionInitMapping` a generic with two forms: 

`SelectionInitMapping<Value | DateTime>` and `SelectionInitMapping<SelectionInitInterval>`

src/compile/selection/index.ts
```diff
@@ -29,12 +30,10 @@ export interface SelectionComponent<T extends SelectionType = SelectionType> {
   name: string;
   type: T;
 
-  // Use conditional typing (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html)
-  // so we have stricter type of init (as the type of init depends on selection type)
   init?: (T extends 'interval'
-    ? SelectionInitInterval
+    ? IntervalSelection['init']['a']
```
kanitw: This seems weird. Why isn't `SelectionInitInterval` working?

src/compile/selection/transforms/project.ts
domoritz: It's a simple type guard.

src/compile/selection/index.ts
domoritz: This get's the right type so when we change `IntervalSelection.init`, the types will be updated correctly.

src/compile/selection/transforms/project.ts
domoritz: Great idea. `SelectionInitMapping` == `Dict` then.

domoritz:
@kanitw I pushed some new changes. Please have a look.

src/selection.ts
```diff
@@ -1,22 +1,16 @@
...
-export type SelectionInit = boolean | number | string | DateTime;
-export type SelectionInitInterval = [boolean, boolean] | [number, number] | [string, string] | [DateTime, DateTime];
-
-export interface SelectionInitMapping {
-  [key: string]: SelectionInit;
-}
-export interface SelectionInitIntervalMapping {
-  [key: string]: SelectionInitInterval;
-}
+export type SelectionInit = Value | DateTime;
```
kanitw: There is nothing specific about SelectionInit.

I think it's better to call this `ValueOrDateTime` and move it to the same place as `Value`?

Also need to update `links.yml`.

src/selection.ts
```diff
@@ -1,22 +1,16 @@
...
+export type SelectionInitInterval = Vector2<boolean> | Vector2<number> | Vector2<string> | Vector2<DateTime>;
```
kanitw: `ValueOrDateTimeTuple`? 

Also need to update links.yml.

src/selection.ts
domoritz: I don't think I want to rename these as it makes it much harder to understand what the type is used for. I prefer if the types have some meaning to them. If we change it, it's also not clear how to link since `Dict<ValueOrDateTime>` is not unambiguously the selection init value.

PR #5176, opened 2019-07-08
chore: refactor optimizationDataflowHelper 

invokesus:
Removes the excessive passing around and reassigning of `mutatedFlag`.

~~Merge after #5170.~~

### DIFF
--- a/src/compile/data/optimize.ts
+++ b/src/compile/data/optimize.ts
@@ -52,29 +52,29 @@ function runOptimizer(optimizer: BottomUpOptimizer | TopDownOptimizer, nodes: Da
 
 function optimizationDataflowHelper(dataComponent: DataComponent, model: Model) {
   let roots = dataComponent.sources;
-  const mutatedFlags: boolean[] = [];
+  const mutatedFlags: Set<boolean> = new Set();
 
-  mutatedFlags.push(runOptimizer(new optimizers.RemoveUnnecessaryNodes(), roots));
+  mutatedFlags.add(runOptimizer(new optimizers.RemoveUnnecessaryNodes(), roots));
 
   // remove source nodes that don't have any children because they also don't have output nodes
   roots = roots.filter(r => r.numChildren() > 0);
 
-  mutatedFlags.push(runOptimizer(new optimizers.RemoveUnusedSubtrees(), getLeaves(roots)));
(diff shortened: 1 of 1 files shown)


src/compile/data/optimize.ts
```diff
@@ -47,35 +47,34 @@ function runOptimizer(optimizer: BottomUpOptimizer | TopDownOptimizer, nodes: Da
...
 }
 
 function optimizationDataflowHelper(dataComponent: DataComponent, model: Model) {
   let roots = dataComponent.sources;
-  let mutatedFlag = false;
+  const mutatedFlags: boolean[] = [];
 
-  // mutatedFlag should always be on the right side otherwise short circuit logic might cause the mutating method to not execute
-  mutatedFlag = runOptimizer(new optimizers.RemoveUnnecessaryNodes(), roots, mutatedFlag);
+  mutatedFlags.push(runOptimizer(new optimizers.RemoveUnnecessaryNodes(), roots));
```
domoritz: Why not just `mutatedFlag = mutatedFlag || runOptimizer(new optimizers.RemoveUnnecessaryNodes(), roots)`? Then we avoid creating an array.

invokesus: That's the pattern I wanted to avoid.
Javascript has short circuit logic, so the 2nd part never runs if the 1st is true.
```
> console.log(true || 1/0 );
true
```

domoritz: Oh, good point. I totally missed that.

src/compile/data/optimize.ts
```diff
@@ -47,35 +47,34 @@ function runOptimizer(optimizer: BottomUpOptimizer | TopDownOptimizer, nodes: Da
...
     }
   });
-  return flags.some(isTrue) || flag;
+  return flags.some(isTrue);
 }
 
 function optimizationDataflowHelper(dataComponent: DataComponent, model: Model) {
   let roots = dataComponent.sources;
-  let mutatedFlag = false;
+  const mutatedFlags: boolean[] = [];
```
domoritz: What if we make this a `Set` and then use `add`? Then we only need to have at most two items in the set.

invokesus: Yeah, that's probably better. 
The return value would then be `<setName>.has(true)`.

domoritz: Let me know when you had a chance to finish this.

invokesus: Pushed changes :)

PR #5077, opened 2019-06-14
Add gradient definition, gradient mark and encoding

domoritz:
This is https://github.com/vega/vega-lite/pull/4970 but on our fork because of https://github.com/vega/vega-lite/issues/5076

PR #5180, opened 2019-07-08
feat: add gradient support for colors

### LINKED ISSUE #4611: Add gradient definition

domoritz:
Fixes #4611


domoritz:
Replaces https://github.com/vega/vega-lite/pull/5077

domoritz:
@kanitw This PR is ready. Can we merge it?
@hungngo97 Where did you submit your examples?

kanitw:
Without examples and docs to demonstrate that the feature works, I don't think this PR is ready. 

Also, please rebase and change merge target to `4.0`.

domoritz:
Done. Rebased and changed the target.

site/docs/types.md
```diff
@@ -42,6 +44,90 @@ Accepts boolean values. For example: `true`, `false`.
...
+```
+
+<a name="LinearGradient" href="#LinearGradient">#</a>
+**Linear Gradient**
+
+A linear gradient interpolates colors along a line, from a starting point to an ending point. By default a linear gradient runs horizontally, from left to right. Use the _x1_, _y1_, _x2_, and _y2_ properties to configure the gradient direction. All coordinates are defined in a normalized [0, 1] coordinate space, relative to the bounding box of the item being colored.
+
+| Name          | Type                          | Description  |
+| :------------ | :---------------------------: | :------------|
+| gradient      | {% include type t="String" %} | {% include required %} The type of gradient. Use `"linear"` for a linear gradient.|
```
domoritz: You just copied the Vega docs. This won't work.

domoritz: This commit made a lot of work for me. Please carefully test your code before pushing it to my branch.

domoritz:
I added the docs

Screen Shot 2019-07-21 at 11 04 37 AM

domoritz:
I made some small changes:

image

kanitw:
Why are there so many commits (including "docs(examples): add `circle_wilkinson_dotplot` example") ?

src/channeldef.ts
```diff
@@ -30,13 +31,15 @@ import {getFullName, QUANTITATIVE, StandardType, Type} from './type';
 import {contains, flatAccessWithDatum, getFirstDefined, internalField, replacePathInField, titlecase} from './util';
 
 export type Value = number | string | boolean | null;
+export type Gradient = GradientLinear | GradientRadial;
```
kanitw: `GradientLinear` and `GradientRadial` will become class names in Altair.  I think Linear/RadialGradient sound more like a proper class name?  (Yes, I think we should fix this in Vega too.)

src/channeldef.ts
```diff
@@ -30,13 +31,15 @@ import {getFullName, QUANTITATIVE, StandardType, Type} from './type';
...
 
 /**
- * Definition object for a constant value of an encoding channel.
+ * Definition object for a constant value (primitive value or gradient definition) of an encoding channel.
  */
-export interface ValueDef<V extends Value = Value> {
+export interface ValueDef<V extends ValueOrGradient = ValueOrGradient> {
   /**
-   * A constant value in visual domain (e.g., `"red"` / "#0099ff" for color, values between `0` to `1` for opacity).
+   * A constant value in visual domain (e.g., `"red"` / `"#0099ff"` / gradient definition for color, values between `0` to `1` for opacity).
```
kanitw: provide a link for "gradient definition"?

src/channeldef.ts
```diff
@@ -61,7 +64,9 @@ export type ChannelDefWithCondition<F extends FieldDef<any>, V extends Value> =
 /**
  * @minProperties 1
  */
-export type ValueDefWithCondition<F extends FieldDef<any>, V extends Value = Value> = Partial<ValueDef<V>> & {
+export type ValueDefWithCondition<F extends FieldDef<any>, V extends ValueOrGradient = ValueOrGradient> = Partial<
```
kanitw: I'm wondering if the default should be `= ValueOrGradient` or ` = Value`?  I'm a bit worried that this might lead to mistakes when we add more channels. That said, if we have to override the default in too many internal codes, I'm fine with this as the default.

src/channeldef.ts
```diff
@@ -104,7 +114,7 @@ export function isConditionalSelection<T>(c: Conditional<T>): c is ConditionalSe
   return c['selection'];
 }
 
-export interface ConditionValueDefMixins<V extends Value = Value> {
+export interface ConditionValueDefMixins<V extends ValueOrGradient = ValueOrGradient> {
```
kanitw: Same here

> I'm wondering if the default should be = ValueOrGradient or = Value? I'm a bit worried that this might lead to mistakes when we add more channels. That said, if we have to override the default in too many internal codes, I'm fine with this as the default.

src/channeldef.ts
domoritz: It's actually nice to be able to type "gradient" and then see the options in autocomplete, though, no?

kanitw: That’s true but it’s reading weird. Most other APIs use linear gradient.

domoritz: I'm renaming it but we also need to rename the Vega schema to make it consistent.

src/channeldef.ts
domoritz: Done

domoritz:
@kanitw tests are passing. Please review and merge.

domoritz:
I squashed everything into one commit.

site/docs/mark/area.md
```diff
@@ -54,6 +54,10 @@ Using `area` mark with one temporal or ordinal field (typically on `x`) and one
 
 <span class="vl-example" data-name="area"></span>
 
+You can set the fill color of an area to a [gradient](/docs/types.html#gradient).
```
kanitw: 1) It's a bit weird that we use different data for `area` and `area_gradient` here in the same subsection (under the same `###`). 

2) area_gradient does use `overlay`, which is in the very next section.  So this example should at least come after area_overlay.

That said, I think it's probably better to just modify `area_overlay` to use gradient, all in one example so we don't add to add too many redundant examples.  (Maybe rename it to `area_gradient_overlay`)?

domoritz: I modified the docs a bit. I didn't want to replace the existing example since it's nice to show the simple `line: true` and the more sophisticated `line: {fill: 'green'}` separately.

PR #5185, opened 2019-07-09
refactor: remove unused code

domoritz:


domoritz:
> Have you checked if isRanged really not used in downstreams project?

[Yes](https://github.com/search?q=org%3Avega+isRanged&type=Code). And even if, the function doesn't use any private stuff so we can easily implement it somewhere else.

PR #5189, opened 2019-07-09
fix: refactor selection initialization and docs

domoritz:
Needed for #5005

### DIFF
--- a/site/docs/selection/init.md
+++ b/site/docs/selection/init.md
@@ -5,7 +5,7 @@ title: Initialzing a Selection
 permalink: /docs/init.html
 ---
 
-To initialize a selection, set the `init` property to a map or array of maps from [projected channels or field names](project.html) to the initial value of the selection. Single selections are initialized with a map to single values, multi selections are initialized with an array of maps to single values, and interval selections are initialized with a single map to arrays of values.
+To initialize a selection, set the `init` property to a map or array of maps from [projected channels or field names](project.html) to the initial value of the selection. Single selections are initialized with a map to single values, multi selections are initialized with a single map or an array of maps to single values, and interval selections are initialized with a single map to arrays of values.
 
 ## Examples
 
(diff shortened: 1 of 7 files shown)


site/docs/selection/init.md
```diff
@@ -5,7 +5,7 @@ title: Initialzing a Selection
 permalink: /docs/init.html
 ---
 
-To initialize a selection, set the `init` property to a map or array of maps from [projected channels or field names](project.html) to the initial value of the selection. Single selections are initialized with a map to single values, multi selections are initialized with an array of maps to single values, and interval selections are initialized with a single map to arrays of values.
+To initialize a selection, set the `init` property to a map or array of maps from [projected channels or field names](project.html) to the initial value of the selection. Single selections are initialized with a map to single values, multi selections are initialized with a single map or an array of maps to single values, and interval selections are initialized with a single map to arrays of values.
```
kanitw: I don't know if users would understand what is "a map", given it's not a JSON construct.  What we are doing is using JSON object to act as a map.

domoritz: Do you prefer to say "mapping" or "dictionary"? Feel free to change this since we are in different time zones 😴

domoritz: Done

domoritz:
@kanitw I updated the description. I think this is ready to go now.

src/compile/selection/transforms/project.ts
```diff
@@ -139,11 +139,11 @@ const project: TransformCompiler = {
       if (scales.has(selCmpt)) {
         log.warn(log.message.NO_INIT_SCALE_BINDINGS);
       } else {
-        const parseInit = <T extends SelectionInitMapping | SelectionInitIntervalMapping>(i: T): T['a'][] => {
+        const parseInit = <T extends SelectionInit | SelectionInitInterval>(i: Dict<T>): T[] => {
```
arvind: This makes _way_ more sense to me. Thanks!!

arvind:
LGTM. @domoritz, could you rebase and then it's ready for merging.

PR #5191, opened 2019-07-10
fix: correct projection `precision`'s type to number (not string)

### LINKED ISSUE #5190: expected precision field of Projection to be a number not a string in VegaLite 3.3.0 schema
I was surprised to find the `precision` field of `Projection` has a type of `string` and not `number`:

https://github.com/vega/schema/blob/master/vega-lite/v3.3.0.json#L8971

```
        "precision": {
          "description": "Sets the threshold for the projection’s [adaptive resampling](http://bl.ocks.org/mbostock/3795544) to the specified value in pixels. This value corresponds to the [Douglas–Peucker distance](http://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm). If precision is not specified, returns the projection’s current resampling precision which defaults to `√0.5 ≅ 0.70710…`.",
          "type": "string"
},
```

domoritz:
Fixes #5190


domoritz:
Related Vega PR: https://github.com/vega/vega/pull/1871

kanitw:
Please name your PR / commit message so it reads like an imperative sentence (e.g., "fix: precision is a number, not a string" -> "fix: correct projection `precision`'s type to number (not string)"

PR #5195, opened 2019-07-13
feat: support color names and hex colors for better auto complete

domoritz:
Screen Shot 2019-07-13 at 1 19 18 PM


domoritz:
Yep, we can add color names. We still need string, though, because of `rgb(0,0,0)`.

domoritz:
We can get all colors from https://developer.mozilla.org/en-US/docs/Web/CSS/color_value with this query `[...document.querySelectorAll('#colors_table code')].map(e => e.innerText).filter(e => !e.startsWith('#'))`.

domoritz:
Screen Shot 2019-07-15 at 9 54 43 AM

src/vega.schema.ts
```diff
@@ -33,7 +33,163 @@ export {VgSortField, VgUnionSortField, VgCompare, VgTitle, LayoutAlign, Projecti
...
+  | 'wheat'
+  | 'whitesmoke'
+  | 'yellowgreen'
+  | 'rebeccapurple';
+
+export type Color =
+  | ColorName
+  | HexColor
+  // RGB, HSL, HSV, HWB, CMYK colors
+  | string;
```
kanitw: should we add Regex pattern for string now that it's just `rgb(...)` / rgba  etc. at this point?

domoritz: I don't think it's worth it (it'll take me a while to get the regex right) at this point as it won't improve autocomplete.

domoritz: The syntax gets pretty subtle so I don't really want to dive into it. See https://developer.mozilla.org/en-US/docs/Web/CSS/color_value.

kanitw:
Okie :)

Issue #4963, opened 2019-05-15
Bar Chart legend does not respect `symbolType` property


johnmdonich:
The bar chart legend only renders as a square `symbolType` and does not respect `symbolType` in the `config` or directly within the color encoding.

### Example spec

```json
{
  "config": {
    "legend": {
      "orient": "bottom",
      "symbolType": "circle"
    }
  },
  "mark": {
    "type": "bar",
    "orient": "horizontal"
  },
  "encoding": {
    "x": {
      "field": "revenue",
      "type": "quantitative"
    },
    "y": {
      "field": "month",
      "type": "nominal"
    },
    "color": {
      "field": "company",
      "type": "nominal",
      "legend": {
        "symbolType": "circle"
      }
    }
  },
  "title": {
    "text": "Stacked Bar"
  },
  "data": {
    "values": [
      {
        "company": "Company 1",
        "month": "2016-01",
        "revenue": 99.74542441836121
      },
      {
        "company": "Company 1",
        "month": "2016-02",
        "revenue": 124.28320211965155
      },
      {
        "company": "Company 2",
        "month": "2016-01",
        "revenue": 64.77394629723233
      },
      {
        "company": "Company 2",
        "month": "2016-02",
        "revenue": 70.07675051264934
      }
    ]
  }
}
```

The above renders the following chart.

![image](https://user-images.githubusercontent.com/2266667/57798448-c802ea00-771a-11e9-87d2-5ca0acacaf3f.png)

[...]

PR #5196, opened 2019-07-13
fix: respect symbolType for legend

domoritz:
Fixes https://github.com/vega/vega-lite/issues/4963


src/compile/legend/properties.ts
```diff
@@ -17,8 +25,39 @@ export function values(legend: Legend, fieldDef: TypedFieldDef<string>) {
...
+  shapeChannelDef:
+    | FieldDefWithCondition<MarkPropFieldDef<string>, Value>
+    | ValueDefWithCondition<MarkPropFieldDef<string>, Value>,
+  markShape: SymbolShape
+): SymbolShape {
+  if (channel !== 'shape') {
+    // use the value from the shape encoding or the mark config if they exist
+    const shape = getFirstConditionValue(shapeChannelDef) || markShape;
+    if (shape) {
+      return shape as SymbolShape;
```
kanitw: What if shape is a custom shape? (Why do we need to cast this as Symbol shape here?)

domoritz: SymbolShape can already be a string. 

```ts
export type SymbolShape =
  | 'circle'
  | 'square'
  | 'cross'
  | 'diamond'
  | 'triangle-up'
  | 'triangle-down'
  | 'triangle-right'
  | 'triangle-left'
  | 'arrow'
  | 'triangle'
  | 'wedge'
  | 'stroke'
  | string;
```

I had to make a few functions generics to remove the `as` here.

domoritz: See 75ffb4d27919f73ebf3ee50e4a65a8c1f052b590

PR #5201, opened 2019-07-19
Refactor: simplify flag typings

domoritz:
* remove `flagKeys` since it's the same as `keys`
* simplify code since the schema generator supports typeof


kanitw:
FYI, semantic commits use lowercase for "refactor".

PR #5204, opened 2019-07-20
fix: use `0.5*a + 0.5*b` for stack mid formula instead of `(a+b)/2` to avoid overflow

kanitw:
This is especially important for temporal in a follow up PR (#5096).


domoritz:
Did we have an integer overflow or what overflow are you talking about?

kanitw:
Integer overflow for time. This is also more flexible for band.

PR #5206, opened 2019-07-20
feat: add `band` property for position field def

### LINKED ISSUE #3819: Band
We should extend `stack` to support `band` for point / text marks.

Basically, currently when we stack points, they use stack-end positions.  However, for Wilkinson Dot Plot and for text labelling, we want to use the middle point instead.

What we can do is to introduce stack band support where a stack_start position will be used when band = 0,  the middle point will be used when band = 0.5, and the end point will be used when band = 1.  (Other values are basically a linear interpolation between the two)  

>~~stack: StackOffset | {offset: StackOffset, band: number}  // Name the latter object StackParams~~

~~(I think this is better than introducing a `stackBand` property as this nesting pattern is more consistent for field definition.)~~ 

If ~~stack is an object and~~ there is a `band`, then the  `stackable` in `valueref.ts` should return a signal that combines the start and end position instead of a simple `fieldRef(channelDef, scaleName, {suffix: 'end'})`.

(See text mark of https://vega.github.io/vega/examples/job-voyager/ for example.) 

- [ ] ~~Modify the code so that stack offset is still read correctly for the rest of the codebase.~~ 
- [ ] Modify the docs
- [ ] Create examples
  - [ ] Replicate stack labels like in https://vega.github.io/vega/examples/job-voyager/
  - [x] Wilkinson dot plot #1060

kanitw:
This property would be important for the upcoming timeUnit PR (#5096). As timeUnit becomes interval, we should be able to set band = 0 to preserve old timeUnit behavior. In this PR, I'm adding `band` for all other cases first so the #5096 would only focus on timeUnit. 

Fix #3819 

See https://github.com/vega/vega-lite/pull/5206/files#diff-374e0cec7400ebae2acf0cd264c3fd57R171 for primary use case. 

cc: @znation


src/channeldef.ts
```diff
@@ -359,6 +359,16 @@ export interface PositionFieldDef<F extends Field>
    * __See also:__ [`impute`](https://vega.github.io/vega-lite/docs/impute.html) documentation.
    */
   impute?: ImputeParams;
+
+  /**
+   * For rect-based marks (`rect`, `bar`, and `image`), mark size relative to bandwidth of [band scales](https://vega.github.io/vega-lite/docs/scale.html#band). If set to `1`, the mark size is set to the bandwidth.  If set to `0.5`, the mark size is half of the bandwidth.
```
domoritz:
```suggestion
   * For rect-based marks (`rect`, `bar`, and `image`), mark size relative to bandwidth of [band scales](https://vega.github.io/vega-lite/docs/scale.html#band). If set to `1`, the mark size is set to the bandwidth. If set to `0.5`, the mark size is half of the bandwidth.
```

src/compile/data/stack.ts
```diff
@@ -127,10 +127,10 @@ export class StackNode extends DataFlowNode {
       return null;
     }
 
-    let dimensionFieldDef: TypedFieldDef<string>;
+    let dimensionFieldDef: PositionFieldDef<string>;
     if (stackProperties.groupbyChannel) {
       const cDef = encoding[stackProperties.groupbyChannel];
-      dimensionFieldDef = getTypedFieldDef(cDef);
+      dimensionFieldDef = getTypedFieldDef(cDef) as PositionFieldDef<string>; // Fair to cast as groupByChannel is always either x or y
```
domoritz: If this is the case, shouldn't `encoding[stackProperties.groupbyChannel]` return a `cDef` with an appropriate type?

src/compile/mark/mixins.ts
```diff
@@ -363,10 +364,12 @@ export function bandPosition(
     }
   }
 
+  const {band = 1} = fieldDef;
+
   return {
     // FIXME: make offset works correctly here when we support group bar (https://github.com/vega/vega-lite/issues/396)
```
domoritz:
```suggestion
    // FIXME: make offset work correctly here when we support group bar (https://github.com/vega/vega-lite/issues/396)
```

src/compile/data/stack.ts
kanitw: Typewise cDef can be a value def / conditional field def.  However, in this case, it can't be a groupby channel if cDef isn't a field. But TS isn't smart enough to know that.  I don't think it's worth fixing here.

domoritz:
Please rebase.

kanitw:
probably should add a bit more docs

PR #5212, opened 2019-07-21
Added example for gradient time series plot

hungngo97:


domoritz:
Please add the example to the gallery, not just as a community example.

PR #5223, opened 2019-07-23
feat: set data: null to ignore the parent's data

### LINKED ISSUE #3316: `data: null` to ignore data from parent + better annotation support
Suppose users want to add an annotation layer with just one mark, we need a way to tell that we don't want to derive data from its parent.

domoritz:
fixes #3316


domoritz:
We need a small fix in the schema generator that I'm working on. Will merge when that's out.

PR #5224, opened 2019-07-23
Implement fit-x and fit-y

### LINKED ISSUE #5133: Setting only `width` when using `autosize: fit` should not use default height in compiled Vega
See the compiled Vega from [this example](https://vega.github.io/editor/#/url/vega-lite/N4KABGBECGCuAuB7AzgSwF4FNIC4oDNV5IAacKAd1QBN4ALXMAFgAYWyJIASZAYzswBbaI0h148AA7IcAelkA3TAHNoAOmVE6sAEZrUiWXwHDFK6AFoANkUyKAzGoBWyRADtS5SNUx8ATqiS8AYeeJAAgmA60H5g-DHwYPTQia5+8MhJAmAAnhYK0FawvlE5WZhgAB75hcXIap6c1CkieMCQsH5Wos3w0LKSiJKwVikhzq4eAL4cUPB+0G7I+Ih+gowA2u2EVvCYfj0psIJqOZgxYAC8l2AATGwskFMAurOQwn4A1qLRB2+Ybl4iGoqDcykYoAgnByEPIUIIqEwVmoomgymws3hkHgOUk2DCqxBbkKjXhUD4hXxYHaCzBmAAyntJIx7CwZnCoZA0sQ2pAAUCieCwpUnhz2ViRW0OZw0co-OY9qJkMdSVjCEiUWE8UMrBjpXNcVTIABHWCLYJ9YJKVWc6CVVDICHYoi60SDYajYLuUXwqbkP1+oA). The `height` is set the default `config.view.height` of `200`, but I'd expect this chart to only "fit" the width (since that is all I specified).

To solve this, I imagine there are two options:
1. We should not apply the config's `height` or `width` when using `"autosize": "fit"` with one of the dimensions specified. 
2. We allow for some value of `"autosize"` that specifies only to `"fit"` along one dimension.

willium:
Fixes #5133

### DIFF
--- a/src/compile/layoutsize/parse.ts
+++ b/src/compile/layoutsize/parse.ts
@@ -110,20 +110,12 @@ function defaultUnitSize(model: UnitModel, sizeType: 'width' | 'height'): Layout
     const range = scaleComponent.get('range');
 
     if (hasDiscreteDomain(scaleType)) {
-      if (isVgRangeStep(range)) {
+      const size = getViewConfigDiscreteSize(config.view, sizeType);
+      if (isVgRangeStep(range) || isStep(size)) {
         // For discrete domain with range.step, use dynamic width/height
         return 'step';
       } else {
-        const size = getViewConfigDiscreteSize(config.view, sizeType);
-        if (isStep(size)) {
-          if (model.fit) {
-            return getViewConfigContinuousSize(config.view, sizeType);
-          } else {
-            return 'step';
-          }
-        } else {
-          return size;
-        }
+        return size;
       }
     } else {
       return getViewConfigContinuousSize(config.view, sizeType);
--- a/test/compile/compile.test.ts
+++ b/test/compile/compile.test.ts
@@ -47,6 +47,79 @@ describe('compile/compile', () => {
     expect(spec.marks.length).toEqual(1); // just the root group
   });
 
+  it('should drop fit in top-level properties for discrete x discrete chart', () => {
+    const spec = compile({
+      data: {
+        values: [{x: 'foo', y: 'bar'}]
+      },
+      autosize: 'fit',
+      mark: 'point',
+      encoding: {
+        x: {field: 'x', type: 'nominal'},
+        y: {field: 'y', type: 'nominal'}
+      }
+    }).spec;
+
+    expect(spec.autosize).toBe('pad');
+  });
+
+  it('should drop fit-y in top-level properties for quantitative x discrete chart', () => {
+    log.wrap(localLogger => {
+      const spec = compile({
+        data: {
+          values: [{x: 1, y: 'bar'}]
+        },
+        autosize: 'fit',
+        mark: 'point',
+        encoding: {
+          x: {field: 'x', type: 'quantitative'},
+          y: {field: 'y', type: 'nominal'}
+        }
+      }).spec;
+
+      expect(localLogger.warns[0]).toEqual(log.message.droppingFit());
+      expect(spec.autosize).toBe('fit-x');
+    });
+  });
+
+  it('should drop fit-y in top-level properties for quantitative x discrete chart', () => {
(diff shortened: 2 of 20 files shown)


kanitw:
Add test cases to improve test coverage?  there are still quite amount of lines on `compile.ts` not covered.

src/compile/layoutsize/parse.ts
```diff
@@ -110,20 +110,12 @@ function defaultUnitSize(model: UnitModel, sizeType: 'width' | 'height'): Layout
     const range = scaleComponent.get('range');
 
     if (hasDiscreteDomain(scaleType)) {
-      if (isVgRangeStep(range)) {
+      const size = getViewConfigDiscreteSize(config.view, sizeType);
```
kanitw: Looking at this I'm not sure why we should read size only from `getViewConfigDiscreteSize(config.view, sizeType)`.

But given we did this previously, I don't think fixing this is within the scope of this PR. 

I'm filing a separate issue (https://github.com/vega/vega-lite/issues/5226), as there might be a wrong code here.

willium:
> Add test cases to improve test coverage? there are still quite amount of lines on `compile.ts` not covered.

What cases are missing re: autosize?

kanitw:
See codecov -- it's 82.53% right now.

willium:
> See codecov -- it's 82.53% right now.

I don't really see any code paths that are worth testing that I haven't tested. feel free to point some out, happy to write them.

kanitw:
Somehow you have to expand compile.ts. (I don't why codecov collapses it by default).

image



I think your tests already covered these branches but I'm not sure why it doesn't look like they did.  Perhaps, try to run jest with coverage locally to see if the lines are well covered?

willium:
they seem to be well covered. I definitely tested all three main cases.

kanitw:
I agree. but it's weird that it's not covered in the coverage reports. 

So I think we better investigate this before merging.  (At least try to get a coverage report locally and see if they are covered. If they are not, maybe you're overlooking something subtle.)

willium:
> I agree. but it's weird that it's not covered in the coverage reports.
> 
> So I think we better investigate this before merging. (At least try to get a coverage report locally and see if they are covered. If they are not, maybe you're overlooking something subtle.)

I'm not sure how to get it running locally... I ran `jest --coverage` and it has the same 82.53%. I also don't really want to spend a bunch of time appeasing code coverage when I know it's mistake. Unless there's something I'm missing?

test/compile/compile.test.ts
```diff
@@ -47,6 +47,124 @@ describe('compile/compile', () => {
...
+          y: {field: 'y', type: 'nominal'}
+        }
+      }).spec;
+
+      expect(localLogger.warns.length).toEqual(0);
+      expect(spec.autosize).toBe('fit');
+    });
+  });
+
+  it('should NOT drop fit-x in top-level properties for specified width chart', () => {
```
kanitw: Actually, you don't need this extra level of `() => {`.  

See this removed code https://github.com/vega/vega-lite/pull/5224/files#diff-9885764e87ed89fc726e6fe7d9a5ceeeL127, for example. 

Hopefully this isn't the cause of the low coverage, but maybe it's relevant.

domoritz:
Please rebase onto master

test/compile/compile.test.ts
willium: fixed in latest commit

willium:
> Please rebase onto master

Rebased on `master`... which destroyed the branch :D. Then rebased on `4.0`, and we're good.

willium:
@kanitw, through no change at all, it seems codecov is at 96.55% now.

kanitw:
🎉

kanitw:
Thanks for your contribution ;)

PR #5237, opened 2019-07-26
fix!: make temporal x-axis left aligned and y-axis bottom baselined by default

kanitw:
(or flipped for reversed scales)


kanitw:
I don't think this is a good default anymore.

PR #5251, opened 2019-07-30
refactor: use flat, flatMap, and trim

domoritz:


domoritz:
Note that I had to add a polyfill for older node versions. Most browsers support flat and flatMap.

Issue #2907, opened 2017-08-20
Conditional axis config


kanitw:
- [x] Remove _future suffixes from examples in https://github.com/vega/vega-lite/pull/3283

kanitw:
The current syntax also feels a bit too many level of nesting for me.  But we can decide about this in the future.  

```
            "axis": {
                "format": "%Y",
                "labelAngle": 0,
                "labelOverlap": false,
                "encoding": {
                    "ticks": {
                        "stroke": {
                            "condition": {
                            "test": {"timeUnit": "monthdate", "field": "value", "equal": {"month": 1, "date": 1}},
                            "value": "black"
                            }
                        }
                    },
                    "labels": {
                        "fill": {
                            "condition": {
                            "test": {"timeUnit": "monthdate", "field": "value", "equal": {"month": 1, "date": 1}},
                            "value": "black"
                            }
                        }
                    }
                }
            }
```

Maybe the "encoding" is not a sufficient level of nesting.   ("ticks", "labels" are just boolean anyway, if they are objects, they are basically "enabled") 

```
            "axis": {
                "format": "%Y",
                "labelAngle": 0,
                "labelOverlap": false,
                "ticks": {
                    "stroke": {
                        "condition": {
                        "test": {"timeUnit": "monthdate", "field": "value", "equal": {"month": 1, "date": 1}},
                        "value": "black"
                        }
                    }
                },
                "labels": {
                    "fill": {
                        "condition": {
                        "test": {"timeUnit": "monthdate", "field": "value", "equal": {"month": 1, "date": 1}},
                        "value": "black"
                        }
                    }
                }
            }
```

kanitw:
@domoritz  Any thoughts about this?

domoritz:
I like your second suggestion.

kanitw:
I feel like the only two cases we need for accessing encode blocks are:

1) Adjusting expression, which might be better done via including expression as a part of format (https://github.com/vega/vega-lite/issues/5122#issuecomment-506870319).

2) Conditional appearance / color for axis/legend parts (e.g., ticks and labels in [the lasagna plot](https://github.com/vega/vega-lite/blob/master/examples/specs/rect_lasagna_future.vl.json) and grid e.g., in [voyager user study plot](https://github.com/vega/vega-lite/blob/master/examples/specs/concat_layer_voyager_result_future.vl.json).

Thus, I think it might be better to just make the following properties support conditional value: 
[...]

PR #5253, opened 2019-07-31
feat: support conditional axis + add examples

kanitw:
part of #2907


build/vega-lite-schema.json
```diff
@@ -560,34 +560,76 @@
...
         "labelAlign": {
-          "$ref": "#/definitions/Align",
-          "description": "Horizontal text alignment of axis tick labels, overriding the default setting for the current axis orientation."
+          "anyOf": [
+            {
+              "$ref": "#/definitions/Align",
+              "description": "Horizontal text alignment of axis tick labels, overriding the default setting for the current axis orientation."
+            },
+            {
+              "$ref": "#/definitions/ConditionalAxisProperty<(def-alias-981995915-5990-6039-981995915-0-9900|undefined|null)>"
```
kanitw: > def-alias-981995915-5990-6039-981995915-0-9900

@domoritz  I don't know why the JSON schema generator doesn't read the name `Align` from Vega and put it here. 

That said, the schema is still technically correct here.

domoritz: I'd need to look more into this but it's not an easy type: `[k in ConditionalAxisProp]?: BaseAxisNoSignals[k] | ConditionalAxisProperty<BaseAxisNoSignals[k]>;`.

site/docs/encoding/axis.md
```diff
@@ -101,6 +101,25 @@ For example, the following plot has a customized x-axis title.
 **TODO** (We have `encoding` property akin to [Vega's axis `encode`](https://vega.github.io/vega/docs/axes/#custom-axis-encodings), but within each element's block, we do not have `enter/update/exit`.)
 -->
 
+{:#conditional}
+
+### Conditional Axis Properties
+
+We can set axis properties (whose type include "ConditionalAxisProperty") to a [conditional value definition](condition.html#value).
```
domoritz:
```suggestion
We can set axis properties (which can be of type "ConditionalAxisProperty") to a [conditional value definition](condition.html#value).
```

src/axis.ts
```diff
@@ -33,8 +35,119 @@ type BaseAxisNoSignals = AxisMixins &
...
+
+export type ConditionalAxisProperty<V extends Value | number[]> = ValueDef<V> & {
+  condition: ConditionalPredicate<ValueDef<V>> | ConditionalPredicate<ValueDef<V>>[];
+};
+
+export function isConditionalAxisValue<V extends Value | number[]>(v: any): v is ConditionalAxisProperty<V> {
+  return v['condition'];
+}
+
 // Vega axis config is the same as vega axis base. If this is not the case, add specific type.
```
```suggestion
// Vega axis config is the same as Vega axis base. If this is not the case, add specific type.
```

PR #5254, opened 2019-07-31
docs: describe options for stack offset

domoritz:


src/transform.ts
```diff
@@ -124,7 +124,9 @@ export interface StackTransform {
    */
   groupby: FieldName[];
   /**
-   * Mode for stacking marks.
+   * Mode for stacking marks. One of `"zero"` (default), `"center"`, or `"normalize"`.
```
kanitw: Missing description for zero. 

We have description for stack in encoding that you might want to copy from
https://vega.github.io/vega-lite/docs/stack.html#encoding

domoritz: I copied the description from Vega.

domoritz: Done

kanitw: yeah, but given this will appear on the same page as the encoding one. Please revise to see if they look consistent

PR #5257, opened 2019-07-31
feat: make dataflow node abstract, implement hash, dependentFields, producedFields for all nodes

domoritz:


domoritz:
Needed for https://github.com/vega/vega-lite/pull/4996

domoritz:
Now with 💯 % test coverage.

domoritz:
It's a feature since we produce more efficient dataflows. The previous code was correct.

Issue #5263, opened 2019-08-05
Support boxplot along continuous scale


michaelaye:
Currently, it seems that the orientation of boxes for boxplots is inferred via which axis is set to type `ordinal` (IIUC).

However, there are use cases for boxplots over irregularly sampled points in time, e.g. at Mars, we study ten-thousands of surface deposits with these distributions of areas, here, at 3 different points in time (time is called Ls = solar longitude for orbiting planetary objects):

![ithaca_histograms](https://user-images.githubusercontent.com/69774/62431691-bfe99380-b6e6-11e9-818b-b4c29f1fe23f.png)

resulting in plots that describe the development of these distributions over a range of time, like this plot, where we use error bars of interquartile ranges, to emulate boxplots:

![Figure 1 (1)](https://user-images.githubusercontent.com/69774/62431725-0b9c3d00-b6e7-11e9-8314-d46db6b6d0c6.png)

The current norm for boxplots is to do this:

```json
{
  "data": {"url": "https://vega.github.io/vega-datasets/data/population.json"},
  "mark": {"type": "boxplot", "extent": "min-max"},
  "encoding": {
    "x": {"type": "ordinal", "field": "age"},
    "y": {"type": "quantitative", "field": "people"}
  }
}
```
creating:

![visualization (1)](https://user-images.githubusercontent.com/69774/62419202-0b913400-b638-11e9-8cb7-451f30f963ea.png)

but in that case, when there'd be data missing over the x-axis, that void would just be filled, as x-axis isn't a quantity (as discussed here: https://github.com/altair-viz/altair/issues/1496).

Jake's work-around in that issue was to impose a domain scale with given bins like so:

```python
x=alt.X('age:O', scale=alt.Scale(domain=np.arange(0, 95, 5).tolist())),
```

but my proposed feature is to simply switch the x-axis type to `quantitative`, so that the boxes are being placed at given data points for x.

However, trying to do that currently, like so:

```json
{
  "data": {"url": "https://vega.github.io/vega-datasets/data/population.json"},
  "mark": {"type": "boxplot", "extent": "min-max"},
  "encoding": {
    "x": {"type": "quantitative", "field": "age"},
    "y": {"type": "quantitative", "field": "people"}
  }
}
```

creates an odd looking plot where bins seem to be merged:

![visualization](https://user-images.githubusercontent.com/69774/62419015-3e392d80-b634-11e9-8136-d75fe6ba6763.png)

[...]

PR #5273, opened 2019-08-06
fix: simplify orient for bar, better default for ranged bar

domoritz:
Improves #5263


domoritz:
This is breaking some examples. Need to look into it.

PR #5274, opened 2019-08-06
 fix: set orient of bar in boxpolot macro

### LINKED ISSUE #4849: The boxes in Boxplot have incorrect size and offset when setting both fields quantitative.
```json
{
  "mark": {"type": "boxplot"},
  "data": {
    "values": [
      {"lat": 47.608013, "sunshine": 69},
      {"lat": 47.608013, "sunshine": 108},
      {"lat": 41.881832, "sunshine": 135},
      {"lat": 41.881832, "sunshine": 187}
    ]
  },
  "encoding": {
    "x": {"field": "lat", "type": "quantitative"},
    "y": {"field": "sunshine", "type": "quantitative"}
  }
}
```

This spec produces this chart:
visualization

### LINKED ISSUE #5263: Support boxplot along continuous scale

domoritz:
fixes #5263, fixes #4849

Also detect the orientation of bar marks from aggregation.

### DIFF
--- a/src/compile/mark/init.ts
+++ b/src/compile/mark/init.ts
@@ -98,10 +98,10 @@ function orient(mark: Mark, encoding: Encoding<string>, specifiedOrient: Orienta
 
   switch (mark) {
     case BAR:
-      if (isFieldDef(x) && isBinned(x.bin)) {
+      if (isFieldDef(x) && (isBinned(x.bin) || (isFieldDef(y) && y.aggregate && !x.aggregate))) {
         return 'vertical';
       }
-      if (isFieldDef(y) && isBinned(y.bin)) {
+      if (isFieldDef(y) && (isBinned(y.bin) || (isFieldDef(x) && x.aggregate && !y.aggregate))) {
         return 'horizontal';
       }
       if (y2 || x2) {
@@ -144,9 +144,9 @@ function orient(mark: Mark, encoding: Encoding<string>, specifiedOrient: Orienta
           return 'horizontal';
         }
       } else if (mark === RULE) {
-        if (encoding.x && !encoding.y) {
+        if (x && !y) {
           return 'vertical';
-        } else if (encoding.y && !encoding.x) {
+        } else if (y && !x) {
           return 'horizontal';
         }
       }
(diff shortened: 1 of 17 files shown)


domoritz:
We could set it only if we have a q x q boxplot...

src/compile/mark/init.ts
```diff
@@ -98,10 +98,10 @@ function orient(mark: Mark, encoding: Encoding<string>, specifiedOrient: Orienta
 
   switch (mark) {
     case BAR:
-      if (isFieldDef(x) && isBinned(x.bin)) {
+      if (isFieldDef(x) && (isBinned(x.bin) || (isFieldDef(y) && y.aggregate && !x.aggregate))) {
```
chanwutk: Should we add the case where there are `x`m `y` and `y2` but not `x2`?

domoritz: Can you give me an example? I don't know how to answer your questions but we should cover all reasonable cases.

chanwutk: For example, if there are `x`, `y`, and `y2` in the encoding channel, it should be `"vertical"`
```javascript
{
  mark: "bar",
  encoding: {
    x: {type: "quantitative", field: "field_for_x"},
    y: {type: "quantitative", field: "field_for_y"},
    y2: {field: "field_for_y2"}
  }
}
```

domoritz: Unfortunately, binned data looks the same: https://vega.github.io/vega-lite/docs/bin.html#binned and the orientation should be different.

chanwutk: I see.

PR #5275, opened 2019-08-06
fix: bug with top-level signal for scale-bound selection

### LINKED ISSUE #4975: Strange behavior when binding selection multiple times
First reported in https://github.com/altair-viz/altair/issues/1514

Situation: concatenation of two charts that share an interval selection bound to x & y

Expected: both panels can pan and zoom in x and y (this is the behavior with Vega-Lite 2.X

Actual: second chart can only pan and zoom along the y axis.

Sample spec:
```json
{
  "hconcat": [
    {
      "mark": "point",
      "encoding": {
        "x": {"type": "quantitative", "field": "Miles_per_Gallon"},
        "y": {"type": "quantitative", "field": "Weight_in_lbs"}
      },
      "selection": {"selector001": {"type": "interval", "bind": "scales"}}
    },
    {
      "mark": "point",
      "encoding": {
        "x": {"type": "quantitative", "field": "Acceleration"},
        "y": {"type": "quantitative", "field": "Horsepower"}
      },
      "selection": {"selector001": {"type": "interval", "bind": "scales"}}
    }
  ]
}
```

([vega editor link](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAutIC4zEgVwE4BsVTkQHoBjWLAZwDoArCgewDtIBfAGnCgAsSmz58AbU4RQEcVAC25ANb5IAB3oBLRgI4SIkAKaNe0VQHN8YzVoAeJyPACeC7fICOGWGuWJ4ygG4O2UAGbK2jjQ8gCyyjjaFAD69lgxAOKwODhMrBpmUDZWtvZOLm4e3r4BQSHyAOrayoZc8DGqMTgARhSsIuLsnVoUwdoknumo6H1Rg-RYAAxTAEy5dg6okKrw2lheKZB+kC2qocsUZFHtLCyd3RKmEpDSWHLLSqvbPVC6+kYmrxYL+cvOrk8xR82zKwQOUAAgiQSP0sAhlOlLllIDkRtZFgVAe4ESCdoFwfIABKTPpKADu6w6ZmRNzGAyGzHR9Im0zmvyWUFW602eB2e0YEMgRxS0VY5wkErAAF0QCwgA))

arvind:
Fixes #4975 which was caused by overriding the top-level signal expression when a scale-bound selection was projected over multiple fields.

### DIFF
--- a/src/compile/selection/transforms/scales.ts
+++ b/src/compile/selection/transforms/scales.ts
@@ -66,7 +66,7 @@ const scaleBindings: TransformCompiler = {
       for (const proj of bound) {
         const mapping = `${stringValue(proj.field)}: ${proj.signals.data}`;
         if (update.indexOf(mapping) < 0) {
-          update = update.substring(0, update.length - 1) + `, ${mapping}}`;
+          update = `${update.substring(0, update.length - 1)}, ${mapping}}`;
         }
       }
       namedSg.update = update;
--- a/src/compile/selection/transforms/scales.ts
+++ b/src/compile/selection/transforms/scales.ts
@@ -59,16 +59,17 @@ const scaleBindings: TransformCompiler = {
     // outer" to from within the units. We need to reassemble this state into
     // the top-level named signal, except no single selCmpt has a global view.
     const namedSg = signals.filter(s => s.name === selCmpt.name)[0];
-    const update = namedSg.update;
+    let update = namedSg.update;
     if (update.indexOf(VL_SELECTION_RESOLVE) >= 0) {
       namedSg.update = `{${bound.map(proj => `${stringValue(proj.field)}: ${proj.signals.data}`).join(', ')}}`;
     } else {
       for (const proj of bound) {
-        const mapping = `, ${stringValue(proj.field)}: ${proj.signals.data}`;
+        const mapping = `${stringValue(proj.field)}: ${proj.signals.data}`;
         if (update.indexOf(mapping) < 0) {
-          namedSg.update = update.substring(0, update.length - 1) + mapping + '}';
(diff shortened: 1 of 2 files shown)


src/compile/selection/transforms/scales.ts
```diff
@@ -59,16 +59,17 @@ const scaleBindings: TransformCompiler = {
...
+    let update = namedSg.update;
     if (update.indexOf(VL_SELECTION_RESOLVE) >= 0) {
       namedSg.update = `{${bound.map(proj => `${stringValue(proj.field)}: ${proj.signals.data}`).join(', ')}}`;
     } else {
       for (const proj of bound) {
-        const mapping = `, ${stringValue(proj.field)}: ${proj.signals.data}`;
+        const mapping = `${stringValue(proj.field)}: ${proj.signals.data}`;
         if (update.indexOf(mapping) < 0) {
-          namedSg.update = update.substring(0, update.length - 1) + mapping + '}';
+          update = update.substring(0, update.length - 1) + `, ${mapping}}`;
```
domoritz:
```suggestion
          update = `${update.substring(0, update.length - 1)}, ${mapping}}`;
```

arvind: I find this less readable, but will commit the suggestion if you prefer it. 

Happy to add a test case, but thought this was a highly specific typo/bug.

domoritz: In this case, I prefer using only the template a bit since it makes whitespaces clearer.

domoritz: I think a small test case would be good since we didn't catch the bug even though the patch has 100% test coverage.

arvind:
Test case added!

Issue #4936, opened 2019-05-04
Disable direct manipulation for widget-bound selections


haldenl:
~It seems like using two selections with one being a filter breaks selection logic. Clicking the scatter circle in the example below throws and error.~
__Even simpler: see Jeff's comment + example below.__

Downgrading to VL 2.* resolves the issue. Doesn't seem to be an issue without aggregation.

cc: @jheer 

Reproduced below:

[link to editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAutIC4zEgO1gWwKYqgGNYAnAZ0gF8AacKeE2DMgMwHsTsCBtOiUCIKgsAlgBt4uEgQFDBkMrjG5C8EWwwFIAZSUr4kPoMpGTEALq0IkbKQDWM+gE8ADvlSRCIkoWWRqUGQiAF7uYABMAAyRYDR0kLgYhGzQIhgA5jJGnmxiHFlyRBqpahoFhc5uWhhs2GmwYv5GQpCiStBaAPIkIulpTRWBeqrqmh6KyqqGhWbGVvIAHuUtsOnpJLjpCGE2uEwDcpDwrjsAjgCuTGqIagBu+PMtbWIdHgASHIoubADuUtNzbJOZbyVbrTbbLR4faPeTHKoeC5XEQ3ET3A5PETtLQAQUIhD0jFKmlMdDi1gm+lGIIUwwMqFkLXhOyCGT8sOsuGwLmOWgajQ5wixLwoqF4g0g3V6-WaYHMplhOjpIOszK0rPS7NlkAARmlXmhZdZbEtUABGACcAHYzYLjf1zdbInaoGkXOd6VBGBkHkbApIXAQzbLyYdntBRWBxRVIABNPbSWWWbVpFEq+Tx0hBp0h0kQUOQAAkZEIAAsuUgPKX4PAXGRkAB6Bv3LYAOj68FL5x1rfUDZL5dszYhAFoxCjcM2AMyt8Kts2tgBWZDK8xgCFgingkcZRFIke4u+sH3IuG+f2kyDNU+d2rxBOUROpV-CgrjCezVtvhUlPT6Y0gABVbQcQBWJHiPKATy+X5-ivAA2ABWN970JBBnzNBdkO1TNLzAS0bTfKV-xQICQLAyh5XzEBKCAA)

```json
{
  "data": {"name": "cars"},
  "transform": [
    {
      "filter": {
        "selection": "Select"
      }
[...]

PR #5277, opened 2019-08-07
feat!: by default, disable direct manipulation for widget-bound selections.

arvind:
As per #4936 and altair-viz/altair#1553, our previous default of leaving direct manipulation interaction enabled for selections bound to input widgets was confusing to users. This PR switches the default such that direct manipulation interaction is disabled by default, but users may choose to explicitly re-enable it by specifying an `on` or `clear` properties.

### DIFF
--- a/examples/compiled/interactive_query_widgets.vg.json
+++ b/examples/compiled/interactive_query_widgets.vg.json
@@ -64,25 +64,11 @@
     {
       "name": "CylYr_Year",
       "init": "1977",
-      "on": [
-        {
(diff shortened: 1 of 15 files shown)


arvind:
@domoritz: I think this is ready to merge? Also, since this is change a default, I've updated the title to indicate a breaking change.

domoritz:
Yes, feel free to merge. Do we need the empty `on` in https://github.com/vega/vega-lite/pull/5277/files#diff-cdccea7c85d33c82998cd590bd519acdR67?

PR #5282, opened 2019-08-10
docs: correct pluralization and capitalization typos

marcprux:
The plural should be "projections", not "projection's"


domoritz:
Good catch. Let's also spell Vega-Lite with capital l and make the pr title semantic so we can merge.

domoritz:
Thank you

Issue #2765, opened 2017-08-04
Lookup on a materialized selection + Indexed Chart Example


kanitw:
We now have look up -- is there anything preventing us from adding it back?

arvind:
Yes, unfortunately -- performing a lookup on a materialized selection (i.e., on a new derived dataset+selection filter). My preference would be to address this in 2.1.

domoritz:
We are now moving towards v4. Can we implement the index chart now?

arvind:
Yes, this would be great to support in v4. I had a branch that had implemented performing lookups on selections but I'd experienced issues where the lookup transform (or timeUnit on the selection) got moved around in ways that I could not understand and thus needed your help @domoritz. Unfortunately, I don't see the branch here on GitHub anymore, and can't find the issue/comment I'd left documenting the issue. 

Will see if I can recreate it soon.

domoritz:
I'm happy to help. Let me know when you have something to demonstrate the issue. We also now have much better debugging of the dataflow. For example, you can comment out https://github.com/vega/vega-lite/blob/262ce13c725bf9021cb1436fab231cc52c61b04b/src/compile/compile.ts#L101 to get a graphical representation of the dataflow.

domoritz:
Done in https://github.com/vega/vega-lite/pull/5286

PR #5286, opened 2019-08-12
Lookup Selections + Index Chart Example

arvind:
Addresses #2765. This is a freshly rebased version of almost year-old code. I've marked it as draft to both verify the changes, but also debug dataflow issues I remember having around where the lookup transform and/or timeunit transforms would get added or rearranged.


arvind:
After investigation, whatever dataflow issues we used to have back in 2018 seemed to have been resolved and the base functionality of this PR appears to work for the included index chart example.

@domoritz, I could use your help though. Right now, for the index chart example included in this PR, we are producing two data sets with this definition (named `data_1` and `data_3`). As far as I can tell, they are respectively being produced when we materialize a selection for the lookup transform, and for the explicit selection filter defined in the spec. It seems wasteful to have this duplication, but I'm not sure how I should change my additions of to the dataflow to be smarter about this. 

```json
{
      "name": "data_1",
      "source": "data_0",
      "transform": [
        {
          "type": "filter",
          "expr": "!(length(data(\"index_store\"))) || (vlSelectionTest(\"index_store\", datum))"
        }
      ]
    }
```

Would you mind taking a look? I'd like to get this resolved before doing unit testing since it may have implications for what gets generated. Thanks!

domoritz:
Vega-Lite's dataset generation works in a few steps. First, we parse and generate a dataflow graph. This graph has dataflow nodes of various kinds. Once we built the complete graph, we run a number of optimizers to remove redundancy. The most important property of an optimizer is to be correct. We are very conservative and there are still a number of fairly simple optimizations we don't do. I suspect that the duplication you are seeing is one of those instances. We should never rely on the optimizer for correctness. If possible, try to write the unit tests in a way where they don't test the generated datasets.

Having said that, we have an optimizer that merges identical nodes at forks, which would cover the case you're seeing. I have to look at the dataflow graph visualization to see what's happening here.

domoritz:
Before optimization:

image

After optimization:

image

It looks like what we need here is an optimizer to move the filter of invalid data up and then it would get merged. I won't have time to do this before the end of the month but I'm planning to spend some time on optimizations then.

domoritz:
Sidenote: the filter invalid on the right side should probably come before the time unit node. Not sure who inserts that node.

domoritz:
Can you squash the commits in this pr?

domoritz:
Can you rebase and squash?

PR #5298, opened 2019-08-14
fix: toggling initialized values from multi selections 

arvind:
Fixes #5199, which revealed several minor bugs:

1. Now that initial values for selections are embedded directly as dataset values (rather than embedded in signal expressions), we no longer need to string escape model names. 

2. Datetime expressions should evaluate to numerical timestamps, rather than strings. 

3. Vega's `modify` expression should also test for object equality (which wasn't previously necessary for our logic because the same object was being reused in every tuple and thus strict equality testing `===` would correctly evaluate to `true`). vega/vega#1909.

4. ~~Although the typings/schema disallow it (and I believe should continue to disallow it, for simplicity/disambiguity), when a multi selection is projected over a single field, it is reasonable to expect people to use a short form initialization as seen in #5199. i.e., `{init: {Origin: ['Japan', 'USA']}}`. Now, we test for this one corner case, and expand it to what it should be `{init: [{Origin: 'Japan'}, {Origin: 'USA'}]}`~~

Issue #5199, opened 2019-07-17
can't toggle initialized multi selection until it's cleared?


gordonwoodhull:
I'm using vega-lite and react-vega for a biggish new project and have quite a few questions and suggestions. I'll start with the most straightforward one to warm up.

I'm using vega-lite 3.3. It's a dashboard with a server-side component. Think of it as a generalization on Nanocubes and crossfilter (and the earlier iPlots). It will be open source but it's not released yet.

The charts update each other a lot, and they go outside of Vega to get new data. This means the selection must be preserved and re-applied with whatever has changed about the data and spec.

Selection initialization with `init` works, but I find that any initialized selection does not toggle, whether or not I explicitly set `"toggle": "event.shiftKey"`. Shift-clicking something new will add it to the selection, but shift-clicking within the selection will not remove anything.

If I clear the selection, this enables toggling for that selection.

Example below with two selections. The behavior is the same with one selection but this makes it really obvious because you can shift-alt-click to toggle the purple selection, but you can only add to the orange selection with shift-click, not remove, until the selection is cleared.

(please excuse the content, I'm using Chicago Crimes as my test data)

I'm eager to contribute once I find my way around!

[...]

PR #5304, opened 2019-08-15
fix!: drop SelectionInitMapping as a possible type for multi selection init.

arvind:
This will reduce confusion and propensity for users to adopt a short hand initialization scheme until such a time that we support columnar initialization (#4766, #5199, #5298).

/cc @gordonwoodhull


domoritz:
I added a `!` to the PR title to indicate the breaking change. It's okay to do this for 4.0.

PR #5307, opened 2019-08-16
fix: improve invalid spec error message

kanitw:
Motivation: a user has a typo in the syntax and was a bit shocked to see just  "Invalid spec" error, which isn't very helpful.

### DIFF
--- a/src/log/message.ts
+++ b/src/log/message.ts
@@ -9,6 +9,7 @@ import {DateTime, DateTimeExpr} from '../datetime';
 import {Mark} from '../mark';
 import {Projection} from '../projection';
 import {ScaleType} from '../scale';
+import {GenericSpec} from '../spec/index';
 import {Type} from '../type';
 import {stringify} from '../util';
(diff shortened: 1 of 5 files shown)


src/log/message.ts
```diff
@@ -17,7 +18,11 @@ import {VgSortField} from '../vega.schema';
  * Collection of all Vega-Lite Error Messages
  */
 
-export const INVALID_SPEC = 'Invalid spec';
+export function invalidSpec(spec: GenericSpec<any, any>) {
+  return `Invalid specification ${JSON.stringify(
+    spec
+  )}.  Make sure the specification include one of the following properties: "mark", "layer", "facet", "hconcat", "vconcat", "concat", or "repeat".`;
```
domoritz:
```suggestion
  )}.  Make sure the specification includes at least one of the following properties: "mark", "layer", "facet", "hconcat", "vconcat", "concat", or "repeat".`;
```

PR #5260, opened 2019-08-04
 feat: support axis/legend `labelExpr` + add example with month initials

kanitw:
Fix #5122
Fix #5249

PR #5322, opened 2019-08-22
feat: support labelExpr in headers

domoritz:
Follow up for https://github.com/vega/vega-lite/pull/5260 to also support headers.

### DIFF
--- a/src/compile/header/assemble.ts
+++ b/src/compile/header/assemble.ts
@@ -113,16 +113,18 @@ function getSort(facetFieldDef: FacetFieldDef<string>, channel: HeaderChannel): 
 }
 
 export function assembleLabelTitle(facetFieldDef: FacetFieldDef<string>, channel: FacetChannel, config: Config) {
-  const {format, labelAngle, labelAnchor, labelOrient} = getHeaderProperties(
-    ['format', 'labelAngle', 'labelAnchor', 'labelOrient'],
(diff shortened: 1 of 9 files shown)


src/compile/header/assemble.ts
```diff
@@ -113,16 +113,18 @@ function getSort(facetFieldDef: FacetFieldDef<string>, channel: HeaderChannel):
...
-  const {format, labelAngle, labelAnchor, labelOrient} = getHeaderProperties(
-    ['format', 'labelAngle', 'labelAnchor', 'labelOrient'],
+  const {format, labelAngle, labelAnchor, labelOrient, labelExpr} = getHeaderProperties(
+    ['format', 'labelAngle', 'labelAnchor', 'labelOrient', 'labelExpr'],
     facetFieldDef,
     config,
     channel
   );
 
+  const tileTextExpr = formatSignalRef(facetFieldDef, format, 'parent', config).signal;
```
kanitw: title? (not tile)

domoritz:
I have some more changes for this that I did on the flight. Namely, support for accessing the raw value. I'll send them in a new PR.

src/compile/header/assemble.ts
domoritz: Uff, soo many typos. Thanks!

PR #5333, opened 2019-08-25
chore: upgrade dependencies

domoritz:


domoritz:
You can ignore the PR build since we are changing the dev dependencies here.

Issue #5340, opened 2019-08-26
Filter data to support overview+detail with ordinal domain


domoritz:
[...]

PR #5342, opened 2019-08-27
docs: use filter by selection instead of binding to domain in minimap

domoritz:
Needed because of #5340.


arvind:
This is a correct solution. Implementing #5340 (which I've labeled an enhancement, rather than a bug fix) would provide another correct solution. But I don't anticipate having any time in the near term to implement it, so I've labeled it as help wanted, and recommend we merge this in for now.

Issue #5287, opened 2019-08-13
Decide what we actually want to return from `compile`


domoritz:
Right now, we return an object with one property `spec`. Initially, we thought we would also return warnings but we instead opted for a logger that gets passed to compile. Should we change the output value for v4 or at least provide a default export for Vega-Lite. 

```ts
import vl from 'vega-lite';

const vgSpec = vl(vlSpec);
```

We could return the dataflow graph for the non-optimized version or the normalized spec from `compile` so that it's actually useful that it is an object.

domoritz:
Done in https://github.com/vega/vega-lite/pull/5555

PR #5344, opened 2019-08-27
feat: export compile function by default, also return normalized spec

domoritz:
Now this works

```js
var {spec, normalized} = vegaLite(vlSpec, options);
```

in addition to

```js
var {spec} = vegaLite.compile(vlSpec, options);
```


src/vl.ts
```diff
@@ -0,0 +1,6 @@
+/**
```
kanitw: Having this file is weird. You can just do the following in index.ts.

```
/**
function compileToSpec(inputSpec: TopLevelSpec, opt: CompileOptions = {}) {
  return compile(inputSpec, opt).spec;
}


export default Object.assign(compileToSpec, {compile, extractTransforms, normalize, version});
```

kanitw:
Yeah, maybe let's not do it. :)

domoritz:
I think that `compile.compile` is not all that useful since it always only returns an object with the spec (#5287). I'd like to deprecate that function. In this PR, I attempted to keep a backward-compatible version around.

domoritz:
Updated. I have not added a new compile function but just export the compile function by default and also make it return the normalized spec.

domoritz:
Replaced by https://github.com/vega/vega-lite/pull/5555

PR #5348, opened 2019-08-27
chore: update to Vega 5.5

domoritz:


domoritz:
Paging @jheer to look at the generated SVGs such as [this one](https://github.com/vega/vega-lite/pull/5348/files?file-filters%5B%5D=.svg&short_path=c6ab747#diff-c6ab74710e2288c7dd4fd9c59f66946e).

domoritz:
It looks like handling of ties in window aggregates has changed in Vega 5.5. Once we know what the right solution is, we go ahead with this PR.

domoritz:
The SVGs change but only in reasonable ways. Items with the same sort value are sorted differently. I think this is good to go. @arvind @kanitw.

PR #5351, opened 2019-08-28
fix: correct references to flattened fields in selection signals.

### LINKED ISSUE #5334: Selections are not working with nested data
For example

```json
{
  "selection": {
    "grid": {
      "type": "interval",
      "bind": "scales"
    }
  },
  "data": {
    "values": [
      {"nested":{ "a": "1", "b": 28}},
      {"nested":{ "a": "2", "b": 55}},
      {"nested":{ "a": "3", "b": 43}},
      {"nested":{ "a": "4", "b": 91}},
      {"nested":{ "a": "5", "b": 81}},
      {"nested":{ "a": "6", "b": 53}},
      {"nested":{ "a": "7", "b": 19}},
      {"nested":{ "a": "8", "b": 87}},     
      {"nested":{ "a": "8", "b": 52}}
    ]
  },
  "mark": "point",
  "encoding": {
    "y": {
      "field": "nested.a",
      "type": "quantitative"
    },
    "x": {
      "field": "nested.b",
      "type": "quantitative"
    }
  }
}
```

from https://github.com/vega/vega-lite/issues/4517#issuecomment-524568226

arvind:
Fields are only flattened if they participate in encoding rules. Thus, we continue to use bracket notation to access datum values in signal expressions as we cannot be guaranteed that the flattened field will be present. However, the field name stored as part of a SelectionProjection will *always* use dot notation as these names are either returned to us flattened by `model.vgField` or users explicitly specify them in the input spec (in which case, dot notation is their only recourse). Thus, when referencing the field via a selection's top-level signal, we just directly access the flattened field name, rather than unpacking it via bracket notation.

Fixes #5334.

### DIFF
--- a/src/compile/data/parse.ts
+++ b/src/compile/data/parse.ts
@@ -322,6 +322,7 @@ export function parseData(model: Model): DataComponent {
   }
 
   head = ParseNode.makeImplicitFromEncoding(head, model, ancestorParse) || head;
+  head = ParseNode.makeImplicitFromSelection(head, model, ancestorParse) || head;
 
   if (isUnitModel(model)) {
     head = GeoJSONNode.parseAll(head, model);
--- a/src/compile/data/parse.ts
+++ b/src/compile/data/parse.ts
@@ -23,8 +23,8 @@ import {
   isJoinAggregate,
   isLoess,
   isLookup,
-  isRegression,
   isPivot,
+  isRegression,
   isSample,
   isStack,
   isTimeUnit,
@@ -43,7 +43,12 @@ import {FilterNode} from './filter';
 import {FilterInvalidNode} from './filterinvalid';
 import {FlattenTransformNode} from './flatten';
 import {FoldTransformNode} from './fold';
-import {ParseNode} from './formatparse';
+import {
+  getImplicitFromEncoding,
+  getImplicitFromFilterTransform,
+  getImplicitFromSelection,
+  ParseNode
+} from './formatparse';
 import {GeoJSONNode} from './geojson';
 import {GeoPointNode} from './geopoint';
 import {GraticuleNode} from './graticule';
@@ -54,8 +59,8 @@ import {JoinAggregateTransformNode} from './joinaggregate';
 import {makeJoinAggregateFromFacet} from './joinaggregatefacet';
 import {LoessTransformNode} from './loess';
 import {LookupNode} from './lookup';
-import {RegressionTransformNode} from './regression';
 import {PivotTransformNode} from './pivot';
+import {RegressionTransformNode} from './regression';
 import {SampleTransformNode} from './sample';
 import {SequenceNode} from './sequence';
 import {SourceNode} from './source';
@@ -145,7 +150,8 @@ export function parseTransformArray(head: DataFlowNode, model: Model, ancestorPa
       transformNode = head = new CalculateNode(head, t);
       derivedType = 'derived';
     } else if (isFilter(t)) {
-      transformNode = head = ParseNode.makeImplicitFromFilterTransform(head, t, ancestorParse) || head;
+      const implicit = getImplicitFromFilterTransform(t);
+      transformNode = head = ParseNode.makeWithAncestors(head, {}, implicit, ancestorParse) || head;
 
       head = new FilterNode(head, model, t.filter);
     } else if (isBin(t)) {
@@ -321,8 +327,10 @@ export function parseData(model: Model): DataComponent {
     head = parseTransformArray(head, model, ancestorParse);
   }
 
-  head = ParseNode.makeImplicitFromEncoding(head, model, ancestorParse) || head;
-  head = ParseNode.makeImplicitFromSelection(head, model, ancestorParse) || head;
+  // create parse nodes for fields that need to be parsed (or flattened) implicitly
+  const implicitSelection = getImplicitFromSelection(model);
+  const implicitEncoding = getImplicitFromEncoding(model);
+  head = ParseNode.makeWithAncestors(head, {}, {...implicitSelection, ...implicitEncoding}, ancestorParse) || head;
 
   if (isUnitModel(model)) {
     head = GeoJSONNode.parseAll(head, model);
--- a/src/compile/data/formatparse.ts
+++ b/src/compile/data/formatparse.ts
@@ -58,6 +58,129 @@ function parseExpression(field: string, parse: string): string {
   }
 }
 
+export function getImplicitFromFilterTransform(transform: FilterTransform) {
+  const implicit: Dict<string> = {};
+  forEachLeaf(transform.filter, filter => {
+    if (isFieldPredicate(filter)) {
+      // Automatically add a parse node for filters with filter objects
+      let val: string | number | boolean | DateTime = null;
+
+      // For EqualFilter, just use the equal property.
+      // For RangeFilter and OneOfFilter, all array members should have
+      // the same type, so we only use the first one.
+      if (isFieldEqualPredicate(filter)) {
+        val = filter.equal;
+      } else if (isFieldRangePredicate(filter)) {
+        val = filter.range[0];
+      } else if (isFieldOneOfPredicate(filter)) {
+        val = (filter.oneOf || filter['in'])[0];
+      } // else -- for filter expression, we can't infer anything
+      if (val) {
+        if (isDateTime(val)) {
+          implicit[filter.field] = 'date';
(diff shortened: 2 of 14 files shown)


domoritz:
Looks good. Are you adding flattening for fields that are only referenced in selections as well?

arvind:
I'm not planning to primarily due to time constraints (with my limited time, I'd rather push forward on code paths I'm already familiar and this felt like a principled fix rather than a hack/workaround).

domoritz:
Okay. Can you explain what you meant by "this PR felt like a principled fix rather than a hack/workaround"?

arvind:
Sure -- I mean that the description/reason I gave when opening this PR feels principled in the sense that selection signals do not need to make any assumptions about what is or isn't happening to the `datum`. Instead, selections can remain agnostic/independent about whether any flattening has occurred and react purely based on what the user specifies (which matches what is returned by `model.vgField`).

domoritz:
Just to clarify, if the user starts with this spec.

```json
{
  "selection": {
    "grid": {
      "type": "multi", "fields": ["nested.b"]
    }
  },
  "data": {
    "values": [
      {"nested":{ "a": "1", "b": 28}},
      {"nested":{ "a": "2", "b": 55}},
      {"nested":{ "a": "3", "b": 43}},
      {"nested":{ "a": "4", "b": 91}},
      {"nested":{ "a": "5", "b": 81}},
      {"nested":{ "a": "6", "b": 53}},
      {"nested":{ "a": "7", "b": 19}},
      {"nested":{ "a": "8", "b": 87}},     
      {"nested":{ "a": "8", "b": 52}}
    ]
  },
  "mark": "point",
  "encoding": {
    "y": {
      "field": "nested.a",
      "type": "quantitative"
    },
    "x": {
      "field": "nested.a",
      "type": "quantitative"
    }
  }
}
```

And now they add a calculate transform to derive fields `nested.aa` and `nested.bb`. 

In the spec below, the user has to know that they have to write `"fields": ["nested\\.bb"]` instead of `"fields": ["nested.bb"]` in the selection. They don't need to change anything in the encoding. 

```json
{
  "selection": {
    "grid": {
      "type": "multi", "fields": ["nested.bb"]
    }
  },
  "data": {
    "values": [
      {"nested":{ "a": "1", "b": 28}},
      {"nested":{ "a": "2", "b": 55}},
      {"nested":{ "a": "3", "b": 43}},
      {"nested":{ "a": "4", "b": 91}},
      {"nested":{ "a": "5", "b": 81}},
      {"nested":{ "a": "6", "b": 53}},
      {"nested":{ "a": "7", "b": 19}},
      {"nested":{ "a": "8", "b": 87}},     
      {"nested":{ "a": "8", "b": 52}}
    ]
  },
  "transform": [{
    "calculate": "datum.nested.a", "as": "nested.aa"
  }, {
    "calculate": "datum.nested.b", "as": "nested.bb"
  }],
  "mark": "point",
  "encoding": {
    "y": {
      "field": "nested.aa",
      "type": "quantitative"
    },
    "x": {
      "field": "nested.aa",
      "type": "quantitative"
    },
    "color": {
      "condition": {"selection": "grid", "value": "red"}
    }
  }
}
```

arvind:
Thanks for that example, @domoritz! That helped me understand the full scope of flattening. And it turns out updating `formatparse` wasn't as hard/as much work as I'd initially feared, yay! Updated this PR to more thoroughly address this issue, with corresponding test cases. So it should be ready for re-review.

domoritz:
The second example in https://github.com/vega/vega-lite/pull/5351#issuecomment-525977832 still doesn't work for me. Clicking the circles doesn't make them red. 

If I replace

```json
{
      "name": "grid_tuple_fields",
      "value": [{"type": "E", "field": "nested.bb"}]
    },
```

with

```json
{
      "name": "grid_tuple_fields",
      "value": [{"type": "E", "field": "nested\\.bb"}]
    },
```

in the generated Vega, it works.

src/compile/data/parse.ts
```diff
@@ -322,6 +322,7 @@ export function parseData(model: Model): DataComponent {
   }
 
   head = ParseNode.makeImplicitFromEncoding(head, model, ancestorParse) || head;
```
domoritz: I believe that we may create unnecessarily flattening for fields that are already flattened in `makeImplicitFromEncoding`. I see three solutions. First, we could extract the code that computes `implicit` for encodings and selections and then call `makeWithAncestors` once for a single implicit dict. Second, makeImplicitFromEncoding could return its `implicit`, which we pass to `makeImplicitFromSelection`, and there we ignore fields that are already flattened otherwise. Second, you could merge `makeImplicitFromEncoding` and `makeImplicitFromSelection` so that they share the same `implicit` dict. I guess this is like the first approach but encapsulates the logic. I'd prefer the first as it's most explicit.

arvind: Thanks for the comment. Yes, I'd thought about this and tested it informally (e.g., using the spec below in which both `makeImplicitFromEncoding` and `makeImplicitFromSelection` are called). As far as I could tell, duplicate flattening does _not_ occur. But, I wasn't sure how to formally test this as a unit test (i.e., focused on just parsing/assembling the `ParseNode`). I'm happy to add a test case that parses the whole model and assembles the full dataflow and test what's in the `transform` array, but that doesn't feel like the right approach. Open to ideas/suggestions here. 

Re: your suggested solutions, option (1) won't work because the existing `fromEncoding` logic expects `TypedFieldDefs` -- selections only hold field and channel names as strings to simplify assembly. I'm happy to merge into a single `makeImplicitFromEncodingAndSelection` such that only a single `implicit` object is populated and passed to `makeWithAncestors` if we prefer that. I chose the current approach because I thought that better separated concerns. 

Let me know what you'd prefer.

arvind:
Thanks, I'd forgotten to send a PR for updating the `vlSelectionTest` function to no longer use a field accessor function. We should see the expected behavior with vega/vega#2002.

src/compile/data/parse.ts
arvind:
```json
{
  "selection": {
    "grid": {
      "type": "multi", "fields": ["nested.a"]
    }
  },
  "data": {
    "values": [
      {"nested":{ "a": "1", "b": 28}},
      {"nested":{ "a": "2", "b": 55}},
      {"nested":{ "a": "3", "b": 43}},
      {"nested":{ "a": "4", "b": 91}},
      {"nested":{ "a": "5", "b": 81}},
      {"nested":{ "a": "6", "b": 53}},
      {"nested":{ "a": "7", "b": 19}},
      {"nested":{ "a": "8", "b": 87}},     
      {"nested":{ "a": "8", "b": 52}}
    ]
  },
  "mark": "point",
  "encoding": {
    "y": {
      "field": "nested.b",
      "type": "quantitative"
    },
    "x": {
      "field": "nested.b",
      "type": "quantitative"
    },
    "color": {
      "condition": {"selection": "grid", "value": "red"}
    }
  }
}
```

does not yield any duplicate flattening (though it should, were it an issue I think?).

domoritz: Which field would you expect to be flattened twice?

I can take care of avoiding duplicate flattening since I am more familiar with the dataflow stuff.

domoritz:
Is it necessary to update Vega or should Vega-Lite generate different spec as I suggested above?

src/compile/data/parse.ts
arvind: Sorry, I think we did a simultaneous update. In the updated spec above, the selection is projected over `nested.b` which is also used in `encoding`. We would expect to see duplicates, but we don't. I think this was your original concern in https://github.com/vega/vega-lite/pull/5351#discussion_r321986174 but perhaps I'm misunderstanding?

arvind:
We should update Vega to keep our logic consistent. Otherwise, we introduce more complexity of needing to remember when we're referring to flattened fields, and when we're referring to nested fields. I'd rather not need to remember how to coordinate that, and just reference flattened fields always.

arvind:
Oh I just noticed that Vega-Lite outputs `nested\\.b` for field refs elsewhere. Which function should I be calling to output the escaped version instead?

domoritz:
We have to remember already since when we access a `field` in Vega and use a `.`, Vega interprets it as a nested access. But if we use `.` to write a field with `as`, the field is flattened and `.` becomes part of the name of the field.

domoritz:
@arvind Please have a look at afc6337 and merge if you think it looks good.

arvind:
That’s a nice solution! Thanks @domoritz!

PR #5362, opened 2019-09-01
refactor: ensure more functions don't modify arguments, remove unused functions

domoritz:
Inspired by https://twitter.com/WrocTypeScript/status/1167907329389387779.

### DIFF
--- a/src/compile/facet.ts
+++ b/src/compile/facet.ts
@@ -121,7 +121,7 @@ export class FacetModel extends ModelWithField {
     parseFacetHeaders(this);
   }
 
-  public assembleSelectionTopLevelSignals(signals: NewSignal[]): NewSignal[] {
+  public assembleSelectionTopLevelSignals(signals: readonly NewSignal[]): readonly NewSignal[] {
     return this.child.assembleSelectionTopLevelSignals(signals);
   }
 
@@ -130,7 +130,7 @@ export class FacetModel extends ModelWithField {
     return [];
   }
 
-  public assembleSelectionData(data: VgData[]): VgData[] {
+  public assembleSelectionData(data: readonly VgData[]): readonly VgData[] {
     return this.child.assembleSelectionData(data);
   }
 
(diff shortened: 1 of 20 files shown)


src/compile/facet.ts
```diff
@@ -121,7 +121,7 @@ export class FacetModel extends ModelWithField {
     parseFacetHeaders(this);
   }
 
-  public assembleSelectionTopLevelSignals(signals: NewSignal[]): NewSignal[] {
+  public assembleSelectionTopLevelSignals(signals: readonly NewSignal[]): readonly NewSignal[] {
     return this.child.assembleSelectionTopLevelSignals(signals);
```
arvind: I'm surprised this works because `assembleSelectionTopLevelSignals` is not pure by design, and definitely makes changes to the `signals` array. As a result, I'm a little worried whether `readonly` poses a danger in being a red herring (e.g., we think functions are pure when they are not).

src/compile/selection/assemble.ts
```diff
@@ -203,7 +204,7 @@ export function assembleSelectionScaleDomain(model: Model, domainRaw: SignalRef)
   return {signal: 'null'};
 }
 
-function cleanupEmptyOnArray(signals: Signal[]) {
+function cleanupEmptyOnArray(signals: readonly Signal[]) {
```
arvind: Another case here, this function is mutating the objects within the array. So labeling the array as `readonly` feels like it's indicating a true rather than shallow immutability. I worry that it gives us a false sense of security/purity when there is more work to be done actually making things pure.

domoritz: Good catch. I'll remove the annotation here.

src/compile/facet.ts
domoritz: Removed

domoritz:
I cleaned up the PR. Please take another look.

kanitw:
Rebase?

domoritz:
Done

PR #5368, opened 2019-09-03
docs: include equalEarth in the projection.md

mattijn:
This adds the `equalEarth` projection to the `projection.md` [file](https://github.com/mattijn/vega-lite/blob/master/site/docs/projection.md). 

I cannot find any other place where there is made a reference to the projection types. So I guess it should be fine like this.


domoritz:
Since Vega-Lite just passes though through to Vega, we already support `equalEarth`. We import Vega's typings so we already have the projection in the schema as well.

PR #5371, opened 2019-09-06
docs: remove sentence about reformatted array

mhyfritz:
:wave: 

The array is formatted the same way as in the previous example.

Cheers,
Markus


domoritz:
Thanks. We now use prettier to format the specs.

Issue #5370, opened 2019-09-05
Support selection to set bin extent?


jheer:
I would like to support smooth zooming (not filtering) within a histogram. As I zoom in, I'd like the data to re-bin to a smaller step size. In Vega I would use signals to do this, but that is not an option with Vega-Lite. We already allow the use of interval selections to set a scale domain. Could we do something similar for the `bin` extent parameter? This would affect both the `bin` transform and the `bin` encoding directive.

Here is an example of what that might look like (ignoring the selection definition for now):

```json
{
  "mark": "bar",
  "encoding": {
    "x": {"field": "a", "bin": {"extent": {"selection": "interval"}},
    "y": {"aggregate": "count"}
  }
}
```

When the selection is empty I would expect `extent` would default to the normal data-driven extent.

Of course, this approach might still have some wrinkles, for example if bin boundaries jump unexpectedly upon zooming. However, the default use of extent `nice`ing and snapping to human-friendly step sizes (1, 5, 10, etc) might go a long way towards keeping things stable.

jheer:
[...]

PR #5372, opened 2019-09-06
feat: allow bin extent to be driven by a selection

arvind:
Addresses feature request #5370. It requires Vega's bin transform to be extended with a ~~`rawExtent`~~ `span` property (which landed in v5.6).


src/vega.schema.ts
```diff
@@ -329,6 +329,7 @@ export type VgEncodeEntry = {[k in VgEncodeChannel]?: VgValueRef | (VgValueRef &
 export interface VgBinTransform extends BaseBin {
```
domoritz: Let's directly import this type from Vega if we can.

arvind: Done with https://github.com/vega/vega-lite/pull/5372/commits/deee163256b105761e7bc27d9c038fc722704283. There's still more work to do because `BaseBin` is very similar to `BaseBin` in vega-typings but the latter uses `SignalRefs`. Not sure if there's a straightforward way to import/reuse and strip out `SignalRefs`. Perhaps easier to restructure vega-typings' `BaseBin` to parameterize `SignalRefs` out.

src/vega.schema.ts
```diff
@@ -1,6 +1,7 @@
 import {
   AggregateOp,
   Align,
+  BinTransform as VgBinTransform,
```
domoritz: Nice. Thanks!

domoritz:
Could you squash the commits?

arvind:
@domoritz: I'm not sure I understand the schema [test failure](https://github.com/vega/vega-lite/pull/5372/checks?check_run_id=285466523#step:6:15) since the schema was automatically generated... Any help would be appreciated!

domoritz:
We are generating a schema that violates the meta schema. I'll take a look.

domoritz:
The problem was that you cannot set `minItems` and `maxItems` for a reference. In the case here it wasn't even correct or necessary since the type already has the information about the length of the array.

PR #5382, opened 2019-09-13
refactor: set the new transforms data field types as FieldName

marcprux:
The field types in the new transforms are strings, when they should be FieldName to match the field types in the other transforms.


domoritz:
Thank you for the PR. I wonder whether we want to get rid of FieldName or also consistently use it in Vega. What do you think @kanitw @arvind?

marcprux:
Having the `FieldName` as a separate type is extremely useful for a consumer of the schema when generating a strongly-typed model from the JSON schema. Even if it is not actively useful for vega or vega-lite, it is also useful for documentary purposes, since you know that a given string is going to be used as the identifier of a data field.

domoritz:
Okay, then let's make sure Vega is consistent as well.

domoritz:
@kanitw should we add a Vega issue?

kanitw:
@domoritz go ahead.

PR #5385, opened 2019-09-16
chore: simplify the deployment of the site

domoritz:
With this PR, we are now building the page locally and then just push the compiled jekyll page with the gh-pages tool.


domoritz:
`yarn deploy:site` (renamed from `deploy:gh`).

domoritz:
I'll fix the docs (it was using the old command)

PR #5391, opened 2019-09-18
chore: lint commit messages

domoritz:
I understand that we may not want to do this since we often squash PRs. 

Merge after #5390


kanitw:
We can try, and turn it off if this becomes too annoying

PR #5398, opened 2019-09-18
fix: use global replace where necessary, fixes #5392

### LINKED ISSUE #5392: Field name doesn't accept two or more dots
I am having trouble with field names that have two or more dots.

According to the field documentation (https://vega.github.io/vega/docs/types/#Field), _To specify field names that contain dots but are not nested lookups, escape the dot inline ("my\\\\.field") or enclose the field name in brackets ("[my.field]")._

I can do this only with one dot. The following example (**a.b**) works well in the Vega Editor:

```json
{
  "data": {
    "values": [
      {
        "color": "default",
        "a.b": 28
      }
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "color",
      "type": "nominal"
    },
    "y": {
      "field": "[a.b]",
      "type": "quantitative"
    }
  }
}
```

However, I can't get something like **a.b.c** working, neither with
```json
"y": {
      "field": "[a.b.c]",
      "type": "quantitative"
    }
```
nor with 
```json
"y": {
      "field": "a\\.b\\.c",
      "type": "quantitative"
    }
```

Is this how it is supposed to work, or is this a bug? 

Hope I was clear, thanks!

domoritz:


src/util.ts
```diff
@@ -376,10 +376,21 @@ export function flatAccessWithDatum(path: string, datum: 'datum' | 'parent' | 'd
...
 
+/**
+ * Replace all ocurrences of a string with another string.
+ *
+ * @param string the string to replace in
+ * @param find the string to replace
+ * @param replace the replacement
+ */
+export function replaceAll(string: string, find: string, replace: string) {
+  return string.replace(new RegExp(find.replace(/[-/\\^$*+?.()|[\]{}]/g, '\\$&'), 'g'), replace);
```
chanwutk: @domoritz why do we need a RegExp in this replacement?

domoritz: We need to replace every occurrence, not just the first one. However, normal strings need to be escaped.

chanwutk: I see

src/util.ts
```diff
@@ -376,10 +376,21 @@ export function flatAccessWithDatum(path: string, datum: 'datum' | 'parent' | 'd
...
 }
 
+/**
+ * Replace all ocurrences of a string with another string.
+ *
+ * @param string the string to replace in
+ * @param find the string to replace
+ * @param replace the replacement
+ */
+export function replaceAll(string: string, find: string, replace: string) {
```
kanitw: call replacement `replacement`, not `replace`. Otherwise, it's quite confusing.

PR #5426, opened 2019-09-30
feat: add quantile transform, update Vega to 5.7

jheer:
- Updates to Vega 5.7.0
- Add quantile transform
- Add quantile test cases
- Add quantile-quantile plot example

This PR violates the "do one thing" maxim, but the Vega 5.7.0 update is necessary to add and test the quantile transform. If Vega 5.7.0 is added in a separate PR (feel free to follow the changes made here), I will update this one to focus on the Quantile transform only.

p.s. I saw some local test failures, all involving TS linting for example files I did not touch. I haven't tried to address them. (_Update_: these seem to have been repaired by CI tasks...)


domoritz:
I also updated the vega-cli dependency and typings (in addition to the peer dependency).

PR #5438, opened 2019-10-04
feat: Add missing parallels property in projection

mattijn:
This PR adds the missing `parallels` property to Vega-Lite.

The property is available in Vega: https://github.com/vega/vega/blob/master/packages/vega-projection/src/projections.js#L31
But not yet available in Vega-Lite.

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -11041,6 +11041,13 @@
         "parallel": {
           "type": "number"
         },
+        "parallels": {
+          "description": "Sets the desired parallels of the projection",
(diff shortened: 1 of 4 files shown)


mattijn:
I don't know how.

domoritz:
You can just import the type from Vega. We are already doing this in other places.

domoritz:
It also seems to be missing from Vega: https://github.com/vega/vega/issues/2153.

domoritz:
Replaced with https://github.com/vega/vega-lite/pull/5542

Issue #5309, opened 2019-08-17
Make responsive charts


domoritz:
It would be great if we could make charts responsive by listening to changes to the surrounding container. I'm not sure how good the support in Vega is. Also see https://github.com/vega/vega/issues/1324

donghaoren:
Here is a prototype for responsive chart in Vega (edited upon a Vega-Lite output): https://bl.ocks.org/donghaoren/023b2246569e8f0615017507b473e55e

megaroeny:
Thanks for opening this issue. Having one simple property for a responsive chart would be much more user-friendly and scalable so it's easier to share.

keckelt:
Not sure if it is in the scope of this issue, but would be nice if faceted charts could also be responsive. Like this one: [Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoAWOgCtY3AHaQANOCgATdHkw5gUVgCcx2XAWJkhtBszacepZJubcjpbbtKpuFRslh5RRUAX3UISAwjAGszUAgIvABPQmQzSHZ0IzUNRO5uMTxGQjM8I1ZkDTCNSAB3Rk0WdPhlXUYlZGzwqGQlFpclajjcqCNuWuGEhMgAM38xTXSAWXQ5NwB9AHFeozTVKGTU9KVuVHb0U33IWEJ0eHahnAAGbqnIFHRNTux4qemxdDsZBiACCgzEaWerz+kABQLEABlGGc8GYAIwAVheIzecOBIIk1BUOFhyBmqOhuMBwIAYq0AMqMABekLAaIATJT-tSxHSlKiSQBNdD8ozoDBKdAUHJ-aZFPAQsxKVhiMQ4sAhEbVN4AD0mMPQ1Gou1oeFZkBarH5agOKXNAEdWCL5TpGFQZQadYxYHFIMaGqVysgrm5-NaSXhuCVNVNtdMkvq3nNgYsSQBhXauqjrAAqdptkEO5pOZ0ll3VkHQXp9OGVqq1VW6FuUc0eYF+ET8yAm+muZW40XNhPwxuQCbjESr3t9mlO6HaAHUGk1njGNSBNUA)

kanitw:
[...]

PR #5441, opened 2019-10-04
feat: support responsive width and height

donghaoren:
This PR addresses #5309. We add a `container` option for the top-level width and height property. The `container` option can be specified independently so a user can have for example responsive width and fixed height.


site/docs/view/size.md
```diff
@@ -43,6 +43,19 @@ For a discrete axis, specifying a fixed size (e.g., width in the following plot)
 
 <span class="vl-example" data-name="bar_size_explicit_bad"></span>
 
+### Specifying Responsive Width and Height
+
+The top-level `width` and/or `height` property can be set to `container` indicating that the width and/or height of the plot should be the same as its container. The `width` and `height` can be set independently, for example, you can have responsive `width` and fixed `height` by setting `width` to `container` and `height` to a number.
```
domoritz:
```suggestion
The top-level `width` and/or `height` properties can be set to `container` indicating that the width and/or height of the plot should be the same as its surrounding container. The `width` and `height` can be set independently, for example, you can have responsive `width` and fixed `height` by setting `width` to `container` and `height` to a number.
```

src/spec/base.ts
```diff
@@ -72,29 +72,31 @@ export interface LayoutSizeMixins {
    *
    * - For a plot with a continuous x-field, width should be a number.
    * - For a plot with either a discrete x-field or no x-field, width can be either a number indicating a fixed width or an object in the form of `{step: number}` defining the width per discrete step. (No x-field is equivalent to having one discrete step.)
+   * - To enable responsive sizing on width, it should be set to `container`
```
domoritz:
```suggestion
   * - To enable responsive sizing on width, it should be set to `container`.
```

src/compile/layoutsize/assemble.ts
```diff
@@ -50,6 +50,13 @@ export function sizeSignals(model: Model, sizeType: 'width' | 'height'): NewSign
     }
     /* istanbul ignore next: Condition should not happen -- only for warning in development. */
     throw new Error('layout size is step although width/height is not step.');
+  } else if (size == 'container') {
+    const isWidth = name.endsWith('width');
+    const expr = isWidth ? 'containerSize()[0]' : 'containerSize()[1]';
+    const defaultValue = isWidth ? model.config.view.continuousWidth : model.config.view.continuousHeight;
+    // make sure the value is valid
+    const safeExpr = `isValid(${expr}) ? ${expr} : ${defaultValue}`;
```
domoritz: We use `isFinite` in other places. It should behave the same except that it also filters out infinite values. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite.

I think Vega uses `isFinite` rather than `Number.isFinite` right now but I'll send a PR to change this.

donghaoren: Good point! Thanks!

src/log/message.ts
```diff
@@ -26,6 +26,9 @@ export function invalidSpec(spec: GenericSpec<any, any>) {
 
 // FIT
 export const FIT_NON_SINGLE = 'Autosize "fit" only works for single views and layered views.';
+export const CONTAINER_SIZE_NON_SINGLE = 'Width/height "container" only works for single views and layered views.';
```
kanitw: If we know if it is width or height that's problematic, we should make that clear. (We can make this log a function that takes `sizeType: 'width' | 'height'` as an input.)

src/spec/base.ts
```diff
@@ -72,29 +72,31 @@ export interface LayoutSizeMixins {
    *
    * - For a plot with a continuous x-field, width should be a number.
    * - For a plot with either a discrete x-field or no x-field, width can be either a number indicating a fixed width or an object in the form of `{step: number}` defining the width per discrete step. (No x-field is equivalent to having one discrete step.)
+   * - To enable responsive sizing on width, it should be set to `container`.
```
kanitw: Please use quotes inside backticks for string values

```
`"container"`
```

(We can omit quotes for property name just like in JS)

src/spec/base.ts
```diff
@@ -72,29 +72,31 @@ export interface LayoutSizeMixins {
...
    */
-  width?: number | Step;
+  width?: number | 'container' | Step;
 
   /**
    * The height of a visualization.
    *
    * - For a plot with a continuous y-field, height should be a number.
    * - For a plot with either a discrete y-field or no y-field, height can be either a number indicating a fixed height or an object in the form of `{step: number}` defining the height per discrete step. (No y-field is equivalent to having one discrete step.)
+   * - To enable responsive sizing on height, it should be set to `container`.
```
kanitw: Please use quotes inside backticks for string values

```
`"container"`
```

(We can omit quotes for property name just like in JS)

src/spec/base.ts
```diff
@@ -72,29 +72,31 @@ export interface LayoutSizeMixins {
...
    *
    * __Default value:__ Based on `config.view.continuousHeight` for a plot with a continuous y-field and `config.view.discreteHeight` otherwise.
    *
-   * __Note:__ For plots with [`row` and `column` channels](https://vega.github.io/vega-lite/docs/encoding.html#facet), this represents the height of a single view.
+   * __Note:__ For plots with [`row` and `column` channels](https://vega.github.io/vega-lite/docs/encoding.html#facet), this represents the height of a single view and the `container` option cannot be used.
```
kanitw: Please use quotes inside backticks for string values

```
`"container"`
```

(We can omit quotes for property name just like in JS)

src/spec/base.ts
domoritz: Good catch.

src/log/message.ts
donghaoren: Actually this isn't being used right now. Is there a function to determine if the view is single or layered?

donghaoren:
I haven't got codecov/patch and codecov/project to pass, how should I fix them?

src/log/message.ts
donghaoren: I tried `isLayerSpec(spec) || isUnitSpec(spec)` like in the `FIT_NON_SINGLE` warning, but it didn't work for column/row encodings.

(btw, this should be a separate issue: the `FIT_NON_SINGLE` warning doesn't trigger for column/row encodings, and the chart won't render. To reproduce, you can add `autosize: "fit"` to https://vega.github.io/editor/#/examples/vega-lite/trellis_stacked_bar)

domoritz:
They are not required checks but nice to have. You have to add tests that cover code that is not yet covered by existing tests. 

See for example https://codecov.io/gh/vega/vega-lite/compare/b878a3b08f48aa0df131820eaa70ed32344c2232...20389f859382c8f48049ef9a2a06869e217047be/src/src/spec/toplevel.ts

You can add a test that makes sure that the message is logged. We have tests like this in the code already if you want to know how to test log messages.

domoritz:
I've just merged white backgrounds into master. Can you merge master into this branch before we merge?

donghaoren:
@domoritz I finished the refactor, this PR is ready for another review

donghaoren:
Note that because I made normalize output `autosize`, many tests has to be changed, and there's also a bunch of updates to the generated normalized example json files.

examples/specs/normalized/airport_connections_normalized.vl.json
```diff
@@ -90,5 +90,6 @@
   ],
   "width": 900,
   "height": 500,
-  "config": {"view": {"stroke": null}}
+  "config": {"view": {"stroke": null}},
+  "autosize": {"type": "pad"}
```
domoritz: Do we need to add this or is it already the default in Vega? It would be good to reduce churn.

donghaoren: Done! Thank for the suggestion.

src/compile/compile.ts
```diff
@@ -122,9 +121,16 @@ export function compile(inputSpec: TopLevelSpec, opt: CompileOptions = {}) {
...
-function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config, autosize: AutoSizeParams, model: Model) {
+function getTopLevelProperties(
+  inputSpec: TopLevel<any>,
+  autosize: AutoSizeType | AutoSizeParams,
+  config: Config,
+  model: Model
+) {
   const width = model.component.layoutSize.get('width');
   const height = model.component.layoutSize.get('height');
+  if (autosize === undefined) autosize = 'pad';
```
domoritz:
```suggestion
  if (autosize === undefined) {
    autosize = 'pad';
  }
```

src/compile/compile.ts
```diff
@@ -122,9 +121,16 @@ export function compile(inputSpec: TopLevelSpec, opt: CompileOptions = {}) {
...
+  if (isString(autosize)) autosize = {type: autosize};
```
domoritz:
```suggestion
  else if (isString(autosize)) {
    autosize = {type: autosize};
  }
```

src/compile/layoutsize/assemble.ts
```diff
@@ -50,6 +50,13 @@ export function sizeSignals(model: Model, sizeType: 'width' | 'height'): NewSign
     }
     /* istanbul ignore next: Condition should not happen -- only for warning in development. */
     throw new Error('layout size is step although width/height is not step.');
+  } else if (size == 'container') {
+    const isWidth = name.endsWith('width');
+    const expr = isWidth ? 'containerSize()[0]' : 'containerSize()[1]';
+    const defaultValue = isWidth ? model.config.view.continuousWidth : model.config.view.continuousHeight;
+    // make sure the value is valid
```
domoritz:
```suggestion
```

src/normalize/index.ts
```diff
@@ -1,28 +1,121 @@
...
+  config?: Config
+): TopLevel<NormalizedSpec> & LayoutSizeMixins {
   if (config === undefined) {
     config = initConfig(spec.config);
   }
 
-  return normalizeGenericSpec(spec, config);
+  let normalizedSpec = normalizeGenericSpec(spec, config);
+
+  // Normalize autosize and deal with width or height == "container"
```
domoritz:
```suggestion
```

src/normalize/index.ts
```diff
@@ -1,28 +1,121 @@
...
+function normalizeGenericSpec(spec: GenericSpec<UnitSpec, ExtendedLayerSpec> | FacetedUnitSpec, config: Config = {}) {
   return normalizer.map(spec, {config});
 }
 
+/** Helper function for normalizeAutoSize */
+function _normalizeAutoSize(autosize: AutosizeType | AutoSizeParams) {
+  return isString(autosize) ? {type: autosize} : autosize || {};
+}
+
+export function normalizeAutoSize(
```
domoritz:
```suggestion
/**
 * Normalize autosize and deal with width or height == "container".
 */
export function normalizeAutoSize(
```

src/normalize/index.ts
```diff
@@ -1,28 +1,121 @@
...
  */
-function normalizeGenericSpec(
-  spec: GenericSpec<UnitSpec, ExtendedLayerSpec> | FacetedUnitSpec,
-  config: Config = {}
-): NormalizedSpec {
+function normalizeGenericSpec(spec: GenericSpec<UnitSpec, ExtendedLayerSpec> | FacetedUnitSpec, config: Config = {}) {
   return normalizer.map(spec, {config});
 }
 
+/** Helper function for normalizeAutoSize */
```
domoritz:
```suggestion
```

src/compile/compile.ts
```diff
@@ -144,9 +153,13 @@ function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config, auto
   }
 
   return {
-    autosize: keys(autosize).length === 1 && autosize.type ? autosize.type : autosize,
+    ...(keys(autosize).length === 1 && autosize.type
+      ? autosize.type == 'pad'
```
kanitw: use `===` ?

src/normalize/index.ts
```diff
@@ -1,28 +1,122 @@
...
+  if (width == 'container' && !(autosize.type == 'fit' || autosize.type == 'fit-x')) {
+    log.warn(log.message.containerSizeNotCompatibleWithAutosize('width'));
+  }
+  if (height == 'container' && !(autosize.type == 'fit' || autosize.type == 'fit-y')) {
+    log.warn(log.message.containerSizeNotCompatibleWithAutosize('height'));
+  }
+
+  const result: TopLevel<NormalizedSpec> & LayoutSizeMixins = {...spec, autosize: autosize};
+  if (isFitCompatible) {
+    if (result.width === undefined && width !== undefined) result.width = width;
```
kanitw: Following Google's convention, Let's avoid one line if's.  


```
if (condition) {
  doSomething()
}
```

src/normalize/index.ts
```diff
@@ -1,28 +1,122 @@
...
+  config?: Config
+): TopLevel<NormalizedSpec> & LayoutSizeMixins {
   if (config === undefined) {
     config = initConfig(spec.config);
   }
 
-  return normalizeGenericSpec(spec, config);
+  let normalizedSpec = normalizeGenericSpec(spec, config);
+
+  normalizedSpec = normalize:(
```
kanitw: Why is there a colon here?

src/normalize/index.ts
```diff
@@ -27,7 +27,7 @@ export function normalize(
 
   let normalizedSpec = normalizeGenericSpec(spec, config);
 
-  normalizedSpec = normalizeAutoSize(
+  normalizedSpec = normalize:(
```
domoritz: Should this have a `:`?

src/compile/compile.ts
```diff
@@ -144,9 +153,13 @@ function getTopLevelProperties(topLevelSpec: TopLevel<any>, config: Config, auto
   }
 
   return {
-    autosize: keys(autosize).length === 1 && autosize.type ? autosize.type : autosize,
+    ...(keys(autosize).length === 1 && autosize.type
```
kanitw: Can we avoid nested ?

src/normalize/index.ts
```diff
@@ -1,28 +1,126 @@
...
+    ...(config ? _normalizeAutoSize(config.autosize) : {}),
+    ..._normalizeAutoSize(spec.autosize)
+  };
+
+  if (autosize.type === 'fit' && !isFitCompatible) {
+    log.warn(log.message.FIT_NON_SINGLE);
+    autosize.type = 'pad';
+  }
+
+  if (width == 'container' && !(autosize.type == 'fit' || autosize.type == 'fit-x')) {
```
dlaliberte, line 101: These warnings seem superfluous.  Using 'container' for the width and height seems to be the best way to be responsive to the container size.   And using autosize.type == 'none' is the only way (?) to specify the padding exactly, rather than adding any auto-computed padding based on what is drawn.  This is necessary to get predictable alignment across charts that are drawn separately.  So we would like to use this combination without getting log warnings.  Is that possible?

dlaliberte (commented):
The 'container' mode for responsive sizing is great, but...

src/normalize/index.ts
domoritz: Definitely. Wouldn't the issues be that the responsive size does not include the padding and therefore the chart is always too wide? If not, then I agree that we should remove this warning.

dlaliberte: Indeed, we would want contains: 'padding' to include the padding in the total.   So the content.width should be container.width - (padding.left + padding.right).  I don't really know if type: 'none' works this way, however.

```
 width: 'container',
 height: 'container',
 padding: { left: 40, right: 0, top: 5, bottom: 20 },
 autosize: { type: 'none', contains: 'padding' }
```

domoritz: It might not in Vega but we can generate the correct output. So if the user specifies autosize none, we can automatically change the container size computation to take the padding into account.

Happy to review a pull request towards this. I think the current behavior warrants a warning.

PR #5445, opened 2019-10-06
fix: set background to white by default

### LINKED ISSUE #5442: Set default background color to white
* Guarantees that charts are legible on any background
* Marks charts usable on slack and twitter (the default background isn't good)

domoritz:
fixes #5442


domoritz:
I did not close this.

PR #5463, opened 2019-10-14
fix: optimizer DuplicateTimeNode should remove bottom-most node first

arvind:
I suspect this will resolve several of the bugs with selections projected over time units. 

The `RemoveDuplicateTimeNode` optimizer is a `BottomUpOptimizer` but, as previously written, would preserve the bottom-most time unit node and consider ones further upstream to be duplicates. As a result, you would get weird orderings like the one found in [interactive_area_brush](https://vega.github.io/editor/#/examples/vega-lite/interactive_area_brush) where the selection filter would come _before_ the time unit it depends on:

```json
"transform": [
  {
    "type": "filter",
    "expr": "!(length(data(\"brush_store\"))) || (vlSelectionTest(\"brush_store\", datum))"
  },
  {
    "type": "formula",
    "as": "yearmonth_date",
    "expr": "datetime(year(datum[\"date\"]), month(datum[\"date\"]), 1, 0, 0, 0, 0)"
  }
```

With the above ordering, initializing the selection would yield buggy behavior as, on the first run of the dataflow, `yearmonth_date` would be undefined when the selection predicate test would occur. 

In this PR, we continue to optimize from the bottom up and remove the bottom-most node when we encounter a duplicate further upstream. As a result, this should often allow time unit nodes to be pushed to the source dataset and reduce the amount of time unit recomputation in derived datasets.

### DIFF
--- a/examples/compiled/interactive_area_brush.vg.json
+++ b/examples/compiled/interactive_area_brush.vg.json
@@ -10,16 +10,22 @@
     {
       "name": "source_0",
       "url": "data/unemployment-across-industries.json",
-      "format": {"type": "json", "parse": {"date": "date"}}
+      "format": {"type": "json", "parse": {"date": "date"}},
+      "transform": [
+        {
+          "type": "formula",
+          "as": "yearmonth_date",
+          "expr": "datetime(year(datum[\"date\"]), month(datum[\"date\"]), 1, 0, 0, 0, 0)"
+        }
+      ]
     },
     {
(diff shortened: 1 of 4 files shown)


arvind:
P.S. The grammatical structure of this commit is weird because if I put the optimizer name first (i.e., `DuplicateTimeNode optimizer should remove bottom-most node first`), the pre-commit hook complained about letter casing.

PR #5473, opened 2019-10-15
feat: add support for multiline text and subtitle

### LINKED ISSUE #4055: Graph subtitle
Vega combined with altair is fantastic to quickly create great looking charts! One small feature that I am missing is the possibility to have a subtitle for a graph. More specifically, to be able to regenerate a plot in  [538 style](https://www.dataquest.io/blog/making-538-plots/) I think this would be the only main feature missing.

### LINKED ISSUE #5431: Support Vega 5.7 properties
https://github.com/vega/vega/pull/2022

* [x] subtitles
* [x] multi-line text
* [x] symbolLimit
* [x] deprecations

domoritz:
Fixes #5431. Fixes #4055.


src/compile/common.ts
```diff
@@ -210,26 +213,25 @@ export function mergeTitleFieldDefs(f1: readonly FieldDefBase<string>[], f2: rea
...
+  if (!isText(v1.value) && !isText(v2.value)) {
     return {
       explicit: v1.explicit,
       value: mergeTitleFieldDefs(v1.value, v2.value)
     };
-  } else if (!isArray(v1.value) && !isArray(v2.value)) {
+  } else if (isText(v1.value) && isText(v2.value)) {
     return {
       explicit: v1.explicit, // keep the old explicit
       value: mergeTitle(v1.value, v2.value)
```
haldenl: Is it not possible to merge `null` and "title"? If so, we need an else case?

domoritz: Good catch. Fixed

PR #5474, opened 2019-10-15
fix: apply background only once

### LINKED ISSUE #5466: The background color should only be set in the config or at the top level
We are currently setting it both at the top-level and in the config that is part of the compiled spec.

domoritz:
fixes #5466

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -210,7 +210,7 @@ export interface Config
    */
   projection?: ProjectionConfig;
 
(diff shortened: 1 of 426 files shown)


src/config.ts
```diff
@@ -373,6 +374,8 @@ export function stripAndRedirectConfig(config: Config) {
   // affect header labels, which also uses `title` directive to implement.
   redirectConfig(config, 'title', 'group-title');
 
+  // Remove background config since we apply it to the spec.
```
kanitw: Why is this comment here, not above in `VL_ONLY_CONFIG_PROPERTIES`?

domoritz: Good catch. Fixed.

PR #5488, opened 2019-10-24
chore: update for Vega 5.7.3

domoritz:


domoritz:
@jheer Can you explain the changes in the SVGs here? 

Screen Shot 2019-10-24 at 16 31 52

domoritz:
Since we bootstrap the intervals, the random state affects what the CIs look like. Something about the randomness must have changed.

PR #5497, opened 2019-10-28
fix: ensure compatibility with TS 3.7

domoritz:
Needed for https://github.com/jupyterlab/jupyterlab/pull/7428.

cc @jasongrout

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "terser": "^4.3.9",
     "ts-jest": "^24.1.0",
     "ts-json-schema-generator": "^0.55.0",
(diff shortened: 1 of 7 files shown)


jasongrout:
Thanks!

jasongrout:
Do you have a timeline for release? (not to hurry or pressure you, just wanting to know when I can check back and resume work on  jupyterlab/jupyterlab#7428).

domoritz:
I'm hoping to get Vega-Lite 4 out in November so we don't block you. We can make a beta release anytime.

PR #5522, opened 2019-11-06
docs: fix link to geographic projection examples

raphodn:


domoritz:
Thanks. Can we fix the target instead since it says maps twice?

raphodn:
@domoritz sorry about that went (much) too fast. should be fixed now. thanks!

domoritz:
Can you add a custom id that doesn’t change when the title changes? Just to be future proof.

Issue #5477, opened 2019-10-16
Enable strictNullCheck


kristw:
Please:

- [x] Check for duplicate issues. Please file separate requests as separate issues on GitHub.
- [x] Describe how to reproduce the bug / the goal of the feature request.

Try to build `react-vega`.

- [x] Use the latest versions of Vega and Vega-Lite.

Yes

- [X] If applicable, include error messages and screenshots, GIF videos (e.g. using https://www.cockos.com/licecap/), or working examples (e.g. by clicking share in the [Vega-Editor](https://vega.github.io/editor/) or https://bit.ly/vega-lite-blocks)

```
../../node_modules/vega-lite/build/src/axis.d.ts:20:81 - error TS2344: Type 'BaseAxisNoSignals[k] | null' does not satisfy the constraint 'string | number | boolean | number[] | null'.
  Type 'BaseAxisNoSignals[k]' is not assignable to type 'string | number | boolean | number[] | null'.
    Type 'string | number | number[] | null | undefined' is not assignable to type 'string | number | boolean | number[] | null'.
      Type 'undefined' is not assignable to type 'string | number | boolean | number[] | null'.
        Type 'BaseAxisNoSignals[k]' is not assignable to type 'number[]'.
          Type 'string | number | number[] | null | undefined' is not assignable to type 'number[]'.
            Type 'undefined' is not assignable to type 'number[]'.

20     [k in ConditionalAxisProp]?: BaseAxisNoSignals[k] | ConditionalAxisProperty<BaseAxisNoSignals[k] | null>;
```

domoritz:
We don’t have strict null checks enabled in our code base. I’ll change this to a feature request.

domoritz:
We can't enable it yet. But the typings work with strict null checks.

domoritz:
It would be good to enable strict null checks. See https://www.figma.com/blog/inside-figma-a-case-study-on-strict-null-checks/ for a discussion.

PR #5535, opened 2019-11-10
feat: more changes to prepare for strictNullCheck support

kristw:
Related to #5477

### DIFF
--- a/src/aggregate.ts
+++ b/src/aggregate.ts
@@ -61,11 +61,11 @@ export function isAggregateOp(a: string | ArgminDef | ArgmaxDef): a is Aggregate
 
 export const COUNTING_OPS: NonArgAggregateOp[] = ['count', 'valid', 'missing', 'distinct'];
 
-export function isCountingAggregateOp(aggregate: string | Aggregate): boolean {
+export function isCountingAggregateOp(aggregate?: string | Aggregate): boolean {
   return isString(aggregate) && contains(COUNTING_OPS, aggregate);
 }
 
-export function isMinMaxOp(aggregate: Aggregate | string): boolean {
+export function isMinMaxOp(aggregate?: Aggregate | string): boolean {
   return isString(aggregate) && contains(['min', 'max'], aggregate);
 }
 
--- a/src/aggregate.ts
+++ b/src/aggregate.ts
@@ -62,11 +62,11 @@ export function isAggregateOp(a: string | ArgminDef | ArgmaxDef): a is Aggregate
 export const COUNTING_OPS: NonArgAggregateOp[] = ['count', 'valid', 'missing', 'distinct'];
 
 export function isCountingAggregateOp(aggregate: string | Aggregate): boolean {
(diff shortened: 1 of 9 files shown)


src/aggregate.ts
```diff
@@ -62,11 +62,11 @@ export function isAggregateOp(a: string | ArgminDef | ArgmaxDef): a is Aggregate
 export const COUNTING_OPS: NonArgAggregateOp[] = ['count', 'valid', 'missing', 'distinct'];
 
 export function isCountingAggregateOp(aggregate: string | Aggregate): boolean {
-  return aggregate && isString(aggregate) && contains(COUNTING_OPS, aggregate);
+  return isString(aggregate) && contains(COUNTING_OPS, aggregate);
 }
 
 export function isMinMaxOp(aggregate: Aggregate | string): boolean {
-  return aggregate && isString(aggregate) && contains(['min', 'max'], aggregate);
```
domoritz: Shouldn't this be

```ts
export function isCountingAggregateOp(aggregate?: string | Aggregate): boolean {
  return isString(aggregate) && contains(COUNTING_OPS, aggregate);
}

export function isMinMaxOp(aggregate?: Aggregate | string): boolean {
  return isString(aggregate) && contains(['min', 'max'], aggregate);
}
```

src/scale.ts
```diff
@@ -406,7 +406,7 @@ export type Domain = number[] | string[] | boolean[] | DateTime[] | 'unaggregate
 export type Scheme = string | SchemeParams;
 
 export function isExtendedScheme(scheme: string | SchemeParams): scheme is SchemeParams {
-  return scheme && !!scheme['name'];
+  return typeof scheme !== 'string' && !!scheme['name'];
```
domoritz:
```suggestion
  return !isString(scheme) && !!scheme['name'];
```

src/util.ts
```diff
@@ -282,7 +282,7 @@ export function deleteNestedProperty(obj: any, orderedProps: string[]) {
   if (orderedProps.length === 0) {
     return true;
   }
-  const prop = orderedProps.shift();
+  const prop = orderedProps.shift() as string;
```
domoritz: I think this is cleaner

```suggestion
  const prop = orderedProps.shift()!;
```

src/util.ts
```diff
@@ -282,7 +282,7 @@ export function deleteNestedProperty(obj: any, orderedProps: string[]) {
   if (orderedProps.length === 0) {
     return true;
   }
-  const prop = orderedProps.shift();
+  const prop = orderedProps.shift()!;
```
domoritz: This violates our linting rules. We should probably change the linting rules or add an exception here.

```suggestion
  const prop = orderedProps.shift()!; // eslint-disable-line @typescript-eslint/no-non-null-assertion
```

What do you think is better? 

Btw, I wish we could figure out a way to run the full test suite on forks but somehow it doesn't work yet. You can make branches in our repo as `krw/*` and send PRs from there.

domoritz:
Thank you.

PR #5544, opened 2019-11-13
feat: add support for new corner radius properties

domoritz:


src/vega.schema.ts
```diff
@@ -320,7 +320,14 @@ export type VgEncodeChannel =
   | 'href'
   | 'cursor'
   | 'defined'
-  | 'cornerRadius';
+  | 'cornerRadius'
+  | 'cornerRadiusTopLeft'
+  | 'cornerRadiusTopRight'
+  | 'cornerRadiusBottomRight'
+  | 'cornerRadiusBottomLeft'
+  | 'scaleX'
```
donghaoren: Are scaleX and scaleY related to this PR? or this file is from Vega?

domoritz: I noticed that they are missing. Not related to this pull request.

PR #5559, opened 2019-11-17
refactor: use null coalescing

domoritz:
Learn more about null coalescing at https://github.com/tc39/proposal-nullish-coalescing.


domoritz:
It’s in typescript so we can use it. We will compile to compatible JavaScript.

domoritz:
I think typescript only accepts stage 3 proposals.

PR #5577, opened 2019-11-20
chore: use vega timeunit transform

haldenl:
Use Vega timeunit instead of converting to less flexible formulas. This also fixes a bug where Vega-Lite wouldn't show February 29th on leap years.

**Before**
before

**After**
after

(The 'gap' is from a small temp range on the one 29th of Feb recorded, from 1.1 to 5.0 degrees C).


test-runtime/interval.test.ts
```diff
@@ -155,8 +155,8 @@ describe('interval selections at runtime in unit views', () => {
     await embed(spec('unit', 1, {type, encodings: ['x']}, {values, x: {type: 'temporal', timeUnit: 'day'}}));
 
     extents = [
-      [1136190528000, 1136361600000],
-      [1136449728000, 1136535264000]
```
haldenl: This is because timeunit was using 2006 to calculate days of week.

test-runtime/interval.test.ts
```diff
@@ -155,8 +155,8 @@ describe('interval selections at runtime in unit views', () => {
...
+      [1325492928000, 1325664000000],
+      [1325752128000, 1325837664000]
```
haldenl: Changed b/c Vega-Lite was using 2006 to calculate days of week.

domoritz:
Thank you @haldenl

PR #5583, opened 2019-11-23
feat: support corner radius for stacked bars

### LINKED ISSUE #5545: Support rounded corners for stacked bars
By using clip and putting rectangles on the same bar in the same group.

Needs https://github.com/vega/vega/issues/2155

donghaoren:
For stacked bars, make the cornerRadius properties apply to the entire stack instead of individual bars. To do so, we use group marks for stacks, and clip the bars within groups.

Fixes #5545


examples/compiled/bar_diverging_stack_population_pyramid.vg.json
```diff
@@ -49,19 +49,69 @@
...
+              "type": "rect",
+              "style": ["bar"],
+              "from": {"data": "stack_group_main"},
+              "encode": {
+                "update": {
+                  "fill": {"scale": "color", "field": "gender"},
+                  "x": {"scale": "x", "field": "sum_signed_people_end"},
+                  "x2": {"scale": "x", "field": "sum_signed_people_start"},
+                  "height": {"field": {"group": "height"}},
+                  "cornerRadiusTopLeft": {"value": 0},
```
domoritz: Let's try not to generate default properties.

examples/compiled/bar_multi_values_per_categories.vg.json
```diff
@@ -51,20 +51,65 @@
   ],
```
domoritz: The SVG changed here in odd ways. There is now a while line around the bar.

src/vega.schema.ts
```diff
@@ -689,6 +689,15 @@ const VG_MARK_CONFIG_INDEX: Flag<keyof BaseMarkConfig> = {
 
 export const VG_MARK_CONFIGS = keys(VG_MARK_CONFIG_INDEX);
 
+// Vega's cornerRadius channels.
+export const VG_CORNERRADIUS_CHANNELS = keys({
+  cornerRadius: 1,
```
domoritz: Why is this an object? We usually only use the set pattern (using 1 as value) with `Flag` since it lets us typecheck against keys of objects.

src/compile/mark/mark.ts
```diff
@@ -83,6 +85,117 @@ function parsePathMark(model: UnitModel) {
...
+      cornerRadiusTopRight: {value: 0},
+      cornerRadiusBottomLeft: {value: 0},
+      cornerRadiusBottomRight: {value: 0}
+    };
+
+    // For bin we have to add bin channels.
+    const groupby: string[] = model.vgField(model.stack.groupbyChannel)
+      ? [model.vgField(model.stack.groupbyChannel)]
+      : [];
+    if (model.fieldDef(model.stack.groupbyChannel) && model.fieldDef(model.stack.groupbyChannel).bin) {
```
domoritz:
```suggestion
    if (model.fieldDef(model.stack.groupbyChannel).?.bin) {
```

src/compile/mark/mark.ts
```diff
@@ -83,6 +85,117 @@ function parsePathMark(model: UnitModel) {
...
+
+    const stackField = (opt: FieldRefOption = {}) => model.vgField(model.stack.fieldChannel, opt);
+    const stackFieldGroup = (func: 'min' | 'max', opt: FieldRefOption) => {
+      const vgFieldMinMax = [
+        stackField({prefix: 'min', suffix: 'start', ...opt}),
+        stackField({prefix: 'max', suffix: 'start', ...opt}),
+        stackField({prefix: 'min', suffix: 'end', ...opt}),
+        stackField({prefix: 'max', suffix: 'end', ...opt})
+      ];
+      return func + '(' + vgFieldMinMax.map(field => `scale('${fieldScale}',${field})`).join(',') + ')';
```
domoritz: Let's use string templates for the full string.

src/compile/mark/mark.ts
```diff
@@ -83,6 +85,117 @@ function parsePathMark(model: UnitModel) {
   }
 }
 
+const STACK_GROUP_PREFIX = 'stack_group_';
+
+function getStackGroups(model: UnitModel) {
```
domoritz: Can you add a comment to explain what the logic is here and that we need this code to support rounded corners for stacked bars?

examples/compiled/bar_multi_values_per_categories.vg.json
donghaoren: Actually there was a white line around the bar before, but now it's cut into half.

donghaoren: Because the mask clips everything, which include the part of the stroke outside.
This'll be a difficult one. I'm almost thinking we should apply the mask only when cornerRadius is used. Any thoughts?

donghaoren: Also, it's a bit weird that stroke applies to the individual bars, but cornerRadius applies to the entire stack.

src/vega.schema.ts
donghaoren: If it' an array, the type will be `string[]`, which causes type check issues later in `getMarkConfig`.

examples/compiled/bar_multi_values_per_categories.vg.json
domoritz: It's somewhat weird but probably what people would expect. Also, you cannot easily set the stroke on the whole bar but also color it for each segment.

If this becomes an issue in the future, we could add a property that says whether stacked bars should be in a single group or not.

src/vega.schema.ts
domoritz: You can say `export const VG_CORNERRADIUS_CHANNELS = ['foo', 'bar'] as const`.

examples/compiled/bar_multi_values_per_categories.vg.json
donghaoren: I'm trying to add the stroke to group, still having some issues (see this image), so let's not merge this one yet.

donghaoren: Vega is adding a 0.5px offset to group strokes (probably to align the plot areas to pixel boundaries). I opened an issue here: https://github.com/vega/vega/issues/2186.

domoritz: You fixed this, right?

donghaoren: Yes, I'm using a workaround --- add another layer for the stroke if specified.

domoritz: We can now remove the workaround, right?

donghaoren: We have to wait until Vega fixes https://github.com/vega/vega/issues/2186 and https://github.com/vega/vega/issues/2197. Thank you for opening https://github.com/vega/vega/issues/2197, that's also necessary.

examples/compiled/facet_custom.vg.json
```diff
@@ -134,19 +134,65 @@
...
+              "clip": {"value": true}
+            }
+          },
+          "marks": [
+            {
+              "type": "group",
+              "encode": {
+                "update": {
+                  "y": {
+                    "signal": "-min(scale('y',parent[\"min_sum_people_start\"]),scale('y',parent[\"max_sum_people_start\"]),scale('y',parent[\"min_sum_people_end\"]),scale('y',parent[\"max_sum_people_end\"]))"
```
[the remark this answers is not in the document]
donghaoren: If would be nice if Vega can somehow separate the group translation and clip rectangle. Right now `x`, `y` serve as both translation and a corner of the clip rectangle.

examples/compiled/facet_custom.vg.json
```diff
@@ -134,19 +134,65 @@
       },
       "marks": [
         {
-          "name": "child_marks",
-          "type": "rect",
-          "style": ["bar"],
-          "from": {"data": "facet"},
+          "type": "group",
```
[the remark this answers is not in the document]
donghaoren: For `x` we can; for `y` it's trickier, because `y == 0` is at the top of the bar, so we'll still need this logic.

examples/compiled/bar_multi_values_per_categories.vg.json
domoritz: I'm looking into fixing a bunch of config typings right now. Could you try to send a pull request to fix https://github.com/vega/vega/issues/2197?

examples/compiled/facet_custom.vg.json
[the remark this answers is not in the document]
domoritz: I filed https://github.com/vega/vega/issues/2207

examples/compiled/facet_custom.vg.json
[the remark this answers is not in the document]
domoritz: Good point. I'll leave it up to you to decide whether it's worth the implementation overhead.

examples/compiled/bar_multi_values_per_categories.vg.json
donghaoren: Done! Let's make this WIP until Vega fixes the issue.

domoritz:
Vega 5.9 is out.

donghaoren:
I found another solution (haven't implemented yet), here is an example spec:

```json
{
  "$schema": "https://vega.github.io/schema/vega/v5.json",
  "background": "white",
  "padding": 5,
  "height": 200,
  "style": "cell",
  "data": [
    {
      "name": "source_0",
      "format": {"type": "csv", "parse": {"date": "date"}, "delimiter": ","},
      "transform": [
        {
          "field": "date",
          "type": "timeunit",
          "units": ["month"],
          "as": ["month_date", "month_date_end"]
        },
        {
          "type": "aggregate",
          "groupby": ["month_date", "weather"],
          "ops": ["count"],
          "fields": [null],
          "as": ["__count"]
        },
        {
          "type": "stack",
          "groupby": ["month_date"],
          "field": "__count",
          "sort": {"field": ["weather"], "order": ["descending"]},
          "as": ["__count_start", "__count_end"],
          "offset": "zero"
        }
      ]
    }
  ],
  "signals": [
    {"name": "x_step", "value": 20},
    {
      "name": "width",
      "update": "bandspace(domain('x').length, 0.1, 0.05) * x_step"
    }
  ],
  "marks": [
    {
      "type": "group",
      "from": {
        "facet": {
          "data": "source_0",
          "name": "stack_group_main",
          "groupby": ["month_date"],
          "aggregate": {
            "fields": [
              "__count_start",
              "__count_start",
              "__count_end",
              "__count_end"
            ],
            "ops": ["min", "max", "min", "max"]
          }
        }
      },
      "encode": {
        "update": {
          "x": {"scale": "x", "field": "month_date"},
          "width": {"scale": "x", "band": true}
        }
      },
      "marks": [
        {
          "type": "group",
          "encode": {
            "update": {
              "cornerRadius": {"value": 10},
              "y": {
                "signal": "min(scale('y',parent[\"min___count_start\"]),scale('y',parent[\"max___count_start\"]),scale('y',parent[\"min___count_end\"]),scale('y',parent[\"max___count_end\"]))"
              },
              "y2": {
                "signal": "max(scale('y',parent[\"min___count_start\"]),scale('y',parent[\"max___count_start\"]),scale('y',parent[\"min___count_end\"]),scale('y',parent[\"max___count_end\"]))"
              },
              "clip": {"value": true},
              "width": {"field": {"group": "width"}}
            }
          },
          "marks": [
            {
              "type": "group",
              "encode": {
                "update": {
                  "y": {"field": {"group": "y"}, "mult": -1},
                  "width": {"field": {"group": "width"}}
                }
              },
              "marks": [
                {
                  "name": "marks",
                  "type": "rect",
                  "style": ["bar"],
                  "from": {"data": "stack_group_main"},
                  "encode": {
                    "update": {
                      "stroke": {"value": "black"},
                      "fill": {"scale": "color", "field": "weather"},
                      "width": {"field": {"group": "width"}},
                      "y": {"scale": "y", "field": "__count_end"},
                      "y2": {"scale": "y", "field": "__count_start"}
                    }
                  }
                }
              ]
            }
          ]
        },
        {
          "type": "rect",
          "encode": {
            "update": {
              "cornerRadius": {"value": 10},
              "y": {
                "signal": "min(scale('y',parent[\"min___count_start\"]),scale('y',parent[\"max___count_start\"]),scale('y',parent[\"min___count_end\"]),scale('y',parent[\"max___count_end\"]))"
              },
              "y2": {
                "signal": "max(scale('y',parent[\"min___count_start\"]),scale('y',parent[\"max___count_start\"]),scale('y',parent[\"min___count_end\"]),scale('y',parent[\"max___count_end\"]))"
              },
              "width": {"field": {"group": "width"}},
              "fill": {"value": null},
              "stroke": {"value": "black"}
            }
          }
        }
      ]
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "band",
      "domain": {"data": "source_0", "field": "month_date", "sort": true},
      "range": {"step": {"signal": "x_step"}},
      "paddingInner": 0.1,
      "paddingOuter": 0.05
    },
    {
      "name": "y",
      "type": "linear",
      "domain": {
        "data": "source_0",
        "fields": ["__count_start", "__count_end"]
      },
      "range": [{"signal": "height"}, 0],
      "nice": true,
      "zero": true
    },
    {
      "name": "color",
      "type": "ordinal",
      "domain": ["sun", "fog", "drizzle", "rain", "snow"],
      "range": ["#e7ba52", "#c7c7c7", "#aec7e8", "#1f77b4", "#9467bd"]
    }
  ],
  "axes": [
    {
      "scale": "y",
      "orient": "left",
      "gridScale": "x",
      "grid": true,
      "tickCount": {"signal": "ceil(height/40)"},
      "domain": false,
      "labels": false,
      "maxExtent": 0,
      "minExtent": 0,
      "ticks": false,
      "zindex": 0
    },
    {
      "scale": "x",
      "orient": "bottom",
      "grid": false,
      "title": "Month of the year",
      "labelAlign": "right",
      "labelAngle": 270,
      "labelBaseline": "middle",
      "labelOverlap": true,
      "encode": {
        "labels": {
          "update": {"text": {"signal": "timeFormat(datum.value, '%b')"}}
        }
      },
      "zindex": 0
    },
    {
      "scale": "y",
      "orient": "left",
      "grid": false,
      "title": "Count of Records",
      "labelOverlap": true,
      "tickCount": {"signal": "ceil(height/40)"},
      "zindex": 0
    }
  ],
  "legends": [
    {
      "title": "Weather type",
      "fill": "color",
      "gradientLength": {"signal": "clamp(height, 64, 200)"},
      "symbolType": "square",
      "encode": {"symbols": {"update": {"stroke": {"value": "black"}}}}
    }
  ]
}
```

At least this doesn't require another `facet` transform just for the strokes. In addition, I found that we can write `"y": {"field": {"group": "y"}, "mult": -1}` in the inner group, it's a bit better than the previous version.

However, neither using `"y": {"field": {"group": "y"}, "mult": -1}`, or implementing `clipY` or `translateChildren` in Vega won't help the stroke issue. Unless Vega fixes https://github.com/vega/vega/issues/2197, we have to add another `rect` mark outside the clip group.

domoritz:
Vega has to eventually fix vega/vega#2197 so I think we can just assume that it will be fixed in the future and go ahead with this pull request.

examples/compiled/bar_diverging_stack_population_pyramid.vg.json
```diff
@@ -49,19 +49,63 @@
...
           "y": {"scale": "y", "field": "age"},
-          "height": {"scale": "y", "band": true}
+          "height": {"scale": "y", "band": true},
+          "x": {
+            "signal": "min(scale('x',datum[\"min_sum_signed_people_start\"]),scale('x',datum[\"max_sum_signed_people_start\"]),scale('x',datum[\"min_sum_signed_people_end\"]),scale('x',datum[\"max_sum_signed_people_end\"]))"
+          },
+          "x2": {
+            "signal": "max(scale('x',datum[\"min_sum_signed_people_start\"]),scale('x',datum[\"max_sum_signed_people_start\"]),scale('x',datum[\"min_sum_signed_people_end\"]),scale('x',datum[\"max_sum_signed_people_end\"]))"
+          },
+          "clip": {"value": true}
```
kanitw: We just want to support clipping for corner radius of stacked bars, but somehow we modify all of bar specification to include aggregate and clipping.  I think that's a bit of an overkill.  

I'd rather generate these complicated stuff only when they're needed. Otherwise, this dilute our story for moving from VL to VG since our generated VG are unnecessarily complicated.

(In general, we should only generate Vega diff / SVG only when necessary.)

donghaoren: We are only doing that for stacked bars --- this one uses stack for the x axis.

kanitw: But if there is no corner radius, we don't really need to do clipping? 

Besides spec complexity, it probably makes it a bit slower to render?
I'm more concerned about spec complexity though.  It's kinda like we're handling corner case here, but by doing that we shouldn't make the common case worse. :)

src/compile/mark/init.ts
```diff
@@ -56,6 +57,12 @@ export function normalizeMarkDef(
     markDef.cursor = cursor(markDef, encoding, config);
   }
 
+  if (markDef.type == 'bar') {
```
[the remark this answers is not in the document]
donghaoren: Moved this into `mark.ts`.

examples/compiled/bar_diverging_stack_population_pyramid.vg.json
donghaoren: I see, good point!

donghaoren:
@domoritz @kanitw this should be ready to go when Vega 5.9 is out.

donghaoren:
Thanks for all the comments and suggestions!

src/compile/mark/mark.ts
```diff
@@ -83,6 +85,170 @@ function parsePathMark(model: UnitModel) {
...
+
+/**
+ * We need to put stacked bars into groups in order to enable cornerRadius for stacks.
+ * If stack is used and the model doesn't have size encoding, we put the mark into groups,
+ * and apply cornerRadius properties at the group.
+ */
+function getStackGroups(model: UnitModel) {
+  // Don't use nested groups when cornerRadius is not specified
+  const hasCornerRadius = VG_CORNERRADIUS_CHANNELS.some(
+    prop => model.markDef[prop] != null || getMarkConfig(prop, model.markDef, model.config) != null
```
kanitw: We use strict comparison (`!==`) in general.  

Maybe we should turn this on for linting in a separate PR.

examples/specs/normalized/stacked_bar_count_corner_radius_stroke_normalized.vl.json
```diff
@@ -0,0 +1,12 @@
...
+  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
+  "data": {"url": "data/seattle-weather.csv"},
+  "mark": {"type": "bar", "cornerRadius": 10, "stroke": "black"},
+  "encoding": {
+    "x": {"timeUnit": "month", "field": "date", "type": "ordinal"},
+    "y": {"aggregate": "count", "field": "*", "type": "quantitative"},
+    "color": {"field": "weather", "type": "nominal"}
+  },
+  "config": {"bar": {"strokeWidth": 2}},
+  "autosize": {"type": "pad"}
```
kanitw: Seeing this file unnecessarily generated makes me wonder if it was a good idea that we did https://github.com/vega/vega-lite/pull/5457.

(With that refactor, now our PRs will become generally more complicated as we always generate a normalized version of all specs.)

This is beyond the scope of this PR though. :)

src/compile/mark/mark.ts
donghaoren: `!=` is actually checking for both `null` and `undefined`, changing to `!==` is probably going to break. Do we have a helper function for this?

donghaoren: oh I'm wrong, `!==` is the right thing here.

examples/specs/normalized/stacked_bar_count_corner_radius_stroke_normalized.vl.json
domoritz: I don't want to always create a normalized version. I don't know why I accepted it. I think we should fix this. I filed https://github.com/vega/vega-lite/issues/5618

kanitw:
I've rebased and force pushed.

PR #5619, opened 2019-12-04
Use babel to create ie11 (es5) compatible build

nyurik:
This creates a build-es5/ dir, similar to Vega main repo.
I am not yet certain if this is required, or if tsc compiler
already creates ie11-compatible code as is.

Please:

- [ ] Make the PR atomic. (Fix one issue at a time.) Multiple relevant issues that must be fixed together? Make atomic commits so we can easily review each issue.

- [ ] Provide a concise title so we can easily copy it to the release note.

  - Mention relevant issues. (e.g., `Fix #1` / `Fix part of #1`)

- [ ] Lint and test (Run `yarn test`)

- [ ] Rebase onto the latest `master` branch

- [ ] Review your changeset first (to ensure code quality)

- For new features
  - [ ] Add new unit tests
  - [ ] Update the documentation under `site/docs/` + add examples

Tips:

- https://medium.com/@greenberg/writing-pull-requests-your-coworkers-might-enjoy-reading-9d0307e93da3 is a nice article about writing a nice pull request.
- Use draft PR for work in progress PRs / when you want early feedback (https://github.blog/2019-02-14-introducing-draft-pull-requests/)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ _site/
 .jekyll-metadata
 .yarnclean
 build/*
+build-es5/*
 coverage
 examples/all-examples.json
 examples/compiled/*.png
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,10 @@
+module.exports = {
+  presets: [
+    ['@babel/preset-env', {
+      targets: {
+        "ie": "11"
+      }
+    }]
+  ],
+  compact: false
+};
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   "scripts": {
     "build": "yarn build:only",
     "build:only": "yarn tsc:src && rollup -c",
-    "postbuild": "terser build/vega-lite.js -cm --source-map content=build/vega-lite.js.map,filename=build/vega-lite.min.js.map -o build/vega-lite.min.js && yarn schema",
+    "postbuild": "terser build/vega-lite.js -cm --source-map content=build/vega-lite.js.map,filename=build/vega-lite.min.js.map -o build/vega-lite.min.js && yarn schema && babel build/vega-lite.js --out-dir build-es5 --config-file ./babel.config.js",
     "prebuild:examples": "yarn build:only",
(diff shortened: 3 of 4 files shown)


domoritz:
I would strongly prefer not to add another dependency and instead just use tsc (if possible).

nyurik:
I would too, but this keeps it consistent with Vega lib itself, making sure it matches exactly the same output requirements/expectations.  With tsc, you will always have to match babel profile with tsc settings - not something that will remain consistent.  Besides, this is not like you are including a new binary, just using some extra tool in the build chain - not something that critical.

But again, it is still unknown if this is needed, will be doing some testing soon.

domoritz:
Makes sense. Maybe we can use babel instead of TSC to transpile Vega-Lite in general.

Vega and Vega-Lite have somewhat different build systems. Vega does not use much Typescript right now for example.

terencehonles:
> Makes sense. Maybe we can use babel instead of TSC to transpile Vega-Lite in general.
> 
> Vega and Vega-Lite have somewhat different build systems. Vega does not use much Typescript right now for example.

Using Babel instead of TSC would probably not be the best idea at this point in time.

[According to the announcement post:](https://devblogs.microsoft.com/typescript/typescript-and-babel-7/)

> # Caveats
> As we mentioned above, the first thing users should be aware of is that Babel won’t perform type-checking on TypeScript code; it will only be transforming your code, and it will compile regardless of whether type errors are present. ...

domoritz:
I see. Then let's just switch to tsc only.

terencehonles:
> I see. Then let's just switch to tsc only.

Just to be clear, the quote I pointing out was TS -> JS via Babel, **not** ES6 -> ES5 via Babel, and this PR would still work. I'm not sure if we want to continue down this route or #5920

terencehonles:
@nyurik looks like your PR effectively got in (although indirectly) :slightly_smiling_face:

domoritz:
Yep. Thank you @nyurik!

nyurik:
it's ok, only 4.5 months :)
Thanks for merging!

domoritz:
🦆 😇

PR #5623, opened 2019-12-04
fix!: switch to vega time format, remove `shortTimeLabels`

haldenl:


domoritz:
This is a `refactor:`, right?

src/guide.ts
```diff
@@ -66,7 +59,7 @@ export interface VlOnlyGuideConfig {
 
 export type GuideEncodingEntry = {[k in VgEncodeChannel]?: ValueDef & ConditionValueDefMixins};
 
-export const VL_ONLY_GUIDE_CONFIG: (keyof VlOnlyGuideConfig)[] = ['shortTimeLabels'];
+export const VL_ONLY_GUIDE_CONFIG: (keyof VlOnlyGuideConfig)[] = [];
```
kanitw: We can remove this variable and its reference?

Issue #5595, opened 2019-11-28
Changes uncovered by Altair 4


jakevdp:
In https://github.com/altair-viz/altair/pull/1818 I begun the process of making Altair wrappers for Vega-Lite 4.0.0-beta.12. This introduced a couple test failures in examples due to properties that are no longer allowed by the schema.

## To add back

- [x] https://altair-viz.github.io/gallery/ridgeline_plot.html fails because ``bounds`` is no longer an allowed property of ``TopLevelUnitSpec``.
  - Ham: this one we should add back.
 

- [x] Any charts that use ``width`` or ``height`` in the view configuration break, because they are no longer allowed properties of ``ViewConfig``
  - Ham: this one we should add back, perhaps as deprecated configs.

## Need to decide whether we want to add these back

- [ ] https://altair-viz.github.io/gallery/anscombe_plot.html and several others fail because ``columns`` is no longer an allowed property of ``TopLevelUnitSpec``.

- [ ] https://altair-viz.github.io/gallery/beckers_barley_trellis_plot.html fails because ``rangeStep`` is no longer an allowed property of ``Scale``.

- [ ] Similarly, https://altair-viz.github.io/gallery/layered_chart_bar_mark.html fails because ``rangeStep`` is no longer an allowed property of ``ScaleConfig``

This is the extent of the issues that came up in a first pass!

kanitw:
I'm adding `bounds` and `config.view.width/height` back in https://github.com/vega/vega-lite/pull/5624. 

However, I don't think we should do `columns` as it's much better to specify:  `alt.Facet('Series', columns=2)`.

kanitw:
For `rangeStep`, width.step is also a much better syntax, so I slight prefer not to add it back.

jakevdp:
Thanks!

kanitw:
:)

jakevdp:
One question: currently altair's default theme uses config.width/config.height. Would you prefer me to change that to config.continuousWidth/continuousHeight in V4, to move away from deprecated configs?

kanitw:
Yes, please. :)

jakevdp:
Great! Already done in https://github.com/altair-viz/altair/commit/0ccd2e4a8c59f9fa5a1009ed4716bd6882187d01

eharkins:
Changing `scale.rangeStep` to `width.step` in Altair version 4.0.1 caused this for me:

```
Error: Unrecognized signal name: "x_step"
[...]

PR #5624, opened 2019-12-04
Add missing properties for Altair 4

kanitw:
Part of #5595


domoritz:
I'll need to add support for the deprecated property in the schema generator.

PR #5625, opened 2019-12-04
fix: adjust default header label align

kanitw:
temp fix for https://github.com/vega/vega/issues/2233


kanitw:
Hmm this doesn't really work. It's still unclear to me how to systematically fix this.

PR #5638, opened 2019-12-05
docs: add responsive bar chart example

kanitw:
the bar_size_responsive doesn't show up...


kanitw:
I'm not sure what it's not showing up .. weird

PR #5642, opened 2019-12-06
docs: improve multiline text examples (remove bad ones + upgrade good ones)

### LINKED ISSUE #5635: Better multiline text example

kanitw:
Fix #5635

(The text multiline one is a bad example, and we don't need it to prevent regression anymore since we have the other two.)


domoritz:
Hmm, the example tests multiline in text marks, axis titles, headers, titles, and subtitles. I still think it's somewhat valuable but I don't have strong opinions.

kanitw:
They are all the same text marks in the scene graphs, so I think I'll merge as is.

domoritz:
In the scenegraph, yes, but not in the Vega-Lite spec. There are somewhat different code paths. Anyway, it's not important.

PR #5647, opened 2019-12-07
fix: don't make every spec normalized

kanitw:


site/_data/versions.yml
```diff
@@ -1,3 +1,4 @@
 vega: 5.9.0
+vega-lite: 4.0.0-beta.12
```
domoritz: This is already fixed in master

PR #5649, opened 2019-12-07
refactor: rename noSignal -> noValueRef, remove custom range config

domoritz:


kanitw:
Please make sure docs for relevant schema doesn't break

PR #5657, opened 2019-12-08
chore: remove now unnecessary timeunit functions

haldenl:
There are a couple functions that are now unnecessary after the move to Vega time format.


kanitw:
Removing useless code! :tada:

Issue #5505, opened 2019-10-30
Center align doesn't work for rect marks


donghaoren:
I'm trying to make tooltips with `rect` marks as backgrounds. In the following spec:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"a": "A", "b": 28},
      {"a": "B", "b": 55},
      {"a": "C", "b": 43},
      {"a": "D", "b": 91},
      {"a": "E", "b": 81},
      {"a": "F", "b": 53},
      {"a": "G", "b": 19},
      {"a": "H", "b": 87},
      {"a": "I", "b": 52}
    ]
  },
  "encoding": {"x": {"field": "a", "type": "ordinal"}},
  "layer": [
    {
      "mark": "bar",
      "encoding": {"y": {"field": "b", "type": "quantitative"}},
      "selection": {
        "hover": {
          "type": "single",
          "on": "mouseover",
          "clear": "mouseout",
          "empty": "none"
        }
      }
    },
    {
      "mark": {
        "type": "rect",
        "color": "#ccc",
        "y": 0,
        "y2": 10,
        "align": "center",
        "width": 30
      },
      "encoding": {
        "opacity": {
          "condition": {"test": {"selection": "hover"}, "value": 1},
          "value": 0
        }
      }
    }
  ]
}
```

![image](https://user-images.githubusercontent.com/1595237/67905247-fa1fd100-fb2d-11e9-8cd9-88ba6b3b0ef2.png)

Problem 1: The `align: "center"` on the `rect` marks have no effect. Since I set `width: 30` and `align: "enter"`, I would expect the rect to be center aligned.

Problem 2: In another attempt, I tried setting `xOffset`, but it has no effect (related to #5502)

andreasplesch:
having the same problem

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/5664

PR #5664, opened 2019-12-10
fix: correctly apply align to `bandPosition`, make center the default align

kanitw:
also don't throw incorrect warning

Fix https://github.com/vega/vega-lite/issues/5505

### DIFF
--- a/src/compile/mark/mixins.ts
+++ b/src/compile/mark/mixins.ts
@@ -322,7 +322,7 @@ export function bandPosition(
     [vgChannel]: ref.fieldRef(fieldDef, scaleName, {}, {band: 0.5})
   };
 
-  if (encoding.size != null || markDef.size != null) {
+  if (encoding.size || (markDef.size !== null && markDef.size !== undefined)) {
     const orient = markDef.orient;
     if (orient) {
       if (getTypedFieldDef(encoding.size) || isValueDef(encoding.size)) {
@@ -341,7 +341,7 @@ export function bandPosition(
     }
   }
 
-  if (defaultSizeRef && defaultSizeRef?.value !== undefined) {
+  if (defaultSizeRef?.value !== undefined) {
     return {
       ...centeredBandPositionMixins,
       [sizeChannel]: defaultSizeRef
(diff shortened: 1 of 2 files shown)


src/compile/mark/mixins.ts
```diff
@@ -314,42 +314,39 @@ export function bandPosition(
...
-        return {
-          ...centeredBandPositionMixins,
-          ...nonPosition('size', model, {vgChannel: sizeChannel})
-        };
-      } else if (isValueDef(model.encoding.size)) {
+  const centeredBandPositionMixins = {
+    [vgChannel]: ref.fieldRef(fieldDef, scaleName, {}, {band: 0.5})
+  };
+
+  if (encoding.size != null || markDef.size != null) {
```
donghaoren: Should we use `!=` or `!==` here?

src/compile/mark/mixins.ts
```diff
@@ -314,42 +314,39 @@ export function bandPosition(
...
+          [sizeChannel]: {value: markDef.size}
         };
       }
     } else {
-      log.warn(log.message.cannotApplySizeToNonOrientedMark(model.markDef.type));
+      log.warn(log.message.cannotApplySizeToNonOrientedMark(markDef.type));
     }
   }
 
+  if (defaultSizeRef && defaultSizeRef?.value !== undefined) {
```
donghaoren: I think `if(defaultSizeRef?.value !== undefined)` is sufficient here

src/compile/mark/mixins.ts
domoritz: `!=` is useful when you also want to support null and undefined.

src/compile/mark/mixins.ts
kanitw: Right, I added ? so I can remove the first part, but then forgot to remove lol

src/compile/mark/mixins.ts
donghaoren: @kanitw said we want to be clearer in terms of semantics with `x !== null && x !== undefined`. Did you guys agree on which one to use?

domoritz: `!= null` is a common pattern and I think it's good to use. In all other cases, we want to be explicit.

PR #5665, opened 2019-12-11
fix: respect offset for mark.x/y

### LINKED ISSUE #5331: xOffset/yOffset cannot be applied to static x/y markdef properties
If I want to place a circle mark just below the bottom of the view, I can do:

```json
{
    "data": {
        "values": [0]
    },
    "mark": {
        "type": "circle",
        "size": 250,
        "yOffset": 25
    },
    "encoding": {
        "y": { "value": "height" }
    }
}
```

which will yield the following correct result:

visualization

However, if want to express the same thing statically in the markdef (without any encodings), I should be able to do:

```json
{
    "data": {
        "values": [0]
    },
    "mark": {
        "type": "circle",
        "size": 250,
        "y": "height",
        "yOffset": 25
    }
}
```

However, the offset is not applied in this case:

visualization-2

It seems that `yOffset` is only applied to a `y` encoding value, but it should also apply to a static `y` markdef property, because otherwise there is no way to position a static annotation against the height of a container.

The same goes for `x` & `xOffset`.

kanitw:
fix #5331

### DIFF
--- a/src/compile/mark/valueref.ts
+++ b/src/compile/mark/valueref.ts
@@ -1,10 +1,8 @@
-import {ValueOrGradientOrText} from './../../channeldef';
-import {array} from 'vega-util';
 /**
  * Utility files for producing Vega ValueRef for marks
  */
 import {SignalRef} from 'vega';
-import {isFunction, isString, stringValue} from 'vega-util';
+import {array, isFunction, isString, stringValue} from 'vega-util';
 import {isCountingAggregateOp} from '../../aggregate';
 import {isBinned, isBinning} from '../../bin';
 import {Channel, getMainRangeChannel, PositionChannel, X, X2, Y, Y2} from '../../channel';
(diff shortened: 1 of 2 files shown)


src/compile/mark/valueref.ts
```diff
@@ -373,7 +372,16 @@ export function midPoint({
...
   }
 
-  return isFunction(defaultRef) ? defaultRef() : defaultRef;
+  const ref = isFunction(defaultRef) ? {...defaultRef(), ...(offset ? {offset} : {})} : defaultRef;
+
+  if (ref) {
+    // for non-position, ref could be undefined.
+    return {
+      ...ref,
+      ...(offset ? {offset} : {})
```
domoritz: What if the offset is 0? 

Should this be:

```suggestion
      ...(offset != null ? {offset} : {})
```

kanitw: If offset is 0, you can safely omit the offset.

domoritz: Good. Just wanted to make sure. Maybe it's worth a comment here since `x ? .. : ...` always makes my alarm bells go off.

PR #5668, opened 2019-12-11
refactor: re-organize mark compiler

kanitw:
This should allow us to read the mark compiler code much easier to read.

- move `mark/mixins` and `mark/valueref` to `mark/encode/*`…

Instead of splitting things by role (as encoding mixins / valueref), we're now grouping encoding mixins by their channels and put channel-specific refs in the same files.  This means less jumping between files when reading the code. 

- distinguish between `defaultRef` and `defaultPos`, which was confusing before.


domoritz:
@jheer does this pr start the refactoring you asked for?

kanitw:
This is definitely a big step towards simpler code base.  

I don't think I want to combine the compiler (inside `compile/` and the type + help methods (right under `src/`) though as right now the split is pretty clean.

PR #5675, opened 2019-12-14
feat: add angle channel for point and text marks

### LINKED ISSUE #4548: Add angle channel for symbols
I couldn't find any previous discussion on this, but apologies if I am repeating previous discussion. 

Would it be practical / desirable to have an orientation channel? This seems one of Bertin's more straightforward retinal variables that is currently absent from Vega-Lite. Personally I would find this very helpful, especially for geographically embedded visualisations.

One use case: To create directional glyphs for representing vector fields. I have managed to [simulate these programmatically](https://gist.github.com/jwoLondon/896e094ac367779e46f1a5d4f9c53d41) by generating a set of SVG path custom shapes, but this is too cumbersome to be a routine approach and requires the data to encode with orientation to be categorical. 

One could imagine a syntax consistent with other channel encodings, such as 

```json
    "shape": {"value": "triangle-up"},
    "size": {"field": "speed","type": "quantitative"},
    "orientation" : {"field": "direction", "type": "quantitative"}
```

leading to output such as

orientationexample

kanitw:
fix #4548

Note that I'm merging this to feature branch as we better settle all angle-related marks all at once to avoid design changes we may need to make


kanitw:
Hmm, I don't know why the SVG looks weird, here is how it's like in my editor: 

image

kanitw:
Note that I also don't understand why the angle looks a bit different from the Vega example: https://vega.github.io/vega/examples/wind-vectors/

But I'm probably too sleep to debug this as this point :p

src/encoding.ts
```diff
@@ -162,14 +162,19 @@ export interface Encoding<F extends Field> {
   strokeWidth?: NumericFieldDefWithCondition<F> | NumericValueDefWithCondition<F>;
 
   /**
-   * Size of the mark.
+   * Size of the marks.
```
domoritz:
```suggestion
   * Size of the mark.
```

or 

```suggestion
   * Size of marks.
```

Issue #5478, opened 2019-10-16
Chrome not rendering marks in vega-lite editor


lelandwilkinson:
Using Google Chrome browser (up to date and refreshed)
Version 77.0.3865.120 (Official Build) (64-bit)

This picture shows the problem. This example was taken from 
https://vega.github.io/vega-lite/examples/point_2d.html

Works fine in Safari. Same problem rendering SVG or Canvas

![image](https://user-images.githubusercontent.com/19478753/66932705-a1b0e580-effd-11e9-88c3-741934998087.png)

domoritz:
The example at https://vega.github.io/editor/#/examples/vega-lite/point_2d works for me. Can you try to clear your cache with SHIFT + CMD + R?

<img width="438" alt="Screen Shot 2019-10-16 at 08 56 26" src="https://user-images.githubusercontent.com/589034/66936624-e1260480-eff2-11e9-83c0-efeebc93ed7e.png">

domoritz:
It was probably some version mismatch in a cached version of some file on the page. I'm glad we could resolve it. 

Happy plotting!

gordonwoodhull:
I have found that sometimes my last-edited spec gets "stuck" in the editor, so it's not parsing/viewing the same thing I asked for. 

In fact, this happened to me when clicking the above link - I got parsing errors on a spec I was looking at earlier and the page didn't display at all, due to an error in error handling.

You can clear the editor state in DevTools Application/Local Storage. Select "redux" and press <kbd>delete</kbd>

lelandwilkinson:
Nope. I even tried "Clear Site Data"  However, going back to 
https://vega.github.io/vega-lite-v1/examples/ <https://vega.github.io/vega-lite-v1/examples/>
and clicking on the scatterplot example launches
https://vega.github.io/vega-editor/?mode=vega-lite&spec=scatter <https://vega.github.io/vega-editor/?mode=vega-lite&spec=scatter>

and it renders just fine. But if I click on the "new editor" link at the top of the page and paste in the same JSON, it fails to render the marks again.

I think you're right that it has something to do with caching, but I'll have to investigate in more detail.

Thx,
L

domoritz:
That's odd. Can you try to open the editor in incognito mode? You can also go to https://vega.github.io/editor/#/reset to reset the saved state.

Could there be an extension that blocks some data (such as an adblocker)?

[...]

PR #5676, opened 2019-12-14
support arc mark 

kanitw:
This is an early stage of the arc mark implementation that I did on the flight. 
It's working!  But still need to make some decisions on the syntax. 

I currently introduced 4 new channels: `angle`, `angle2`, `radius` (for `outerRadius`) and `radius2` (for `innerRadius`).

Note that this diverges from Vega, which uses `start/endAngle` and `inner/outerRadius`. 
This syntax is consistently with the rest of VL (x/x2, lat/lat2).  
Plus, in Vega, innerRadius can actually be larger than outerRadius, so the words for inner/outer doesn't really mean much. (@jheer seemed to approved this direction when I last chatted to him about arc mark.) 

However, this syntax is still quite problematic in a few ways.

## 1) Angle 

As I consider only the arc marks to begin with, I call the channel angle instead of theta as it's a bit more friendly to lay people.  (Angle and radius go together better as `theta` normally goes with `r` or `rho`.) 

Angle is totally fine for arc marks alone.  However, the name `angle` would not work if we want to consistently apply this to text marks and allow layer encoding shorthand like this:

```
{
  "data": {
    "values": [
      {"id": 1, "field": 4},
      {"id": 2, "field": 6},
      ...
    ]
  },
  "encoding": {
    "angle": {"field": "field", "type": "quantitative"},
    "color": {"field": "id", "type": "nominal"}
  },
  "layer": [
    {"mark": "arc"},
    {"mark": "text", radiusOffset: 50}
  ]
}
```

The reason is that text already has `angle` for rotation angle in Vega.
Vega instead [uses `theta` for this polar coordinate angle](https://vega.github.io/vega/examples/radial-plot/).

This means that we can't use the spec above (as angle will actually rotate the text, not placing it at the right location). 

To avoid this problem, I'm leaning towards calling this pair of channels for arc marks `theta` and `theta2` in Vega-Lite too.  This means that we can do the spec above (just change `angle` to `theta`). 

If we agree to do so, we still need to decide what to do when users specify arc marks with `angle` (which will be available as a channel via autocompletion). 

To summarize, here are our options

- 1a) Call the channel `theta`. Ignore `angle` for arc marks.

- 1b) Call the channel `theta`. Automatically rename `angle` to `theta` and throw a warning that it's called theta for arc marks.

- 1c) Call the channel `theta`. Allow `angle` as an alias, but still advertise `theta` as the primary way since `theta` would work with text.

- 1d) Call the channel `angle`.  If users want to label, too bad they can't use the layer encoding shorthand. 

I'm leaning toward either 1b or 1c and slightly prefer 1c. 

## 2) Radius 

Calling the channels `radius` and `radius2` is probably the right things for a few reasons:

1. Consistent with other channels in VL that has the main and secondary ones (e.g., x/x2)
2. inner/outer doesn't really mean much as described
3. text marks also call this `radius`, so this would work great with layer encoding shorthand

If user only specifies one `radius` without `radius2`, I think users would mean the `outerRadius` in general.  

However, to create a donut chart, this means we have to set `radius2` alone:

```
{
  "data": {
    "values": [
      {"id": 1, "field": 4},
      {"id": 2, "field": 6},
      ...
    ]
  },
  "mark": {"type": "arc", "radius2": 50},
  "encoding": {
    "angle": {"field": "field", "type": "quantitative"},
    "color": {"field": "id", "type": "nominal"}
  }
}
```

Note that we set the default `radius` to be `min(width/height)/2`, so it's a bit awkward to manually set the number for radius if we only want to adjust `innerRadius`. 

Plus, Vega config would still have `innerRadius` and `outerRadius`.

I'm pretty still convinced that `radius` and `radius2` should stay (given the reasons above). 

The question is whether we allow `innerRadius` and `outerRadius` as aliases for `radius2` and `radius` at all.  

I know aliases are annoying, but we're in a situation where it's at at least always available in config anyway, so we might as well make donut spec looks a bit cleaner like this: `"mark": {"type": "arc", "innerRadius": 50},`

cc: @donghaoren @domoritz @jheer @jakevdp @haldenl @arvind -- if you have time, let me know what you think.


lelandwilkinson:
One thing I worry about when designing a new feature for a system is whether it will come back to haunt us when we discover it is an instance of a more general feature and will force massive refactoring to implement the more general feature. I don't think this will happen with arc mark, even though you know I've advocated going more vigorously into non-rectangular coordinate systems. Also, I'm sure you know, I spent a lot of effort trying to minimize the number of marks in GG because I hate redundancy and love expressiveness in software design. But I trust your instincts and expect arc mark might later generalize in a simple way. Do keep in mind, also, that annotations and decorations (axes, labels, legends, ...) are critical in many applications and among the most difficult things to implement in different coordinate systems. I'm sure that everyone who gets into graphics below a surface level, like all of you, have run into that problem.

lelandwilkinson:
Forgive me for a tangential comment, but I've heard a few opinions here about pie charts that are unsupportable. Statisticians (and designers like Tufte) rant about pie charts and their negativism creeps into blogs that users take for gospel. These rants are not based on solid psychological research and they are made by people who have no training in psychology. Allow me a story.

Some years ago I spoke to a committee at the National Academy on statistical graphics. Bill Cleveland couldn't make it and asked me to fill in. So I summarized the Cleveland research and other psychological experiments on perception of statistical graphics. As I belittled pie charts (based on Cleveland's finding regarding angles), a member interrupted me. "That judgment isn't supported by what we know about perception." 

The person was John Swets. 

I was taken aback, and went home to look more widely at the evidence. Here it is, including more recent studies:

Simkin, D., & Hastie, R. (1987). An Information-Processing Analysis of Graph Perception. Journal of the American Statistical Association, 82(398), 454

Hollands, J.G., and Spence, I. (1998). Judging proportion with graphs: The summation model. Applied Cognitive Psychology, 12, 173-190. 

Lewandowsky and Spence (1998). The perception of statistical graphs. Sociological Methods and Research, 18. 200-242.

Spence, Ian (1990). Visual psychophysics of simple graphical elements. Journal of Experimental Psychology: Human Perception and Performance, 16, 683-692. 

Spence, Ian, and Lewandowsky, Stephan (1991). Displaying proportions and percentages. Applied Cognitive Psychology, 5, 61–77.

Spence, Ian (2005). No Humble Pie: The Origins and Usage of a Statistical Chart. Journal of Educational and Behavioral Statistics, 30, No. 4, pp. 353–368. 

These researchers are experts in perception and cognition. Bottom line: Swets was right. For accurate judgments on proportion-of-whole data, pie charts are preferred to divided bars and other rectangular representations as long as there aren't more than about 7 categories. Furthermore, as Rob Kosara (not a psychologist, but a damn good researcher on graphic perception) has shown in several studies, pie slices are judged by area, not angle. See here for a summary if you're not familiar with this research:

https://eagereyes.org/pie-charts

How can we get ordinary users to pay attention to this research instead of citing sources who have no expertise in this area? Well, you are thought leaders on visualization so it starts with you. I stole some of your Github space because I don't know anywhere else where people care as much about visualization. Thanks.

donghaoren:
In my mental model, the polar coordinate system is implicitly triggered when we use `theta` and `radius` channels.

What happens to other marks, like `rect`, `line`, `area`, `rule`, `tick`, `text`, and composite marks like `boxplot`? Should we use the same morphing mechanism as rect -> arc (i.e., a horizontal line/rule becomes an arc)?  Should there be an option to enable / disable morphing?

Weather Radials (http://www.weather-radials.com/) is a good example of `rule` and `point` marks in polar coordinates.

For a text marks, should it be upright or rotate according to its `theta`? Should we allow people to have text along the circle?

What should `x`, `y` encoding do when polar coordinates is used? Perhaps they should be disabled, but I can see cases where you want to put labels at a specific (x, y) location (i.e., top-left of the chart).

ghalimi:
An arc mark is required to implement support for [circos charts](http://circos.ca/), and these are superior to any known alternative for a wide range of applications. I cannot wait to use this mark with Vega-Lite. Great work @kanitw!

kanitw:
@lelandwilkinson  Thanks for your comment. I think one of the best way to include these references is to include them in our documentation. We'll do that before merging this PR.

> Also, I'm sure you know, I spent a lot of effort trying to minimize the number of marks in GG because I hate redundancy and love expressiveness in software design.

Definitely.  That's why we try to make sure we can re-use marks like area (unlike ggplot2 that has area and ribbon).

kanitw:
> What happens to other marks, like rect, line, area, rule, tick, text, and composite marks like boxplot? Should we use the same morphing mechanism as rect -> arc (i.e., a horizontal line/rule becomes an arc)? Should there be an option to enable / disable morphing?

By using `theta`, `radius` channel, we can make these behaviors like you suggests in the future.  
However, implementing such behavior is beyond the scope of this PR.  (At least, we need to add such support in Vega first.) 


> For a text marks, should it be upright or rotate according to its theta? Should we allow people to have text along the circle?

As mentioned above:  "text already has angle for rotation angle in Vega.
Vega instead uses theta for this polar coordinate angle."


> What should x, y encoding do when polar coordinates is used? Perhaps they should be disabled, but I can see cases where you want to put labels at a specific (x, y) location (i.e., top-left of the chart).

They will be combined (added as offset).  I don't think they should be disabled as you might want to place arcs at different cartesian coordinates. For example, one can make symbol map with pies like this:

image

marcprux:
We're very excited to finally get arc marks! Is there anything I can do to help move this draft PR forward?

kanitw:
Closing as this be replaced with #6201.

lelandwilkinson:
I've lost the issue on github, but I thought you might want to know that the problem I reported with the vega-lite editor not rendering properly on my version of the Chrome browser went away when I updated Chrome recently to Version 83.0.4103.116 (Official Build) (64-bit).

Sorry I can't be more specific, because it was a while ago and I can't see any issues on github now with Chrome. Maybe Google changed something.

Best,
Lee Wilkinson

clemens-tolboom:
Guess it was https://github.com/vega/vega-lite/issues/5478 as it was the only on on https://github.com/lelandwilkinson?tab=overview&from=2019-12-01&to=2019-12-31 :-p

Issue #5660, opened 2019-12-09
Escaping of square brackets does not work


jakevdp:
I have been trying to follow https://vega.github.io/vega-lite/docs/field.html to escape square brackets in field names, and cannot get it to work at all. Here is a representative spec ([vega editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEDGD2B2AzAlgc0gLjMSA3ZApgO6baRHIAmALgBakAsADEwDRS0Fq3WkDMLAL7tIAWwCGAJwDWpHNViwANtWQAHUvACuSpYOHgolcdXFzI8caIKlIx0wFpxADgCcAVj7RXAJld8ANgAjVyZ3IJ8+F38AnyCAdj5KPnj450gDCDEpWSxINVhkeF5WQ0gCeDhKIvQsUAgsgA9zagBPNRs8gEctcWLkU1VcGxEUAiVKW0aAHWmAOlxxJS0bTIbIVpb2zqgevtVB5GHIUcIJ21bZgG1ERVmAXQzDNcgAEgBnaE4JWx5qNXeGAA9EDhqhxHNUANaFognNkLAgZ9vuJQQRwQ4lAMCKC+HMGHMmHMAFbvBAnMr2cTvAjUd5yQxZKlONyebx+QIhMIRKJuQJxRLJVLpLBXRkNHCNBZLFakACMIlaN0UjywfDWDTIUsWyx2PkVytgqrAco1Esg2plOz4BtuRsYgnF92eIEEQA))
```json
{
  "data": {"values": [
      {"x.value": 1, "y[foo]": 3},
      {"x.value": 2, "y[foo]": 1},
      {"x.value": 3, "y[foo]": 4}
    ]},
  "mark": "point",
  "encoding": {
    "x": {"type": "quantitative", "field": "x\\.value"},
    "y": {"type": "quantitative", "field": "y\\[foo\\]"}
  }
}
```
Escaping of the ``"."`` seems to work as expected, but no amount of backslashes causes the y encoding to work properly.

domoritz:
Looks like we are removing the escaping for field references in scale domains and encode blocks.

domoritz:
I think this is an issue with Vega. We use a utility method to split paths and it removes the escaping.

```
$ vega.splitAccessPath('y\\[foo\\]')
["y[foo]"]
```

I would expect

```
$ vega.splitAccessPath('y\\[foo\\]')
["y\\[foo\\]"]
```

[...]

PR #5730, opened 2020-01-09
fix: support escaping of brackets

domoritz:
Fixes https://github.com/vega/vega-lite/issues/5660

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -293,6 +293,10 @@ export function titlecase(s: string) {
   return s.charAt(0).toUpperCase() + s.substr(1);
 }
 
+function escapePatchAccess(string: string) {
+  return replaceAll(replaceAll(replaceAll(string, ']', '\\]'), '[', '\\['), '.', '\\.');
+}
+
 /**
  * Converts a path to an access path with datum.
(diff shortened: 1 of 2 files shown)


src/util.ts
```diff
@@ -293,6 +293,10 @@ export function titlecase(s: string) {
   return s.charAt(0).toUpperCase() + s.substr(1);
 }
 
+function escapePatchAccess(string: string) {
+  return replaceAll(replaceAll(replaceAll(string, ']', '\\]'), '[', '\\['), '.', '\\.');
```
kanitw: Minor: This isn't very readable

domoritz: What do you suggest? Make a function to replace multiple characters or just a regex?

kanitw: a function? (or ignore this comment for now). 

I'm more interested to understand what's going on.

domoritz: The regex is not that difficult and we use regexes in other places for these kinds of things.

domoritz:
@kanitw https://github.com/vega/vega-lite/issues/5660 explains the issue.

PR #5759, opened 2020-01-19
docs: add missing `index` property for expression in the axis docs

kanitw:
Let's see if https://github.com/vega/vega/pull/2314 gets any edits before merging.


kanitw:
https://github.com/vega/vega/pull/2314/files is merged, without edits.

PR #5761, opened 2020-01-19
fix: make conditional axis config work 

kanitw:
the add example should be updated to show the right behavior


examples/compiled/line_conditional_grid_dash_config.vg.json
```diff
@@ -0,0 +1,132 @@
...
+  "data": [
+    {
+      "name": "source_0",
+      "url": "data/stocks.csv",
+      "format": {"type": "csv", "parse": {"date": "date"}},
+      "transform": [
+        {"type": "filter", "expr": "datum.symbol==='GOOG'"},
+        {
+          "type": "filter",
+          "expr": "inrange(time(datetime(year(datum[\"date\"]), 0, 1, 0, 0, 0, 0)), [time(datetime(2006, 0, 1, 0, 0, 0, 0)), time(datetime(2008, 0, 1, 0, 0, 0, 0))])"
```
domoritz: Is there a Vega time unit expression we could use for this? It seems inconsistent that we are using Vega everywhere else except here.

kanitw: IDK, but that's not the scope of this PR.  

This output is from the unchanged the codebase for predicate.

kanitw: Also, https://vega.github.io/vega/docs/expressions/#timeUnitSpecifier is fo r using with timeFormat, it doesn't preserve "numeric" part of datetime.

domoritz: I wonder whether we should add an expression function in Vega. Anyway, it's out of scope.

PR #5763, opened 2020-01-19
fix: enable example tests and fix numerous issues

domoritz:
Turns out, our example tests were not doing anything. This pull request enables them to work and fixes a bunch of errors that were happening.

This pull request also removes the conditional order from one example since we don't support that.


domoritz:
Needs https://github.com/vega/vega-lite/issues/5764

domoritz:
Needs https://github.com/vega/vega-lite/pull/5761

PR #5765, opened 2020-01-20
docs: change `line_conditional_grid_dash` to even more advanced `line_conditional_axis`

kanitw:
also simplify its predicate


kanitw:
Hmm, the hack to use whitespace instead of `labelOffset` doesn't work for SVG generation. (It worked OK on Chrome.)

site/docs/encoding/axis.md
```diff
@@ -117,7 +117,9 @@ Note that each axis tick, grid line, and label instance is backed by a data obje
 - `label` - the string label
 - `value` - the data value
 
-For example, we can adjust the grid dash in a line chart based on whether a particular grid line falls on a year boundary:
+### Example: Conditional Axis Properties and Multi-Line Axis Label
+
+In the following example, we adjust the `gridDash` and `tickDash` properties in a line chart based on whether a particular grid line falls on a year boundary. We also use the `labelExpr` property to show multi-line labels for year and month, showing the year number only for January of each year.
 
 <div class="vl-example" data-name="line_conditional_grid_dash"></div>
```
domoritz: Please rename references as well. 

```suggestion
<div class="vl-example" data-name="line_conditional_axis_"></div>
```

domoritz:
I think we can merge this despite bad offset btw.

kanitw:
> I think we can merge this despite bad offset btw.

Why? It's bad. I'd rather wait to fix the offset.

domoritz:
Okay. Let’s wait then.

kanitw:
This is rebased, now that we support `labelOffset`.

domoritz:
> Hmm, the hack to use whitespace instead of labelOffset doesn't work for SVG generation. (It worked OK on Chrome.)

Vega now always trims text so the hack does not work anymore.

kanitw:
line_conditional_axis_config.vl.json isn't correct yet, but it should be correct with Vega 5.10.  I'll track that in https://github.com/vega/vega-lite/pull/6033

PR #5795, opened 2020-01-25
fix: correctly initialize selections with dates

### LINKED ISSUE #5794: Selection initialization does not work with all dates
Selection initialization is supposed to accept Vega-Lite datetime objects but it fails with dates that use strings like "January". 

Screen Shot 2020-01-24 at 20 16 56

[Spec](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoAWOgCtY3AHaQANOCgATdHkw5gkVgCcx2LTvR883eAGtYdeLAqQAvuoiQ8R9EtgAzbiNUMwBtA39GMTxkIzNIbTxWVDpYAE9Udm4xAF48gHIAcQB5YsL8twBdDygxdDTYsI0IUAg22vrGnFDm9rADDCNbeIklZDcavv7eqchYZDFkeDweFX0Zqah503XNza80wnGcOcYlakW1Db3IMfQjZFg8M29WZEm9tshleNRuVnm3CocQ+nygyCU8G4mjO1FgYUgAA9INVrvszswzAZkfpIA17mYAEwABmJAA5VFA-koWPEAFK+Vj3NJqcwxMwARlcrjR7R5YPcvKp92GJ0I3DOzzR-L6qNmEKhMPOWLRSKxkEiC008US40pByO8RiqHFPlMgpuLNxmrE2rFRkY8D1UDwh2OUAAjkyacwdIwqG4NjKIBa2q0+l4fH5AsEwhEojE4rjtksVj8Tts3K45RHBqKoEZWJdQeDIdDYeqcf0NYwtTqdM6De6vGhTehzcGwDLKiBXEA)

cc @arvind

domoritz:
Fixes #5794

I also refactored things a bit. We now have separate functions for parsing. We also use numbers instead of strings in a few more places.


src/datetime.ts
```diff
@@ -52,7 +52,7 @@ export interface DateTime {
   month?: Month | string;
 
   /**
-   * Integer value representing the date from 1-31.
+   * Integer value representing the date (day of the month) from 1-31.
```
kanitw: By definition, dates mean days of month. Plus, we already say 1-31.  So these parentheses are a bit overkill.

domoritz: I appreciate the additional clarification since JavaScript uses `day` for the day of the month. With the clarification, you can ctrl+f for `day` and find the right name. Also, date is not generally understood as the day fo the month. Where did you see this?

kanitw: > since JavaScript uses day for the day of the month.

No, it does not. We do match JS. Day is day of week, as it should be. 

image


> Also, date is not generally understood as the day fo the month. Where did you see this?

This is well defined in the dictionary. 

image

src/datetime.ts
```diff
@@ -232,55 +231,91 @@ export function dateTimeExpr(d: DateTime | DateTimeExpr, normalize = false, toJS
...
   }
+}
+
+/**
+ * Return Vega expression for a date time expression.
+ *
+ * @param d the internal date time object with expression.
+ * @returns the Vega expression.
+ */
+export function dateTimeExprToExpr(d: DateTimeExpr) {
```
arvind: Would it be better to consolidate this function with the one above it? In particular, I find this function name `dateTimeExprToExpr` difficult to parse/understand.

src/datetime.ts
```diff
@@ -232,55 +231,91 @@ export function dateTimeExpr(d: DateTime | DateTimeExpr, normalize = false, toJS
...
+/**
+ * @param d the date time.
+ * @returns the timestamp.
+ */
+export function dateTimeToTimestamp(d: DateTime) {
+  const parts: (string | number)[] = dateTimeParts(d, true);
 
   if (d.utc) {
-    return `utc(${unitsString})`;
+    return +new Date(Date.UTC(...(parts as [any, any])));
```
arvind: I want to double check: will this work when `parts` contains some `string` elements?

src/datetime.ts
arvind: I agree, I don't think the additional clarification hurts.

domoritz: > No, it does not. We do match JS. Day is day of week, as it should be.

Oh, that's confusing. I missed that `getDate` returns the day of the month but the parameter to the constructor is called `day`. 🤦‍♂ 

Screen Shot 2020-01-26 at 10 05 32

> This is well defined in the dictionary.

I didn't find https://en.wikipedia.org/wiki/Calendar_date clarified this well. I asked @ldegreef what she thinks a "date" was and she thought that date refers to the whole time specifier (like here: https://en.wikipedia.org/wiki/ISO_8601#Dates). Moreover, in this pull request, I replaced a few occurrences of `day` with `date` since were assigned numbers above 7, which indicates that we are confused ourselves about what `day` and `date` mean. Of course, it's too late, but maybe `dayOfMonth` and `dayOfWeek` or `date` and `weekday` would have been better names to avoid confusion.

src/datetime.ts
domoritz: No, it doesn't need to but this shouldn't be an issue since we normalize strings to numbers. Let me check what happens if a user doesn't respect the schema.

domoritz: Looks like js accepts this: `new Date("2019", "3", "2")` so we are good. Thanks for checking.

src/datetime.ts
domoritz: I'd like to but we want to normalize if d is a DateTime and not normalize if it's a DateTimeExpr. Unfortunately, these two types can't reliably be distinguished at runtime so I think it's best to have two functions.

src/datetime.ts
domoritz: I added some code to always support numbers even if they are in strings.

test/datetime.test.ts
```diff
@@ -1,185 +1,176 @@
...
       }).toThrow();
     });
 
     it('should use utc expression if utc is specified', () => {
-      const d = {
+      const d: DateTime = {
         year: 2007,
-        day: 'monday',
+        day: 1,
         utc: true
```
haldenl: is this already in the docs?

domoritz: Is what? utc?

PR #5803, opened 2020-01-26
fix: make cornerRadius work for stacked bars with timeUnit

### LINKED ISSUE #5800: cornerRadius doesn't work for stacked bars with timeUnit
Failing spec:

```json
{
  "mark": {"cornerRadiusTopLeft": 2, "cornerRadiusTopRight": 2, "type": "bar"},
  "encoding": {
    "x": {"field": "date", "type": "temporal", "timeUnit": "year"},
    "y": {"field": "price", "type": "quantitative", "aggregate": "average"},
  }
}
```

Symptom: bars don't show up at all.
Note: If we remove the cornerRadius, the spec will work properly.

kanitw:
fix #5800

image

### DIFF
--- a/examples/compiled/stacked_bar_count_corner_radius_config.vg.json
+++ b/examples/compiled/stacked_bar_count_corner_radius_config.vg.json
@@ -48,7 +48,7 @@
         "facet": {
           "data": "source_0",
           "name": "stack_group_main",
-          "groupby": ["month_date"],
+          "groupby": ["month_date", "month_date_end"],
           "aggregate": {
             "fields": [
               "__count_start",
--- a/src/compile/mark/mark.ts
+++ b/src/compile/mark/mark.ts
@@ -174,8 +174,8 @@ function getStackGroups(model: UnitModel) {
       }
     }
 
+    // For bin and time unit, we have to add bin/timeunit -end channels.
     const groupByField = model.fieldDef(model.stack.groupbyChannel);
-    // For bin we have to add bin channels.
     const groupby: string[] = vgField(groupByField) ? [vgField(groupByField)] : [];
 
     if (groupByField?.bin || groupByField?.timeUnit) {
--- a/src/compile/mark/mark.ts
+++ b/src/compile/mark/mark.ts
@@ -174,12 +174,12 @@ function getStackGroups(model: UnitModel) {
       }
     }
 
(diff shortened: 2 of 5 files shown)


src/compile/mark/mark.ts
```diff
@@ -174,12 +174,12 @@ function getStackGroups(model: UnitModel) {
       }
     }
 
+    const groupByField = model.fieldDef(model.stack.groupbyChannel);
     // For bin we have to add bin channels.
```
donghaoren: Consider moving this comment to before the if statement, and describe timeUnit?

examples/compiled/stacked_bar_count_corner_radius_config.vg.json
```diff
@@ -48,7 +48,7 @@
         "facet": {
           "data": "source_0",
           "name": "stack_group_main",
-          "groupby": ["month_date"],
+          "groupby": ["month_date", "month_date_end"],
```
donghaoren: For this particular spec, `month_date_end` is not being used, but I guess it doesn't hurt to add it there?

src/compile/mark/mark.ts
kanitw: Why moving?   Unlike Vega codebase, we prefer to declare things where they are used.

examples/compiled/stacked_bar_count_corner_radius_config.vg.json
kanitw: Yeah, probably doesn't matter. We do generate this as an output from time unit earlier anyway.

src/compile/mark/mark.ts
kanitw: (if you agree, please go ahead and merge. :D)

donghaoren: I thought the comment is describing the what the if statement is doing?

kanitw: oh you mean this if below.

kanitw:
> I just found in the current version (before this PR), if I use "x": {"timeUnit": "month", "field": "date", "type": "ordinal"} it's going to work, but "x": {"timeUnit": "month", "field": "date", "type": "temporal"} is not going to work. It seems like bin is set differently based on whether the type is temporal or ordinal?

Yes, ordinal timeUnit would use ordinal scales rather than continuous scales. :)

PR #5808, opened 2020-01-27
refactor: make channel type more specific

### LINKED ISSUE #5173: Make channel type more specific
In https://github.com/vega/vega-lite/blob/75cbdd52c19ebab8e2b55a101ca1aa782d0ed686/src/encoding.ts#L377, we assume that the channel is of type `Channel | string`. However, it should be of type `keyof Encoding<string>` and `isChannel` should also be a guard for `keyof Encoding<string>`. This way, we would know the type of `channelDef` in https://github.com/vega/vega-lite/blob/75cbdd52c19ebab8e2b55a101ca1aa782d0ed686/src/encoding.ts#L406. 

This is currently blocking https://github.com/vega/vega-lite/pull/5005.

domoritz:
Basically, the idea is to have `Channel` be the type for `Encoding` (normalized) rather than `ExtendedEncoding`.

fixes #5173

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -9,7 +9,14 @@ import {Mark} from './mark';
 import {EncodingFacetMapping, EncodingFacetMapping as ExtendedFacetMapping} from './spec/facet';
 import {Flag, keys} from './util';
 
-export type Channel = keyof Encoding<any> | keyof ExtendedFacetMapping<any>;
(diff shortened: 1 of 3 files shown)


domoritz:
Replaced by https://github.com/vega/vega-lite/pull/6712

PR #5811, opened 2020-01-28
feat: add `band` for `x2`/`y2` 

kanitw:
So we can draw rect from the start of timeunit to its end


src/compile/mark/encode/valueref.ts
```diff
@@ -193,30 +192,33 @@ export function midPoint({
...
-            const channel2 = channel === X ? X2 : Y2;
-            log.warn(log.message.channelRequiredForBinned(channel2));
+      if (isBinning(channelDef.bin) || (band && channelDef.timeUnit)) {
+        // Use middle only for x an y to place marks in the center between start and end of the bin range.
+        // We do not use the mid point for other channels (e.g. size) so that properties of legends and marks match.
+        if (
+          (contains([X, Y], channel) &&
+            isTypedFieldDef(channelDef) &&
+            contains([QUANTITATIVE, TEMPORAL], channelDef.type)) ||
+          contains([X2, Y2], channel)
```
kanitw: Need to check what to do for ordinal case

examples/compiled/layer_falkensee.vg.json
```diff
@@ -127,7 +127,9 @@
         "update": {
           "fill": {"scale": "color", "field": "event"},
           "x": {"scale": "x", "field": "year_start"},
-          "x2": {"scale": "x", "field": "year_end"},
+          "x2": {
```
kanitw: This is definitely wrong.

kanitw:
Let's revisit this in https://github.com/vega/vega-lite/issues/6248

PR #5812, opened 2020-01-28
docs: fix {:#repeat-ref} anchor in field.md

keckelt:
This `repeat-ref` keyword  is [showing up in the docs](https://vega.github.io/vega-lite/docs/field.html):
image

As the number is in the code anyway, I thought it is safe to remove it but it is referenced in [vega-lite/site/_data/link.yml#L299-L300](https://github.com/keckelt/vega-lite/blob/e0c6eac22954d91dfe0061f73e927cd0a6d9e498/site/_data/link.yml#L299-L300) and I'm not really sure what that does.

I can remove the anchor there as well or change the code here so that the anchor works again (with a little help 😉  )


kanitw:
The right fix is to add line breaks between the headers and the anchor.

kanitw:
Thanks for noticing the issue! :)

PR #5780, opened 2020-01-21
fix: properly support utc timeunits.

haldenl:
Fixes #5777

PR #5821, opened 2020-01-29
docs: correct + update utc related examples

haldenl:
Looks like #5780 didn't properly regenerate examples.


domoritz:
Screen Shot 2020-01-29 at 11 03 01

kanitw:
The examples are correct, not sure which one is incorrect. I'd recommend looking at each of them in the docs.

domoritz:
Okay. I was just confused that the first example changes the label angle. It doesn't look good to me.

kanitw:
> Okay. I was just confused that the first example changes the label angle. It doesn't look good to me.

The old label was bad anyway.

PR #5826, opened 2020-01-30
feat: support signal ref in (conditional) channel def

kanitw:


src/channeldef.ts
```diff
@@ -75,40 +76,53 @@ export type ValueDefWithCondition<F extends FieldDef<any>, V extends ValueOrGrad
...
+
+export type NumericValueOrSignalWithCondition<F extends Field> = ValueOrSignalWithCondition<
   MarkPropFieldDef<F, StandardType>,
   number
 >;
 
 export type TypeForShape = 'nominal' | 'ordinal' | 'geojson';
 
-export type ShapeValueDefWithCondition<F extends Field> = StringValueDefWithCondition<F, TypeForShape>;
+export type ShapeValueOrSignalWithCondition<F extends Field> = StringValueOrSignalWithCondition<F, TypeForShape>;
```
kanitw: Question: Should we be accurate and use the name `Shape(/Color/...)ValueOrSignalWithCondition` ?

Or should we preserve the old name (e.g., `ShapeValueDefWithCondition`) as we currently hide signals?

domoritz: Can we name the type after what it is used for?

kanitw: Not sure what you mean, but thinking more,  the right thing to do is to keep the accurate name in the TS code.  

If we want to hide "signals", we should do that in the renaming process, so it's not confusing when we read TS code.

PR #5835, opened 2020-02-01
chore: bump vega-typings

kanitw:


yarn.lock
```diff
@@ -7885,13 +7885,20 @@ vega-transforms@4.6.0:
...
 
-vega-typings@0.12.0, vega-typings@~0.12.0:
+vega-typings@0.12.0:
   version "0.12.0"
   resolved "https://registry.yarnpkg.com/vega-typings/-/vega-typings-0.12.0.tgz#bf998252b7f72bb2f22d6d5a02813e0bd3685562"
   integrity sha512-K+IoUTTtXW3E1Qhr/y+JgLRxy476viAm6DeM8IiVrA8vvuLA3FTzHaeI7TCnaWEwk9xxLJBtdVKKC5FGbp0Nyw==
   dependencies:
     vega-util "^1.12.1"
 
+vega-typings@~0.12.1:
```
kanitw: @domoritz  this is a bit weird?

domoritz: That's from Vega. 

Screen Shot 2020-02-01 at 18 51 17

kanitw: yeah, but what should we do in this case?

PR #5855, opened 2020-02-04
feat: Experimental Signal Support for Signal Ref in Vega-Lite

kanitw:
- [x] make sure we don't pollute the schema

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -2469,8 +2469,8 @@
     "ColorGradientFieldDefWithCondition": {
       "$ref": "#/definitions/FieldDefWithCondition<MarkPropFieldDef,(Gradient|string|null)>"
     },
-    "ColorGradientValueOrSignalWithCondition": {
-      "$ref": "#/definitions/ValueOrSignalWithCondition<MarkPropFieldDef,(Gradient|string|null)>"
+    "ColorGradientValueWithCondition": {
+      "$ref": "#/definitions/ValueWithCondition<MarkPropFieldDef,(Gradient|string|null)>"
     },
     "ColorName": {
       "enum": [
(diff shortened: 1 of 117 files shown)


build/vega-lite-schema.json
```diff
@@ -2984,7 +2689,7 @@
               "$ref": "#/definitions/NumericFieldDefWithCondition"
             },
             {
-              "$ref": "#/definitions/NumericValueDefWithCondition"
+              "$ref": "#/definitions/NumericValueWithCondition"
```
kanitw: FYI, this is the only con I have about this PR, we need to rename `*ValueOrSignal` to something else.

However, we can't use ValueDef because 
given the hiding, `NumericValueOrSignalWithCondition` just refers to 
`NumericValueDefWithCondition` that already exists. We can't use the same name twice. Doing so will make JSON schema invalid.

kanitw:
FYI, I have double-checked, this PR should be free from introducing "signal" to the schema, except for one declaration that is not referenced anywhere (that's a schema generator bug).

To facilitate future revert, please do not squash.

build/vega-lite-schema.json
```diff
@@ -798,7 +798,7 @@
         "labelPadding": {
```
domoritz: Can we remove the schema updates from this commit so we can revert it easily?

src/axis.ts
```diff
@@ -239,12 +251,12 @@ export interface Axis extends AxisConfigBaseWithConditional, Guide {
    *
    * __Default value__: `undefined`
    */
-  tickMinStep?: number;
+  tickMinStep?: number | SignalRef; // Vega already supports Signal
```
domoritz: What does the comment mean?

src/compile/axis/parse.ts
```diff
@@ -246,7 +247,9 @@ function parseAxis(channel: PositionScaleChannel, model: UnitModel): AxisCompone
         axisComponent.set(property, configValue, false);
       }
     } else if (
-      isConditionalAxisValue<any>(configValue) // need to cast as TS isn't smart enough to figure the generic parameter type yet
+      isConditionalAxisValue<any>(configValue) ||
+      isSignalRef(configValue)
+      // need to cast as TS isn't smart enough to figure the generic parameter type yet
```
domoritz: I don't see a cast.

src/legend.ts
```diff
@@ -88,12 +89,12 @@ export interface Legend extends BaseLegendNoSignals, LegendMixins, Guide {
    *
    * __Default value__: `undefined`
    */
-  tickMinStep?: number;
+  tickMinStep?: number | SignalRef;
```
domoritz: This is also a Vega property, right? Why do we need to override here?

build/vega-lite-schema.json
kanitw: It won't help anyway because we change import in many places later. 

The one that should be reverted is "fix: casting"

src/axis.ts
kanitw: But it is not a part of BaseAxis that we derive from.

kanitw: We should change Vega-typings such that Axis's extension could be refered in VL, but I don't have a good name for the extension (...) below.

```
export interface Axis extends BaseAxis {
  ...
}
```

In any case, I think that's beyond the scope of this PR.

src/compile/axis/parse.ts
kanitw: will fix.

src/legend.ts
kanitw: Same issue with axis, let's handle later.

PR #5861, opened 2020-02-05
fix: make offset work for rect band position

kanitw:


kanitw:
The error is weird and seems irrelevant to this PR:

```
[!] Error: 'hasOwnProperty' is not exported by node_modules/vega-util/build/vega-util.js
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module
build/src/util.js (5:9)
3: import deepEqual_ from 'fast-deep-equal';
4: import stableStringify from 'fast-json-stable-stringify';
5: import { hasOwnProperty, isNumber, isString, splitAccessPath, stringValue, writeConfig } from 'vega-util';
  
```

Issue #5838, opened 2020-02-02
Vega-Lite 4.1.1: ref names url-encode special characters


jakevdp:
This means that automated schema validation tools like Python's ``jsonschema`` package do not work correctly.

Example:

Here is a reference:

https://github.com/vega/schema/blob/34f2d40982b3b2dd3ae69322d49487a50f61fe8f/vega-lite/v4.1.1.json#L2223

Here is the definition that should match in a string equality sense:

https://github.com/vega/schema/blob/34f2d40982b3b2dd3ae69322d49487a50f61fe8f/vega-lite/v4.1.1.json#L5630

domoritz:
That’s unfortunate. The RFC says that references need to be URIs but many schema validates don’t support it. I’m inclined to revert this change but then we have a schema that’s invalid according to the standard...

Anyway, I think I’ll revert until at least VSCode catches up (they don’t support encoded references either).

domoritz:
We should file an issue with the json schema python package, though.

jakevdp:
Oh, OK I'll double-check. I didn't dig into precisely why it broke Altair (it may be my code not jsonschema) but this URL encoding of references definitely has not been part of the schema prior to 4.1.

Was it an intended change, then?

jakevdp:
Ah, looks like jsonschema itself is fine, I just need to change how Altair handles this:
```python
import jsonschema

schema = {
    "definitions": {
        "Foo<hello>": {"type": "string"}
    },
    "properties": {
        "x": {"$ref": "#/definitions/Foo<hello>"},
        "y": {"$ref": "#/definitions/Foo%3Chello%3E"}
    }
}

spec = {"x": "x value", "y": "y value"}

jsonschema.validate(spec, schema)
```

I'll wait until you confirm this was an intended change before I start adapting Altair to handle this.

domoritz:
I mentioned this a few weeks ago on slack. It’s a fix since the schema previously wasn’t valid. We can roll this change back for now if it’s causing issues.

jakevdp:
Turns out this is pretty straightforward to accommodate within Altair now that I know what's going on.

Sorry for the noise!

domoritz:
Thanks for updating Altair.

PR #5868, opened 2020-02-06
docs: fix references

domoritz:

PR #5869, opened 2020-02-06
docs: fix references, simplify schema

domoritz:
Improves #5868


domoritz:
Fixes https://github.com/vega/vega-lite/issues/5838

Issue #5871, opened 2020-02-07
More axis styling support


kanitw:
Just like how we have `axisBand`, we may want to customize our axis differently for different types of axes.

- [x] Like `axisBand`, we should add `config.axisQuantitative`, `config.axisTemporal` (time and utc)

- [ ] Moreover, for the same type of axis, one may want to apply various axis style, so we could add "style" to axis as mixins just like mark style (e.g., `axis: {style: 'foo'}`)

- [ ] The open question is how to specify axis style in the config as 
config.style is currently only for marks. Two options:
  - a) config.axisStyle["name"] - seperate from mark style
  - b) Use `config.style["name"]` like mark style, but apply them correctly depending on who refer to it.

@donghaoren @domoritz and I prefer b) since it will be easier to teach that all style properties read from `config.style`.  (On the other hand, it's weird to teach that style in axes is from "axisStyle" while style in marks is from just plain "style" without a prefix.)

kanitw:
Will be fixed in #5916

PR #5873, opened 2020-02-07
feat: support config.axisQuantitative/Temporal

kanitw:
Part of https://github.com/vega/vega-lite/issues/5871

### DIFF
--- a/src/compile/axis/config.ts
+++ b/src/compile/axis/config.ts
@@ -1,6 +1,6 @@
 import {PositionScaleChannel} from '../../channel';
 import {Config} from '../../config';
-import {ScaleType} from '../../scale';
+import {isContinuousToContinuous, ScaleType} from '../../scale';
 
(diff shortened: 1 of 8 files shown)


src/compile/axis/config.ts
```diff
@@ -12,6 +12,10 @@ export function getAxisConfig(
   // configTypes to loop, starting from higher precedence
   const configTypes = [
     ...(scaleType === 'band' ? ['axisBand'] : []),
+    ...(isContinuousToContinuous(scaleType) ? ['axisQuantitative'] : []),
```
domoritz: `isContinuousToContinuous` returns true for time and utc scales. Is this intended?

kanitw: Right, this is dumb.  We should fix this.

domoritz: @donghaoren realized this

PR #5874, opened 2020-02-08
chore: update schema generator

domoritz:
This version of the schema generator supports @hidden for types.


build/vega-lite-schema.json
```diff
@@ -3388,10 +3388,10 @@
     "ConditionalValueDef<number[]>": {
       "anyOf": [
         {
-          "$ref": "#/definitions/ConditionalPredicate%3CValueDef%3Cnumber%5B%5D%3E%3E"
```
domoritz: Not sure why these were encoded before.

PR #5877, opened 2020-02-09
refactor: fix comments for mark props

domoritz:


kanitw:
This isn't really docs, just internal comments.

PR #5882, opened 2020-02-11
fix: don't output gradientLength when unnecessary

kanitw:


domoritz:
Fixes https://github.com/vega/vega-lite/issues/4508

PR #5886, opened 2020-02-11
docs: improve scale domain docs

kanitw:


src/scale.ts
```diff
@@ -435,16 +435,19 @@ export interface Scale {
   type?: ScaleType;
 
   /**
-   * Customized domain values.
+   * Customized domain values, either constant values or dynamic values driven by a selection.
```
domoritz:
```suggestion
   * Customized domain values, either constant values, or dynamic values driven by a selection.
```

src/scale.ts
```diff
@@ -435,16 +435,19 @@ export interface Scale {
...
    *
-   * For _quantitative_ fields, `domain` can take the form of a two-element array with minimum and maximum values. [Piecewise scales](https://vega.github.io/vega-lite/docs/scale.html#piecewise) can be created by providing a `domain` with more than two entries.
-   * If the input field is aggregated, `domain` can also be a string value `"unaggregated"`, indicating that the domain should include the raw data values prior to the aggregation.
+   * 1) Constant `domain` for _quantitative_ fields can take one of the following forms:
    *
-   * For _temporal_ fields, `domain` can be a two-element array minimum and maximum values, in the form of either timestamps or the [DateTime definition objects](https://vega.github.io/vega-lite/docs/types.html#datetime).
+   * - a two-element array with minimum and maximum values.
+   * - an array with more than two entries, for [Piecewise  quantitative scales](https://vega.github.io/vega-lite/docs/scale.html#piecewise). (Alternatively, the `domainMid` property can be set for a diverging scale.)
```
domoritz:
```suggestion
   * - an array with more than two entries, for [Piecewise quantitative scales](https://vega.github.io/vega-lite/docs/scale.html#piecewise). Alternatively, the `domainMid` property can be set for a diverging scale.
```

src/scale.ts
```diff
@@ -435,16 +435,19 @@ export interface Scale {
...
+   * - a string value `"unaggregated"`, if the input field is aggregated, to indicate that the domain should include the raw data values prior to the aggregation.
    *
-   * For _ordinal_ and _nominal_ fields, `domain` can be an array that lists valid input values.
+   * 2) Constant `domain` for _temporal_ fields can be a two-element array minimum and maximum values, in the form of either timestamps or the [DateTime definition objects](https://vega.github.io/vega-lite/docs/types.html#datetime).
```
domoritz:
```suggestion
   * 2) Constant `domain` for _temporal_ fields can be a two-element array with minimum and maximum values, in the form of either timestamps or the [DateTime definition objects](https://vega.github.io/vega-lite/docs/types.html#datetime).
```

src/scale.ts
kanitw: I think "either" better has no comma between.

domoritz: Right. My bad.

Issue #5823, opened 2020-01-29
Ability to augment domain 


kanitw:
Currently we can either set scale `domain` to a fixed value or make it data driven. 

However, in streaming settings, one might want to set initial/default domain, but allows domain to exceed the initial/default domain. 

For example, if we set this initial domain to [0, 100] and all data falls within [0,100], then that's the domain.

```
{
  "data": {
    "values": [
      {"a": "A", "b": 28}, 
      {"a": "B", "b": 55}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "a", "type": "ordinal"},
    "y": {"field": "b", "type": "quantitative", "scale": {"domainInit/Default/???": [0, 100]}}
  }
}
```

![image](https://user-images.githubusercontent.com/111269/73401398-eeb08300-429f-11ea-9625-70a766333e63.png)

but if the data include values beyond the initial domain, the domain should then be updated:

![image](https://user-images.githubusercontent.com/111269/73401470-0f78d880-42a0-11ea-8081-a3b97f2a96e8.png)

We can already achieve this in Vega by doing:
```
domain: {
        "fields": [
          {"data": "data_0", "field": "b"},
          [0, 100]
        ]
      }
````

However, we need a way to specify this in Vega-Lite.

There are a few possible ways:

1) Add a new `domainXXX` property (`domainBase/domainInit/domainDefault`) for this method

2) Add another flag to read the field (e.g., `domainFromData` or something similar)

3) Allow nesting to blend 1 or 2.  

```
domain: {
  values: [0, 100]
  data: true,
}
```

**Comment by kanitw** [comment]
4. The other way to think about this is to think of default domain as `domain: {data: true}` or something. 

we could then support:

```
domain: {
  union: [{data: true}, [0, 100]]
} 
```

domoritz:
I don't think people think of unioning domains so maybe we should name it slightly differently.

kanitw:
I don't think I agree.  Domain and range are already math-terms, so union would still make sense?.  

That being said, option 4 is honestly probably not a good idea since no body would need it besides to merge with data domain. 

5. The other way would be to call it `domainBase` or something similar -- basically having domainBase = [0, 100] means domainMin = min(0, min of data) and domainMax = max(0, max of data).

kanitw:
Maybe `domainPad`? because it's basically scale padding, but directly on the domain value.

domoritz:
[...]

PR #5887, opened 2020-02-11
feat: add domain unionWith support

kanitw:
Fix https://github.com/vega/vega-lite/issues/5823

Merge https://github.com/vega/vega-lite/pull/5886 first

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -4993,6 +4993,44 @@
       ],
       "type": "string"
     },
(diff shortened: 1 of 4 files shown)


domoritz:
Need to rebase.

domoritz:
LGTML but please add an example to the docs.

kanitw:
It's a very special case, I don't know if we need a whole example for it.  Feel free to submit a PR if you think we really need one.

PR #5585, opened 2019-11-23
feat: support system colors

domoritz:
this PR ports https://github.com/vega/vega/pull/2088 to Vega-Lite. 

Test config

```json
{
  "signals": [
    {
      "name": "colors",
      "value": {
        "blue": "red",
        "orange": "#f58518",
        "red": "#e45756",
        "teal": "#72b7b2",
        "green": "#54a24b",
        "yellow": "#eeca3b",
        "purple": "#b279a2",
        "pink": "#ff9da6",
        "brown": "#9d755d",
        "gray0": "red",
        "gray1": "#111",
        "gray2": "#222",
        "gray3": "#333",
        "gray4": "#444",
        "gray5": "#555",
        "gray6": "#666",
        "gray7": "#777",
        "gray8": "yellow",
        "gray9": "#999",
        "gray10": "#aaa",
        "gray11": "#bbb",
        "gray12": "#ccc",
        "gray13": "orange",
        "gray14": "#eee",
        "gray15": "#fff"
      }
    }
  ]
}
```

Example

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
  "width": 200,
  "height": {"step": 16},
  "data": {"url": "data/movies.json"},
  "encoding": {
    "y": {
      "field": "Major_Genre",
      "type": "nominal",
      "axis": null
    }
  },
  "layer": [{
    "mark": {"type": "bar", "color": "colors.gray13"},
    "encoding": {
      "x": {
        "aggregate": "mean",
        "field": "IMDB_Rating",
        "type": "quantitative",
        "scale": {"domain": [0, 10]},
        "title": "Mean IMDB Ratings"
      }
    }
  }, {
    "mark": {"type": "text", "align": "left", "x": 5},
    "encoding": {
      "text": {
        "field": "Major_Genre",
        "type": "nominal"
      },
      "detail": {"aggregate": "count", "type": "quantitative"}
    }
  }]
}
```

image

PR #5888, opened 2020-02-11
feat: support (experimental) signal-based color/fontSize config

kanitw:
replaces #5585

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,4 +1,5 @@
-import {Color, RangeConfig, RangeScheme} from 'vega-typings';
+import {scheme} from 'vega-scale';
+import {Color, InitSignal, NewSignal, RangeConfig, RangeScheme} from 'vega-typings';
 import {isObject, mergeConfig} from 'vega-util';
 import {AxisConfigMixins, isConditionalAxisValue} from './axis';
 import {CompositeMarkConfigMixins, getAllCompositeMarks} from './compositemark';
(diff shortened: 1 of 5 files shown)


src/config.ts
```diff
@@ -265,8 +278,123 @@ export const defaultConfig: Config = {
...
+      stroke: {signal: 'gridColor'}
+    }
+  },
+  axis: {
+    domainColor: {signal: 'domainColor'},
+    gridColor: {signal: 'gridColor'},
+    tickColor: {signal: 'domainColor'}
+  },
+  legend: {
+    gradientStrokeColor: 'gridColor',
```
kanitw: we definitely re-use domainColor for both tick and domain by default, so it might make sense to define this `domainColor` signal, but I'm not sure about `gridColor`.

src/config.ts
```diff
@@ -103,7 +104,25 @@ export function isVgScheme(rangeScheme: string[] | RangeScheme): rangeScheme is
   return rangeScheme && !!rangeScheme['scheme'];
 }
 
+export type ColorConfig = {[name: string]: Color};
+
+export type FontSizeConfig = {[name: string]: number};
+
 export interface VLOnlyConfig {
+  /**
+   * Default color signals
```
domoritz:
```suggestion
   * Default color signals.
```

src/config.ts
```diff
@@ -103,7 +104,25 @@ export function isVgScheme(rangeScheme: string[] | RangeScheme): rangeScheme is
...
+   *
+   * @hidden
+   */
+  color?: boolean | ColorConfig;
+
+  /**
+   * Default font size signals
```
domoritz:
```suggestion
   * Default font size signals.
```

src/config.ts
```diff
@@ -188,6 +207,11 @@ export interface Config
 
   /** An object hash that defines key-value mappings to determine default properties for marks with a given [style](https://vega.github.io/vega-lite/docs/mark.html#mark-def). The keys represent styles names; the values have to be valid [mark configuration objects](https://vega.github.io/vega-lite/docs/mark.html#config). */
   style?: StyleConfigIndex;
+
+  /**
+   * @hide
```
domoritz:
```suggestion
   * @hidden
```

PR #5890, opened 2020-02-11
feat: support config.font

kanitw:
Note that since this feature doesn't involve signal, I'm not hiding it.


kanitw:
image

domoritz:
Hmm, I'd like to hide it until we get an okay from @arvind and @jheer. In the near future, we may want to do this in Vega.

Issue #5854, opened 2020-02-03
Vega-Lite 4.1.1: some layered specs are invalid


jakevdp:
Several dozen of Altair's examples are invalid under the Vega-Lite 4.1.1 schema; the common feature seems to be that they contain layered specs. An example is the [Bar and Tick Chart](https://altair-viz.github.io/gallery/layered_chart_bar_mark.html) (view in [vega editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAutIC4yghSA3WAbArgKYDOKYA2uBhMJAOYD2eZATAKwA0UADgE4MArQgGN4ZSEi6Riwhr0Ks2AXw5VqtRs1QAWAOxS+gkWNSQARpCky5C1G10q1GDU1xkAzAAYD-IaPHCllDW8qzujtQ09K6s+jy+xuLQQdKyoagAjACcEZEuWmDuABw+Rv6mClZptmBFuerRBdmlfiZQAGYpITV69c6NbjotiaZ0XdVkdU4AulS5kLiwAJ6EvGSU6k6YhAB2stAAljtjqOiRmAAeZLTtB4S4yaaGrSnwS9w1kDsMALZHzH1qJAltdILd7o9ghMpG8PuIAI74WA7eAHRCorAKJRbMCAqA-WC8ADW4jMhMsOMgAHcDtB4AALUHEeCEbhkbSebHUPpnIG7fZHE5oHGXUHgh7iZ6JGHvT7fP47AGqc6YEGnMF3CWjGIyuGmRHI1Hog6YyBcyJ4yAE4mg2S4OTieTJKwHABeNXcADYYfSDsIiTsSKRUCxdZ9Uf6zU5zbMlEA)).

Here is how I checked the validity under the schema; it results in a validation error under the v4.1.1 schema, but is fine under the v4.0.2 schema.
```python
import jsonschema
import json
import urllib.request
with urllib.request.urlopen('https://vega.github.io/schema/vega-lite/v4.1.1.json') as f:
    schema = json.load(f)

spec = {
  "data": {
    "values": [
      {"goal": 25, "project": "a", "score": 25},
      {"goal": 47, "project": "b", "score": 57},
      {"goal": 30, "project": "c", "score": 23},
      {"goal": 27, "project": "d", "score": 19},
      {"goal": 38, "project": "e", "score": 8},
      {"goal": 19, "project": "f", "score": 47},
      {"goal": 4, "project": "g", "score": 8}
    ]
[...]

PR #5891, opened 2020-02-12
fix: revert #5797

domoritz:
Fixes https://github.com/vega/vega-lite/issues/5854


domoritz:
Alternatively, we could add `width` and `height` to unit specs and mark them as deprecated.

Issue #5893, opened 2020-02-12
Log scale breaks rule mark  


christophe-petitjean:
If I switch y encoding to log scale when I have two layers (1 rule, 1 tick), the rule is coming from the top instead of starting from the bottom of the chart. If I use linear scale no problem, no problem either If I use a bar mark instead of a rule.

<img width="1672" alt="image" src="https://user-images.githubusercontent.com/354149/74327849-2d106c80-4d8d-11ea-8ed0-3ed589b3a715.png">

<img width="1403" alt="image" src="https://user-images.githubusercontent.com/354149/74327869-37cb0180-4d8d-11ea-921f-f1dbaca9e2f1.png">

[ Spec I used ](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAusDOBTeTIC4yghSAJADYCeAjFmANrh4TCQBKKADgPZICW8bATiZUgBhBCgDmfEmAoAaKELYBXAHbxKZAJxkAvjNp0Gzdlx79BI+OMlgATJDnClq9QAYXu-XkOsO3SedEJfjAAZnt5JzVsADYPOnomHxN-bGFA6wAWcMcVKLBYvXicROM-M1SLK2CAVmyFXNdtTwBdWjioAHdOaHgAC0oMgHYADkKoXpROMV68jOHBscgAYzZlADMpygYAN04UDq3IJHheNgBrFEET2GUkFlheFGdtdpgEWEPlWABbS9TiciQV5PFbQTjKMRbTyQAAeUKKkA2KCI0EERl8pgEYzoRz4eQYSJR5ki2T40BQvEE5KQSyeYIhQOxeEg8BILD+UGUbG+4NgREgnna+AE2Fw8SOSz5HIYrPZWEgRDYkKFzMJqIqJKZ+FlHMgAEdFDd4NwEJxtpdBUzlmxFZTRYi9kTUujkmYHDrBFyeV9+U0IK8iLASBTKDQDJ58N8HmdBLxFEQLXQVWKcVHeDHUsaljHPILaK1tEA)

domoritz:
Here is a simplified spec.

```json
{
  "data": {
    "values": [
      {"Repository": "Category 1", "Count": 191},
      {"Repository": "Category 2", "Count": 100},
      {"Repository": "Category 3", "Count": 6},
      {"Repository": "Category 4", "Count": 6},
      {"Repository": "Category 5", "Count": 10}
    ]
  },
  "encoding": {
[...]

PR #5897, opened 2020-02-12
fix: warn about using rule with log scales

domoritz:
Fixes https://github.com/vega/vega-lite/issues/5893


src/compile/mark/rule.ts
```diff
@@ -27,8 +27,8 @@ export const rule: MarkCompiler = {
         range: orient !== 'vertical' // include x2 for horizontal or line segment rule
       }),
       ...encode.pointOrRangePosition('y', model, {
-        defaultPos: orient === 'vertical' ? 'zeroOrMin' : 'mid',
-        defaultPos2: 'zeroOrMax',
+        defaultPos: orient === 'vertical' ? 'zeroOrMax' : 'mid',
```
kanitw: Why do we need to swap this? and only for Y, not X?

domoritz: We should swap those as well

kanitw: Ok, but that still doesn't answer why we need to swap.  I'm sure discover an example where it matters?

PR #5898, opened 2020-02-12
fix: bring back width/height/view for unit specs inside layers (as deprecated)

kanitw:
Fix https://github.com/vega/vega-lite/issues/5854

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -7943,72 +7943,6 @@
       ],
       "description": "Any specification in Vega-Lite."
     },
-    "GenericUnitSpec<Encoding,AnyMark>": {
-      "additionalProperties": false,
-      "description": "Base interface for a unit (single-view) specification.",
-      "properties": {
-        "data": {
-          "anyOf": [
-            {
-              "$ref": "#/definitions/Data"
-            },
-            {
-              "type": "null"
-            }
-          ],
-          "description": "An object describing the data source. Set to `null` to ignore the parent's data source. If no data is set, it is derived from the parent."
-        },
-        "description": {
-          "description": "Description of this mark for commenting purpose.",
-          "type": "string"
-        },
-        "encoding": {
-          "$ref": "#/definitions/Encoding",
-          "description": "A key-value mapping between encoding channels and definition of fields."
-        },
-        "mark": {
-          "$ref": "#/definitions/AnyMark",
-          "description": "A string describing the mark type (one of `\"bar\"`, `\"circle\"`, `\"square\"`, `\"tick\"`, `\"line\"`,\n`\"area\"`, `\"point\"`, `\"rule\"`, `\"geoshape\"`, and `\"text\"`) or a [mark definition object](https://vega.github.io/vega-lite/docs/mark.html#mark-def)."
-        },
-        "name": {
-          "description": "Name of the visualization for later reference.",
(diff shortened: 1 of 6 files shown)


src/spec/unit.ts
```diff
@@ -42,7 +42,7 @@ export type NormalizedUnitSpec = GenericUnitSpec<Encoding<Field>, Mark | MarkDef
 /**
  * A unit specification, which can contain either [primitive marks or composite marks](https://vega.github.io/vega-lite/docs/mark.html#types).
  */
-export type UnitSpec = GenericUnitSpec<CompositeEncoding, AnyMark>;
+export type UnitSpec = GenericUnitSpec<CompositeEncoding, AnyMark> & DeprecatedFrameMixins;
```
kanitw: It's a bit funny that UnitSpec and UnitSpecWithFrame are pretty much the same, but one has "deprecated frame mixins".

build/vega-lite-schema.json
```diff
@@ -16716,8 +16650,108 @@
...
+        "transform": {
+          "description": "An array of data transformations such as filter and new field calculation.",
+          "items": {
+            "$ref": "#/definitions/Transform"
+          },
+          "type": "array"
+        },
+        "view": {
+          "$ref": "#/definitions/ViewBackground",
+          "deprecated": "Please avoid using width in a unit spec that's a part of a layer spec."
```
kanitw: So @deprecated doesn't even show up in the description / editor.

Should we instead use `__Deprecated:__ <description>`?

domoritz: Hmm. Yes, we probably should. Eventually, it would be nice if the editor and the docs understood @deprecated and @default but I guess we don't have that for now.

src/spec/unit.ts
domoritz: Oh well.

src/spec/base.ts
```diff
@@ -113,6 +113,23 @@ export interface FrameMixins extends LayoutSizeMixins {
   view?: ViewBackground;
 }
 
+export type DeprecatedFrameMixins = {
+  /**
+   * __Deprecated:__ Please avoid using width in a unit spec that's a part of a layer spec.
```
domoritz:
```suggestion
   * __Deprecated:__ Please avoid using width in a unit spec that's part of a layer spec.
```

src/spec/base.ts
```diff
@@ -113,6 +113,23 @@ export interface FrameMixins extends LayoutSizeMixins {
...
+   */
+  width?: FrameMixins['width'];
+
+  /**
+   * __Deprecated:__ Please avoid using width in a unit spec that's a part of a layer spec.
```
domoritz:
```suggestion
   * __Deprecated:__ Please avoid using width in a unit spec that's part of a layer spec.
```

src/spec/base.ts
```diff
@@ -113,6 +113,23 @@ export interface FrameMixins extends LayoutSizeMixins {
...
+   */
+  height?: FrameMixins['height'];
+
+  /**
+   * __Deprecated:__ Please avoid using width in a unit spec that's a part of a layer spec.
```
domoritz:
```suggestion
   * __Deprecated:__ Please avoid using width in a unit spec that's part of a layer spec.
```

PR #5912, opened 2020-02-14
refactor: remove namespace usage

kristw:
`namespace` cause issue when trying to use the lib with `babel` + `typescript`

```ts
import { ScaleType } from 'vega-lite/build/src/scale';

const x = ScaleType.LINEAR;
// Error: Cannot read property 'LINEAR' of undefined
```

Please:

- [X] Make the pull requests (PRs) atomic. (Fix one issue at a time.) Multiple relevant issues that must be fixed together? Make atomic commits so we can easily review each issue.
- [X] Provide a concise title so we can easily copy it to the release note.
  - Use imperative mood and present tense.
  - Mention relevant issues. (e.g., `Fixes #1` / `Fixes part of #1`)
- [ ] Lint and test (Run `yarn test`)
- [X] Rebase onto the latest `master` branch.
- [] Review your changes before sending the PR (to ensure code quality).
- For new features:
  - [ ] Add new unit tests.
  - [ ] Update the documentation under `site/docs/` + add examples.

Tips:

- https://medium.com/@greenberg/writing-pull-requests-your-coworkers-might-enjoy-reading-9d0307e93da3 is a nice article about writing a nice PR.
- Use draft PR for work in progress PRs / when you want early feedback (https://github.blog/2019-02-14-introducing-draft-pull-requests/)

PR #5913, opened 2020-02-14
refactor: remove namespace usage

domoritz:
`namespace` cause issue when trying to use the lib with `babel` + `typescript`

```ts
import { ScaleType } from 'vega-lite/build/src/scale';

const x = ScaleType.LINEAR;
// Error: Cannot read property 'LINEAR' of undefined
```

### DIFF
--- a/src/scale.ts
+++ b/src/scale.ts
@@ -10,44 +10,32 @@ import {Type, TYPE_INDEX} from './type';
 import {contains, Flag, keys} from './util';
 import {ScaleInterpolate, ScaleInterpolateParams} from './vega.schema';
 
-export namespace ScaleType {
+type ValueOf<T> = T[keyof T];
+export const ScaleType = {
   // Continuous - Quantitative
-  export const LINEAR: 'linear' = 'linear';
-  export const LOG: 'log' = 'log';
-  export const POW: 'pow' = 'pow';
-  export const SQRT: 'sqrt' = 'sqrt';
-  export const SYMLOG: 'symlog' = 'symlog';
+  LINEAR: 'linear',
+  LOG: 'log',
+  POW: 'pow',
+  SQRT: 'sqrt',
+  SYMLOG: 'symlog',
+
   // Continuous - Time
-  export const TIME: 'time' = 'time';
-  export const UTC: 'utc' = 'utc';
+  TIME: 'time',
+  UTC: 'utc',
 
   // Discretizing scales
-  export const QUANTILE: 'quantile' = 'quantile';
-  export const QUANTIZE: 'quantize' = 'quantize';
-  export const THRESHOLD: 'threshold' = 'threshold';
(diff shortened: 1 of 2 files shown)


domoritz:
Replaces #5912

domoritz:
@kristw We need to fix https://github.com/vega/ts-json-schema-generator to support this change. 

```
Error: Unknown node "TimeUnit" (ts.SyntaxKind = 216) at /Users/dominik/Code/vega-lite/src/predicate.ts(50,13)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

kristw:
ok.... Need to dig into the schema generator. May take some time.

src/timeunit.ts
```diff
@@ -3,61 +3,62 @@ import {isObject, isString} from 'vega-util';
...
+  UTCHOURSMINUTESSECONDS: 'utchoursminutesseconds',
+  UTCMINUTESSECONDS: 'utcminutesseconds',
+  UTCSECONDSMILLISECONDS: 'utcsecondsmilliseconds',
+  UTCQUARTER: 'utcquarter',
+  UTCYEARQUARTER: 'utcyearquarter',
+  UTCQUARTERMONTH: 'utcquartermonth',
+  UTCYEARQUARTERMONTH: 'utcyearquartermonth'
+} as const;
+
+type ValueOf<T> = T[keyof T];
```
domoritz: Please extract this into utils.

kristw: Do you have a place for storing these util types?
I looked into `utils` and it is like a mixed bags of typeguards, etc. but if that is a preferred place, sure.

domoritz: Yeah, let's put it in utils.

src/scale.ts
```diff
@@ -10,44 +10,32 @@ import {Type, TYPE_INDEX} from './type';
...
-  | typeof ScaleType.TIME
-  | typeof ScaleType.UTC
-  | typeof ScaleType.QUANTILE
-  | typeof ScaleType.QUANTIZE
-  | typeof ScaleType.THRESHOLD
-  | typeof ScaleType.BIN_ORDINAL
-  | typeof ScaleType.ORDINAL
-  | typeof ScaleType.POINT
-  | typeof ScaleType.BAND;
+export type ScaleType = ValueOf<typeof ScaleType>;
```
domoritz, line 38: This is just `keyof typeof` so let me see how I can support that in the schema generator.

domoritz:
@kristw I updated the schema generator so we can get rid of namespaces now. As a heads up, I'm about to send a refactoring of time unit in a bit.

domoritz:
replaced by https://github.com/vega/vega-lite/pull/6355

src/scale.ts
domoritz: It's not. I was wrong.

PR #5914, opened 2020-02-14
fix: use record type and add missing properties to AXIS_PROPERTY_TYPE

domoritz:


src/axis.ts
```diff
@@ -353,10 +361,12 @@ export const AXIS_PROPERTY_TYPE: {
   titleX: 'main',
   titleY: 'main',
 
-  tickWidth: 'both',
+  encode: 'both', // we hide this in Vega-Lite
+  scale: 'both',
   tickCount: 'both',
+  tickWidth: 'both',
```
kanitw: why both for `tickWidth`?

src/axis.ts
```diff
@@ -321,30 +320,39 @@ export const AXIS_PROPERTY_TYPE: {
...
   labelOpacity: 'main',
   labelOverlap: 'main',
   labelPadding: 'main',
   labels: 'main',
+  labelSeparation: 'main',
   maxExtent: 'main',
   minExtent: 'main',
   offset: 'both',
   position: 'main',
+  tickBand: 'main',
```
kanitw: tickBand is a shorthand for bandPosition+tickExtra - it should be "both".

src/axis.ts
```diff
@@ -321,30 +320,39 @@ export const AXIS_PROPERTY_TYPE: {
...
   tickColor: 'main',
+  tickDash: 'main',
+  tickDashOffset: 'main',
   tickExtra: 'main',
```
kanitw: tickExtra actually should be "both".

domoritz:
@kanitw updated

PR #5916, opened 2020-02-15
feat: support axis `style` property, which reads from axis config

kanitw:
Fix https://github.com/vega/vega-lite/issues/5871


src/axis.ts
```diff
@@ -230,6 +230,14 @@ export interface AxisPropsWithConditionAndSignal {
 export type AxisConfig = VlOnlyGuideConfig & AxisConfigBaseWithConditionalAndSignal;
 
 export interface Axis extends AxisConfigBaseWithConditionalAndSignal, Guide {
+  /**
+   * A string or array of strings indicating the name of custom styles to apply to the axis. A style is a named collection of axis property defined within the [style configuration](https://vega.github.io/vega-lite/docs/mark.html#style-config). If style is an array, later styles will override earlier styles. Any [axis properties](https://vega.github.io/vega-lite/docs/encoding.html#mark-prop) explicitly defined within the `encoding` will override a style default.
+   *
+   * __Default value:__ (none)
+   * __Note:__ Any specified style will augment the default style. For example, a x-axis mark with `"style": "foo"` will receive from `config.axisX` and `config.style.foo` (the specified style `"foo"` has higher precedence).
```
```suggestion
   * __Note:__ Any specified style will augment the default style. For example, an x-axis mark with `"style": "foo"` will use `config.axisX` and `config.style.foo` (the specified style `"foo"` has higher precedence).
```

kanitw:
@domoritz  The cost of doing this shared `config.style` is that we need to do a lot of casting.

src/config.ts
```diff
@@ -171,7 +171,7 @@ export interface VLOnlyConfig {
 }
 
 export interface StyleConfigIndex {
-  [style: string]: MarkConfig;
+  [style: string]: MarkConfig | Axis;
```
domoritz: Let's add

```
  'group-title': MarkConfig
```

so we don't need to cast below

test/config.test.ts
```diff
@@ -151,13 +151,13 @@ describe('config', () => {
       expect(output.style.cell['height']).not.toBeDefined();
       expect(output.style.cell['fill']).toBe('#eee');
 
-      expect(output.style.bar.opacity).toEqual(0.5);
+      expect(output.style.bar['opacity']).toEqual(0.5);
```
domoritz: Should not be needed if we make the type more specific above.

test/config.test.ts
```diff
@@ -151,13 +151,13 @@ describe('config', () => {
...
     });
 
     it('should redirect config.title to config.style.group-title and rename color to fill', () => {
       expect(output).not.toHaveProperty('title');
-      expect(output.style['group-title'].fontWeight).toBe('bold');
-      expect(output.style['group-title'].fill).toBe('red');
+      expect(output.style['group-title']['fontWeight']).toBe('bold');
```
domoritz: same as above

PR #5920, opened 2020-02-15
feat: build ES5 bundle

terencehonles:
As per @domoritz comments in #5619 I updated the tsc config to output a ES5 compliant module. Unfortunately when rolling the package up with rollup `vega-utils` is still ES6 and babel is still needed like @nyurik did in #5619 to transpile *that* and possibly other ES6 dependencies.

I'm not sure if @domoritz would prefer this over #5619, but it's another stab at producing an ES5 compliant bundle.

Please:

- [x] Make the pull requests (PRs) atomic. (Fix one issue at a time.) Multiple relevant issues that must be fixed together? Make atomic commits so we can easily review each issue.
- [x] Provide a concise title so we can easily copy it to the release note.
  - Mention relevant issues. (e.g., `Fixes #1` / `Fixes part of #1`)
- [x] Lint and test (Run `yarn test`)
- [x] Rebase onto the latest `master` branch.
- [x] Review your changes before sending the PR (to ensure code quality).


babel.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
domoritz: Since we are already running tsc with es5 output, is it necessary to also run babel? Couldn't we just bundle up the generated files?

The issue you pointed at with vega-util could maybe be resolved by requiring es5 compiled filed from eslint in rollup. What do you think?

domoritz:
Just to explain my hesitation here. Maintaining Vega-Lite and all the related packages is taking up a lot of my time and any increase in complexity in the build process is putting an extra burden on me. I definitely want to support es5 but in a way where we don't add new compile paths that can break.

.gitignore
```diff
@@ -1,10 +1,12 @@
 _site/
 !build/vega-lite-schema.json
+!build-es5/vega-lite-schema.json
```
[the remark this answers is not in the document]
terencehonles: I don't know. I was just copying what was in the build in case it would also be needed

babel.config.js
terencehonles: I'm not sure how to get roll-up to pull the ES5 files if that's possible. I didn't know if all the dependencies had ES5 builds and how to pull the right files when building an ES5 bundle.

terencehonles:
I understand that, but inserting Babel into the tool chain seems like the least burdensome way to add an ES5 bundle. Without that I think you'll always have to be very careful what dependencies you're pulling in and how they're built. 

When implementing this PR I already saw how many moving parts there were. I thought just updating the tsc config would be sufficient, but I didn't realize the rolled up files weren't from tsc, but rollup. When I updated that config I was no longer building the ES5 bundle I was trying to build and it wasn't tsc's fault.

I'm not sure if tsc can roll up the files by itself, but it seems like it will do the right thing as far as it's output. I'm not sure if just changing the output format would work or if you actually need rollup.

Another thing to possibly look at is ng-packagr. The config for that is pretty terse, and the angular package format defines a number of bundles to produce which may cover what you're already generating and then add a couple more package formats including ES5. Obviously this is changing the build tooling, but it might make things simpler and allow a lot more ways to consume Vega lite.

domoritz:
Let's see. The easiest way, in this case, might be to just post-process the individual files in build/src and the compiled files. This could be a post-build step that's easy to manage since nothing else depends on it. How does that sound?

terencehonles:
The individual files in build/src are already ready to go based on the tsc config changes. The rolled up file can be post processed with the rollup Babel plugin or via Babel itself like in #5619. So to me it sounds like you're suggesting what this or #5619 is doing (correct me if I'm wrong)

Or are you just suggesting this should be at a different stage in the build pipeline?

domoritz:
Yep, I am proposing to do what https://github.com/vega/vega-lite/pull/5619 does but also for the files in build/src. Rather than going through a separate tsc step, though, I suggest to go trough babel. 

My thinking is that we are es6 for everything except the last step where we take stuff in `build` and transpile it to `es5-build`.

terencehonles:
@domoritz I'm not sure I'm following (and possibly because you closed #5619) did you want #5619 adjusted so it just transpiles all the files in the `build-es5/src` directory in addition to the `build-es5/vega-lite.js`? Is this so there isn't the TSC change? I think configuring babel to create ES5 copies of each of the files seems like more potential problems then having [the 1 TSC config file](https://github.com/vega/vega-lite/blob/ac33037d9791325134cc0f6c864c300cf38bf6a9/src/tsconfig.src-es5.json) which just sets a different output path and enables two options that are needed for ES5

With this change the files in the `build-es5/src/` **are** already ES5, but the problem is they still import modules which are not ES5, and those need to be transpiled (not the files in `build-es5/src/`). If all of the dependencies outputted a ES5 bundle I think you could change the rollup resolver to pick the right one (I believe that's essentially analogous how the Angular compiler works and why they created the Angular Package Format (and it does break down in cases where dependencies don't ship ES5 bundles in the APF))

domoritz:
I think we either 

1) transpile build/src and the bundle with babel (babel does es5 -> es5 for everything) or
2) transpile ts -> es5 js and add a rollup config to bundle with es5 files from all dependencies (no babel involved)

I'm sorry that I went back and forth on this a bit but there were some new things I learned when looking at the different PRs.

terencehonles:
For 1. That seems to still involve TSC, Babel, and rollup (unless I'm misunderstanding), and not sure how that differs from this (other being different config for different tools).

For 2. That's going to be TSC and rollup, but I think it would be harder to resolve the right paths based on the output mode. I can look at it, but if 1. seems to be like this PR I'd like to understand what the difference is before chasing something I'm already anticipating to be difficult (at least for my limited experience with rollup).

domoritz:
For 1) I’m thinking that we wouldn’t need tsc since we go from the files in build/src.

domoritz:
Closing since there hasn't been up update in two weeks. Please send a new pull request.

PR #5928, opened 2020-02-18
fix: correctly apply axis config

kanitw:
The old code sometimes drop config when there is a default rule for undefined config value.


domoritz:
Can you write a test for it? I'm not sure what the cases are this pull request is trying to fix.

PR #5933, opened 2020-02-18
feat(experimental): support legend color signals

kanitw:
Note: signals are experimental, thus they won't show up in the schema.


domoritz:
Can I use this to dynamically change the domainMid for a diverging color scale? I think that would be useful.

src/compile/legend/assemble.ts
```diff
@@ -40,31 +41,45 @@ export function assembleLegends(model: Model): VgLegend[] {
...
+      out.stroke = {value: 'transparent'};
+    }
 
-      if (labelExpr !== undefined) {
-        let expr = labelExpr;
-        if (legend.encode?.labels?.update?.text && isSignalRef(legend.encode.labels.update.text)) {
-          expr = replaceAll(labelExpr, 'datum.label', legend.encode.labels.update.text.signal);
-        }
+    if (legend.fill) {
+      // If top-level fill is defined, for non color channel's legend, we need remove fill.
```
domoritz:
```suggestion
      // If top-level fill is defined, for non color channel's legend, we need to remove fill.
```

src/legend.ts
```diff
@@ -3,20 +3,77 @@ import {
...
+  | 'fillColor'
+  | 'gradientStrokeColor'
+  | 'labelColor'
+  | 'strokeColor'
+  | 'symbolFillColor'
+  | 'symbolStrokeColor'
+  | 'titleColor';
+
+export const SIGNAL_LEGEND_PROP_INDEX: {
+  [prop in SignalLegendProp]: {
```
domoritz: Use a `Record` type here.

PR #5939, opened 2020-02-19
feat(experimental): support signal ref for most of header properties

kanitw:
This is not particularly important, but it's super easy so I'd rather not leave it as a technical debt.

Note that since we manually map header groups, we mostly get these for free. :)

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -7892,12 +7892,16 @@
         },
         "labelFont": {
           "description": "The font of the header label.",
-          "type": "string"
+          "type": [
+            "string"
+          ]
         },
         "labelFontSize": {
           "description": "The font size of the header label, in pixels.",
           "minimum": 0,
-          "type": "number"
+          "type": [
+            "number"
+          ]
         },
         "labelFontStyle": {
           "$ref": "#/definitions/FontStyle",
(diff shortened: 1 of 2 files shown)


build/vega-lite-schema.json
```diff
@@ -7892,28 +7892,36 @@
         },
         "labelFont": {
           "description": "The font of the header label.",
-          "type": "string"
+          "type": [
```
domoritz: Hmm, the schema generator should be updated to not generate arrays with a single type. Can you file an issue?

src/header.ts
```diff
@@ -40,12 +50,12 @@ export interface CoreHeader extends FormatMixins {
   /**
    * The anchor position for placing the title. One of `"start"`, `"middle"`, or `"end"`. For example, with an orientation of top these anchor positions map to a left-, center-, or right-aligned title.
    */
-  titleAnchor?: TitleAnchor;
+  titleAnchor?: TitleAnchor; // We don't allow signal for titleAnchor since there is a dependent logic
```
domoritz: What does this mean? Can you point to the logic and what it depends on?

kanitw: It means we read the value of titleAnchor to determine other values. 

> Can you point to the logic and what it depends on?

Too much work. One can always add `| SignalRef` here and see where it breaks.

build/vega-lite-schema.json
kanitw: https://github.com/vega/ts-json-schema-generator/issues/337

src/header.ts
domoritz: ok

Issue #2760, opened 2017-08-03
Limit facet to only have layer or unit inside


kanitw:
The semantic of facet would not be what people expect. 

(I'll provide more example soon)

domoritz:
I feel having a repeat could be a useful thing to have inside a facet. Admittedly, you could do the inverse: a repeat of a facet. Anyway, I would like to see some examples of how the output is unexpected and an argument that there aren't cases where you would want arbitrary composition.

kanitw:
I think Repeat of facet is way better visualization than facet of repeat as
the latter would be extremely confusing for readers (non-creator).

arvind:
A while ago, I had created a facet of concat plots to see two different views of each facet.

kanitw:
I think longer term we certainly should support all of them. 

However, currently there are some amount of corner cases that we do not support well. 

- [ ] https://github.com/vega/vega-lite/issues/2759 -- facet of concat of facet_column (I haven't test if it works, but suspect that it probably won't.) 

- [ ] https://github.com/vega/vega-lite/issues/2775 Cross Nested Facet

So instead of delaying our release more, I'd rather support less things, but do them well.

kanitw:
In #2977, we now limit facet to only have layer or unit inside.  I'm changing this issue to become "Support generic spec inside facet" and move it out from 2.0 milestone.

domoritz:
- [ ] Support faceted concat https://github.com/vega/vega-lite/issues/5261

kanitw:
Here is a cool example of facet concat use case: https://observablehq.com/@datadesk/california-covid-19-disparities-stacked-bars.

PR #5961, opened 2020-02-23
fix: allow facet encodings inside facets

### LINKED ISSUE #5960: Nested facets invalid in Vega-Lite 4.4
I'm attempting to upgrade Altair to Vega-Lite 4.4, and am blocked by the fact that some specs that were valid under 4.0 are now invalid under the 4.4 schema.
```python
import jsonschema
import json
import urllib.request
with urllib.request.urlopen('https://vega.github.io/schema/vega-lite/v4.0.0.json') as f:
    schema4_0 = json.load(f)
with urllib.request.urlopen('https://vega.github.io/schema/vega-lite/v4.4.0.json') as f:
    schema4_4 = json.load(f)

spec = {
  "data": {"url": "data.csv"},
  "facet": {"column": {"field": "B", "type": "nominal"}},
  "spec": {
    "encoding": {"row": {"field": "A", "type": "nominal"}},
    "mark": "point"
  }
}
jsonschema.validate(spec, schema4_0)  # pass
jsonschema.validate(spec, schema4_4)  # fail
```
It appears that encoding-based facets can no longer be nested within facet specs. Is this change intended? If so, we will have to rework a number of Altair examples, tests, and documentation in order to move beyond Vega-Lite 4.0.

domoritz:
fixes #5960

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -6180,7 +6180,7 @@
           "description": "An object describing the data source. Set to `null` to ignore the parent's data source. If no data is set, it is derived from the parent."
         },
         "description": {
-          "description": "Description of this mark for commenting purpose.",
+          "description": "Description of this specification for commenting purposes.",
           "type": "string"
         },
         "encoding": {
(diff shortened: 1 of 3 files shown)


kanitw:
This was intentional since there are many cases that facet of facet are broken, and thus should not be counted as officially supported yet -- see https://github.com/vega/vega-lite/issues/2760#issuecomment-320377512.  

Facet spec are not allowed inside facet spec, either.  So I don't think facet encoding should be allowed.

domoritz:
I see. Let's close it then.

PR #5963, opened 2020-02-23
fix: support signals in view properties

### LINKED ISSUE #5962: Signals should also work for view properties, not just config
https://github.com/vega/vega-lite/pull/5888 introduced experimental support for signals for configs. However, one may want to set signals on normal properties.

```json
{
  "data": {
    "values": [
      {"a": "A", "b": 28}, {"a": "B", "b": 55}, {"a": "C", "b": 43},
      {"a": "D", "b": 91}, {"a": "E", "b": 81}, {"a": "F", "b": 53},
      {"a": "G", "b": 19}, {"a": "H", "b": 87}, {"a": "I", "b": 52}
    ]
  },
  "mark": "bar",
  "view": {
    "fill": {"signal": "'red'"}
  },
  "encoding": {
  }
}
```

But this leads to 

```
...
  "encode": {"update": {"fill": {"value": {"signal": "'red'"}}}},
```

domoritz:
Fixes #5962

### DIFF
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -7,7 +7,6 @@ import {
   SignalRef,
   Title as VgTitle
 } from 'vega-typings';
-import {hasOwnProperty} from 'vega-util';
 import {
   Channel,
   FACET_CHANNELS,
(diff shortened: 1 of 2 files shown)


src/compile/model.ts
```diff
@@ -311,12 +310,10 @@ export abstract class Model {
...
-    for (const property in baseView) {
-      if (hasOwnProperty(baseView, property)) {
-        const value = baseView[property];
-        if (value !== undefined) {
-          e[property] = {value};
-        }
+    for (const property of keys(baseView)) {
+      const value = baseView[property];
+      if (value !== undefined) {
+        e[property] = isSignalRef(value) ? value : {value};
```
kanitw: there is a helper method that we use everywhere.  valueOrSignal or the other way round.

domoritz: `signalOrValueRef`, which seems like the wrong name. I'll send a pull request.

domoritz:
Done

PR #5964, opened 2020-02-23
refactor: get rid of some hasOwnProperty

domoritz:


src/compile/selection/index.ts
```diff
@@ -69,12 +69,9 @@ export function forEachSelection(
 ) {
   const selections = model.component.selection;
   if (selections) {
-    for (const name in selections) {
-      if (hasOwnProperty(selections, name)) {
-        const sel = selections[name];
-        const success = cb(sel, compilers[sel.type]);
-        if (success === true) break;
-      }
+    for (const sel of vals(selections)) {
```
kanitw: Honestly, I don't like this since it requires generating an intermediate array to loop over. 

Why is `hasOwnProperty` a bad thing?

domoritz: The main thing here is that we actually only need the values and we can get much better types if we use vals. 

Screen Shot 2020-02-23 at 13 32 59

domoritz: Performance-wise, there isn't much difference here since the object doesn't have many keys.

PR #5984, opened 2020-02-28
Document single-value as parameter behavior for bin

marcprux:


kanitw:
Nice! Thanks for catching, let's merge after the schema gets updated again.

PR #5989, opened 2020-02-28
refactor: upgrade Vega and use typings from Vega

domoritz:
Reverts typings imports from https://github.com/vega/vega-lite/pull/5855


kanitw:
I'm curious why these CI error bars examples still change even thought we already fix the seed for generation, but that's beyond the scope of this PR.

Issue #6000, opened 2020-03-02
Cannot bind to scale of the same field


domoritz:
```json
{
  "data": {"url": "data/cars.json"},
  "mark": "point",
  "selection": {
    "grid": {"type": "interval", "resolve": "global", "bind": "scales"}
  },
  "encoding": {
    "x": {"field": "Acceleration", "type": "quantitative"},
    "y": {"field": "Acceleration", "type": "quantitative"},
    "color": {"field": "Origin", "type": "nominal"}
  }
}
```

![Kapture 2020-03-01 at 16 37 06](https://user-images.githubusercontent.com/589034/75637376-ed45e380-5bda-11ea-95c8-d0e1fd886487.gif)

domoritz:
I see this has already been thought of in https://github.com/vega/vega-lite/blob/fa9409b07ee25803a2e690333279cb37ec0c1120/src/compile/selection/transforms/project.ts#L125 by @arvind in https://github.com/vega/vega-lite/pull/4068.

kanitw:
When someone fixes this, please add

```
// it('should bind both scales in diagonal repeated views', () => {
    //   const model = parseModel({
    //     repeat: {
    //       row: ['Horsepower', 'Acceleration'],
    //       column: ['Miles_per_Gallon', 'Acceleration']
    //     },
    //     spec: {
    //       data: {url: 'data/cars.json'},
    //       mark: 'point',
    //       selection: {
    //         grid: {
    //           type: 'interval',
    //           resolve: 'global',
    //           bind: 'scales'
    //         }
    //       },
    //       encoding: {
    //         x: {field: {repeat: 'column'}, type: 'quantitative'},
    //         y: {field: {repeat: 'row'}, type: 'quantitative'},
    //         color: {field: 'Origin', type: 'nominal'}
    //       }
    //     }
    //   });

    //   model.parseScale();
    //   model.parseSelections();

    //   const scales = assembleScalesForModel(model.children[3]);
    //   expect(scales.length === 2).toBe(true);
    //   expect(scales[0]).toHaveProperty('domainRaw');
    //   expect(scales[1]).toHaveProperty('domainRaw');
    //   expect(scales[0].domainRaw).toEqual({signal: 'grid["Acceleration"]'});
    //   expect(scales[1].domainRaw).toEqual({signal: 'grid["Acceleration"]'});
    // });
```

to `scales.test.ts`.

Right now I'll remove the commented code so we don't get lint warning.

[...]

PR #6002, opened 2020-03-02
refactor: implement repeat as a normalization to reduce number of models

### LINKED ISSUE #4947: Refactor Repeat as a Macro?
Our repeat should have the same behavior as using fold (unpivot) and then facet into row/column. We should investigate whether it makes sense to clean up the implementation and implement repeat as a macro.

domoritz:
Fixes #4947

This also fixes the layer type, which incorrectly extended generic layer spec.


domoritz:
Rebased to master. Looking pretty good now.

examples/compiled/geo_repeat.vg.json
```diff
@@ -85,128 +83,152 @@
   ],
   "projections": [
     {
-      "name": "projection",
-      "size": {"signal": "[width, height]"},
-      "fit": {
-        "signal": "[child__repeat_row_population_geojson_0, child__repeat_row_engineers_geojson_0, child__repeat_row_hurricanes_geojson_0]"
-      },
+      "name": "__row_population_projection",
```
kanitw: projection doesn't get merged correctly.

examples/compiled/geo_repeat.vg.json
```diff
@@ -30,7 +28,7 @@
         {
           "type": "geojson",
           "geojson": "geo",
-          "signal": "child__repeat_row_population_geojson_0"
+          "signal": "__row_population_geojson_0"
```
kanitw: To minimize review churns, I'd rather have consistent naming pattern with the past. You can send another PR to change / simplify it later, but avoid it in this giant PR.

examples/compiled/geo_repeat.vg.json
```diff
@@ -85,128 +84,158 @@
   ],
   "projections": [
     {
-      "name": "projection",
-      "size": {"signal": "[width, height]"},
-      "fit": {
-        "signal": "[child__repeat_row_population_geojson_0, child__repeat_row_engineers_geojson_0, child__repeat_row_hurricanes_geojson_0]"
-      },
+      "name": "__repeat_row_population_projection",
```
domoritz: Projections are not merged because layout sizes are not merged correctly here.

kanitw: Then fix the layout size to merge correctly? 

I don't think we should merge this without fixing it.

examples/compiled/geo_repeat.vg.json
```diff
@@ -30,7 +29,7 @@
         {
           "type": "geojson",
           "geojson": "geo",
-          "signal": "child__repeat_row_population_geojson_0"
+          "signal": "__repeat_row_population_geojson_0"
```
kanitw: The name is still changing.

examples/compiled/geo_repeat.vg.json
domoritz: Filed as https://github.com/vega/vega-lite/issues/6024 since it's orthogonal

examples/compiled/geo_repeat.vg.json
```diff
@@ -85,22 +84,40 @@
   ],
   "projections": [
     {
-      "name": "projection",
-      "size": {"signal": "[width, height]"},
-      "fit": {
-        "signal": "[child__repeat_row_population_geojson_0, child__repeat_row_engineers_geojson_0, child__repeat_row_hurricanes_geojson_0]"
-      },
+      "name": "child__repeat_row_population_projection",
```
domoritz: These changes come from an orthogonal issue: https://github.com/vega/vega-lite/issues/6024

examples/compiled/geo_repeat.vg.json
domoritz: Fixed

kanitw:
Please rebase after https://github.com/vega/vega-lite/pull/6026 is merged.

domoritz:
TODO:
* [ ] Make sure we still use the repeat config
* [ ] Clean up `extractCompositionLayout`

domoritz:
Merge after https://github.com/vega/vega-lite/pull/6029

domoritz:
Okay, I decided to just drop support for repeat configs. I don't think anyone really used it anyway. We could hack it through style maybe but I think this is an acceptable exception.

examples/compiled/interactive_panzoom_splom.vg.json
```diff
@@ -118,22 +118,25 @@
   "marks": [
     {
       "type": "group",
-      "name": "child__repeat_row_Horsepower__repeat_column_Miles_per_Gallon_group",
+      "name": "child__repeat_row_Horsepowerchild__repeat_column_Miles_per_Gallon_group",
```
kanitw: The name is a bit weird here

kanitw:
I think you just caused even more renames lol

domoritz:
I think it's only these ones, right? 

Screen Shot 2020-03-08 at 18 28 00

kanitw:
> I think it's only these ones, right?

IDK, because I haven't looked at them -- too many.

domoritz:
I looked through the generated SVGs and this looks good to me. Anything left @kanitw?

examples/compiled/interactive_layered_crossfilter_discrete.vg.json
```diff
@@ -183,6 +182,7 @@
   ],
   "signals": [
     {"name": "childWidth", "value": 200},
+    {"name": "childHeight", "value": 200},
```
kanitw: Why does this become childHeight?

kanitw:
> I looked through the generated SVGs and this looks good to me. Anything left @kanitw?

It'd be great to look at generated Vega too. 

Just added one comment after looking at a few Vega specs.

test/compile/repeat.test.ts
```diff
@@ -1,10 +1,8 @@
-import {replaceRepeaterInEncoding} from '../../src/compile/repeater';
+import {replaceRepeaterInEncoding, replaceRepeaterInFacet} from '../../src/compile/repeater';
```
kanitw: Why does this file remains here at all given there is no model left?

src/spec/map.ts
```diff
@@ -76,18 +88,19 @@ export abstract class SpecMapper<
...
       spec: this.map(spec.spec, params) as any
     };
   }
 
-  protected mapRepeat(spec: GenericRepeatSpec<UI, LI>, params: P): GenericRepeatSpec<UO, GenericLayerSpec<UO>> {
+  protected mapRepeat(spec: RepeatSpec, params: P): GenericSpec<UO, any, RO, FO> {
     return {
       ...spec,
-      spec: this.map(spec.spec, params)
+      spec: this.map(spec.spec as any, params)
```
kanitw: comment why `as any` here (even if it's the same as in `mapFacet` -- they are two different functions).

src/log/message.ts
```diff
@@ -85,13 +85,9 @@ export function columnsNotSupportByRowCol(type: 'facet' | 'repeat') {
...
 
-// CONCAT
+// CONCAT / REPEAT
 export const CONCAT_CANNOT_SHARE_AXIS =
-  'Axes cannot be shared in concatenated views yet (https://github.com/vega/vega-lite/issues/2415).';
-
-// REPEAT
-export const REPEAT_CANNOT_SHARE_AXIS =
-  'Axes cannot be shared in repeated views yet (https://github.com/vega/vega-lite/issues/2415).';
+  'Axes cannot be shared in concatenated or concatenated views yet (https://github.com/vega/vega-lite/issues/2415).';
```
kanitw: concatenated or concatenated ? lol

src/compile/unit.ts
```diff
@@ -72,31 +71,21 @@ export class UnitModel extends ModelWithField {
...
 
-    const encodingWithRepeaterReplaced = replaceRepeaterInEncoding(spec.encoding ?? {}, repeater);
+    const encoding = spec.encoding ?? {};
 
-    this.markDef = initMarkdef(spec.mark, encodingWithRepeaterReplaced, config, {
+    this.markDef = initMarkdef(spec.mark, encoding, config, {
       graticule: spec.data && isGraticuleGenerator(spec.data)
     });
-    const encoding = (this.encoding = initEncoding(encodingWithRepeaterReplaced, this.markDef));
+    const normalizedEncoding = (this.encoding = initEncoding(encoding, this.markDef));
```
kanitw: Let's not call this normalized as it doesn't happen in the normalized phase. Instead, just call this encoding and refer to the old `encoding` above via `spec.encoding`.

src/compile/selection/transforms/scales.ts
```diff
@@ -34,12 +34,6 @@ const scaleBindings: TransformCompiler = {
       const extent = {selection: selCmpt.name, field: proj.field};
       scale.set('selectionExtent', extent, true);
       bound.push(proj);
-
-      // Bind both x/y for diag plot of repeated views.
-      if (model.repeater && model.repeater.row === model.repeater.column) {
```
kanitw: this seems problematic?

scripts/rename-schema.sh
```diff
@@ -7,8 +7,8 @@ perl -pi -e s,'<Field\,','<',g build/vega-lite-schema.json
 perl -pi -e s,'<StandardType>','',g build/vega-lite-schema.json
 
 perl -pi -e s,'CompositeEncoding','Encoding',g build/vega-lite-schema.json
-perl -pi -e s,'Generic(.*)<FacetedUnitSpec\,LayerSpec>','\1',g build/vega-lite-schema.json
-
+perl -pi -e s,'Generic(.*)<FacetedUnitSpec\,LayerSpec\,?.*\,FieldName>','\1',g build/vega-lite-schema.json
+perl -pi -e s,'Generic(.*)<FacetedUnitSpec\,LayerSpec\,?.*\,Field>','Normalized\1',g build/vega-lite-schema.json
```
kanitw: Can you explain why do we need this rename? Looks very complicated.

src/normalize/core.ts
```diff
@@ -55,43 +58,80 @@ export class CoreNormalizer extends SpecMapper<NormalizerParams, FacetedUnitSpec
...
+          const childRepeater = {
+            repeat: repeatValue,
+            row: rowValue,
+            column: columnValue
+          };
+
+          const child = this.map(childSpec, {...params, repeater: childRepeater});
+
+          // TODO: remove child__repeat_ prefix
+          child.name =
```
kanitw: This line is wrong, you should only include `child__repeat` only once in the beginning.  (Hence the name changes in the diff.)

examples/compiled/repeat_splom_cars.vg.json
```diff
@@ -46,22 +45,25 @@
...
+  "signals": [
+    {"name": "childWidth", "value": 200},
+    {"name": "childHeight", "value": 200}
+  ],
   "layout": {"padding": 20, "columns": 2, "bounds": "full", "align": "all"},
   "marks": [
     {
       "type": "group",
-      "name": "child__repeat_row_Displacement__repeat_column_Horsepower_group",
+      "name": "child__repeat_row_Displacementchild__repeat_column_Horsepower_group",
```
kanitw: see https://github.com/vega/vega-lite/pull/6002#discussion_r390498678

examples/compiled/interactive_layered_crossfilter_discrete.vg.json
domoritz: Do you see an issue in the normalized (concat) spec? The height here is just the default, no?

test/compile/repeat.test.ts
domoritz: We still need to replace repeater in encoding and facet in normalization. Should I move the test?

src/log/message.ts
domoritz: Ups. Will fix.

src/compile/selection/transforms/scales.ts
domoritz: This is https://github.com/vega/vega-lite/issues/6000, right?

scripts/rename-schema.sh
domoritz: This is the similar to before. Just renaming generic specs with FieldName (string) to be normalized spec and generic spec with Field (string or repeat) to just be a spec.

test/compile/repeat.test.ts
kanitw: Yes, to where the function is placed now.

src/compile/selection/transforms/scales.ts
kanitw: Perhaps, but what's the effect of removing that?

test/compile/repeat.test.ts
domoritz: I'm moving the files and tests then

src/compile/unit.ts
domoritz: Done

examples/compiled/interactive_panzoom_splom.vg.json
```diff
@@ -1895,7 +1895,6 @@
       "name": "child__repeat_row_Horsepower__repeat_column_Horsepower_y",
       "type": "linear",
       "domain": {"data": "data_2", "field": "Horsepower"},
-      "domainRaw": {"signal": "grid[\"Horsepower\"]"},
```
kanitw: Is this change intentional? (and why?)

src/compile/selection/transforms/scales.ts
domoritz: Let's try: https://github.com/vega/vega-lite/pull/6124

examples/compiled/interactive_panzoom_splom.vg.json
domoritz: Answer is above.

src/normalize/core.ts
```diff
@@ -55,43 +58,81 @@ export class CoreNormalizer extends SpecMapper<NormalizerParams, FacetedUnitSpec
...
+        }
+      }
+    }
+
+    const columns = isArray(repeat) ? spec.columns : repeat.column ? repeat.column.length : 1;
+
     return {
-      ...spec,
-      spec: this.map(spec.spec, params)
+      data: data ?? childSpec.data, // data from child spec should have precedence
```
kanitw: The code here doesn't really give precedence to child spec?

src/normalize/core.ts
```diff
@@ -55,43 +58,81 @@ export class CoreNormalizer extends SpecMapper<NormalizerParams, FacetedUnitSpec
...
+          };
+
+          const child = this.map(childSpec, {...params, repeater: childRepeater});
+
+          // TODO: remove child__repeat_ prefix
+          child.name =
+            'child' +
+            (repeatValue ? `__repeat_repeat_${varName(repeatValue)}` : '') +
+            (rowValue ? `__repeat_row_${varName(rowValue)}` : '') +
+            (columnValue ? `__repeat_column_${varName(columnValue)}` : '');
```
kanitw: The naming logic here seems wrong for nested repeat? 

Existing repeatValue should be treated as prefix to the name, not inserted in the middle.

PR #6015, opened 2020-03-03
feat: add `scale.reverse` and `config.scale.xReverse`

kanitw:
I'm not adding yReverse since it's unclear when that'd be useful.  We can add that later. 

Note that this is stylistically different from axis config where we have `config.axisX`, `config.axisY`. 

However, people may want to customize different axis styles, more so than scales, which people mostly will just change range values and xReverse (for RTL).


build/vega-lite-schema.json
```diff
@@ -13594,6 +13594,10 @@
         "useUnaggregatedDomain": {
           "description": "Use the source data range before aggregation as scale domain instead of aggregated data for aggregate axis.\n\nThis is equivalent to setting `domain` to `\"unaggregate\"` for aggregated _quantitative_ fields by default.\n\nThis property only works with aggregate functions that produce values within the raw data domain (`\"mean\"`, `\"average\"`, `\"median\"`, `\"q1\"`, `\"q3\"`, `\"min\"`, `\"max\"`). For other aggregations that produce values outside of the raw data domain (e.g. `\"count\"`, `\"sum\"`), this property is ignored.\n\n__Default value:__ `false`",
           "type": "boolean"
+        },
+        "xReverse": {
```
domoritz: Why not `reverseX`? Sounds more natural to me.

kanitw: we use condition for specificity as prefix elsewhere (e.g., `barBandPadding`)

PR #6025, opened 2020-03-06
refactor: implement repeat as a normalization to reduce number of models 

kanitw:
Extending/replacing #6002, let's see if this makes the spec diff less scary.


kanitw:
This seems good, but causing extra spec changes. Let's close this PR and do https://github.com/vega/vega-lite/pull/6026 separately instead.

Issue #6024, opened 2020-03-06
Layout size signals are not merged correctly


domoritz:
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
  "data": {"url": "data/weather.csv"},
  "repeat": {"column": ["temp_max", "precipitation"]},
  "spec": {
    "mark": "line",
    "encoding": {
      "y": {
        "aggregate": "mean",
        "field": {"repeat": "column"},
        "type": "quantitative"
      },
      "x": {"timeUnit": "month", "field": "date", "type": "ordinal"}
    }
  }
}
```

Produces different signals for width because the scales are different even though they should be the same.

kanitw:
Fixed in https://github.com/vega/vega-lite/pull/6026

PR #6026, opened 2020-03-06
fix: make layoutsize always merge width and height

kanitw:
fix https://github.com/vega/vega-lite/issues/6024

Replacing https://github.com/vega/vega-lite/pull/6025 since this is causing other spec changes beyond the original #6002.


examples/compiled/concat_bar_scales_discretize_2_cols.vg.json
```diff
@@ -77,11 +75,11 @@
       "style": "cell",
       "encode": {
         "update": {
-          "width": {"signal": "concat_0_width"},
+          "width": {"signal": "width"},
```
kanitw: This seems wrong.

kanitw:
Will be replaced with https://github.com/vega/vega-lite/pull/6028

Issue #6023, opened 2020-03-05
Remove concat type from concat model and use only column instead


domoritz:
vconcat and hconcat are just concat with different numbers of rows

kanitw:
I don't see why we need to move this to normalizer yet?  
Moving everything to normalize also means we will generate many more normalized specs, bloating our PRs. 

Since there is already only one ConcatModel, I don't see why this is gonna be useful.

domoritz:
We have some code that distinguishes in assemble and it is probably not as clean as it could be. But maybe some refactoring is enough. I’ll take a look at some point to see what makes sense.

kanitw:
> We have some code that distinguishes in assemble and it is probably not as clean as it could be.

This sounds pretty vague, can you point to a specific problematic line?

domoritz:
I will remove the concat type from the model and instead only use column. That should simplify things enough.

PR #6027, opened 2020-03-06
refactor: remove concat type from concat model

domoritz:
Fixes https://github.com/vega/vega-lite/issues/6023

We now consistently treat a concat spec with 1 column as a vconcat.

### DIFF
--- a/src/compile/layoutsize/parse.ts
+++ b/src/compile/layoutsize/parse.ts
@@ -19,14 +19,24 @@ export function parseLayerLayoutSize(model: Model) {
 
 export const parseRepeatLayoutSize = parseLayerLayoutSize;
 
+function sizeTypeToMerge(columns: number | undefined) {
+  switch (columns) {
+    case 1:
+      return 'width';
+    case undefined:
+      return 'height';
+    default:
+      return undefined;
+  }
+}
+
 export function parseConcatLayoutSize(model: ConcatModel) {
   parseChildrenLayoutSize(model);
   const layoutSizeCmpt = model.component.layoutSize;
 
-  const columns = model.layout.columns;
-  const sizeTypeToMerge = columns === 1 ? 'width' : columns === undefined ? 'height' : undefined;
-  if (sizeTypeToMerge) {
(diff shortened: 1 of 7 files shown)


src/compile/model.ts
```diff
@@ -209,7 +209,7 @@ export abstract class Model {
 
     this.description = spec.description;
     this.transforms = normalizeTransform(spec.transform ?? []);
-    this.layout = isUnitSpec(spec) || isLayerSpec(spec) ? {} : extractCompositionLayout(spec, type, config);
+    this.layout = type === 'layer' || type === 'unit' ? {} : extractCompositionLayout(spec, type, config);
```
kanitw: why is this related?

src/compile/layoutsize/parse.ts
```diff
@@ -19,16 +19,12 @@ export function parseLayerLayoutSize(model: Model) {
...
-  hconcat: 'height'
-};
-
 export function parseConcatLayoutSize(model: ConcatModel) {
   parseChildrenLayoutSize(model);
   const layoutSizeCmpt = model.component.layoutSize;
 
-  const sizeTypeToMerge = SIZE_TYPE_TO_MERGE[model.concatType];
+  const columns = model.layout.columns;
+  const sizeTypeToMerge = columns === 1 ? 'width' : columns === undefined ? 'height' : undefined;
```
kanitw: change `concatType` to a getter (which determines type based on column)? 

So here we can make the code a bit more readable?

src/compile/model.ts
domoritz: extractCompositionLayout now has a type for column while it was any before.

src/compile/layoutsize/parse.ts
domoritz: Would that actually be more readable? I feel this code is very clear about what it does and you don't need to look at another function. I can make this a case statement, though.

kanitw: still irrelevant to this PR, but sure I can ignore it.

src/compile/model.ts
kanitw: still irrelevant to this PR, but sure I can ignore it.

domoritz: Okay

PR #6029, opened 2020-03-07
fix: update schema generator, fix layer spec type

domoritz:


src/spec/layer.ts
```diff
@@ -18,7 +18,14 @@ export interface GenericLayerSpec<U extends GenericUnitSpec<any, any>> extends B
 /**
  * A full layered plot specification, which may contains `encoding` and `projection` properties that will be applied to underlying unit (single-view) specifications.
  */
-export interface LayerSpec extends GenericLayerSpec<UnitSpec> {
+export interface LayerSpec extends BaseSpec, FrameMixins, ResolveMixins {
```
kanitw: Please add comment explaining why LayerSpec, unlike other spec, does not extend its own generic spec.

domoritz:
Notice how I deleted the one line in the rename script. The hack was needed because the layer spec type was incorrect.

PR #6031, opened 2020-03-07
refactor: make repeat use the same layout parser as concat

kanitw:
This should extract irrelevant changes from https://github.com/vega/vega-lite/pull/6002/files


domoritz:
Ups, I didn't see the test failures. Fixing in https://github.com/vega/vega-lite/pull/6035

Issue #4727, opened 2019-03-14
vl2vg STDIN error in Windows


crouthamela:
**OS:**
Windows 7

**Versions:**
vega-cli - 5.2.0 (from npm)
vega-lite - 3.0.0-rc15 (from npm)

I have a simple Vega-Lite diagram:
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v3.json",
  "width": 500,
  "data": {
    "values": [
      {"Customer": "Customer1", "Bandwidth": 250},
      {"Customer": "Customer2", "Bandwidth": 50}
    ]
  },
  "mark": "bar",
  "encoding": {
    "color": {"field": "Customer", "type": "nominal"},
    "x": {
      "field": "Bandwidth",
      "scale": {"domain": [0, 500]},
      "type": "quantitative",
      "axis": {"title": "Circuit Bandwidth Guarantee (Mbps)"}
    }
  }
}
```

When I run `vl2vg < test.json`, I receive the following error:
```
H:\Documents>vl2vg < test.json
C:\Users\ancro\AppData\Roaming\npm\node_modules\vega-lite\bin\vl2vg:35
    throw err;
    ^

Error: ENOENT: no such file or directory, open 'H:\dev\stdin'
```

This appears similar to [this vega issue](https://github.com/vega/vega/issues/1276).

domoritz:
We should use `process.stdin`. Can you send a PR?

crouthamela:
Sorry, I'm not really a coder, just a network engineer. This ends up causing asciidoctor-diagram to fail on Windows though, so I certainly hope someone can make the change.

jheer:
In addition to having vl2vg use `process.stdin`, it looks like each utility function (e.g., `vl2pdf`) needs to be rewritten to work cross-platform. I'd recommend dropping the bash scripts and using node instead. At least, this is what Vega does.

SoftwareApe:
Could we use https://github.com/mbostock/rw here?

That way you only need to do:

    var rw = require('rw');

and later 

    rw.readFile(...

The rest can stay the same. With `rw`'s readFile `/dev/stdin` works on Windows.

domoritz:
Yes, we could but we usually prefer not to add dependencies if we can.

stenzengel:
I have the same problem. I am also using asciidoctor-diagram. Vega works correctly, only vega-lite has problems. 

The milestone was removed recently. Will this issue be solved?

Does anybody know a workaround?

[...]

PR #6034, opened 2020-03-08
feat: add vl2pdf, test the cli, support windows

domoritz:
Fixes https://github.com/vega/vega-lite/issues/4727

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -106,8 +106,9 @@ jobs:
           bin/vl2pdf examples/specs/bar.vl.json > bar.pdf
 
       - name: Compile from stdin
+        shell: cmd
         run: |
-          cat examples/specs/bar.vl.json | bin/vl2vg -p > bar-stdin.vg.json
-          cat examples/specs/bar.vl.json | bin/vl2png > bar-stdin.png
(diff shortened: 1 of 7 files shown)


domoritz:
@stenzengel do you know why windows is failing? How do I test stdin?

stenzengel:
For my tests I used `vl2vg < test.json` as in the original description of #4727. (After this worked without error, also my original use case with `asciidoctor-diagram` worked). Another possibility should be using `type` instead of `cat`.

domoritz:
That still doesn't seem to work @stenzengel.

Screen Shot 2020-03-08 at 18 13 03

domoritz:
Nothing seems to work. Note that github actions use powershell. Can you help me get this test working @stenzengel?

stenzengel:
@domoritz Yes, the problem is that GitHub Actions use Powershell. I had the same problem (`The '<' operator is reserved for future use`) after switching to Powershell. I have [found](https://stackoverflow.com/questions/2148746/the-operator-is-reserved-for-future-use) the following workaround, which worked for me:
```
Get-Content test.vl.json | vl2vg.cmd > test.vg.json
```
Please note, that I also had to add the ".cmd" extension to "vl2vg".

Hope that helps.

PR #6047, opened 2020-03-09
feat: make `background` and `padding` support (experimental) signal ref

kanitw:


src/spec/base.ts
```diff
@@ -81,7 +81,7 @@ export interface LayoutSizeMixins {
    *
    * __See also:__ [`width`](https://vega.github.io/vega-lite/docs/size.html) documentation.
    */
-  width?: number | 'container' | Step;
+  width?: number | 'container' | Step; // Vega also supports SignalRef for width and height. However, we need to know if width is a step or not in VL and it's very difficult to check this at runtime, so we intentionally do not support SignalRef here.
```
domoritz: What's so difficult to check at runtime?

src/spec/toplevel.ts
```diff
@@ -44,23 +45,23 @@ export interface TopLevelProperties {
...
+  padding?: Padding | SignalRef;
 
   /**
    * How the visualization size should be determined. If a string, should be one of `"pad"`, `"fit"` or `"none"`.
    * Object values can additionally specify parameters for content sizing and automatic resizing.
    *
    * __Default value__: `pad`
    */
-  autosize?: AutosizeType | AutoSizeParams;
+  autosize?: AutosizeType | AutoSizeParams; // Vega actually supports signal for autosize. However, we need to check autosize at compile time to infer the rest of the spec. Thus VL's autosize wont support SignalRef for now.
```
domoritz:
```suggestion
  autosize?: AutosizeType | AutoSizeParams; // Vega actually supports signal for autosize. However, we need to check autosize at compile time to infer the rest of the spec. Thus VL's autosize won't support SignalRef for now.
```

src/spec/base.ts
kanitw: Good luck making all the dependent logic living in signal world.

PR #6055, opened 2020-03-10
feat: support custom `formatType` + refactor format code

kanitw:
These are multiple atomic commits, please do not squash. 

(It's  also likely easier to review a commit at a time).


src/channeldef.ts
```diff
@@ -1037,6 +1040,7 @@ export function channelCompatibility(
   }
 }
 
+// TODO: rename to isFieldDefForTimeFormat
```
domoritz: Why not do it?

src/compile/axis/encode.ts
```diff
@@ -1,35 +1,32 @@
...
 
-  let labelsSpec: any = {};
+  const text = formatSignalRef({
+    fieldDef,
+    field: 'datum.value',
+    format,
+    formatType,
+    config,
+    isUTCScale: model.getScaleComponent(channel).get('type') === ScaleType.UTC,
+    omitTimeFormatConfig: true, // Vega axes automatically determine format each label automatically so we don't apply config.timeFormat
```
domoritz:
```suggestion
    omitTimeFormatConfig: true, // Vega axes automatically determine the format for each label so we don't apply config.timeFormat
```

src/channeldef.ts
kanitw: Done

Issue #6064, opened 2020-03-11
Sorting by nominal field fails in some circumstances


jakevdp:
This specification does not show the expected sort order in the y axis ([vega editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAutIC4yghSA3WAbArgKYDOKYA2uBhMNnmQIwA0UusARobmZAIKQBfJlWq0c3VACYWkNpwlQAQoOHUadBQGYZcrj34CRYALpUhVSAFtYAJwDWPdrciqohAHYBjAPbQAlu4A5mToGJAAHiGQAGZ+XNA84i5Q8ACeAA6EPACO+LDu8H6IhVhZ5tSQqVGx8Yn0MmmZPN42-u71UMQt8NVxuAmoshx6AuVhPrgtvbWDutwNGVmD7t6WAfSGEIYCQA)):
```json
{
  "data": {
    "values": [
      {"val": 1, "label": "A"},
      {"val": 2, "label": "B"},
      {"val": 3, "label": "A"}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "val", "type": "quantitative"},
    "y": {"field": "val", "type": "ordinal", "sort": {"field": "label"}},
    "color": {"field": "label", "type": "nominal"}
  }
}
```
![visualization - 2020-03-10T173712 567](https://user-images.githubusercontent.com/781659/76370954-cb90de80-62f5-11ea-98e7-4315df2dc7d5.png)

This bug only shows up with a specific combination of features; for example, the sort works correctly when doing any of the following:

- Removing the ``x`` encoding ([vega editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAutIC4yghSA3WAbArgKYDOKYA2uBhMNnmQIwA0UusARobmZAIKQBfJlWq0c3VACYWkNpwlQAQoOHUadBQGYZcrj34CRYALpUhVSAFtYAJwDWPdrciqohAHYBjAPbQAlu4A5mToGJAAniGQAGZ+XNA84i5Q8OEADoQ83jb+7vQyxNnwUbHxPLrcAubUkD642SVxuAmoshx6MqkZPO7elgH0hhCGAkA)):
   ![visualization - 2020-03-10T195507 073](https://user-images.githubusercontent.com/781659/76377794-10724080-6309-11ea-94c7-02f4337d970a.png)

[...]

PR #6065, opened 2020-03-11
fix: always use sort min even if chart is stacked

domoritz:
Fixes https://github.com/vega/vega-lite/issues/6064

### DIFF
--- a/src/compile/scale/domain.ts
+++ b/src/compile/scale/domain.ts
@@ -362,14 +362,11 @@ function parseSingleChannelDomain(
   }
 }
 
-function normalizeSortField(sort: EncodingSortField<string>, isStacked: boolean) {
+function normalizeSortField(sort: EncodingSortField<string>) {
   const {op, field, order} = sort;
   return {
-    // Apply default op
-    op: op ?? (isStacked ? 'sum' : DEFAULT_SORT_OP),
-    // flatten nested fields
+    op: op ?? DEFAULT_SORT_OP,
     ...(field ? {field: util.replacePathInField(field)} : {}),
-
     ...(order ? {order} : {})
   };
 }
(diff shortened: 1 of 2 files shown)


kanitw:
What if the sort field is quantitative? Sum is probably better?

domoritz:
I’m not sure it is better, actually. We use min everywhere else so using sum all of a sudden could come as a surprise. I’d vote for consistency over smartness here.

kanitw:
I think it should be min if the sort field is a groupby, but sum if the sort field is not a groupby?

domoritz:
That would also work but I am not yet convinced that `min` isn't easier and is right for almost all cases. Note that if you sort by the stacked field, we don't have to do any aggregation.

kanitw:
> Note that if you sort by the stacked field, we don't have to do any aggregation.

When you sort by a stacked quantitative field, you have to sum them.

domoritz:
If we sort by the stacked field, we don't need any aggregation since we group by the color already, no?

kanitw:
> If we sort by the stacked field, we don't need any aggregation since we group by the color already, no?

The sorted value should be summed across color groups.  That's not the result of the original aggregation.

domoritz:
Oh right, then yes, we want to sum up the data. Let's revise this pull request.

domoritz:
Okay, so the issue is more nuanced

Should use min

```
{
  "data": {
    "values": [
      {"val": 1, "val2": 1.1, "label": "A"},
      {"val": 2, "val2": 1.8, "label": "B"},
      {"val": 3, "val2": 3.2, "label": "A"}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "val2", "type": "quantitative"},
    "y": {"field": "val", "type": "ordinal", "sort": {"field": "label"}},
    "color": {"field": "label", "type": "nominal"}
  }
}
```


Should use sum

```
{
  "data": {
    "values": [
      {"val": 1, "val2": 1.1, "label": "A"},
      {"val": 2, "val2": 1.8, "label": "B"},
      {"val": 3, "val2": 3.2, "label": "A"}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "val2", "type": "quantitative"},
    "y": {"field": "val", "type": "ordinal", "sort": {"field": "val2"}},
    "color": {"field": "label", "type": "nominal"}
  }
}
```

```
{
  "data": {
    "values": [
      {"val": 1, "val2": 1.1, "label": "A"},
      {"val": 2, "val2": 1.8, "label": "B"},
      {"val": 3, "val2": 3.2, "label": "A"}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "val2", "type": "quantitative"},
    "y": {"field": "val", "type": "ordinal", "sort": "x"},
    "color": {"field": "label", "type": "nominal"}
  }
}
```

domoritz:
I'll send a new pull request.

PR #6071, opened 2020-03-12
fix: don't sum up stacked dimension fields

domoritz:
Fixes https://github.com/vega/vega-lite/issues/6064

### DIFF
--- a/src/stack.ts
+++ b/src/stack.ts
@@ -34,7 +34,7 @@ export interface StackProperties {
   groupbyChannel: 'x' | 'y';
 
   /** Field for groupbyChannel. */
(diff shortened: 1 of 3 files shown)


src/stack.ts
```diff
@@ -32,6 +33,9 @@ export interface StackProperties {
   /** Dimension axis of the stack. */
   groupbyChannel: 'x' | 'y';
 
+  /** Field for groupbyChannel. */
+  groupbyField: FieldName;
```
kanitw: `groupbyField?` ? (It can be undefined)

domoritz: Fixed in follow up commit

Issue #6066, opened 2020-03-11
Transform groupby does not escape compound fields


jakevdp:
Consider this specification ([vega editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAutIC4yghSA3WAbArgKYDOKYA2uBhMJAOYBOA9vgA5kCMANFK4Q8SYA7MpACCuWCIC+XKtVqMW7VN179BI1JADihJgzqFIs+RkXM2nHpD4DhogGINC0E3Oo16llWABMNnaaogDK8IRYxqaeaN7KZAHq9lpQAHIAlgDGANbuZl5KVqiJthoO2gDq6bi4eTEW8cWBZSmQAMJMuFEeng1FYADMzcmiAEpZudG9cf1DScHaACpMALYmZgC6VFOQ8AxSxABmBmuolAr5ULhMTNn9pSM91JCHzKdol5hwiGToMRjYPBEUhnT7TISwFbGbQSKSQGwAfUE+AYmWhsVgRk4AAYbJkEIQ6AYAJ5kACs0im-3MkAhUNEyzWiORqPRtEx6IA7HiCUSGKTVNjKWCIFtqVTPJBsoQBVA6cYnpLDulCLhoCCKJAkSxWQA6DnwqDalFo3X48J80lizzSMwSv7PTGMQkEsjkWhMXyQYj4JlQZWqtzaY16g02WAa3ZMRCSLHSDaKzCFVgAI1l5C1LNN5sJJMg1ogttFPUgK1gDFy2hT5fhVEghCEmSY0HSQjovzMkAAHr8Xiq1aIQ9neXmbPBiXxRAYWxDahLILLaAGB9p4NG8GGoOPJ9oAI74KTwdKII+RdaFkDSIA)):
```
{
  "data": {
    "values": [
      {"group": 1, "person": "Alan"},
      {"group": 1, "person": "George"},
      {"group": 1, "person": "Fred"},
      {"group": 2, "person": "Steve"},
      {"group": 2, "person": "Nick"},
      {"group": 2, "person": "Will"},
      {"group": 2, "person": "Cole"},
      {"group": 3, "person": "Rick"},
      {"group": 3, "person": "Tom"}
    ]
  },
  "transform": [
    {
      "lookup": "person",
      "from": {
        "data": {
          "values": [
            {"name": "Alan", "_source": {"age": 10, "category": 5}},
            {"name": "Tom", "_source": {"age": 7, "category": 10}}
[...]

PR #6072, opened 2020-03-12
fix: escape paths in dimension fields

domoritz:
Fixes https://github.com/vega/vega-lite/issues/6066


test/compile/data/aggregate.test.ts
```diff
@@ -289,4 +289,17 @@ describe('compile/data/aggregate', () => {
       expect(agg1.producedFields()).toEqual(new Set(['a_mean', 'b_mean']));
     });
   });
+
+  describe('assemble', () => {
```
domoritz:
```suggestion
  describe('assemble()', () => {
```

PR #6074, opened 2020-03-12
docs: Adding a new tool for embedding vega-lite in flutter-web.

Abhilash-Chandran:
Added a new entry referring the vega_embed_flutter package which is flutter widget to embed vega-lite specs in flutter web pages.


domoritz:
Thank you. Can you shorten the text a bit and use active voice? Then we can merge.

PR #6076, opened 2020-03-12
fix: `labelOffset` type

kanitw:


src/axis.ts
```diff
@@ -206,34 +206,34 @@ export type AxisConfigBaseWithConditionalAndSignal = Omit<BaseAxisNoValueRefs, C
 
 export interface AxisPropsWithConditionAndSignal {
   domainColor?: BaseAxisNoValueRefs['domainColor'] | SignalRef;
-  labelAlign?: BaseAxisNoValueRefs['labelAlign'] | ConditionalAxisLabelAlign;
-  labelBaseline?: BaseAxisNoValueRefs['labelBaseline'] | ConditionalAxisLabelBaseline;
+  labelAlign?: BaseAxisNoValueRefs['labelAlign'] | ConditionalAxisLabelAlign | SignalRef;
```
domoritz: Would it be better to rewrite `BaseAxisNoValueRefs` to not remove SignalRefs? I feel it's dangerous to add SignalRefs back when they may or may not be allowed in Vega.

domoritz: Actually, take a look at ExcludeMappedValueRefButKeepSignal. Why is this change even necessary?

kanitw: Because Type script is stupid ... if you type 

type X = AxisPropsWithConditionAndSignal[xxx]


_Some_ (not all) properties don't get signals. 


(We should file a bug)

domoritz: I don't get what the bug is.

kanitw: Ok maybe it's not TS bug, but the problem is "butKeepSignal" doesn't really always keep signals.

domoritz: Let's fix that then rather than adding signals back by hand.

PR #6080, opened 2020-03-12
fix: correct style merging logic

kanitw:


src/compile/common.ts
```diff
@@ -81,7 +81,7 @@ export function getStyleConfig<P extends keyof AnyMarkConfig | keyof Axis>(
   for (const style of styles) {
     const styleConfig = styleConfigIndex[style];
 
-    if (styleConfig) {
+    if (styleConfig && styleConfig[p as string] !== undefined) {
```
domoritz: Why not `p in styleConfig`?

PR #6083, opened 2020-03-12
feat: support axis config based on both orientation and type (e.g., `config.axisXTemporal`)	

kanitw:
Merge https://github.com/vega/vega-lite/pull/6082 first

### DIFF
--- a/test/compile/axis/parse.test.ts
+++ b/test/compile/axis/parse.test.ts
@@ -191,7 +191,8 @@ describe('Axis', () => {
 
       const axisComponent = parseUnitAxes(model);
       expect(axisComponent['x'].length).toEqual(1);
-      expect(axisComponent['x'][0].implicit.labelColor).toEqual(undefined); // blue would get applied with Vega's logic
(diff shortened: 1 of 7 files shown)


test/compile/axis/parse.test.ts
```diff
@@ -130,6 +130,70 @@ describe('Axis', () => {
...
+          axisX: {labelColor: 'blue'},
+          style: {
+            foo: {labelColor: 'red'}
+          }
+        }
+      });
+
+      const axisComponent = parseUnitAxes(model);
+      expect(axisComponent['x'].length).toEqual(1);
+      expect(axisComponent['x'][0].implicit.labelColor).toEqual(undefined); // blue would get applied with Vega's logic
```
donghaoren: Shouldn't this be `blue`? If not, could you revise the comment ` blue would get applied with Vega's logic`?

kanitw: This is intentionally not "blue" since config.axisX is a Vega property, so we don't need to output redundant "blue" in the output axis.

That's why I put the comment, but perhaps it's not clear yet.

PR #6084, opened 2020-03-12
fix: make predicate value support experimental signals

kanitw:
Merge https://github.com/vega/vega-lite/pull/6083 first


src/predicate.ts
```diff
@@ -117,7 +118,7 @@ export interface FieldRangePredicate extends FieldPredicateBase {
    * @maxItems 2
    * @minItems 2
    */
-  range: (number | DateTime | null)[];
+  range: (number | DateTime | null | SignalRef)[];
```
domoritz: Should this be `(number | DateTime | null | SignalRef)[]` or `(number | DateTime | null)[] | SignalRef`?

kanitw: Ah yes, technically both.

kanitw:
I added signal support for range, take a look again?

kanitw:
I guess this should be fine.

PR #6088, opened 2020-03-13
fix: make conditional axis property support conditional signals

kanitw:


src/axis.ts
```diff
@@ -181,8 +181,8 @@ export const CONDITIONAL_AXIS_PROP_INDEX: Record<
   title: null // title supports signal, let's use it.
 };
 
-export type ConditionalAxisProperty<V extends Value | number[]> = ValueDef<V> & {
-  condition: ConditionalPredicate<ValueDef<V>> | ConditionalPredicate<ValueDef<V>>[];
+export type ConditionalAxisProperty<V extends Value | number[]> = (ValueDef<V> | SignalRef) & {
```
domoritz: This would add signals to every property even if the property does not support signals, though.

domoritz: Or do we always know that the property supports signals because that's how we implement conditionals in the first place?

kanitw: Yes, conditional axis property must either match to an encoding channel (which always support signal) or to a property that supports signal.

PR #6095, opened 2020-03-15
fix: make concat/repeat avoid using width/height

kanitw:


kanitw:
I'm merging this into https://github.com/vega/vega-lite/pull/6002, so let's merge this PR by hand to avoid dumb conflicts.

domoritz:
👍

PR #6096, opened 2020-03-15
fix: make legend title work with multi-view resolution

### LINKED ISSUE #6087: Explicit legend title: null gets overridden by multi-view resolution.
According to the encoding docs, setting the legend `title` property to `null` should cause no title to be shown. This works fine with a single view. However, in a multi-view setup with a shared legend it seems that *all* participating encodings need to have `"title": null`, otherwise a default title is used. 

Is the current behavior by design, or a bug? My expectation was that a single *explicit* setting would disable the title, assuming no other participating encodings specify the title.

kanitw:
fix #6087

### DIFF
--- a/src/compile/legend/parse.ts
+++ b/src/compile/legend/parse.ts
@@ -209,7 +209,7 @@ function getProperty<K extends keyof LegendComponentProps>(
       ) as LegendComponentProps[K];
 
     case 'title':
-      return (fieldDefTitle(fieldDef, model.config, {allowDisabling: true}) || undefined) as LegendComponentProps[K];
+      return fieldDefTitle(fieldDef, model.config, {allowDisabling: true}) as LegendComponentProps[K];
 
     case 'type':
       return type({legend, channel, timeUnit, scaleType, alwaysReturn: false}) as LegendComponentProps[K];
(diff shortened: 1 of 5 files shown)


src/compile/legend/parse.ts
```diff
@@ -209,7 +209,7 @@ function getProperty<K extends keyof LegendComponentProps>(
       ) as LegendComponentProps[K];
 
     case 'title':
-      return (fieldDefTitle(fieldDef, model.config, {allowDisabling: true}) || undefined) as LegendComponentProps[K];
+      return fieldDefTitle(fieldDef, model.config, {allowDisabling: true}) as LegendComponentProps[K];
```
domoritz: What? So does `fieldDefTitle` return null or ''? Should we refactor the function instead?

kanitw: Yes, but null or "" is fine.  Previously we only do this to minimize the Vega output so the title doesn't output anything, but it leads to confusion in the merging process.  At this point, it's not worth the simplification. That said, if you strongly think we should simplify the output, you can add some logic to assemble. (I'd rather not.)

kanitw:
Actually this is failing test ... sigh (dumb github action)

PR #6104, opened 2020-03-15
feat: support "width" / "height" in the range array

### LINKED ISSUE #5501: Make x-/y-scale range support "width" and "height"
Right now we support 

```
"mark": {"type": "bar", ..., "y2": "height", "y2Offset": 20},
```

and 

```
"encoding": {
  "y2": {"value": "height"}
}
```

But 

scale.range = `["height", 40]` doesn't work

kanitw:
fix #5501

This could be important for making a layer not using the whole width height.

Even if we have signals, users don't always know the width/height signal names in multi-view settings without looking at vega, so this will still be useful.

### DIFF
--- a/src/compile/scale/range.ts
+++ b/src/compile/scale/range.ts
@@ -111,11 +111,7 @@ export function parseRangeForChannel(channel: ScaleChannel, model: UnitModel): E
                   if (v === 'width' || v === 'height') {
                     // get signal for width/height
 
-                    // Note that, just like default range logic below, these range signals are temporary
-                    // as they can be merged and renamed.
-                    // (We do not have the right size signal here since parseLayoutSize() happens after parseScale().)
-                    // We will later replace these temporary names with
-                    // the final name in assembleScaleRange()
+                    // Just like default range logic below, we use SignalRefWrapper to account for potential merges and renames.
 
                     const sizeSignal = model.getName(v);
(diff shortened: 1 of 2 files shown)


src/compile/scale/range.ts
```diff
@@ -105,7 +105,27 @@ export function parseRangeForChannel(channel: ScaleChannel, model: UnitModel): E
...
+            if (isArray(specifiedScale.range) && (channel === 'x' || channel === 'y')) {
+              return makeExplicit(
+                specifiedScale.range.map(v => {
+                  if (v === 'width' || v === 'height') {
+                    // get signal for width/height
+
+                    // Note that, just like default range logic below, these range signals are temporary
+                    // as they can be merged and renamed.
+                    // (We do not have the right size signal here since parseLayoutSize() happens after parseScale().)
+                    // We will later replace these temporary names with
```
domoritz: I'm confused by this. If we later replace these temporary names, why do we need them? The idea with the SignalRefWrapper is that you don't have to explicitly rename later but that the closure handles the renaming.

kanitw: It’s the same comment from below that i copy from.

domoritz: Ahh, I didn't see this on Github. The comment is wrong then. Please delete it or fix it.

PR #6106, opened 2020-03-15
fix: rule orientation for binned data (e.g., error bar for binned data)

### LINKED ISSUE #5489: Slanted error bars when data is binned
When plotting error bars on quantitative data that is binned, the error bars show up slanted.
Here's a minimalist reproduction that I've tested in the Vega online editor (this is the barley yield error bar example, just that the y-axis shows binned years rather than variety):
```json
{
  "encoding": {"y": {"field": "variety", "type": "ordinal"}},
  "layer": [
    {
      "mark": {"type": "errorbar", "extent": "ci"},
      "encoding": {
        "x": {
          "type": "quantitative",
          "field": "yield",
          "scale": {"zero": false}
        },
        "y": {"type": "quantitative", "bin": true, "field": "year"}
      }
    },
    {
      "mark": {"type": "point", "color": "black", "filled": true},
      "encoding": {
        "x": {"type": "quantitative", "aggregate": "mean", "field": "yield"},
        "y": {"type": "quantitative", "bin": true, "field": "year"}
      }
    }
  ]
}
```
In this example I'd like to plot the barley yield for binned years. (It's binning only over single years, which doesn't really make sense, but illustrates the problem).
Here's the chart this specification produces for me:
image

Is there a way to work around this issue?

kanitw:
fix #5489

image

### DIFF
--- a/src/compile/mark/init.ts
+++ b/src/compile/mark/init.ts
@@ -149,7 +149,7 @@ function orient(mark: Mark, encoding: Encoding<string>, specifiedOrient: Orienta
     // falls through
     case RULE:
       // return undefined for line segment rule and bar with both axis ranged
-      // we have to ignore that case that the data are already binned though
+      // we have to ignore the case that the data are already binned
       if (x2 && !(isFieldDef(x) && isBinned(x.bin)) && y2 && !(isFieldDef(y) && isBinned(y.bin))) {
         return undefined;
       }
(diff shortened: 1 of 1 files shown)


src/compile/mark/init.ts
```diff
@@ -149,7 +149,8 @@ function orient(mark: Mark, encoding: Encoding<string>, specifiedOrient: Orienta
     // falls through
     case RULE:
       // return undefined for line segment rule and bar with both axis ranged
-      if (x2 && y2) {
+      // we have to ignore that case that the data are already binned though
```
domoritz:
```suggestion
      // we have to ignore the case that data is already binned
```

PR #6109, opened 2020-03-15
feat: fix axis disabling for layer + add `config.axis.disable`

### LINKED ISSUE #3814: No easy way to disable all axes / legends by default.
When we want to create multiple layers with no axis at all, this becomes quite tedious.

We also can't just allow `config.axis=null` because if some axis is enabled, we still need to read the default from `config.axis`.  

So perhaps we want config like `config.axis.enabled`, same for legend.

### LINKED ISSUE #5407: Merge axis disabling
See this [simple boxplot spec with `axis: null`](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoAWOgCtY3AHaQANOCgYATgGtsYYJDwBPQsn2R23AB6Ex3PJAC+6iJAAm6PJhyHWWsQt8IhJyKloGZjZOHiFaEU9vWGQ8WFJE9FJUbgpGZFh5RRUXDUhkJXhud0Ylan1QCDdreo1GqAAzPLF3CwBJAFkAEQAhAH0AJS8autc2qBMzCwBHVnQlPGYpqjVWxsh0a0ZYfSVWMTFdpw0rpyA). The axis should not be included, yet it remains. Similar behavior occurs with `errorbar` and `errorband` marks.

kanitw:
fix #5407, fix #3814


src/axis.ts
```diff
@@ -240,7 +240,12 @@ export interface AxisPropsWithConditionAndSignal {
 
 export type AxisConfig = VlOnlyGuideConfig &
   AxisConfigBaseWithConditionalAndSignal &
-  Pick<Axis, 'labelExpr' | 'tickCount' | 'style'>;
+  Pick<Axis, 'labelExpr' | 'tickCount' | 'style'> & {
+    /**
+     * Disable axis by default
```
```suggestion
     * Disable axis by default.
```

src/compile/legend/parse.ts
```diff
@@ -86,17 +86,20 @@ function isExplicit<T extends string | number | object | boolean>(
...
+      return !!legend?.values;
     case 'title':
       // title can be explicit if fieldDef.title is set
       if (property === 'title' && value === fieldDef.title) {
         return true;
       }
   }
   // Otherwise, things are explicit if the returned value matches the specified property
-  return value === legend[property];
+  return value === (legend || {})[property];
```
domoritz:
```suggestion
  return value === (legend ?? {})[property];
```

kanitw: why is that better?  When it produces the same result, `||` is probably better performance and leads to simpler JS output  code.

PR #6123, opened 2020-03-16
docs: clarify how repeat wrap works

domoritz:


src/spec/repeat.ts
```diff
@@ -24,8 +24,8 @@ export interface GenericRepeatSpec<U extends GenericUnitSpec<any, any>, L extend
     ResolveMixins {
   /**
    * Definition for fields to be repeated. One of:
-   * 1) An array of fields to be repeated. If `"repeat"` is an array, the field can be referred using `{"repeat": "repeat"}`
-   * 2) An object that mapped `"row"` and/or `"column"` to the listed of fields to be repeated along the particular orientations. The objects `{"repeat": "row"}` and `{"repeat": "column"}` can be used to refer to the repeated field respectively.
+   * 1) An array of fields to be repeated. If `"repeat"` is an array, the field can be referred to as `{"repeat": "repeat"}`. The repeated views are laid out in a wrapped row. You can set the number of columns to control the wrapping.
```
kanitw: > The repeated views are laid out in a wrapped row.

This is not always true.  Do you mean "by default"?

domoritz: I think it is. You can have a wrapped row with one column and then you get a column.

kanitw: oh right, sorry.

Issue #6129, opened 2020-03-19
Support labelFontWeight for facet header guide


jheer:
According to the [`header` docs](https://vega.github.io/vega-lite/docs/header.html) there is no `labelFontWeight` option for header labels, while there is a `titleFontWeight`. FWIW I recently ran into a use case where I wanted to change the label font weight, but couldn't do so easily (I think I would need to change the style config). It would be helpful to make this configurable.

kanitw:
Thanks @jheer  -- will be fixed in https://github.com/vega/vega-lite/pull/6130

PR #6130, opened 2020-03-19
feat: add missing header `labelBaseline/FontWeight/LineHeight`

kanitw:
Fix https://github.com/vega/vega-lite/issues/6129


src/header.ts
```diff
@@ -130,6 +134,13 @@ export interface CoreHeader extends FormatMixins {
    */
   labelAlign?: Align | SignalRef;
 
+  /**
+   * The vertical text baseline for the header labels. One of `alphabetic` (default), `top`, `middle`, `bottom`, `line-top`, or `line-bottom`.
```
domoritz: Should strings be quoted?

PR #6133, opened 2020-03-19
docs: link to docs pages for  config and datasets

domoritz:
Fixes vega/vega-lite-api#30

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -16075,7 +16075,7 @@
         },
         "config": {
           "$ref": "#/definitions/Config",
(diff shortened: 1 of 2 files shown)


kanitw:
While this PR has a good intention, I think this PR is a bit problematic. 

1) Config is already well linked in the table

image

2) Even if we link just `datasets` and `config`, there are hundreds of other properties that do not get linked in the same way. So it doesn't really solve the problem. 


I think the right fix is to annotate type like `Config` to have a `@see/@docs`  or something tag its docs URL, so we can add it to the schema.  Then we can leverage these in the table and eliminate `link.yml`.

domoritz:
Let’s close it. Vega lite api should use our link file then.

PR #6136, opened 2020-03-20
fix: don't do conditional test for invalid values when `mark.invalid` = null

kanitw:


kanitw:
I'll merge for now, since this isn't causing any spec diff. It should be safe to merge.  Feel free to send follow up comments.

PR #6141, opened 2020-03-22
fix: correct repeat child name logic and data precedence

kanitw:
Fix follow-up comments in https://github.com/vega/vega-lite/pull/6002

### DIFF
--- a/src/normalize/core.ts
+++ b/src/normalize/core.ts
@@ -104,6 +104,7 @@ export class CoreNormalizer extends SpecMapper<NormalizerParams, FacetedUnitSpec
           };
 
           const childName =
+            (childSpec.name || '') +
             repeaterPrefix +
             'child__' +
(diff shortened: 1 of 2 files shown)


src/normalize/core.ts
```diff
@@ -103,13 +103,17 @@ export class CoreNormalizer extends SpecMapper<NormalizerParams, FacetedUnitSpec
             column: columnValue
           };
 
-          const child = this.map(childSpec, {...params, repeater: childRepeater});
-
-          child.name =
+          const childName =
+            (childSpec.name || '') +
```
domoritz:
```suggestion
            (childSpec.name ?? '') +
```

kanitw: why is that more correct?  I don't think we really should use `??` everywhere when it doesn't matter?

PR #6142, opened 2020-03-22
feat: support repeat layer

kanitw:
Merge #6141 first

Implementation for https://github.com/vega/vega-lite/issues/1274


src/spec/repeat.ts
```diff
@@ -15,10 +16,19 @@ export interface RepeatMapping {
   column?: string[];
 }
 
+export interface LayerRepeatMapping extends RepeatMapping {
+  /**
+   * An array of fields to be repeated layers.
```
domoritz:
```suggestion
   * An array of fields to be repeated as layers.
```

PR #6147, opened 2020-03-23
refactor:  extract Position/Legend/ScaleMixins in channeldef

kanitw:
This would be useful for future changes to datum


kanitw:
subsumed by https://github.com/vega/vega-lite/pull/6149

PR #6149, opened 2020-03-23
refactors: make ChannelDef type more accurate + etc. 

kanitw:

PR #6150, opened 2020-03-24
feat: support datum in encoding

### LINKED ISSUE #1601: Encoding's `datum` for constant values in data domain that still pass through scales

kanitw:
Replaces #4201 

Fix #1601

Merge #6149 first!

# TODO

- check all
  - [x] isFieldDef
  - [x] getFieldDef
  - [x] model.fieldDef
  - [x] hasConditionalFieldDef
  - [x] channelHasField
  - [x] forEachFieldDef
  - [x] reduceFieldDef
  - [x] support datum for lat/long
  
  - [x] repeat in datum 

# Examples 

- [x] simple layer datum
- [x] repeat_layer
- [x] layer_bar_annotations

# Later

- [ ] docs
- [ ] Make datum support date time object
- [x] formatSignalRef?
- [ ] we should remove ValueDef from lat/long https://github.com/vega/vega-lite/issues/6154
- [x] weird issue with Encoding<string> can't be assigned to Encoding><Field>


domoritz:
I'd expect this to work

```
{
  "data": {
    "values": [
      {"a": "A", "b": 28},
      {"a": "B", "b": 55},
      {"a": "C", "b": 43},
      {"a": "D", "b": 91},
      {"a": "E", "b": -81},
      {"a": "F", "b": 53},
      {"a": "G", "b": -19},
      {"a": "H", "b": 87},
      {"a": "I", "b": 52}
    ]
  },
  "layer": [
    {
      "mark": "line",
      "encoding": {
      }
    },
    {"mark": "rule", "encoding": {"y": {"datum": 0}}}
  ]
}
```

But I am getting `[Warning] Dropping {"datum":0} from channel "y" since it does not contain any data field, value, or signal.`

kanitw:
> {
>   "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
>   "data": {
>     "values": [
>       {"a": "A", "b": 28},
>       {"a": "B", "b": 55},
>       {"a": "C", "b": 43},
>       {"a": "D", "b": 91},
>       {"a": "E", "b": -81},
>       {"a": "F", "b": 53},
>       {"a": "G", "b": -19},
>       {"a": "H", "b": 87},
>       {"a": "I", "b": 52}
>     ]
>   },
>   "layer": [
>     {
>       "mark": "line",
>       "encoding": {
>         "x": {"field": "a", "type": "ordinal"},
>         "y": {"field": "b", "type": "quantitative"}
>       }
>     },
>     {"mark": "rule", "encoding": {"y": {"datum": 0}}}
>   ]
> }

good catch, fixed.

domoritz:
I also just found the issue :-P

domoritz:
We are off the domain line by half a pixel here but I guess that's a known issue. 

Screen Shot 2020-03-25 at 13 59 15

kanitw:
> We are off the domain line by half a pixel here but I guess that's a known issue.
> 
> Screen Shot 2020-03-25 at 13 59 15

Yeah, and I'd rather suggest people to use conditional gridColor for the zero baseline.  So I think that's not the point of this PR.

domoritz:
In the example above, we are drawing the rule for each item in the dataset. Can we optimize the spec when there is not field encoding?

Screen Shot 2020-03-25 at 14 02 11

domoritz:
It's actually more than an optimization if we use opacity for the rule.

domoritz:
Actually, the right spec is

```
{
  "layer": [
    {
      "data": {
        "values": [
          {"a": "A", "b": 28},
          {"a": "B", "b": 55},
          {"a": "C", "b": 43},
          {"a": "D", "b": 91},
          {"a": "E", "b": -81},
          {"a": "F", "b": 53},
          {"a": "G", "b": -19},
          {"a": "H", "b": 87},
          {"a": "I", "b": 52}
        ]
      },
      "mark": "line",
      "encoding": {
      }
    },
    {"data": {"values": [{}]}, "mark": "rule", "encoding": {"y": {"datum": 0}}}
  ]
}
```

src/channeldef.ts
```diff
@@ -357,6 +376,18 @@ export interface ScaleMixins {
   scale?: Scale | null;
 }
 
+export interface DatumDef<F extends Field = string, V extends Value = Value> extends Partial<TypeMixins<Type>> {
+  /**
+   * A constant value in data domain.
+   */
+  datum?: F extends RepeatRef ? V | RepeatRef : V;
+  // only apply Repeatref is field (F) can be RepeatRef
```
domoritz:
```suggestion
  // only apply Repeatref if field (F) can be RepeatRef
```

PR #6151, opened 2020-03-24
docs: add fruit bar example

domoritz:


site/_data/examples.json
```diff
@@ -512,10 +516,6 @@
         "name": "concat_layer_voyager_result",
         "title": "Comparative Likert Scale Ratings",
         "description": "Comparing Likert scale ratings between two conditions. (Figure 10. from @kanitw et al.'s \"Voyager 2: Augmenting Visual Analysis with Partial View Specifications\" -- http://idl.cs.washington.edu/files/2017-Voyager2-CHI.pdf)."
-      },
-      {
```
domoritz: We had this example already in the gallery.

kanitw: It's ok to show an example twice, as long as they appear in the right sections.

domoritz: It was twice in the same section. 

Screen Shot 2020-03-24 at 15 28 19

kanitw: hahaha ok.

PR #6152, opened 2020-03-25
perf: remove vega-scale dependency

kanitw:
This reduces minified file size from 325kb to 264kb


src/config.ts
```diff
@@ -311,7 +310,17 @@ export const defaultConfig: Config = {
   concat: {spacing: DEFAULT_SPACING}
 };
 
-const tab10 = scheme('tableau10');
+const tab10 = [
+  '#4c78a8',
+  '#f58518',
+  '#e45756',
+  '#72b7b2',
+  '#54a24b…eeca3b',
```
domoritz: This looks wrong, no?

src/config.ts
```diff
@@ -311,7 +310,17 @@ export const defaultConfig: Config = {
   concat: {spacing: DEFAULT_SPACING}
 };
 
-const tab10 = scheme('tableau10');
+const tab10 = [
```
domoritz: Add a comment where this comes from.

PR #6156, opened 2020-03-25
chore: add timer (in ms) to build-example script, so we can benchmark

kanitw:
The console will now output

```
Compiling bar (~160 ms)
```

instead of 

```
Compiling bar
```

Note that `build:example` is much faster than building each example in the parallel version (`build:examples`)

### DIFF
--- a/scripts/build-example.sh
+++ b/scripts/build-example.sh
@@ -4,10 +4,13 @@ set -eo pipefail
 
 dir=${dir-"examples/compiled"}
 
+# mac osx doesn't have ms time, need to use python -- https://apple.stackexchange.com/a/314573
+start=`python -c 'from time import time; print int(round(time() * 1000))'`
+end=`python -c 'from time import time; print int(round(time() * 1000))'`
+timertime=$((end-start))
+
 for name in "$@"
 do
-  echo "Compiling $name" # to $dir (nopatch=$nopatch, forcesvg=$forcesvg)"
-
   # compile normalized example if $skipnormalize is not set
   if [ ! -n "$skipnormalize" ]; then
     #build full vl example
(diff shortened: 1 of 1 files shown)


scripts/build-example.sh
```diff
@@ -16,15 +19,24 @@ do
...
   # 1) Vega spec has changed
   # 2) The SVG file does not exist (new example would not have vg file diff)
   # or 3) the forcesvg environment variable is true
-  if (! git diff $nopatch --exit-code $dir/$name.vg.json || [ ! -f $dir/$name.svg ] || $forcesvg)
-  then
-    rm -f examples/compiled/$name.svg
-    node_modules/vega-cli/bin/vg2svg --seed 123456789 examples/compiled/$name.vg.json > examples/compiled/$name.svg -b .
-  fi
+
+  # if (! git diff $nopatch --exit-code $dir/$name.vg.json || [ ! -f $dir/$name.svg ] || $forcesvg)
```
domoritz: So why not delete this?

scripts/build-example.sh
```diff
@@ -16,15 +19,24 @@ do
 
   # compile Vega example
   rm -f examples/compiled/$name.vg.json
+
+  start=`python -c 'from time import time; print int(round(time() * 1000))'`
```
domoritz: Why not just use `time`?

domoritz: Or at least `date +%s`.

kanitw: Both of them do not have millisecond precision.

kanitw: See the comment on the top of the file

scripts/build-example.sh
kanitw: forgot too uncomment, sorry

scripts/build-example.sh
```diff
@@ -4,10 +4,13 @@ set -eo pipefail
 
 dir=${dir-"examples/compiled"}
 
+# mac osx doesn't have ms time, need to use python -- https://apple.stackexchange.com/a/314573
+start=`python -c 'from time import time; print int(round(time() * 1000))'`
```
domoritz: Python 2 is dead. Let's use Python 3's print function.

domoritz: Actually, perl is ways faster `perl -MTime::HiRes=time -e 'printf "%d\n", time*1000'`.

PR #6158, opened 2020-03-25
refactor: update type guards to use `in` where reasonable

kanitw:


src/transform.ts
```diff
@@ -609,50 +609,50 @@ export interface LoessTransform {
...
-  return t['flatten'] !== undefined;
+  return 'flatten' in t;
 }
 export function isCalculate(t: Transform): t is CalculateTransform {
-  return t['calculate'] !== undefined;
+  return 'calculate' in t;
 }
 
 export function isBin(t: Transform): t is BinTransform {
-  return !!t['bin'];
```
domoritz: What if `bin: false`? We should not change this one here.

kanitw: bin's type is `true | BinParams`. 

If we worry about falsy for the primary properties, then maybe we shouldn't change this for all transforms?

For example, 

`{aggregate: null/undefined}` is probably not a reasonable AggregateTransform?

domoritz: Bin is the only one I'm really worried about since people may expect that `false` works since we support `true`.

kanitw: But why would you do that in transform? 
If so, does that make it a bin transform or not? Perhaps an inactive bin transform? So “in” is still correct.

domoritz: Ohh, nm, I looked wrong.

domoritz:
Runtime tests are still failing.

Issue #4226, opened 2018-10-19
vconcat removes grid lines


SyntaxRules:
I have a simple chart with gridlines:

![image](https://user-images.githubusercontent.com/4117657/47231747-07d34000-d38b-11e8-8088-ab5421360b20.png)
[...]

PR #6173, opened 2020-03-26
refactor: use view height and width instead of gridScale

domoritz:
Relevant issue: https://github.com/vega/vega-lite/issues/4226


kanitw:
I think there was a reason why we added it. 

Maybe it's not true anymore, but given this massive diff, I'd rather not merge unless you are willing inspect all of these and make sure.

kanitw:
There are definitely some broken ones:

image

domoritz:
I'm looking into it.

domoritz:
We added it to gridlines are spanning the whole width/height of the view.

domoritz:
The examples now work correctly.

kanitw:
Have you looked at all output? (esp. layer and multi-view?) 

I'm suspecting this could be wrong for layer and multi-views, so we should at least check *all* of them.

domoritz:
Let's not do it for now.

PR #6181, opened 2020-03-27
docs: add examples for tickband

### LINKED ISSUE #6171: How to place gridlines between ticks for ordinal data
When using an ordinal encoding, I might still want gridlines to make it easy to see what marks belong together. 

```json
{
  "mark": "tick",
  "encoding": {
    "y": {
      "field": "Cylinders",
      "type": "ordinal",
      "axis": {"grid": true}
    }
  }
}
```

But rather than having lines behind the marks, I'd like the lines between marks. 

image

Similar to binned data

image

cc @ldegreef

domoritz:
Fixes #6171


site/docs/encoding/axis.md
```diff
@@ -84,6 +84,10 @@ By default, Vega-Lite automatically sets the axis extent (the space axis ticks a
 
 {% include table.html props="ticks,tickBand,tickColor,tickCount,tickDash,tickExtra,tickMinStep,tickOffset,tickOpacity,tickRound,tickSize,tickWidth,values" source="Axis" %}
 
+Using `tickBand`, we can change the axis ticks and gridlines to be drawn between marks.
```
kanitw: Add `#### Example: ...` ?

domoritz: I added a heading to another example that had it missing.

PR #6185, opened 2020-03-28
feat: add aria properties to guides and add aria-label channel

domoritz:
Needs https://github.com/vega/vega/pull/2380

### DIFF
--- a/src/compile/mark/encode/aria.ts
+++ b/src/compile/mark/encode/aria.ts
@@ -1,0 +1,58 @@
+import {entries} from '../../../util';
+import {getMarkPropOrConfig} from '../../common';
+import {UnitModel} from './../../unit';
+import {wrapCondition} from './conditional';
+import {textRef} from './text';
+import {tooltipData} from './tooltip';
+
+export function aria(model: UnitModel) {
+  const {mark, markDef, config} = model;
+
+  const ariaHidden = getMarkPropOrConfig('ariaHidden', markDef, config);
+
+  // we can ignore other aria properties if ariaHidden is true
+  if (ariaHidden === true) {
+    return {
+      ariaHidden: {
+        value: true
+      }
+    };
+  }
+
+  return {
+    ...(ariaHidden ? {ariaHidden} : {}),
+    ariaRoleDescription: {value: mark},
+    ariaRole: {value: 'graphics-symbol'},
+    ...ariaLabel(model)
+  };
+}
+
+export function ariaLabel(model: UnitModel) {
+  const {encoding, markDef, config, stack} = model;
+  const channelDef = encoding.ariaLabel;
+
+  if (channelDef) {
+    return wrapCondition(model, channelDef, 'ariaLabel', cDef => textRef(cDef, model.config));
+  }
+
+  // Use default from mark def or config if defined.
+  // Functions in encode usually just return undefined but since we are defining a default below, we need to check the default here.
+  const ariaLabelValue = getMarkPropOrConfig('ariaLabel', markDef, config);
+  if (ariaLabelValue != null) {
+    return {
+      ariaLabel: {
+        value: ariaLabelValue
+      }
+    };
+  }
+
+  const data = tooltipData(encoding, stack, config);
+
+  return {
+    ariaLabel: {
+      signal: entries(data)
+        .map(({key, value}) => `${key} + " is " + (${value})`)
+        .join(' + ", " + ')
+    }
+  };
+}
--- a/src/compile/mark/encode/tooltip.ts
+++ b/src/compile/mark/encode/tooltip.ts
@@ -15,6 +15,7 @@ import {
 import {Config} from '../../../config';
 import {Encoding, forEach} from '../../../encoding';
 import {StackProperties} from '../../../stack';
+import {entries} from '../../../util';
 import {getMarkPropOrConfig} from '../../common';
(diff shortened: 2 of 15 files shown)


domoritz:
To try

```json
{
  "data": {
    "values": [
      {"a": "A", "b": 28}, {"a": "B", "b": 55}, {"a": "C", "b": 43},
      {"a": "D", "b": 91}, {"a": "E", "b": 81}, {"a": "F", "b": 53},
      {"a": "G", "b": 19}, {"a": "H", "b": 87}, {"a": "I", "b": 52}
    ]
  },
  "mark": {
    "type": "bar",
	"ariaLabel": "a label for all marks"
  },
  "encoding": {
    "x": {"field": "a", "type": "ordinal", "axis": {"ariaLabel": "hello"}},
  }
}

```

src/compile/mark/encode/aria.ts
```diff
@@ -0,0 +1,58 @@
...
+  const channelDef = encoding.ariaLabel;
+
+  if (channelDef) {
+    return wrapCondition(model, channelDef, 'ariaLabel', cDef => textRef(cDef, model.config));
+  }
+
+  // Use default from mark def or config if defined.
+  // We usually just return undefined but since we are defining a default below, we need to check the default here.
+  const ariaLabelValue = getMarkPropOrConfig('ariaLabel', markDef, config);
+  if (ariaLabelValue) {
```
kanitw, line 41: what if the value is `""`?

src/encoding.ts
```diff
@@ -283,6 +283,11 @@ export interface Encoding<F extends Field> {
    */
   url?: StringFieldDefWithCondition<F> | StringValueOrSignalDefWithCondition<F>;
 
+  /**
+   * A string that labels the axis group for accessibility.
+   */
+  ariaLabel?: StringFieldDefWithCondition<F> | StringValueOrSignalWithCondition<F> | StringFieldDef<F>[];
```
kanitw, line 289: I thought we said that the initialize version doesn't need array?

src/compile/mark/encode/aria.ts
```diff
@@ -0,0 +1,58 @@
...
+export function ariaLabel(model: UnitModel) {
+  const {encoding, markDef, config, stack} = model;
+  const channelDef = encoding.ariaLabel;
+
+  if (channelDef) {
+    return wrapCondition(model, channelDef, 'ariaLabel', cDef => textRef(cDef, model.config));
+  }
+
+  // Use default from mark def or config if defined.
+  // We usually just return undefined but since we are defining a default below, we need to check the default here.
```
kanitw, line 39: > We usually just return undefined but since we are defining a default below, we need to check the default here.

Who are "we"?  Make it clear if it's "this method" or something.

src/compile/mark/encode/tooltip.ts
```diff
@@ -63,17 +63,15 @@ export function tooltip(model: UnitModel, opt: {reactiveGeom?: boolean} = {}) {
   }
 }
 
-export function tooltipRefForEncoding(
+export function tooltipData(
```
kanitw, line 66: Instead of splitting this method, how about adding `entryMapper` and `delimiter` as optional params?

src/compile/mark/encode/aria.ts
domoritz: Fixed

src/encoding.ts
domoritz: I forgot to change it back. Fixed.

src/compile/mark/encode/tooltip.ts
domoritz: I doubt that it's cleaner. At least now we can test tooltipData independently. Let me know if you feel strongly about this.

src/compile/mark/encode/tooltip.ts
```diff
@@ -62,17 +63,15 @@ export function tooltip(model: UnitModel, opt: {reactiveGeom?: boolean} = {}) {
...
   encoding: Encoding<string>,
   stack: StackProperties,
   config: Config,
-  {reactiveGeom}: {reactiveGeom?: boolean} = {}
```
kanitw, line 69: Why do you need to change this line?

src/compile/mark/encode/tooltip.ts
```diff
@@ -116,12 +115,26 @@ export function tooltipRefForEncoding(
...
+export function tooltipRefForEncoding(
+  encoding: Encoding<string>,
+  stack: StackProperties,
+  config: Config,
+  {reactiveGeom}: {reactiveGeom?: boolean} = {}
+) {
+  return tooltipRefData(tooltipData(encoding, stack, config, reactiveGeom));
+}
+
+export function tooltipRefData(data: Dict<string>) {
```
kanitw, line 137: Why is this `tooltipRefData`?  This name sounds weird.

I think the two lines here should be just a part of `tooltipRefForEncoding`.

domoritz: Good idea. Done.

src/compile/mark/encode/tooltip.ts
domoritz: fixed

kanitw:
We can merge once we update to a new Vega version with relevant support.

domoritz:
Replaced by #6391

PR #6191, opened 2020-03-28
docs: datum + misc

kanitw:


kanitw:
hmms looks like the schema change is problematic.

domoritz:
What's problematic? Anything I need to fix in the generator?

PR #6194, opened 2020-03-28
feat(examples): add Multi-Series Line Chart with Halo

kanitw:
Note that I intentionally call example as feature -- so we can feature them in a release note.


examples/specs/line_color_halo.vl.json
```diff
@@ -0,0 +1,29 @@
...
+    "groupby": ["date"]
+  }],
+  "repeat": {
+    "layer": ["AAPL", "AMZN", "GOOG", "IBM", "MSFT"]
+  },
+  "spec": {
+    "layer": [{
+      "mark": {"type": "line", "stroke": "white", "strokeWidth": 4},
+      "encoding": {
+        "x": {"field": "date", "type": "temporal"},
```
domoritz: Do you want to pull x and y out of the individual layers?

kanitw: Just one encoding, probably not worth the additional complexity to read the code.

PR #6197, opened 2020-03-28
docs: improve co2 concentration to be more efficient

kanitw:
We now use aggregate instead of window.

The code is a bit longer, but it's easier to understand.


kanitw:
test failing somehow

domoritz:
In transform extract

kanitw:
Don't merge yet, maybe I can fix it.

PR #6200, opened 2020-03-29
feat: feature branch for angle related features 

### LINKED ISSUE #4548: Add angle channel for symbols

kanitw:
fix #4548


kanitw:
Now that we'll use `theta` for arc marks, there won't be a breaking change here. Since we already approved #5675, I'll merge this once test passes, so I don't have to keep rebasing the feature branch.

PR #6201, opened 2020-03-29
feat: arc mark

kanitw:
This may replace #5676 

- [x] modify all the specs to use theta/theta2

- [x] example: arc radial
- [x] example: layering text 
  - [x] add radius/theta for text 

- [x] support outerRadius / innerRadius in markDef and config -- pending   https://github.com/vega/vega/pull/2484

- [x] docs
  - [x] encoding.md
  - [x] arc.md 

- [x] radius(2)Offset

- [x] redirect theta/theta2/radius/radius2 config

- [x] test
  - [x] bin arc position
  - [x] radius scale range 

## LATER

- [ ] normalize angle to theta when reasonable https://github.com/vega/vega-lite/issues/6229


examples/compiled/layer_falkensee.vg.json
```diff
@@ -126,7 +126,9 @@
       "encode": {
         "update": {
           "fill": {"scale": "color", "field": "event"},
-          "x": {"scale": "x", "field": "year_start"},
+          "x": {
```
kanitw: this is wrong

site/docs/encoding.md
```diff
@@ -70,6 +70,7 @@ An integral part of the data visualization process is encoding data with visual
 The keys in the `encoding` object are encoding channels. Vega-Lite supports the following groups of encoding channels
 
 - [Position Channels](#position): `x`, `y`, `x2`, `y2`, `xError`, `yError`, `xError2`, `yError2`
+- TODO: add angular channels
```
kanitw: Add docs

src/compile/mark/encode/valueref.ts
```diff
@@ -323,3 +341,29 @@ export function widthHeightValueRef(channel: Channel, value: ValueOrGradientOrTe
   }
   return signalOrValueRef(value);
 }
+
+export function getValueFromMarkDefAndConfig({
```
kanitw: Hmm, this should be extracted. I'll extract and rebase later.

examples/compiled/layer_falkensee.vg.json
domoritz: Is it still wrong?

kanitw: no diff anymore.

domoritz:
Yet another example where not using the outline would be better :-P

Screen Shot 2020-03-31 at 21 07 11

examples/specs/arc_donut.vl.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
+  "description": "A simple donut chart with embedded data.",
+  "data": {
+    "values": [
+      {"id": 1, "field": 4},
```
domoritz, line 6: id and field are weird names. Let's replace with `category` and `value`.

src/compile/mark/base.ts
```diff
@@ -8,7 +9,7 @@ export interface MarkCompiler {
   /**
    * Underlying vega Mark type for the Vega-Lite mark.
```
domoritz, line 10:
```suggestion
   * Underlying Vega Mark type for the Vega-Lite mark.
```

src/compile/mark/base.ts
```diff
@@ -1,3 +1,4 @@
+import {Mark} from 'vega-typings/types';
```
domoritz, line 1: Can we import from Vega instead?

test/compile/scale/range.test.ts
```diff
@@ -264,6 +265,20 @@ describe('compile/scale', () => {
...
+      it('should use default radius.', () => {
+        const model = parseUnitModelWithScaleExceptRange({
+          mark: 'text',
+          encoding: {
+            radius: {field: 'x', type: 'quantitative'}
+          }
+        });
+        const r = parseRangeForChannel('radius', model);
+        expect(r.value[0]).toEqual(0);
+        expect(r.value[1]).toBeInstanceOf(SignalRefWrapper);
```
domoritz, line 278:
```suggestion
        expect(r.value[1]).toEqual({signal: ...});
```

PR #6204, opened 2020-03-30
fix: only apply descending stack order for y-stack

kanitw:
It's weird to have the stack from right to left for x-stack.  

(It makes sense to flip y as the colors go from top to bottom.)

Plus, This is also causing incorrect rendering in the arc branch.


kanitw:
Since this is definitely the right thing to do and it's blocking many other PRs, I'll merge first. 

Feel free to submit follow up comments.

Issue #6209, opened 2020-03-30
 VL 4.8 breaks autosize:fit calculation in concat views autosize:fit sometimes doesn't work with concat views


jheer:
[...]

PR #6226, opened 2020-03-31
revert: let concat use top-level width/height when feasible

kanitw:
reverting  85264bb

part of #6209


domoritz:
Can you add a test case?

kanitw:
Ok, adding interactive_seattle_weather_autosize.vl.json

domoritz:
Maybe a smaller example? That example is 600loc in Vega.

domoritz:
Thanks!

PR #6228, opened 2020-03-31
feat(example): layer_text_heatmap_joinaggregate – use joinaggregate to determine label color threshold

domoritz:


kanitw:
Let's use `feat` for new examples, so we can "feature" them in the release log.

kanitw:
Nice example, btw.

Issue #6240, opened 2020-04-02
LegendLayout schema is empty


marcprux:
VL 4.9.0's schema's (https://github.com/vega/schema/blob/master/vega-lite/v4.9.0.json#L10787) `LegendLayout` def is not being generated, and `LegendConfig.layout` is an odd `not` construct:

```json
    "layout": {
        "description": "Legend orient group layout parameters.",
          "not": {
        }
```

The last schema to have the `layout` property defined correctly was 4.1.1 (https://github.com/vega/schema/blob/master/vega-lite/v4.1.1.json#L9515): 

```json
        "layout": {
          "$ref": "#/definitions/LegendLayout",
          "description": "Legend orient group layout parameters."
        }
```

So it would seem that this issue was introduced around 4.2.0.

marcprux:
This may have something to do with `export interface LegendLayout extends BaseLegendLayout` in https://github.com/vega/vega/blob/eced44eb31fc4d64d36b165a360430fee450ad9e/packages/vega-typings/types/spec/config.d.ts#L451 

If I manually remove `extends BaseLegendLayout` and re-generate the typings, it does generate a ref:

```shell

marc@bix vega-lite % yarn schema && grep --after-context=3 'layout": ' build/vega-lite-schema.json
yarn run v1.22.4
$ mkdir -p build && ts-json-schema-generator -f tsconfig.json -p src/index.ts -t TopLevelSpec --no-type-check --no-ref-encode > build/vega-lite-schema.json && yarn renameschema && cp build/vega-lite-schema.json site/_data/
$ scripts/rename-schema.sh
✨  Done in 4.15s.
        "layout": {
          "$ref": "#/definitions/LegendLayout",
          "description": "Legend orient group layout parameters."
        },
```

marcprux:
If I rename `BaseLegendLayout.offset` to `BaseLegendLayout.offsetXXX` in `config.d.ts`, then the `LegendLayout` ref is generated successfully. Perhaps the typescript generator doesn't like the property name `offset`, or it is conflicting with another `offset` property somewhere?

marcprux:
Perhaps related to https://github.com/vega/ts-json-schema-generator/issues/344 ? It should be noted that the schema is generated fine with `margin?: number | SignalRef`, but fails with `offset?: number | SignalRef`, so it seems like the keyword `offset` is conflicting with something.

marcprux:
[...]

PR #6242, opened 2020-04-02
fix: don't remove value refs from legend config since it shouldn't have them in the first place

domoritz:
Needs https://github.com/vega/vega/pull/2499
Fixes https://github.com/vega/vega-lite/issues/6240


kanitw:
Sounds reasonable, but it's failing tests.

domoritz:
Yes, because it needs vega/vega#2499 :-P

kanitw:
still failinng

domoritz:
After rebasing, tests pass now.

PR #6246, opened 2020-04-03
perf: skip empty & unnecessary axis config

kanitw:
No need to loop over them if they don't exist

### DIFF
--- a/src/compile/axis/parse.ts
+++ b/src/compile/axis/parse.ts
@@ -13,7 +13,7 @@ import {
 } from '../../channeldef';
 import {Config} from '../../config';
 import {isQuantitative} from '../../scale';
-import {contains, getFirstDefined, keys, normalizeAngle, titlecase} from '../../util';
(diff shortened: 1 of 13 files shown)


src/compile/axis/parse.ts
```diff
@@ -287,28 +287,34 @@ function parseAxis(channel: PositionScaleChannel, model: UnitModel): AxisCompone
...
+    } else {
+      const {configValue = undefined, configFrom = undefined} = isAxisProperty(property)
+        ? getAxisConfig(property, model.config, axisConfigTypes, axis?.style)
+        : {};
+
+      if (value !== undefined && configValue === undefined) {
+        // only set property if it is explicitly set or has no config value (otherwise we will accidentally override config)
+        axisComponent.set(property, value, explicit);
+      } else if (
+        // Cases that we need to implicit values
```
domoritz, line 304:
```suggestion
        // Cases that we need implicit values
```

domoritz: I don't know what this message was supposed to say.

PR #6256, opened 2020-04-03
fix: make binSpacing work for values other than 0 and 1

kanitw:
The old logic forgot to split the spacing to both sides while taking `axis.translate` into account. 

Also shift the bins for binSpacing = 0 a bit.


src/compile/mark/encode/position-rect.ts
```diff
@@ -55,6 +55,8 @@ export function rectPosition(model: UnitModel, channel: 'x' | 'y', mark: 'bar' |
     !hasDiscreteDomain(scaleType)
   ) {
     const band = getBand({channel, fieldDef, stack, markDef, config});
+    const axis = model.component.axes[channel]?.[0];
+    const axisTranslate = axis ? axis.get('translate') ?? 0.5 : 0.5;
```
domoritz, line 59: Does this work?

```suggestion
    const axisTranslate = axis?.get('translate') ?? 0.5;
```

domoritz: Also, can you add a comment on why we add 0.5?

PR #6270, opened 2020-04-05
fix: make theta and radius include zero by default

kanitw:
Merge https://github.com/vega/vega-lite/pull/6270 first


domoritz:
#6270 is a circular reference @kanitw

PR #6278, opened 2020-04-05
Misc Axis Refactor

kanitw:
- [x] simplify labelAngle / labelBaseline

### DIFF
--- a/src/compile/axis/config.ts
+++ b/src/compile/axis/config.ts
@@ -1,14 +1,71 @@
+import {ScaleType} from 'vega-typings/types';
+import {array} from 'vega-util';
 import {Axis} from '../../axis';
+import {PositionScaleChannel} from '../../channel';
 import {Config} from '../../config';
+import {isQuantitative} from '../../scale';
+import {titlecase} from '../../util';
 import {getStyleConfig} from '../common';
 
+function getAxisConfigFromConfigTypes(configTypes: string[], config: Config) {
+  // TODO: add special casing to add conditional value based on orient signal
+  return Object.assign.apply(null, [{}, ...configTypes.map(configType => config[configType])]);
+}
+
+export type AxisConfigs = ReturnType<typeof getAxisConfigs>;
+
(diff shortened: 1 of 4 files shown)


src/compile/axis/config.ts
```diff
@@ -1,14 +1,71 @@
...
 
+function getAxisConfigFromConfigTypes(configTypes: string[], config: Config) {
+  // TODO: add special casing to add conditional value based on orient signal
+  return Object.assign.apply(null, [{}, ...configTypes.map(configType => config[configType])]);
+}
+
+export type AxisConfigs = ReturnType<typeof getAxisConfigs>;
+
+export function getAxisConfigs(channel: PositionScaleChannel, scaleType: ScaleType, orient: string, config: Config) {
+  const typeBasedConfigTypes =
```
domoritz, line 18: convert to if?

domoritz: The chain is really hard to read.

src/compile/axis/properties.ts
```diff
@@ -11,7 +11,7 @@ import {NOMINAL, ORDINAL, Type} from '../../type';
 import {contains, normalizeAngle} from '../../util';
 import {isSignalRef} from '../../vega.schema';
 import {UnitModel} from '../unit';
-import {getAxisConfig} from './config';
+import {AxisConfigs, getAxisConfig} from './config';
 
 // TODO: we need to refactor this method after we take care of config refactoring
```
domoritz, line 16: Still relevant

src/compile/axis/config.ts
kanitw: nah, I disagree for this one.

src/compile/axis/properties.ts
kanitw: will remove in a subsequent refactor PR, so we don't have to rebase this one.

src/compile/axis/config.ts
domoritz: Why do you think this ternary is better than some ifs?

PR #6279, opened 2020-04-06
refactor: simplify legend parsing code

domoritz:
* Don't use reduce for large code blocks. 
* Use existing LEGEND_SCALE_CHANNELS


domoritz:
It should not. I wonder whether the output is dependent on the order of some of the properties.

domoritz:
There we go, no changes.

PR #6289, opened 2020-04-06
Refactor axis angle / labelAlign/Baseline

kanitw:


domoritz:
Do you want the example to change in a refactor pull request?

kanitw:
Normally no, but there is only one changing example and it seems to look better, so maybe it's fine?

src/util.ts
```diff
@@ -397,7 +397,7 @@ export function isInternalField(name: string) {
  * Normalize angle to be within [0,360).
  */
 export function normalizeAngle(angle: number) {
-  return ((angle % 360) + 360) % 360;
+  return angle !== undefined ? ((angle % 360) + 360) % 360 : undefined;
```
domoritz, line 400: I think normal ifs make code more understandable. Ternaries make sense when you assign to a variable. 

```suggestion
  if (angle === undefined) {
  	return undefined;
  }
  return ((angle % 360) + 360) % 360;
```

PR #6295, opened 2020-04-07
docs: update wind example

### LINKED ISSUE #6231: Windmap image broken
Screen Shot 2020-03-31 at 23 33 38

domoritz:
towards #6231


kanitw:
Convert to draft? since I still can't preview the SVGs and you say in https://github.com/vega/vega-lite/issues/6231#issuecomment-610682431 that this doesn't fix the problem.

domoritz:
I fixed the data parsing. This still does not solve the image generation for some reason but it's an improvement nonetheless.

domoritz:
I filed https://github.com/vega/vega/issues/2662 for the rendering issue.

kanitw:
FWIW, I'm not convinced that we should project it since we don't show other maps elements anyway, so adding projection only adds complexity to this example.


If we do overlay this windmap on top of actual maps, then it makes more sense to project?

jwoLondon:
I think projection is important at this scale because it is generally bad practice to perform a (lng,lat) -> (x,y) projection when geographic distance between neighbouring cells noticeably varies.

kanitw:
Thanks @jwoLondon.  Ok let's merge this.

PR #6305, opened 2020-04-08
refactor: make axis rule logic not require casting + fix uncovered issues

kanitw:


src/compile/axis/assemble.ts
```diff
@@ -194,7 +194,7 @@ export function assembleAxisSignals(model: Model): NewSignal[] {
   for (const channel of POSITION_SCALE_CHANNELS) {
     if (axes[channel]) {
       for (const axis of axes[channel]) {
-        if (!axis.get('gridScale')) {
+        if (!axis.get('disable') && !axis.get('gridScale')) {
```
kanitw, line 197: FYI, this is the line that's causing example changes.

I added this because without it, it's also changing several more examples.  

Adding this condition is more correct since the affected examples shouldn't need the width/height signals anyway.

PR #6309, opened 2020-04-09
 refactor: consolidate labelAlign logic for x and y

kanitw:


kanitw:
The example changes seem to be improving.

PR #6312, opened 2020-04-09
feat: make axis orient/labelAngle support signal

kanitw:


examples/specs/rect_labelangle_orient_signal_future.vl.json
```diff
@@ -0,0 +1,41 @@
...
+  "data": { "url": "data/github.csv"},
+  "mark": "circle",
+  "encoding": {
+    "y": {
+      "field": "time",
+      "type": "ordinal",
+      "timeUnit": "day",
+      "sort": ["mon", "tue", "wed", "thu", "fri", "sat", "sun"],
+      "axis": {
+        "orient": "right",
```
kanitw, line 13: @haldenl  Once Vega supports it, make orient signal and test.

examples/specs/rect_labelangle_orient_signal_future.vl.json
```diff
@@ -0,0 +1,41 @@
...
+        "labelAngle": {"signal": "yLabelAngle"}
+      }
+    },
+    "x": {
+      "field": "time",
+      "type": "ordinal",
+      "timeUnit": "hours",
+      "axis": {
+        "orient": "top",
```
kanitw, line 22: @haldenl  Once Vega supports it, make orient signal and test.

examples/specs/circle_labelangle_orient_signal_future.vl.json
```diff
@@ -0,0 +1,41 @@
...
+  "data": { "url": "data/github.csv"},
+  "mark": "circle",
+  "encoding": {
+    "y": {
+      "field": "time",
+      "type": "ordinal",
+      "timeUnit": "day",
+      "sort": ["mon", "tue", "wed", "thu", "fri", "sat", "sun"],
+      "axis": {
+        "orient": "right",
```
kanitw, line 13: @haldenl Once Vega supports it, make orient signal and test.

examples/specs/circle_labelangle_orient_signal_future.vl.json
```diff
@@ -0,0 +1,41 @@
...
+        "labelAngle": {"signal": "yLabelAngle"}
+      }
+    },
+    "x": {
+      "field": "time",
+      "type": "ordinal",
+      "timeUnit": "hours",
+      "axis": {
+        "orient": "top",
```
kanitw, line 22: @haldenl Once Vega supports it, make orient signal and test.

kanitw:
I've rebased to master. 
I think this is mergable since we're correctly passing orient to Vega.
(We can revert the casting commit later.)

src/compile/axis/config.ts
```diff
@@ -1,20 +1,56 @@
...
-function getAxisConfigFromConfigTypes(configTypes: string[], config: Config) {
+function getAxisConfigFromConfigTypes(
+  configTypes: string[],
+  config: Config,
+  channel: 'x' | 'y',
+  orient: string | SignalRef
+) {
   // TODO: add special casing to add conditional value based on orient signal
-  return Object.assign.apply(null, [{}, ...configTypes.map(configType => config[configType])]);
+  return Object.assign.apply(null, [
```
domoritz, line 18: Use `...` instead.

src/compile/axis/config.ts
```diff
@@ -1,20 +1,56 @@
...
+    {},
+    ...configTypes.map(configType => {
```
domoritz, line 20: This map is getting pretty big. Let's pull it out and use a normal for loop.

src/compile/axis/config.ts
```diff
@@ -1,20 +1,56 @@
...
+      if (configType === 'axisOrient') {
+        const orient1 = channel === 'x' ? 'bottom' : 'left';
+        const orientConfig1 = config[channel === 'x' ? 'axisBottom' : 'axisLeft'] || {};
+        const orientConfig2 = config[channel === 'x' ? 'axisTop' : 'axisRight'] || {};
+
+        const props = new Set([...keys(orientConfig1), ...keys(orientConfig2)]);
```
domoritz, line 26: No need to make a set here if you only loop over it.

src/compile/axis/config.ts
kanitw: Why would using loop be better?

Using map and then `assign`ing once also means the merging is done in native code, not JS and thus faster.

src/compile/axis/config.ts
kanitw: I already use ... below, so this comment doesn't make sense.

Plus, you can't do ... to merge/assign an array of object.

src/compile/axis/config.ts
kanitw: I disagree. It's the right semantic.

src/compile/axis/config.ts
domoritz: We could if we do it in a for loop. Anyway, not needed.

src/compile/axis/config.ts
domoritz: A loop is easier to step into when using a debugger.

src/compile/axis/config.ts
domoritz: Is it because you want to remove duplicates? If you iterate over values below, an array makes as much sense as a set, imo.

src/compile/axis/properties.ts
```diff
@@ -167,58 +169,111 @@ export function getLabelAngle(
   }
 }
 
+export function normalizeAngleExpr(angle: SignalRef) {
+  return `(((${angle.signal} % 360) + 360) % 360)`;
```
haldenl, line 173: Should parenthesize `${angle.signal}` to preserve evaluation order

e.g., 1 + 359 % 360 !== (1 + 359) % 360

src/compile/axis/properties.ts
```diff
@@ -167,58 +169,111 @@ export function getLabelAngle(
...
+            `(45 < ${a} && ${a} < 135) || (225 < ${a} && ${a} < 315) ? "middle" :` +
+            `(${a} <= 45 || 315 <= ${a}) === ${orientIsTop} ? "bottom" : "top"`
+        };
+      }
+
+      if ((45 < angle && angle < 135) || (225 < angle && angle < 315)) {
+        return 'middle';
+      }
+
+      if (isSignalRef(orient)) {
```
haldenl, line 200: This looks unreachable since `isSignalRef(orient)` implies `channel === 'y'`. (I think we need a signal check on the `channel` assignment line).

src/compile/axis/properties.ts
```diff
@@ -167,58 +169,111 @@ export function getLabelAngle(
...
   if (angle !== undefined) {
     if (channel === 'x') {
-      return (45 < angle && angle < 135) || (225 < angle && angle < 315)
-        ? 'middle'
-        : (angle <= 45 || 315 <= angle) === (axisOrient === 'top')
-        ? 'bottom'
-        : 'top';
+      if (isSignalRef(angle)) {
+        const a = normalizeAngleExpr(angle);
+        const orientIsTop = isSignalRef(orient) ? `(${orient.signal} === "top")` : orient === 'top';
```
haldenl, line 188: I believe we need to parenthesize `${orient.signal}` here as well.

src/compile/axis/properties.ts
```diff
@@ -167,58 +169,111 @@ export function getLabelAngle(
...
+        return {
+          signal:
+            `(45 < ${a} && ${a} < 135) || (225 < ${a} && ${a} < 315) ? "middle" :` +
+            `(${a} <= 45 || 315 <= ${a}) === ${orientIsTop} ? "bottom" : "top"`
```
haldenl, line 192: Parenthesize here too.

haldenl: (And etc. for other occurrences.)

src/compile/header/parse.ts
```diff
@@ -1,16 +1,17 @@
...
 import {assembleAxis} from '../axis/assemble';
 import {FacetModel} from '../facet';
 import {parseGuideResolve} from '../resolve';
 import {getHeaderProperty} from './common';
 import {HeaderChannel, HeaderComponent} from './component';
 
-export function getHeaderType(orient: AxisOrient) {
-  if (orient === 'top' || orient === 'left') {
+export function getHeaderType(orient: AxisOrient | SignalRef) {
+  if (orient === 'top' || orient === 'left' || isSignalRef(orient)) {
```
haldenl, line 14: Why do we default to `header` if orient is a signal?

test/compile/axis/config.test.ts
```diff
@@ -0,0 +1,20 @@
+import {getAxisConfigs} from '../../../src/compile/axis/config';
+
+describe('getAxisConfigs', () => {
+  it('correctly get axis types for orient signal', () => {
+    const {vgAxisConfig} = getAxisConfigs(
+      'x',
+      'point',
+      {signal: 'a'},
+      {
+        axisTop: {labelBaseline: 'line-bottom', labelPadding: 2},
```
haldenl, line 10: Can we change one of these config properties to a signal to make sure objects work okay.

src/compile/axis/properties.ts
kanitw: Why does `isSignalRef(orient)` implies `channel === 'y'`?

kanitw: FWIW, this is *not* unreachable -- look at the test coverage.

src/compile/header/parse.ts
kanitw: Because header / footer can't be signal in Vega.  We should add some comments here.

src/compile/axis/properties.ts
kanitw: Good catch.

src/compile/axis/properties.ts
kanitw: Can you come up with a case that will break this?
Otherwise, I'd rather have the output readable.

src/compile/axis/properties.ts
kanitw: Can you come up with a case that will break this?
Otherwise, I'd rather have the output readable.

test/compile/axis/config.test.ts
kanitw: not sure what you mean by "objects work okay".

haldenl: b/c we use `stringValue` to signalfy the config.

haldenl: if one of the properties is a signal it will stringify to `"{ 'signal': '...' }"`. Should make sure this works properly.

src/compile/axis/config.ts
```diff
@@ -1,20 +1,56 @@
...
+
+        const conditionalOrientAxisConfig = {};
+        for (const prop of props.values()) {
+          conditionalOrientAxisConfig[prop] = {
+            // orient is surely signal in this case
+            signal: `${orient['signal']} === "${orient1}" ? ${stringValue(orientConfig1[prop])} : ${stringValue(
```
kanitw, line 32: wrong when the config is a signal

src/compile/axis/properties.ts
kanitw: Actually ternary operator will break this.

src/compile/axis/properties.ts
kanitw: Let's deal with this separately in https://github.com/vega/vega-lite/issues/6374.

src/compile/axis/properties.ts
kanitw: Let's deal with this in https://github.com/vega/vega-lite/issues/6374

src/compile/axis/properties.ts
kanitw: Let's deal with this in #6374

src/compile/axis/properties.ts
kanitw: I'll remove the `channel` inference code so it's clear.

PR #6322, opened 2020-04-12
fix:  apply offset to both range positions when position2 is unspecified

### LINKED ISSUE #6320: Inconsistency with thetaOffset and theta2Offset for arc mark
The Vega-Lite 4.10 schema has (as yet, undocumented) `thetaOffset` and `theta2Offset`. This is certainly useful, and behaves as I would expect when just `thetaOffset` is specified for radial charts where the segment angle is equal for all marks but radius varies:

```json
{
  "data": {
    "values": [
      {"dir": "1N","strength": 10},
      {"dir": "2NE","strength": 12},
      {"dir": "3E","strength": 10},
      {"dir": "4SE","strength": 12},
      {"dir": "5S","strength": 10},
      {"dir": "6SW","strength": 12},
      {"dir": "7W","strength": 10},
      {"dir": "8NW","strength": 12}
    ]
  },
  "encoding": {
    "theta": {"field": "dir","type": "ordinal"},
    "radius": {"field": "strength","type": "quantitative"},
    "color": {"field": "dir","type": "nominal"}
  },
  "mark": {
    "type": "arc",
    "stroke": "white",
    "thetaOffset":-0.39
  }
}
```
thetaOffset1

Its behaviour is analogous to specifying `xOffset` in a Cartesian vertical bar chart.

However, when the segment angle varies, such as a pie chart, to apply an offset we have to supply both `thetaOffset` and `theta2Offset`. 

```json
{
  "data": {
    "values": [
      {"category": "A","value": 4},
      {"category": "B","value": 6},
      {"category": "C","value": 10},
      {"category": "D","value": 3},
      {"category": "E","value": 7},
      {"category": "F","value": 8}
    ]
  },
  "encoding": {
    "theta": {"field": "value","type": "quantitative"},
    "color": {"field": "category","type": "nominal"}
  },
  "mark": {
    "type": "arc",
    "thetaOffset": -0.3,
    "theta2Offset": -0.3
  }
}
```

thetaOffset2

Needing to provide both seems inconsistent given the data-driven specification only identifies `theta` encoding. I think a more consistent approach would be apply `thetaOffset` to both the start and end angles unless (a) the main encoding uses both `theta` and `theta2` channels; or (b) both `thetaOffset` and `theta2Offset` are supplied. This would be more consistent with the Cartesian equivalents.

I note also that if only `thetaOffset` is applied (try removing the `theta2Offset` line above), it would appear that it is actually `theta2` that is offset :

thetaOffset2

And if only `theta2Offset` is supplied, we get a misleading overlapping of later segments on earlier ones (set the mark opacity to <1 to confirm this happens):

thetaOffset3

kanitw:
fix #6320

### DIFF
--- a/test/compile/mark/arc.test.ts
+++ b/test/compile/mark/arc.test.ts
@@ -41,6 +41,25 @@ describe('Mark: Arc', () => {
     });
   });
 
+  describe('for pie with thetaOffset', () => {
+    // This is a simplified example for stacked text.
+    // In reality this will be used as stacked's overlayed marker
+    const model = parseUnitModelWithScaleAndLayoutSize({
+      mark: {type: 'arc', thetaOffset: 5},
+      encoding: {
+        theta: {field: 'field', type: 'quantitative'},
+        color: {field: 'id', type: 'nominal'}
+      }
+    });
+
+    const props = arc.encodeEntry(model);
+
+    it('applies thetaOffset to both startAngle and endAngle', () => {
+      expect(props.startAngle).toEqual({scale: 'theta', field: 'field_end', offset: 5});
+      expect(props.endAngle).toEqual({scale: 'theta', field: 'field_start', offset: 5});
+    });
+  });
+
   describe('for radial histogram', () => {
     // This is a simplified example for stacked text.
     // In reality this will be used as stacked's overlayed marker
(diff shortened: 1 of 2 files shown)


test/compile/mark/arc.test.ts
```diff
@@ -41,6 +41,25 @@ describe('Mark: Arc', () => {
     });
   });
 
+  describe('for pie with thetaOffset', () => {
+    // This is a simplified example for stacked text.
+    // In reality this will be used as stacked's overlayed marker
```
domoritz, line 46:
```suggestion
    // In reality this will be used as stacked's overlayed marks.
```

PR #6324, opened 2020-04-12
fix: allow applying offset to binned rect/bar position

kanitw:


test/compile/mark/bar.test.ts
```diff
@@ -301,6 +301,35 @@ describe('Mark: Bar', () => {
...
+      expect(props.y2).toEqual([
+        fieldInvalidTestValueRef(y, 'y'),
+        {scale: 'y', field: 'bin_maxbins_10_Horsepower', offset: 5}
+      ]);
+      expect(props.y).toEqual([
+        fieldInvalidTestValueRef(y, 'y'),
+        {
+          scale: 'y',
+          field: 'bin_maxbins_10_Horsepower_end',
+          offset: defaultBarConfig.binSpacing + 5
```
donghaoren, line 326: Is `yOffset` supposed to only offset the `y` property? Do we also have `y2Offset`?

I'm think about the difference between the following scenarios:
- `yOffset` only on `y`, keep `y2`, change `height`
- `yOffset` on `y`, keep `height`, change `y2`

PR #6344, opened 2020-04-16
fix: don't generate label align when angle is 0 for top or bottom axis

### LINKED ISSUE #6341: Setting label angle removes label flush
Setting the label angle removes label flush. We could say that this is by design but it's a bit jarring. Not sure whether this is a regression either. 

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoAWOgCtY3AHaQANOCgATZAgBOjQnh4qckAIJhYjVITHIw7dLrBIneMAHdmiMGnbJNbU0wTXQ8ejUNSFDw7DBQCAhICnQxVh04gG0NRPjITFMzNSh2OIAmAA4AX1U8gqgAIWLIUpwAVjaaurjIAGFm1rBpAGYanMTgfJ6AEQG4gE4ARi7J+sgAUTmcCuXa1Z6AMS2wNtH1XIh90wBxY8X5lanTAAljioB2R7WASWO2sqq4wAuhoxkkMLoANY9Ry6SJJZBKeDcTSMJTUOIJRKQAAemMgADNGMgxJoYc08ABPQjIHoAR1Y6CURnCRiozXQOMYsHxYnQ-jEZnRdjiAAYqoCIIDAUA)

domoritz:
fixes #6341

### DIFF
--- a/src/compile/axis/assemble.ts
+++ b/src/compile/axis/assemble.ts
@@ -158,6 +158,10 @@ export function assembleAxis(
       setAxisEncode(axis, 'labels', 'text', {signal: expr});
     }
 
+    if (axis.labelAlign === null) {
+      delete axis.labelAlign;
+    }
+
     // Remove unnecessary encode block
     if (axis.encode) {
       for (const part of AXIS_PARTS) {
--- a/examples/compiled/bar.vg.json
+++ b/examples/compiled/bar.vg.json
@@ -93,7 +93,6 @@
       "grid": false,
       "title": "a",
       "labelAngle": 0,
-      "labelAlign": "center",
       "labelBaseline": "top",
       "labelOverlap": true,
       "zindex": 0
--- a/examples/compiled/bar_negative.vg.json
+++ b/examples/compiled/bar_negative.vg.json
@@ -106,7 +106,6 @@
       "labelAngle": 0,
       "labelPadding": 4,
       "ticks": false,
-      "labelAlign": "center",
       "labelBaseline": "top",
       "labelOverlap": true,
       "zindex": 0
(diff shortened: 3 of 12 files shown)


src/compile/axis/assemble.ts
```diff
@@ -158,6 +158,11 @@ export function assembleAxis(
       setAxisEncode(axis, 'labels', 'text', {signal: expr});
     }
 
+    // clear default label align so label flush still works
+    if (axis.labelAngle === 0 && (orient === 'top' || orient === 'bottom') && axis.labelAlign === 'center') {
+      delete axis.labelAlign;
```
kanitw, line 163: This is too hacky.  Setting it in parse and removing it later would make it really hard to reason about. 
Also, what if I make labelFlush = false, shouldn't we still set textAlign here?

, and wouldn't be correct in all cases. 

I'd rather make labelAlign/labelBaseline also depends on labelFlush. 
Basically if labelFlush is true, then don't set labelAlign for x / labelBaseline for y.

You can extract labelFlush to be AxisRuleParams, just like we do with labelAngle.

domoritz: I did that but when I returned `undefined` in the case where I didn't want to set it, then the resolution logic would fill in the wrong value. I could return `null` instead of undefined, I guess.

domoritz: The issue with `null` is that we need to clean it up later.

kanitw: I think cleaning up null is more reasonable than setting other values then delete?

At least, we can consistently set null for both signal and normal cases.  (And only clean up in normal cases.)

kanitw:
@domoritz -- can you do this by forking from https://github.com/vega/vega-lite/pull/6312?

PR #6346, opened 2020-04-17
docs: add liminoid.io to ecosystem list

hyphaebeast:


domoritz:
Thank you!

hyphaebeast:
🙏 I hope your CMU transition is going as well as it can during a global pandemic 🤞

domoritz:
Thank you. I'm a bit nervous about it but hopeful that things will be better by the fall.

PR #6358, opened 2020-04-19
refactor: use keyof typeof to reduce duplication

domoritz:
Merge after #6355
This pull request should reduce code duplication. I also replaced some strings with references to variables. More could be done but this is a start.


kanitw:
> Merge after #6357



#6357 is outdated and closed

domoritz:
Sorry, the right one is #6355

src/channel.ts
```diff
@@ -66,32 +66,36 @@ export const HREF: 'href' = 'href';
 
 export const URL: 'url' = 'url';
 
-export type PositionChannel = 'x' | 'y' | 'x2' | 'y2';
-
-const POSITION_CHANNEL_INDEX: Flag<PositionChannel> = {
+const POSITION_CHANNEL_INDEX = {
```
kanitw, line 69: FWIW, you declared 

```
const PositionScaleChannel = {
  x: 1,
  y: 1
} as const;
```

below, so perhaps we should be consistent.

src/compile/axis/properties.ts
```diff
@@ -2,7 +2,7 @@ import {Align, AxisOrient, Orient, ScaleType, SignalRef} from 'vega';
 import {isArray} from 'vega-util';
 import {Axis} from '../../axis';
 import {isBinned, isBinning} from '../../bin';
-import {PositionScaleChannel, X} from '../../channel';
+import {PositionScaleChannel, X, X2, Y, Y2} from '../../channel';
```
kanitw, line 5: Please either revert this file or rebase it to https://github.com/vega/vega-lite/pull/6312.

src/compile/facet.ts
```diff
@@ -245,7 +245,7 @@ export class FacetModel extends ModelWithField {
         as.push(`distinct_${field}`);
       }
     } else {
-      for (const channel of ['x', 'y'] as ScaleChannel[]) {
+      for (const channel of [X, Y] as ScaleChannel[]) {
```
kanitw, line 248: POSITION_SCALE_CHANNNELS?

src/compile/axis/properties.ts
kanitw: actually https://github.com/vega/vega-lite/pull/6372

src/channel.ts
domoritz: Let me fix the one below. I went back and forth between the two styles.

src/compile/axis/properties.ts
domoritz: Hopefully it's easy enough to resolve the conflicts here. I am happy to do the rebase if needed.

src/compile/facet.ts
domoritz: Done. I also replaced a few more places I found in the code.

domoritz:
I addressed your comments @kanitw

PR #6360, opened 2020-04-19
fix: always set axis/legend disable for correct merging

### LINKED ISSUE #6340: Conflicting axis property warning no longer raised and default behaviour has changed
Up to and including 4.10.1, Vega-lite was being charmingly forgiving of my poor spec, where I had different axis between layers and was mistakenly resolving as shared, giving me this warning and covering for my mistake by displaying the first axis.

```
WARN Conflicting axis property "disable" (false and true). Using false.
```

With 4.10.2 and 4.10.4, there is no warning and no axis is rendered, hence less forgiving.  Not a major issue but I thought worth raising in case unintended change or has broader implications. I see other issues related to axis disable, but believe this is a somewhat nuanced extension.

Simple test case follows...
```
{
  "data": {
    "values": [
      {
        "y": 1
      }
    ]
  },
  "layer": [
    {
      "mark": "circle",
      "encoding": {
        "y": {
          "field": "y",
          "type": "quantitative",
          "axis": {
            "title": "Y"
          }
        }
      }
    },
    {
      "mark": "square",
      "encoding": {
        "y": {
          "field": "y",
          "type": "quantitative",
          "axis": null
        }
      }
    }
  ],
  "resolve": {
    "scale": {
      "y": "shared"
    }
  }
}
```

kanitw:
fix #6340

### DIFF
--- a/src/compile/axis/parse.ts
+++ b/src/compile/axis/parse.ts
@@ -233,9 +233,9 @@ function parseAxis(channel: PositionScaleChannel, model: UnitModel): AxisCompone
 
   const axisConfigs = getAxisConfigs(channel, scaleType, orient, model.config);
 
-  if ((axis !== undefined && !axis) || getAxisConfig('disable', config, axis?.style, axisConfigs).configValue) {
-    // if axis is explicitly falsy, disable it and include no other properties.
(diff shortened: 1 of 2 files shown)


domoritz:
Screen Shot 2020-04-19 at 09 45 54

PR #6379, opened 2020-04-21
feat: support time units with day

domoritz:
merge after #6378


domoritz:
Yes, we were renaming time units because `day` and `datum` are not super obvious names.

PR #6386, opened 2020-04-21
feat: add option to allow custom formatter functions

domoritz:


domoritz:
@kanitw is implementing this differently

PR #6387, opened 2020-04-21
fix!: move customFormatTypes from compiler option to config

kanitw:


site/docs/config.md
```diff
@@ -50,7 +50,30 @@ A Vega-Lite `config` object can have the following top-level properties:
...
 These two config properties define the default number and time formats for text marks as well as axes, headers, and legends:
 
-{% include table.html props="numberFormat,timeFormat" source="Config" %}
+{% include table.html props="numberFormat,timeFormat,customFormatTypes" source="Config" %}
+
+{:#custom-format-type}
+
+### Providing Custom Formatters
+
+To customize how Vega-Lite formats numbers or text, you can register a new formatter by (1) setting the `customFormatTypes` config to `true` and then (2) registering [an expression function](https://vega.github.io/vega/docs/api/extensibility/#expressions) that takes a data point and a format parameter as input to the Vega View:
```
domoritz, line 59: The format parameter is optional, right? Maybe say that.

src/guide.ts
```diff
@@ -34,7 +34,7 @@ export interface FormatMixins {
   format?: string | object;
 
   /**
-   * The format type for labels (`"number"` or `"time"` or a [registered custom format type](https://vega.github.io/vega-lite/usage/compile.html#format-type)).
+   * The format type for labels (`"number"` or `"time"` or a [registered custom format type](https://vega.github.io/vega-lite/usage/config.html#custom-format-type)).
```
domoritz, line 37: Fix this above as well: `When used with a [custom `"formatType"`](https://vega.github.io/vega-lite/usage/compile.html#format-type), this value will be passed as `format` alongside `datum.value` to the registered function.`

PR #6391, opened 2020-04-22
feat: add aria properties to guides and add description channel, update to Vega 5.11

domoritz:
This pull request adds support for aria properties to Vega-Lite. Specifically, it
* hides grid axes by default
* adds new aria properties (aria and description) to marks and guides
* adds a `config.aria` property to remove default annotations
* updates the docs

merge after https://github.com/vega/vega-lite/pull/6409

~~Depends on https://github.com/vega/vega/pull/2549~~


kanitw:
Hmm having "chore: update for Vega 5.11" as chore commit will miss the fact that we add domain/grid/tickCap.

kanitw:
Should there be `config.aria` so one can turn it on/off for all of mark/axis/legend? 

Also, have we check if this is working well for facet headers?  (It's ok to deal with that in a separate PR though.)

domoritz:
> Should there be config.aria so one can turn it on/off for all of mark/axis/legend?

Wouldn't that be inconsistent with the Vega property `aria`? The Vega property sets `aria-hidden: true`, which is different from not setting aria properties. I'd mirror the Vega behavior instead.

domoritz:
Faceted charts work mostly but we can definitely improve it. I filed https://github.com/vega/vega-lite/issues/6443.

domoritz:
The part I'm least sure about is in https://github.com/vega/vega-lite/pull/6391/files?file-filters%5B%5D=.ts#diff-c051714e2f2736dc8791c76f8c39a6bcR9. @kanitw you can focus on that since there are a lot of changes in this pull request.

site/docs/encoding.md
```diff
@@ -329,13 +324,39 @@ The example below show how the href channel can be used to provide links to exte
...
-{% include table.html props="condition" source="ValueDefWithCondition<MarkPropFieldDef,(string|null)>" %}
+{% include table.html props="condition" source="ValueDefWithCondition<StringFieldDef,Text>" %}
+
+{:#description}
+
+## Description Channel
+
+By setting the `description` channel, you can add a text description to the mark for ARIA accessibility (SVG output only). The `"aria-label"` attribute will be set to this description.
+
+By default, Vega-Lite generatea a description based on the encoding similar to [default tooltips]({{ site.baseurl }}/docs/tooltip.html#encoding). To disable automatic descriptions, set [`config.aria`](config.html#aria-config) to false.
```
kanitw: We should also mention `mark.aria`?

src/compile/legend/assemble.ts
```diff
@@ -39,10 +39,20 @@ export function assembleLegends(model: Model): VgLegend[] {
     }
   }
 
-  return vals(legendByDomain)
+  const legends = vals(legendByDomain)
     .flat()
     .map(assembleLegend)
     .filter(l => l !== undefined);
+
+  if (model.config.aria === false) {
```
kanitw, line 47: In general, making side effect after a `map` should be avoided if possible.

Let's make `config` another parameter of `assembleLegend` instead.

src/compile/mark/encode/aria.ts
```diff
@@ -0,0 +1,79 @@
...
+    return {};
+  }
+
+  const ariaRoleDesc = getMarkPropOrConfig('ariaRoleDescription', markDef, config);
+
+  if (ariaRoleDesc != null) {
+    return {ariaRoleDescription: {value: ariaRoleDesc}};
+  }
+
+  return mark in VG_MARK_INDEX ? {} : {ariaRoleDescription: {value: mark}};
```
kanitw, line 40: bars and points, for example, are not VG marks.  Should they get no `ariaRoleDescription`?

site/docs/config.md
```diff
@@ -162,3 +163,9 @@ Each of the view composition configurations (`concat` and `facet`) supports the
 {% include table.html props="columns,spacing" source="CompositionConfig" %}
 
 Repeat uses the same configuration as concatenation.
+
+{:#aria-config}
+
+## ARIA Configuration
+
+`config.aria` is a boolean flag indicating if ARIA default attributes should be included for marks and guides (SVG output only). If false, the `"aria-hidden"` attribute will be set for all guides, removing them from the ARIA accessibility tree and Vega-Lite will not generate default descriptions for marks.
```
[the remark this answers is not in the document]
domoritz: Good point. Changed.

src/compile/mark/encode/aria.ts
domoritz: Vega marks don't need a default because Vega already has one.

PR #6399, opened 2020-04-25
refactor: remove utc format type since we don't support it

domoritz:


kanitw:
Vega actually does support it (see https://vega.github.io/vega/docs/axes/), so this PR is fixing the wrong thing. 
We should support it rather than not.

domoritz:
I think I disagree. In Vega-Lite, we rely on the scale being `utc`. I don't really want to change this right now.

kanitw:
> I think I disagree. In Vega-Lite, we rely on the scale being utc. I don't really want to change this right now.

Ok this is pretty confusing, but why should only only have to use UTC scale only in Vega-Lite, but not in Vega?  (Vega has both utc scale and utc format type as well.)

domoritz:
It is confusing but Vega-Lite is smart in that it automatically picks the right format type based on the scale type. I have not seen any problems so I don't want to change things.

kanitw:
ok whatever.

Issue #6382, opened 2020-04-21
Throw a warning if `customFormatTypes` is not enabled


kanitw:

domoritz:
The warning should appear when a custom format type is dropped. I think this would greatly improve debugging. Should we bump the priority?

kanitw:
Let's also drop it properly (https://github.com/vega/vega-lite/pull/6400#discussion_r415105388).

kanitw:
will be fixed in https://github.com/vega/vega-lite/pull/6448

PR #6400, opened 2020-04-25
fix: don't set guide label formats in encode blocks

### LINKED ISSUE #6398: Change how we generate formats to allow Vega to generate accessible labels
Vega's automatic description of guides does not take into account when the labels of an axis are formatted through an expression in the encode block. 

Unfortunately, Vega-Lite makes heavy use of it.

## Case 1

```json
{
  "mark": "line",
  "encoding": {
    "x": {"field": "date", "type": "temporal", "axis": {"format": "%a"}},
  }
}
```

generates

```json
    {
      "scale": "x",
      "orient": "bottom",
      "grid": false,
      "title": "date",
      "labelFlush": true,
      "labelOverlap": true,
      "tickCount": {"signal": "ceil(width/40)"},
      "encode": {
        "labels": {
          "update": {"text": {"signal": "timeFormat(datum.value, '%a')"}}
        }
      },
      "zindex": 0
    },
```

## Case 2

```json
{
  "mark": "bar",
  "encoding": {
    "x": {
      "timeUnit": "month",
      "field": "date",
      "type": "temporal"
    },
    "y": {
      "aggregate": "mean",
      "field": "precipitation",
      "type": "quantitative"
    }
  }
}
```

generates

```json
    {
      "scale": "x",
      "orient": "bottom",
      "grid": false,
      "title": "date (month)",
      "labelFlush": true,
      "labelOverlap": true,
      "encode": {
        "labels": {
          "update": {
            "text": {
              "signal": "timeFormat(datum.value, timeUnitSpecifier([\"month\"], {\"year-month\":\"%b %Y \",\"year-month-date\":\"%b %d, %Y \"}))"
            }
          }
        }
      },
      "zindex": 0
    },
```

In case 1, we could change Vega-Lite to use the standard format parameter of an axis but case 2 is a bit trickier. The issue is that Vega-Lite has a different default for time unit specifiers.

Maybe I'm not understanding time formatting correctly but I don't see a way to achieve what we need with just `format` and `formatType`. 

This, for example, isn't valid (and omits that we want to format a `month`).

Screen Shot 2020-04-24 at 09 39 17

Adding @haldenl who revised time unit support in Vega-Lite.

domoritz:
fixes #6398

~~Merge after #6391~~


src/compile/legend/assemble.ts
```diff
@@ -74,12 +73,7 @@ export function assembleLegend(legendCmpt: LegendComponent) {
   }
 
   if (labelExpr !== undefined) {
-    let expr = labelExpr;
-    if (legend.encode?.labels?.update?.text && isSignalRef(legend.encode.labels.update.text)) {
-      expr = replaceAll(labelExpr, 'datum.label', legend.encode.labels.update.text.signal);
-    }
-
-    setLegendEncode(legend, 'labels', 'text', {signal: expr});
+    setLegendEncode(legend, 'labels', 'text', {signal: labelExpr});
```
kanitw, line 76: Just like in axis, I don't think we should drop this.

src/compile/format.ts
```diff
@@ -1,114 +1,150 @@
...
+
+export function formatGuide(
+  fieldOrDatumDef: FieldDef<string> | DatumDef<string>,
+  type: Type,
+  format: string | object,
+  formatType: string,
+  config: Config,
+  omitTimeFormatConfig: boolean // axis doesn't use config.timeFormat
+) {
+  if (isFieldOrDatumDefWithCustomTimeFormat(fieldOrDatumDef, config)) {
```
kanitw, line 133: What if the spec has a  custom `formatType` but config doesn't allow it?

It seems like we're not handling this here?  

(If you agree this is wrong, we can do this in a separate PR since I think previously we didn't handle this 100% correct either.)

test/compile/axis/assemble.test.ts
```diff
@@ -51,23 +51,6 @@ describe('compile/axis/assemble', () => {
       expect(axis.title).toBe('Max of a, Min of b');
     });
 
-    it('correctly applies labelExpr', () => {
```
kanitw, line 54: Don't remove this.

domoritz:
> test is still failing

It depends on the new Vega release.

src/compile/axis/assemble.ts
```diff
@@ -151,12 +151,7 @@ export function assembleAxis(
     }
 
     if (labelExpr !== undefined) {
-      let expr = labelExpr;
-      if (axis.encode?.labels?.update && isSignalRef(axis.encode.labels.update.text)) {
-        expr = replaceAll(labelExpr, 'datum.label', axis.encode.labels.update.text.signal);
-      }
-
-      setAxisEncode(axis, 'labels', 'text', {signal: expr});
+      setAxisEncode(axis, 'labels', 'text', {signal: labelExpr});
```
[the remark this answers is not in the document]
domoritz: You can still use format but it's directly set to the axis now. I'm not following what the issue is here.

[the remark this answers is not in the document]
kanitw: Ok, datum.label shouldn't exist anymore so this is correct (no need to replace).

src/compile/legend/assemble.ts
kanitw: NVM, it's ok to remove this.

kanitw:
Some comments are still unaddressed.


Btw, I wonder which feature in the pending Vega this branch require?

kanitw:
I also don't think we should merge this in the aria branch since it's a fix that can be done before.  (We can merge into `vg5.11`)

domoritz:
> I also don't think we should merge this in the aria branch since it's a fix that can be done before. (We can merge into vg5.11)

I did this because it's a fix that is relevant to accessibility support. Without this change, guide VoiceOver descriptions are not formatted correctly. 

My vote would be for merging the ARIA changes before merging this change. Makes sense?

src/compile/format.ts
```diff
@@ -1,114 +1,150 @@
...
-      return {signal: `isValid(${field}) ? ${field} : ""+${field}`};
+      return vgField(fieldOrDatumDef, {expr});
     }
+  } else {
+    return datumDefToExpr(fieldOrDatumDef);
   }
-  return undefined;
+}
+
+export function formatCustomType({
```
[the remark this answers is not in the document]
domoritz: Fixed

src/compile/axis/parse.ts
```diff
@@ -305,10 +304,8 @@ function parseAxis(channel: PositionScaleChannel, model: UnitModel): AxisCompone
 
     const axisEncodingPart = guideEncodeEntry(axisEncoding[part] ?? {}, model);
 
-    const value = part === 'labels' ? encode.labels(model, channel, axisEncodingPart) : axisEncodingPart;
-
-    if (value !== undefined && keys(value).length > 0) {
-      e[part] = {update: value};
+    if (axisEncodingPart !== undefined && keys(axisEncodingPart).length > 0) {
```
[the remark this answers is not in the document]
domoritz: Right, this does not work anymore

```json
{
  "mark": "line",
  "encoding": {
    "x": {"field": "date", "type": "temporal", "axis": {"format": "%Y %b"}},
    "y": {"field": "price", "type": "quantitative", "axis": {"format": "0.2f", "formatType": "myFormat"}}
  },
  "config": {
    "customFormatTypes": true
  }
}
```

src/compile/format.ts
kanitw: Let's handle this in a separate PR for https://github.com/vega/vega-lite/issues/6382

domoritz:
Thank you @kanitw ❤️

PR #6402, opened 2020-04-25
feat: support setting scale range based on another field

### LINKED ISSUE #6392: Legend with disabled scale
I want precise control of the mapping of an encoding channel (e.g. color), by directly passing encoded values in the input data. Vega-Lite makes it possible to [disable the scale](https://vega.github.io/vega-lite/docs/scale.html#disable) which is exactly what I want.

So far so good, except that if I try that, [the legend disappears](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEC2CGBOBrSAuMwoBcCeAHApqlAMYCW8xANoWAL4A04U+AdsQPYAmpLA5kaAgRI2AUyFY8hNJBbtoPWJUiMJwgGal8lTkUgdl4uqqH72ldvDFrJBPXIUslKoxq069xFzcgBnYkrSYCwArpSURrRMDEyQnLCYsAKQAG5KIfi+RADaRsBmyjIAGipQhVAAgqX6egDE6uoADM2NkDES+QZ6AJrV5ZAAQtVeMrUtjQ1tALpRtEA):

```
{
  "mark": { "type": "circle" },
  "encoding": {
    "y": {
      "type": "nominal",
      "field": "col"
    },
    "color": {
      "type": "nominal",
      "field": "c",
      "scale": null
    }
  },
  "data": {"values": [
    {"col": "X", "l": "A", "c": "#ff0000"},
    {"col": "Y", "l": "B", "c": "#0000ff"}]}
}
```
visualization(66)

Okay, well, I guess that makes sense because otherwise the legend would just show the color values themselves, which is not particularly useful.

But that's precisely the problem: it would make sense to be able to customize which *field* is used for the legend labels, so that I can use some field (here, `c`) for the color but some other field (here, `l`) for the legend labels. That is, I want this:

visualization(67)

Which [can be made to work](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEC2CGBOBrSAuMwoBcCeAHApqlAMYCW8xANoWAL4A04U+AdsQPYAmpLA5kaAgRI2AUyFY8hNJBbtoPWJUiMJwgGal8lTkUgdl4uqqH72ldvDFrJBPXIUslKoxq069ykxMgBnYkrS6FDwsHxBANqQAMTq6gAMifEqUNFJ8XGQALp0RrRMDEyQnLCYsAKQAG5KAK74vkQRRsBmyjIAGimQbVAAgl3EerEJSZCFEi0GegCaXT2QAEIDQ+mZtFn5tEA), but only by manually spelling out all the color values in the color channel scale range, and then praying they don't get out of sync with the data itself.

Ideally, what I'm looking for is a way to (1) set a legend where the legend labels come from a specific field (or Vega expression), and (2) have that legend show up even if the color scale is `null`. Something like this for example:

```
    "color": {
      "type": "nominal",
      "field": "c",
      "scale": null,
      "legend": { "values": "datum.l" }
    }
```

kanitw:
fix #6392


kanitw:
@domoritz  schema doesn't get pushed for the failing diff. Did you break in the script recently?

domoritz:
Looks like there are still unstaged changes. Let's output what it is. 

Screen Shot 2020-04-25 at 12 46 31

PR #6405, opened 2020-04-26
chore: fix comment about tuple store types

domoritz:


src/compile/selection/transforms/project.ts
```diff
@@ -11,7 +11,7 @@ export const TUPLE_FIELDS = '_tuple_fields';
 
 /**
  * Do the selection tuples hold enumerated or ranged values for a field?
- * Ranged values can be left-right inclusive (R) or left-inclusive, right-exclusive (R-LE).
+ * Ranged values can be exclusive (E),  left-right inclusive (R) or left-inclusive, right-exclusive (R-RE).
```
kanitw, line 14: > Ranged values can be exclusive (E)

- E = enumerate according to the line above
- Enumerate is not range

domoritz: So 
* E = enumerated
* R = ranged but left-right left-right inclusive
* R-RE = ranged left-inclusive, right-inclusive

kanitw: I think so, but @arvind would know better.

domoritz:
Merging. @arvind please let us know if we misunderstood the comment.

PR #6408, opened 2020-04-26
fix: various small code fixes

domoritz:
Discovered in https://lgtm.com/projects/g/vega/vega-lite


kanitw:
Hmm this is actually a refactor.

domoritz:
The merged legend change is not, I think. But I’d have to look more closely to confirm that it could have caused problems.

PR #6426, opened 2020-04-30
chore: include PNGs in the compiled files

kanitw:


kanitw:
Let's do https://github.com/vega/vega-lite/pull/6446 instead.

kanitw:
Given we can't really run example compilation consistenly across platform, I think it's better to setup png in `build:examples` process _for now_. 

We could consider migrating to using jest for generating both SVGs and PNGs later.

PR #6438, opened 2020-05-02
fix: revise `encoding.color = {signal:...}` to be `{value: {signal:...}}`

### LINKED ISSUE #6433: Revise syntax for using signal in an encoding constant `value`.
Seeing  datum signal in https://github.com/vega/vega-lite/pull/6432 as   `encoding.color = {datum: {signal: ...}`

I think we should revise `encoding.color = {signal:...}` to be `encoding.color = {value: {signal:...}}` so it's consistent and easier to teach, though a bit more verbose.   

Basically, once we have signal as an official feature, we can just say `value` is a constant value in visual domain and `datum` is constant value in data domain, both of which support signals as their values.  (Instead of saying value signal use `{signal:...}` but datum signal needs nesting.)

kanitw:
fix #6433

Note that we'll still support conditional axis signal like this:

```
gridColor: {
  condition: {
    test: '<is major grid>',
    signal: 'majorGridColor' // this can also be `value: "#ddd"`
  },
  signal: 'gridColor'  // this can also be `value: "#eee"`
}
```

### DIFF
--- a/src/encoding.ts
+++ b/src/encoding.ts
@@ -1,4 +1,4 @@
-import {AggregateOp, SignalRef} from 'vega';
+import {AggregateOp} from 'vega';
 import {array, isArray} from 'vega-util';
 import {isArgmaxDef, isArgminDef} from './aggregate';
 import {isBinned, isBinning} from './bin';
@@ -46,7 +46,7 @@ import {
   ChannelDef,
   ColorGradientDatumDefWithCondition,
   ColorGradientFieldDefWithCondition,
-  ColorGradientValueOrSignalDefWithCondition,
+  ColorGradientValueDefWithCondition,
   DatumDef,
   Field,
   FieldDef,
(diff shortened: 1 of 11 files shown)


src/encoding.ts
```diff
@@ -188,7 +188,7 @@ export interface Encoding<F extends Field> {
   color?:
     | ColorGradientFieldDefWithCondition<F>
     | ColorGradientDatumDefWithCondition<F>
-    | ColorGradientValueOrSignalDefWithCondition<F>;
```
kanitw, line 191: I didn't like the fact that we have to do `ColorGradientValueOrSignalDefWithCondition` because once we expose signal as an official feature, value and signal will show up together in the docs.  

If want to keep signal def but separate value and signal in the property table, then we have to declare so many repetitive types.
For every variant of value or signal def (e.g., `ColorGradient`), we have to do:

```
XXXValueDefWithCondition
XXXSignalDefWithCondition
```

domoritz:
Nice. I like that you delete more code than add.

PR #6446, opened 2020-05-03
test: include snapshot tests

kanitw:


domoritz:
Tests are failing. Do you see the failures locally as well.

kanitw:
So it seems like node-canvas doesn't generate images with the same size on Linux (compared to my mac output).  

I think I'll switch to https://github.com/vega/vega-lite/pull/6426.

PR #6448, opened 2020-05-04
fix: drop custom format type if disabled + throw warning

### LINKED ISSUE #6382: Throw a warning if `customFormatTypes` is not enabled

kanitw:
Fix #6382


kanitw:
> Looks good. Just want to confirm that we no longer support formatTypes in the options of compile?

Yep.

Issue #6464, opened 2020-05-07
unselectedOpacity should be a legend property, not just a legend config


domoritz:

kanitw:
Why?

kanitw:
(FWIW, I think unselectedOpacity is kinda different from other legend property, so I'm not sure if I like it.)

domoritz:
What if I want different opacities for different legends?

kanitw:
Then you can override? Just like you can override any config?

domoritz:
That's the point. I cannot override the opacity for one specific legend only. 

<img width="1199" alt="Screen Shot 2020-05-10 at 10 02 27" src="https://user-images.githubusercontent.com/589034/81505604-5e71d900-92a5-11ea-8861-dc02c967df2c.png">

kanitw:
Ok got you know.  This is a dupilcate of https://github.com/vega/vega-lite/issues/5921, but I think the description in #5921 is better so I'll close this one.

PR #6465, opened 2020-05-07
docs: add missing legend docs, move properties to where they belong

domoritz:


domoritz:
`strokeDash` and `strokeWidth` are the Vega properties used to map a scale to the stroke property. They are not normal legend properties so they don't make sense in Vega-Lite. See https://vega.github.io/vega/docs/legends/.

site/docs/encoding/legend.md
```diff
@@ -62,23 +62,23 @@ _See also:_ This [interactive article](https://beta.observablehq.com/@jheer/a-gu
 
 ### General
 
-{% include table.html props="aria,cornerRadius,description,direction,fillColor,legendX,legendY,offset,orient,padding,strokeColor,strokeWidth,type,tickCount,unselectedOpacity,values,zindex" source="Legend" %}
```
domoritz, line 65: See https://github.com/vega/vega-lite/issues/6464 about unselectedOpacity.

Issue #5992, opened 2020-02-29
Including a field in a tooltip breaks aggregating that field


jowens:
This is the standard cars example but adding a `color` encoding (`Origin`). I wanted to aggregate (mean) on the `y` encoding (`Miles_per_Gallon`). If I include this encoding in the tooltip, the aggregation disappears. The below two pictures show the difference if I change the tooltip between `Miles_per_Gallon` to `Miles_per_Gallo`.

I think my desirable behavior would be if I include a field that is aggregated, the tooltip would show `mean(field)` instead of `field`.

Hopefully I didn't miss an existing issue that looks at this (I didn't see anything in the vl docs either).

https://gist.github.com/jowens/da4845936b35ee7ade554ca93e6741ab , with first few lines below.

```json
{
  "config": {"view": {"continuousWidth": 400, "continuousHeight": 300}},
  "data": {"name": "data-f02450ab61490a1363517a0190416235"},
  "mark": "point",
  "encoding": {
    "color": {"type": "nominal", "field": "Origin"},
    "x": {"type": "temporal", "field": "Year"},
    "y": {
      "type": "quantitative",
      "aggregate": "mean",
      "field": "Miles_per_Gallon"
    },
    "tooltip": [
      {"type": "quantitative", "field": "MxNxK"},
      {"type": "nominal", "field": "metric"},
      {"type": "quantitative", "field": "Miles_per_Gallon"},
      {"type": "nominal", "field": "schedule"}
    ]
  },
  "$schema": "https://vega.github.io/schema/vega-lite/v4.0.2.json",
  "datasets": {
    "data-f02450ab61490a1363517a0190416235": [
```

<img width="1148" alt="Vega_Editor" src="https://user-images.githubusercontent.com/976970/75614689-39682980-5af0-11ea-801e-efadeab52042.png">
<img width="1120" alt="Vega_Editor" src="https://user-images.githubusercontent.com/976970/75614692-3ff6a100-5af0-11ea-9436-6565e3bd9b5e.png">

jakevdp:
This is working as intended, I believe: the way you specify that you'd like to group by a particular field before aggregating is to put that field in the encoding. When you put ``Miles_per_gallon`` in the tooltip encoding, you are saying that you want to group by this field.

[...]

PR #6475, opened 2020-05-10
fix: support fields with spaces in them

domoritz:
See https://github.com/vega/vega-lite/issues/5992#issuecomment-626370269


jowens:
Hoping this also applies to fields with `-` in them, a problem which bit me today.

domoritz:
It should, yes.

PR #6476, opened 2020-05-10
docs: motivate argmin/argmax

domoritz:
Suggested by @jowens.

### DIFF
--- a/site/docs/transform/aggregate.md
+++ b/site/docs/transform/aggregate.md
@@ -125,6 +125,8 @@ The supported **aggregation operations** are:
 
 ## Argmin / Argmax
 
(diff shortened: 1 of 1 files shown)


site/docs/transform/aggregate.md
```diff
@@ -125,6 +125,8 @@ The supported **aggregation operations** are:
 
 ## Argmin / Argmax
 
+Sometimes, you may not want to find the minimum or maximum of a field but the value from a field corresponds to the minimum or maximum value in another fields. In these cases you can use the argmin and argmax aggregates.
```
jowens, line 128: "Sometimes, you may not want to find the minimum or maximum of a field, but instead the value from a field that corresponds to the minimum or maximum value in another field." ?

domoritz: Thanks

PR #6500, opened 2020-05-16
chore: update to vega 5.12

haldenl:


domoritz:
Let us know when this pull request is ready. Make sure to check [the changelog](https://github.com/vega/vega/releases/tag/v5.12.0) for any changes that need to be propagated to Vega-Lite.

kanitw:
There are so many examples diff, so we need to carefully examine what's happening.

domoritz:
A lot of it is a change to svg styles. We can look at the pngs to see what the visible changes are.

haldenl:
Looks like the svg diff is caused by a vega bug that results in fill being misattributed. I get these when I import the example svgs into a `.sketch` file. Curiously the black background does not appear on github.

image

https://github.com/vega/vega/issues/2618

jheer:
This looks like potentially a different issue from the Vega bug, which led to style leakage to other SVGs. Here it looks like Sketch is not respecting the document level SVG stylesheet, such that default fills are being applied where fill none in the stylesheet should apply?

haldenl:
gotcha, that would make sense. I'm not seeing this in any other application.

jheer:
Hmm, looks like I'm seeing it in Chrome as well if I load standalone SVGs generated by vg2svg. That is not good. Will investigate further and fix.

jheer:
Just released fixes as part of Vega v5.12.1. The internal stylesheet for Vega SVG uses a custom class for the Vega root element to localize the CSS rules, and also moves the `style` tag into a top level `defs` tag. I don't know if this fixes the issue you are seeing in Sketch, but I can confirm that Chrome properly renders standalone SVG files produced by Vega.

haldenl:
Thanks @jheer! I'm still seeing fill issue issue in sketch but the other problems have been resolved.

package.json
```diff
@@ -126,12 +126,12 @@
...
     "tslib": "~1.11.2",
     "vega-event-selector": "~2.0.3",
-    "vega-expression": "~2.6.4",
-    "vega-util": "~1.13.2",
+    "vega-expression": "~2.6.5",
+    "vega-util": "~1.14.0",
     "yargs": "~15.3.1"
   },
   "peerDependencies": {
-    "vega": "^5.11.1"
```
domoritz, line 134: Why update the Vega dependency? Do we need something from Vega 5.12?

PR #6501, opened 2020-05-16
fix: prefer the axis with temporal type to be the non-aggregated axis in composite marks

### LINKED ISSUE #6491: Errorband doesn't work on X axis
[Works](https://vega.github.io/editor/#/examples/vega-lite/layer_line_errorband_ci)
[Doesn't work](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JACyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICUJNohSZ8gL5K4sqG2U1ZZNKACebkADMacQcr6AJpwpoogmO44cPqYiDhs2kh6Spg0CHAAqrJ0+u6h2iBWNiCCSPmFqADaoMjaANZekdH6cNraiUxIsoG2AB5xspj6UDRFtvaOzq6aIH1evv6B6ACyNIJwEAD60dpbAOLJgnLhzTHoAI4M3WnqaaSndBv6K6GyAARs3u9rGxDvu3eh0Exw+AAoAJwAVgApO8AMIASQgAEoisUFLVTI10MJZDEJg4nC4vPNZkgyGRtHw4voMt1wosAi91psdm0DkcTqkoucQFcbnQ1DQHsUrABdKxAA)

chanwutk:
fix #6491


src/compositemark/common.ts
```diff
@@ -281,12 +282,26 @@ export function compositeMarkOrient<M extends CompositeMark>(
...
+          // x is temporal
+          if (isFieldOrDatumDefForTimeFormat(y) && isMarkDef(mark) && mark.orient) {
+            // both x and y are temporal and orient is specified
+            return mark.orient;
+          }
 
-        // default orientation = vertical
-        return 'vertical';
+          // x is temporal but y is not
+          return 'vertical';
```
domoritz, line 293: If orient is specified, shouldn't we use it?

chanwutk: If both x and y are temporal we should follow the orient. But, if only x is temporal, shouldn't the orient be 'vertical'?

kanitw:
Minor comment: Can you make PR names follow conventional commits as well, just in case we squash?

src/compositemark/common.ts
domoritz: Well, if the user overrides it, we should respect it or remove the orient but also show a warning. 

Let's write some tests for the different cases to make sure they work.

chanwutk: Should we make the user specified orient be the first priority, or just in the temporal case?

domoritz: It should always be the priority but we should remove it and warn the user if we cannot satisfy their request.

PR #6502, opened 2020-05-16
fix: make size in errorbar also set the length of its ticks

### LINKED ISSUE #6495: Setting size of errorbar removes the ticks
I am trying to make errorbars a bit thicker. Using ```size```in ```encoding``` works, but removes the ticks. The json below (modified from https://vega.github.io/vega-lite/docs/errorbar.html) gives no ticks as expeced when testing in https://vega.github.io/editor/

```json
{
  "layer": [
    {
      "mark": {"type": "errorbar", "ticks": true},
      "encoding": {
        "x": {
          "field": "yield",
          "type": "quantitative",
          "scale": {"zero": false}
        },
        "y": {"field": "variety", "type": "ordinal"},
        "size": {"value": "2"}
      }
    },
    {
      "mark": {"type": "point", "filled": true, "color": "black"},
      "encoding": {
        "x": {"field": "yield", "type": "quantitative", "aggregate": "mean"},
      }
    }
  ]
}
```

chanwutk:
Fix #6495

`size` in errorbar's `encoding` becomes `thickness` in mark definition of `ticks`.

### DIFF
--- a/src/compositemark/errorbar.ts
+++ b/src/compositemark/errorbar.ts
@@ -69,7 +69,10 @@ export interface ErrorExtraEncoding<F extends Field> {
   yError2?: SecondaryFieldDef<F> | ValueDef<number>;
 }
 
-export type ErrorEncoding<F extends Field> = Pick<Encoding<F>, PositionChannel | 'color' | 'detail' | 'opacity'> &
+export type ErrorEncoding<F extends Field> = Pick<
+  Encoding<F>,
+  PositionChannel | 'color' | 'detail' | 'opacity' | 'size'
+> &
   ErrorExtraEncoding<F>;
 
 export type ErrorBarPartsMixins = PartsMixins<ErrorBarPart>;
@@ -131,15 +134,20 @@ export function normalizeErrorBar(
     tooltipEncoding
   } = errorBarParams(spec, ERRORBAR, config);
 
+  const {size, ...encodingWithoutContinuousAxisAndSize} = encodingWithoutContinuousAxis;
   const makeErrorBarPart = makeCompositeAggregatePartFactory<ErrorBarPartsMixins>(
     markDef,
     continuousAxis,
     continuousAxisChannelDef,
-    encodingWithoutContinuousAxis,
+    encodingWithoutContinuousAxisAndSize,
     config.errorbar
   );
 
-  const tick: MarkDef = {type: 'tick', orient: ticksOrient};
+  const tick: MarkDef = {
+    type: 'tick',
+    orient: ticksOrient,
+    ...(size && 'value' in size ? {thickness: size.value} : {})
+  };
 
   const layer = [
     ...makeErrorBarPart({
@@ -164,7 +172,10 @@ export function normalizeErrorBar(
       },
       positionPrefix: 'lower',
       endPositionPrefix: 'upper',
-      extraEncoding: tooltipEncoding
+      extraEncoding: {
+        ...tooltipEncoding,
+        ...(size ? size : {})
+      }
     })
   ];
 
--- a/test/compositemark/errorbar.test.ts
+++ b/test/compositemark/errorbar.test.ts
@@ -2,7 +2,7 @@ import {AggregateOp} from 'vega';
 import {ErrorBarCenter, ErrorBarExtent} from '../../src/compositemark/errorbar';
 import {defaultConfig} from '../../src/config';
 import * as log from '../../src/log';
-import {isMarkDef} from '../../src/mark';
+import {isMarkDef, MarkDef} from '../../src/mark';
 import {normalize} from '../../src/normalize';
 import {isLayerSpec, isUnitSpec} from '../../src/spec';
 import {TopLevelUnitSpec} from '../../src/spec/unit';
@@ -539,6 +539,36 @@ describe('normalizeErrorBar with raw data input', () => {
       {field: 'center_people', title: 'Median of people', type: 'quantitative'},
       {field: 'age', type: 'ordinal'}
     ]);
+  });
+
+  it("should move size (value def) in tick's encoing channel to its mark definition as thickness", () => {
+    const thickness = 5;
+    const outputSpec = normalize(
+      {
+        data: {url: 'data/population.json'},
+        mark: {type: 'errorbar', ticks: true},
+        encoding: {
+          x: {field: 'age', type: 'ordinal'},
+          y: {field: 'people', type: 'quantitative'},
+          size: {value: thickness}
+        }
+      },
+      defaultConfig
+    );
+
+    expect(isLayerSpec(outputSpec)).toBeTruthy();
+    if (isLayerSpec(outputSpec)) {
+      outputSpec.layer.forEach(l => {
+        if (isUnitSpec(l) && (l.mark as MarkDef).type === 'tick') {
+          const {mark} = l;
+          if (isMarkDef(mark) && mark.type === 'tick') {
+            expect(mark.thickness).toBe(thickness);
+          }
+        }
+      });
+    } else {
+      expect(false).toBe(true);
+    }
   });
 });
 


src/compositemark/errorbar.ts
```diff
@@ -69,7 +69,10 @@ export interface ErrorExtraEncoding<F extends Field> {
   yError2?: SecondaryFieldDef<F> | ValueDef<number>;
 }
 
-export type ErrorEncoding<F extends Field> = Pick<Encoding<F>, PositionChannel | 'color' | 'detail' | 'opacity'> &
```
chanwutk, line 72: @kanitw @domoritz I'm not sure why `size` is not in the `ErrorEncoding`. Is it not supposed to be in the encoding channel of error bar?

src/compositemark/errorbar.ts
```diff
@@ -131,15 +134,20 @@ export function normalizeErrorBar(
...
-    encodingWithoutContinuousAxis,
+    encodingWithoutContinuousAxisAndSize,
     config.errorbar
   );
 
-  const tick: MarkDef = {type: 'tick', orient: ticksOrient};
+  const tick: MarkDef = {
+    type: 'tick',
+    orient: ticksOrient,
+    ...(size && 'value' in size ? {thickness: size.value} : {})
```
chanwutk, line 149: @domoritz @kanitw I'm not sure what is the good way for getting the size's value.

src/compositemark/errorbar.ts
kanitw: Isn't `ErrorEncoding` shared between ErrorBar and ErrorBand?

src/compositemark/errorbar.ts
kanitw: This is fine, but you can use isValueDef type guard or something?

kanitw: Provided that you wanna keep this (see general comment)

Also, this seems like markDef.size won't be applied correctly?

domoritz:
> I think mapping size to error bar's thickness is confusing. If you ask me in English, what's the size of the error bar, I would think that you mean length (which encodes the uncertainty), not thickness.

> Another alternative to consider is to just ignore size (drop it), but instead introduce a thickness property for just ErrorBar's markdef and apply it to both the underlying rule and ticks.

I think I disagree. The length of the errorbar represents the spread of the data. The size channel to me should map to the thickness of the lines. I don't want to add another encoding channel when we already have one that is sensible. We already use size for example to represent the thickness of bars or trails.

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JACyUAVhDYA7EABoQAEzjQATjRyZ289AEEABBBoIcguIaZJ1h2DcyGA7nRiHETOMtXLDypJhUiioBKKigxEiCDGpoANqgYSD6wUxoAEwAHAC+ColoIABCqWgArKW5+egAwiWoYgDMlSBJACJ1AJwAjM1JAKJ1mT15LQUAYnWlTSNJAOJ1XR29BQASgwDsy+gAkpPp2QC6uSDI6gDWBdbqwXCyUGzKNLJkaKAAHq8gAGY0cILKBRQSkwAE8cHACmx1I9ZFFgkg3jQIJ9BEhPIJ9M8LGgAAzZY4gz4-P4A9BpYFgiHoACODCQsh0gR0pBAxxMAC8qaBif9LsFQeCCrT6YyAjQWfjskA)
Add https://vega.github.io/editor/#/examples/vega-lite/trail_color

kanitw:
>  I don't want to add another encoding channel when we already have one that is sensible.

I'm not suggesting an encoding channel.  This should be just the mark property, and we already have `thickness` property for ticks. 

> We already use size for example to represent the thickness of bars or trails.

Note that the size of ticks are their width, not thickness.  And ticks are parts of error bars. 


Another problem with `size` keyword is that `size` is an encoding channel.  But we can't map tick's thickness to a field (yet).  So we can't really support size encoding channel for errorbar. 
But it would be surprising that `size` works in both mark property and encoding elsewhere, but not for error bars. 


If we just call this `thickness`, then it's at least consistent with `ticks` and doesn't add confusion related to `size`'s encoding channel.

domoritz:
I see. So you are saying let's not support size as an encoding channel and instead support thickness as a mark property? 

I'm okay with that as well. But I feel supporting size as an encoding channel and mapping it to thickness of the lines makes sense as well.

chanwutk:
We can already set thickness and size of ticks [here](https://vega.github.io/editor/?#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JACyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICTJLrgBPKTPkBfJXFlQ2ymrLJpQ5jyABmNOILK+sSm-pheSuE4cPps2i6ySHo2diCCSOZw2mgA2qDI2gDW3lEx6DhsrpiKvjSCgnBBqJjaDHCpDk4J7pogAB7eSGRk2nyYZSAIcEjySn4BTSDm-oE1pfoAjgwzmHRqNKQ10ElloABeWWxoPkkQ7ZF0DfoAQqYN5gAEAJorQalOgji3hCgja+iY6SgxRSdnypmKvXW6Cy2jiJmy9j641k1XQEEwyjgxDWNChEBKMFJhVkcAg5NQAEYlBAaBc0AAmAAMKXsjmcrh6oAGvXmq3QywWa3M0U22xxe12hwemCe6FeZk+P0lMIAujYgA).
But, one problem with setting `thickness` and `size` in `ticks` is that the `thickness` and `size` will get overridden by `size` in `encoding` [here](https://vega.github.io/editor/?#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JACyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICTJLrgBPKTPkBfJXFlQ2ymrLJpQ5jyABmNOILK+sSm-pheSuE4cPps2i6ySHo2diCCSOZw2mgA2qDI2gDW3lEx6DhsrpiKvjSCgnBBqJjaDHCpDk4J7pogAB7eSGRk2nyYZSAIcEjySn4BTSDm-oE1pfoAjgwzmHRqNKQ10ElloABeWWxoPkkQ7ZF0DfoAQqYN5gAEAJorQalOgji3hCgja+iY6SgxRSdnypmKvXW6Cy2jiJmy9j641k1XQEEwyjgxDWNChEBKMFJhVkcAg5NQAEYlBAaBc0AAmAAMKXsjmcrh6oAGvXmq3QywWa3M0U22xxe12hwemCe6FeZk+P0lqRZbN6ILBqAAzDCALo2IA).

kanitw:
1)  To clarify, let's do *not* add size to encoding for error bars. As said you shouldn't add encoding size if it doesn't work with `field`.

2) Then the question is what do we add to mark definition. 
I still think that thickness is better. 

2.1) If we add `mark.size`, in 

```
   {
      "mark": {
        "type": "errorbar",
	    "size": {"value": 3}
        "ticks": {"thickness": 1, "size": 20}
      },
      "encoding": {
        ...
      }
    }
```

`mark.size` will override `mark.ticks.thickness`, not `mark.ticks.size`.

2.2) While a similar challenge happens with `rule`, namely that in

```
   {
      "mark": {
        "type": "errorbar",
	    "thickness": {"value": 3}
        "rule": {"size": 1}
      },
      "encoding": {
        ...
      }
    }
```
`mark.thickness` will override `mark.rule.size`.   

That being said, we would agree in English that rule's size is its thickness (whereas in 2.1, tick's size is by definition not its thickness).

kanitw:
Given this conclusion, I think it's better to close this PR and start a new one to add the `thickness`.  Feel free to reopen if you disagree.

kanitw:
Given this conclusion, I think it's better to close this PR and start a new one to add the `thickness` and always drop `size` encoding  from error bar.  Feel free to reopen if you disagree.

chanwutk:
@kanitw Shouldn't `mark.size` override `mark.ticks.size`, and `mark.thickness` override `mark.ticks.thickness` and `mark.rule.size`?

chanwutk:
I wonder if it should be the other way around?
`mark.ticks.size` override `mark.size`
`mark.ticks.thickness` and `mark.rule.size` override `mark.thickness`

I also think that we might not need `mark.size`. `mark.size` only have an effect on ticks. And for `mark.size` to have an effect, we have to also set `mark.ticks` to `true` first. Would it make more sense to not have `mark.size` at all, and set ticks' size from `mark.ticks.size`?

PR #6503, opened 2020-05-17
docs: explain how to achieve toggling without shift-key

morberg:
This is an attempt to explain how to achieve toggling without requiring the user to hold down the shift-key.

See the discussion in https://github.com/altair-viz/altair/issues/2151. I read and re-read this part of the documentation but couldn’t for the life of me figure out how to enable toggling without using a modifier key.


arvind:
Looks good to me, thank you @morberg!

PR #6518, opened 2020-05-20
chore: upgrade to ts 3.9

domoritz:


src/vega.schema.ts
```diff
@@ -141,25 +141,24 @@ export type VgScale = Pick<BaseScale, 'type'> & {
...
+      Omit<QuantileScale, 'type'> &
+      Omit<QuantizeScale, 'type'> &
+      Omit<ThresholdScale, 'type'> &
+      Omit<BinOrdinalScale, 'type'> &
+      // Sequential
+      Omit<SequentialScale, 'type'> &
+      // Discrete
+      Omit<BandScale, 'type'> &
+      Omit<PointScale, 'type'> &
+      Omit<OrdinalScale, 'type'>,
```
donghaoren, line 161: Why `omit` type individually given that we already omit in the outer `Omit`?

src/compile/data/parse.ts
```diff
@@ -119,7 +119,7 @@ function parseRoot(model: Model, sources: SourceNode[]): DataFlowNode {
 
     if (model.data === null) {
       // data: null means we should ignore the parent's data so we just create a new data source
-      const source = new SourceNode([]);
+      const source = new SourceNode({values: []});
```
donghaoren, line 122: I don't see any change to `SourceNode` constructor, is this going to break?

src/vega.schema.ts
domoritz: Good catch. The previous code was wrong because the type can never be satisfied if you don't omit the `type`. Typescript fixed a bug that previously existed.

src/compile/data/parse.ts
domoritz: `[]` is a not a valid `Data` (the expected type). I don't know why the code accepted an empty array before.

kanitw:
I don't know what's `unknown` yet, but I assume they are reasonable.

src/data.ts
```diff
@@ -83,7 +84,7 @@ export type DataSource = UrlData | InlineData | NamedData;
 
 export type Data = DataSource | Generator;
 
-export type InlineDataset = number[] | string[] | boolean[] | object[] | string | object;
+export type InlineDataset = number[] | string[] | boolean[] | Dict<unknown>[] | string | Dict<unknown>;
```
danmarshall, line 87: the `object[]` syntax was working ok, now I need to `extend Dict<unknown>` ...

domoritz: Can you expand on this comment? I don’t understand the context.

danmarshall: I ended up doing an inline cast to `Dist<any>` [here](https://github.com/microsoft/vegalite-for-powerbi/blob/93aebf4ec2f360f800d828e4a4432101fa040d9f/src/visual.ts#L98). Let's see if any other users start seeing compiler issues with the removal of `object`.

PR #6551, opened 2020-05-28
docs: use modern js in example

domoritz:


kanitw:
Some people may still prefer JS that doesn't need compiler?

domoritz:
The js in the example doesn't use anything that most modern browsers not already support.

PR #6564, opened 2020-05-31
docs: document values aggregate

domoritz:
See https://github.com/vega/vega/pull/2674

### DIFF
--- a/site/docs/transform/aggregate.md
+++ b/site/docs/transform/aggregate.md
@@ -98,8 +98,9 @@ The supported **aggregation operations** are:
 
 | Operation | Description |
 | :-- | :-- |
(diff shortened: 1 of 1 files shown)


site/docs/transform/aggregate.md
```diff
@@ -98,8 +98,9 @@ The supported **aggregation operations** are:
 
 | Operation | Description |
 | :-- | :-- |
-| count | The total count of data objects in the group. <span class="note-line">**Note:** _'count'_ operates directly on the input objects and return the same value regardless of the provided field. Similar to SQL's `count(*)`, count can be specified with a `field` `"*"`. |
+| count | The total count of data objects in the group. <span class="note-line">**Note:** _'count'_ operates directly on the input objects and return the same value regardless of the provided field. |
```
donghaoren, line 101: Should we still use `field: "*"`?

domoritz: I don't think so.

kanitw: no, it's not necessary.

PR #6588, opened 2020-06-05
fix: revert to object for inlineDataset element

danmarshall:
The use of `Dict<unknown>` as an InlineDataset element would not allow the use of an interface to represent an array element. Reverting to `object`.

Note: I'm not able to run `yarn test` locally - is this because I'm using Windows?

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -10174,7 +10174,7 @@
         },
         {
           "items": {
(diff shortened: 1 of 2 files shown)


domoritz:
> Note: I'm not able to run yarn test locally - is this because I'm using Windows?

We don't develop on windows but things should work (I think) in WSL.

PR #6608, opened 2020-06-14
fix: tell npx where to find vega scripts

domoritz:


domoritz:
cc @johnmdonich

domoritz:
@johnmdonich checked this change.

Issue #6606, opened 2020-06-13
Layered Chart: Binning a calculated field causes "Infinite extent for field"


shabbirahussain:
# Steps to reproduce:

1. Add a transformed field with transform_calculate (a simple `new_x = 'datum.x'` will do). 
2. Create a layered chart with one layer using the newly transformed field.
3. Thing should look good so far.
4. Now use bins on the newly transformed field. This will cause "Infinite extent for field".

Note: Using bins on original `x` field works just fine.

```json
{
  "config": {"view": {"continuousWidth": 400, "continuousHeight": 300}},
  "layer": [
    {
      "mark": "rect",
      "encoding": {
        "color": {"type": "quantitative", "aggregate": "count"},
        "x": {"type": "quantitative", "bin": true, "field": "x"},
        "y": {"type": "quantitative", "field": "Horsepower"}
      },
      "transform": [{"calculate": "datum.Displacement", "as": "x"}]
    },
    {
      "mark": "circle",
      "encoding": {
        "x": {"type": "quantitative", "field": "Displacement"},
        "y": {"type": "quantitative", "field": "Horsepower"}
      }
    }
  ],
  "data": {"url": "https://vega.github.io/vega-datasets/data/cars.json"},
  "$schema": "https://vega.github.io/schema/vega-lite/v4.8.1.json"
}
```

https://vega.github.io/editor/#/gist/cfdf00ecbbed9187a32f5488ffe356fd/spec.json

domoritz:
Note that this is only a warning and can be ignored. I wonder what exactly is causing it, though. 

Also related https://github.com/vega/vega-lite/issues/6266

shabbirahussain:
I think I know the issue. This is being caused during vega-light -> vega schema translation. 

The transformation order being generated have an 'extent' and 'bin' transformations (which uses the newly calculated field) before the 'calculate' transformation which actually generates that field.

Current Vega schema being generated [Not-Working]: https://vega.github.io/editor/#/gist/7c5350f9a6da919ceee71ecb8eb59eaf/spec.json

Expected [Working]: https://vega.github.io/editor/#/gist/392d8d0d20e8bb738664812ac508f7f7/spec.json

domoritz:
Right. The bin is supposed to come after the calculate.

https://github.com/vega/vega-lite/blob/master/src/compile/data/parse.ts#L261

But this is the dataflow we are getting here.  

[...]

Issue #2270, opened 2017-04-23
Correctly merge extent / bin node for shared bin scale


kanitw:

kanitw:
Since in #2415, we can move repeat and concat with shared axes to Post 2.0.  
This means that we can punt on fixing this issue, which is quite complicated, until post 2.0 too.  

If you disagree, feel free to fix it, but I don't plan to fix it within this summer.

domoritz:
Attempt in https://github.com/vega/vega-lite/pull/6625

shabbirahussain:
Pre-binning data might be the only viable option here. One or the other case would fail.

PR #6625, opened 2020-06-16
fix: add bin after transforms

domoritz:
Fixes https://github.com/vega/vega-lite/issues/6606

See #2270


domoritz:
Surprisingly, only a few examples are broken here. Somehow we are not using the same scale/bins for x and y. 

Screen Shot 2020-06-15 at 17 32 43

domoritz:
Which is probably correct since the filter is before the bin. 

Screen Shot 2020-06-15 at 17 38 35

So we get this dataflow. 

Screen Shot 2020-06-15 at 17 38 08

So, now the question is, do we want to
* move the bin before the filter in layered charts
* add syntax to move transforms post transforms coming from the encoding
* add filter to encodings (probably not)
* change the example to bin explicitly before the filter

domoritz:
Note that this change also breaks cross filter. 

Screen Shot 2020-06-15 at 17 43 10

Closing for now since I don't have time to fix this.

Issue #5764, opened 2020-01-20
Order channel does not support conditional


domoritz:
The order channel cannot be conditional but we use it in an example. This looks like a regression in Vega-Lite 4. It breaks the example tests in master. 

[...]

PR #6632, opened 2020-06-17
feat: support conditional value def for order

### LINKED ISSUE #4864: Make `order` use Vega's `zindex` rather than sort
- This allows more efficient re-drawing. 

- This will make order support conditional encoding (https://github.com/vega/vega-lite/issues/5764)

----- 

Open Question:

- What does order = array of field defs mean for zindex?
- What does value def order mean for non z-index order? (e.g., stacking.)

kanitw:
fix #4864 
fix https://github.com/vega/vega-lite/issues/5764

Basically we now map order's `field` to `sort` but order's conditional value defs to `zindex`.  

To avoid unnecessary complexity, my current implementation does not allow mixing field in one condition but value in another. 
I have looked at existing examples but I can't see when we would really need to mix field and value for order.

That said, if we have convincing use cases, we can extend `zindex` support condition of a field (but not an array of field would be too complicated).


examples/specs/line_color_interactive.vl.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
+  "description": "Stock prices of 5 Tech Companies over Time.",
+  "data": {"url": "data/stocks.csv"},
+  "interactive": {"highlight": true, "select": true},
```
domoritz, line 5: I don't think you want this in this changeset, right?

PR #6638, opened 2020-06-18
feat: make `type` optional

kanitw:
Merge #6648 first 

This will immensely improve usability!

The new type docs as copied below should summarize the change/suggested use case. (4.14.0 is the next version which will include this PR.) 

Feel free to suggest additions to the default rules. 

cc: @jakevdp  -- I know you asked for this a while ago.  :)
 
-----

Since Vega-Lite 4.14.0, Vega-Lite automatically infers data types in many cases as discussed below. However, type is required for a field if: 
(1) the field is not nominal and the field encoding has no specified `aggregate` (except `argmin` and `argmax`), `bin`, scale type, custom `sort` order, nor `timeUnit`
(2) if you wish to use an ordinal scale for a field with `bin` or `timeUnit`.

__Default value:__
1) For a data `field`, `"nominal"` is the default data type unless the field encoding has `aggregate`, `channel`, `bin`, scale type, `sort`, or `timeUnit` that satisfies the following criteria:
- `"quantitative"` is the default type if (1) the encoded field contains `bin` or `aggregate` except `"argmin"` and `"argmax"`, (2) the encoding channel is `latitude` or `longitude` channel or (3) if the specified scale type is [a quantitative scale](https://vega.github.io/vega-lite/docs/scale.html#type).
- `"temporal"` is the default type if (1) the encoded field contains `timeUnit` or (2) the specified scale type is a time or utc scale
- `ordinal""` is the default type if (1) the encoded field contains a [custom `sort` order](https://vega.github.io/vega-lite/docs/sort.html#specifying-custom-sort-order) or (2) the specified scale type is an ordinal/point/band scale.
2) For a constant value in data domain (`datum`):
- `"quantitative"` if the datum is a number
- `"nominal"` if the datum is a string
- `"temporal"` if the datum is [a date time object](https://vega.github.io/vega-lite/docs/datetime.html)


domoritz:
- [ ] Fix examples/compiled/area.png

src/channeldef.ts
```diff
@@ -1072,14 +1116,12 @@ export function initFieldDef(fd: FieldDef<string, any>, channel: Channel) {
...
     const newType = defaultType(fieldDef as TypedFieldDef<any>, channel);
-    log.warn(log.message.missingFieldType(channel, newType));
-
     fieldDef['type'] = newType;
   }
 
   if (isTypedFieldDef(fieldDef)) {
-    const {compatible, warning} = channelCompatibility(fieldDef, channel);
-    if (!compatible) {
+    const {compatible, warning} = channelCompatibility(fieldDef, channel) || {};
```
domoritz, line 1123:
```suggestion
    const {compatible, warning} = channelCompatibility(fieldDef, channel) ?? {};
```

kanitw: Meh, It doesn't matter here and `||` leads to better ES5 output.

PR #6639, opened 2020-06-18
docs: use nominal for bar charts

domoritz:
I did not look at all examples but some.


kanitw:
Hmm it's changing label overlap.

domoritz:
Ahh, see, another thing we do. We don't want to remove labels when we use nominal but we do it for ordinal.

kanitw:
lol I also just notice fruits in linux

image

PR #6643, opened 2020-06-18
fix: remove transformextract test 

kanitw:
We need to keep extractTransformForEncoding, but that shouldn't benefit from these tests.


domoritz:
I suspect that we don't need to keep all of the code then, either.

kanitw:
> I suspect that we don't need to keep all of the code then, either.

I already remove the unused part of the code. See the diff. 

The transformextract pretty much does nothing besides calling extractTransformFromEncoding.

saulshanabrook:
Just FYI I was using this downstream (https://github.com/Quansight/ibis-vega-transform/blob/fba79d4437df5d481603609a51c5f990b6a90ad2/src/compiler.ts#L22) and the 4.14.0 broke that with these changes broke that.

I will copy the file locally into my project then, to keep using the functionality.

PR #6648, opened 2020-06-20
fix: apply labelOverlap to ordinal time unit without custom sort, not any ordinal

kanitw:
I don't think turning on labelOverlap for ordinal is good in general.

It makes sense for timeUnit ordinal because single time unit months/days/weeks are mostly uniform, so we can skip some (even though not are months have same number of days)
However, if it's something non-uniform, it's probably a bad idea.

For example, we have been using Cylinders as ordinal
But it skips some numbers (like 7) so hiding some labels hide the facts that some numbers are not a valid domain value / not imputed.

Similarly for grades,  suppose there are A+, A, A-, B+, B, B-, C+, C, C-, D, F.  (D and F has no plus minus) (edited)
We can't reliably skip just some labels.


domoritz:
I still think as a default it’s okay to skip ordinal since you know that the missing label is something between the surrounding ones. I also haven’t seen it happening too often that labels get skipped.

kanitw:
Our principle for sensible defaults so far has been conservative, trying to do smart things only if it works for most of the time. 

Skipping ordinal is probably more on the controversial side and is conflicting with our principle. 

Plus, if you think it doesn't happen that often, the ordinal bit isn't very useful.  

In any case, I'll merge this for now since (1) I improve the logic to be more correct in general for timeUnit and when there is a sort property, (2) there are a number of depending PRs (including https://github.com/vega/vega-lite/pull/6638, which has a cleaner diff if we remove labelOverlap for ordinal).

Feel free to send a follow up PR after we merge https://github.com/vega/vega-lite/pull/6638 if you strongly think we need to keep this behavior.

PR #6654, opened 2020-06-21
docs: boxplot with pre-calculated summaries

### LINKED ISSUE #4343: Support pre-aggregated box plot
(or at least create an example that shows an equivalent data)

kanitw:
fix #4343

FYI, the data in this file isn't the real penguin data but I'm putting my guesses since the data is not available in the editor yet.  We may want to update the inline data toe match the dataset.


kanitw:
Updated with real data ;)

PR #6674, opened 2020-06-23
fix: don't add guide properties to tooltips, use encoding titles for boxplot titles

domoritz:


src/compositemark/common.ts
```diff
@@ -58,7 +58,7 @@ export function filterTooltipWithAggregatedField<F extends Field>(
 } {
   const {tooltip, ...filteredEncoding} = oldEncoding;
   if (!tooltip) {
-    return {filteredEncoding: oldEncoding};
+    return {filteredEncoding};
```
kanitw, line 61: Why this this change related to "fix: don't add guide properties to tooltips"?

src/compositemark/common.ts
```diff
@@ -123,11 +123,16 @@ export function getCompositeMarkTooltip(
     }
   );
 
+  const tooltipFieldDefs = fieldDefs(encodingWithoutContinuousAxis).map(fd =>
+    // remove guides as they don't work for tooltips
+    omit(fd, ['legend', 'axis', 'header'] as any[])
```
kanitw, line 128: What about scale?  

I also look at `tooltip.ts`, it seems like our tooltip logic reads `title` for its key.  So simply drop guides here is also incorrect.

src/compositemark/common.ts
domoritz: It's not. It just makes the code less confusing.

src/compositemark/common.ts
domoritz: Oh, we read the title off the guide? We should move the title to the encoding property then.

kanitw: Yeah, I'd say look at tooltip.ts carefully and make sure we keep all important bits.

src/channeldef.ts
```diff
@@ -885,8 +897,8 @@ export function title(
   config: Config,
   {allowDisabling, includeDefault = true}: {allowDisabling: boolean; includeDefault?: boolean}
 ) {
-  const guide = getGuide(fieldOrDatumDef) ?? {};
-  const guideTitle = guide.title;
+  const guide = getGuide(fieldOrDatumDef);
```
kanitw, line 900: We can get rid of guide variable?

src/compositemark/boxplot.ts
```diff
@@ -291,7 +291,7 @@ export function normalizeBoxPlot(
           type: continuousAxisChannelDef.type,
           ...(title !== undefined ? {title} : {}),
           ...(scale !== undefined ? {scale} : {}),
-          ...(axis !== undefined ? {axis} : {})
+          ...(isEmpty(axisWithoutTitle) ? {} : {axis: axisWithoutTitle})
```
kanitw, line 294: Add comment why we need `axisWithoutTitle` here.  (It's unclear to me.)

src/compositemark/common.ts
```diff
@@ -171,7 +175,7 @@ export function makeCompositeAggregatePartFactory<P extends PartsMixins<any>>(
           type: continuousAxisChannelDef.type,
           ...(title !== undefined ? {title} : {}),
           ...(scale !== undefined ? {scale} : {}),
-          ...(axis !== undefined ? {axis} : {})
+          ...(isEmpty(axisWithoutTitle) ? {} : {axis: axisWithoutTitle})
```
kanitw, line 178: Add comment why we need axisWithoutTitle here. (It's unclear to me.)

PR #6682, opened 2020-06-25
feat: allow shared type, scale, axis, legend in shared encoding

### LINKED ISSUE #6675: A more readable way to express shared scale and guide properties
Currently, our syntax for scales and guide prioritizes single views and thus include `scale`, `axis`, and `legend` inside a single view's encoding, even when we want to express a shared scale/guide.

This design allows users to rapidly compose unit charts with scale/guide definitions, which is quite useful for rapid visual/exploratory data analysis.

However, when we use Vega-Lite in production code, this design can lead to a code that is hard to read and maintain, esp. for advanced charts with several layers.

For example, if we have a spec where we define _n_ unit views as layer1-5, in which each layer (e.g., layer1) can have sub-layers, so this spec isn't necessary very short:

```
var layer1 = <unit/layer spec A>;
var layer2 = <unit/layer spec B>;
...
var layer5 = <unit/layer spec E>;


var spec = {
  layer: [layer1, ..., layer5]
  // more readable if we can express shared scale/axis/legend here 
}
```

It would be better to express a shared scale/axis definition in the shared `spec` rather than hiding it in one of the unit view, which sometimes confuses our users (including myself) where we specify scale and axis. 

To help users write better code, we should provide an alternative syntax that better matches the shared structure.  (This is another way of doing things, but the benefit outweighs the redundancy.) 

There are several examples that can benefit from this: 

- https://vega.:.io/vega-lite/examples/bar_layered_weather.html
- https://vega.github.io/vega-lite/examples/facet_bullet.html
- https://vega.github.io/vega-lite/examples/wheat_wages.html
- https://vega.github.io/vega-lite/examples/concat_layer_voyager_result.html
- https://vega.github.io/vega-lite/examples/layer_likert.html

Here are a few alternatives:

## A) Allow encoding with just scale/axis/legend (without field/datum/value) in layer encoding

For example, we could modify https://vega.github.io/vega-lite/examples/wheat_wages.html to be:

```
...,
encoding: {
  x: {"axis": {"tickCount": 5, "format": "d"}}, 
  y: {"axis": { "zindex": 1 }}
},
layer: [{
  ...
}],
...
```

Given that we already have a shared encoding with exactly this schema (except that field is required), I prefer this solution.

## Solution B) Allow Resolve to specify Shared Properties

Resolve could support scale/axis/legend definitions, in addition to `"shared"|"independent":

```
...,
resolve: {
  axis: {
    x: {"tickCount": 5, "format": "d"},
    y: {"zindex": 1}
  }
},
layer: [{
  ...
}],
...
```

kanitw:
fix #6675

I've checked all the SVG/VG diff.


examples/compiled/bar_layered_weather.vg.json
```diff
@@ -140,7 +140,7 @@
           "fill": {"value": "#ccc"},
           "ariaRoleDescription": {"value": "bar"},
           "description": {
-            "signal": "\"id: \" + (isValid(datum[\"id\"]) ? datum[\"id\"] : \"\"+datum[\"id\"]) + \"; Temperature (F): \" + (format(datum[\"record.low\"], \"\")) + \"; record.high: \" + (format(datum[\"record.high\"], \"\"))"
+            "signal": "\"id: \" + (isValid(datum[\"id\"]) ? datum[\"id\"] : \"\"+datum[\"id\"]) + \"; record.low: \" + (format(datum[\"record.low\"], \"\")) + \"; record.high: \" + (format(datum[\"record.high\"], \"\"))"
```
kanitw, line 143: This is more accurate.

examples/compiled/area_horizon.vg.json
```diff
@@ -83,7 +83,7 @@
           "fill": {"value": "#4c78a8"},
           "opacity": {"value": 0.3},
           "description": {
-            "signal": "\"x: \" + (format(datum[\"x\"], \"\")) + \"; y: \" + (format(datum[\"ny\"], \"\"))"
+            "signal": "\"x: \" + (format(datum[\"x\"], \"\")) + \"; ny: \" + (format(datum[\"ny\"], \"\"))"
```
kanitw, line 86: This is more accurate.

examples/compiled/interactive_line_hover.vg.json
```diff
@@ -11,7 +11,7 @@
       "name": "hover_store",
       "values": [
         {
-          "unit": "layer_0_layer_0",
+          "unit": "layer_0",
```
kanitw, line 14: This is changing because I change unnecessary nesting in the VL spec.

examples/compiled/layer_candlestick.vg.json
```diff
@@ -79,7 +79,7 @@
           ],
           "ariaRoleDescription": {"value": "bar"},
           "description": {
-            "signal": "\"Date in 2009: \" + (timeFormat(datum[\"date\"], '%m/%d')) + \"; open: \" + (format(datum[\"open\"], \"\")) + \"; close: \" + (format(datum[\"close\"], \"\"))"
+            "signal": "\"Date in 2009: \" + (timeFormat(datum[\"date\"], '%m/%d')) + \"; Price: \" + (format(datum[\"open\"], \"\")) + \"; close: \" + (format(datum[\"close\"], \"\"))"
```
kanitw, line 82: This is more accurate.

examples/compiled/layer_line_co2_concentration.vg.json
```diff
@@ -97,7 +97,7 @@
             "update": {
               "stroke": {"scale": "color", "field": "decade"},
               "description": {
-                "signal": "\"decade: \" + (isValid(datum[\"decade\"]) ? datum[\"decade\"] : \"\"+datum[\"decade\"]) + \"; Year into Decade: \" + (format(datum[\"scaled_date\"], \"\")) + \"; CO2 concentration in ppm: \" + (format(datum[\"CO2\"], \"\"))"
+                "signal": "\"scaled_date: \" + (format(datum[\"scaled_date\"], \"\")) + \"; CO2 concentration in ppm: \" + (format(datum[\"CO2\"], \"\")) + \"; decade: \" + (isValid(datum[\"decade\"]) ? datum[\"decade\"] : \"\"+datum[\"decade\"])"
```
domoritz, line 100: It used to say "Year into Decade". Is this description still there somewhere?

examples/compiled/layer_line_errorband_2d_horizontal_borders_strokedash.vg.json
```diff
@@ -54,7 +54,7 @@
...
           },
           "description": {
-            "signal": "\"Miles per Gallon (95% CIs): \" + (format(datum[\"lower_Miles_per_Gallon\"], \"\")) + \"; upper_Miles_per_Gallon: \" + (format(datum[\"upper_Miles_per_Gallon\"], \"\")) + \"; Year (year): \" + (timeFormat(datum[\"year_Year\"], '%b %d, %Y')) + \"; Ci1 of Miles_per_Gallon: \" + (format(datum[\"upper_Miles_per_Gallon\"], \"\")) + \"; Ci0 of Miles_per_Gallon: \" + (format(datum[\"lower_Miles_per_Gallon\"], \"\")) + \"; Mean of Miles_per_Gallon: \" + (format(datum[\"center_Miles_per_Gallon\"], \"\"))"
+            "signal": "\"Miles_per_Gallon: \" + (format(datum[\"lower_Miles_per_Gallon\"], \"\")) + \"; upper_Miles_per_Gallon: \" + (format(datum[\"upper_Miles_per_Gallon\"], \"\")) + \"; Year (year): \" + (timeFormat(datum[\"year_Year\"], '%b %d, %Y')) + \"; Ci1 of Miles_per_Gallon: \" + (format(datum[\"upper_Miles_per_Gallon\"], \"\")) + \"; Ci0 of Miles_per_Gallon: \" + (format(datum[\"lower_Miles_per_Gallon\"], \"\")) + \"; Mean of Miles_per_Gallon: \" + (format(datum[\"center_Miles_per_Gallon\"], \"\"))"
```
domoritz, line 57: The description used to be "Miles per Gallon (95% CIs)", which was better, no?

kanitw: No, because value described here is the lower value, so it's weird if lower description is 95% (without the word lower) and the upper is just upper.  Thus, I move "Miles per Gallon (95% CIs)" to be just axis title.

Note that somehow the default for lower doesn't say "lower", unlike upper. 
This is a bug in error bar code, but isn't caused by this PR.

examples/compiled/layer_line_co2_concentration.vg.json
kanitw: In the axis title, but I can't just make it field title because the "min" scaled_date. Let me hide aria for the text labels, then we can bring this back.

examples/compiled/layer_line_errorband_2d_horizontal_borders_strokedash.vg.json
domoritz: Sounds good.

examples/compiled/layer_line_co2_concentration.vg.json
domoritz: ok

PR #6684, opened 2020-06-25
refactor: make DataSourceType an Enum

### LINKED ISSUE #6609: Make DataSourceType an enum

znation:
Use an enum with string values, instead of using an union type
with string literals. This should give better compile time type checking
and error messages.

Fixes #6609

### DIFF
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -470,13 +470,7 @@ export abstract class Model {
   }
 
   public getDataName(type: DataSourceType) {
-    switch (type) {
-      case DataSourceType.Raw: return this.getName("raw");
-      case DataSourceType.Main: return this.getName("main");
-      case DataSourceType.Row: return this.getName("row");
-      case DataSourceType.Column: return this.getName("column");
-      case DataSourceType.Lookup: return this.getName("lookup");
-    }
+    return DataSourceType[type];
   }
 
   /**
--- a/src/compile/model.ts
+++ b/src/compile/model.ts
@@ -469,13 +469,23 @@ export abstract class Model {
     return varName((this.name ? this.name + '_' : '') + text);
   }
 
+  public getDataName(type: DataSourceType) {
+    switch (type) {
+      case DataSourceType.Raw: return this.getName("raw");
+      case DataSourceType.Main: return this.getName("main");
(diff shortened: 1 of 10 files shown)


src/data.ts
```diff
@@ -147,10 +147,16 @@ export function isGraticuleGenerator(data: Partial<Data> | Partial<VgData>): dat
...
-export const RAW = 'raw' as const;
+export enum DataSourceType {
+  Raw = 'raw',
+  Main = 'main',
+  Row = 'row',
+  Column = 'column',
+  Lookup = 'lookup',
+};
+
+export const MAIN = DataSourceType.Main as const;
```
znation, line 158: Now I'm wondering if we should stop exporting MAIN and RAW separately, since the whole enum now has a type name to refer to. So consumers of it should probably use `DataSourceType.Raw` instead of `RAW`. I'll make that change too.

src/data.ts
```diff
@@ -147,10 +147,13 @@ export function isGraticuleGenerator(data: Partial<Data> | Partial<VgData>): dat
...
-
-export const MAIN = 'main' as const;
-export const RAW = 'raw' as const;
+export enum DataSourceType {
+  Raw = 'raw',
+  Main = 'main',
+  Row = 'row',
+  Column = 'column',
+  Lookup = 'lookup'
+}
```
domoritz, line 156: I don't think we even need the names here anymore. Let's just let ts figure out the enum values (probably numbers).

src/compile/model.ts
```diff
@@ -469,13 +469,23 @@ export abstract class Model {
     return varName((this.name ? this.name + '_' : '') + text);
   }
 
+  public getDataName(type: DataSourceType) {
+    switch (type) {
```
domoritz, line 473: You can lookup the name on the enum, I think. Try `return DataSourceType [type]`.

znation: Thanks! I'm coming back to TypeScript after about 6 years away. I'm a little rusty still.

domoritz: It has changed a lot.

domoritz:
Moved to https://github.com/vega/vega-lite/pull/6686

PR #6686, opened 2020-06-25
refactor: make DataSourceType an Enum

### LINKED ISSUE #6609: Make DataSourceType an enum

domoritz:
Use an enum with string values, instead of using an union type
with string literals. This should give better compile time type checking
and error messages.

Fixes #6609

Replaces https://github.com/vega/vega-lite/pull/6684/


examples/compiled/area_density_stacked.vg.json
```diff
@@ -44,7 +44,7 @@
       "type": "group",
       "from": {
         "facet": {
-          "name": "faceted_path_main",
+          "name": "faceted_path_Main",
```
kanitw, line 47: Better keep the lowercase

znation: Yup, I am realizing that is the cause of most of my test failures. Will keep lowercase.

domoritz:
Screen Shot 2020-06-25 at 17 24 08

znation:
@domoritz I see code coverage checks are failing, but I did add a unit test for all of the new code added (as far as I can tell). Not sure why it's reporting that coverage is still too low. Otherwise, I think this PR is ready to go in. Should we override the coverage requirement here, or is there any other test coverage I should add?

domoritz:
Since the code that isn't covered isn't new, we can accept your patch. I'll take a look if you think it's ready.

znation:
@domoritz Yes, I think it's ready. Please take a look, thanks!

domoritz:
Thank you

PR #6687, opened 2020-06-26
refactor: use correct error logger

domoritz:


src/compile/data/optimize.ts
```diff
@@ -15,7 +15,7 @@ export function checkLinks(nodes: readonly DataFlowNode[]): boolean {
   for (const node of nodes) {
     for (const child of node.children) {
       if (child.parent !== node) {
-        console.error('Dataflow graph is inconsistent.', node, child);
+        log.error('Dataflow graph is inconsistent.', node, child);
```
kanitw, line 18: When do we throw error and when do we use log.error?

domoritz: In this case we should not crash because maybe the chart still works. But it's pretty bad if we have an inconsistent dataflow.

PR #6688, opened 2020-06-26
feat: support cursor configuration for interval selection

donghaoren:
This is a step towards #4155.


src/compile/selection/interval.ts
```diff
@@ -94,6 +110,31 @@ const interval: SelectionCompiler<'interval'> = {
...
+    const cursorEvents = [
+      ...(selCmpt.viewCursor
+        ? [
+            {events: 'scope:mouseover', update: "'crosshair'"},
+            {events: 'scope:mouseout', update: JSON.stringify(selCmpt.viewCursor)}
+          ]
+        : []),
+      ...(selCmpt.mark.cursor
+        ? [
+            {events: `@${selCmpt.name + BRUSH}:mouseover`, update: JSON.stringify(selCmpt.mark.cursor)},
```
domoritz, line 124: This will use `"` while the code below uses `'` quotes. Why don't we just use `value` or a string template?

domoritz:
This should be redone.

PR #6689, opened 2020-06-26
fix: do not apply axis/legend title to tooltip title

kanitw:
Rationale:  if one sets `axis.title = ""` to hide the title, then tooltip should still display the field name

If one really wants the field title to be "", they should set fielddef.title directly.

Overall, this reduces unnecessary complexity in our code. 

This reverts part of https://github.com/vega/vega-lite/pull/6674.


domoritz:
Makes sense. But we still use the channel def title.

kanitw:
I've checked to make sure that SVG diffs are improvement to descriptions.  Previously, so many descriptions use empty strings as the keys.

domoritz:
Awesome. Thanks for fixing it.

PR #6693, opened 2020-06-26
refactor: simplify description signals

kanitw:
The old signal was unnecessarily complicated


kanitw:
@domoritz I looked at a few diff and it seems like the example SVGs are outdated.

PR #6696, opened 2020-06-28
fix: boxplot should include description on the box layer, not mid tick

kanitw:


kanitw:
This is pretty straightforward, so I'll merge this first as there are dependent PRs.  

Feel free to send follow up comments / fixes.

PR #6707, opened 2020-06-28
refactor: eliminate warnings in tests & eliminate warnings in tests

kanitw:


src/compile/data/optimize.ts
```diff
@@ -15,7 +15,7 @@ export function checkLinks(nodes: readonly DataFlowNode[]): boolean {
   for (const node of nodes) {
     for (const child of node.children) {
       if (child.parent !== node) {
-        log.error('Dataflow graph is inconsistent.', node, child);
+        // log.error('Dataflow graph is inconsistent.', node, child);
```
domoritz, line 18: I really need this.

kanitw: You can turn it back on when you debug.  There is no point showing this to users since it's unclear what they should do when they see this message.

PR #6710, opened 2020-07-03
ci: use pat to build automated commits

domoritz:
With this change, we are now building even automated commits so that we don't end up in a situation where we miss that a test failed but not in the last commit.


kanitw:
Thanks, it's hard to look at the code and know if it's doing the right thing.  

Let's merge this and revert if it is broken.

PR #6712, opened 2020-07-04
refactor: use stricter channel types

domoritz:
Fixes https://github.com/vega/vega-lite/issues/5173

### DIFF
--- a/src/encoding.ts
+++ b/src/encoding.ts
@@ -4,7 +4,6 @@ import {isArgmaxDef, isArgminDef} from './aggregate';
 import {isBinned, isBinning} from './bin';
 import {
   ANGLE,
-  Channel,
   CHANNELS,
   COLOR,
   DESCRIPTION,
@@ -40,7 +39,8 @@ import {
   X,
   X2,
   Y,
-  Y2
+  Y2,
+  Channel
 } from './channel';
 import {
   binRequiresRange,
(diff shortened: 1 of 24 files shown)


src/scale.ts
```diff
@@ -854,32 +852,3 @@ export function channelSupportScaleType(channel: Channel, scaleType: ScaleType):
       return scaleType === 'ordinal'; // shape = lookup only
   }
 }
-
-export function getSupportedScaleType(channel: Channel, fieldDefType: Type) {
```
domoritz, line 858: this code is unused

test/scale.test.ts
```diff
@@ -96,57 +88,4 @@ describe('scale', () => {
       }
     });
   });
-
```
domoritz, line 99: remove test for unused code

src/encoding.ts
```diff
@@ -429,15 +432,16 @@ export function extractTransformsFromEncoding(oldEncoding: Encoding<any>, config
             }
           }
         }
+
         // now the field should refer to post-transformed field instead
-        encoding[channel] = newFieldDef;
+        encoding[channel as any] = newFieldDef;
```
domoritz, line 437: need this to resolve: Expression produces a union type that is too complex to represent.ts(2590)

src/encoding.ts
```diff
@@ -576,9 +582,9 @@ export function fieldDefs<F extends Field>(encoding: EncodingWithFacet<F>): Fiel
   return arr;
 }
 
-export function forEach<U extends Partial<Record<Channel, any>>>(
+export function forEach<U extends Record<any, any>>(
```
Saba9, line 585: Why go from the more strict type of `Partial<Record<Channel, any>>>` to the less strict `Record<any, any>`?

domoritz: I don't know exactly why but we actually get better type inference where we use `forEach`.

Issue #6719, opened 2020-07-07
legend's symbolOpacity does not work


swanderz:
### background

I can't pass a different opacity to the legend symbols than what is given to all the marks. I want the legend symbols to have 100% opacity. On [SO](https://stackoverflow.com/questions/62768337/set-opacity-for-marks-but-not-in-legend),  @jakevdp pointed me here!

### reprex ([Vega Editor link](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygK4CcA28QAspQA4Gc4B6IgNwFMBzCAOkoEspNEAjG+gezKogFpIY+clHxEBEIgGMIyfDQBW+DgDsQAXwA0IALYyA1vFBQAnrnIZJ9ZJNTmtHXBEsn4ABhoBGTSHLLJHMHplSkMQf1QOZFCAM3pyVDAMAHlkegZVLRMzDGUObSCIdC1bSl9EhBB8Y20WDlQkx2djeC9vAA8YuISMAAlIoVwOAHdyKMzTczgQAEdECGUoRmh6CnUtZorY+PKQAFl6W3wAfTNkI4BxQoiMkCzJmbmFpcXVtTegA))
```json
{
  "data": {"url": "https://vega.github.io/vega-datasets/data/cars.json"},
  "mark": {"type": "circle", "opacity": 0.1},
  "encoding": {
    "color": {
      "field": "Origin",
      "type": "nominal",
      "legend": {"symbolOpacity": 1}
    },
    "x": {"field": "Horsepower", "type": "quantitative"},
    "y": {"field": "Miles_per_Gallon", "type": "quantitative"}
  }
}
```

domoritz:
`symbolStrokeColor` also doesn't work as expected. 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
  "description": "A bubbleplot showing horsepower on x, miles per gallons on y, and binned acceleration on size.",
  "data": {"url": "data/cars.json"},
  "mark": "point",
  "encoding": {
    "x": {"field": "Horsepower", "type": "quantitative"},
    "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
    "size": {"field": "Acceleration", "type": "quantitative", "legend": {
      "symbolStrokeColor": "red"
    }}
  }
}

```

domoritz:
Thank you for the bug report @swanderz. Fixed in https://github.com/vega/vega-lite/pull/6724.

[...]

PR #6724, opened 2020-07-11
fix: respect symbolOpacity, gradientOpacity, symbolFillColor, and symbolStrokeColor

domoritz:
Fixes https://github.com/vega/vega-lite/issues/6719


examples/compiled/circle_binned.vg.json
```diff
@@ -157,7 +157,6 @@
   "legends": [
     {
       "size": "size",
-      "gradientLength": {"signal": "clamp(height, 64, 200)"},
```
donghaoren, line 160: Could you explain why this is removed?

examples/compiled/concat_marginal_histograms.vg.json
```diff
@@ -441,7 +441,6 @@
     {
       "fill": "color",
       "gradientLength": {"signal": "clamp(concat_1_height, 64, 200)"},
-      "symbolType": "square",
```
donghaoren, line 444: Why is this removed?

src/compile/legend/parse.ts
```diff
@@ -126,6 +125,13 @@ export function parseLegendForChannel(model: UnitModel, channel: NonPositionScal
   };
 
   for (const property of LEGEND_COMPONENT_PROPERTIES) {
+    if (
+      (legendType === 'gradient' && property.startsWith('symbol')) ||
+      (legendType === 'symbol' && property.startsWith('gradient'))
+    ) {
+      continue;
```
donghaoren, line 132: So we are skipping gradient properties for symbol legend and vice versa, and that's why we are seeing some symbolType and gradientLength properties being removed, right?

domoritz: Yes!

examples/compiled/circle_binned.vg.json
domoritz: The legend is a symbol legend and doesn't use gradient length. 

You found the code below: https://github.com/vega/vega-lite/pull/6724/files#r453887311

examples/compiled/concat_marginal_histograms.vg.json
domoritz: Same as above

examples/compiled/point_angle_windvector.vg.json
```diff
@@ -87,7 +87,6 @@
   "legends": [
     {
       "size": "size",
-      "gradientLength": {"signal": "clamp(height, 64, 200)"},
```
Saba9, line 90: Do you know why the PNG image generated for this spec changed?

domoritz: Because the png generation isn't stable. I've seen this in a few places. 

Just ignore it when the SVG doesn't change either.

domoritz: It would be good to figure out how we can minimize the noise @kanitw

PR #6737, opened 2020-07-13
docs: Adding title example to tooltip docs

ArvindEnvoy:
The documentation for `tooltip` did not make any references to the `title` parameter, so I am adding an example here.


domoritz:
Thanks. Replaced by https://github.com/vega/vega-lite/pull/6738

PR #6740, opened 2020-07-17
fix: errorbar thickness and size

### LINKED ISSUE #6495: Setting size of errorbar removes the ticks

chanwutk:
Fix #6495 from the discussion in #6502 
- Drop size encoding in errorbar
- `mark.size` sets `mark.ticks.size` if `mark.ticks.size` is not specified
- `mark.thickness` sets:
  - `mark.ticks.thickness` if `mark.ticks.thickness` is not specified
  - `mark.rule.size` if `mark.rule.size` is not specified

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -6021,6 +6021,14 @@
             }
           ]
         },
+        "size": {
+          "description": "Size of the ticks of an error bar",
+          "type": "number"
+        },
+        "thickness": {
+          "description": "Thickness of the ticks and the bar of an error bar",
+          "type": "number"
+        },
         "ticks": {
           "anyOf": [
             {
@@ -6073,6 +6081,14 @@
               "$ref": "#/definitions/MarkConfig"
             }
           ]
+        },
+        "size": {
+          "description": "Size of the ticks of an error bar",
+          "type": "number"
+        },
+        "thickness": {
+          "description": "Thickness of the ticks and the bar of an error bar",
+          "type": "number"
         },
         "ticks": {
           "anyOf": [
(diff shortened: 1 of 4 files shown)


examples/specs/normalized/errorbar_2d_vertical_ticks_normalized.vl.json
```diff
@@ -29,7 +29,6 @@
       "mark": {
         "type": "tick",
         "orient": "horizontal",
-        "aria": false,
```
domoritz, line 32: We should only expose one accessibility element per discrete item to screen readers. Make sure that aria false stays set.

domoritz:
Do you know why we make ticks blue and with lower opacity? That doesn't seem to make sense as a default. 

image

chanwutk:
Right, do we want to change the default to black?

domoritz:
Hmm, good question. I think the color of the ticks and the rule should be the same and they should have 100% opacity. Maybe for consistency, the color should be blue (default mark color). @kanitw do you agree?

PR #6754, opened 2020-07-20
fix: fix axis translate signal for binned bar marks

### LINKED ISSUE #6743: Signal for axis translate breaks binned marks
```
{
  "mark": "bar",
  "encoding": {
    "x": {
      "bin": true,
      "field": "IMDB_Rating",
      "type": "quantitative"
    },
    "y": {
      "aggregate": "count",
      "type": "quantitative"
    }
  },
  "config": {
    "axis": {
      "translate": { "signal": "0.5" }
    }
  }
}
```

Generates:

```
Expression parse error: (!isValid(datum["bin_maxbins_10_IMDB_Rating"]) || !isFinite(+datum["bin_maxbins_10_IMDB_Rating"]))?0:_scale("x", datum["bin_maxbins_10_IMDB_Rating"])+((xAxisFlipsForRtl && locale.characterDirection === "rightToLeft" ? true : false ? NaN : [object Object]0.5))
```

Need to add signal support for translate.

haldenl:
fixes #6743

### DIFF
--- a/src/compile/mark/encode/position-rect.ts
+++ b/src/compile/mark/encode/position-rect.ts
@@ -211,7 +211,7 @@ function getBinSpacing(
   channel: PositionChannel | PolarPositionChannel,
   spacing: number,
   reverse: boolean | SignalRef,
-  translate: number,
+  translate: number | SignalRef,
   offset: number | SignalRef
 ) {
   if (isPolarPositionChannel(channel)) {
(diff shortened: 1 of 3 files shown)


src/compile/mark/encode/position-rect.ts
```diff
@@ -220,22 +220,19 @@ function getBinSpacing(
 
   const spacingOffset = channel === 'x' || channel === 'y2' ? -spacing / 2 : spacing / 2;
 
-  if (isSignalRef(reverse)) {
+  if (isSignalRef(reverse) || isSignalRef(offset) || isSignalRef(translate)) {
```
haldenl, line 223: simplify the logic of this method by collapsing into 2 cases:

1. If there is a signal for any of `reverse`, `offset`, or `translate`
2. If none are signals.

Tests are modified to reflect this change.

test/compile/mark/bar.test.ts
```diff
@@ -345,15 +345,15 @@ describe('Mark: Bar', () => {
...
+        {scale: 'y', field: 'bin_maxbins_10_Horsepower', offset: {signal: '0.5 + (a + -0.5)'}}
       ]);
       expect(props.y).toEqual([
         fieldInvalidTestValueRef(y, 'y'),
         {
           scale: 'y',
           field: 'bin_maxbins_10_Horsepower_end',
           offset: {
-            signal: '1 + a'
+            signal: '0.5 + (a + 0.5)'
```
domoritz, line 356: This looks worse than before. Any chance you can simplify the expression?

haldenl: i spent a good amount of time trying, but the complexity wasn't worth it imo. Here's the non-signal expression:

`translate + (reverse ? -offset - spacingOffset : +offset + spacingOffset)`

The problem I ran into is that there are 3 values (`translate`, `offset`, and `spacing`) and 2 of them can be signals (`translate` and `offset`). You'd have to conditionally combine `spacing` and the non-signal (if there is one). On top of that, `offset` and `spacing` are modified (turned negative) if `reverse` is true, but `translate` is not.

Instead, this PR takes the approach of the formula

`translate +  (reverse ? -1 : 1) * (offset + spacing)` and collapses when values are missing. 

But let me know if you see a better solution I might have overlooked.

domoritz: I guess what's nice is that we use the same expressions.

PR #6755, opened 2020-07-21
fix: incorrect boxplot when using temporal field

### LINKED ISSUE #4524: Unexpected offset of boxplots when filtering aggregated temporal data

### LINKED ISSUE #4925: Boxplot missing/broken whiskers when using timeUnit
When I try to group data by weekday or month (or any `timeUnit`) and use `"extent": "min-max"`, whiskers are displayed, but when using the default 1.5 IQR, no whiskers are drawn. If I switch to `ordinal` or `nominal` types, detached whiskers are drawn in a new column. This seems to be related to issue #4524 

image

image

image

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v3.2.1.json",
  "hconcat": [
    {
      "data": {"url": "https://vega.github.io/vega-datasets/data/seattle-temps.csv"},
      "mark": {"type": "boxplot",  "extent": "min-max"},
      "encoding": {
        "x": {"type": "temporal", "field": "date", "timeUnit": "month"},
        "y": {"type": "quantitative", "field": "temp"}
      }
    },
    {
      "data": {"url": "https://vega.github.io/vega-datasets/data/seattle-temps.csv"},
      "mark": "boxplot",
      "encoding": {
        "x": {"type": "temporal", "field": "date", "timeUnit": "month"},
        "y": {"type": "quantitative", "field": "temp"}
      }
    },
    {
      "data": {"url": "https://vega.github.io/vega-datasets/data/seattle-temps.csv"},
      "mark": "boxplot",
      "encoding": {
        "x": {"type": "ordinal", "field": "date", "timeUnit": "month"},
        "y": {"type": "quantitative", "field": "temp"}
      }
    }
  ]
}
```

chanwutk:
Fix #4524 
Fix #4925

Boxplot did not include `bins` and `timeUnits` transform into `filteredLayersMixins`' transform. So, I included them.

### DIFF
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -104,6 +104,8 @@ export function normalizeBoxPlot(
 
   const boxPlotType = getBoxPlotType(extent);
   const {
+    bins,
+    timeUnits,
     transform,
     continuousAxisChannelDef,
     continuousAxis,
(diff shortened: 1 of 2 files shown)


chanwutk:
At line 244, do we need the if statement `if (boxPlotType === 'tukey')`?
`boxPlotType` is always either `'tukey'` or `'min-max'`, so in this line it is going to always be `'tukey'` because it is under another if (`if (boxPlotType !== 'min-max')`).

domoritz:
> At line 244, do we need the if statement if (boxPlotType === 'tukey')?
boxPlotType is always either 'tukey' or 'min-max', so in this line it is going to always be 'tukey' because it is under another if (if (boxPlotType !== 'min-max')).

Good catch. Let's simplify the code. 

We can also simplify `if (outlierLayersMixins && filteredWhiskerSpec)` since `filteredWhiskerSpec` is always true.

src/compositemark/boxplot.ts
```diff
@@ -300,17 +302,14 @@ export function normalizeBoxPlot(
       }
     })[0];
 
+    const filteredLayersMixinsTransforms = [...bins, ...timeUnits, joinaggregateTransform];
     if (outlierLayersMixins && filteredWhiskerSpec) {
```
domoritz, line 306:
```suggestion
    if (outlierLayersMixins) {
```

domoritz: If you simplify the code above.

domoritz:
Thank you!

PR #6784, opened 2020-08-13
refactor: simplify optimizer code

domoritz:


src/compile/data/optimize.ts
```diff
@@ -58,19 +58,21 @@ export function isTrue(x: boolean) {
  *
  * @param optimizer The optimizer instance to run.
  * @param nodes A set of nodes to optimize.
- * @param flag Flag that will be or'ed with return valued from optimization calls to the nodes.
  */
 function runOptimizer(optimizer: BottomUpOptimizer | TopDownOptimizer, nodes: DataFlowNode[]): boolean {
-  const flags = nodes.map(node => {
+  const mutatedFlags: Set<boolean> = new Set();
```
donghaoren, line 63: A set of boolean? Why not just a single boolean that you flip to true when hitting a true in the for loop?

domoritz: 🤦

domoritz: I was just copying code from below that I could also simplify.

PR #6801, opened 2020-08-18
docs: add global development example

domoritz:
Merge after https://github.com/vega/vega-lite/pull/6804


domoritz:
Yeah, it's even slightly improved but pretty much the same example.

PR #6804, opened 2020-08-19
feat: default type for order channel

domoritz:


src/channeldef.ts
```diff
@@ -280,7 +280,7 @@ export interface TypeMixins<T extends Type> {
    * The type of measurement (`"quantitative"`, `"temporal"`, `"ordinal"`, or `"nominal"`) for the encoded field or constant value (`datum`).
    * It can also be a `"geojson"` type for encoding ['geoshape'](https://vega.github.io/vega-lite/docs/geoshape.html).
    *
-   * Since Vega-Lite 4.14.0, Vega-Lite automatically infers data types in many cases as discussed below. However, type is required for a field if:
```
kanitw, line 283: Why removing this clause?

src/channeldef.ts
```diff
@@ -911,6 +911,9 @@ export function defaultType<T extends TypedFieldDef<Field>>(fieldDef: T, channel
     case 'shape':
     case 'strokeDash':
       return 'nominal';
+
+    case 'order':
```
kanitw, line 915: Add comment to explain why we need to add this ordinal thing? 

(Why does it matter? Does it solve any issue?)

src/channeldef.ts
domoritz: We don't mention specific versions in other places, do we?

src/channeldef.ts
domoritz: Answered offline: the default was nominal but nominal causes a warning.

PR #6820, opened 2020-08-26
fix: add color encoding to boxplot outliers

### LINKED ISSUE #6818: Box plot outliers are not colored
Please:

- [x] Provide an example spec in JSON, wrapped by triple backticks

---

Hi!

According [to the docs](https://vega.github.io/vega-lite/docs/boxplot.html#color-size-and-opacity-encoding-channels) of box plots 
> The color is applied to only the box, the median tick, and the outlier points. 

First, the outlier points are not colored, but probably should be to directly see where they belong to (orange here):
image
https://vega.github.io/editor/#/examples/vega-lite/boxplot_2D_vertical

Second, the median tick is not colored like the box (or the outliers) as described - but shouldnt be anyway imo.

domoritz:
Fixes #6818

Adds the boxplot's color encoding to the outliers' encoding. If no color encoding is specified nothing is added.


domoritz:
Thank you @keckelt!

PR #6825, opened 2020-08-31
feat: support `expr` in header

kanitw:


domoritz:
Tests are failing but looks reasonable otherwise.

Issue #6706, opened 2020-06-28
MarkDef "url" parameter doesn't work


marcprux:
The documentation for [image mark](https://vega.github.io/vega-lite/docs/image.html#properties) says that the mark properties accept some "special" properties, like "aspect" and "url". "aspect" does indeed work, but the following spec reports that "url" is not allowed (and it does not render the image): 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
  "description": "A simple bar chart with a background image.",
  "data": {
    "values": [
      {"a": "A", "b": 28},
      {"a": "B", "b": 55},
      {"a": "C", "b": 43},
      {"a": "D", "b": 91},
      {"a": "E", "b": 81},
      {"a": "F", "b": 53},
      {"a": "G", "b": 19},
      {"a": "H", "b": 87},
      {"a": "I", "b": 52}
    ]
  },
  "layer": [
    {
      "data": null,
      "mark": {
        "type": "image",
        "url": "data/ffox.png",
        "x": 100,
        "y": 100,
        "aspect": false,
        "width": 300,
        "height": 100
      }
    },
    {
      "mark": "bar",
      "encoding": {
        "x": {"field": "a", "type": "ordinal", "axis": {"labelAngle": 0}},
        "y": {"field": "b", "type": "quantitative"}
      }
    }
  ]
}
```

Embedding the image via a value setting of the "url" encoding is a work-around for including a static background image:

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
  "description": "A simple bar chart with a background image.",
  "data": {
    "values": [
      {"a": "A", "b": 28},
      {"a": "B", "b": 55},
      {"a": "C", "b": 43},
      {"a": "D", "b": 91},
      {"a": "E", "b": 81},
      {"a": "F", "b": 53},
      {"a": "G", "b": 19},
      {"a": "H", "b": 87},
      {"a": "I", "b": 52}
    ]
  },
  "layer": [
    {
      "data": null,
      "mark": {
        "type": "image",
        "x": 100,
        "y": 100,
        "width": 300,
        "height": 300
      },
      "encoding": {
        "url": { "value": "data/ffox.png" }
      }
    },
    {
      "mark": "bar",
      "encoding": {
        "x": {"field": "a", "type": "ordinal", "axis": {"labelAngle": 0}},
        "y": {"field": "b", "type": "quantitative"}
      }
    }
  ]
}
```

[...]

PR #6844, opened 2020-09-03
fix: MarkDef "url" parameter doesn't work

Zyh533:
This pull request is to fix the issue #6706 . I noticed that the image can be normally shown when the 'url' parameter in the mark encoding which type is 'image' in the generated vega sepc. Therefore,when parsed the vega-lite spec into the vega sepc, I add a new property 'urlImag' in src/compile/mark/image.ts , if the 'url' parameter is not undefined, it will be added into the mark encoding.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,37 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <option name="LINE_SEPARATOR" value="&#10;" />
+    <JSCodeStyleSettings>
+      <option name="FORCE_SEMICOLON_STYLE" value="true" />
+      <option name="SPACE_BEFORE_FUNCTION_LEFT_PARENTH" value="false" />
+      <option name="FORCE_QUOTE_STYlE" value="true" />
+      <option name="ENFORCE_TRAILING_COMMA" value="Remove" />
+      <option name="SPACES_WITHIN_OBJECT_LITERAL_BRACES" value="true" />
+      <option name="SPACES_WITHIN_IMPORTS" value="true" />
+    </JSCodeStyleSettings>
+    <TypeScriptCodeStyleSettings>
+      <option name="FORCE_SEMICOLON_STYLE" value="true" />
+      <option name="SPACE_BEFORE_FUNCTION_LEFT_PARENTH" value="false" />
+      <option name="FORCE_QUOTE_STYlE" value="true" />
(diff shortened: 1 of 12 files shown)


src/compile/data/assemble.ts
```diff
@@ -268,5 +268,38 @@ export function assembleRootData(dataComponent: DataComponent, datasets: Dict<In
     }
   }
 
+  // make the aggregate type transform in the first order
+  // to make it effective when there are more than one transforms in the data
+  for(const d of data){
```
domoritz, line 273: Please apply the prettier formatter (`yarn format`)

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,37 @@
+<component name="ProjectCodeStyleConfiguration">
```
domoritz, line 1: please remove these files

src/compile/data/assemble.ts
```diff
@@ -268,5 +268,38 @@ export function assembleRootData(dataComponent: DataComponent, datasets: Dict<In
...
+    swapTransformTypeIndex(d,"aggregate",0);
+  }
+
   return data;
 }
+
+// swap the index of transform for the type to the targetIndex
+function swapTransformTypeIndex(data:VgData,typeName:string,targetIndex:number) {
+  // if the data.transform is exist and its length >=2
```
domoritz, line 282: remove all comments that describe the code and only keep the ones that describe the why

src/compile/data/assemble.ts
```diff
@@ -268,5 +268,38 @@ export function assembleRootData(dataComponent: DataComponent, datasets: Dict<In
     }
   }
 
+  // make the aggregate type transform in the first order
```
domoritz, line 271: what does this have to do with the URL issue?

src/compile/mark/encode/image.ts
```diff
@@ -0,0 +1,12 @@
+import { UnitModel } from "../../unit";
+
+// encode the image url
+export function imageUrl(model: UnitModel) {
+  const { markDef } = model;
+  // if the mark definition contains the property of url as the image url to be shown
+  if(markDef["url"] != undefined){
```
domoritz, line 7: 'url' in markDef?

domoritz, line 7: combine the whole if into a ternary

domoritz (changes requested):

domoritz:
https://github.com/vega/vega/pull/2871 has a cleaner fix

PR #6852, opened 2020-09-07
feat: rollup everything

domoritz:
* Use rollup to build and bundle everything


scripts/create-example-pages
```diff
@@ -7,7 +7,6 @@
 
 const fs = require('fs');
 const examples = require('../site/_data/examples.json');
-const svg2png = require('svg2png-many');
```
kanitw, line 10: Why changing these?

domoritz: Yeah, somewhat unrelated but I noticed we are compiling pngs anyway so we don't need the extra library anymore.

PR #6867, opened 2020-09-14
docs: update Facet Bullet Chart

eitanlees:
Currently, the faceted bullet chart example looks like this

visualization (54)

I added the `labelAlign` parameter to clean up the y axis

visualization (53)


domoritz:
I think the right fix is in Vega. See https://github.com/vega/vega/issues/2233.

PR #6916, opened 2020-09-30
refactor: convert selections from Dict to array of named objects.

arvind:
Note: I'm waiting on converting all the example specs + test cases until more of the refactoring is done to reduce churn. But this PR allows us to examine changes to the compiler source code.


arvind:
I think for the initial refactoring, I'll just push the commits and issue the PR on the main feature branch (`as/selection-params`). Otherwise, the PRs are going to be very noisy with schema/example changes that aren't correct.

PR #6917, opened 2020-09-30
refactor: align selection syntax with params syntax

arvind:
This PR brings the selection syntax in line with params. Future PRs will continue refactoring as well as other feature changes described in our [params proposal](https://github.com/vega/vega-lite-params-proposal). 

To summarize this PR:
* Selections are now defined as an array of named objects, rather than an object with named properties. This change should hopefully make clearer that selection names are not special/reserved keywords.
* Selection properties are nested within a `select` property. This property can also be a string of the selection type (i.e., `single`, `multi`, `interval`) to concisely adopt defaults.
* Selection initialization now occurs via `value` rather than `init` and occurs outside the `select` block, to align it with parameters.
* Selection binding similarly occurs outside the `select` block, again to align it with parameters.


src/compile/selection/transforms/project.ts
```diff
@@ -63,42 +68,46 @@ const project: TransformCompiler = {
       return {[range]: sg};
     };
 
+    const type = selCmpt.type;
+    const init = selDef.value;
```
domoritz, line 72:
```suggestion
    const {type, init} = selCmpt;
```

src/compile/selection/transforms/project.ts
```diff
@@ -152,16 +161,15 @@ const project: TransformCompiler = {
...
-    if (selDef.init) {
+    if (init) {
       const parseInit = <T extends SelectionInit | SelectionInitInterval>(i: Dict<T>): T[] => {
         return proj.items.map(p => (i[p.channel] !== undefined ? i[p.channel] : i[p.field]));
       };
 
-      if (selDef.type === 'interval') {
-        selCmpt.init = parseInit(selDef.init);
+      if (type === 'interval') {
+        selCmpt.init = parseInit(selDef.value as SelectionInitIntervalMapping);
```
domoritz, line 170: Why is the type case needed here? Is `selDef.value` not compatible with what `parseInit` expects?

src/selection.ts
```diff
@@ -229,32 +183,74 @@ export interface IntervalSelectionConfig extends BaseSelectionConfig {
...
    * - `"single"` -- to select a single discrete data value on `click`.
    * - `"multi"` -- to select multiple discrete data value; the first value is selected on `click` and additional values toggled on shift-`click`.
    * - `"interval"` -- to select a continuous range of data values on `drag`.
    */
-  type: T;
+  select:
+    | SelectionType
+    | ({
+        type: SelectionType;
+      } & (T extends 'single'
```
domoritz, line 207: Idea: maybe add more generic arguments to BaseSelectionDef to avoid this complicated type logic. Alternatively, we could have a base type that we inherit from but I'm not sure whether the scheme generator would inherit the comments (which I suppose is the goal of having the type BaseSelectionDef).

domoritz (approved):

src/selection.ts
arvind: Yeah I hear you. This logic should simplify once we collapse `single` and `multi` into a single `point` selection type. I tried to simplify further and this is where I got to – but you're savvier with TypeScript, so I'm open to suggestions :)

domoritz: I'll refactor this when it's in master. This works for now.

src/compile/selection/transforms/project.ts
arvind: Wait no, this change is incorrect.

src/compile/selection/transforms/project.ts
arvind: The type guard broke because we now read the selection type off `selCmpt` not `selDef`. And I wasn't able to get the type guard working again with alternate forms of the `if` statement. Definitely open to feedback though.

`if (isString(selDef.select) ? selDef.select === 'interval' : isObject(selDef.select) ? selDef.select.type === 'interval' : false) {`

src/compile/selection/transforms/project.ts
domoritz: Ups, yes. Should be:

```
const {type} = selCmpt;
const {value: init} = selDef;
```

arvind: Not sure I understand why we prefer that? Your suggestion is harder for me to read/understand.

domoritz: Your original code is fine. I don't have strong preferences either way.

src/compile/selection/transforms/project.ts
domoritz: I see. It's hard to say for me without diving into the types so let's defer fixing this until we refactor the types of `BaseSelectionDef` as discussed below.

arvind: That sounds good to me :)

PR #6919, opened 2020-10-01
feat: top-level selections

arvind:


src/parameter.ts
```diff
@@ -29,9 +30,10 @@ export interface Parameter {
   bind?: Binding;
 }
 
-export function assembleParameterSignals(params: Parameter[]) {
+export function assembleParameterSignals(params: (Parameter | TopLevelSelectionDef)[]) {
   const signals: (NewSignal | InitSignal)[] = [];
   for (const param of params || []) {
+    if (isParameterSelection(param)) continue;
```
arvind, line 36: @kanitw, my hunch is that these changes shouldn't be required if I break out a `Generic`/`Normalized` equivalent for `TopLevelProperties` but I'm not entirely sure how to do that and/or what the knock on effect of changes would be. I'd love some help. But I'm also down to leave this as is, because it's a pretty small/localized issue and doesn't even strike me as particularly hacky.

arvind: I'm going to mark this as resolved for now based on Slack discussion that we will continue working on making it possible to define params/selections at any level of the spec tree.

examples/specs/brush_table.vl.json
```diff
@@ -6,7 +6,7 @@
     "window": [{"op": "row_number", "as": "row_number"}]
   }],
   "hconcat": [{
-    "selection": [{"name": "brush", "select": "interval"}],
+    "params": [{"name": "brush", "select": "interval"}],
```
kanitw, line 9: I feel like `"select": "interval"` reads kinda weird.  Wouldn't it be clearer if this is `selection`?

I could see that it doesn't make sense as the first commit in https://github.com/vega/vega-lite/pull/6926 because having selection inside selection is kinda weird.

src/parameter.ts
kanitw: Do we skip parameter selection here because selection has a different codepath? 

If so, this seems fine. But let's add a comment.

kanitw (approved):
The select -> selection, if we wanna do it, could be in a different PR though.  So i'm approving this.

examples/specs/brush_table.vl.json
arvind: I prefer `select` both because it's the verb form (and so reads more naturally, i.e., "select an interval") and because it's consistent with `bind` (rather than `binding`) which occurs at the same level.

src/parameter.ts
arvind: Yes, separate code path. Comment added with https://github.com/vega/vega-lite/pull/6919/commits/8ddace24bf594503103211f97c5c86762d48d3d3.

examples/specs/brush_table.vl.json
domoritz: I agree with @arvind.

Issue #6920, opened 2020-10-02
tooltip causes aggregate fail


eflister:
i made a line graph following the method from https://vega.github.io/vega-lite/examples/interactive_line_hover.html to label each line.  

![image](https://user-images.githubusercontent.com/366563/94880156-12bbf980-0417-11eb-9579-59953b0749f9.png)

it works as expected until i add a field to the tooltip.  then, instead of the aggregation eliminating all but the last label for each line, labels show up for every datapoint.

![image](https://user-images.githubusercontent.com/366563/94880029-bce75180-0416-11eb-9316-8eb19fbdf71d.png)

https://gist.github.com/eflister/d743fa9ef6641d26a1d09c41265aba1c
note that removing the `pct` field from the `tooltip` in line 4611: `{"field": "pct"}` yields the expected aggregate behavior.

domoritz:
Tooltips add another aggregate dimension. If you don't want that, you need to provide an aggregation function.

eflister:
is this documented anywhere?  i don't see anything on https://vega.github.io/vega-lite/docs/tooltip.html

>  If you don't want that, you need to provide an aggregation function.

i'm new, i don't know what you're referring to.  is it this?  https://vega.github.io/vega/docs/api/view/#view_tooltip

i'm using `vega-embed` so i guess that means i'm already using `vega-tooltip`...

domoritz:
A tooltip will be categorical by default and therefore be used for the group by. The aggregate I was referring to is https://vega.github.io/vega-lite/docs/aggregate.html. That page also mentions that "all fields without aggregation function specified are treated as group-by fields". I hope that clears things up. Let me know if I misunderstood your questions or issue.

eflister:
the aggregate page doesn't say anything about tooltips.  would you mind pointing me to a worked example of what you're describing?  i don't see how to "provide an aggregation function" other than what i did normally (`max`/`argmax`).  it seems like it would be a really common situation that people would run into tooltips breaking their aggregations, shouldn't the docs have to discuss it specifically?

domoritz:
You are right that the docs don't explicitly mention that tooltips. "all fields without aggregation function specified are treated as group-by fields" says that all fields without aggregate (which includes tooltips) are treated as group-by fields. 

Here is an example to demonstrate the difference. 

[...]

PR #6921, opened 2020-10-02
docs: document that encoding a field as a tooltip groups by the field

domoritz:
Fixes https://github.com/vega/vega-lite/issues/6920


kanitw:
Wondering if this comment really make sense for only tooltip, given adding an unaggregated field to *any* encoding?

domoritz:
I generally agree. People seem to struggle a lot more for tooltips than other encodings, though.

domoritz:
Should we still merge this?

donghaoren (approved):
An example might help here? I had to read the sentence a few times to comprehend it...

domoritz:
I rewrote the text and added examples. What do you think now @donghaoren?

PR #6925, opened 2020-10-04
docs: add BBC Visual and Data Journalism cookbook port to examples page

aezarebski:
I add a link to a collection of examples described below to the list of external examples on the website.

**Project description** 

This project is a port of the [BBC ggplot2 cookbook](https://bbc.github.io/rcookbook/) into Vega-Lite. This
makes use of the `gapminder` dataset which is included as a CSV in this
repository. The goal is to have useful examples of vega-lite specifications
recreating the main plots in the BBC cookbook. These /should/ be are easy to
reuse. If getting the last 20% of the visualisation to agree makes up 80% of the
code, then I will opt for only getting it 80% correct.


domoritz:
Thank you

PR #6926, opened 2020-10-05
feat!: selections as parameters

arvind:


domoritz:
This pull request is waiting for https://github.com/vega/vega-lite/pull/7116, which is blocked by https://github.com/vega/vega/pull/3008.

domoritz (approved):

examples/specs/brush_table.vl.json
```diff
@@ -6,23 +6,19 @@
...
-      }
-    },
+    "params": [{"name": "brush", "select": "interval"}],
     "mark": "point",
     "encoding": {
       "x": {"field": "Horsepower", "type": "quantitative"},
       "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
       "color": {
-        "condition": {"selection": "brush", "field": "Cylinders", "type": "ordinal"},
+        "condition": {"param": "brush", "field": "Cylinders", "type": "ordinal"},
```
kanitw, line 15: I know we have found a smart way to unify variable and selection into `param`. And that sort of make sense. 

But seeing these code diff makes me wonder if the old code on the left makes it clearer what happens behind the scene (that brush is a selection parameter). 

So that's the cost of this unification.

domoritz: I agree with the comment but I think keeping `param` is better since one might use a param to toggle whether all marks have a different color. In that case there would be no fields or type but still a `param`. It's nice to be able to *upgrade* to a selection param later without having to change the key of the property.

kanitw: > one might use a param to toggle whether all marks have a different color.  
> In that case there would be no fields or type but still a param.

But we already have `"condition": {"test": "a", ...}`, which already works when a is a variable parameter. 

Basically, with this unification, we are adding two ways to using variable parameter as a predicate. 
 
> It's nice to be able to upgrade to a selection param later without having to change the key of the property.

Can you give an example of an "upgrade" scenario that may happen as you write analysis code?

I feel like it's hard to imagine one because variable parameter is operating as a scalar variable.  Meanwhile, selection has underlying a data collection.  

So conditioning with variable parameter and with selection parameter aren't exactly the same kind of operation.  While variable parameter is its own standalone thing that is just plugged in as we apply, conditioning with selection means that it interacts with the data point.  

It's worth noting the term "parameter" doesn't naturally convey to me that it can "interact" with the data field whereas the term "selection" makes that quite clear. (We can interpret the code `"selection": "brush"` as "in the 'brush' selection".  To interpret that `"param": "brush"` in the same way requires a bit more leap (by looking the declaration of brush to know that it's a selection parameter). 

I know that people can still learn the concept that selection is a parameter type. 
However, thinking more about this, I can't help but think again that the language might be more obvious to learn if we don't unify parameter+selection.

domoritz: > But we already have "condition": {"test": "a", ...}, which already works when a is a variable parameter.

Test expects an expression. Param expects a param name and generated the correct check automatically (which for non selection params is the same). 

> Can you give an example of an "upgrade" scenario that may happen as you write analysis code?

I was thinking of a case where I incrementally write an interactive spec. I start out with just trying to change the color of a mark and then continuing to define the selection.

kanitw: > I was thinking of a case where I incrementally write an interactive spec. I start out with just trying to change the color of a mark and then continuing to define the selection.

I see. To make this more concrete, can you think of a series of specs that you may write in this scenario? 

> > But we already have "condition": {"test": "a", ...}, which already works when a is a variable parameter.
> Test expects an expression. Param expects a param name and generated the correct check automatically (which for non selection params is the same).

But variable can be referred in an expression.  Vega also checks if a signal name exists and throw error when users referred to undeclared signals.  Vega-Lite would in turn get similar errors.

kanitw: To clarify why this kinda feels counterintuitive for me: 

> I feel like it's hard to imagine one because variable parameter is operating as a scalar variable. Meanwhile, selection has underlying a data collection.


It's kinda like in Swift / Java, you begin by declaring a variable to be scalar or collection from the beginning.  If suddenly you change from scalar to collection, you shouldn't expect that the rest of the code still works since the semantic of the variable has changed.

PR #6927, opened 2020-10-05
refactor: combine `single` and `multi` selections as `point` selections

arvind:
Review/merge after #6919.


kanitw (approved):
Question: I assume that the new point selection stores data as a collection like multi?  

(I can't remember if the old single selection stores data as a scalar or a collection with at most one item.)

arvind:
> Question: I assume that the new point selection stores data as a collection like multi?

Yes. 

> (I can't remember if the old single selection stores data as a scalar or a collection with at most one item.)

The latter.

PR #6935, opened 2020-10-06
fix: set `defaultLabelOverlap` to "greedy" for symlog scale type

### LINKED ISSUE #6934: Symlog scale should use greedy label overlap
See https://github.com/vega/vega/issues/2914. When the scale type is symlog, we should set label overlap to greedy as well.

baogianghoangvu:
Closes #6934 

- Set `defaultLabelOverlap` in Axis and Legend to "greedy" for symlog scale type
- Add corresponding tests

Thanks for reviewing!


domoritz:
Merged via https://github.com/vega/vega-lite/pull/6936

PR #6939, opened 2020-10-07
docs: fix documentation for mark text baseline

john-guerra:
The documentation was broken for the [mark text baseline property](https://vega.github.io/vega-lite/docs/text.html) the documentation formatting was broken. Took me a lot of time to figure this out, but I think this should fix it


domoritz:
Thank you. Merging in https://github.com/vega/vega-lite/pull/6940.

Issue #6823, opened 2020-08-31
Schema shouldn't include normalized spec / FieldName


kanitw:
Somehow it currently does. 

Search for `Normalized` in https://raw.githubusercontent.com/vega/vega-lite/master/build/vega-lite-schema.json

kanitw:
https://github.com/vega/vega-lite/commit/5c0b29951c901758c51701b458995d97fd85c92d#r41899248 is probably the culprit (as part of https://github.com/vega/vega-lite/pull/6002/)

PR #6950, opened 2020-10-09
fix: correct which specs are normalized

domoritz:
Repeat specs are normalized by filling in the field name.

Part of https://github.com/vega/vega-lite/issues/6823

### DIFF
--- a/scripts/rename-schema.sh
+++ b/scripts/rename-schema.sh
@@ -9,8 +9,8 @@ perl -pi -e s,'<ExprRef>','',g build/vega-lite-schema.json
 perl -pi -e s,'\,ExprRef>','>',g build/vega-lite-schema.json
 
 perl -pi -e s,'CompositeEncoding','Encoding',g build/vega-lite-schema.json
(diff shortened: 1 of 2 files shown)


scripts/rename-schema.sh
```diff
@@ -9,8 +9,8 @@ perl -pi -e s,'<ExprRef>','',g build/vega-lite-schema.json
 perl -pi -e s,'\,ExprRef>','>',g build/vega-lite-schema.json
 
 perl -pi -e s,'CompositeEncoding','Encoding',g build/vega-lite-schema.json
-perl -pi -e s,'Generic(.*)<FacetedUnitSpec\,LayerSpec\,?.*\,FieldName>','\1',g build/vega-lite-schema.json
```
kanitw, line 12: Being normalized means it can only has FieldName, but not vice versa.  (normalized means much more.)  

I think you mean more like nonrepeated spec?

kanitw (changes requested):

scripts/rename-schema.sh
domoritz: Eventually, normalized specs shouldn't even appear in the schema and I'm working on that. In the meantime, this is flipping the naming to be more correct.

domoritz:
Done better in #6965

PR #6953, opened 2020-10-10
feat: enable `interval` selections for cartographic projections

arvind:
Kapture 2020-10-10 at 15 40 57

Chained to #7933.

<details>
  <summary>📦 Published PR as canary version: <code>5.2.1--canary.6953.50ab21f.0</code></summary>
  <br />

  ```bash
  npm install vega-lite@5.2.1--canary.6953.50ab21f.0
  # or 
  yarn add vega-lite@5.2.1--canary.6953.50ab21f.0
  ```
</details>


src/compile/selection/assemble.ts
```diff
@@ -114,23 +114,29 @@ export function assembleTopLevelSignals(model: UnitModel, signals: Signal[]) {
...
 export function assembleUnitSelectionData(model: UnitModel, data: readonly VgData[]): VgData[] {
   const dataCopy = [...data];
+  const unit = unitName(model, {escape: false});
+
   for (const selCmpt of vals(model.component.selection ?? {})) {
-    const init: VgData = {name: selCmpt.name + STORE};
+    const store: VgData = {name: selCmpt.name + STORE};
+
+    if (selCmpt.project.hasSelectionId) {
+      store.transform = [{type: 'collect', sort: {field: SELECTION_ID}}];
```
dvmoritzschoefl, line 123: @arvind regarding https://github.com/vega/vega-lite/pull/7932 and this PR I have a question:
- Is the transform for datum.values[0] missing here? because you mentioned that the transform for the lasso must include this formula to correctly work with 'vlSelectionTuples'

`
"transform": [
          {"type": "formula", "expr": "datum.values[0]", "as": "_vgsid_"},
          {"type": "collect", "sort": {"field": "_vgsid_"}}
        ]
`

arvind: Yes, good eye! In https://github.com/vega/vega-lite/pull/7933 and https://github.com/vega/vega/pull/3397 (both of which this PR depends on), I introduced a more optimized format for selection stores when only the `SELECTION_ID` is being stored. In particular, after those two PRs, the selection store holds tuples in the form `{_vgsid_, unit}` sorted by `_vgsid_`. This structure allows us to skip the `formula` transform, as well all the array generation that would previously happen for the `fields` and `values` keys of the older tuples. The `{unit, fields, values}` tuple format still applied for non-selectionID based stores. 

Hope that helps!

src/compile/unit.ts
```diff
@@ -281,7 +281,7 @@ export class UnitModel extends ModelWithField {
     if (this.encoding.x || this.encoding.y) {
       return 'cell';
     } else {
-      return undefined;
+      return 'view';
```
kanitw, line 284: Any reasons why we're adding `view` here?

kanitw (commented):
Have some questions from a first skim. Will look more and add more comments.

src/compile/selection/interval.ts
```diff
@@ -1,104 +1,198 @@
...
-    selCmpt.project.items.forEach((proj, i) => {
-      const channel = proj.channel;
-      if (channel !== X && channel !== Y) {
-        warn('Interval selections only support x and y encoding channels.');
-        return;
+    if (!model.hasProjection) {
+      // Proxy scale reactions to ensure that an infinite loop doesn't occur
+      // when an interval selection filter touches the scale.
+      if (!scales.defined(selCmpt)) {
+        const triggerSg = name + SCALE_TRIGGER;
```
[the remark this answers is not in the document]
arvind: Short for signal. This is a convention used throughout `src/compile/selection/` code. Do you want me to update this PR to fix it everywhere, send a separate PR, or keep it as is?

src/compile/selection/interval.ts
```diff
@@ -1,104 +1,198 @@
...
+        const scaleTriggers = channels.map(proj => {
+          const channel = proj.channel as 'x' | 'y';
+          const {data: dname, visual: vname} = proj.signals;
```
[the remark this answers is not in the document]
arvind: I wanted the variable name to clearly indicate that it was referencing the name of the visual signal.

src/compile/selection/parse.ts
```diff
@@ -52,9 +52,10 @@ export function parseUnitSelection(model: UnitModel, selDefs: SelectionParameter
       events: isString(defaults.on) ? parseSelector(defaults.on, 'scope') : array(duplicate(defaults.on))
     } as any);
 
+    const def_ = duplicate(def); // defensive copy to prevent compilers from causing side effects
```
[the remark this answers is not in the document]
arvind: Yes, this PR introduces side effects — namely, if the interval selection parser detects a map projection, it rewrites the defaults from `x/y` projection to instead `lat/lon` and updates the predicate field to `SELECTION_ID` ready for subsequent selection parsers to operate. This was the cleanest, most scoped/localized way I could think of to introduce brushing on maps without triggering a more massive refactor since the rest of the interval selection components expect standard cartesian coordinate systems. So my preference would be to keep the defensive copy since its occurring just on the selection definitions.

src/compile/unit.ts
arvind: I'm trying to page this in since it's been a while but I believe it's to force a transparent background to be able to capture events. Without a `'view'` style, there is no background specified on the overall view and so brushes only trigger if someone mouses down on a mark.

arvind: Yes, https://github.com/vega/vega/pull/3480

src/compile/selection/interval.ts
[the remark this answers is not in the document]
domoritz: Keeping as is is fine but let's update to `triggerSignal` in a follow up pull request. sg could be misinterpreted as scenegraph.

src/compile/selection/interval.ts
```diff
@@ -1,104 +1,198 @@
...
+import {LayoutSizeIndex} from '../layoutsize/component';
+import {isUnitModel} from '../model';
 import {UnitModel} from '../unit';
 import {assembleInit} from './assemble';
 import {SelectionProjection, TUPLE_FIELDS} from './project';
 import scales from './scales';
 
 export const BRUSH = '_brush';
 export const SCALE_TRIGGER = '_scale_trigger';
+export const GEO_INIT_TICK = 'geo_interval_init_tick'; // Workaround for https://github.com/vega/vega/issues/3481
```
domoritz, line 19: Can you add something to the issue about how we need to update Vega-Lite when the issue is fixed in Vega?

domoritz (approved):
Looks good overall. Very exciting change!

src/compile/selection/interval.ts
arvind: Done!

domoritz:
🎉

PR #6965, opened 2020-10-19
fix: remove normalized spec from schema

### LINKED ISSUE #6823: Schema shouldn't include normalized spec / FieldName

domoritz:
fixes #6823

Needs https://github.com/vega/ts-json-schema-generator/pull/549

### DIFF
--- a/src/compositemark/index.ts
+++ b/src/compositemark/index.ts
@@ -53,20 +53,18 @@ export function remove(mark: string) {
   delete compositeMarkRegistry[mark];
 }
 
-export type CompositeEncoding<F extends Field = Field> = Encoding<F> & ErrorExtraEncoding<F>;
+export type CompositeEncoding<F extends Field> = Encoding<F> & ErrorExtraEncoding<F>;
 
 export type PartialIndex<T extends Encoding<any>> = {
   [t in keyof T]?: Partial<T[t]>;
 };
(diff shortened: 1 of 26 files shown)


domoritz:
Also addresses https://github.com/vega/vega-lite/commit/5c0b29951c901758c51701b458995d97fd85c92d#r41899248

src/compositemark/index.ts
```diff
@@ -53,20 +53,18 @@ export function remove(mark: string) {
   delete compositeMarkRegistry[mark];
 }
 
-export type CompositeEncoding<F extends Field = Field> = Encoding<F> & ErrorExtraEncoding<F>;
+export type CompositeEncoding<F extends Field> = Encoding<F> & ErrorExtraEncoding<F>;
```
kanitw, line 56: Can you explain why do we remove the default `= Field` here?

domoritz: I found it less confusing to explicitly say the type. It makes me more confident that we didn't miss using string somewhere.

domoritz: Are you okay with this?

PR #6982, opened 2020-10-20
fix: supportMark accepts ExtendedChannel

isaacparker2:
Seems like this was just overlooked when channel type defs were changed, lemme know if I'm wrong.


domoritz:
I guess we didn't need support for the extended channel but this is definitely better so let's merge it. Thanks!

PR #6989, opened 2020-10-23
ci: test node 15

domoritz:
Needs https://github.com/Automattic/node-canvas/issues/1688


domoritz:
Will need to wait until all dependencies properly support node 15.

PR #7031, opened 2020-11-06
feat: don't support type or aggregation for detail and key

domoritz:
Extension of https://github.com/vega/vega-lite/pull/6921


domoritz:
It make the schema a bit stricter and sensible but also less flexible if I want to quickly change e.g. color to detail.

Probably not quite what we want/need right now.

PR #7052, opened 2020-11-18
fix!: change default projection to equalEarth

### LINKED ISSUE #5367: Change default projection type
Now that Vega includes the [equal earth projection](https://en.wikipedia.org/wiki/Equal_Earth_projection), could It be included in Vega-Lite as one of the built-in projection types?

VL4 might also be an opportunity to change the default projection type away from mercator as this is rarely the most appropriate for the kinds of mapping Vega-lite is used for. Personally, I'd argue that the identity projection makes most sense as the default as it is semantically closest to 'unprojected' when a projection type is not specified. But if that is not suitable, the equalEarth projection would be far superior as a default compared to the mercator.

domoritz:
fixes #5367

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -20364,7 +20364,7 @@
         },
         "type": {
           "$ref": "#/definitions/ProjectionType",
-          "description": "The cartographic projection to use. This value is case-insensitive, for example `\"albers\"` and `\"Albers\"` indicate the same projection type. You can find all valid projection types [in the documentation](https://vega.github.io/vega-lite/docs/projection.html#projection-types).\n\n__Default value:__ `mercator`"
+          "description": "The cartographic projection to use. This value is case-insensitive, for example `\"albers\"` and `\"Albers\"` indicate the same projection type. You can find all valid projection types [in the documentation](https://vega.github.io/vega-lite/docs/projection.html#projection-types).\n\n__Default value:__ `equalEarth`"
         }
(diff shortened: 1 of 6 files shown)


kanitw (approved):
`fix!` rather than `feat!`?

domoritz:
Why is it a fix? I consider this an enhancement (therefore feat!).

willium (approved):
Looks good! Would be nice to add an example that lets you try different projections with a drop down (as is available in Vega)

domoritz:
Something like this:

```json
{
  "width": 500,
  "height": 300,
  "params": [
    {
      "name": "projection",
      "value": "equalEarth",
      "bind": {
        "input": "select",
        "options": [
          "albers",
          "albersUsa",
          "azimuthalEqualArea",
          "azimuthalEquidistant",
          "conicConformal",
          "conicEqualArea",
          "conicEquidistant",
          "equalEarth",
          "equirectangular",
          "gnomonic",
          "mercator",
          "naturalEarth1",
          "orthographic",
          "stereographic",
          "transverseMercator"
        ]
      }
    }
  ],
  "data": {
    "url": "data/world-110m.json",
    "format": {"type": "topojson", "feature": "countries"}
  },
  "projection": {"type": {"expr": "projection"}},
  "mark": {"type": "geoshape", "fill": "lightgray", "stroke": "gray"}
}
```

Unfortunately, expr doesn't work here yet.

PR #7057, opened 2020-11-18
refactor: make expr replacement code stricter

domoritz:
* Remove `ExprOrSignalRef` type
* Write stricter output type for `replaceExprRefInIndex`
* Remove unused function

merge after https://github.com/vega/vega-lite/pull/7058


src/compile/model.ts
```diff
@@ -238,16 +238,6 @@ export abstract class Model {
       legends: {}
     };
   }
-  private initTitle(title: TitleParams<ExprOrSignalRef>) {
```
kanitw, line 241: Can you explain why we can remove this method?

domoritz: It's only used in the constructor and I replaced that call with a call to `replaceExprRef`, which seems to do the same thing.

Issue #6997, opened 2020-10-26
Point sizes on charts with binned data are NOT updating properly


jpkli:
[...]

PR #7063, opened 2020-11-19
feat: support getBinStepSignal for prebinned data

domoritz:
Fixes https://github.com/vega/vega-lite/issues/6997

### DIFF
--- a/src/compile/scale/range.ts
+++ b/src/compile/scale/range.ts
@@ -72,17 +72,27 @@ export function parseUnitScaleRange(model: UnitModel) {
 function getBinStepSignal(model: UnitModel, channel: 'x' | 'y'): SignalRefWrapper {
   const fieldDef = model.fieldDef(channel);
 
-  if (fieldDef && fieldDef.bin && isBinning(fieldDef.bin)) {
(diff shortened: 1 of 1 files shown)


willium (approved):
Pretty elegant solution! A bit worried it may be brittle to use signals like this... are there other places this pattern is used?

domoritz:
What do you mean by "this pattern"?

jpkli:
This seems did not fix #6997. The encoding field for the point size may not have the _bin_ params, but the position and size of the points are still depended on the _x_ and _y_ encodings that are using pre-binned data, so this PR cannot handle such cases. 

I think we probably need to add additional cases in the [_sizeRangeMax function_](https://github.com/vega/vega-lite/blob/master/src/compile/scale/range.ts#L382) to handle this. When both fields for the _x_ and _y_ channels are using pre-binned data, we need to generate the signal for the point and circle sizes.

domoritz:
Did you try it? I actually used your example and it worked.

jpkli:
Actually, it works! I think I didn't set up my vega editor correctly when I tried earlier today. Now I can see it did fix #6997. Thank you very much!

jpkli (approved):
Looks good. Thanks!

Issue #7073, opened 2020-11-30
Lookup transform should permit custom default values


marcprux:
Consider the following successful lookup:

<img width="95" alt="Screen Shot 2020-11-30 at 14 23 15" src="https://user-images.githubusercontent.com/659086/100654297-9743cf00-3317-11eb-8993-02239617c709.png">

```json
{
  "$schema" : "https://vega.github.io/schema/vega-lite/v4.17.0.json",
  "mark" : "text",
  "width" : "container",
  "height" : "container",
  "data" : {
    "name" : "lookup_groups"
  },
  "transform" : [
    {
      "from" : {
        "data" : {
          "name" : "lookup_people"
        },
        "fields" : [
          "age",
          "height"
        ],
        "key" : "name"
      },
      "lookup" : "person"
    },
    {
      "calculate": "['person=' + datum['person'], 'group=' + datum['group'], 'age=' + datum['age'], 'height=' + datum['height']]",
      "as": "summary"
    }
  ],
  "encoding" : {
    "text" : {
      "field" : "summary"
    }
  },
  "datasets" : {
    "lookup_groups" : [
      {
        "group" : 1,
        "person" : "Alan"
      }
    ],
    "lookup_people" : [
      {
        "age" : 25,
        "height" : 180,
        "name" : "Alan"
      }
    ]
  }
}
```

When the lookup fails, we can customize the default for the unmatched fields with the `default` parameter:

<img width="165" alt="Screen Shot 2020-11-30 at 14 25 15" src="https://user-images.githubusercontent.com/659086/100654490-df62f180-3317-11eb-9efa-f6cb9ae58893.png">

```json
{
  "$schema" : "https://vega.github.io/schema/vega-lite/v4.17.0.json",
  "mark" : "text",
  "width" : "container",
  "height" : "container",
  "data" : {
    "name" : "lookup_groups"
  },
  "transform" : [
    {
      "from" : {
        "data" : {
          "name" : "lookup_people"
        },
        "fields" : [
          "age",
          "height"
        ],
        "key" : "name"
      },
      "default": "DEFAULT_LOOKUP",
      "lookup" : "person"
    },
    {
      "calculate": "['person=' + datum['person'], 'group=' + datum['group'], 'age=' + datum['age'], 'height=' + datum['height']]",
      "as": "summary"
    }
  ],
  "encoding" : {
    "text" : {
      "field" : "summary"
    }
  },
  "datasets" : {
    "lookup_groups" : [
      {
[...]

PR #7074, opened 2020-11-30
fix: allow any value as default value in lookup transforms

domoritz:
Fixes https://github.com/vega/vega-lite/issues/7073


marcprux (approved):
LGTM!

PR #7114, opened 2020-12-14
feat: normalize old selection syntax for backwards compatibility

### LINKED ISSUE #6947: Add normalizer to support old selection syntax with new params syntax

arvind:
Resolves #6947.

### DIFF
--- a/src/normalize/selectioncompat.ts
+++ b/src/normalize/selectioncompat.ts
@@ -27,7 +27,9 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<NormalizerParam
       });
     }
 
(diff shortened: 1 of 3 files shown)


src/normalize/selectioncompat.ts
```diff
@@ -0,0 +1,33 @@
...
+
+      params.push({
+        name,
+        value: init,
+        select,
+        bind
+      });
+    }
+
+    if (params.length) spec.params = params;
```
domoritz, line 30:
```suggestion
    if (params.length) {
      spec.params = params;
    }
```

domoritz (approved):

PR #7115, opened 2020-12-14
fix: interactive legends should be toggleable by default

arvind:
This PR offers a cleaner approach for toggling interactive legend items. By default, direct manipulation is disabled for bound point selections. Under our previous approach, however, this combination would yield errant references to undefined signals. Under this proposed approach, toggling is also considered a direct manipulation operation that is disabled by default but then explicitly re-enabled (if needed) by the bind-legend selection operator.


src/compile/selection/legends.ts
```diff
@@ -24,7 +24,12 @@ const legendBindings: SelectionCompiler<'point'> = {
   },
 
   parse: (model, selCmpt, selDef) => {
-    disableDirectManipulation(selCmpt, selDef);
+    // Allow legend items to be toggleable by default even though direct manip is disabled.
```
domoritz, line 27:
```suggestion
    // Allow legend items to be toggleable by default even though direct manipulation is disabled.
```

domoritz (approved):

PR #7125, opened 2020-12-21
ci: avoid duplicate jobs

domoritz:
* run tests on prs or checks on push


domoritz:
Not sure why the TOC is being skipped.

Issue #7126, opened 2020-12-22
Unintuitive typescript API for boxplot


willcrichton:
I'm using react-vega and vega-lite to create a boxplot in my React/Typescript codebase. Here's the program I arrived at:

```typescript
import React from 'react';
import {VegaLite} from 'react-vega';
import {Mark} from 'vega-lite/build/src/mark';
import {BOXPLOT} from 'vega-lite/build/src/compositemark/boxplot';
import {Type} from 'vega-lite/build/src/type';

import {PROGRAM_DATA} from './data';

export let AnalysisRoute = () => {
  let spec = {
    mark: BOXPLOT,
    encoding: {
      x: {field: 'language', type: Type.nominal},
      y: {field: 'ntokens', type: Type.quantitative}
    },
    data: {name: 'data'}
  };

  return <div>
    <h2>Analysis</h2>
    <VegaLite spec={spec} data={{data: PROGRAM_DATA}} />
  </div>;
};
```

The issues I encountered:
* Import paths are weird and not documented. If it's not exported from `index.ts` then I have to search around for the appropriate place to import.
* Finding the right type to use is hard. I sadly can't just put `mark: "boxplot"`. I had to search through several files to find the right incantation of `BOXPLOT`.

These issues are partially API design, partially documentation (#5247). Re-exporting more utilities from `index.ts` and additional documentation can help.

domoritz:
You can use strings.

```ts
const spec: TopLevelSpec = {
  mark: 'boxplot',
  encoding: {
    x: {field: 'language', type: 'nominal'},
    y: {field: 'ntokens', type: 'quantitative'}
  },
  data: {name: 'data'}
};
```

I haven't run the code above but I think the problem is that Typescript infers `'boxplot'` as `string` when you create `spec`. To fix it, you can try to either set the type of `spec` to `TopLevelSpec` or append `as const`.

Let me know if that works. If yes, then I would say the issue is only documentation, not API design.

domoritz:
@willcrichton I just created https://github.com/vega/vega-lite/pull/7127 to add some docs to the website. Can you take a look and suggest improvements?

willcrichton:
Ah ok that was correct. The specific issue was that I'm using `react-vega` which goes through `vega-embed`. `vega-embed` expects a type `VisualizationSpec` which is either a Vega spec or a Vega-Lite spec. For whatever reason, TS assumes a Vega spec instead of a Vega-Lite spec by default, and hence the type error.

[...]

PR #7127, opened 2020-12-22
docs: add typescript docs

### LINKED ISSUE #5247: Docs for typescript users
Right now we have types exported, but may need to document how to start using Vega-Lite with Typescript.  (People sometimes get confused which Spec to import, etc.)

domoritz:
fixes #5247

### DIFF
--- a/site/usage/typescript.md
+++ b/site/usage/typescript.md
@@ -1,0 +1,46 @@
+---
+layout: usage
+menu: usage
+title: Vega-Lite and Typescript
+permalink: /usage/typescript.html
+---
+
+Vega-Lite is written in [TypeScript](https://www.typescriptlang.org). You can use the types provided by Vega-Lite to type check your specifications and the compiler API.
+
+The main function for Vega-Lite is [`compile`](compile.html). Vega-Lite directly exports this function. The first argument to `compile` is a [Vega-Lite specification]({{ site.baseurl }}/docs/spec.html) of type `TopLevelSpec`. You can use the type, which Vega-Lite directly exports, to type check your specification. Vega-Lite also exports a type `Config` for the [Vega-Lite configuration]({{ site.baseurl }}/docs/config.html).
+
+Below is a short code snippet that uses the `compile` function and the various types exported by Vega-Lite.
+
+```ts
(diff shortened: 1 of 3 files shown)


site/usage/typescript.md
```diff
@@ -0,0 +1,46 @@
+---
+layout: usage
+menu: usage
+title: Vega-Lite and Typescript
+permalink: /usage/typescript.html
+---
+
+Vega-Lite is written in [TypeScript](https://www.typescriptlang.org). You can use the types provided by Vega-Lite to type check your specifications and the compiler API.
+
+The main function for Vega-Lite is [`compile`](compile.html). Vega-Lite directly exports this function. The first argument to `compile` is a [Vega-Lite specification]({{ site.baseurl }}/docs/spec.html) of type `TopLevelSpec`. You can use the type, which Vega-Lite directly exports, to type check your specification. Vega-Lite also exports a type `Config` for the [Vega-Lite configuration]({{ site.baseurl }}/docs/config.html).
```
willcrichton, line 10: Re: the issue in #7126, it would be helpful to note that using wrappers on vega-lite like vega-embed need to have their types constrained to avoid inference errors. Maybe that documentation should go on vega-embed instead?

domoritz: I don't think that's the case. 

```ts
import {compile} from 'vega-lite';

const vegaLiteSpec = {
  $schema: 'https://vega.github.io/schema/vega-lite/v4.json',
  data: {
    values: [
      {a: 'A', b: 28},
      {a: 'B', b: 55},
      {a: 'C', b: 43},
      {a: 'D', b: 91},
      {a: 'E', b: 81},
      {a: 'F', b: 53},
      {a: 'G', b: 19},
      {a: 'H', b: 87},
      {a: 'I', b: 52}
    ]
  },
  mark: 'bar',
  encoding: {
    x: {field: 'a', type: 'nominal', axis: {labelAngle: 0}},
    y: {field: 'b', type: 'quantitative'}
  }
} as const;

const vegaSpec = compile(vegaLiteSpec).spec;
```

The issue without `as const` is that Typescript widens the type of `"bar"` to `string` rather than the literal `"bar"`.

PR #7141, opened 2021-01-03
refactor: rename `Parameter/SelectionDef` -> `Variable/SelectionParameter`

kanitw:
If we gonna make selection a kind of parameter, we should do this.

That said, as I try to write docs for this, I'm having second thoughts about this. 
I'm no longer liking this unification. (Thus, I'm making just a draft PR.)

### DIFF
--- a/src/selection.ts
+++ b/src/selection.ts
@@ -3,6 +3,7 @@ import {isObject} from 'vega-util';
 import {SingleDefUnitChannel} from './channel';
 import {FieldName, PrimitiveValue} from './channeldef';
 import {DateTime} from './datetime';
+import {ParameterBase} from './parameter';
 import {Dict} from './util';
 
 export const SELECTION_ID = '_vgsid_';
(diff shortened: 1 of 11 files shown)


domoritz (approved):

src/selection.ts
```diff
@@ -299,6 +295,6 @@ export function isLegendStreamBinding(bind: any): bind is LegendStreamBinding {
   return isLegendBinding(bind) && isObject(bind);
 }
 
-export function isParameterSelection(param: any): param is SelectionDef {
+export function isInteractiveParameter(param: any): param is SelectionParameter {
```
arvind, line 298: I'm not sure why this was changed? Variable parameters can also be interactive? Either the previous name or `isSelectionParameter` makes sense to me.

arvind (changes requested):

src/selection.ts
domoritz: I prefer `isSelectionParameter`.

arvind:
Superseded by #7172.

PR #7146, opened 2021-01-03
fix: labelBaseline cannot be null

kanitw:
This is what's failing the Vega validation in the master branch and in https://github.com/vega/vega-lite/pull/7144


domoritz:
The example doesn't look right anymore.

kanitw:
So null is actually significant, otherwise the "default" align isn't preserved.  I'm adding a new fix in https://github.com/vega/vega-lite/pull/7147

Issue #4371, opened 2018-12-15
A predicate to check whether a selection is empty or not


kanitw:
Our selection can include `empty: "all" | "none"`.  However, sometimes, we want to apply each of the modes to a different part of the same visualization.  

For example, in https://vega.github.io/vega-lite/examples/interactive_bar_select_highlight.html:

- We want the `fillOpacity` to activate the selection predicate both when there are selected data for the `select` selection and when there is none.  
- However, we want the strokeWidth to appear only when there are selected data for the `select` selection, but **not** when there is none.   Currently, I had to add a hack where I have to combine an expression with the selection predicate: 

```
           test": {
            "and": [
              {"selection": "select"},
              "length(data(\"select_store\"))"
            ]
          }
```

Note that the expression ` "length(data(\"select_store\"))"` is a hack as I know the underlying data store name in Vega.

It would be more natural to provide a syntax for this as a part of a [selection predicate](https://vega.github.io/vega-lite/docs/filter.html#selection-predicate) (and a [condition definition](
https://vega.github.io/vega-lite/docs/condition.html)).

The tricky part is that the new syntax that we can add here may be overriding the "empty" defined.

Some alternatives (given a selection named `select`):

### a) Add `{empty: "sel_name"}` to the selection logical chain

```json
{
  "condition": {"selection": {"or": ["select", {"empty": "select"}], "value": ...}
}
```

```json
{
  "condition": {"selection": {"and": ["select", {"not": {"empty": "select"}}], "value": ...}
}
```

### b) Add `{empty:  "sel_name"}` to the predicate chain

```json
{
  "condition": {"test": {"or": [{"selection": "select"}, {"empty": "select"}], "value": ...}
}
```

```json
{
  "condition": {"test": {"and": [{"selection": "select"}, {"not": {"empty": "select"}}], "value": ...}
}
```

### c) Add boolean empty to both the selection predicate and condition predicate

```js
{
  "condition": {
    // Don't trigger predicate if empty (equivalent to "and not empty") 
    "selection": "select", "empty": false,  
    "value": ...
  } 
}
```

```js
{
  "condition": {
    //Trigger predicate if empty (equivalent to "or empty")
    "selection": "select", "empty": true,  
    "value": ...
  } 
}
```

For this c) design  we can also add `"empty"` to the [selection predicate](https://vega.github.io/vega-lite/docs/filter.html#selection-predicate) too: ` {"selection": "select", "empty": false}`

-----

[...]

PR #7155, opened 2021-01-05
feat: move selection emptiness check to predicate

arvind:
Addresses #4371 — emptiness checks on the predicate allow us to get rid of the hacky `length(data("brush_store"))` hacks of yore. This PR, however, _moves_ the emptiness check to the selection predicate rather than allowing emptiness to be tested with the predicate in _addition_ to being defined on the selection as before (#3038). 

Although this approach can introduce some occasional duplication, I think it's preferable to have only one way of doing things and because keeping the selection-level emptiness definition would yield a somewhat confusing set of `"empty"` property values (`"all" | "none"` at the selection level, and then booleans at the predicate level).

Adding `empty` to the predicate level necessitates one other breaking change: selection predicates can only be composed alongside other predicate types (i.e., within a `condition.test` block or `filter`) rather than being composable within the `selection` property itself (i.e., we could previously do `"selection": {"and": ["foo", "bar"]}`).


examples/compiled/selection_type_interval_invert.vg.json
```diff
@@ -298,7 +298,7 @@
         "update": {
           "fill": [
             {
-              "test": "!(length(data(\"pts_store\"))) || (!(vlSelectionTest(\"pts_store\", datum)))",
+              "test": "!(!length(data(\"pts_store\")) || vlSelectionTest(\"pts_store\", datum))",
```
domoritz, line 301: This example is now different from before. Is that intended? 

The docs say "Notice, all marks are initially colored grey. This is because empty selections are treated as containing all data values." so I think either the example or the docs need to change.

domoritz (approved):
Looks good except for one issue with an example.

domoritz:
@kanitw since you filed https://github.com/vega/vega-lite/issues/4371, can you take a look?

arvind:
Yes, fixes #4371.

PR #7187, opened 2021-01-27
docs: update to reflect unified variable + selection parameters

arvind:
We used to use the word "parameter" throughout the docs, so I've replaced those in favor of "property" to avoid confusion.


domoritz:
Is this pull request supposed to be complete? I see a bunch of missing references and empty charts. I can help fix up some of them but want to check before I go into it.

arvind:
It should be complete. Eg [here’s the param-expr spec](https://github.com/vega/vega-lite/pull/7187/files#diff-1f1b374968aeb124e20287765dd4861fb49ae6287207f11551cfd86dd556b322) that’s 404ing in your screenshot. It was showing up correctly on my side, but I totally believe I set something up wrong, or didn’t put specs in the right place, or something similarly silly :D

domoritz:
I'll check again after I get some sleep to make sure it'n not my setup.

kanitw:
We probably need to re-think about what should show up under "parameters" on the left? 

Currently, the things that show up are only properties of selection parameters, which feels a bit weird if the header is parameter. 

image

kanitw:
- [ ] condition.md needs an update to match the current structure (it still refers to "selection")

domoritz:
> condition.md needs an update to match the current structure (it still refers to "selection")

working on that one

Issue #6994, opened 2020-10-25
Revise the band property


kanitw:
Currently it can mean either band size or band position.  Maybe we need to be clearer to avoid confusion.  

- [ ] rename `timeUnitBand` to `timeUnitBandSize`

kanitw:
Will be fixed in https://github.com/vega/vega-lite/pull/7190

PR #7190, opened 2021-01-28
feat!: split field's `band` into field's `bandPosition` & mark's `width/height: {band: ...}`

kanitw:
Also rename `config.mark.timeUnitBand` -> `timeUnitBandSize`

### DIFF
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -1,4 +1,4 @@
-import {Gradient, SignalRef, Text} from 'vega';
+import {Gradient, ScaleType, SignalRef, Text} from 'vega';
 import {isArray, isBoolean, isNumber, isString} from 'vega-util';
 import {Aggregate, isAggregateOp, isArgmaxDef, isArgminDef, isCountingAggregateOp} from './aggregate';
 import {Axis} from './axis';
@@ -14,6 +14,7 @@ import {
   FACET,
   FILL,
   FILLOPACITY,
+  getSizeChannel,
   HREF,
   isScaleChannel,
   isSecondaryRangeChannel,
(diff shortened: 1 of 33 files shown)


domoritz:
Fixes https://github.com/vega/vega-lite/issues/6994

src/channeldef.ts
```diff
@@ -472,14 +475,12 @@ export interface PositionBaseMixins {
 
 export interface BandMixins {
   /**
-   * For rect-based marks (`rect`, `bar`, and `image`), mark size relative to bandwidth of [band scales](https://vega.github.io/vega-lite/docs/scale.html#band), bins or time units. If set to `1`, the mark size is set to the bandwidth, the bin interval, or the time unit interval. If set to `0.5`, the mark size is half of the bandwidth or the time unit interval.
-   *
-   * For other marks, relative position on a band of a stacked, binned, time unit or band scale. If set to `0`, the marks will be positioned at the beginning of the band. If set to `0.5`, the marks will be positioned in the middle of the band.
+   * Relative position on a band of a stacked, binned, time unit or band scale. If set to `0`, the marks will be positioned at the beginning of the band. If set to `0.5`, the marks will be positioned in the middle of the band.
```
domoritz, line 478:
```suggestion
   * Relative position on a band of a stacked, binned, time unit, or band scale. If set to `0`, the marks will be positioned at the beginning of the band. If set to `0.5`, the marks will be positioned in the middle of the band, etc.
```

domoritz (approved):
LGTM

src/channeldef.ts
kanitw: I don't think we need etc at the end.

domoritz: Maybe say "and so on." then or start with "For example..."? When I read it I was surprised to only see two examples but the text makes it sound like this is all you can set.

kanitw:
Tests are all passing now.  Since a lot has changed since you approved, it might be worth another review pass, just in case.

kanitw:
oops it's still failing ...

domoritz (approved):
A bit more test coverage would be good.

PR #7196, opened 2021-01-31
docs: rename property types to property value types to reduce confusion with props

domoritz:


domoritz:
I don't know why I merged properties and params in my head. Ignore me.

PR #7204, opened 2021-02-02
fix: normalize selections across multi-views and selection emptiness

arvind:


arvind:
Hmm, looks like these changes are causing a lot more specs to be normalized even though there isn't any real change in the specification. Need to investigate why. @vega/vl-dev, would love any pointers to what I'm doing wrong.

arvind:
Ok, should be fixed.

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
   return bin;
 }
 
-function normalizePredicate(op: any) {
+function normalizePredicate(op: any, normParams: NormalizerParams) {
   // Normalize old compositions of selection names (e.g., selection: {and: ["one", "two"]})
   const normalizeSelectionComposition = (o: LogicalComposition<string>) => {
-    return normalizeLogicalComposition(o, param => ({param} as any));
+    return normalizeLogicalComposition(o, param => {
+      const empty = normParams.emptySelections[param] || true;
```
domoritz, line 146: Should this use `??`? What happens if `normParams.emptySelections[param]` returns false?

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
+      const pred = {param, empty};
+      normParams.selectionPredicates[param] = normParams.selectionPredicates[param] || [];
```
domoritz, line 148: same as above, why not `??`

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
 
-      params.push({
-        name,
-        value: init,
-        select,
-        bind
-      });
+          // Propagate emptiness forwards and backwards
+          normParams.emptySelections[name] = empty !== 'none';
+          for (const pred of Object.values(normParams.selectionPredicates[name] ?? {})) {
```
domoritz, line 58:
```suggestion
          for (const pred of vals(normParams.selectionPredicates[name] ?? {})) {
```

From util.

src/normalize/base.ts
```diff
@@ -43,5 +45,7 @@ export interface NormalizerParams {
   repeater?: RepeaterValue;
   repeaterPrefix?: string;
   selections?: TopLevelSelectionParameter[];
+  emptySelections?: Dict<boolean>;
```
domoritz, line 48: Should this be `Set` instead? Or can the value ever be `false`?

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
-      if (select.type === 'single') {
-        select.type = 'point';
-        select.toggle = false;
-      } else if (select.type === 'multi') {
-        select.type = 'point';
-      }
+    if (selection) {
+      return {
+        ...rest,
+        params: Object.entries(selection).map(([name, selDef]) => {
```
domoritz, line 47:
```suggestion
        params: entries(selection).map(([name, selDef]) => {
```

From util

src/normalize/selectioncompat.ts
domoritz: If it can never be false, we should use a `Set`.

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
+function normalizePredicate(op: any, normParams: NormalizerParams) {
   // Normalize old compositions of selection names (e.g., selection: {and: ["one", "two"]})
   const normalizeSelectionComposition = (o: LogicalComposition<string>) => {
-    return normalizeLogicalComposition(o, param => ({param} as any));
+    return normalizeLogicalComposition(o, param => {
+      const empty = normParams.emptySelections[param] || true;
+      const pred = {param, empty};
+      normParams.selectionPredicates[param] = normParams.selectionPredicates[param] || [];
+      normParams.selectionPredicates[param].push(pred);
+      return pred as any;
```
domoritz, line 150: Why do we need as any here?

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
-    return {...bin, extent: {...rest, param: selection}};
+    const {selection: param, ...rest} = ext;
+    return {...bin, extent: {...rest, param}};
   }
 
   return bin;
 }
 
-function normalizePredicate(op: any) {
+function normalizePredicate(op: any, normParams: NormalizerParams) {
```
domoritz, line 142: I know this is old code but why is op any?

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
-        ...normalizeChannelDef(enc),
-        ...(enc.condition
+    return spec;
+  }
+}
+
+function normalizeTransforms(spec: any, normParams: NormalizerParams) {
+  const {transform: tx, ...rest} = spec;
+  if (tx) {
+    const transform = tx.map((t: any) => {
```
domoritz, line 74: Why is t any?

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
+      if (isFilter(t)) {
+        return {filter: normalizePredicate(t, normParams)};
+      } else if (isBin(t) && isBinParams(t.bin)) {
+        return {
+          ...t,
+          bin: normalizeBinExtent(t.bin)
+        };
+      } else if (isLookup(t)) {
+        const {selection: param, ...from} = t.from as any;
```
domoritz, line 83: Why is t.from any?

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
+    spec = normalizeTransforms(spec, normParams);
+    return super.map(spec, normParams);
+  }
+
+  public mapLayerOrUnit(spec: FacetedUnitSpec<Field> | LayerSpec<Field>, normParams: NormalizerParams) {
+    spec = normalizeTransforms(spec, normParams);
+
+    if (spec.encoding) {
+      const encoding = {};
+      for (const [channel, enc] of Object.entries(spec.encoding)) {
```
domoritz, line 32:
```suggestion
      for (const [channel, enc] of entries(spec.encoding)) {
```

from util

src/normalize/selectioncompat.ts
```diff
@@ -13,105 +14,141 @@ export class SelectionCompatibilityNormalizer extends SpecMapper<
...
+      spec = {...spec, encoding};
     }
 
-    return super.map(spec, params);
+    return super.mapLayerOrUnit(spec, normParams);
   }
 
-  public mapUnit(spec: UnitSpec<Field>) {
+  public mapUnit(spec: UnitSpec<Field>, normParams: NormalizerParams) {
     const {selection, ...rest} = spec as any;
```
domoritz, line 43: This is old code but why does spec need to be any?

domoritz (approved):

src/normalize/base.ts
arvind: It can be false, yes.

PR #7208, opened 2021-02-05
fix: smoothen `value` gradient for var param -> point selection

### LINKED ISSUE #7206: Smooth the "value"  for selection

arvind:
Fixes #7206

### DIFF
--- a/src/compile/selection/project.ts
+++ b/src/compile/selection/project.ts
@@ -63,6 +63,7 @@ const project: SelectionCompiler = {
     };
 
     const type = selCmpt.type;
+    const cfg = model.config.selection[type];
     const init =
       selDef.value !== undefined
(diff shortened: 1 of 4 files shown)


src/compile/selection/project.ts
```diff
@@ -161,7 +166,9 @@ const project: SelectionCompiler = {
 
     if (init) {
       selCmpt.init = (init as any).map((v: SelectionInitMapping | SelectionInitIntervalMapping) => {
-        return proj.items.map(p => (v[p.channel] !== undefined ? v[p.channel] : v[p.field]));
+        // selections can be initialized either with a full object that maps projections to values
```
domoritz, line 169:
```suggestion
        // Selections can be initialized either with a full object that maps projections to values
```

domoritz (approved):

PR #7226, opened 2021-02-10
docs: revise contributing guide and add list of design principles

domoritz:


CONTRIBUTING.md
```diff
@@ -14,13 +14,20 @@ If you find a bug in the code or a mistake in the [documentation](https://vega.g
 
 See our [issue](.github/ISSUE_TEMPLATE.md) and [pull request](.github/PULL_REQUEST_TEMPLATE.md) templates for more information.
 
-### Looking for a Task to Contribute
+## Design principles
 
-You can find [tasks with the "Good first issue" label in the issue tracker :pray:](https://github.com/vega/vega-lite/labels/Good%20first%20issue%20%3Ababy%3A). Please add a comment in issues if you are planning to work on a major task.
+Vega-Lite is a compiler and a declarative language to describing interactive multi view graphics. As a compiler, Vega-Lite compiles to the lower-level [Vega](https://vega.github.io/vega/) specifications. The Vega-Lite project inherits many of the design principles of Vega, especially the declarative design. In addition, the development of Vega-Lite follows these principles that we have established over the years.
+
+- **Smart defaults.** Any property that is not specified should have a (somewhat obvious) default value, which users can override. For example, Vega-Lite automatically synthesizes scales and guides (axes and legends).
```
kanitw, line 21: better say sensible default?

CONTRIBUTING.md
```diff
@@ -14,13 +14,20 @@ If you find a bug in the code or a mistake in the [documentation](https://vega.g
...
+- **Remain backwards compatible when possible.** Even if we change a major feature, we aim to support the old syntax. However, the old syntax may not be supported by the JSON schema.
```
kanitw, line 22: Is this really true? Do we want to keep the normalizer for old selection forever? (Probably not?))

CONTRIBUTING.md
```diff
@@ -14,13 +14,20 @@ If you find a bug in the code or a mistake in the [documentation](https://vega.g
...
+- **Backwards compatibility concerns the input, not the output.** Just like Vega, we may change how a declarative specification is interpreted.
+- **Vega-Lite never needs to see the data.** All decisions about how to compile to Vega have to be made without access to the data that will be used in the chart. This principle enables the compiled Vega charts to work with any dataset that follows the same schema (field names and types).
+- **Generate readable Vega.** We aim to generate readable Vega specifications that do not contain unnecessary properties (e.g. Vega's defaults).
```
kanitw, line 25: This is also not really *always* true. It's more like generate readable Vega when possible.

CONTRIBUTING.md
```diff
@@ -14,13 +14,20 @@ If you find a bug in the code or a mistake in the [documentation](https://vega.g
 
 See our [issue](.github/ISSUE_TEMPLATE.md) and [pull request](.github/PULL_REQUEST_TEMPLATE.md) templates for more information.
 
-### Looking for a Task to Contribute
+## Design principles
```
kanitw, line 17: Some of these items you listed down here feels more like contract to users (regarding versioning)) more than design principles. 

What we had in https://medium.com/hci-design-at-uw/introducing-vega-lite-438f9215f09e feels more like the genuine design principles of the language to me.

kanitw (changes requested):

CONTRIBUTING.md
domoritz: Done

CONTRIBUTING.md
domoritz: But probably until Vega-Lite 6. Backward compatibility is really important for JupyterLab so I think saying it here is a good signal that we care.

CONTRIBUTING.md
domoritz: That's why I say we **aim** to generate readable Vega.

CONTRIBUTING.md
domoritz: I think a mix of both are good in the contributing guidelines. Maybe design principles is not as good as a term as development principles or development goals.

domoritz: I added "Favor composition over templates" as it was missing. 

"Leverage Vega’s performance, flexibility across platforms, and expressivity." doesn't make too much sense in this context, I think.

domoritz:
I made some changes. What do you think now @kanitw?

CONTRIBUTING.md
kanitw: But the header is still misleading.

CONTRIBUTING.md
```diff
@@ -14,13 +14,22 @@ If you find a bug in the code or a mistake in the [documentation](https://vega.g
...
+
+- **Provide sensible defaults, but allow customization.** Any property that is not specified should have a (somewhat obvious) default value, which users can override. For example, Vega-Lite automatically synthesizes scales and guides (axes and legends).
+- **Favor composition over templates.** While chart templates (as found in spreadsheet programs) can be convenient, they limit the visualization types that can be created. Instead, Vega-Lite uses a compositional approach, describing a visualization based on the properties of graphical marks. Any new building block should be compatible with the existing building blocks.
+- **Support gradual specification.** Most people write specifications incrementally, making one atom change at a time such as changing a property. Any intermediate step should be valid and produce some visualization.
```
arvind, line 23:
```suggestion
- **Support gradual specification.** Most people write specifications incrementally, making one atomic change at a time such as changing a single property. Any intermediate step should be valid and produce some visualization.
```

CONTRIBUTING.md
domoritz: Added

CONTRIBUTING.md
domoritz: Added

domoritz:
We can iterate on this in master.

PR #7138, opened 2021-01-03
docs: docs for parameter

kanitw:
No matter what we plan to do for 5.0, this is useful for pre 5.0.

PR #7230, opened 2021-02-12
docs: various docs fixes

domoritz:


site/docs/params/params.md
```diff
@@ -2,7 +2,7 @@
 layout: docs
 menu: docs
 title: Parameter / Interaction
-permalink: /docs/parameter.html
+permalink: /docs/params.html
```
domoritz, line 5: @arvind we had to pages at the same url. Where should this page go? It's not linked from anywhere.

arvind: It looks like this page was created in #7138, and merged before we had resolved the discussion about whether (and how) or not to unify params and selections. And, because it was merged into master, I didn’t have access to it in the selection-params feature branch (ergo the duplicate docs). The content below describing example param definitions should be merged into parameter.md. Will try to send a PR later today if no one gets to it by then.

domoritz: I just merged them.

arvind: 🙌

site/docs/parameter/parameter.md
```diff
@@ -37,10 +37,18 @@ Both variable and selection parameters support the following properties:
 
 ### Variable Parameters
 
-Variables are the simplest form a parameter can take. Variable parameters allow for a value to be defined once and then reused throughout the rest of the specification. As the name suggests, variable values can be modified dynamically either by [binding the variable](bind.html) to input widgets (e.g., sliders or drop down menus) or by specifying its `expr` property:
+Variables are the simplest form a parameter can take. Variable parameters allow for a value to be defined once and then reused throughout the rest of the specification. As the name suggests, variable values can be modified dynamically either by [binding the variable](bind.html) to input widgets (e.g., sliders or drop down menus), by specifying its `expr` property, or externally as a signal through the [Vega view API](https://vega.github.io/vega/docs/api/view/#signals):
```
kanitw, line 40: > externally as a signal through the [Vega view API]

This seems quite a leap. We should at least include an example?

kanitw: > by specifying its `expr` property

I think "modified" feels like a side effect, so binding and set via view  APIs feels like a match for "modified", but I think expr feels like more like declaring reactive dependencies.

src/parameter.ts
```diff
@@ -9,11 +9,6 @@ export interface VariableParameter {
    */
   name: ParameterName;
 
-  /**
```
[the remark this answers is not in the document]
domoritz: Yep. We didn't use it for any other param either.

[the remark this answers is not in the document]
domoritz: We briefly talked about it on Slack.

kanitw (approved):
@domoritz  if my further edit is good we can merge this.

PR #7256, opened 2021-02-23
docs: consolidate selection docs to improve information architecture of param docs

kanitw:
Fixes #7427


domoritz (approved):

arvind (approved):
LGTM. The only one I'm not sure about is `selection-value.md`. The property (`value`) is applicable to both variable and selection parameters, and has the same function for both. The page also talks about variable params (but doesn't include an example). I'm fine either way, but I just wanted to flag this.

PR #7263, opened 2021-02-24
feat: use vega-expression and vega-event-selector from vega

domoritz:
Bundle size goes from 768k to 713k (272k to 247k minified). 

Needs https://github.com/vega/vega/pull/3095 and means we always need Vega installed (which we had as a peer dependency already anyway).


domoritz:
Actually, would be better with https://github.com/vega/vega/pull/3154 since we then can still not rely on Vega.

domoritz:
Closing in favor of https://github.com/vega/vega-lite/pull/7718

Issue #7264, opened 2021-02-24
Expr does not work for legend title


mattijn:
I've the following chart that renders correctly, even though it issues a schema validation warning.
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "data": {
    "url": "https://vega.github.io/vega-lite/data/us-10m.json",
    "format": {"type": "topojson", "feature": "states"}
  },
  "params": [
    {
      "name": "selection",
      "select": {"type": "point", "fields": ["key"]},
      "value": [{"key": "hurricanes"}],
      "bind": {
        "input": "select",
        "options": ["population", "engineers", "hurricanes"]
      }
    }
  ],
  "transform": [
    {
      "lookup": "id",
      "from": {
        "data": {"url": "data/population_engineers_hurricanes.csv"},
        "key": "id",
        "fields": ["population", "engineers", "hurricanes"]
      }
    },
    {"fold": ["population", "engineers", "hurricanes"]},
    {"filter": {"param": "selection"}}
  ],
  "projection": {"type": "albersUsa"},
  "mark": "geoshape",
  "encoding": {
    "color": {
      "field": "value",
      "type": "quantitative",
      "legend": {"title": {"signal": "selection_key"}}
    }
  }
}
```

![signal_legend](https://user-images.githubusercontent.com/5186265/109069951-321f3000-76f2-11eb-948c-e6944773050e.gif)
The schema validation comes from this line:
```json
"legend": {"title": {"signal": "selection_key"}}
```

While reading the new docs (congrats again!) I have the feeling that I should be able to replace the `signal` with an `expr`, as such:
```json
"legend": {"title": {"expr": "selection"}}
``` 
but then the legend title does not render anymore. 
![expr_legend](https://user-images.githubusercontent.com/5186265/109071802-79a6bb80-76f4-11eb-8784-313a711ae5d7.gif)

Is this something that could/should be supported with an `expr`?

* * * 

[...]

PR #7265, opened 2021-02-25
feat: support expr in guide titles

domoritz:
Fixes https://github.com/vega/vega-lite/issues/7264

### DIFF
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -53,7 +53,7 @@ import {CompositeAggregate} from './compositemark';
 import {Config} from './config';
 import {DateTime, dateTimeToExpr, isDateTime} from './datetime';
 import {Encoding} from './encoding';
-import {ExprRef, isExprRef} from './expr';
+import {ExprRef, isExprRef, replaceExprRef} from './expr';
 import {Guide, GuideEncodingConditionalValueDef, TitleMixins} from './guide';
 import {ImputeParams} from './impute';
 import {Legend} from './legend';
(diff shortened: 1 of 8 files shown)


src/channeldef.ts
```diff
@@ -943,11 +943,11 @@ export function title(
   }
 }
 
-export function getGuide(fieldDef: TypedFieldDef<string> | SecondaryFieldDef<string> | DatumDef): Guide {
+export function getGuide(fieldDef: TypedFieldDef<string> | SecondaryFieldDef<string> | DatumDef): Guide<SignalRef> {
   if (isPositionFieldOrDatumDef(fieldDef) && fieldDef.axis) {
-    return fieldDef.axis;
+    return replaceExprRef(fieldDef.axis);
```
domoritz, line 948: @kanitw is this the right place to replace the expressway ref?

domoritz:
Ping @kanitw @arvind

src/channeldef.ts
kanitw: I searched a bit and replaceExprRef seems to happen consistently during `init()`.

A related problem is that this getGuide is now like anti-pattern because it was written before we make expr/signal generic type of axis/legend (but not FieldDef).

domoritz: I'm not following. What is the anti pattern here?

src/channeldef.ts
```diff
@@ -943,11 +943,11 @@ export function title(
   }
 }
 
-export function getGuide(fieldDef: TypedFieldDef<string> | SecondaryFieldDef<string> | DatumDef): Guide {
+export function getGuide(fieldDef: TypedFieldDef<string> | SecondaryFieldDef<string> | DatumDef): Guide<SignalRef> {
```
kanitw, line 946: `getGuide` is called in general utility function, not just inside init(). 

How do we know that this is `Guide<SignalRef>`, not `Guide<Expr>`?

kanitw (changes requested):
I think we might need to do a bigger refactor for getGuide?

src/channeldef.ts
domoritz: I guess if we call this function after init, the guide should only have Expr refs, no?

domoritz:
Done entirely in #7438

PR #7282, opened 2021-03-05
docs: fix links to outdated selection docs

### LINKED ISSUE #7281: V5 Schema requires list for a selection parameter value and contains broken links
The schema, as illustrated with this [query widget example](https://vega.github.io/vega-lite/examples/interactive_query_widgets.html), appears to require that the initial value of the selection specified via `value`, is an array of objects rather than a single object. Not sure if this is a bug or by design, but I can't see the rationale for requiring an array here.

The schema also references  now out of date (and broken) https://vega.github.io/vega-lite/docs/project.html and https://vega.github.io/vega-lite/docs/init.html pages.

domoritz:
fixes #7281


site/_layouts/docs.html
```diff
@@ -2,176 +2,176 @@
 layout: page_submenu
 
 sidebar:
-- text: Overview
```
arvind, line 5: Not a big deal, but these changes are curious. As far as I can tell, the only thing that changed is the tab spacing. But I'm not sure why/what caused it?

arvind (approved):

site/_layouts/docs.html
domoritz: I set my text editor to auto format with prettier. I made one change

```
  - text: Value
    url: selection-value
```

became just

```
  - text: Value
    url: value
```

sorry for the noise in the pull request.

PR #7296, opened 2021-03-10
feat: support images on maps

### LINKED ISSUE #6580: Image on a map: NaN width/height
When creating a map using Vega-Lite and circle markers, all is well (very pretty!). After changing the marker type to "image", adding "width" and "height" values, as well as adding a "url" encoding, it appears the images that are put on the map do not receive proper width/height. 

The same usage of images appears to work fine on regular (non-geo) charts. To reproduce, paste in the online editor:

```json
{
  "layer": [
    {
      "mark": {
        "type": "geoshape",
        "fill": "#ddd",
        "stroke": "#fff",
        "strokeWidth": 1
      },
      "data": {
        "url": "data/us-10m.json",
        "format": {"type": "topojson", "feature": "states"}
      }
    },
   
    {
      "mark": {"type": "image", "width":10, "height":10},
      "data": {"url": "data/airports.csv"},
      "encoding": {
        "latitude": {"field": "latitude", "type": "quantitative"},
        "longitude": {"field": "longitude", "type": "quantitative"},
        "url": {"datum": "data/ffox.png", "type": "nominal"}
      }
    }
  ],
  "projection": {"type": "albersUsa"},
  "width": 900,
  "height": 500,
  "config": {"view": {"stroke": null}}
}
````

Result in (most recent) Chrome:

image

image

### PR DESCRIPTION
**Pull request by domoritz**
fixes #6580

### DIFF
--- a/src/compile/mark/encode/position-range.ts
+++ b/src/compile/mark/encode/position-range.ts
@@ -12,7 +12,8 @@ import {pointPosition, pointPositionDefaultRef} from './position-point';
 import * as ref from './valueref';
 
 /**
- * Utility for area/rule position, which can be either point or range. (One of the axes should be point and the other should be range.)
+ * Utility for area/rule position, which can be either point or range.
+ * (One of the axes should be point and the other should be range.)
  */
 export function pointOrRangePosition(
   channel: 'x' | 'y',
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/compile/mark/encode/position-range.ts [hunk 1]
```diff
@@ -88,7 +89,18 @@ function pointPosition2OrSize(
...
-    return {[vgChannel]: {field: model.getName(channel)}};
+    const vgSizeChannel = getSizeChannel(channel);
+
+    const size = model.markDef[vgSizeChannel];
+    if (size != null) {
+      return {
+        [vgSizeChannel]: {value: size}
+      };
+    } else {
+      return {
```
kanitw, line 100: This case not covered by test. 

It's probably useful to update the comment since the old comment above is outdated (now there are two subcases)

> // use geopoint output if there are lat2/long2 and there is no point position2 overriding lat2/long2.

domoritz: Done

domoritz:
image

PR #7309, opened 2021-03-15
docs(example): Normalized (Percentage) Stacked Bar Chart With Labels

kanitw:


domoritz:
Doesn't look right

image

kanitw:
It seems correct now after rebasing

domoritz:
Wanna change the text color to white? Otherwise lgtm.

PR #7310, opened 2021-03-15
docs: example non-linear histogram

kanitw:
Thanks @Saba9


examples/compiled/histogram_nonlinear.vg.json
```diff
@@ -0,0 +1,120 @@
+{
+  "$schema": "https://vega.github.io/schema/vega/v5.json",
+  "description": "A histogram for a data source that provides non-linear bins. Thanks to [@Saba9](https://github.com/Saba9) who helps create this example.",
```
domoritz, line 3:
```suggestion
  "description": "A histogram for a data source that provides non-linear bins. Thanks to [@Saba9](https://github.com/Saba9) who helped create this example.",
```

examples/specs/histogram_nonlinear.vl.json
```diff
@@ -0,0 +1,72 @@
...
+      },
+      {
+        "startTime": "8.33",
+        "endTime": "12.50",
+        "residency": 0
+      },
+      {
+        "startTime": "12.50",
+        "endTime": "16.67",
+        "residency": 31.17
```
domoritz, line 19: If you convert this to actual ratios, we can use the format `.2f%` below.

examples/specs/histogram_nonlinear.vl.json
```diff
@@ -0,0 +1,72 @@
...
+      "sort": null
+    },
+    "x2": {"field": "endTime"},
+    "y": {
+      "title": "",
+      "field": "residency",
+      "type": "quantitative",
+      "scale": {"domain": [0, 100]},
+      "axis": {
+        "labelExpr": "datum.label + '%'"
```
domoritz, line 68: Switch to using format `.2f%`

domoritz (changes requested):

examples/specs/histogram_nonlinear.vl.json
kanitw: what do you mean?

examples/specs/histogram_nonlinear.vl.json
kanitw: There are cases where input data is already in percentage, so I don't think I agree we need to do this here.  I'll merge for now.  If you feel strongly about this, feel free to submit a PR.

Issue #6912, opened 2020-09-29
Support strict null checks for exported interfaces


domoritz:

domoritz:
https://github.com/vega/vega-embed uses strict null checks and it seems to work. Please report specific types exported that don't work with strict null checks on.

PR #7352, opened 2021-04-02
fix: Remove <null> from Value to fix TopLevelSpec consumed by code using strict mode.

walterra:
Related to #6912.

When you consume `TopLevelSpec` like `import type { TopLevelSpec } from 'vega-lite';` in a project with `strictNullChecks` set to `true`, you'll run into the following error:

```
ERROR
      node_modules/vega-lite/build/src/vega.schema.d.ts:34:19 - error TS2344: Type 'null' does not satisfy the constraint 'SignalRef | ExprRef'.

      34     value?: Value<null>;
                           ~~~~

      Found 1 error.
```

The problem seems to be that the generic `ES` for the `Value` type doesn't include `null`. Changing that type to `Value<ES extends ExprRef | SignalRef | null = ExprRef | SignalRef | null>` fixes it.

An alternative would be to remove the `<null>` from the code that references `Value` since the type definition includes `ExprRef | SignalRef` as a default anyway.

Let me know which version would make more sense and I'll update the PR.

### DIFF
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -90,7 +90,7 @@ import {isSignalRef} from './vega.schema';
 
 export type PrimitiveValue = number | string | boolean | null;
 
-export type Value<ES extends ExprRef | SignalRef | null = ExprRef | SignalRef | null> =
+export type Value<ES extends ExprRef | SignalRef = ExprRef | SignalRef> =
   | PrimitiveValue
   | number[]
   | Gradient
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -90,7 +90,7 @@ import {isSignalRef} from './vega.schema';
 
 export type PrimitiveValue = number | string | boolean | null;
 
-export type Value<ES extends ExprRef | SignalRef = ExprRef | SignalRef> =
+export type Value<ES extends ExprRef | SignalRef | null = ExprRef | SignalRef | null> =
   | PrimitiveValue
   | number[]
   | Gradient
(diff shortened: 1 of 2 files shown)


domoritz:
I think removing the null is cleaner.

Can you check why we have null there?

walterra:
Great, pushed an update that just removes the `<null>` part. I know too little about the history of the code to know why it's there. The type for `Value` will allow `null` because it's part of the `PrimitiveValue` union type bit so it should be fine.

walterra:
It was introduced here: https://github.com/vega/vega-lite/commit/b404b86de7a23e7ac3d330751d8b4ab83d5396a4

Looking at that change, I think it's safe to remove the `<null>` part.

domoritz:
Can you rebuild the schema `yarn schema`?

walterra:
Just did that, but it didn't produce any changes I could commit. Is there a way to trigger the tests again?

domoritz:
> Just did that, but it didn't produce any changes I could commit. Is there a way to trigger the tests again?

Great if there are no changes! Before and after, is there any externally observable change? Can you assign a value that could not be assigned before or vice versa?

To run the test in your fork, you may need to add a github token. Buf it there are no changes, then we should be good. 

Let me know and I will review the changes carefully.

domoritz:
I copied this pull request to https://github.com/vega/vega-lite/pull/7357

walterra:
Thanks for merging and doing the follow up!

walterra:
Do you have a usual release schedule? I'm just wondering if you can estimate when this fix will make it into a release? Thanks!

domoritz:
We do not and I want to get a few more things into the 5.1 release. I expect a release in the next two weeks, though.

walterra:
Sounds great, thanks!

PR #6672, opened 2020-06-22
feat: support autosize fit for faceted plots

drgould:
Based on discussion in #5219 and stealing some code from #5225, this adds support for specifying the outer plot dimensions and sizing the inner facets automatically which subsequently allows unlocking `autosize: fit(-x/-y)` for faceted plots.

Note: this is only for specs which explicitly define the facet operator at the top level.

- [x] Add tests
- [x] Update docs

PR #7356, opened 2021-04-04
feat: support autosize fit for faceted plots

domoritz:
Copy of https://github.com/vega/vega-lite/pull/6672 on original repo.


domoritz:
This needs to be picked up by someone else.

PR #7357, opened 2021-04-04
fix: remove <null> from Value to fix TopLevelSpec consumed by code using strict mode

domoritz:
Copy of https://github.com/vega/vega-lite/pull/7352

### DIFF
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -90,7 +90,7 @@ import {isSignalRef} from './vega.schema';
 
 export type PrimitiveValue = number | string | boolean | null;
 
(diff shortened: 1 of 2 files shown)


walterra:
Looks like all tests passed! 🥳 

> Great if there are no changes! Before and after, is there any externally observable change? Can you assign a value that could not be assigned before or vice versa?

The generated spec didn't change and since `null` was already part of the allowed values as part of `PrimitiveValue` there should also be no change in how to use the TS interface.

domoritz:
That is not correct.

```ts
const v: VgValueRef = {
  value: {signal: 'string'}
};
```

used to not be allowed and now is

Issue #7228, opened 2021-02-11
Add automatic dodge / repel of overlapping text marks


joelostblom:
When using text marks, the text labels can overlap so that they are impossible to read. This is prohibitive both for annotation of individual points and direct labelling of e.g. categorical lineplots, which can be preferred to using legends. There is an R package that solves this called [ggrepel](https://ggrepel.slowkow.com/) by using a force direct layout of the labels and it would be great with something similar in vega-lite, possibly building upon the [vega code used to lay out plots in the bee swarm plot example](https://vega.github.io/vega/examples/beeswarm-plot/)?

Example of what I mean (from ggrepel)

![image](https://imgur.com/ii9ova8.gif)

Related to https://github.com/vega/vega-lite/issues/5637 and also somewhat to https://github.com/vega/vega-lite/issues/3421

---

Please:

- [x] Check for duplicate issues. Please file separate requests as separate issues on GitHub.
- [x] Describe the feature's goal, motivating use cases, and its expected behavior.
- [] ~If you are proposing a new syntax, please provide at least one example spec, wrapped by triple backticks like this:~

You are encouraged to prototype multiple alternative syntaxes for your proposed feature. Doing so often leads to a better design.

- [x] If applicable, include screenshots, GIF videos (e.g. using https://www.cockos.com/licecap/), or working example (e.g. example Vega specification for the requested feature)

jcmkk3:
I think that you’re in luck. [vega-label](https://github.com/vega/vega-label) sounds like what you want. It was merged into vega a little while back and it looks like @chanwutk is working on a PR (#7222) to bring it into vega-lite.

joelostblom:
That's great! Thanks for pointing me to that PR!

nholland20:
Is this actually implemented in Vega-lite? I'm not seeing documentation on how to use the repel feature for labels? I only see it in vega.

domoritz:
See https://github.com/vega/vega-lite/pull/7378 by @chanwutk

jonfroehlich:
Just a friendly bump on this as well. Note how both the Los Angeles Clippers/Lakers overlap and so do the New York Knicks and Brooklyn Nets. Would be nice if there was an elegant way to deal with this. Sounds like https://github.com/vega/vega-lite/pull/7378 is perfect.

![image](https://github.com/vega/vega-lite/assets/1621749/e7cffd2b-a7ff-41f5-a1a6-ee190c61ca7f)

[...]

PR #7378, opened 2021-04-11
feat: add label encoding

kanitw:
Redo of #7222 so we have a proper review.


src/channeldef.ts
```diff
@@ -198,6 +199,72 @@ export type TextDef<F extends Field> =
...
+   * __Default value:__ `0` (only avoid the base mark)
+   */
+  avoidParentLayer?: 'all' | number;
+
+  /**
+   * Mark definition of the label text mark without the `'type'` property.
+   *
+   * __See also:__ [`text` mark](https://vega.github.io/vega-lite/docs/text.html#properties)
+   */
+  mark?: Omit<MarkDef<'text'>, 'type'>;
```
kanitw, line 240: We should declare a different name here. Otherwise, it'll show up as is type in the schema, which will be bad for Altair.

src/channeldef.ts
```diff
@@ -198,6 +199,72 @@ export type TextDef<F extends Field> =
...
+   * The layers of marks that the labels will avoid.
+   * `'all'` means all the marks that are children of the root layer.
+   * number `n` means all the marks that are children of the nth ancestor layer.
+   *
+   * __Default value:__ `0` (only avoid the base mark)
+   */
+  avoidParentLayer?: 'all' | number;
+
+  /**
+   * Mark definition of the label text mark without the `'type'` property.
```
kanitw, line 236: Just `type` in fine.  No need to do quotes for property names.

src/channeldef.ts
```diff
@@ -198,6 +199,72 @@ export type TextDef<F extends Field> =
...
+  position?: LabelPosition[];
+
+  /**
+   * The layers of marks that the labels will avoid.
+   * `'all'` means all the marks that are children of the root layer.
+   * number `n` means all the marks that are children of the nth ancestor layer.
+   *
+   * __Default value:__ `0` (only avoid the base mark)
+   */
+  avoidParentLayer?: 'all' | number;
```
kanitw, line 233: I think this API sounds confusing.  

As it is described, this sounds more like avoidAncestorLayer (not parent, since parent should mean `1` only).

src/channeldef.ts
```diff
@@ -198,6 +199,72 @@ export type TextDef<F extends Field> =
...
+  mark?: Omit<MarkDef<'text'>, 'type'>;
+
+  /**
+   * The padding in pixels by which a label may extend past the chart bounding box.
+   */
+  padding?: number;
+
+  /**
+   * The labeling method to use for area marks.
+   * One of `'floodfill'`, `'reduced-search'`, and `'naive'`.
```
kanitw, line 249: I think we need to explain more what these options means and when should they use them?

src/channeldef.ts
```diff
@@ -198,6 +199,72 @@ export type TextDef<F extends Field> =
...
+   * This property only applies when the base mark is a group mark containing area marks.
+   *
+   * __Default value:__ `'reduced-search'`
+   */
+  method?: AreaLabelMethod;
+
+  /**
+   * The anchor position of labels for line marks, where one line receives one label.
+   * One of `'start'` or `'end'` (the default).
+   * This property only applies when the base mark is a group mark containing line marks.
```
kanitw, line 259: > when the base mark is a group mark containing line marks.

This doesn't sound sensible for Vega-Lite?

src/channeldef.ts
```diff
@@ -198,6 +199,72 @@ export type TextDef<F extends Field> =
...
+export type LabelPosition = {
+  /**
+   * distance of a label from its base mark.
+   * if negative, the label position is insize.
+   */
+  offset: number;
+
+  /**
+   * anchor direction of a label from its base mark.
+   * The value could be one of `'left '` , `'right '` , `'top '` , `'middle '` ,
```
kanitw, line 214: Please use double quotation marks ("), not single (') for all docs.

examples/compiled/bar_corner_radius_end.vg.json
```diff
@@ -49,6 +49,7 @@
   "marks": [
     {
       "type": "group",
+      "name": "stackgroup",
```
kanitw, line 52: Why do we need to add all of these "stackgroup"?

src/channeldef.ts
```diff
@@ -198,6 +199,72 @@ export type TextDef<F extends Field> =
   | FieldOrDatumDefWithCondition<StringDatumDef<F>, Text>
   | ValueDefWithCondition<StringFieldDef<F>, Text>;
 
+/**
+ * Position of a label relative to its base mark's position, which includes anchor direction (anchor) and distance (offset).
+ */
+export type LabelPosition = {
+  /**
+   * distance of a label from its base mark.
```
kanitw, line 207: Distance -> D should be a capital letter.   
Please revise other docs comment so that all sentences begin with a capital letter.

src/channeldef.ts
```diff
@@ -198,6 +199,72 @@ export type TextDef<F extends Field> =
...
+export type LabelDefMixins = {
+  /**
+   * A list of possible positions for a label to be placed relative to its base mark.
+   */
+  position?: LabelPosition[];
+
+  /**
+   * The layers of marks that the labels will avoid.
+   * `'all'` means all the marks that are children of the root layer.
+   * number `n` means all the marks that are children of the nth ancestor layer.
```
kanitw, line 229: Number

src/channeldef.ts
kanitw: What about just `avoid: 'all-marks' | 'base-mark'`?

(I wonder if we really need the children of the nth ancestor because that sounds confusing.)

src/compile/layer.ts
```diff
@@ -120,17 +129,32 @@ export class LayerModel extends Model {
...
   public assembleMarks(): any[] {
-    return assembleLayerSelectionMarks(
+    const marks = assembleLayerSelectionMarks(
       this,
       this.children.flatMap(child => {
         return child.assembleMarks();
       })
     );
+
+    // move label marks to the top
```
kanitw, line 139: Can you explain more why we need to move label marks to the top here?

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
   ];
 }
 
+export function getLabel(model: UnitModel, data: string): LabelMark[] {
+  if (!model.encoding.label) {
+    return [] as LabelMark[];
```
kanitw, line 372: Why do we need `as LabelMark[]` here?

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
...
+  }
+
+  const {mark} = model;
+  if (!supportLabel(mark)) {
+    log.warn(log.message.dropChannelOnMark(mark, 'label'));
+    return [] as LabelMark[];
```
kanitw, line 378: Why do we need `as LabelMark[]` here?

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
...
+  }
+
+  const {label} = model.encoding;
+  const {position, avoidParentLayer, mark: _mark, method, lineAnchor, ...textEncoding} = label;
```
kanitw, line 382: Please rename `_mark` to `labelMark`.

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
...
+          ...(omit(
+            encodeBaseEncodeEntry(textModel, {
+              align: 'ignore',
+              baseline: 'ignore',
+              color: 'include',
+              size: 'ignore',
+              orient: 'ignore',
+              theta: 'ignore'
+            }),
+            ['x', 'y', 'angle', 'radius', 'theta']
```
kanitw, line 422: Add comment to explain why we omit these properties?

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
...
+          ) as TextEncodeEntry),
```
kanitw, line 423: Why do we need to cast to `as TextEncodeEntry`? Can we avoid casting?

Same for below.

src/channeldef.ts
```diff
@@ -198,6 +199,72 @@ export type TextDef<F extends Field> =
...
+   * anchor direction of a label from its base mark.
+   * The value could be one of `'left '` , `'right '` , `'top '` , `'middle '` ,
+   * `'bottom '` , `'top-left '` , `'top-right '` , `'bottom-left '` , or  '`bottom-right '`
+   */
+  anchor: LabelAnchor;
+};
+
+export type LabelDefMixins = {
+  /**
+   * A list of possible positions for a label to be placed relative to its base mark.
```
kanitw, line 222: What's the default behavior?

src/channeldef.ts
kanitw: Actually

```
Mark definition of the label text mark without the `'type'` property.
```

describes what the type looks like, not what the type is for. 

I think it should be something like:

```
A label mark definition for customizing the label's text mark. 
This definition includes all [properties of a text mark](http://vega.github.io/vega-lite/docs/text.html#properties) except `type`. 
```

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
...
+    case 'trail': {
+      const defaultLineAnchor = DEFAULT_LINE_ANCHOR[orient];
+      const anchorDirection = defaultLineAnchor.direction[lineAnchor];
+      return {
+        ...common,
+        lineAnchor,
+        ...(position
+          ? {anchor, offset}
+          : {
+              anchor: defaultLineAnchor.anchor(anchorDirection) as LabelAnchor[],
```
kanitw, line 493: Why do we need to cast `as LabelAnchor[]` here?

kanitw (commented):
See comments below.

Also, please:
- Avoid unnecessary casting. (If we need to cast, we should add a comment to explain why? -- Usually we don't really need to cast.) 
- Make sure all the docs have proper sentence capitalization and using  double quotes (") for string values and no quotes for property names.

src/normalize/pathoverlay.ts
```diff
@@ -151,7 +170,7 @@ export class PathOverlayNormalizer implements NonFacetUnitNormalizer<UnitSpecWit
           ...pick(markDef, ['clip', 'interpolate', 'tension', 'tooltip']),
           ...lineOverlay
         },
-        encoding: overlayEncoding
+        encoding: omit(overlayEncoding, ['label'])
```
kanitw, line 173: We need a comment here to explain why omitting.

src/normalize/pathoverlay.ts
```diff
@@ -164,7 +183,8 @@ export class PathOverlayNormalizer implements NonFacetUnitNormalizer<UnitSpecWit
           ...pick(markDef, ['clip', 'tooltip']),
           ...pointOverlay
         },
-        encoding: overlayEncoding
+        // drop label from encoding when having detail (grouped line chart)
```
kanitw, line 186: I'm not sure if I understand this comment. 

What do you mean "when having detail (grouped line chart)"? 
Given the code below doesn't do any conditional check.

src/normalize/pathoverlay.ts
```diff
@@ -122,7 +140,8 @@ export class PathOverlayNormalizer implements NonFacetUnitNormalizer<UnitSpecWit
           ...markDef
         }),
         // drop shape from encoding as this might be used to trigger point overlay
-        encoding: omit(encoding, ['shape'])
+        // drop label from encoding when not having detail (connected scatter plot)
```
kanitw, line 143: What do you mean "when having detail (connected scatterplot)"?
Given the code below doesn't do any conditional check.

src/normalize/pathoverlay.ts
```diff
@@ -36,7 +37,13 @@ function getPointOverlay(
   markConfig: LineConfig<ExprRef | SignalRef> = {},
   encoding: Encoding<string>
 ): MarkConfig<ExprRef | SignalRef> {
-  if (markDef.point === 'transparent') {
+  if (
+    markDef.point === 'transparent' ||
+    (markDef.type === 'line' &&
```
kanitw, line 42: Add comment to explain what's this new condition is for?

src/log/message.ts
```diff
@@ -18,6 +18,13 @@ import {Type} from '../type';
 import {stringify} from '../util';
 import {VgSortField} from '../vega.schema';
 
+export function dropChannelOnMark(mark: Mark, channel: Channel) {
```
kanitw, line 21: We already have `incompatibleChannel`. Please use that one instead.

src/log/message.ts
```diff
@@ -18,6 +18,13 @@ import {Type} from '../type';
...
+  return `${mark} does not support ${channel}: dropping ${channel}`;
+}
+
+export const ROUNDED_CORNER_STACKED_BAR_WITH_AVOID =
+  'Labels on stacked bar chart can only avoid the base mark (stacked bars)';
```
kanitw, line 26: Can you explain why we have this limitation? 

Does this mean that labels stacked bar charts also can't avoid "all marks"?

src/encoding.ts
```diff
@@ -273,6 +275,8 @@ export interface Encoding<F extends Field> {
    */
   text?: TextDef<F>;
 
+  label?: LabelDef<F>;
```
kanitw, line 278: Missing docs.

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
...
+  }
+}
+
 /**
  * If scales are bound to interval selections, we want to automatically clip
  * marks to account for panning/zooming interactions. We identify bound scales
  * by the selectionExtent property, which gets added during scale parsing.
  */
-function scaleClip(model: UnitModel) {
+export function scaleClip(model: UnitModel) {
```
kanitw, line 517: Why do we have to export this?

src/compile/mark/mark.ts
```diff
@@ -355,15 +524,15 @@ function scaleClip(model: UnitModel) {
  * If we use a custom projection with auto-fitting to the geodata extent,
  * we need to clip to ensure the chart size doesn't explode.
  */
-function projectionClip(model: UnitModel) {
+export function projectionClip(model: UnitModel) {
```
kanitw, line 527: Why do we have to export this?

src/compile/mark/mark.ts
```diff
@@ -355,15 +524,15 @@ function scaleClip(model: UnitModel) {
...
   const projection = model.component.projection;
   return projection && !projection.isFit ? true : undefined;
 }
 
 /**
  * Only output interactive flags if we have selections defined somewhere in our model hierarchy.
  */
-function interactiveFlag(model: UnitModel) {
+export function interactiveFlag(model: UnitModel) {
```
kanitw, line 535: Why do we have to export this?

src/channel.ts
```diff
@@ -472,6 +480,23 @@ const ALL_MARKS: Record<Mark, 'always'> = {
...
+  area: 'always',
+  bar: 'always',
+  line: 'always',
+  trail: 'always',
+  rect: 'always'
+} as const;
+
+export type LabelSupportingMark = keyof typeof LABEL_SUPPORTING_MARKS;
+
+export function supportLabel(mark: Mark): mark is LabelSupportingMark {
```
kanitw, line 496: Do we need to declare this new method? 

Can we use `getSupportedMark` since it seems to be the same thing?

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
   ];
 }
 
+export function getLabel(model: UnitModel, data: string): LabelMark[] {
```
kanitw, line 370: Rename to getLabelMark(s)?

src/compile/unit.ts
```diff
@@ -79,6 +79,8 @@ export class UnitModel extends ModelWithField {
   public readonly selection: SelectionParameter[] = [];
   public children: Model[] = [];
 
+  public label: {mark: LabelMark; level: number}[] = [];
```
kanitw, line 82: This label property sounds weird.  

- It's singular, but it's an array.
- What's the `level` really for? I think we can name the `level` property to be more descriptive, or at least add some comments.

src/compile/mark/mark.ts
kanitw: This method never seem to return array with more than one item.  Do we really need an array here?  Can't the return value be LabelMark or null?

src/compile/unit.ts
kanitw: Seeing down the `getLabel` never returns array with more than one item. Thus, I wonder if this `label` property really needs to be an array.

kanitw (commented):
Some more comments. 

It's worth being careful about singularity/plurality for label(s). 
The label sometimes an array and sometimes not, but I wonder if we really need an array in general.

src/compile/mark/mark.ts
```diff
@@ -38,7 +44,15 @@ const markCompiler: Record<Mark, MarkCompiler> = {
   trail
 };
 
-export function parseMarkGroups(model: UnitModel): any[] {
+export interface LabelMark extends TextMark {
+  transform: [LabelTransform];
```
kanitw, line 48: Do we need an array here? Is there ever more than one LabelTransform?

src/compile/mark/mark.ts
```diff
@@ -213,37 +231,46 @@ function getGroupsForStackedBarWithCornerRadius(model: UnitModel) {
...
+        marks: [
+          {
+            type: 'group',
+            encode: {update: innerGroupUpdate},
+            marks: [mark, ...labels]
+          }
+        ]
+      }
+    ],
+    label: [] as LabelMark[]
```
kanitw, line 272: why casting?

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
...
+  ];
+}
+
+const DEFAULT_LINE_ANCHOR = {
+  horizontal: {
+    direction: {
+      begin: 'bottom',
+      end: 'top'
+    },
+    anchor: (direction: string) => ['-left', '', '-right'].map(a => direction + a)
```
kanitw, line 439: > : string

'top' | 'bottom'?

(Minor)

src/compile/mark/mark.ts
```diff
@@ -340,12 +367,154 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
...
+  },
+  vertical: {
+    direction: {
+      begin: 'left',
+      end: 'right'
+    },
+    anchor: (direction: string) => ['top-', '', 'bottom-'].map(a => a + direction)
```
kanitw, line 446: > : string

'top' | 'bottom'?

(Minor)

src/channeldef.ts
chanwutk: Do you mean that we should not use the name `mark`?

src/channeldef.ts
chanwutk: I think that `avoid: 'all-marks' | 'base-mark'` is better too. But, there is a problem when users use a `line` mark with `point: true`. Using `'base-mark'` before normalizing should mean both the line and the point. However, We do not have a way to represent that after we normalize the spec.
So, I decide to use `avoidParentLayer` (or `avoidAncestorLayer`) because we can increment the avoid level by one whenever we have a normalization for line with point.

examples/compiled/bar_corner_radius_end.vg.json
chanwutk: I don't remember the reason I put it there (I might saw it from pathgroup). I will remove it.

src/compile/layer.ts
chanwutk: If a label mark is under another mark, the labels in the label mark would be behind that mark if they overlap.

src/compile/mark/mark.ts
chanwutk: I add this before I annotate the function return type. I will fix it.

src/compile/mark/mark.ts
chanwutk: Same for the above comment.

src/compile/mark/mark.ts
chanwutk: I defined the return type to be `LabelMark` which is a `TextMark` with a label-transform.
I think we could avoid this cast by loosen the definition of `LabelMark` to be just a mark with a label-transform.

src/channeldef.ts
chanwutk: The defaults are different if depending on the mark type of the base mark. Should we list all of them?

src/compile/mark/mark.ts
chanwutk: This is a type for an output mark in Vega. If I'm not mistaken, a transform can only be an array?

src/channeldef.ts
chanwutk: Here are some alternatives that I could think of:

- label
- textDef
- labelDef
- textMark
- labelMark

src/channeldef.ts
domoritz: Also, don't put spaces before the last ".

src/log/message.ts
chanwutk: There is a typo here. This limitation only applies to stack bar charts with rounded corners.
In Vega, a label mark can only avoid other marks that comes before itself.
We have this limitation because Vega-Lite compiles a stack bar chart with rounded corners into a group mark. So, we have to put the label mark inside the group mark. For this reason, we cannot pull the label mark out to avoid other marks that comes after the stacked bar chart.
On the other hand, a regular stacked bar chart is just a normal bar chart without grouping. So, we can push the label mark to be behind all the other marks to avoid them.

src/compile/layer.ts
chanwutk: Also in Vega, a text mark with label transform can only avoid the marks that comes before itself. To be able to avoid the marks that comes after itself, we need to push the text mark to the top.

src/channeldef.ts
domoritz: Just define a new type 

```
type LabelMarkDef = Omit<MarkDef<'text'>, 'type'>
...
mark?: LabelMarkDef,
```

chanwutk: I see.

src/channeldef.ts
kanitw: I'm not sure if I totally understand the following point yet:

> So, I decide to use avoidParentLayer (or avoidAncestorLayer) because we can increment the avoid level by one whenever we have a normalization for line with point.



But this sounds the the syntax is designed to optimize for a corner case.

If anything,  `avoid: 'all-marks' | 'base-mark' | {ancestor: number}`` would probably make it more readable.

src/channeldef.ts
kanitw: If it's not too complicated, let's try to list them.

src/channeldef.ts
kanitw: Not sure why you mark this as resolved as you still mention "a group mark".

src/channeldef.ts
```diff
@@ -198,6 +199,87 @@ export type TextDef<F extends Field> =
...
+  mark?: LabelMarkDef;
+
+  /**
+   * The padding in pixels by which a label may extend past the chart bounding box.
+   */
+  padding?: number;
+
+  /**
+   * The labeling method to use for area marks.
+   * One of `"floodfill"`, `"reduced-search"`, and `"naive"`.
```
kanitw, line 250: For both one of here and the explanation paragraphs below, I'd begin with "reduced-search" since it's the default.

src/channeldef.ts
```diff
@@ -198,6 +199,87 @@ export type TextDef<F extends Field> =
...
+   * This property only applies when the mark is a multi-series area.
```
kanitw, line 251: Do you mean stacked area?

src/channeldef.ts
```diff
@@ -198,6 +199,87 @@ export type TextDef<F extends Field> =
...
+   *
+   * If `"floodfill"`: For each area, the label is placed at the center of the largest possible rectangle that can fit into the area.
```
kanitw, line 253: This  'if "xxx":' pattern reads weird. 


To make it easier to read that there are 3 values and describe them right away. 
I'd do

```
The labeling method to use for stacked area marks. One of: 

- `"reduced-search"`: ...

- `"floodfill"`: ...

- `"naive"`: ...
```

src/compile/layer.ts
kanitw: These are great explanations to add to the comment in the code.

src/channeldef.ts
chanwutk: I see. The syntax you suggested `avoid: 'all-marks' | 'base-mark' | {ancestor: number}` sounds great to me. I will make the change.

src/channeldef.ts
chanwutk: It will look like this...

__Default value:__
- `bar`:
  - If vertical bar chart: `[
      {anchor: "top", offset: 2},
      {anchor: "top", offset: -2}
    ]`
  - If horizontal bar chart: `[
      {anchor: "right", offset: 2},
      {anchor: "right", offset: -2}
    ]`
- `line` or `trail`:
  - If vertical line/trail chart:
    - If `lineAnchor` is `"begin"`: `[
        {anchor: "bottom-left", offset: 2},
        {anchor: "bottom", offset: 2},
        {anchor: "bottom-right", offset: 2},
      ]`
    - If `lineAnchor` is `"end"`: `[
        {anchor: "top-left", offset: 2},
        {anchor: "top", offset: 2},
        {anchor: "top-right", offset: 2},
      ]`
  - If horizontal line/trail chart:
    - If `lineAnchor` is `"begin"`: `[
        {anchor: "top-left", offset: 2},
        {anchor: "left", offset: 2},
        {anchor: "bottom-left", offset: 2},
      ]`
    - If `lineAnchor` is `"end"`: `[
        {anchor: "top-right", offset: 2},
        {anchor: "right", offset: 2},
        {anchor: "bottom-right", offset: 2},
      ]`
- `rect`: `[{anchor: "middle", offset: 0}]`
- `circle` or `point` or `square`: `[
    {anchor: "top-right", offset: 2},
    {anchor: "top", offset: 2},
    {anchor: "top-left", offset: 2},
    {anchor: "left", offset: 2},
    {anchor: "bottom-left", offset: 2},
    {anchor: "bottom", offset: 2},
    {anchor: "bottom-right", offset: 2},
    {anchor: "right", offset: 2}
  ]`

aantn:
Any updates on this?

kanitw:
Do we have examples for bar and stacked bar?

chanwutk:
We don't have it right now, but I can add more.

examples/specs/label_multi_line.vl.json
```diff
@@ -0,0 +1,10 @@
+{
```
kanitw, line 1: line_color_label

kanitw (commented):
Great progress with a lot of awesome examples here.  I spent some time reviewing the examples here. 

I think we should:

- use _label as a suffix 
- try to add at least example per mark type, to see that we have thorough think through this. 

I'll find more time to review the actual code.

examples/compiled/label_multi_line.vg.json
```diff
@@ -0,0 +1,173 @@
...
+    },
+    {
+      "name": "marks_label",
+      "type": "text",
+      "style": ["text"],
+      "sort": {"field": "datum[\"year_date\"]"},
+      "from": {"data": "pathgroup"},
+      "encode": {
+        "update": {
+          "fill": {"value": "black"},
```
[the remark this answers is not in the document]
kanitw: The tricky bit is that we shouldn't apply color to area's label, nor to stacked bar's labels.

examples/specs/label_multi_line.vl.json
chanwutk: We already have a `line_color_label` example.
Its this one: https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEzjQATjRyZ289AFkGgnfwhxNagATDZcS7CTrMlgO50Y1pEziCIlRSpImCiooAzqgmiBwZyYbFAA1n5QEMQgAL5KmOpIshAAZmzqCGgA2qD5NMbmUcpBDAiUEACeCExsggCEALzdAOQAkgBC+n0ZALpKcLJQbMo0smRooLOCRcsglT7KUS1tHQGYzThwUbJsCAtIkUqC5NM7qLJGgumZIIJIzTWo5SDI6gSURspymMzmCyWoRAAA8NltBI9oqCQEcTlFMIgcEVroc6HdakFTu9mvCaNsojhNFAUWjTugAI4MXI6YI6Uh4zAE9BUmg0jJvBSgaazeaLDZw6FIMhkdR8TFRZBwpQIpF1BVZY701FYnGREkbaWy+X00COMhKwkK96qynU2laqJMll0II0DmCj5fH5-AFA6F0qJQGjqKAEwWgP0bQPoTEwzABa40Mh6D5wfIJpTKSUAFneIoh4oDcHjZIp6D27RuqMd6HOl1kuLe6XGLaUs1klShoGI5JcGwg2TYCXpz0ErzeQA
Also, our label-encoding version cannot replace this version because we cannot encode color into the text label.

site/docs/label.md
```diff
@@ -1,8 +1,72 @@
 ---
 layout: docs
 menu: docs
-title: Labeling (Coming Soon!)
+title: Labeling
 permalink: /docs/label.html
 ---
 
-Labeling will be supported soon. Please subscribe to [our repository](https://github.com/vega/vega-lite) to hear about our releases.
+Labels can provide details of a particular data point as a text annotation directly near the data point. Users can add text labels to marks of the following types: area, bar, line, trail, rect, circle, point, and square. To annotate marks with labels, users can encode fields with label encoding channel.
```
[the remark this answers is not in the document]
chanwutk: We do not support arc mark yet.
Vega also does not support arc, so we might need some more discussion about how would we implement this.

examples/specs/label_stacked_area.vl.json
```diff
@@ -0,0 +1,21 @@
...
+  "transform": [
+    {
+      "density": "Body Mass (g)",
+      "groupby": ["Species"],
+      "extent": [2500, 6500]
+    }
+  ],
+  "encoding": {
+    "x": {"field": "value", "type": "quantitative", "title": "Body Mass (g)"},
+    "y": {"field": "density", "type": "quantitative", "stack": "zero"},
```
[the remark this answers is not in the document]
chanwutk: I copied the spec from this example: https://vega.github.io/vega-lite/examples/area_density_stacked.html
If this is a bug, should the correct behavior be: if an area chart encode color or other details, they should be stacked?

examples/compiled/label_multi_line.vg.json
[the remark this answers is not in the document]
chanwutk: Could we have an option in the label encoding channel `detail: true | false`.
Then, the label would encode all the detail from the base mark that encode the same field.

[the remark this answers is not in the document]
chanwutk: Or, this might work:
```js
{
  ...
  "encode": {
    "color": {"field": "a", "type": "nominal"},
    "label": {
      "field": "a",
      // or
      "from_detail": "color",  // need a better name

      ...other configs
    }
  }
}
```

examples/specs/label_multi_line.vl.json
kanitw: > Also, our label-encoding version cannot replace this version because we cannot encode color into the text label.

Why shouldn't we support this? 

It will be so confusing why we have so many similar label examples, but done in different ways due to limitations of the `label` encodings.

site/docs/label.md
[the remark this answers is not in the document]
kanitw: You mean label in Vega doesn't support arc, right?

site/docs/label.md
```diff
@@ -1,8 +1,86 @@
...
+      mark: ...,
+      padding: ...,
+      method: ...,
+      lineAnchor: ...,
+    }
+  }
+}
+```
+
+## Configurations
```
kanitw, line 44: Are these configurations as in theme?  We use the term "properties" in other pages.

kanitw: Also, you need to define that this table lists properties of label field definition, in addition to standard  field definition property. 

See the language we use in encoding.md (https://vega.github.io/vega-lite/docs/encoding.html#text-field-def). 

- [ ] Please also add a "label" to the "text and tooltip" section in the encoding.md 
- [ ] rename the section to "text, label, and tooltip" 
- [ ] link to this page, just like how tooltip links to its own page.

src/channeldef.ts
```diff
@@ -197,6 +198,154 @@ export type TextDef<F extends Field> =
...
+   *     {anchor: "top", offset: 2},
+   *     {anchor: "top-left", offset: 2},
+   *     {anchor: "left", offset: 2},
+   *     {anchor: "bottom-left", offset: 2},
+   *     {anchor: "bottom", offset: 2},
+   *     {anchor: "bottom-right", offset: 2},
+   *     {anchor: "right", offset: 2}
+   *   ]`
+   */
+  position?: LabelPosition[];
```
kanitw, line 290: Do we have label rotation?   If not, we should have it. 

for vertical bar, I'd argue that the default text should be rotated by 90 (or -90) degree because the label will better scale for arbitrary length of text -- see Tableau's label for example.

examples/compiled/label_multi_line.vg.json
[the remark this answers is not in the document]
kanitw: That's one approach, and we need a good name for it: 
```
"label": {
  "field": ...,
  "apply/encode": "color" // similar to your from_detail above, but 
}
```

However, I think I like this format more: 

```
"label": {
  "field": ...,
  "color": ... where color can be `null` (use the mark's color) or `string` (color name/hex`)
}
```

src/channeldef.ts
chanwutk: We do not have that in Vega, right now.
Do you think it is ok to not have it right now, and we can add it later after we implement label rotation in Vega?

examples/compiled/label_multi_line.vg.json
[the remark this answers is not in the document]
chanwutk: I see. Can it be like this:
```js
"label": {
  "field": ...,
  "color": "inherit" | {value: "#fffff"}  
}
```

site/docs/label.md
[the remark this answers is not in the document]
chanwutk: Yes

src/channeldef.ts
kanitw: > Do you think it is ok to not have it right now?

I think it's pretty bad if we can't rotate the text for vertical bars, but we also have to account for how hard it is implement there, and whether you have time for it. 

(If you don't have time, then please at least file an issue in Vega, and give some hints for how could someone else go implement the feature.)

examples/compiled/label_multi_line.vg.json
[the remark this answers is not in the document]
kanitw: It can, but do we think that "inherit" should be used more often than custom color? (If custom color is more often, then this `{value: ...}` seems an overkill.)


By the way, maybe the term `inherit` isn't bad.  Plus, thinking more, we also have `mark` already, so maybe the first format you proposed is possibly better and `inherit` could be a good keyword for it

```
"label": {
  "field": ...,
  "inherit": "color" | ["color", "opacity"] // similar to your from_detail above
}
```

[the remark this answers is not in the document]
kanitw: @domoritz  opinions?

site/docs/label.md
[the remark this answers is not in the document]
chanwutk: Yes. It does not support arc.

examples/compiled/label_multi_line.vg.json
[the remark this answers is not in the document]
chanwutk: We also have to think about (1) how encoding inherits each encoding channel and (2) what is the default encoding channel to inherit for each mark type.
For example:
(1) In a line chart, the color in the label should inherit the line's stroke color.
(2) As you mentioned, area should not inherit color by default, but line should.

src/channeldef.ts
```diff
@@ -199,6 +200,155 @@ export type TextDef<F extends Field> =
...
+  }
+}
+
+export type LabelDefMixins = {
+  /**
+   * A list of possible positions for a label to be placed relative to its base mark.
+   * The default value depends on the mark type.
+   *
+   * __Default value:__
+   * - `bar`:
```
kanitw, line 248: This default won't work with bars with negative values.

src/normalize/pathoverlay.ts
```diff
@@ -36,7 +37,11 @@ function getPointOverlay(
   markConfig: LineConfig<ExprRef | SignalRef> = {},
   encoding: Encoding<string>
 ): MarkConfig<ExprRef | SignalRef> {
-  if (markDef.point === 'transparent') {
+  if (
+    markDef.point === 'transparent' ||
+    // if the main mark is a single line/trail/area chart, create an invisible point overlay for label.
+    (!markDef.point && encoding.label && pathGroupingFields(markDef.type, encoding).length <= 0)
```
kanitw, line 43: - [ ] Probably need to update the condition here to not always does this.
- [ ] It feels hacky that the main compiler needs to depends on a mechanism inside normalize.

src/normalize/pathoverlay.ts
```diff
@@ -122,7 +137,8 @@ export class PathOverlayNormalizer implements NonFacetUnitNormalizer<UnitSpecWit
           ...markDef
         }),
         // drop shape from encoding as this might be used to trigger point overlay
-        encoding: omit(encoding, ['shape'])
+        // If the main mark is multi-series line/trail or stacked area, label the main mark.
+        encoding: omit(encoding, ['shape', ...(isMultiSeriesPath ? [] : ['label' as const])])
```
kanitw, line 141: Need to fix the condition here.  (#7883)

src/stack.ts
```diff
@@ -194,9 +194,9 @@ export function stack(
   // If the dimension has offset, don't stack anymore
 
   // Should have grouping level of detail that is different from the dimension field
-  const stackBy = NONPOSITION_CHANNELS.reduce((sc, channel) => {
-    // Ignore tooltip in stackBy (https://github.com/vega/vega-lite/issues/4001)
-    if (channel !== 'tooltip' && channelHasField(encoding, channel)) {
+  // Ignore tooltip in stackBy (https://github.com/vega/vega-lite/issues/4001)
+  const stackBy = STACKABLE_CHANNELS.reduce((sc, channel) => {
+    if (channelHasField(encoding, channel)) {
```
kanitw, line 199: @chanwutk  -- Can you explain why do we need to remove tooltip here?

sid-kap:
@kanitw I'm really looking forward to this feature, is there anything I can do to help get this feature ready?

smashmirrorcardboardface:
here also - is there anything we can do to help?

tmespe:
Is this still being worked on?

durd07:
Is this still being worked on? It has been opened for so long.

salimmj:
Could you please update or clarify this?

It seems like it was supposed to be a redo of https://github.com/vega/vega-lite/pull/7222 which got reverted but then re-reverted? 

I'm curious when we'll get this feature: https://github.com/vega/vega-lite/issues/7228

PR #7391, opened 2021-04-16
docs: suggest using view.resize with streaming data

vicatcu:
Minor documentation updates related PR https://github.com/vega/vega/pull/3172

### DIFF
--- a/site/tutorials/streaming.md
+++ b/site/tutorials/streaming.md
@@ -91,7 +91,7 @@ vegaEmbed('#chart', vlSpec).then(function (res) {
 });
 ```
 
-As a word of caution, in some cases, streaming data might cause some unexpected results when rendering. If your rendered chart overflows your container in the presence of new incoming data, for example, you may need to configure [Autosize](https://vega.github.io/vega-lite/docs/size.html#autosize) functionality or explicitly use the [view.resize](https://vega.github.io/vega/docs/api/view/#view_resize) API to explicitly resize when new data is incorporated.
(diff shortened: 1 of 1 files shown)


domoritz:
Thanks. Can you also link to the alternative of using resize in the spec?

site/tutorials/streaming.md
```diff
@@ -91,7 +91,7 @@ vegaEmbed('#chart', vlSpec).then(function (res) {
 });
 ```
 
-As a word of caution, in some cases, streaming data might cause some unexpected results when rendering. If your rendered chart overflows your container in the presence of new incoming data, for example, you may need to configure [Autosize](https://vega.github.io/vega-lite/docs/size.html#autosize) functionality or explicitly use the [view.resize](https://vega.github.io/vega/docs/api/view/#view_resize) API to explicitly resize when new data is incorporated.
+As a word of caution, in some cases, streaming data might cause unexpected results when rendering. If your rendered chart overflows your container in the presence of new incoming data, for example, you may need to configure [Autosize](https://vega.github.io/vega-lite/docs/size.html#autosize) functionality or explicitly use the [view.resize](https://vega.github.io/vega/docs/api/view/#view_resize) API to explicitly resize when new data is incorporated.
```
domoritz, line 94:
```suggestion
New data may change the layout but Vega does not always resize the chart. Because of this optimization, your axes may be clipped off for example. To resolve this issue, you need to configure [autosize](https://vega.github.io/vega-lite/docs/size.html#autosize) or explicitly use [view.resize](https://vega.github.io/vega/docs/api/view/#view_resize) to resize when the data updates.
```

PR #7399, opened 2021-04-21
fix(docs): point "create a histogram" link to proper id on Binning documentation

mullim:
This fixes the "create a histogram" link in the Binning documentation (https://vega.github.io/vega-lite/docs/bin.html)

Before:
- https://vega.github.io/vega-lite/docs/bin.html#example

After:
- https://vega.github.io/vega-lite/docs/bin.html#histogram


domoritz:
Thank you

PR #7438, opened 2021-05-07
feat: allow signal or expression refs in more places

domoritz:


src/compositemark/common.ts
```diff
@@ -109,21 +109,22 @@ export function filterTooltipWithAggregatedField<F extends Field>(
...
-  const fiveSummaryTooltip: StringFieldDef<string>[] = tooltipSummary.map(
-    ({fieldPrefix, titlePrefix}): StringFieldDef<string> => {
+  const fiveSummaryTooltip: StringFieldDef<string, ExprRef>[] = tooltipSummary.map(
+    ({fieldPrefix, titlePrefix}): StringFieldDef<string, ExprRef> => {
       const mainTitle = withFieldName ? ` of ${getTitle(continuousAxisChannelDef)}` : '';
       return {
         field: fieldPrefix + continuousAxisChannelDef.field,
         type: continuousAxisChannelDef.type,
-        title: isSignalRef(titlePrefix) ? {signal: `${titlePrefix}"${escape(mainTitle)}"`} : titlePrefix + mainTitle
+        // TODO: this was signal: ...
```
kanitw, line 126: This seems unfinished?

marr:
Hi, is this still something that is planned? I'm a Vega beginner, and maybe I'm missing something simple. My data has a numeric _value_ field and a nominal "unit" field. I render a faceted view based on another column called "variable" and use a temporal x axis. I would like to have the chart have a title like "`variable (unit)`" -- the unit is repeated for each row, so I'm not sure how I'd do that.

I will try putting together an example in the editor.

domoritz:
Yes, still planned. I need to pick it up again.

src/compositemark/common.ts
kanitw: What about this one?

Issue #7374, opened 2021-04-09
Interval selections don't work with aggregations (`Expression parse error`)


joelostblom:
When I create a selection of type `'single'` or `'multi'`, I can use either the `fields` or `encodings` key to specify what the selection should select on. However, when I switch to a selection of type `'interval'`, it only works when using `ecodings`. Trying to use `fields` throws the error: `Expression parse error: ? selector033_scale_trigger : {}`. The chart below can be used to reproduce the error, [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAFgAY2AacaYsiygAlMiRoRQBmDgF9p3OgENCCvCCgKAtphQhFygLQA2AJxg6YTACY2EgOyHMAVkcAjR9du3MptgEY3ABzGlpYSIHIgGgoATgDWOi4xINyYUJB0JEioIBAADgpgcIQAnqoEGcTQqmgKADak2shsAHS23JSYtZhglVA6HV09ENE2YRE19Y0tvhEAHqoluY0gAI6kCkRFSnBo2tzwnXQ6AJqYSRGl2YvLaxvEysS7ySAKCAjRmAhKy5CkROERRgiBBiFCOLggAbdXqqKFDEYSMJXYpLHQkQiYaITZ4HWp0SgoADaIFOSQAurJ5EoFB1CATsnoFEZTOYrDZ7E5XO42J5vGA-IFgqEiaBSdEUL5jMZDHJRWdxchJdLZST5RKpbYVWL1cYAlq1chrGxHNIKdIgA):

```json
{
  "config": {"view": {"continuousWidth": 400, "continuousHeight": 300}},
  "data": {"name": "data-69cdce20376e55b52077e9c01b589223"},
  "mark": "bar",
  "encoding": {
    "opacity": {
      "condition": {"value": 0.7, "selection": "selector033"},
      "value": 0.1
    },
    "x": {"type": "quantitative", "field": "Year"},
    "y": {"type": "quantitative", "aggregate": "count"}
  },
  "height": 50,
  "selection": {"selector033": {"type": "interval", "fields": ["Year"]}},
  "datasets": {
    "data-69cdce20376e55b52077e9c01b589223": [
      {
        "Year": 1996
      },
      {
        "Year": 1996
      },
      {
        "Year": 1997
      },
      {
        "Year": 1998
      },
      {
[...]

PR #7442, opened 2021-05-07
fix: interval selection should gracefully fail if no valid projections

arvind:
Partially addresses #7374. Our [documentation](https://vega.github.io/vega-lite/docs/selection.html#project) already states that interval selections can only be projected via `encodings` but our types/schema did not reflect that. On top of that, this PR allows the interval selection to fail more gracefully when invalid projections are provided — we always threw a warning about invalid projections, but now the spec will render out a (static) visualization instead of throwing obscure parse errors.


src/compile/selection/project.ts
```diff
@@ -63,15 +63,15 @@ const project: SelectionCompiler = {
     };
 
     const type = selCmpt.type;
-    const cfg = model.config.selection[type];
+    const cfg = model.config.selection[type] as PointSelectionConfig;
```
domoritz, line 66: What is `selCmpt.type` here? If it's telling us the type, do we still need `as PointSelectionConfig`?

domoritz (approved):

src/compile/selection/project.ts
arvind: `selCmpt.type` is either `point` or `interval`. But, now, `fields` is only a property on the former, and `encodings` is on both. So, doing the typecast here felt like the cleanest fix to me rather than adding duplicative code simply for a `selCmpt.type === 'point'` type guard.

arvind: Alternatively, I can put this typecast in at [L102](https://github.com/vega/vega-lite/pull/7442/files#diff-87e9e2677b96aab8a505667f2d85f1da7d3d205da7c9596524aa67f45ad9d11cR102):

```diff
-fields = cfg.fields;
+fields = (cfg as PointSelectionConfig).fields;
```

Let me know if that's preferable.

domoritz: I think it's best not to assert a type when we don't know it yet. I changed it in 3987a99fb.

PR #7445, opened 2021-05-08
refactor: recursively replace ExprRef with SignalRef

domoritz:


domoritz:
After this, I think we can remove a lot of the signalRef methods in common.ts. 

As a follow up, look at these PRs to see what we can revert: #6825, #6826, #6847, #6849, #6850, #6806, and #6805.

PR #7461, opened 2021-05-14
chore: fix selection types in tests

domoritz:
Without these changes, `yarn tsc` fails. It also makes sense since only points selections have `fields`. I added a check to the CI to prevent this issue in the future.


domoritz:
@arvind, I am going ahead and merge this. Please let me know if there are any issues.

Issue #7465, opened 2021-05-15
Unable to interact with Translate and zoom documentation.


Abhilash-Chandran:
This is an issue in the documentation page for translate and zoom.

The interactive selection creates a random html in the page and the example code section doesn't change revealing the documentation.

https://vega.github.io/vega-lite/docs/selection.html#translate

A small gif showing the issue.

![1BkyRhL9QR](https://user-images.githubusercontent.com/4489849/118354713-bf915400-b56c-11eb-8fd5-1ed013464f95.gif)

domoritz:
This happens when the example file name in the docs is wrong. Can you help fix it?

Abhilash-Chandran:
I saw that error in the console.logs but I couldn't locate such a file in the vega-lite repo. It must be from Vega repo then. How do I know what is the correct file name for this section.?

domoritz:
We probably renamed a file and forgot to update the reference. Look at the example files and find the appropriate one.

Abhilash-Chandran:
I think [this](https://github.com/vega/vega-lite/tree/master/examples/specs) is where its located. I will look into it.

domoritz:
Looks like the issue is actually that we request `https://vega.github.io/vega-lite/examples/selection_zoom_undefined_shift-drag.vl.json`. The issue is in the select/js. Maybe we just need to have unique ids (`type` and `event` appear twice). I'd suggest trying `type_translate` etc.

Abhilash-Chandran:
Ah ok. Makes sense. I will try to fix that.

PR #7467, opened 2021-05-15
docs: fix id's of translate and zoom selection

Abhilash-Chandran:
This should fix the issue #7465 which is caused by the conflicting id's of the two dropdowns used with same id's in two different section.


domoritz:
Did you try `yarn site` to confirm that this fix works?

Abhilash-Chandran:
I will set up a codespace to do it now.. :)

Abhilash-Chandran:
I just tested the changes and results are recorded as short video below. Hope this is fine.


https://user-images.githubusercontent.com/4489849/118368240-4b71a300-b5a2-11eb-898c-5f01e35403ac.mp4

domoritz:
Thank you

PR #7533, opened 2021-06-17
fix: docs broken encoding type link

thcipriani:
The anchor `type` no longer exists on /docs/encoding.html

The "Type" link on the "Encoding" page points to /docs/type.html.
Update this page to point to the same location.


thcipriani:
Maybe I should have opened this as an issue since I didn't setup tests to run on the fork. I followed "update this page on github", hopefully this is helpful; if not, feel free to close.

PR #7563, opened 2021-07-01
fix: resolve theta to independent by default for faceted charts

domoritz:
Thanks for @leibatt for the issue report.


kanitw (approved):

src/compile/resolve.ts
```diff
@@ -4,7 +4,9 @@ import {Resolve, ResolveMode} from '../resolve';
 import {isConcatModel, isFacetModel, isLayerModel, Model} from './model';
 
 export function defaultScaleResolve(channel: ScaleChannel, model: Model): ResolveMode {
-  if (isLayerModel(model) || isFacetModel(model)) {
+  if (isFacetModel(model)) {
+    return channel === 'theta' ? 'independent' : 'shared';
```
kanitw, line 8: What about radius? I think radius probably should be independent too. 

Similarly, I think theta/concat should *not* be shared for concat.

kanitw (changes requested):
Thanks for adding this. It'd be nice to also cover relevant cases here in the same PR.

src/compile/resolve.ts
domoritz: I don't know whether I agree. I think radius is actually more easily comparable across facets so it makes sense to share by default. 

I agree that theta and radius should be independent in concat. I'll make that change.

PR #7600, opened 2021-07-19
docs: add baseline bar chart example

MartinA731:
New example of using baseline on a bar chart.


kanitw:
@domoritz  What happens to the auto gen Vega/SVG?

domoritz:
Since this is coming from a fork, the person who sends the pull request needs to set up a gh token as described in https://github.com/vega/vega-lite/blob/master/.github/PULL_REQUEST_TEMPLATE.md. Otherwise, the github action does not have the permissions to push changes.

What you can do is copy the branch into the main fork and make a new pull request.

There may be a better way but I haven't had the cycles to look into it.

PR #7641, opened 2021-08-07
fix: add explicit sort order to brush table example

RobinL:
If this fix is acceptable, this closes #7102 .

The behaviour described in the issue is surprising, but I don't think this is strictly bug. It's more an ambiguity in the example spec.

The example spec requires that 20 records are shown, but does not specify which 20 records should be shown when more than 20 meet the filter criterion (either a brush, or no user selection).

This would be best solved if the dataset has a unique id per record.  It does not, but the `Name` field can serve this purpose since it's unique.    

The surprising behaviour disappears when a sort order is specified - which is what I've suggested in this PR.

PR #7653, opened 2021-08-17
fix: add explicit sort order to brush table example #7641

kanitw:
replacing https://github.com/vega/vega-lite/pull/7641


examples/specs/brush_table.vl.json
```diff
@@ -19,7 +19,7 @@
   }, {
     "transform": [
       {"filter": {"param": "brush"}},
-      {"window": [{"op": "rank", "as": "rank"}]},
+      {"window": [{"op": "rank", "as": "rank"}], "sort": [{"field": "Name","order":"ascending"}]},
```
domoritz, line 22:
```suggestion
      {"window": [{"op": "rank", "as": "rank"}], "sort": [{"field": "Name", "order":"ascending"}]},
```

Issue #7621, opened 2021-07-30
Non-cartesian plots shouldn't include view stroke by default 


kanitw:
![image](https://user-images.githubusercontent.com/111269/127719142-87dc72a0-89a4-46e3-90ff-49ffee29c593.png)

domoritz:
Why not? In a multi view chart it makes it easier to see different radii. 

I'm playing devils advocate here. Let's consider whether the issue really are pie charts or single views more broadly.

kanitw:
> Why not? In a multi view chart it makes it easier to see different radii.

Yes, but (1) I bet people use non-small multiple pie charts way more often than multi-views 
and (2) for pie chart in particular, the radius is fixed. But I think you want to consider [Radial chart](http://vega.github.io/vega-lite/examples/arc_radial.html), which FWIW we did hide the stroke in our example. 

> whether the issue really are pie charts or single views more broadly.

I think the line I'd draw is whether something is a Cartesian plot (has x/y).  For example, I think 
maps like this [one](http://vega.github.io/vega-lite/examples/geo_choropleth.html) shouldn't have default view stroke by default.

domoritz:
That makes sense. I like it if we remove the view stroke from maps as well.

PR #7665, opened 2021-08-28
fix: don't include view stroke by default for non-cartesian plot

kanitw:
fix https://github.com/vega/vega-lite/issues/7621


src/compile/layer.ts
```diff
@@ -100,6 +101,17 @@ export class LayerModel extends Model {
...
   }
 
+  public assembleGroupStyle(): string | string[] {
+    const uniqueStyles = new Set<string>();
+    for (const child of this.children) {
+      for (const style of array(child.assembleGroupStyle())) {
+        uniqueStyles.add(style);
+      }
+    }
+    const styles = Array.from(uniqueStyles);
```
domoritz, line 111: I think you can check the length of the set and return the first item of the set instead. No need to make it an array.

domoritz (approved):
Please simplify the set code but otherwise looks good.

src/compile/layer.ts
kanitw: To get the first item you need to get the iterator and call next, which seems equally expensive as this but result in a more complicated code. So I think this is actually better.

domoritz: I disagree. The array call makes a copy while creating an iterator is highly optimized.

domoritz: I guess since we may return an array, it doesn't make a big difference.

kanitw: Most of the time N=1 or a few.  Maybe iterator is marginally better, but produces way longer code that doesn't justify the benefit.

PR #7684, opened 2021-09-06
Support offset channel for grouped bar / jittering

kanitw:


src/compile/scale/properties.ts
```diff
@@ -271,6 +272,11 @@ export function padding(
       const {type, orient} = markDef;
       if (type === 'bar' && !(isFieldDef(fieldOrDatumDef) && (fieldOrDatumDef.bin || fieldOrDatumDef.timeUnit))) {
         if ((orient === 'vertical' && channel === 'x') || (orient === 'horizontal' && channel === 'y')) {
+          const sizeChannel = getSizeChannel(channel);
```
kanitw: This is not exactly offset, but it's a relevant bug that bar should respect relative band size.

src/compile/scale/range.ts
```diff
@@ -307,8 +312,64 @@ function defaultRange(channel: ScaleChannel, model: UnitModel): VgRange {
...
+    if (isStep(size)) { // step is for offset
+      const step = getOffsetStep(size, offsetScaleType);
+      if (step) {
+        return step;
+      }
+    }
+    // otherwise use the position
+    return [0, {signal: `bandwidth('${positionScaleName}')`}];
+  } else { // continuous scale
+    throw new Error(`Cannot use ${channel} scale if ${positionChannel} scale is not discrete.`)
```
kanitw: We should never reach this line because we won't even generate the scale in the first place.

src/scale.ts
```diff
@@ -180,10 +180,33 @@ export interface ScaleConfig<ES extends ExprRef | SignalRef> {
...
+
+  /**
+   * Default inner padding for `x` and `y` band scales with nested `xOffset` and `yOffset` encoding.
+   *
+   * __Default value:__ `0.2`
+   *
+   * @minimum 0
+   * @maximum 1
+   */
+   nestedOffsetPaddingInner?: number | ES;
```
kanitw: Any opinion about this config name?

src/spec/base.ts
```diff
@@ -51,11 +51,26 @@ export interface DataMixins {
...
 export interface Step {
   /**
    * The size (width/height) per discrete step.
    */
   step: number;
+
+  /**
+   * Whether to apply the step to position scale or offset scale when there are both `x` and `xOffset` or both `y` and `yOffset` encodings.
+   */
+  for?: StepFor;
```
kanitw, line 65: Any opinion about this property name?

kanitw:
@domoritz  This should be ready and battle tested with many examples.   Please think of a different example we should test or add (or let me know if any of the examples I included for testing should be included in the gallery or docs too.)

src/stack.ts
```diff
@@ -166,18 +166,33 @@ export function stack(
   const stackedFieldDef = encoding[fieldChannel] as PositionFieldDef<string> | PositionDatumDef<string>;
   const stackedField = isFieldDef(stackedFieldDef) ? vgField(stackedFieldDef, {}) : undefined;
 
-  let dimensionChannel: 'x' | 'y' | 'theta' | 'radius' = getDimensionChannel(fieldChannel);
-  let dimensionDef = encoding[dimensionChannel];
+  const dimensionChannel: 'x' | 'y' | 'theta' | 'radius' = getDimensionChannel(fieldChannel);
+  const groupbyChannels: StackProperties['groupbyChannels'] = [];
+  const groupbyFields: FieldName[] = [];
```
domoritz, line 171: Could we make this a `Set<FieldName>`?

src/channel.ts
```diff
@@ -335,6 +341,28 @@ export function getOffsetChannel(channel: Channel) {
...
+    case Y:
+      return 'yOffset';
+  }
+  return undefined;
+}
+
+export function getMainChannelFromOffsetChannel(channel: OffsetScaleChannel): PositionScaleChannel {
+  switch (channel) {
+    case 'xOffset':
+      return 'x';
```
domoritz, line 360:
```suggestion
      return X;
```

src/channel.ts
```diff
@@ -335,6 +341,28 @@ export function getOffsetChannel(channel: Channel) {
...
+    case 'yOffset':
+      return 'y';
```
domoritz, line 362:
```suggestion
      return Y;
```

src/compile/data/stack.ts
```diff
@@ -195,54 +197,61 @@ export class StackNode extends DataFlowNode {
...
+        if (bin) {
+          // As we can only impute one field at a time, we need to calculate
+          // mid point for a binned field
+          transform.push({
+            type: 'formula',
+            expr:
+              `${bandPosition}*` +
+              vgField(dimensionFieldDef, {expr: 'datum'}) +
+              `+${1 - bandPosition}*` +
+              vgField(dimensionFieldDef, {expr: 'datum', binSuffix: 'end'}),
```
domoritz, line 241: Let's write this as one template string. 

```suggestion
              `${bandPosition}*${vgField(dimensionFieldDef, {expr: 'datum'})}+${1 - bandPosition}*${vgField(dimensionFieldDef, {expr: 'datum', binSuffix: 'end'})}`
```

src/log/message.ts
```diff
@@ -141,6 +149,14 @@ export function projectionOverridden<ES extends ExprRef | SignalRef>(opt: {
 
 export const REPLACE_ANGLE_WITH_THETA = 'Arc marks uses theta channel rather than angle, replacing angle with theta.';
 
+export function offsetNestedInsideContinuousPositionScaleDropped(mainChannel: PositionScaleChannel) {
+  return `${mainChannel}Offset dropped because ${mainChannel} is continuous`;
+}
+
+export function replaceOffsetWithMainChannel(mainChannel: PositionScaleChannel) {
+  return `There is no ${mainChannel} encoding. Replacing ${mainChannel}Offset encoding as ${mainChannel}.`;
```
domoritz, line 157: Nice!

src/spec/base.ts
domoritz: `forScale` or `applyTo` or `along` maybe?

domoritz (approved):
- Offset doesn't support title. Should we remove it from the schema?
- Shouldn't the grouped bar chart have some padding between the bars? It looks pretty tight right now. Also, how do you add padding and still make sure that bars are centered?

Other than that, this is awesome!

src/spec/base.ts
kanitw: I think

`{step: 20, for: "position"}`

reads better than 

`{step: 20, forScale/applyTo/along: "position"}`?

kanitw:
> Offset doesn't support title. Should we remove it from the schema?
Their is no guide, but the field itself still has title, which can be used for tooltip.

src/spec/base.ts
domoritz: Full examples

```js
{
  width: {step: 23, for: 'position'},
  mark: 'bar',
  encoding: {
    x: {field: 'x', type: 'nominal'},
    xOffset: {field: 'subx', type: 'nominal'}
  }
}
```

```js
{
  width: {step: 23, for: 'offset'},
  mark: 'bar',
  encoding: {
    x: {field: 'x', type: 'nominal'},
    xOffset: {field: 'subx', type: 'nominal'}
  }
}
```

Another option may be `forChannel`. I am okay with `for`.

src/scale.ts
domoritz: Why is it `nested` and not just offset padding?

kanitw:
> Shouldn't the grouped bar chart have some padding between the bars? It looks pretty tight right now. 

That's sort of intentional. If you google "grouped bar", even for D3, it seems like the norm is to have near-zero padding. 
Plus, given padding in the underlying D3 scales is a ratio, not absolute pixel, trying to do padding for both x and xOffset can be pretty messy.

If you strongly think we should change this, let me know.  Otherwise, I think we can merge this. 

> Also, how do you add padding and still make sure that bars are centered?

Good question.  You can add `padding` to x/yOffset scale.  (Good that you ask because I then noticed a bug for this padding.  It's not fixed.  

I also added `config.scale.offsetBandPaddingInner/Outer` just in case someone doesn't like the zero default.

src/spec/base.ts
kanitw: I think `for` is also better than `forChannel` because it's for position (which is a category of channel).  "for channel position" is swapping the order of the words anyway.  

If it's unclear if a longer word is clearer, I generally prefer brevity

src/scale.ts
kanitw: renamed to  `bandWithNestedOffsetPadding`   --> This is padding of the x/y scale, when they have x/yOffset nested, but the padding is for x/y scales, not x/yOffset scales.  (offsetPaddingInner/Outer is then for x/yOffset scales)

domoritz:
> That's sort of intentional. If you google "grouped bar", even for D3, it seems like the norm is to have near-zero padding.
Plus, given padding in the underlying D3 scales is a ratio, not absolute pixel, trying to do padding for both x and xOffset can be pretty messy, esp. as a default behavior.

Wow, that's surprising. I feel some padding is better since the bars show different categories. But I guess not padding makes the grouping clearer. I don't feel strongly about it so we can leave as it is as long as it's easy to add padding.

wylieconlon:
I'm not sure how the release cycle works on vega-lite, but it would be great if you could publish a release with this included!

kanitw:
Yes, I'm fixing one more bug that's relevant to this and then we'll release soon.

PR #7716, opened 2021-09-21
Update the default padding of labels in multi-series line cart to be Infinity

chanwutk:
Merge this PR to #7378 when Vega-Lite uses Vega 5.21.0


domoritz:
Do you want to send a pull request to update Vega-Lite to Vega 5.21?

chanwutk:
Yes, I will do that :)

kanitw:
Can you clarify why it should be infinity?

chanwutk:
@kanitw For a multi-series line chart, labels are at the end of each line. So, they would extend past the chart's bounding box.
And because we don't know the length of each label, we cannot limit the padding of the chart.

chanwutk:
I will merge this branch to #7378. Then, we can move our discussion about infinity padding for multi-series line chart in that branch.

PR #7731, opened 2021-09-30
docs: fix types page title

gordonje:
Seems like the title of this page should be "Property Types", not "Parameter Types". When I stumbled upon it, I initially confused it for a list of available (Vega Lite parameters)[https://vega.github.io/vega-lite/docs/parameter.html]. FWIW: A nearly [identical page in Vega's docs](https://vega.github.io/vega/docs/types/) is also currently titled "Parameter Types".

### DIFF
--- a/site/docs/types.md
+++ b/site/docs/types.md
@@ -1,7 +1,7 @@
 ---
 layout: docs
 menu: docs
-title: Parameter Types
+title: Property Types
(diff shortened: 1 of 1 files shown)


domoritz:
Good catch. I think we started with a copy of the page.

domoritz (approved):

PR #7745, opened 2021-10-07
perf: use more sets and reduce usage of contains

domoritz:
Merge after #7744


domoritz:
Agreed. I was just curious and needed a somewhat mindless task for an evening.

Issue #7758, opened 2021-10-16
Documentation: Add Clojure specific specs to example gallery ?


behrica:
The Clojure Data Science community is making heavy use of vega lite for plotting.
It has become the dominant  technology for doing plots in Clojure.

The adaption of vega (lite) in Clojure would be eased if there would be somewhere a gallery of examples, using the Clojure syntax for the Json lite specs, as an example:

```clojure
{:$schema "https://vega.github.io/schema/vega-lite/v5.json"
 :data {:url "data/cars.json"}
 :description "A scatterplot showing horsepower and miles per gallons for various cars."
 :encoding {:x {:field "Horsepower" :type "quantitative"}
            :y {:field "Miles_per_Gallon" :type "quantitative"}}
 :mark "point"}
 ```
 
 We could either create an Clojure specific example galery ourselves (reusing/copying the VL examples content of this repository),
or teh official  example gallery could be changed, to become "polyglot",
so showing the examples "not only in Json".

This could be done similar to some  "polyglot examples of some APIs, just for illustration:

![image](https://user-images.githubusercontent.com/127361/137589171-d1e1be66-65e8-414d-916f-2cf6cd4fdc26.png)

My question / proposal would be if you think it would be a good idea to change basically all the "individual example web pages", like: https://vega.github.io/vega-lite/examples/point_2d.html

to become "polyglot" and show the example not only in Json.

Clojure could be a first starting point.

kanitw:
Thanks for doing this. Sorry I'm a bit late to the party here.

Polyglot example is a great initiative and I think it would an amazing contribution to our communities. I've been hoping we'll have this for a long time.

I hope we can find a mechanism that can make this sustainable without introducing significant maintenance burdens to Vega-Lite maintainers and contributors.

Here are some thoughts:

- I hope we can minimize code reviews in the main Vega-Lite repos.  Note that Vega-lite maintainers do not have expertise for many of the wrapper languages. So if we include code examples in other languages in the main repo, we will have to review a number of PRs, but practically we'll just do "rubber stamps".  Basically we don't really understand the diff, but have more burdens to look at them and probably will click merge buttons without understanding much (and if we're busy, we may unintentionally block people when we are busy). 

- For this reason, I strongly prefer that we don't include examples for other languages in the main Vega-Lite repo. I'd prefer that we do automation during site build` or even automation within the website (just run JS code to convert).  

[...]

PR #7774, opened 2021-10-26
Proposal to make the vega lite examples page polyglot

behrica:
This is a very early draft proposal to make the vega lite examples page polyglot.
(see discussion here: https://github.com/vega/vega-lite/issues/7758)
The PR contains:

- change to site/_includes/example.html to show the examples in different languages in a tabbed view
- all vega lite spec files in EDN format (mostly used in Clojure)  They are autogenerated from all current examples json specs
   (the auto generation process has been started from file site/_data/examples.json
- empty files for ".R" ".python" ".hanami"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ test/log
 test/log/difflist.json
 yarn-debug.log*
 yarn-error.log*
+*.edn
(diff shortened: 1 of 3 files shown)


kanitw:
Hi, 

Thanks so much for doing this. I've been hoping that we can do something like this for a long time. 

To help facilitate reviews, can we avoid generating empty files in the PR?  
(If we don't have an example for a particular example in a language, I think it's ok to not include the file.  It's better to have fewer files to reviews.)

behrica:
I searched a bit but could not find how to do this lines:

```
{% include_relative {{ include.spec }}.vl.hanami %}
```

and have it "working", if the file is not present.
Is there an other jekyll directive which allows this ?

behrica:
I restarted the PR from scratch.
This time with generation of the edn files as part of the build.

behrica:
I will wait untul this is merged:
https://github.com/vega/vega-lite/pull/7807

Then we have an easier way tu bring "jet" and "puget" intote dev environment.

PR #7782, opened 2021-11-02
fix: tick should respect step size

### LINKED ISSUE #7724: Tick should respect step size
Currently the default tick length is fixed no matter what the step size is.

kanitw:
fix #7724


kanitw:
Replaced with https://github.com/vega/vega-lite/pull/9395

PR #7794, opened 2021-11-10
docs: fix empty toggle

ChristopherDavisUCI:
Replaced `'selection_filter_true' + this.value` with `'selection_filter_' + this.value`


domoritz:
Thank you!

Issue #7796, opened 2021-11-11
domainMin and domainMax not always respected


rfbowen:
`domainMin` and `domainMax` are not always respected, despite `domain` behaving correctly.

[...]

PR #7799, opened 2021-11-12
fix: default nice to false if domainMin/domainMax are set

maclockard:
fixes https://github.com/vega/vega-lite/issues/7796


site/_data/versions.yml
```diff
@@ -1,4 +1,4 @@
 vega: 5.21.0
 vega-lite: 5.1.1
-vega-embed: 6.19.1
+vega-embed: 6.20.0
```
maclockard, line 3: I have no idea why this bumped when I ran `yarn install && yarn test`

src/scale.ts
```diff
@@ -673,7 +673,7 @@ export interface Scale<ES extends ExprRef | SignalRef = ExprRef | SignalRef> {
    *
    * For temporal fields with time and utc scales, the `nice` value can be a string indicating the desired time interval. Legal values are `"millisecond"`, `"second"`, `"minute"`, `"hour"`, `"day"`, `"week"`, `"month"`, and `"year"`. Alternatively, `time` and `utc` scales can accept an object-valued interval specifier of the form `{"interval": "month", "step": 3}`, which includes a desired number of interval steps. Here, the domain would snap to quarter (Jan, Apr, Jul, Oct) boundaries.
    *
-   * __Default value:__ `true` for unbinned _quantitative_ fields; `false` otherwise.
+   * __Default value:__ `true` for unbinned _quantitative_ fields without explicit domain bounds; `false` otherwise.
```
maclockard, line 676: Was unsure of how to best phrase this

maclockard:
@domoritz I added `GH_PAT` secret personal access token to my fork with the following permissions: 
Screen Shot 2021-11-12 at 10 59 08 AM
but it looks like actions still arn't running? is there a step I'm missing?

domoritz:
I'm not sure. Maybe the instructions aren't complete. If you could investigate, that would be great.

site/_data/versions.yml
domoritz: This happens when you run `yarn build:versions`. We usually run it before releases.

domoritz:
I had to approve running the tests since this is your first contribution to Vega-Lite.

site/_data/versions.yml
maclockard: ah, okay, I'll revert this file then.

src/scale.ts
domoritz: Sounds good.

domoritz:
Moved to https://github.com/vega/vega-lite/pull/7802.

PR #7803, opened 2021-11-15
Adding support for themes in render.js

versipellis:
Adds requirement for vega-themes, renders via vega.parse.

### DIFF
--- a/bin/render.js
+++ b/bin/render.js
@@ -5,6 +5,7 @@ const path = require('path');
 const args = require('./args');
 const read = require('./read');
 const vegaLite = require('..');
+const themes = require('vega-themes');
 
 function load(file) {
   return require(path.resolve(file));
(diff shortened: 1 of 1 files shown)


versipellis:
@arvind

versipellis:
@arvind took a shot at adding themes support; for your review-

This renders the theme using `vega.parse()` using the theme specified per `vega-embed` in `usermeta.embedOptions.theme`. Uses `vega-themes` to get the appropriate config JSON and passes that as a config input into `vega.parse()`. I can't figure out why the background fails to parse properly, however - initially I assumed that background color was being specified in the spec (which overwrites config passed to `vega.parse()`), but in my unit tests this remained the case even without a background color passed in the spec. Also noting that this PR does add `vega-themes` as a `require()`.

bin/render.js
```diff
@@ -47,8 +48,9 @@ module.exports = (type, callback, opt) => {
 
   // instantiate view and invoke headless render method
   function render(vlSpec) {
+    const specTheme = themes[vlSpec.usermeta.embedOptions.theme];
```
domoritz, line 51: Should we also support a command line argument to set the theme? 

And if we support Vega-Lite, should we also support Vega?

PR #7870, opened 2021-12-24
feat: support isolatedModules

### LINKED ISSUE #7869: Support TS isolatedModules
When I try to `yarn build` my Nextjs project using typescript and vega-lite, I get this error:
```
The following mandatory changes were made to your tsconfig.json:

	- isolatedModules was set to true (requirement for babel)

Failed to compile.

./node_modules/vega-lite/src/compositemark/index.ts:28:9
Type error: Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.

  26 | } from './errorbar';
  27 |
> 28 | export {BoxPlotConfig} from './boxplot';
     |         ^
  29 | export {ErrorBandConfigMixins} from './errorband';
  30 | export {ErrorBarConfigMixins} from './errorbar';
  31 |
error Command failed with exit code 1.
```
I think this is pointing to this line: https://github.com/vega/vega-lite/blob/22057b18d54e71be4cc7c959c5f83d43d765a697/src/compositemark/index.ts#L28
Do we need to update these `export` statements to `export type`?

I doubt that the vega-lite codebase is not compatible with Babel, since vega-lite is currently producing ES2015 output. So maybe my TypeScript configuration is wrong?

Here is my `tsconfig.json`:
```
{
  "compilerOptions": {
    "target": "es6",
    "lib": [
      "dom",
      "dom.iterable",
      "esnext"
    ],
    "allowJs": true,
    "skipLibCheck": true,
    "strict": false,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "esModuleInterop": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "jsx": "preserve",
    "baseUrl": ".",
    "isolatedModules": true,
    "downlevelIteration": true
  },
  "include": [
    "next-env.d.ts",
    "**/*.ts",
    "**/*.tsx"
  ],
  "exclude": [
    "node_modules"
  ]
}
```

sid-kap:
Fixes #7869 

This makes it easier to use vega-lite in NextJS TypeScript projects, since NextJS uses SWC/Babel which require isolatedModules

### DIFF
--- a/src/compositemark/index.ts
+++ b/src/compositemark/index.ts
@@ -25,9 +25,9 @@ import {
   normalizeErrorBar
 } from './errorbar';
 
-export {BoxPlotConfig} from './boxplot';
-export {ErrorBandConfigMixins} from './errorband';
-export {ErrorBarConfigMixins} from './errorbar';
+export type {BoxPlotConfig} from './boxplot';
+export type {ErrorBandConfigMixins} from './errorband';
+export type {ErrorBarConfigMixins} from './errorbar';
 
 export type CompositeMarkNormalizerRun = (
   spec: GenericUnitSpec<any, any>,
--- a/src/spec/base.ts
+++ b/src/spec/base.ts
@@ -12,7 +12,7 @@ import {LayoutAlign, RowCol} from '../vega.schema';
 import {isConcatSpec, isVConcatSpec} from './concat';
 import {isFacetMapping, isFacetSpec} from './facet';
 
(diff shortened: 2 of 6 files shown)


domoritz:
Looks good. Please fix the failure in the test. I think you need to add an `export {}` or something like that.

sid-kap:
Thanks so much, and happy holidays!

PR #7885, opened 2021-12-27
Fix types, docs, examples 

kanitw:
Merging into the label feature branch. 

I also created https://github.com/vega/vega-lite/milestone/73 as a milestone for label.
Let's plan to fix all P1 before we release.


domoritz:
@kanitw @chanwutk what's the status of this pull request?

chanwutk:
I actually have not worked on this for quite a while. I remember that our last discussion was on designing syntax and defaults for bar and line labeling. Seems like there are a lot of updates since then. I would be happy to meet, if that is helpful.

Issue #7883, opened 2021-12-27
Multi-series line label incorrect 


kanitw:
```json
{
  "data": {"url": "data/stocks.csv"},
  "mark": "line",
  "encoding": {
    "x": {"timeUnit": "year", "field": "date"},
    "y": {"aggregate":"mean", "field": "price", "type": "quantitative"},
    "color": {"field": "symbol", "type": "nominal", "legend": null},
    "label": {"field": "symbol"}
  }
}
```

<img width="818" alt="image" src="https://user-images.githubusercontent.com/111269/147504234-7302d7be-13be-49fa-b2e9-4f453fa700fe.png">

PR #7906, opened 2022-01-06
fix: label's lineAnchor uses 'start' instead of 'begin'

chanwutk:
fix the last issue in #7883

### DIFF
--- a/site/docs/label.md
+++ b/site/docs/label.md
@@ -81,9 +81,9 @@ Encoding a field to labels in multi-series line chart results the labeling algor
 
 <span class="vl-example" data-name="label_multi_line"></span>
 
-As a default, the labeling algorithm place each label at the end of each line. You can configure it to place each label at the beginning of each line by setting `lineAnchor: 'begin'`.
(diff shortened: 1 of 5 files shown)


examples/compiled/line_color_label_start.vg.json
```diff
@@ -1,3 +1,4 @@
+start
```
kanitw, line 1: This is weird.  Why is "start" added here?

kanitw (changes requested):

site/docs/label.md
```diff
@@ -81,9 +81,9 @@ Encoding a field to labels in multi-series line chart results the labeling algor
 
 <span class="vl-example" data-name="label_multi_line"></span>
 
-As a default, the labeling algorithm place each label at the end of each line. You can configure it to place each label at the beginning of each line by setting `lineAnchor: 'begin'`.
+As a default, the labeling algorithm place each label at the end of each line. You can configure it to place each label at the start of each line by setting `lineAnchor: 'start'`.
 
-<span class="vl-example" data-name="line_color_label_begin"></span>
+<span class="vl-example" data-name="line_color_label_start"></span>
```
kanitw, line 86: Maybe the example we want is the "end" one?

PR #7917, opened 2022-01-11
docs: add version tag

mcmcgrath13:
This is perhaps half an issue, half a PR.

On the Vega docs, there are useful tags all over to say when a feature was introduced within a major version.  I was helping someone debug an issue with `xOffset` today and it looks like this was introduced in version `5.2.0`, but that isn't mentioned in the docs (leading to confusion when it didn't work).   It would be helpful if these tags could be added where applicable.

Please:

- [x] Make the pull requests (PRs) atomic (fix one issue at a time). Multiple relevant issues that must be fixed together? Make atomic commits so we can easily review each issue.
- [x] Provide a concise title as a [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties") so we can easily copy it to the release note.
- [ ] Lint and test (Run `yarn test`).
- [ ] If you send a pull request from a fork, make sure that GitHub actions run successfully. Make sure to add a [`GH_PAT` secret](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) for a [personal access token](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token) with public repository permissions.
- [ ] Review your changes before sending the PR (to ensure code quality).
- For new features:
  - [ ] Add new unit tests.

Tips:

### DIFF
--- a/site/docs/encoding.md
+++ b/site/docs/encoding.md
@@ -215,7 +215,7 @@ In addition to the general [field definition properties](#field-def), field defi
 
 {:#positon-offset}
 
-## Position Offset Channels
+## Position Offset Channels {% include tag ver="5.2" %}
 
 `xOffset` and `yOffset` position channels determine additional offset to the `x` or `y` position.
 


domoritz:
I agree version tags are useful but only if they are somewhat complete (e.g. cover at least the minor versions since the last major release). Therefore, I won't merge this pull request.

Ideally, we would auto generate these tags. We have the schema so we could theoretically do it. 

Could you send a pull request to add version tags for 5.1 and 5.2?

mcmcgrath13:
I'm thinking it will be easiest to look at the features in the release notes for 5.1 and 5.2 and use those.  Do you know if that would miss anything that really should get a tag?

domoritz:
It could but hopefully not.

One thing I worry is maintainability. After this work is done it will fall on @kanitw and me to maintain the version tags. Thinking more, I don't think it's worth your time right now until we have a sustainable maintenance plan in place. 

I'll close the pull request for now but we can keep talking about the idea. I am all for it if we find a sustainable way.

mcmcgrath13:
Agree, this needs to be sustainable.  I just pushed a few options on how to approach, which might inform some of that down the road.

1) add the tags on headers on docs where necessary
2) add the span html for the tag in the jsdoc string (pretty clunky)
3) add a `version` or similar extra tag to the json schema which can then be used programmatically in the jekyll templates (this is probably the nicest from a maintenance perspective, but I'm not sure about the potential downstream impacts of adding a custom property to the json schema - per the standard, it should be okay, but standards say a lot of things...).

image

The top tag is with the jsdoc key and the lower with the jsdoc inline html.  Adding the `default` to the jsdoc keys instead of markdown might be worth exploring either way.  The styling could look the same, but then it's available programmatically and if version tags are utilized down the road, they could be added in a nicer looking way more easily.

domoritz:
I agree that adding the version to the schema (via the comments) is the cleanest option. I don't think this should cause too many issues as we already add custom annotations for e.g. the default value. 

I would consider adding the version tags this way. For users, if they don't see a version tag, they have to check when a property was introduced.

I want to hear from @kanitw before I ask you to do this work, though.

PR #7920, opened 2022-01-13
style: use optional chaining to simplify code

domoritz:


kanitw (approved):
P42 produces so many noises

PR #7933, opened 2022-01-20
feat: optimize ID-driven selections

arvind:
With geo-interval (#6953) and lasso (#7932) selections on the horizon, there are going to be not only many more ID-driven selections than before, but these selections are likely going to capture many more IDs than before.  As a result, this PR cherry picks selection ID-based optimizations from #6953 and improves on them in a more cross-cutting fashion. 

In particular, stores for ID-driven selections use a flattened structure with tuples in the form of `{unit, _vgsid_}` sorted by `_vgsid_`. And inclusion testing now occurs via a new Vega expression function that binary searches through the store to determine a hit. 

This PR relies on updates introduced in vega/vega#3397 (so runtime tests may not pass until that PR is merged and released).

<details>
  <summary>📦 Published PR as canary version: <code>5.2.1--canary.7933.9a516dd.0</code></summary>
  <br />

  ```bash
  npm install vega-lite@5.2.1--canary.7933.9a516dd.0
  # or 
  yarn add vega-lite@5.2.1--canary.7933.9a516dd.0
  ```
</details>


arvind:
Woo hoo, with Vega 5.22 released, this PR is now ready for review which should unblock the much more exciting geo-interval (https://github.com/vega/vega-lite/pull/6953) and lasso (https://github.com/vega/vega-lite/pull/7960) features!

domoritz (approved):
Looks great. One question about the format. We store selection ids in sorted order so we can do a binary search. Why don't we store the ids in an object that would make it even easier to search and edit (e.g. some kind of set)?

arvind:
> Why don't we store the ids in an object that would make it even easier to search and edit (e.g. some kind of set)?

That's a good idea but I don't think Vega has support for Sets/OrderedSets? I might've missed something though. But, in the event I haven't, I think it's a good idea for us to consider in the future or if/when we move selections down to the Vega level.

domoritz:
Vega itself could definitely use Sets but they cannot be serialized as JSON. The solution for that is usually to represent a set as an object with values `1` or `true` as in https://github.com/vega/vega/blob/main/packages/vega-util/src/toSet.js. But yes, let's consider it for when we move selections to Vega.

PR #7946, opened 2022-01-26
docs: add ecosysem section for tools that scale Vega-Lite

jonmmease:
I would like to add my new project [VegaFusion](https://vegafusion.io/) to the ecosystem page.

I noticed that a few related projects that focus on scaling Vega-Lite/Altair visualizations to larger datasets were not already present, so added a new section named "Tools for Scaling Vega-Lite Visualizations" with descriptions of `altair-transform` (cc @jakevdp) and `ibis-vega-transform` (cc @saulshanabrook) and VegaFusion.

Happy to add something on [Scalable Vega](https://github.com/vega/scalable-vega) if that makes sense, and to reorganize/reword as you see fit.  Thanks!

### DIFF
--- a/site/ecosystem.md
+++ b/site/ecosystem.md
@@ -28,6 +28,12 @@ We mark featured plugins and tools with a <span class="octicon octicon-star"></s
 - [VizLinter](https://vizlinter.idvxlab.com/), an online editor that detects and fixes encoding issues based on vega-lite-linter.
 - [Datapane](https://github.com/datapane/datapane), a Python framework for building interactive reports from open-source visualization formats such as Vega-Lite.
 
+## Tools for Scaling Vega-Lite Visualizations
+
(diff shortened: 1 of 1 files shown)


domoritz:
Thank you. Let's add scalable Vega as well since it's a neat little demo of how to run computation somewhere else.

jonmmease:
> Let's add scalable Vega...

Done!  Also, I took the liberty of adding the "featured" badge to VegaFusion since it's a new project, but let me know if that's not appropriate or if any of the other entries should also be featured.

domoritz:
I've intended the featured label for projects that are/have been inspiring, are unique (hard to define), and somewhat maintained. It's a super subjective label but I think VegaFusion is a good candidate so let's add it.

saulshanabrook:
Thanks @jonmmease!

PR #7984, opened 2022-02-17
chore: set up auto for versioning/release management

hydrosquall:
## Motivation

Something changed about Github scopes in the past week  that is preventing me from pushing new commits onto #7972  (see [message](https://a.cl.ly/YEuB0Pnn)) .

Opening up a new PR seems to resolve the problem.

- Unified versioning / release process through the Vega org: 
  - less time spent spent to rebuild packages locally/releasing/managing changelogs = more time for new features

## Changes

- Replicate the setup in https://github.com/vega/vega-tooltip/pull/566
- Left the documentation + schema publishing steps intact, updated that section of the workflow to only trigger those when new releases are created
- Figure out a way to dry-run a test of the documentation publishing step from `stable`

## Pre-merge checklist for a repo admin

- [x] Create a branch called `next` from the current head of `master` https://github.com/vega/vega-lite/tree/next
   -  [x] make `next` the base branch of this PR instead of `master`
   -  [x] make `next` the base branch for the whole repo 
-  [x] Create a branch called `stable` from the latest tag  `v5.2.0`  https://github.com/vega/vega-lite/tree/stable
-  [x] Make a tagged release of the latest tag (This is done: https://github.com/vega/vega-lite/releases/tag/v5.2.0 )
-  [x] Create tags called `skip-release` and `released` in this repo (the first is used to tell auto not to release anything if this PR is released, the second is used to tag PRs that have been released).


scripts/bump.sh
```diff
@@ -7,12 +7,12 @@ if [ "$#" -lt 1 ]; then
   exit 1
 fi
 
-if [ "$(git rev-parse --abbrev-ref HEAD)" != "master" ]; then
+if [ "$(git rev-parse --abbrev-ref HEAD)" != "stable" ]; then
```
hydrosquall, line 10: I am partly tempted to remove this file completely to avoid confusing people, but I left it intact just in case there are reasons to version Vega-lite manually outside of what is being calculated using semantic commit messages.

domoritz: I'm okay with removing it. I made the file because I wanted a way to bump the version and update all the required files without having to remember the commands. I don't think we need it anymore with your process.

hydrosquall: Ok! If we need it again, it can be retrieved from git history. Removed in https://github.com/vega/vega-lite/pull/7984/commits/334cfeb7f806331b979691cf74e85e57acd31360

PR #8021, opened 2022-03-06
Fix typo in timeunit 'weekdayhours'

willium:
Quick typo fix, but probably will be a breaking change, so will need a version bump for this


domoritz:
It's just in the schema so I think it's mostly a bug we can fix. We should bump a minor version.

domoritz:
Rebased in https://github.com/vega/vega-lite/pull/9206

PR #8031, opened 2022-03-07
chore: update master to next in CI scripts and docs

domoritz:
<details>
  <summary>📦 Published PR as canary version: <code>5.2.1--canary.8031.0044b51.0</code></summary>
  <br />

  ```bash
  npm install vega-lite@5.2.1--canary.8031.0044b51.0
  # or 
  yarn add vega-lite@5.2.1--canary.8031.0044b51.0
  ```
</details>

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -35,7 +35,7 @@ jobs:
           sudo mv parallel sem /usr/local/bin
 
       - name: Format
-        if: github.ref != 'refs/heads/master'
+        if: github.ref != 'refs/heads/next'
         run: yarn format
 
       - name: Build Schema
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@ name: Test
 on:
   push:
     branches:
-      - master
+      - next
   pull_request:
 
 jobs:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ You can find [tasks with the "Good first issue" label in the issue tracker :pray
 
 The website is under `site/` and the documentation is under `site/docs/`. We use GitHub Pages to publish our documentation when we release a new version. To contribute changes to the documentation or website, simply submit a pull request that changes the corresponding markdown files in `site/`.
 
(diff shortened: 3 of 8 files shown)


.github/workflows/check.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           sudo mv parallel sem /usr/local/bin
 
       - name: Format
-        if: github.ref != 'refs/heads/master'
+        if: github.ref != 'refs/heads/next'
```
domoritz, line 38: @hydrosquall do we still need this?

.github/workflows/test.yml
```diff
@@ -3,7 +3,7 @@ name: Test
 on:
   push:
     branches:
-      - master
+      - next
```
domoritz, line 6: Was this not updated for a reason?

hydrosquall: It was an oversight, good to include it now

.github/workflows/check.yml
hydrosquall: I think so, assuming the goal here is to warn someone if they tried to open a PR with some unformatted code

hydrosquall (approved):
Looks good, thanks. After this goes in, I'm think might consider deleting the `master` branch altogether (after merging the commits it absorbed accidentally to next first)

domoritz:
Go for it.

hydrosquall:
Unfortunately, there seems to be a bug with Github, I've queried for all PRs that may have master as the base (TIL via [Stackoverflow](https://stackoverflow.com/questions/20547198/how-can-i-filter-all-github-pull-requests-for-a-specific-target-branch)), and the list is empty

- https://github.com/vega/vega-lite/pulls?q=is%3Aopen+is%3Apr+base%3Amaster

However, the "delete" button isn't working in my UI.

Image 2022-03-07 at 8 18 12 PM

Hopefully this is a caching issue that will go away tomorrow.

domoritz:
Hmm, I'm still getting this error today.

hydrosquall:
@domoritz me too...

One option might be to rename the branch to something like `master-do-not-use-see-pull-8031`, since the "rename" function appears to still be working. I'm going to do that as a starting point, since hard to find is almost as good as deleted.

hydrosquall:
Image 2022-03-09 at 9 13 43 PM

Success! (the `master` name persisted in the delete action, despite my going through the rename step first).

domoritz:
Lol and 🎉

Issue #7635, opened 2021-08-03
Vega-tooltip image support


ehknight:
Please:

- [x] Check for duplicate issues. Please file separate requests as separate issues.

I could find any duplicate issues

- [x] Describe how to reproduce the bug.

*(x-post from Altair issue here: https://github.com/altair-viz/altair/issues/2474)*

My understanding is that to render tooltips, Vega-Lite and Altair uses the `vega-tooltip` plugin. Per that project's [README](https://github.com/vega/vega-tooltip), one awesome feature that library has is the ability to render images as part of the tooltip with the image key. From the docs:

 > Supports special keys title (becomes the title of the tooltip) and image (used as the url for an embedded image)

However, this does not seem to be supported in Altair, because it is does not seem to be [added yet](https://github.com/altair-viz/altair/issues/2474#issuecomment-875919974) to Vega-Lite. This would be a very useful feature to support.

- [x] Use the latest versions of Vega and Vega-Lite.
- [x] Provide an [minimal, reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) spec in JSON, wrapped by triple backticks like this:

Works:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.8.1.json",
  "config": {
    "view": {
      "continuousHeight": 300,
      "continuousWidth": 400
    }
  },
  "data": {
    "name": "data-724890d4e94cac4976cbd99ef3589be4"
  },
  "datasets": {
    "data-724890d4e94cac4976cbd99ef3589be4": [
      {
        "a": 1.0,
        "b": 1.0
      }
    ]
  },
  "encoding": {
    "tooltip": [
      {
        "field": "a",
        "title": "My Cool Tooltip",
        "type": "quantitative"
      }
    ],
    "x": {
      "field": "a",
      "type": "quantitative"
    },
    "y": {
      "field": "b",
      "type": "quantitative"
    }
  },
  "mark": "circle",
  "selection": {
    "selector003": {
      "bind": "scales",
      "encodings": [
        "x",
        "y"
      ],
      "type": "interval"
    }
  }
}
```

Does not work:
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.8.1.json",
  "config": {
    "view": {
      "continuousHeight": 300,
      "continuousWidth": 400
    }
  },
  "data": {
    "name": "data-724890d4e94cac4976cbd99ef3589be4"
  },
  "datasets": {
[...]

PR #8048, opened 2022-03-17
docs: add example of rendering images in tooltips

joelostblom:
I am not sure about whether this is the correct place for this information, but it is very useful with image tooltips and it took me quite some time to figure out how to create them by going through the issues, so I think it would be a valuable addition somewhere in the docs.

Example from this comment:

- https://github.com/vega/vega-lite/issues/7635#issuecomment-925892820


site/docs/tooltip.md
```diff
@@ -64,6 +64,14 @@ To avoid that the tooltips groups the data, add an aggregate to the tooltip enco
 
 <div class="vl-example" data-name="bar_tooltip_aggregate"></div>
 
+## Tooltip image
+
+To display an image in a tooltip, you need to use the special field name `image`, which is an indicator to the Vega Tooltip plugin to render an image instead of displaying text in the tooltip. The image tooltip can be specified either by setting the `tooltip` property of the mark definition (as detailed above) or by passing the field *as an array* to the tooltip encoding channel, as in the example below:
```
domoritz: We need to clarify that this only works with Vega-Tooltip and add a link.

domoritz (changes requested):
I'll have to hold off on merging until we fix https://github.com/vega/vega-lite/issues/8049

site/docs/tooltip.md
joelostblom: Fixed!

examples/compiled/text_tooltip_image.svg
```diff
@@ -0,0 +1 @@
...="0 0 92 70"><rect width="92" height="70" fill="white"/><g fill="none" stroke-miterlimit="10" transform="translate(36,5)"><g class="mark-group role-frame root" role="graphics-object" aria-roledescription="group mark container"><g transform="translate(0,0)"><path class="background" aria-hidden="true" d="M0.5,0.5h20v60h-20Z" stroke="#ddd"/><g><g class="mark-text role-mark marks" role="graphics-object" aria-roledescription="text mark container"><text aria-label="image: data/ffox.png" role="graphics-symbol" aria-roledescription="text mark" text-anchor="middle" transform="translate(10,33)" font-family="sans-serif" font-size="11px" fill="black">data/ffox.png</text><text aria-label="image: data/gimp.png" role="graphics-symbol" aria-roledescription="text mark" text-anchor="middle" transform="translate(10,53)" font-family="sans-serif" font-size="11px" fill="black">data/gimp.png</text><text aria-label="image: data/7zip.png" role="graphics-symbol" aria-roledescription="text mark" text-anchor="middle" transform="translate(10,13)" font-family="sans-serif" font-size="11px" fill="black">data/7zip.png</text></g></g><path class="foreground" aria-hidden="true" d="" display="none"/></g></g></g></svg>
```
domoritz, line 1: should we remove the frame?

site/docs/tooltip.md
```diff
@@ -64,6 +64,14 @@ To avoid that the tooltips groups the data, add an aggregate to the tooltip enco
 
 <div class="vl-example" data-name="bar_tooltip_aggregate"></div>
 
+## Tooltip image
+
+To display an image in a tooltip you need to be using the [Vega Tooltip plugin](#plugin). Vega Tooltip requires the special field name `image` to be used to indicate that the field values should be rendered as images instead of displayed as text in the tooltip. The image tooltip can be specified either by setting the `tooltip` property of the mark definition (as detailed above) or by passing the field _as an array_ to the tooltip encoding channel, as in the example below:
```
domoritz, line 69:
```suggestion
To display an image in a tooltip you can use the [Vega Tooltip plugin](#plugin). Vega Tooltip requires the special field name `image` to indicate that the field values should be rendered as images instead of displayed as text. The image tooltip can be specified either by setting the `tooltip` property of the mark definition (as detailed above) or by passing the field _as an array_ to the tooltip encoding channel, as in the example below:
```

domoritz (approved):

examples/compiled/text_tooltip_image.svg
joelostblom: Yes, it was bothering me too, fixed!

PR #8104, opened 2022-04-15
feat: Lasso Specification Compiler

dvmoritzschoefl:
This is the vega-lite implementation for lasso selections. It is compatible with the current vega master branch.


src/compile/selection/lasso.ts
```diff
@@ -0,0 +1,96 @@
...
+          enter: {
+            fill: {value: fill},
+            fillOpacity: {value: fillOpacity},
+            stroke: {value: stroke},
+            strokeWidth: {value: strokeWidth},
+            strokeDash: {value: strokeDash}
+          },
+          update: {
+            path: {
+              test: true,
```
arvind, line 75: Hi @dvmoritzschoefl, what is `test: true` for here?

src/selection.ts
```diff
@@ -7,15 +7,15 @@ import {ParameterName} from './parameter';
 import {Dict} from './util';
 
 export const SELECTION_ID = '_vgsid_';
-export type SelectionType = 'point' | 'interval';
+export type SelectionType = 'point' | 'interval' | 'lasso';
```
arvind, line 10: For some context, when first designing selections, we intentionally chose `interval` to emphasize _what_ was being selected (rather `brush` which would have focused on _how_ that selection was occurring). 

I think a similar analogy applies here. Rather than `lasso` (which, to me, describes _how_ the selection is performed), I wonder if we can find names that better describe _what_ is being selected?

Off the top of my head, the closest candidate I can think of is `region`.

src/selection.ts
```diff
@@ -7,15 +7,15 @@ import {ParameterName} from './parameter';
...
 export type SelectionResolution = 'global' | 'union' | 'intersect';
 
 export type SelectionInit = PrimitiveValue | DateTime;
 export type SelectionInitInterval = Vector2<boolean> | Vector2<number> | Vector2<string> | Vector2<DateTime>;
-
+export type SelectionInitLasso = number[];
 export type SelectionInitMapping = Dict<SelectionInit>;
 export type SelectionInitIntervalMapping = Dict<SelectionInitInterval>;
-
+export type SelectionInitLassoMapping = Dict<SelectionInitLasso>;
```
arvind, line 18: Can these selections be initialized? I can imagine populating them with IDs somehow — though doing so will fall outside the scope of this PR, and is something we should think about generally for all ID-driven selections (e.g., `point` and geo-intervals). Putting that aside, I'm not sure it'd make sense to initialize the lasso specifically (i.e., pixel coordinates)? 

I'm guessing @dvmoritzschoefl, you were following a similar thought process because I don't see any additional references to these two types? So perhaps we can delete them, and make sure the lasso selection type doesn't allow `value`?

dvmoritzschoefl: True, I forgot to delete these 2.
I dont think there is a way to initialize the store anyway now, or am I missing something here?

dvmoritzschoefl:
@arvind I added runtime and compile tests for the lasso selection. I also refactored **lasso** to **region** in the comments and classes.

src/selection.ts
```diff
@@ -200,6 +198,25 @@ export interface IntervalSelectionConfig extends BaseSelectionConfig<'interval'>
...
+   */
+  mark?: BrushConfig;
+
+  /**
+   * An array of field names whose values must match for a data tuple to
+   * fall within the selection.
+   *
+   * __See also:__ The [projection with `encodings` and `fields` section](https://vega.github.io/vega-lite/docs/selection.html#project) in the documentation.
+   */
+  fields?: FieldName[];
```
kanitw, line 217: How should this property work?  Shouldn't the fields always match the x and y encodings?

kanitw (commented):
Thanks for the contribution. (Sorry for the super delayed review)

It seems like something is wrong with the generated PNGs? 

image

src/selection.ts
kanitw, line 217: How should this property work?  Shouldn't the fields always match the x and y encodings? 

It would be good to add an example for this property if it should be supported, or otherwise we should remove them.

kanitw (changes requested):
Thanks for contribution. Sorry for super delayed review. 

Let's update `select.md` to include these awesome features so people can learn about them beyond just from the example list.

src/selection.ts
dvmoritzschoefl: Hi, looking at the vega counterpart for this PR
https://github.com/vega/vega/pull/3388/files#diff-547caeeef0ceae19c314c72edcd93ab9aac4a9462ada8fbae25ec7d37fad147b

the intersection test will always use x and y coordinates, if this is what you meant.
I think I took the interval selection config as a starting point, can I just remove the fields property if it is not needed?

kanitw: > Shouldn´t the fields property be always [SELECTION_ID] like in the default config for the point selection?

Point selection indeed uses `[SELECTION_ID]` by default. However, one can configure the `fields` to be different for point selection, thus having the `fields` property for users to customize makes sense.  

So the question is whether customizing this for lasso ever make sense.  
If it should always be a constant value and can't be customized, we shouldn't expose it in the specification. 

If it can be customized, better add an example to demonstrate at least one practical use case.

arvind: @dvmoritzschoefl: the typings here generate the JSON Schema and/or serve as the public interface for what Vega-Lite users believe can be specified. You're quite right that the lasso selection's predicate uses `SELECTION_ID`, but it does not make sense to customize it to use other fields (i.e., your predicate function is hard coded to `vlSelectionIdTest`) or for customizing the field to affect how lassoing works operationally/mechanically (i.e., how we can get a uni-dimensional brush by specifying `"encodings": ["x"]` for interval selections).

So I think @kanitw is quite right that this property should be dropped here — great catch @kanitw!

dvmoritzschoefl: Ok, thanks for the clarification that makes sense. So removing both this property and the initial config for it should do the trick?

dvmoritzschoefl:
> Thanks for the contribution. (Sorry for the super delayed review)
> 
> It seems like something is wrong with the generated PNGs?
> 
> image

This seems to be the incorrect location anyway for the images. I think I generated them in the wrong place once and did not notice, the only place where those are generated is by running the runtime tests right? (Then I can just delete them and it run it again)

src/selection.ts
dvmoritzschoefl: @kanitw It seems that the CLI is generating the faulty images after a new commit. Is there an easy way to test why it is generating empty images? I tested the examples locally and they seem correct. Also the other images are correctly created in the runtime tests.

dvmoritzschoefl: @kanitw @arvind Was I correct under the assumption that the 2 images were in the wrong folder? Is it resolved then or is there another thing to do before this can be merged?

arvind:
Closing this in favor of #8988 which freshens the code, expands runtime test coverage, and fixes a few bugs along the way. 

THANK YOU @dvmoritzschoefl for all your contributions here, and I'm _very_ sorry it's taken us such a long time to actually make progress on getting it merged.

PR #8154, opened 2022-05-10
feat: enable default tooltip for boxplot outliers

joelostblom:
This is a bit of a issue/discussion, but I wanted to show which change I think will work so I made a PR right away. I think it is quite useful to see hover info on boxplot outliers to get more into about these points. Having this enable by default facilitates getting this info right away without adding a tooltip encoding and is in line with the default tooltip added for the box and whiskers. One drawback could be that the tooltip gets cluttered if there are a lot of columns in the dataframe, and that it might not be possible to turn it off completely (but this is also the case for the box and it is easy to set it to a single field).

Please:

- [ ] Make the pull requests (PRs) atomic (fix one issue at a time). Multiple relevant issues that must be fixed together? Make atomic commits so we can easily review each issue.
- [ ] Provide a concise title as a [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties") so we can easily copy it to the release note.
- [ ] Lint and test (Run `yarn test`).
- [ ] If you send a pull request from a fork, make sure that GitHub actions run successfully. Make sure to add a [`GH_PAT` secret](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) for a [personal access token](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token) with public repository permissions.
- [ ] Review your changes before sending the PR (to ensure code quality).
- For new features:
  - [ ] Add new unit tests.

Tips:

### DIFF
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -289,7 +289,7 @@ export function normalizeBoxPlot(
 
   const outlierLayersMixins = partLayerMixins<BoxPlotPartsMixins>(markDef, 'outliers', config.boxplot, {
     transform: [{filter: `(${fieldExpr} < ${lowerWhiskerExpr}) || (${fieldExpr} > ${upperWhiskerExpr})`}],
-    mark: 'point',
+    mark: {"type": "point", "tooltip": {"content": "data"}},
     encoding: {
       [continuousAxis]: {
         field: continuousAxisChannelDef.field,


domoritz (approved):

src/compositemark/boxplot.ts
```diff
@@ -289,7 +289,7 @@ export function normalizeBoxPlot(
 
   const outlierLayersMixins = partLayerMixins<BoxPlotPartsMixins>(markDef, 'outliers', config.boxplot, {
     transform: [{filter: `(${fieldExpr} < ${lowerWhiskerExpr}) || (${fieldExpr} > ${upperWhiskerExpr})`}],
-    mark: 'point',
+    mark: {type: 'point', tooltip: {content: 'data'}},
```
kanitw, line 292: While I think this is useful, I think this makes the tool behave inconsistently. 

Personally I think tooltip is useful to be turned on by default in general, but I think we previously turned it off since it's distracting for some use cases.

Since we don't turn it on by default elsewhere, it's weird that this is the only place we turn on by default. 

Instead of turning tooltip on always, we should read boxplot's tooltip property and turn it on based on the specified tooltip property?

kanitw (changes requested):

src/compositemark/boxplot.ts
joelostblom: Hello @kanitw , just a ping on my latest comment here, what do you think?

kanitw: I think it's important that we respect the tooltip property (which is default to false), so I'd prefer that we read the tooltip property.

domoritz: It turns out we have a default tooltip for box plots already, which as @kanitw said should not be the case. Let's fix this behavior and then also add tooltips for the outliers.

joelostblom: I looked a bit more at this today and it seems like all composite marks in Vega-Lite currently have the tooltip on by default, see for example:

- https://vega.github.io/vega-lite/examples/layer_point_errorbar_ci.html
- https://vega.github.io/vega-lite/examples/layer_line_errorband_ci.html

I personally find that useful and it makes sense to me that boxplots also have a default tooltip since it is a composite mark. I can also see an argument for always having the tooltip off by default for all composite marks (although I personally find it more useful to have them on as they are currently), but I also think it makes sense to treat these differently from regular marks and there is an argument to be made that tooltips are more needed for complex marks. In either case, a change removing all tooltips from all composite marks seems to go beyond the scope of what I am trying to fix in this PR (which is consistency within the boxplot itself and making sure the box and outlier points are treated the same).

PR #8156, opened 2022-05-11
docs: Explicit temporal type

danmarshall:
Since this spec is used as the example for the [temporal encoding type in the docs](https://vega.github.io/vega-lite/docs/type.html#temporal) it's nice to have it spelled out.


domoritz:
Merging in https://github.com/vega/vega-lite/pull/8352

PR #8189, opened 2022-05-23
feat: export `defaultTitleFormatter` as a top-level export

jkillian:
Exports `defaultTitleFormatter` from the top-level of the `vega-lite` package. This makes it easier for people using the [`fieldTitle` option of `compile` ](https://vega.github.io/vega-lite/usage/compile.html#field-title) to write a custom implementation that uses the default definition as a starting point.

Does this seems like a reasonable change? Let me know if there are any other changes also needed here!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,7 @@
 import pkg from '../package.json';
 export const version = pkg.version;
 
+export {defaultTitleFormatter} from './channeldef';
 export {compile} from './compile/compile';
 export type {Config} from './config';
 export {normalize} from './normalize';


domoritz:
Maybe it's better to have you import from the file directly since we may change the location of the method in the future and I don't want to promise that we don't change this method. Would that work for you? 

```
import {defaultTitleFormatter} from `vega-lite/src/channeldef`;
```

jkillian:
I think a submodule import like that is reasonable, I've just been unable to get things to work in our node.js v16 backend. 

Our codebase is in TypeScript, but we compile it down to commonjs modules via TS before running. If I attempt to import the published channeldef ES module, with something like `import {defaultTitleFormatter} from "vega-lite/build/src/channeldef.js;"` then gets compiled down to something like `const channeldef_js_1 = require("vega-lite/build/src/channeldef.js");` This file is not a CJS module though and node throws an error since we're importing a ES module with `require`:

```
./project/node_modules/vega-lite/build/src/channeldef.js:12
import { isArray, isBoolean, isNumber, isString } from 'vega-util';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

So, in brief, because our application is CJS based, I can't figure out any way to consume the published ES module submodules. Perhaps it would be possible to publish CJS modules in addition to the ES modules?

(Though I do think `defaultTitleFormatter` would maybe be a reasonable top-level export! Since `fieldTitle` is an option for `compile`, it would be nice to have some way of getting at the default title in there. Alternatively, perhaps, the default title could be passed to the `fieldTitle` callback? But that's a bigger API change.)

domoritz:
Can you run Babel to convert esm to cjs? I don't think I want to add the export right now unless there is more need for it.

jkillian:
Yep, I can attempt something like that, I definitely understand the hesitation to add more complexity to your build process. 

I do wonder if there's something that should be slightly tweaked in your process though? I _think_ (if I understand correctly) for correct NodeJS semantics, ES modules should either have a `.mjs` extension or the `package.json` should contain [a `"type": "module"` field](https://nodejs.org/api/packages.html#type). Part of the difficulty here might come from the fact that the published `.js` files use ES module syntax but aren't known to NodeJS to be ES modules.

domoritz:
Yep, that could be the case. We'd need to audit all Vega packages to make sure they are consistent. I'm all for cleaning that up but it's a tricky thing to get right (I recently spent a few days making Apache arrow as close as I think we can get to working well with different bundles).

domoritz:
Thanks. ESM is pretty good so you can look into adopting it if that's an option.

PR #8225, opened 2022-06-13
docs(examples): Bar Chart with Full-Height Hover Targets for Tooltip

kanitw:
image

<details>
  <summary>📦 Published PR as canary version: <code>5.2.1--canary.8225.9a33014.0</code></summary>
  <br />

  ```bash
  npm install vega-lite@5.2.1--canary.8225.9a33014.0
  # or 
  yarn add vega-lite@5.2.1--canary.8225.9a33014.0
  ```
</details>


examples/specs/interactive_bar_full_height_hover.vl.json
```diff
@@ -0,0 +1,30 @@
...
+        }
+      ],
+      "mark": {"type": "bar", "color": "#eee", "tooltip": true},
+      "encoding": {
+        "x": {"bin": true, "field": "IMDB Rating"},
+        "opacity": {
+          "condition": {"test": {"param": "hover", "empty": false}, "value": 0.5},
+          "value": 0
+        },
+        "detail": [{"aggregate": "count"}]
```
domoritz, line 19: This feels odd to me to prevent many bars per x. I wonder whether a distinct transform would be cleaner.

domoritz (commented):
I love the example but the implementation doesn't make me happy.

examples/specs/interactive_bar_full_height_hover.vl.json
kanitw: I don't understand your point here.  
This is a histogram, so there is only one bar per x and I need to add this line, so the count shows up on the tooltip.

kanitw: Now that I say it, I'll rename the example name to histogram.

domoritz: Oh I see. I didn't realize that was for the tooltip. All good then.

kanitw:
Merging since Dom is approving in the text.

PR #8242, opened 2022-06-23
chore: publish script is not just for prereleases and canaries

domoritz:
<details>
  <summary>📦 Published PR as canary version: <code>5.2.1--canary.8242.8790c53.0</code></summary>
  <br />

  ```bash
  npm install vega-lite@5.2.1--canary.8242.8790c53.0
  # or 
  yarn add vega-lite@5.2.1--canary.8242.8790c53.0
  ```
</details>

### DIFF
--- a/.github/workflows/publish-to-npm.yml
+++ b/.github/workflows/publish-to-npm.yml
@@ -1,4 +1,4 @@
-name: Publish to NPM (canaries and prereleases)
+name: Publish to NPM
 
 on:
   push:


hydrosquall (approved):
The naming was from an earlier version of the script, where I only wanted to do some tasks on pre-release and canary, and some tasks only on stable. 

This was specifically when we were concerned about only publishing the docs + new schema only for major changes. That is no longer relevant, since we opted in https://github.com/vega/vega-lite/pull/7984 to put those "stable release only" tasks into a separate workflow that is triggered based on creating releases (which are created by Auto) rather than by checking the branch name.

https://github.com/vega/vega-lite/blob/next/.github/workflows/release-docs-and-schema.yml

In any case, changing this task name makes sense.

Issue #8300, opened 2022-07-18
Expose `normalizedNumberFormat` for normalized % stack


kanitw:
For normalized stacked charts (like [bar](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICONjgaC17WVJnyAvkszaksiADM22hGgDaoNzSCmHDa+qqYDAiUAJ5wSNoABAC8SQkATAAMWSAOoFBIglDmanBhapGUEHAAHsmpaQkA-AkA5ABiiAVwLQmorQCyXS2KIEgQ+mRwssohOQC6SsjaANZooJjROKXoTPEjmGxsQTQ4aE4McA4gAO40yvRrIBDBp6gAjADsV1NQbMo0sjIj2ijyQZDI2j4wX0EEiIwCcEEyn0W1MglKjjo6JRpmKmCsI2eSCgq3Qsk8yGEAC9Sldqo8EUj9GDaUpfoJPAyaIjkehJtNZkpoF1Hs5AdsfCAAMQANg+YjeAE4AMwjKX5AAcGplbxAczsBrsQA)), we automatically generate tooltips that calculates ratio / %.

<img width="445" alt="image" src="https://user-images.githubusercontent.com/111269/179632912-9e509c03-62ff-4594-ab9e-0b06514b8a93.png">

However, we haven't applied any % formatting to this % tooltip yet. 

We should

1) Expose `config.normalizedNumberFormat` (or `percentFormat` or sth) for setting formatter for this % tooltip
  - Probably also add `config.normalizedNumberFormatType` so this `config.normalizedNumberFormat` has its own format type. 
2) Set the default `config.normalizedNumberFormat` (`defaultConfig` in `config.ts`) ---  probably  to`.2%`?

For testing, perhaps:

1) modify `stacked_bar_normalize.vl.json` to include `tooltip: true` in mark to test the default format
and 
2) add `stacked_bar_normalize_custom_format.vl.json` to test customized format in the config.

kanitw:
cc: @domoritz @jheer @arvind 

Any opinion for the config name?

[...]

PR #8303, opened 2022-07-20
Feat: allow for `normalizedNumberFormat` for `stack: "normalize"` tooltips

lsh:
This PR addresses #8300. As mentioned in the referenced issue, the spelling of the `normalizedNumberFormat` field may be subject to change (per feedback of @domoritz @jheer).

**Important**: The `normalizedNumberFormat` should only be applied to tooltips that are `stack: "normalize"`. Furthermore, this PR intentionally does not implement the `normalizedNumberFormat[Type]` encoding for `src/compile/legend/encode.ts` because @kanitw and I could not think of a non positional situation where `stack` is used.


lsh:
Note: this PR will either need to rebase if #8302 is merged first, or vice versa.

kanitw:
> @kanitw and I could not think of a non positional situation where `stack` is used.

Looking at the code more, it's also worth noting that `stack?:` is defined in `PositionBaseMixins` and thus won't be used by non-position encoding.

kanitw (commented):
Add some tests so codecov won't complain?

lsh:
Closing in favor of #8307

PR #8349, opened 2022-08-09
fix: fix thin bar problem (#8350)

kanitw:
- https://github.com/vega/vega-lite/issues/8350 Fix thin bar problem 

See the added `bar_group_thin` example before and after the behavior fix commit. 

(Other Vega diffs do not produce image diffs, so this shouldn't introduce any regression.) 

<details>
  <summary>📦 Published PR as canary version: <code>5.4.1--canary.8349.4227eff.0</code></summary>
  <br />

  ```bash
  npm install vega-lite@5.4.1--canary.8349.4227eff.0
  # or 
  yarn add vega-lite@5.4.1--canary.8349.4227eff.0
  ```
</details>


src/compile/mark/encode/position-rect.ts
```diff
@@ -156,7 +162,8 @@ function positionAndSize(
     If band is 0.6, the the x/y position in such case should be `(1 - band) / 2` = 0.2
    */
 
-  const defaultBandAlign = scale?.get('type') !== 'band' || !('band' in sizeMixins[vgSizeChannel]) ? 'middle' : 'top';
+  const defaultBandAlign =
+    scale?.get('type') === 'band' && isRelativeBandSize(bandSize) && !hasFixedSizeMixins ? 'top' : 'middle';
```
yhoonkim, line 166: Can you explain if 
`('band' in sizeMixins[vgSizeChannel])` is equivalent `isRelativeBandSize(bandSize) && !hasFixedSizeMixins`?

I wonder if this logic change is correct and necessary

yhoonkim (approved):
Made a minor comment due to my lack of the context. Feel free to merge if you are sure about it.

src/compile/mark/encode/position-rect.ts
kanitw: Yeah, since we output `{signal: max(0.25, bandwidth('x'))}`  instead of `{"scale": "x", "band": 1}`, the old logic that checks for "band" doesn't work anymore. 

I tracked how `band` ` in `sizeMixins[vgSizeChannel]` was generated and found that  

`band` in `sizeMixins[vgSizeChannel]` only happens if the "bandSize" is relative and there is no size from mark/encoding (i'll rename the variable to `hasSizeFromMarkOrEncoding`).

Issue #8324, opened 2022-07-27
Allow config to set default scale.zero per marktype


yhoonkim:
`scale.zero` is `true` as default for x and y channels if the quantitative field is not binned and no custom domain is provided; false otherwise.   (https://vega.github.io/vega-lite/docs/scale.html )
Though, for some range-based-marktype (e.g., point, box plot). `zero: true` might not be useful default; if the data are mostly distributed 10M~11M and if the scale includes zero, the point marks for those data will look like just a single dot. 

[Even the example scatter plot in VL editor uses zero: false](https://vega.github.io/editor/#/examples/vega-lite/point_color_with_shape). If we delete it, it looks like this:
![image](https://user-images.githubusercontent.com/6269456/181369985-d479e653-83cf-4660-b329-0be553570673.png)

--
**Suggestion**
Assuming that many VL users may be already used to the current default, I would like to suggest adding `config.scale.zero` to change the default instead of changing system default.

```json
{
  "config": { "scale": {"zero": "always" } } 
}
```
-> Keep the system defulat: `scale.zero` is `true` as default for x and y channels if the quantitative field is not binned and no custom domain is provided; false otherwise.   

```json
{
  "config": { "scale": {"zero": "range-mark-only" } } 
}
```
-> `scale.zero` is `true` as default for y(/x) channel if the quantitative field is not binned and no custom domain is provided **and it is vertical (/horizontal) bar and area**; false otherwise.  

CC; @kanitw

kanitw:
For implementation, you can look at `zero()` in `src/compile/scale/properties.ts`.

kanitw:
Btw, look at existing behavior, I just realize that "always" isn't very accurate. 

So maybe:

```
zero: "all-marks" | "bar-and-area-only" 
```

kanitw:
cc: @domoritz @arvind  @jheer  any opinion about the config name?

domoritz:
If we only have two options, I wonder whether we could consider a Boolean. 

Otherwise I think all marks, and range marks makes sense. I don't think we need to day range marks only.

kanitw:
FWIW, https://github.com/vega/vega/issues/1471 is related to this. 
(Adding dynamic zero with threshold is probably gonna be a better default than always off.)

In terms of sequencing, I think add this config first would be easier (since it's simpler).

kanitw:
Some more thoughts: 

[...]

PR #8354, opened 2022-08-11
feat: Allow config to set default scale.zero per marktype

yhoonkim:
#8324 Allow config to set default scale.zero per marktype.

- config.scale.zero is true as default.
- When `config.scale.zero===false`, zero is false as default except for the non-ranged bar/area chart.

<details>
  <summary>📦 Published PR as canary version: <code>5.5.1--canary.8354.6729853.0</code></summary>
  <br />

  ```bash
  npm install vega-lite@5.5.1--canary.8354.6729853.0
  # or 
  yarn add vega-lite@5.5.1--canary.8354.6729853.0
  ```
</details>


src/compile/scale/properties.ts
```diff
@@ -120,7 +120,8 @@ function parseUnitScaleProperty(model: UnitModel, property: Exclude<keyof (Scale
                 domainMax: specifiedScale.domainMax,
                 markDef,
                 config,
-                hasNestedOffsetScale: channelHasNestedOffsetScale(encoding, channel)
+                hasNestedOffsetScale: channelHasNestedOffsetScale(encoding, channel),
+                hasRangeChannels: encodingHasRangeChannels(encoding)
```
kanitw, line 124: The fact that `encodingHasRangeChannels(encoding)` doesn't take in `channel` seems alarming to me. 

Basically, Y2 shouldn't affect x-scale's zero.  

I think you should 
- undo adding `encodingHasRangeChannels`
- rename `hasRangeChannels` to `hasRangeChannel` and just check if there is x2 for x or y2 for x.  (And maybe radius2 for radius.)

site/docs/encoding/scale.md
```diff
@@ -444,7 +444,7 @@ To provide themes for all scales, the scale config (`config: {scale: {...}}`) ca
 
 #### Padding
 
-{% include table.html props="bandPaddingInner,barBandPaddingInner,rectBandPaddingInner,bandWithNestedOffsetPaddingInner,offsetBandPaddingInner,bandPaddingOuter,bandWithNestedOffsetPaddingOuter,offsetBandPaddingOuter,continuousPadding,pointPadding" source="ScaleConfig" %}
+{% include table.html props="bandPaddingInner,barBandPaddingInner,rectBandPaddingInner,bandWithNestedOffsetPaddingInner,offsetBandPaddingInner,bandPaddingOuter,bandWithNestedOffsetPaddingOuter,offsetBandPaddingOuter,continuousPadding,pointPadding,zero" source="ScaleConfig" %}
```
kanitw, line 447: Why is `zero` added to the padding table?  This seems wrong?

src/scale.ts
```diff
@@ -415,6 +415,14 @@ export interface ScaleConfig<ES extends ExprRef | SignalRef> {
    * Reverse x-scale by default (useful for right-to-left charts).
    */
   xReverse?: boolean | ES;
+
+  /**
+   * Override the default `scale.zero` for [`continuous`](https://vega.github.io/vega-lite/docs/scale.html#continuous) scale except for non-ranged bar or area charts.
```
kanitw, line 420: > [`continuous`](https://vega.github.io/vega-lite/docs/scale.html#continuous) scale

scale**s**

kanitw: > except for non-ranged bar or area charts.

If you agree with the other comment to always do "zero" for size scale. 

except for (1) x/y-scales of non-ranged bar or area charts and (2) size scales.

src/compile/scale/properties.ts
```diff
@@ -418,18 +422,21 @@ export function zero(
     }
   }
 
-  // If there is no custom domain, return true only for the following cases:
+  const defaultZero = scaleConfig?.zero === undefined ? true : scaleConfig?.zero;
```
kanitw, line 425: Instead of adding this line, probably cleaner to add `zero: true` to `defaultScaleConfig`.

kanitw (changes requested):

kanitw (approved):

Issue #5013, opened 2019-06-01
Custom sort order not respected in ordered plot 


kanitw:
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v3.json",
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"a": "A","b": 28}, {"a": "B","b": 55}
    ]
  },

  "layer": [{
    "mark": "rule",
    "encoding": {
      "x": {"field": "a", "type": "ordinal", "sort": ["B", "A"]},
      "y": {"field": "b", "type": "quantitative"}
    }
  },{
    "mark": "circle",
    "encoding": {
      "x": {"field": "a", "type": "ordinal"},
      "y": {"field": "b", "type": "quantitative"}
    }
  }]
}
```

![image](https://user-images.githubusercontent.com/111269/58753677-fcdfa280-8477-11e9-80ad-d089ee1ec0a9.png)

The order should be "B, A", not "A, B". 

A similar problem can be seen in https://github.com/vega/vega-lite/pull/5012.

domoritz:
Is this a duplicate of https://github.com/vega/vega-lite/issues/4932?

domoritz:
Or of https://github.com/vega/vega-lite/issues/3153?

kanitw:
They are all similar but slightly different:

- #3153 uses `domain` in the spec, this issue uses `sort`.
- #4932 specifies `sort` in all layers, this issue specifies only one.

kanitw:
I think the solution for this particular one is to consider `sort` as one of the component (or part of the scale component) that gets respected when we later derive the CalculateNode (for sort index).

kanitw:
If we extend Vega domain sort to support either (a) sorting by a calculated field based on each unique value (b) custom sort order. 

Then such domain `sort` could be merged and we can easily fix this issue.

ben741:
I'm encountering something similar, where adding "color" to an encoding within a layered plot breaks sorting. I'm not sure if it has the same root cause as described here, or whether it should be treated as a separate issue.

 Here's my code, where deleting `"color": {"field": "highlight"},` causes the sort to work properly:
```json
{
  "data": {
    "values": [
      {"name": "a", "val": 2, "highlight": 1, 
       "lower": 1.5,"upper": 2.5},
      {"name": "c", "val": 1, "highlight": 0,
      "lower": 0.5,"upper": 1.5},
      {"name": "b", "val": 3, "highlight": 0,
[...]

PR #8567, opened 2022-11-23
fix: correctly handle merging of domains with more than one sort object

MufaroMakiwa:
Partially addresses https://github.com/vega/vega-lite/issues/5013

### DIFF
--- a/test/compile/scale/domain.test.ts
+++ b/test/compile/scale/domain.test.ts
@@ -1085,6 +1085,38 @@ describe('compile/scale', () => {
     );
 
     it(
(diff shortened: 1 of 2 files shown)


test/compile/scale/domain.test.ts
```diff
@@ -1084,6 +1084,38 @@ describe('compile/scale', () => {
       })
     );
 
+    it(
+      'should warn if sorts conflict even if one sort object is not default',
```
jonathanzong, line 1088: how about `should use non-min aggregation when sort ops conflict`

PR #8601, opened 2022-12-17
docs: Add example of color legend that can be changed by color picker input.

dwootton:
Added a colorpicker input example that allows user to dynamically change the color of their color scales.


dwootton:
Adding changes to site as well:

domoritz:
Thank you

Issue #7854, opened 2021-12-14
duplicate parameter definition error in layered charts (because of `"point": true`)


mattijn:
Original observed in this Altair issue comment: https://github.com/altair-viz/altair/pull/2528#issuecomment-992989438.
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgDghgThC2DOIBcBtUA7OBTZIAWAlgOZ4A2xeALiADQjxalYDG1SolAnmDkuAPYF01OsybRcsfgFcGMkSH7pJMuQDcsUEAF9tAXTprmS5hDZp8WCmwCMABjt1Y0ANbIO3XiHLocdMILCyABmEKQM2nRY6MYAJkJE7iAAHknBBIyxuLFmfiCUBLBYAKroBGwgnFgSdFw8uOj8sEJhOnScSRBERFBYRLmS1cp06Zm4YFAEzHl1XgCO0hDC5WYEGjq6NKDOUG7s+Z64vaxtINFxCWkEpKSd3b39lF5FS7Qgo6RZfBNTM4d8CyWBUoq3WkXolCg-BcvFAxnQ8QKSiSajC0i8ACNSBBmG5-NA4LhCCRyCQFFhYGAuCEwhFDGjeOhpDdwal9h8viAck83gUiqVyrgqjUDvU+E9KfwYLddPo6NyIElpFBbnwFQB6eCUfi4+AAOmY8DUpwAJPBmHgKYq+FRKGB4Eh1eqNP09URynhpBi9QR+JqLVbnX0IABachPZ0AVj1ACY9XY9QAreDI7RAA)

```json
{
  "params": [
    {
      "name": "highlight",
      "select": {"type": "point", "clear": "mouseout", "on": "mouseover"}
    }
  ],
  "vconcat": [
    {
      "height":100,
      "mark": {"type": "line", "point": true},
      "encoding": {
        "x": {"field": "date", "timeUnit": "year", "type": "nominal"},
        "y": {"aggregate": "mean", "field": "price", "type": "quantitative"}
      }
    },
    {
      "mark": {"type": "rect"},
      "encoding": {
        "fill": {"aggregate": "mean", "field": "price", "type": "quantitative"},
        "stroke": {
          "condition": {"value": "black", "param": "highlight", "empty": false},
          "value": null
        },
        "x": {"field": "date", "timeUnit": "year", "type": "temporal"}
      }
    }
  ],
[...]

PR #8662, opened 2023-01-21
fix: move name from layer chart to one of the inside layers

ChristopherDavisUCI:
This is an attempt to fix part of #7854, specifically, the part mentioned in https://github.com/vega/vega-lite/issues/7854#issuecomment-1315551841

In @mattijn's example I linked, the spec gets copied behind the scenes to form a layer chart (because of the `"point": true`).  Because the spec has a `name` property, that also gets duplicated.  Here we prevent the `name` property from getting duplicated.

I'm very happy to hear whether or not this is a reasonable solution.  Thank you for your time!


domoritz:
Thank you. Makes sense to me.

Issue #8446, opened 2022-10-04
chart with repeat and a defined name disables the independency of the repeated channel (eg y-axis)


mattijn:
When a `name` is defined within a `layer` within a `repeat` `spec` as such:
```json 
 "repeat": ["temp_max", "precipitation", "wind"],
  "spec": {
    "layer": [
      {
        "name": "FOO",
```
Than the axis of the chart get confused, see the following screenshot:
<img width="821" alt="image" src="https://user-images.githubusercontent.com/5186265/193812425-1cfd2549-ff79-41dd-9dce-e9a97e82816d.png">
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICAdzhr42ylAjEQAXyXa4OM5jQBtEJkQ4A+sgAPRRAcJygaHDo1dnklYxpZZRAAXSUIFyg0UEEkAE84bQ9QWSQEOH0AMQB5KuDkbQBrLK9cl31w7ShBcocQOFkoNmUEsmagzRAAMxo4QST0VW9gzBoygFVZOn0EOXp7JVzm6dn50CcXNX1z132QJDIyJwol9DKkeV7BwTZCieO5-TfKDROS3CA0ABe5QmxCQggY0LEAAY7KiFKB6k0JphWtCQMJZD0lP1BsNZKMJuNQP95gYlkoVutNm5XrsYLdDn8ZgCJtdLug+W5evdHnwXiA3h8lF8fkdubSgSCPqjkqigA)

VL-spec:
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "data": {"url": "data/weather.csv"},
  "repeat": ["temp_max", "precipitation", "wind"],
  "spec": {
    "layer": [
      {
        "name": "FOO",
        "mark": {"type": "circle"},
        "encoding": {
          "x": {"field": "date", "timeUnit": "month"},
          "y": {"field": {"repeat": "repeat"}, "aggregate": "mean"},
          "color": {"field": "location"},
          "size": {"value": 50}
        }
      },
      {
        "mark": {"type": "line"},
        "encoding": {
          "x": {"field": "date", "timeUnit": "month"},
[...]

Issue #8348, opened 2022-08-08
repeat chart with top level defined brush selection corrupts cross-view interaction


mattijn:
This issue is linked to this comment: https://github.com/vega/vega-lite/issues/8230#issuecomment-1201358652.
Example is based on a repeated chart such as this example: https://vega.github.io/vega-lite/examples/interactive_layered_crossfilter.html

Upon manually defining a `"name":"CHART"` property for the view defining the `param` + `filter` the units-field in the `brush_store` is fixed to the name `"CHART"`, as can be seen in the data-viewer below. No matter which view is being brushed. This makes the brush selection corrupt cross-view.

<img width="700" alt="image" src="https://user-images.githubusercontent.com/5186265/183510923-eefe892e-a9ec-4e04-916e-d644d004efc3.png">

Spec that errors:
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICAM2FksEfgCYA1lJnylxttuSY0oHEm0Q4Hg5h+6KqBIAC+EUqY2kiyEM6uaADaoFBIglAMgmpBGGw6EAAUIQwIlCFwAJSKIEgQ+pg0CH5hALpK2nA4cGr+UGyCpfKoSSo0EOqyUH5KynDZAJ41jc0grWFKEN1Q-gDuNMr0aACMYgAMSvA05u6ox2cXIItw2smgyNo2+kzeNXBTbGUNFkZH8AA9-MYaPNlP5Ot1euh+oMEPINiAmMD-MgwZi4mgrGdIiAlppamQyJ0KKEkflZO50ciXP5iOkGLkAMQANh54Uinm8SAQ9RGoFkQtyTG0DAgMBqvkEcCgt1AmAW3X0wMC2lZeiU-36QJBItGEPWbSiMTiCQQbxAUME2v8XhitvQUplcoi7RA4tW6AAwgAJACCACUACo1D5fd2-fUAo2gskQslQmFwro9W4gZFDcJKPHYpC44EiwnE0mgJAUqk5fT9Bj0gu5gbMsm69n6FyxMgtb0RIA)

* * *

Comparison to _not_ using a manual defined `name` property:

[...]

PR #8663, opened 2023-01-22
fix: Add prefix to names in a layer

ChristopherDavisUCI:
This is an attempt to resolve the issues @mattijn raised in #8348 and #8446.  (I believe this PR is independent from my other recent PR, #8662.)

I'm very happy to hear comments:
* Is this approach, making sure there are no duplicate names, the correct approach?
* Is this the right place to make the change?
* Is this a reasonable name format (probably not): `params.repeaterPrefix + spec.name`
* Is there a more natural way to write this logic in JavaScript? 
```
name: spec.name && params.repeaterPrefix ? params.repeaterPrefix + spec.name : undefined
```

A test is currently failing because of this line:
https://github.com/vega/vega-lite/blob/4556109808e023eb2e53507d46c1792725e382b4/test/normalize/toplevelselection.test.ts#L89
Is it correct to have both `normalized.concat[0].layer[0].name` and `normalized.concat[1].layer[0].name` equal to `"a"`?  This PR would change that, so it would break this test.

Thanks again for any comments!


ChristopherDavisUCI:
Hi @domoritz, thanks for the reviews of my other PRs.  I changed this one to "ready for review" also, although please see my many questions in the [original comment](https://github.com/vega/vega-lite/pull/8663#issue-1552112691) above.

src/normalize/core.ts
```diff
@@ -65,6 +65,7 @@ export class CoreNormalizer extends SpecMapper<NormalizerParams, FacetedUnitSpec
 
     const specWithReplacedEncoding = {
       ...spec,
+      name: spec.name && params.repeaterPrefix ? params.repeaterPrefix + spec.name : undefined,
```
domoritz, line 68: This will remove the name when there is no repeater prefix, which is definitely wrong (see changed example in this diff).

domoritz (changes requested):

ChristopherDavisUCI:
Thanks @domoritz for proposing the fix in #8733!  I will close this, but please let me know if I misunderstood anything.

PR #8688, opened 2023-02-01
docs: remove the duplicate 'Expr' on the sidebar

HaveF:
I find "Expr" on the sidebar of doc is duplicate with its parent 'Parameter' link.

No evidence was found that this was done intentionally. So I just remove it.

image

image

### DIFF
--- a/site/_layouts/docs.html
+++ b/site/_layouts/docs.html
@@ -146,9 +146,6 @@ sidebar:
     sub-sidebar:
       - text: Value
         url: value
-      - text: Expr
-        url: parameter
-        anchor: '#expr'
       - text: Bind
         url: bind
       - text: Select


domoritz:
Is this file the autogenerated one?

Please use a semantic commit message for the pr title.

HaveF:
Ok, I refine the PR message. Do you think that is ok? 

I check the code, I find `docs_toc.md` is auto-generated by script, so I remove it.

domoritz:
Hmm, I don't think they are the same. The latter one links to a specific section #expr (or at least is supposed to I think).

HaveF:
`docs_toc.md` is generated from `site/_layouts/docs.html` by [scripts/generate-toc](https://github.com/vega/vega-lite/blob/next/scripts/generate-toc#L13)

PR #8690, opened 2023-02-02
docs: fix overloaded url for value.html

domoritz:
https://vega.github.io/vega-lite/docs/value.html was used for two pages (value on encodings and in params) and so one isn't accessible.


domoritz:
Merging since I think this is the correct fix. @kanitw @arvind feel free to follow up.

Issue #8523, opened 2022-10-29
repeat operator with defined width and height does not accept layer


mattijn:
This spec [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgTgpgDhCGAuIBcoA2sCeEzINogFUBlAfQHEwB7AZ2pABoQB1SsVAEwHcBLdicqrRABdRgGNKqAK4BbAHZ4QASQCyAEQBCJAEoJucgOYMQ2yvHgQ5JACqUZCShGo69hkQF9G1GGOSge7PAAFsgATAAM4YxBENwGQYhIAIwAzFEg7AiwfiBSbMggCfBQ1EgA9GVi7HIAdABW1Hyo3ABuYDVyEPBlclAyZS0QBrAAtJnwsNRd1AACLUk1oQCcNeFl47BlMpQt3E711JQKniD2YADWBc2dxpYS7PpGKCASqKw547I5kDAIV5jYEDuE4ADxyACN9Mh4GApBBGAAzPYcb7QOCJF6SWTHE4YHKwAwGSDDCwFGRwBSI5HsVG-DHoLA4E7wbjwVAQAoqCkAAkoCO5xG5sDk7G5LDYXF4EG5FBodGBJwAJNQxDF7AUiiVygMhrAagZWUEpOCatxKGUVWrNoNhiNmhYBgBWRarA5HIFAA):
```json
{
  "repeat": {
    "layer": ["US_Gross", "Worldwide_Gross"],
    "column": ["IMDB_Rating", "Rotten_Tomatoes_Rating"]
  },
  "spec": {
    "width": 200,
    "height": 130,
    "data": {
      "url": "https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/movies.json"
    },
    "mark": "line",
    "encoding": {
      "color": {"datum": {"repeat": "layer"}},
      "x": {"bin": true, "field": {"repeat": "column"}},
      "y": {
        "aggregate": "mean",
        "field": {"repeat": "layer"},
        "title": "Mean of US and Worldwide Gross"
      }
    }
  },
  "$schema": "https://vega.github.io/schema/vega-lite/v5.2.0.json"
}
```
Renders correctly. But officially the `layer` cannot be defined, since the schema does not accept it:

<img width="476" alt="image" src="https://user-images.githubusercontent.com/5186265/202025352-5c7c1f08-a22f-4ea7-8cb0-a42fc056a0b8.png">

[...]

PR #8676, opened 2023-01-28
fix: Allow width and height in LayerRepeatSpec

ChristopherDavisUCI:
This is an attempt to fix the issue raised in #8523 by allowing the `spec` parameter to be a `UnitSpecWithFrame`, in addition to the current options of `LayerSpec` and `UnitSpec`.

Here is an example of a spec that raises warnings currently but does not raise a warning in this PR.  (I deleted the `$schema` parameter from Mattijn's original spec.)  [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgTgpgDhCGAuIBcoA2sCeEzINogFUBlAfQHEwB7AZ2pABoQB1SsVAEwHcBLdicqrRABdRgGNKqAK4BbAHZ4QASQCyAEQBCJAEoJucgOYMQ2yvHgQ5JACqUZCShGo69hkQF9G1GGOSge7PAAFsgATAAM4YxBENwGQYhIAIwAzFEg7AiwfiBSbMggCfBQ1EgA9GVi7HIAdABW1Hyo3ABuYDVyEPBlclAyZS0QBrAAtJnwsNRd1AACLUk1oQCcNeFl47BlMpQt3E711JQKniD2YADWBc2dxpYS7PpGKCASqKw547I5kDAIV5jYEDuE4ADxyACN9Mh4GApBBGAAzPYcb7QOCJF6SWTHE4YHKwAwGSDDCwFGRwBSI5HsVG-DHoLA4E7wbjwVAQAoqCkAAkoCO5xG5sDk7G5LDYXF4EG5FBodGBwKAA)

Thanks for looking and for any feedback!

PR #8723, opened 2023-02-20
fix: allow setting width/height in repeated layers

domoritz:
Replaces https://github.com/vega/vega-lite/pull/8676
Fixes https://github.com/vega/vega-lite/issues/8523


arvind (approved):
LGTM

PR #8733, opened 2023-02-20
fix: add repeat prefix to name in normalization

### LINKED ISSUE #8446: chart with repeat and a defined name disables the independency of the repeated channel (eg y-axis)

domoritz:
fixes #8446

replaces #8663


domoritz:
@arvind I am a bit lost here what the expected behavior is. In https://github.com/vega/vega-lite/blob/ba46bd2b538be46686e90d8d6da032d160fac6e6/test/normalize/toplevelselection.test.ts#L62, we assume that the names of repeated specs stay as they are. However, if we look at a simple spec without layer in repeat as in [Open the Chart in the Vega Editor](http://localhost:8080/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICCNsRpwIUmfIC+S7XBxw1aUFDaCGCeagDaIAFUAZQACAHFtNggIRRAAdTZdZQB3GmU4cMjokABdOxAIJyhXEFkkBDh9ADM2NljkbQBrfWFZSqU4WXdlGlkyEoAPEqZetExtBjglKvNBZX0ASQBZABEAIRCAJTVe-qVMAE8nfQBHBiRZTDod0hB8g5KkMjIHCkxK9AqL2Jm4OZKHE4XOh3J5vHd9kcPiAzhcruorrd9nRBNCls5ZCE2FUQsEQhdlCEEklUulMlEYvlQYkSqpMF4AY5nJh9KCvLZIcd0LI2Ahekg9DYhTYgA), we seem to be changing the name of the view to be unique. I can make the names unique but that breaks the nested name handling in params.

Screenshot 2023-02-20 at 18 50 17

In this pull request, I updated the test to expect specific names.

arvind:
Thanks for pinging and issuing this PR! As far as I can tell/page in, I think both that test code path and the [`string[][]` definition](https://github.com/vega/vega-lite/blob/main/src/selection.ts#L252) of `views` in a `TopLevelSelection` exist to reflect the behavior that this PR fixes (i.e., that sometimes view names changed but not always, and thus the only way to address some nested views uniquely was by providing a full path). 

I think simplifying things like this PR does is a great idea. We could also consider dropping the `string[][]` part of `views` as an additional simplification. 

However, I do want to flag that we currently have [an example changing](https://github.com/vega/vega-lite/pull/8733/files#diff-b8b58618fc1fae9b2510b7fbf7d832e4ee66c6a1aa71dfb8fb4747918929ee01) in an unexpected/unintended way. So there may be some other place in the codebase that is baking in this assumption as well? (And may suggest we might want to mark this as a breaking change in case users have baked this assumption in their use cases as well?)

joelostblom:
Would this also fix #8348 as was mentioned in the superseeded PR https://github.com/vega/vega-lite/pull/8663?

domoritz:
@arvind Thanks for taking a look.

> However, I do want to flag that we currently have [an example changing](https://github.com/vega/vega-lite/pull/8733/files#diff-b8b58618fc1fae9b2510b7fbf7d832e4ee66c6a1aa71dfb8fb4747918929ee01) in an unexpected/unintended way.

The example changing was a fluke, I think, because https://github.com/vega/vega-lite/pull/8601 didn't fully build before merging. The image is correct now in main: 
.

> I think simplifying things like this PR does is a great idea. We could also consider dropping the string[][] part of views as an additional simplification.

Let's do that. I updated the pull request and the examples. I think we should definitely only release this in a minor release but I would consider is somewhat of a bug fix for a broken behavior. I also don't think we should promise consistency in how the names for nested specs are generated but we could promise to not change them in a patch release. Does that sound good to you? Also cc @kanitw for input. 

> Would this also fix https://github.com/vega/vega-lite/issues/8348 as was mentioned in the superseeded PR https://github.com/vega/vega-lite/pull/8663?

I think so. At least the example from the first comment work as far as I can tell. 

Screenshot 2023-03-02 at 00 50 26

domoritz:
patch to stay backwards compatible: https://github.com/vega/vega-lite/pull/8752

arvind (approved):
> I think we should definitely only release this in a minor release but I would consider is somewhat of a bug fix for a broken behavior. I also don't think we should promise consistency in how the names for nested specs are generated but we could promise to not change them in a patch release. Does that sound good to you?

This sounds like a good balance to me! And I think releasing this in a minor release makes sense, especially in light of your idea of pushing backwards compatibility to the normalizer.

PR #8752, opened 2023-03-02
feat: stay backwards compatible

domoritz:
See comment in https://github.com/vega/vega-lite/pull/8733#issuecomment-1444171786


arvind (approved):
Oh, this is a clever way of making sure we fix the bug without breaking compatibility! Nice idea!

PR #8810, opened 2023-03-27
chore: upgrade deps and fix build

domoritz:

PR #8811, opened 2023-03-28
Issues with regenerated example images

jonmmease:
I'm not sure that we'd want to merge this PR as-is, but following up on the discussion in https://github.com/vega/vega-lite/pull/8810#issuecomment-1485836800, I wanted to regenerate the baseline images to see what looks like it's changed since they were last run.  I did this by deleting all of the specs and images from `examples/compiled` and running `scripts/build_examples.sh`

I'm seeing some concerning differences / regressions. I think my development environment is set up properly, but it would definitely be good for someone else to double check what I'm seeing. There are roughly three kinds of regressions showing up here:

#### Shifted axis labels
Screenshot 2023-03-28 at 10 03 26 AM

#### Shifted arc labels
Screenshot 2023-03-28 at 10 03 39 AM

#### Skinny arcs
Screenshot 2023-03-28 at 10 03 49 AM

#### Skinny bars
Screenshot 2023-03-28 at 10 04 06 AM

----
Since the vega specs didn't change, I guess these must be differences in Vega since the images were last generated. But I'm not seeing any issues in the Vega chart editor (which is on Vega 5.24.0).  I wonder if it's an image export issue 🤔


domoritz:
New pull request in https://github.com/vega/vega-lite/pull/8812 that runs updates on the CI instead.

jonmmease:
Sounds good, happy to move discussion over there

PR #8838, opened 2023-04-14
fix: correctly apply stacked to bar with quantitative x and quantitative y axes

kanitw:
Our default stack logic didn't read orient(ation). This PR fixes that.


fandu-db (approved):
LGTM! Thanks Ham!

Issue #8841, opened 2023-04-15
Website shows broken example


domoritz:
On https://vega.github.io/vega-lite/. 

<img width="1120" alt="Screenshot 2023-04-14 at 22 48 04" src="https://user-images.githubusercontent.com/589034/232179491-c49c2a69-cd74-4f7a-9ddb-689ce51bdedd.png">

domoritz:
Adding @arvind since he worked on the XSS fix.

PBI-David:
Broken examples half way down the page here too: https://vega.github.io/vega-lite/docs/condition.html

hydrosquall:
👋 I think the issue might be related to the fact that there are 2 copies of `vega-scale` used in vega-lite, as well as multiple copies of `vega-scenegraph` . For the scale registry that we added to work correctly, we make the assumption that `vega-functions` is interacting with only 1 version of `vega-scale` . If there are multiple copies, the upstream code might not succeed at looking up the correct copy.

- Vega-scale: https://a.cl.ly/rRugBWkr
- Vega-scenegraph: https://a.cl.ly/6qubXvYN

After all, the examples in regular vega: 
https://vega.github.io/vega/examples/bar-chart/

Or the same example from the website when opened in the [editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEzjQATjRyZ289AEEABBBoIcguIaZJ1h2DcyGA7nRiHETOMtXLDypJhUiioBKKigxEiCDGpoANqgYSD6wUxoAEwAHAC+ColoIABCqWhiYrn56ADCJagALADMFSBJACK1AJwAjM1JAKK1mT15LQUAYrViTSNJAOK1XR29BQASgwDsy+gAkpPp2QC6uSDI6gDWBdbqwXCyUGzKNLJkaKAAHq8gAGY0cILKBRQSkwAE8cHACrI2AgnlFgkg3jQIJ9BEhPIJ9M8LGgAAzZY4gz4-P4A9BpYFgiHoACODCQsh0gR0pBA+OyQA) are displaying correctly.

[...]

PR #8855, opened 2023-04-21
fix: repair docs site examples

hydrosquall:
## Motivation

- Quick fix for #8841

## Changes

- Ran `yarn why vega-scale` to find that both vega-scale `7.2.0` and `7.3.0` were present, when only was expecting
- Pining repo to `7.3.0` would prevent different libraries from accessing different versions.

## Testing

- Build site locally, confirm that the currently broken examples render correctly
- I'm not able to update the lockfile locally - something about python or C dependencies erroring out. Maybe this is something that can be regenerated and exported from CI, or by a maintainer with a working local environment.

## Notes

- Longer term fix -> try to adjust dependencies (possibly upstream) so that only one version of vega-scale is needed
- TODO: look into visual regression testing to detect this type of issue programmatically in the future

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "ts-jest": "^29.0.5",
     "ts-json-schema-generator": "^1.2.0",
     "typescript": "~4.9.5",
-    "vega-cli": "^5.22.1",
+    "vega-cli": "^5.24.0",
     "vega-datasets": "~2.7.0",
     "vega-embed": "^6.21.3",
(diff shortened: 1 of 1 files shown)


domoritz:
Good find. Looks like the issue is that Vega depends on a specific version. I think we need to update Vega.

domoritz:
Superseded by https://github.com/vega/vega-lite/pull/8857

PR #8871, opened 2023-04-27
fix: remove incorrect stack transform from charts with one linear Q-axis one non-linear Q-axis.

kanitw:
Remove incorrect stack transform from charts with one linear Q-axis one non-linear Q-axis.


src/stack.ts
```diff
@@ -104,15 +104,6 @@ function potentialStackedChannel(
       // if there is no explicit stacking, only apply stack if there is only one aggregate for x or y
       if (xAggregate !== yAggregate) {
         return xAggregate ? x : y;
-      } else {
```
kanitw, line 107: This logic was just wrong. Instead, we should rely on the orient logic below.

ZacharyBys (approved):

examples/compiled/bar_q_qpow.png
ZacharyBys, line 1: Are these two images supposed to be before and after? I don't see any changes

domoritz (approved):
Looks good. There is a failing test for bar_q_qpow.vl.json, though.

examples/compiled/bar_q_qpow.png
kanitw: If you see the 4th (generated commits), you can see that the actual code changes in the 3rd commit fixes the problem.

kanitw: image

PR #8885, opened 2023-05-06
Add support for tiles to geoshape mark

### LINKED ISSUE #8767: include tiles option for geoshape mark
With an imperative proof of concept in https://github.com/vega/vega-lite/issues/5758#issuecomment-1462683219 for tile support within Vega-Lite. I'm proposing something as the following for declarative tiles support:

```json
"mark": {"type": "geoshape", "tiles": true} 
```
Where `true` defaults to `"tiles": {"url": "https://tile.opentopomap.org/", "attribution": "(C) OpenStreetMap contributors"}` 

similar to `"mark": {"type": "line", "points": true}`.

This can also be extended with providers that require access-tokens or API keys. 

The imperative approach of the poc above can then be included within the extended vega-lite specification.

binste:
This PR is far from being ready to be merged as many corner cases need to be handled, tests and documentation written, etc. But I have a few questions on how to proceed which are probably easier to answer if you can see the code. The goal is to add a `tile` property to the `geoshape` mark which adds tiles from arbitrary tile XYZ tile servers as `image` marks. See #8767 for more context and all the credits for the calculations goes to @mattijn who provided the initial specs!

## Current status
I already got the layering to work so that the following spec (notice the `"tile": true` in `"mark"`):

```json
{
  "data": {
    "url": "https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/world-110m.json",
    "format": {"feature": "countries", "type": "topojson"}
  },
  "mark": {
    "type": "geoshape",
    "fillOpacity": 0.1,
    "stroke": "orange",
    "strokeWidth": 2,
    "tile": true
  },
  "width": 400,
  "height": 400,
  "projection": {
    "type": "mercator",
    "scale": 300,
    "center": [0, 40],
    "rotate": [-10, 0, 0]
  }
}
```

produces an extended VL spec in the Vega Editor which is very close to what @mattijn originally came up with. The main change is that I rewrote the calculations so that they do not depend on an input zoom level but instead the zoom level is determined based on `scale` in `projection`:

```json
{
  "width": 400,
  "height": 400,
  "layer": [
    {
      "mark": {
        "type": "image",
        "clip": true,
        "height": {"expr": "tile_size"},
        "width": {"expr": "tile_size"}
      },
      "encoding": {
        "url": {"field": "url", "type": "nominal"},
        "x": {"field": "x", "scale": null, "type": "quantitative"},
        "y": {"field": "y", "scale": null, "type": "quantitative"}
      },
      "data": {
        "sequence": {"start": 0, "stop": 4, "as": "a"},
        "name": "tile_list"
      },
      "transform": [
        {"calculate": "sequence(0, 4)", "as": "b"},
        {"flatten": ["b"]},
        {
          "calculate": "base_tile_url + zoom_ceil + '/' + (datum.a + dii_floor + tiles_count) % tiles_count + '/' + ((datum.b + djj_floor)) + '.png'",
          "as": "url"
        },
        {"calculate": "datum.a * tile_size + dx + tile_size / 2", "as": "x"},
        {"calculate": "datum.b * tile_size + dy + tile_size / 2", "as": "y"}
      ]
    },
    {
      "projection": {
        "type": "mercator",
        "scale": 300,
        "center": [0, 40],
        "rotate": [-10, 0, 0]
      },
      "data": {
        "url": "https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/world-110m.json",
        "format": {"feature": "countries", "type": "topojson"}
      },
      "mark": {
        "type": "geoshape",
        "fillOpacity": 0.1,
        "stroke": "orange",
        "strokeWidth": 2
      },
      "encoding": {}
    }
  ],
  "params": [
    {"name": "base_tile_size", "value": 256},
    {"name": "base_tile_url", "value": "https://tile.openstreetmap.org/"},
    {"name": "prScale", "expr": "300"},
    {
      "name": "zoom_level",
      "expr": "log((2 * PI * prScale) / base_tile_size) / log(2)"
    },
    {"name": "zoom_ceil", "expr": "ceil(zoom_level)"},
    {"name": "tiles_count", "expr": "pow(2, zoom_ceil)"},
    {
      "name": "tile_size",
      "expr": "base_tile_size * pow(2, zoom_level - zoom_ceil)"
    },
    {"name": "base_point", "expr": "invert('projection', [0, 0])"},
    {"name": "dii", "expr": "(base_point[0] + 180) / 360 * tiles_count"},
    {"name": "dii_floor", "expr": "floor(dii)"},
    {"name": "dx", "expr": "(dii_floor - dii) * tile_size"},
    {
      "name": "djj",
      "expr": "(1 - log(tan(base_point[1] * PI / 180) + 1 / cos(base_point[1] * PI / 180)) / PI) / 2 * tiles_count"
    },
    {"name": "djj_floor", "expr": "floor(djj)"},
    {"name": "dy", "expr": "round((djj_floor - djj) * tile_size)"}
  ]
}
```

This VL spec renders fine:

image

## Questions
There are two related issues around the parameters in the above spec where I'd be very grateful for guidance from @domoritz or any VL maintainer.

* The current approach relies on value parameters which in my understanding need to be at the top-level of a specification, i.e. I cannot have them in a `LayerSpec`. I managed to push them to the top-level of the processed view as in the example above, but this does not work if the above spec is wrapped into an `hconcat` chart. Any ideas (existing mechanisms?) how I could push these parameters to the top of any spec, independent of how deeply nested the chart spec is which has `"tile": true`?
* The parameters do show up in the extended Vega-Lite specs in the Vega Editor. However, this is not the spec which is passed to `vega.parse`. The spec which is eventually passed lost the parameters section. I think this happens here https://github.com/vega/vega-lite/blob/main/src/compile/compile.ts#L122. `spec` has the parameters, `vgSpec` does not. You can see the behavior I described in this screen recording:

https://user-images.githubusercontent.com/23366411/236632406-5842518a-4c11-4884-b01d-34a52a8539d5.mov

If it's not possible to do what I want with the parameters, I could maybe rewrite the generated spec so that all expressions are inlined but that would lead to a lot of duplicated calculations and makes the resulting VL and Vega specs less readable in my opinion.

Any help is greatly appreciated! Thanks.

Linked issue: Closes #8767.

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -144,6 +144,9 @@
         },
         {
           "$ref": "#/definitions/TickConfig"
+        },
+        {
+          "$ref": "#/definitions/GeoshapeConfig"
         }
       ]
     },
@@ -7625,7 +7628,7 @@
           "type": "string"
         },
         "geoshape": {
-          "$ref": "#/definitions/MarkConfig",
+          "$ref": "#/definitions/GeoshapeConfig",
           "description": "Geoshape-Specific Config"
         },
         "header": {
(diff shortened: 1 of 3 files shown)


mattijn:
Nice work already @binste! Regarding the required lifting of parameters to top-level, I would prefer to move the mechanism we introduced in Altair, https://github.com/altair-viz/altair/pull/2702 to Vega-Lite so this lifting can be done when serialization it (as macro?) into extended Vega-Lite. 
In hindsight it would have been better to do that in Vega-Lite directly, but at the moment of writing we (at least I) were not aware of the extended Vega-Lite principle (Vega-Medium?).

This also would mean that Vega-Lite can better controls the naming definition of parameter views.

binste:
So far I only saw the gap in the Vega Editor, not in VS Code with the same settings. Maybe the placement of the images is sometimes slightly different. If I add 1 pixel to the height and width of the image marks then the gap is also closed in the Vega Editor.

PR #8886, opened 2023-05-06
Add 'yarn schema' command to CONTRIBUTING.md

binste:


domoritz:
Thanks for the pull request. I think we should rebuild the schema on build instead. What do you think?

binste:
Sounds good, I'd find that more convenient. To only compile VL without the schema one can still use `build:only`.  I adjusted `package.json` and reverted the change in `CONTRIBUTING.md`.

domoritz:
Yeah, we need to comprehensively look through the scripts. I don't have time right now but will look at some point to resolve the issue you found.

domoritz:
I think https://github.com/vega/vega-lite/pull/9877 addressed this.

PR #8894, opened 2023-05-09
feat: Make bars without secondary channel thicker

yhoonkim:
Added the logic making the bar's size thicker if there is no secondary channel.


src/compile/mark/encode/position-rect.ts
```diff
@@ -110,6 +111,9 @@ function defaultSizeRef(
       return {value: scaleRange.step - 2};
     }
   }
+  if (!hasFieldDef) {
+    return {signal: `0.8 * ${sizeChannel}`};
```
yhoonkim, line 115: Wonder if we extract `{signal: `0.8 * ${sizeChannel}`}` as the constant. But I could not find a good place to place if we extract...

kanitw: Very good question. For consistency with existing similar style, we can respect `config.scale.bandPaddingInner/ barBandPaddingInner/rectBandPaddingInner`.

Basically, we can think of axis without a field as categorical axis with one unique value. 

So we can do: 

```
const padding = getFirstDefined(bandPaddingInner, mark === 'bar' ? barBandPaddingInner : rectBandPaddingInner);
return {signal: `${1 - padding} * ${sizeChannel}`}; 
```

examples/compiled/facet_bullet.svg
kanitw, line 1: For this example, we may want to set config.scale.barBandPaddingInner = 0

src/compile/mark/encode/position-rect.ts
```diff
@@ -110,6 +115,19 @@ function defaultSizeRef(
       return {value: scaleRange.step - 2};
     }
   }
+  if (!hasFieldDef) {
+    const {bandPaddingInner, barBandPaddingInner, rectBandPaddingInner} = config.scale;
+    const padding = getFirstDefined(bandPaddingInner, mark === 'bar' ? barBandPaddingInner : rectBandPaddingInner); // this part is like paddingInner in scale.ts
+    if (isSignalRef(padding)) {
+      return {signal: `(1 - (${padding.signal})) * ${sizeChannel}`};
+    } else if (isNumber(padding)) {
+      return {signal: `(1 - (${padding})) * ${sizeChannel}`};
```
kanitw, line 124:
```diff
- signal: `(1 - (${padding})) * ${sizeChannel}`
+ signal: `${1-padding} * ${sizeChannel}`
```

kanitw: I think we don't need these parentheses for this number case.

kanitw (approved):
Last few nits before merging. :)

Issue #8143, opened 2022-05-06
Boolean field selection issue when selection is bound to legend


hannahburkhardt:
When binding a selection for a boolean field to the legend, I can select "true" as expected, but for "false", I have to click between the symbol and the word. Clicking on the symbol (circle in the example below) or the string "false" produces the following error in the log: `[Error] Cannot read properties of undefined (reading '0')`

[Example](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAFgAY2AacaYsiygAlMiRoRQBmDgF9p3OgENCCvCFIAnADYoQYwgAdKyAPTGwdKADoAVpTqZNcNOstRMhY1H0BbY2kwICgC0isqU7pQAAmgAjJYATACclmzGoQrGAEYKWpgAnjaU0CByIN45ANY6+hAk4tyYUJB0JEioPJoQ6qrwDnQ6aDlYhHkg3CP6mDpQEN4kCtqlEPoKYHAjqgQtxMXtg5qkU8gx3OGamGA7UDpnF4RdbPHxJdz7hyhsCaWUjAqTPVhNP1kCBMmMQBMjiAZnMoAsXiAAB4Avo6PKYHLgyE6LotOGLbijdq9IFowH9cZ5f4ggCOpAURHWSicU1kpwcdzgu1At0uDyeqmxIO8pE0xHBJLolBQAG1QSAALrcTIkYEgc4IRr9NkQ9QMygwLreWX4BZgUVKKGhUjeSyDdTDPIAXidAHIAOKaBSZdTQV3ghTSkFg6RKkAAEkoYEYmHKOj0hhMfgCCksCHWjFImUsXOMUZj5WTgSCjkImD8AFYEilCsVpEA):

```json
{
  "config": {"view": {"continuousWidth": 400, "continuousHeight": 300}},
  "data": {
    "url": "https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/barley.json"
  },
  "mark": "point",
  "encoding": {
    "color": {"field": "variety", "type": "nominal"},
    "opacity": {
      "condition": {"value": 1, "selection": "selector022"},
      "value": 0.2
    },
    "shape": {"field": "b", "type": "nominal"},
    "x": {"field": "year", "type": "ordinal"},
    "y": {"field": "yield", "type": "quantitative"}
  },
  "selection": {
[...]

PR #8800, opened 2023-03-23
fix: handle falsy legend values in selections

jonmmease:
Proposed fix for https://github.com/vega/vega-lite/issues/8143.

The problem this PR addresses is that when a boolean column is present in the legend, and a selection is bound to the legend, clicking on the `false` entry does nothing.

The problem is this expression:
```
datum.value || item().items[0].items[0].datum.value
```
When `datum.value` is falsy, it's value is never used. Instead, I think this expression should be

```
datum.value != null? datum.value: item().items[0].items[0].datum.value
```
so that the datum's value is used as long as it's not null or undefined.

---
Here is the result of this change for the repro in #8143

[...]

PR #8895, opened 2023-05-10
fix: handle falsy legend values in selections (take 2)

jonmmease:
Replacement for https://github.com/vega/vega-lite/pull/8800, which was reverted.  Fixes https://github.com/vega/vega-lite/issues/8143.

This implementation also handles `null` legend values by using the `isDefined` expression function.

### DIFF
--- a/src/compile/selection/legends.ts
+++ b/src/compile/selection/legends.ts
@@ -72,7 +72,7 @@ const legendBindings: SelectionCompiler<'point'> = {
           ...(!selCmpt.init ? {value: null} : {}),
           on: [
             // Legend entries do not store values, so we need to walk the scenegraph to the symbol datum.
-            {events, update: 'datum.value || item().items[0].items[0].datum.value', force: true},
+            {events, update: 'isDefined(datum.value)? datum.value: item().items[0].items[0].datum.value', force: true},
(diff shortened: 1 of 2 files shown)


src/compile/selection/legends.ts
```diff
@@ -72,7 +72,7 @@ const legendBindings: SelectionCompiler<'point'> = {
           ...(!selCmpt.init ? {value: null} : {}),
           on: [
             // Legend entries do not store values, so we need to walk the scenegraph to the symbol datum.
-            {events, update: 'datum.value || item().items[0].items[0].datum.value', force: true},
+            {events, update: 'isDefined(datum.value)? datum.value: item().items[0].items[0].datum.value', force: true},
```
domoritz, line 75: Just a small formatting difference to how we usually write ternaries. 

```suggestion
            {events, update: 'isDefined(datum.value) ? datum.value : item().items[0].items[0].datum.value', force: true},
```

domoritz (approved):

src/compile/selection/legends.ts
jonmmease: Done in [ee2703f](https://github.com/vega/vega-lite/pull/8895/commits/ee2703f6386776432a0a3e4a86f2aa69f6874b84)

domoritz:
Thank you

PR #8899, opened 2023-05-12
docs: fix position-offset permalink in encoding.md

dkcheng:
Current docs have a typo in the [Position Offset permalink](https://vega.github.io/vega-lite/docs/encoding.html#positon-offset). Fixing this to "position-offset" so the [reference here](https://github.com/vega/vega-lite/blob/309af9ac314b5c83ecf89afbda6069a63fd7a53e/site/docs/encoding.md?plain=1#L84) resolves properly.


domoritz:
Thank you

PR #8906, opened 2023-05-16
fix: respect bandWithNestedOffsetPaddingInner for grouped bar within discretized temporal axis

kanitw:
Having some gap should make it clearer this is a grouped bar. 

Before:

image

After:

image


fandu-db (approved):
So much better with the gap!

domoritz (approved):

Issue #4060, opened 2018-07-19
Time channel


domoritz:
Vega supports timer events. It would be nice to have a channel for that. 

For example https://vega.github.io/vega/examples/hypothetical-outcome-plots/. 

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {
    "url": "data.json"
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "a", "type": "ordinal"},
    "y": {"field": "b", "type": "quantitative"},
	"time": {"field": "time", "type": "ordinal", "interval": 1000}
  }
}
```

* We only support ordinal (iterate through) and quantitative (do we need a step size?).
* How should users define an interval?

g3o2:
Is timer more like an encoding or more like a selection/interaction?

eeyshen:
Is there any progress on this? Thanks.

domoritz:
http://www.ycwu.org/Files/Data-GIF.pdf gives a good overview over the design space.

joelostblom:
I just stumbled upon [Gemini](https://github.com/uwdata/gemini), which looks really cool! Is the plan to support animations directly from the VL spec in the future via Gemini?

domoritz:
We do plan to add animation support without recommendation but probably not via Gemini but native Vega support.

domoritz:
This idea is implemented in Animated Vega-Lite: https://arxiv.org/abs/2208.03869.

PBI-David:
Moving examples can be found here. This looks amazing. Will this also be available in Vega?

http://vis.csail.mit.edu/pubs/animated-vega-lite/

domoritz:
We're working on a plan to create prs to add the features from the paper to Vega-Lite.

gut4:
any updates on this?

PBI-David:
For anyone else looking for timer support in VL, it can be achieved as shown in the nice looking clock example on Stack Overflow here: https://stackoverflow.com/questions/78106640/vega-lite-clock-chart/78117784

domoritz:
@jonathanzong is working on https://github.com/vega/vega-lite/pull/8921. All updates are over there.

apb-reports:
Check out this example also where I switch between 4 different layers with the use of a timer event. A few warning in the log but it works well. https://stackoverflow.com/questions/78126558/animated-vega-lite-charts

joelostblom:
[...]

PR #8921, opened 2023-05-25
feat: frame animations with time encoding and timer param

jonathanzong:
This change implements basic features of [Animated Vega-Lite](https://vis.csail.mit.edu/pubs/animated-vega-lite/). With this change, users can create frame animations using a `time` encoding, a `timer` point selection, and a `filter` transform.

This change does _not_ include more complex features e.g.: interpolation, custom predicates, rescale, interactive sliders, or data-driven pausing.

- Adds a `time` encoding channel
- Adds `isTimerSelection` function to check if a selection is an animation selections
- Builds `_curr` animation dataset for `timer` selections to store the current animation frame
- Adds animation signals to track the elapsed time (`anim_clock`), current animation value (`anim_value`), current position in the animation field's domain (`t_index`), etc.
- When `time` encoding is present, updates associated marks' `from.data` to use the animation dataset (current frame).

Relevant issue: https://github.com/vega/vega-lite/issues/4060

Coauthor: @joshpoll

## Example specs

Hop example:

```json
{
  "data": {
    "url": "data/seattle-weather.csv"
  },
  "mark": "tick",
  "config": {
    "tick": {
      "thickness": 3
    }
  },
  "params": [
    {
      "name": "date",
      "select": {
        "type": "point",
        "fields": [
          "date"
        ],
        "on": "timer"
      }
    }
  ],
  "transform": [
    {
      "filter": {
        "param": "date"
      }
    }
  ],
  "encoding": {
    "y": {
      "field": "precipitation",
      "type": "quantitative"
    },
    "time": {
      "field": "date"
    }
  }
}
```

Gapminder:
```json
{
  "data": {
    "url": "data/gapminder.json"
  },
  "mark": "point",
  "params": [
    {
      "name": "avl",
      "select": {
        "type": "point",
        "fields": [
          "year"
        ],
        "on": "timer"
      }
    }
  ],
  "transform": [
    {
      "filter": {
        "param": "avl"
      }
    }
  ],
  "encoding": {
    "color": {
      "field": "country"
    },
    "x": {
      "field": "fertility",
      "type": "quantitative"
    },
    "y": {
      "field": "life_expect",
      "type": "quantitative"
    },
    "time": {
      "field": "year"
    }
  }
}
```


jonathanzong:
@domoritz @arvind this is ready for review. we were thinking hold off on docs and examples until a later PR, but let us know if you'd rather us include it here

domoritz:
Thanks. Can we see whether you can run the formatting action as well so we know the formatting doesn't break.

jonathanzong:
i was able to run `yarn format` with no problems, we've been using the eslint config

domoritz:
Great. I was hoping we could get the GitHub action for checking formatting to work somehow.

joshpoll:
I think it's running in the Runtime, Linting, and Coverage test.

Screen Shot 2023-05-30 at 10 48 44 AM

src/compile/selection/point.ts
```diff
@@ -1,11 +1,49 @@
...
+      on: [{events: [{signal: 'anim_clock'}], update: 'now()'}]
+    },
+    {
+      name: 'eased_anim_clock',
+      // update: 'easeLinear(anim_clock / max_range_extent) * max_range_extent'
+      update: 'anim_clock' // TODO: replace with above once easing functions are implemented in vega-functions
+    },
+
+    // scale signals
+    // TODO(jzong): uncomment commented signals when implementing interpolation
```
arvind, line 37: There still appear to be some commented signals, so I'm not sure what the status of this TODO is?

src/compile/selection/point.ts
```diff
@@ -1,11 +1,49 @@
...
+import {SelectionCompiler, TUPLE, isTimerSelection, unitName} from '.';
 import {SELECTION_ID} from '../../selection';
 import {vals} from '../../util';
 import {BRUSH} from './interval';
 import {TUPLE_FIELDS} from './project';
 
+export const CURR = '_curr';
+
+const animationSignals = (selectionName: string, scaleName: string): Signal[] => {
+  return [
```
arvind, line 12: Are these selection-specific signals, or global signals (i.e., one set of signals that are used by all selections)? I _think_ it's the former based on some of the references within. In which case, I think they'll all need to be prefixed by the selection name. Otherwise, if you have multiple timer selections within the same unit spec, you're going to get a duplicate signal error.

src/compile/selection/point.ts
```diff
@@ -1,11 +1,49 @@
...
+    // timer signals
+    {
+      name: 'anim_clock',
```
arvind, line 15: Can we avoid hardcoding signal names and instead make use of exported CONSTs? That'll make any future refactoring just a little bit easier.

src/compile/selection/point.ts
```diff
@@ -1,11 +1,49 @@
...
+      init: '0',
+      on: [
+        {
+          events: {type: 'timer', throttle: 16.666666666666668},
```
arvind, line 19: Where does this throttle come from? (I'm guessing it's something like the browser's requestNextAnimFrame?) I wonder if there's a way to not hardcode this value...

src/compile/selection/point.ts
```diff
@@ -1,11 +1,49 @@
...
-import {SelectionCompiler, TUPLE, unitName} from '.';
+import {SelectionCompiler, TUPLE, isTimerSelection, unitName} from '.';
 import {SELECTION_ID} from '../../selection';
 import {vals} from '../../util';
 import {BRUSH} from './interval';
 import {TUPLE_FIELDS} from './project';
 
+export const CURR = '_curr';
+
+const animationSignals = (selectionName: string, scaleName: string): Signal[] => {
```
arvind, line 11: How different will these signals be for when we generalize the implementation (eg to have timer-driven interval signals)? Is it worth extracting them to their own selection module and working them in that way for reusability? Happy to punt on that for now, but it'd be useful to spec things through and externalize them while they're more fresh in your head.

src/compile/selection/point.ts
```diff
@@ -52,21 +92,38 @@ const point: SelectionCompiler<'point'> = {
...
-                update: `${test} ? {${update}} : null`,
-                force: true
-              }
-            ]
-          : []
-      }
-    ]);
+    if (isTimerSelection(selCmpt)) {
+      // timer event: selection is for animation
+      // TODO(jzong) scale name currently hardcoded to 'time'
```
arvind, line 97: Status of this TODO?

src/compile/selection/toggle.ts
```diff
@@ -9,11 +9,18 @@ const toggle: SelectionCompiler<'point'> = {
   },
 
   signals: (model, selCmpt, signals) => {
-    return signals.concat({
-      name: selCmpt.name + TOGGLE,
-      value: false,
-      on: [{events: selCmpt.events, update: selCmpt.toggle}]
-    });
+    if (isTimerSelection(selCmpt)) {
```
arvind, line 12: If toggle is always set to `false` in the case of timer selections, it might be preferable to augment `defined` above? i.e., saying that `toggle` can never be instantiated on a timer selection—so the resultant signals/etc. are all bypassed, and has the knock on effects throughout the selection codebase.

src/compile/unit.ts
```diff
@@ -271,7 +272,15 @@ export class UnitModel extends ModelWithField {
       marks = assembleUnitSelectionMarks(this, marks);
     }
 
-    return marks.map(this.correctDataNames);
+    const marksWithCorrectedDataNames = marks.map(this.correctDataNames);
+    if ('time' in this.encoding) {
+      marksWithCorrectedDataNames.forEach(mark => {
+        if (mark.from?.data) {
+          mark.from.data = mark.from.data + CURR;
```
arvind, line 279: Why does this happen here? Could/should it happen in `this.correctDataNames` (which, besides keeping mark-data naming all encapsulating in one function, might also benefit other parts of the codebase?).

src/compile/selection/point.ts
joshpoll: these signals are for a follow-up PR

src/compile/selection/toggle.ts
joshpoll: That makes sense. Is there documentation for the semantics of `defined`? If not I can try to add it in this or a follow-up PR.

joshpoll: Actually maybe I'll defer to jz on this since it's `false` in both branches. I'm not 100% sure what the behavior of this is supposed to be.

src/compile/unit.ts
joshpoll: I moved `correctDataNames` into unit model since it requires inspecting `this.encoding` and it's only used for this method.

src/compile/selection/point.ts
jonathanzong: this is 60 frames / second converted into milliseconds / frame. we could make that more legible by writing it as `1 / 60 * 1000` and adding a comment

src/compile/scale/range.ts
```diff
@@ -292,6 +293,13 @@ function defaultRange(channel: ScaleChannel, model: UnitModel): VgRange {
       ];
     }
 
+    case TIME: {
+      if (scaleType === 'band') {
+        return {step: 500};
+      }
+      return [0, 5000]; // e.g. linear
```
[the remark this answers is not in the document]
jonathanzong: address by josh in https://github.com/vega/vega-lite/pull/8921/commits/8bac921a26899841e057af7a0f72c3e57531c155

src/compile/selection/point.ts
jonathanzong: addressed by josh in https://github.com/vega/vega-lite/pull/8921/commits/2d55ccf7c77f398bdf51d8c8b2e9c31d40bfef66

src/compile/selection/point.ts
jonathanzong: resolved in https://github.com/vega/vega-lite/pull/8921/commits/41ae4b530c4ebf22c9d3e10d4557d8d248def96d

src/compile/unit.ts
jonathanzong: resolved in https://github.com/vega/vega-lite/pull/8921/commits/3b63808eaf1d31e09d6cd203164e39ac2a28e711

src/compile/selection/parse.ts
```diff
@@ -91,7 +91,7 @@ export function parseSelectionPredicate(
     }
   }
 
-  const fn = selCmpt.project.hasSelectionId ? 'vlSelectionIdTest(' : 'vlSelectionTest(';
+  const fn = selCmpt.project.hasSelectionId && !isTimerSelection(selCmpt) ? 'vlSelectionIdTest(' : 'vlSelectionTest(';
```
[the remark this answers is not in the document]
jonathanzong: I think there are reasons, but I couldn't really remember why it was necessary to add this check here and it doesn't affect the current examples, so I reverted this line in https://github.com/vega/vega-lite/pull/8921/commits/25349cd1fde4b764583d3d3380eb360c06a01b56

src/compile/selection/toggle.ts
jonathanzong: using defined feels like the right thing, thanks: https://github.com/vega/vega-lite/pull/8921/commits/cac66c9bab5d35e103da1712b2e8c8180a4b4436

src/compile/selection/point.ts
jonathanzong: i think deciding what's selection-specific vs global matters for two situations:

1. multiple timer selections in the same unit spec

i don't think this should be allowed, because what does it mean to animate the a single set of marks two different ways? what's the best way to disallow this?

2. multiple timer selections across multiple unit specs in a multi-view

we ideally want the animation clock `ANIM_CLOCK` to be global, so we can share a common timer across multiple units. but the problem is that the animation clock needs to know the `MAX_RANGE_EXTENT` to know when to wrap back around to 0, but this is calculated from the scale for each unit. imo for now this is fine since we didn't consider multi-view in the scope of the paper, but we'll have to figure this out when we think about multi-view / scale resolution. the extent of the clock needs to be determined from the unioned scale or the max of the independent scale extents

src/compile/selection/point.ts
jonathanzong: i feel like i will have a lot of questions about what changes to make to selections to expose predicates so i think i will opt to punt this from this PR's scope but try to write some thoughts down in a doc

jonathanzong:
first round of comments has been addressed. pending todos:

- question here https://github.com/vega/vega-lite/pull/8921#discussion_r1336304428
- find some time to talk about how runtime testing works

mattijn:
https://github.com/vega/vega-lite/pull/8921/commits/c72cb7db0c6272360a8c93d3a92bd1351a915a81 was an attempt to trigger the deployment preview of https://github.com/vega/vega-lite/pull/9294..

mattijn:
Hi @jonathanzong and @joshpoll!👋 Would you mind updating this PR so the latest changes from the main repository are also included in this branch?

Im not brave enough to do it myself🙈, but I'm sensing that this is the reason why the new deployment preview is not yet triggering.

mattijn:
Just thinking about making a map with the trajectory of the upcoming eclipse🥳

joelostblom:
> Just thinking about making a map with the trajectory of the upcoming eclipse🥳

Such a cool idea! Please share if you create it, I would love to see an animated VL chart for this.

Btw, @jonathanzong are you waiting for a review on this branch or are you planning to add more commits (I saw you were still adding more since you last requested a review).

joshpoll:
I rebased

jonathanzong:
We are indeed just waiting for a review

arvind:
Yes, apologies, it's been on my docket for a while but I've been underwater with various other deadlines. My plan is to wrap this up this month 🤞

mattijn:
Thanks for rebasing @joshpoll! But that didn't trigger the deployment review either. 

@hydrosquall, do you know if these deployment previews only work upon opening a new PR and not on existing PRs?

hydrosquall:
Hm, it looks like I might have to add eligible branch prefixes one at a time, until I figure out how the "include" and "exclude" rules compose (which takes precedence). If you push another commit to this branch that modifies anything in the `src/` directory, a preview should build.

joelostblom:
@arvind Just a ping on this one when you get a chance. I re-requested the review so it shows up correctly in your queue.

@jonathanzong Could you make any minor modification to one of the files in `src` for the preview to build as per the comment from @hydrosquall ? It can be something non-significant like modifying a comment.

jonathanzong:
i have rebased, fixed checks, and verified that the animation works as expected in local editor (for what i hope is the last time for this pr?)

src/compile/selection/point.ts
arvind: Re: (2) a global `ANIM_CLOCK`, I agree that we didn't consider multi-view in the paper, but I feel a little strange to introduce this constrain in production since, from an end user perspective, it might seem like an arbitrary limitation? Could you humor me and help me think through how complex this would be to implement? If the issue is primarily about how to resolve `MAX_RANGE_EXTENT` for wrapping around back to 0, are there more than these two cases we would need to handle:

1. No global `ANIM_CLOCK`: so each unit can have only one timer selection, and they are independent and loop independently.
2. A shared `ANIM_CLOCK`: they would wrap around at the same time; that means, you essentially would need to max over the constituent `MAX_RANGE_EXTENT` and then use that to determine the wrap around. The effect would be that one timer would "pause" while the other(s) finish, and then they wrap.

These two cases feel like they map well to VL's existing notion of `resolve` (which is used for scales/axes) and which we overload for selection predicates in terms of `union`/`intersect`. So I feel like there's conceptual machinery already in place that we might be able to piggy back off?

examples/specs/animated_gapminder.vl.json
```diff
@@ -0,0 +1,43 @@
+{
+  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
+  "data": {
+    "url": "data/gapminder.json"
+  },
+  "mark": "point",
+  "params": [
+    {
+      "name": "avl",
```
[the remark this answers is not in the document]
jonathanzong: https://github.com/vega/vega-lite/pull/8921/commits/0c7df9a9ec0ab77822bc6e67f76e2d4c97a8e98d

src/compile/selection/point.ts
jonathanzong: logging and unit tests added

https://github.com/vega/vega-lite/pull/8921/files/25d1b1c02b4aac8c2cbc72758eb36dda7cedb2ef..c2b4b59bad9f2e6b6b3fc180c5009e56dd61fd92

jonathanzong: option 2 sounds like what we want eventually. i looked back at what we did for the paper and we did implement `layer` and vconcat` in the prototypes. but in our prototypes these were two separate implementations with not a lot of overlap. i also spent most of the day trying to get facet to work, and it almost works modulo the range resolution stuff, but i have no idea how to do it. we knew when we were doing the paper that conceptually there's overlap, but i suspect there isn't overlap in the implementation (at least right now) since we're using signals and note stores, and can't make use of `vlSelectionResolve` directly

all in all, i think it'd be a pretty big lift to get multi-view working in this PR. when we set out to do this PR, we thought of it as the minimal set of features to get people making basic animations that would cover the most common use case (we also aren't scoping in things like interpolation which i would prioritize over multi-view). i'm hesitant to increase the scope at this point because of 1) how hard it is for me to build velocity in this codebase due to complexity 2) the accessibility libraries being more of a personal priority

jonathanzong:
i have added an error explaining that facet, layer, and concat animations are unsupported when a user tries to create them (and unit tests for this error).

how serious are we about codecov? if you look at what's left, it's codecov asking for:
- unit test for consts that i added
- unit tests for one-line util functions for the runtime tests
- things that i'm pretty sure already have coverage

can we just bypass this? it's a little ridiculous

arvind (approved):
🎉🎉🎉 

I think we're good to go here. And I agree, fine to bypass the codecov—it seems to be failing on exceedingly minor things.

PR #8940, opened 2023-06-09
feat: add `extent` transform

lsh:
Adds an interface to use Vega's [`extent` transform](https://vega.github.io/vega/docs/transforms/extent/).


site/docs/transform/extent.md
```diff
@@ -0,0 +1,59 @@
...
+  ...
+  "transform": [
+    {"extent": ...} // Fold Transform
+     ...
+  ],
+  ...
+}
+```
+
+## Fold Transform Definition
```
domoritz, line 22:
```suggestion
## Extent Transform Definition
```

src/compile/data/extent.ts
```diff
@@ -0,0 +1,42 @@
+import {ExtentTransform as VgExtentTransform} from 'vega';
+import {ExtentTransform} from '../../transform';
+import {duplicate, hash} from '../../util';
+import {DataFlowNode} from './dataflow';
+
+/**
+ * A class for flatten transform nodes
```
domoritz, line 7:
```suggestion
 * A class for extent transform nodes
```

src/transform.ts
```diff
@@ -390,6 +390,19 @@ export interface FoldTransform {
...
 
+export interface ExtentTransform {
+  /**
+   * The field of which to get the extent.
+   */
+  extent: FieldName;
+
+  /**
+   * The output parameter produced by the extent transform.
+   * __Default value:__ `"extent"`
```
domoritz, line 401: Can there really be a default? Would that mean you pass undefined or null? Maybe we should not make this optional.

domoritz (approved):
Looks good. I'd like to have a quick check from @kanitw before merging.

src/transform.ts
lsh: I agree. Updated to fix this.

site/docs/transform/extent.md
```diff
@@ -0,0 +1,59 @@
...
+
+And the transform:
+
+```json
+"transform": [
+  {"extent": "b", "param": "b_extent"}
+]
+```
+
+this example produces the output `param` will look like:
```
kanitw, line 48: Probably better say the output `b_extent` param will be `[19, 91]`? 

(The following JSON in the block isn't really produced anywhere?)

src/compile/data/extent.ts
```diff
@@ -0,0 +1,40 @@
...
+    super(parent);
+    this.transform = duplicate(transform);
+  }
+
+  public dependentFields() {
+    return new Set([this.transform.extent]);
+  }
+
+  public producedFields() {
+    return new Set([this.transform.param]);
```
kanitw, line 24: This isn't a field produced.

PR #8962, opened 2023-06-25
fix: show gridline of first axis only for dual axes charts

kanitw:
Rationale: Better transitions -- at least adding the second axis won't suddenly hide all gridlines


src/compile/axis/parse.ts
```diff
@@ -93,9 +93,14 @@ export function parseLayerAxes(model: LayerModel) {
       delete child.component.axes[channel];
     }
 
-    // Suppress grid lines for dual axis charts (https://github.com/vega/vega-lite/issues/4676)
+    // Show gridlines for first axis only for dual-axis chart
     if (resolve.axis[channel] === 'independent' && axes[channel] && axes[channel].length > 1) {
+      let first = true;
       for (const axisCmpt of axes[channel]) {
```
domoritz, line 99: With `.entries()`, you get the index and can use that instead of the somewhat ugly `first` variable.

domoritz (approved):

domoritz:
Good change overall (as discussed on our hike).

PR #8988, opened 2023-07-06
feat: Region/Lasso Selections

arvind:
Branching off #8104 to test if that fixes CI errors we were experiencing previously (and to validate freshness of the previous PR, fix bugs, etc.). 

I will be sending a separate PR for updating the documentation because I think a bigger change is needed than what falls in scope for this feature/PR. 

HUGE thanks to @dvmoritzschoefl for the contribution!


kanitw (approved):

domoritz:
🎉

PR #9015, opened 2023-07-18
feat: Improve color scheme type

mcnuttandrew:
The JSON Schema that vega-lite generates is wonderful, but there are a few places where the types are slightly under-defined because they are defined as part of the vega type system. 

One such example is color schemes, which are more permissive than should be allowing any string. While the included description is helpful for identifying the allow values ("A string indicating a color [scheme](https://vega.github.io/vega-lite/docs/scale.html#scheme) name (e.g., `"category10"` or `"blues"`) or a [scheme parameter object](https://vega.github.io/vega-lite/docs/scale.html#scheme-params)...."), this is not very helpful systems ingesting the schema computationally, such as for structure editing.

Fortunately, a possible workaround is to explicitly import the wonderful vega-typings package (which was already an implied dependency, cf yarn.lock). This allows use of the more explicit type described by vega-typings. 

Please:

- [X] Make the pull requests (PRs) atomic (fix one issue at a time). Multiple relevant issues that must be fixed together? Make atomic commits so we can easily review each issue.
- [X] Provide a concise title as a [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties") so we can easily copy it to the release note.
- [X] Lint and test (Run `yarn test`).
- [X] Review your changes before sending the PR (to ensure code quality).
- For new features:
  - [ ] Add new unit tests.

Tips:


package.json
```diff
@@ -127,6 +127,7 @@
     "vega-datasets": "^2.7.0",
     "vega-embed": "^6.22.1",
     "vega-tooltip": "^0.32.0",
+    "vega-typings": "^0.24.1",
```
domoritz, line 130: Since we already depend on Vega, I don’t think we need this. It just makes updates more annoying.

examples/compiled/interactive_point_domainRaw_binding.png
domoritz, line 1: Why does this change?

src/scale.ts
```diff
@@ -458,7 +459,7 @@ export interface SchemeParams {
    *
    * For the full list of supported schemes, please refer to the [Vega Scheme](https://vega.github.io/vega/docs/schemes/#reference) reference.
    */
-  name: string | SignalRef;
+  name: string | ColorScheme | SignalRef;
```
domoritz, line 462: This is great. Should we even allow string?

domoritz (approved):

src/scale.ts
mcnuttandrew: Good point! I went with a more permissive type because I wasn't sure what the down stream deps were from this type, but if this seems fine to you then a more assertive seemed great!

examples/compiled/interactive_point_domainRaw_binding.png
mcnuttandrew: accidentally got generated bc my auto formatter surprisingly decided to remove a type, who knows

mcnuttandrew: i think i already have?

domoritz:
Thank you!

PBI-David:
Will this still allow things like the following which I can do in Vega?
```

"range": {
        "scheme": [
          "#1860cc",
          "#18d2d9",
          "#18cc36"
        ]
      }
```

domoritz:
Screenshot 2023-07-20 at 08 47 50

We actually didn't even support that in the first place, but should!

domoritz:
I filed https://github.com/vega/vega-lite/issues/9022

domoritz:
@PBI-David you can already set the range to a list: https://vega.github.io/vega-lite/docs/scale.html#2-setting-the-range-property-to-an-array-of-valid-css-color-strings

Issue #9001, opened 2023-07-13
`encoding.order` sort the path not the stack for Q-Q area chart


yhoonkim:
![image](https://github.com/vega/vega-lite/assets/6269456/6883d3c5-f542-49f8-9c65-38a1e1bc746e)

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "data": {
    "values": [
      {"a": 1, "b": 2, "c": "c1"},
      {"a": 2, "b": 3, "c": "c1"},
      {"a": 3, "b": 1, "c": "c1"},
      {"a": 1, "b": 2, "c": "c2"},
      {"a": 3, "b": 1, "c": "c2"},
      {"a": 2, "b": 2, "c": "c2"}
    ]
  },
  "mark": "area",
  "encoding": {
    "x": {"field": "a", "type": "quantitative"},
    "y": {"field": "b", "type": "quantitative"},
    "color": {"field": "c"},
    "order": {"field": "c"}
  }
}
```

The main reason is that Area chart is not stacking with x:Q and y:Q by default so that "order" channel apply the sort on the path rather than the stacking order. 

I suggest to add "stack" by default (stacking the color even without `stack` def like bar).

joelostblom:
This was fixed in #9018, so closing. The example above now looks like expected:

<img width="921" height="514" alt="Image" src="https://github.com/user-attachments/assets/423efc14-ddb5-4ddf-82e5-39be95076762" />

PR #9018, opened 2023-07-19
feat: Area with plain quantitative fields on both axes get stacked by default

yhoonkim:
Address this issue: https://github.com/vega/vega-lite/issues/9001


examples/specs/area_cumulative_freq.vl.json
```diff
@@ -14,7 +14,8 @@
     },
     "y": {
       "field": "Cumulative Count",
-      "type": "quantitative"
+      "type": "quantitative",
+      "aggregate": "max"
```
yhoonkim, line 18: Note that this PR fixes this example which uses `window` transform and plot the chart but without aggregation. The backed data is not aggregated so internally it draws many more unnecessary data points.

Those unnecessary data points break the area chart with the new default stack in area.

src/normalize/pathoverlay.ts
```diff
@@ -130,7 +132,7 @@ export class PathOverlayNormalizer implements NonFacetUnitNormalizer<UnitSpecWit
     // FIXME: determine rules for applying selections.
 
     // Need to copy stack config to overlayed layer
-    const stackProps = stack(markDef, encoding);
+    const stackProps = stack(initMarkdef(markDef, encoding, config), encoding);
```
kanitw, line 135: This is technically a tech debt, since normalizer shouldn't call methods from an init phase. 

So I'd at least add

```ts
// FIXME: normalizer shouldn't call methods from an init phase. 
```

examples/specs/area_cumulative_freq.vl.json
kanitw: (nit) A more efficient fix is probably aggregate the count before calling window, but feel free to file a follow up ticket if u don't wanna fix now.

kanitw (approved):

joelostblom:
@yhoonkim @kanitw I have noticed issues with densities getting stacked by default when using the area mark and I believe it comes from this PR, see https://github.com/vega/vega-lite/issues/9170 for an example. This is undesirable because it makes it hard to compare the two distributions with each other. Is there any way to make this PR compatible with the fact that we don't want to stack densities by default (while one could use a line for the density transform, it is also common to use an area and it is unexpected that they differ as well)? Stacking the density areas used to [require the stack to be explicitly set like in this example](https://vega.github.io/vega-lite/examples/area_density_stacked.html).

PR #9019, opened 2023-07-19
fix: Escape Quotes in tooltip `title`

lsh:
Currently titles such as `"\"mytitle\""` will cause the editor to break because it generates fields that look like:
```
({ ""mytitle"":  ... })
```
This PR escapes quotes to allow this case to pass.


kanitw (commented):
Can you add an example in `examples/specs` to show that this really does make it renderable?

kanitw (approved):

src/compile/mark/encode/tooltip.ts
```diff
@@ -88,7 +88,7 @@ export function tooltipData(
         };
 
     const title = fieldDef.title || defaultTitle(fieldDef, formatConfig);
-    const key = array(title).join(', ');
+    const key = array(title).join(', ').replaceAll(/"/g, '\\"');
```
domoritz, line 91: Huh, do we not have a utility method for this already? Does this problem not happen for other texts, axis labels or titles?

lsh: I didn't see one, but maybe I didn't look in the right place. I'm not sure the other areas have the same problem since this one specifically generates a dictionary rather than acts as a value.

PR #9066, opened 2023-08-22
fix: update ts and fix typing issues

domoritz:
@arvind it looks like `fields` wasn't in interval selection configs but [the docs](https://vega.github.io/vega-lite/docs/selection.html) and tests say it's a shared config option. Does this look right to you that I moved `filelds`?

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -12472,6 +12472,13 @@
           },
           "type": "array"
         },
+        "fields": {
+          "description": "An array of field names whose values must match for a data tuple to fall within the selection.\n\n__See also:__ The [projection with `encodings` and `fields` section](https://vega.github.io/vega-lite/docs/selection.html#project) in the documentation.",
+          "items": {
+            "$ref": "#/definitions/FieldName"
+          },
+          "type": "array"
+        },
         "mark": {
           "$ref": "#/definitions/BrushConfig",
           "description": "An interval selection also adds a rectangle mark to depict the extents of the interval. The `mark` property can be used to customize the appearance of the mark.\n\n__See also:__ [`mark` examples](https://vega.github.io/vega-lite/docs/selection.html#mark) in the documentation."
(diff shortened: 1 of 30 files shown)


lsh (approved):
Looks good, thanks!

joelostblom:
This PR added the `fields` param to the interval selection config, but [the docs say that interval selections only work with `encodings` and not `fields`](https://vega.github.io/vega-lite/docs/selection.html#current-limitations). After this PR, `fields` now show up as an accepted parameter in interval selections, but setting it does not seem to have an effect [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICUJNohSZ8gL5KcppAghoA2qFkO4+ptoYQYikAg4QTgoTDRQTABPHE90GllMOG1iJD0lADMaYOUnVGcQAAk2MzgcNgB3ZICAWRoQiAB9WO1GgHE0wTkQAF0rKx6lZG0Aa31yhPClOFkoNmUEsgiQAA9lrJz9YtLyqu0A6Nj9AEcGJES6NRpSEBsQKPXswWV9Oobm5PbO7qVDuJBTudMJdgTc7nMuvtNCA5rIFsDutC7NoHF4fH4AhtnvoAMJRYRw5JOX4xf4lBbuPR3VKCBj-MjaOAPfpWIA). Should we remove `fields` from the interval config until it works to create a spec such as in the linked chart?

arvind:
Sorry that this slipped off my radar. `fields` is not a valid property of interval selections—they only take `encodings` as @joelostblom mentioned.

domoritz:
I see. We should undo the specific change then. https://vega.github.io/vega-lite/docs/selection.html#selection-props is incorrect then, right?

Screenshot 2023-09-29 at 16 34 10

PR #9068, opened 2023-08-23
feat: switch from clone to built in structured clone

domoritz:
* update jest
* remove dependencies (no more commonjs deps)


lsh (approved):
Looks good to me. I'm not super keen on the use of `any`, but I can't think of a better way to achieve the same result.

kanitw:
Why is this a feat?

PR #9099, opened 2023-09-20
ci: lint pull request messages

domoritz:
Using https://github.com/marketplace/actions/semantic-pull-request


jonmmease:
This looks good, but I don't see GitHub Actions actually running it

domoritz:
I think it only runs after this is merged which makes it hard to test admittedly.

jonmmease (approved):
Let's give it a try

PR #9101, opened 2023-09-20
ci: monthly updates

domoritz:
Reduce update cadence.

To reduce even further, we can use [groups](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) but this is the first step.


arvind (approved):
Hallelujah! 🙌 🙏🏼

Issue #9078, opened 2023-09-04
Grouped density charts have artifacts (again)


jonmmease:
This was originally reported in https://github.com/hex-inc/vegafusion/issues/381, but it appears to be a Vega-Lite issue.  Given this grouped density Vega-Lite spec:

```json
{
  "config": {"view": {"continuousWidth": 300, "continuousHeight": 300}},
  "data": {
    "url": "https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/movies.json"
  },
  "mark": {"type": "area"},
  "encoding": {
    "color": {"field": "Major_Genre", "type": "nominal"},
    "x": {"field": "IMDB_Rating", "type": "quantitative"},
    "y": {"field": "density", "type": "quantitative"}
  },
  "transform": [
    {"filter": {"field": "Major_Genre", "oneOf": ["Action", "Drama"]}},
    {
      "density": "IMDB_Rating",
      "groupby": ["Major_Genre"],
      "as": ["IMDB_Rating", "density"]
    }
  ],
  "$schema": "https://vega.github.io/schema/vega-lite/v5.14.1.json"
}
```

The resulting density curves have a lot of artifacts.

![visualization (1)](https://github.com/vega/vega-lite/assets/15064365/cd5171ad-4cb2-4a1f-9e38-89c4cca8a8ff)

A related case was reported in https://github.com/vega/vega-lite/issues/8049, and the suggested fix was to add an explicit `steps` argument to the Vega `kde` transform that Vega-Lite produces. This change was made in https://github.com/vega/vega-lite/pull/8088, but this doesn't seem to take care of the issue universally.

Rather than (or in addition to) `steps` I think what we need is to specify `"resolve": "shared"` in the kde transform so that the generated x-values are in sync across groups. This chart is fixed when manually adding ``"resolve": "shared"`` to the generated Vega spec:

[...]

PR #9106, opened 2023-09-20
fix: use resolve "shared" instead of steps to fix artifacts in grouped density transform

jonmmease:
Closes https://github.com/vega/vega-lite/issues/9078

### DIFF
--- a/examples/compiled/area_density_stacked_fold.vg.json
+++ b/examples/compiled/area_density_stacked_fold.vg.json
@@ -24,7 +24,8 @@
           "extent": [10, 500],
           "counts": true,
           "steps": 200,
-          "as": ["value", "density"]
+          "as": ["value", "density"],
+          "resolve": "shared"
         },
         {
           "type": "impute",
--- a/examples/compiled/area_density_facet.vg.json
+++ b/examples/compiled/area_density_facet.vg.json
@@ -15,7 +15,7 @@
           "groupby": ["Species"],
           "extent": [2500, 6500],
           "as": ["value", "density"],
-          "steps": 200
+          "resolve": "shared"
         },
         {
           "type": "impute",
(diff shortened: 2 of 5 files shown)


examples/compiled/area_density_stacked_fold.vg.json
```diff
@@ -24,7 +24,8 @@
           "extent": [10, 500],
           "counts": true,
           "steps": 200,
-          "as": ["value", "density"]
+          "as": ["value", "density"],
+          "resolve": "shared"
```
domoritz, line 28: Should this one change?

domoritz (approved):

examples/compiled/area_density_stacked_fold.vg.json
jonmmease: Yeah, the input Vega-Lite spec has:

```json
    {
      "density": "value",
      "bandwidth": 10,
      "groupby": ["Measurement"],
      "extent": [10, 500],
      "counts": true,
      "steps": 200
    }
```
Since `groupby` is defined we get `"resolve": "shared"`. `"steps": 200` is still there because it's explicitly specified in the Vega-Lite transform.

joelostblom:
I'm teaching a class starting middle of next week which would benefit from having this included. Is there any chance for a patch release of VL that includes this fix before then? It is not critical as I can work around it by teaching `resolve_scale` manually or use and older version of Altair, but just wanted to check-in, in case anyone has the bandwidth.

domoritz:
Yes. I think either @kanitw and @lsh make a release after https://github.com/vega/vega-lite/pull/9111 before the weekend or I will make a release with this fix.

joelostblom:
Awesome, thank you!!

domoritz:
https://github.com/vega/vega-lite/releases/tag/v5.15.1

joelostblom:
Is there a build process that takes some time before the URL https://vega.github.io/schema/vega-lite/v5.15.1.json goes live? Or is there a separate manual step for that?

domoritz:
There is a build process but I got a failure email. @kanitw can you check?

domoritz:
Reran it and worked now.

PR #9111, opened 2023-09-29
fix: correct the formula for interpolating between bin start and end (interpolatedSignalRef)

kanitw:
The new formula correctly makes bandPosition = 0 represent start and bandPosition = 1 represents end.
(The existing code wasn't consistent with the simplified formula when bandPosition = 0 or 1)

```ts
if (bandPosition === 0 || bandPosition === 1) {
    ref.scale = scaleName;
    const field = bandPosition === 0 ? start : end;
    ref.field = field;
  }
```


lsh:
Code looks good but I'd encourage checking to make sure this doesn't result in any stale documentation. Alternatively if this is now an invariant, maybe we should be documenting it?

kanitw:
Btw, after checking more, I can confirm that the resulting diff is correct. 

(The formula was indeed wrong.  One example is that in the old code, band=0.99 is wider than band=1, simply because the offset to add 1 pixel is applied in the wrong direction due to the wrong formula.

## Before

image

image

## After 

image

lsh (approved):
Looked through the diff and no complaints.

Issue #9025, opened 2023-07-21
`timeUnitBandPosition` on axes


nicolaskruchten:
If I could work up a pull request, would there be any appetite for a `timeUnitBandPosition`-like property that could be set on axes, so as to display, say, month names at the midpoint of the months, such that monthly bars would be centered over the labels?

nicolaskruchten:
To elaborate a little bit... :)

## The Problem

The current way to do axis labels for temporal axes looks like this when using monthly bars: the bars span the months and the gridlines and labels are at the month starts, labelled with the month name. (in the example below I'm dropping the month of November on purpose... read on!)

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "data": {"url": "data/movies.json"},
  "transform": [{"filter": "month(datum['Release Date']) != 10"}],
  "mark": "bar",
  "encoding": {
    "x": {"timeUnit": "month", "field": "Release Date", "type": "temporal"},
    "y": {"aggregate": "mean", "field": "US Gross"}
  },
  "height": 400,
  "width": 600
}
```

<img width="704" alt="image" src="https://github.com/vega/vega-lite/assets/203523/1cc114c6-e130-4324-8e70-09f9a65ecd60">

The problem here is that some people find this labelling confusing: they associate one bar to one month, and here the month labels straddle bars, forcing readers to disambiguate which bar belongs to which month. Intuitively, part of the problem here is that we are labelling an instant (the beginning of the month of February, so February 1 @ 00:00) as the month itself "Feb".

We could switch the encoding to be ordinal, as [suggested in the docs](https://vega.github.io/vega-lite/docs/timeunit.html#time-unit-with-ordinal-fields) but this can be misleading if a value is missing, for example as above and below where I'm dropping November:


```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "data": {"url": "data/movies.json"},
  "transform": [{"filter": "month(datum['Release Date']) != 10"}],
  "mark": "bar",
  "encoding": {
    "x": {"timeUnit": "month", "field": "Release Date", "type": "ordinal"},
    "y": {"aggregate": "mean", "field": "US Gross"}
  },
  "height": 400,
  "width": 600
}
```

<img width="695" alt="image" src="https://github.com/vega/vega-lite/assets/203523/10de23a4-9f78-4cfe-a14c-a8f8397cbc8a">

## Possible Workarounds

[...]

PR #9126, opened 2023-10-05
feat: make bar & rect marks support bandPosition (allow center-align bars and ticks)

kanitw:
feat: make bar & rect marks support bandPosition (allow center-align bars and ticks)!

(Address concerns discussed in #9025.)

Example: (left = old style, right = center align bar and ticks) 

image

This also works for grouped bars

image


fandu-db (approved):
LGTM!

PR #9132, opened 2023-10-11
feat: 'move' cursor added over brush mark

alliefeldman:
To create affordance for brush panning, switch the user's mouse cursor to 'move' when dragging/hovering over a displayed interval selection.

Screenshot 2023-10-11 at 11 27 21 AM

Contributes to #4155


arvind:
This is an exciting contribution! As we work through the feature and associated PRs, I thought I'd offer one gut check (which you may have already thought of and done, but just in case): not all interval selections will necessarily be pannable/zoomable (i.e., users may choose to set `translate` or `zoom` to `false`). Thus, it may make more sense for cursors to be properties of the translate/zoom selection compilers (e.g., hooking into the `marks` callback and making cursor overrides as need be) than to have them be situated in the interval selection compiler/config (as I _think_ this PR is currently proposing—though I may be mistaken).

PR #9136, opened 2023-10-12
docs(examples): Add example showing how to set the zorder via a conditional order encoding

joelostblom:
This adds an example of how to make a point appear on top when it is hovered over, instead of it being covered by other points based on the initial plotting order. This was discussed in https://github.com/vega/vega-lite/issues/4864 and https://github.com/vega/vega-lite/issues/1684, but I couldn't find an example that demonstrates this behavior. I would presume that setting the zorder like this is more performant than having a second layer that responds to a filter transform linked to to the selection, which is what I have used myself in the past.

image
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygK4CcA28QAspQA4Gc4B6IgYzADsA6AK3zAFNUBLAN2SooaiItwFsirBgHMIAWkgx83fAAFWARioAmAJxUADESkQyEZPlr4A9hRABfADQh+BgNbxQUAJ64GGUs2SlUHmya4EF6u8IrWIAwUpCZgzBQiTiAxqCbISTEUcVDMZklByBD8GAVFAPqKKiogNqwQqIgecMlp9TWR-LihcABm9TIRdQ1NICLIDC6WAciM6QggQ43wmjaZ2bnm86XFzdsVVe2LTYo2DJ3dfagDEfjMAF5NoGvMOXlbBkUlH-z71bX1SzgABZNCsOl1Jr1+gxBgDjqCIgAPJI9ZhMMAYAASaRkuBMAHcGOkbK53BgAI6ICAUHIwHLCKYgSGgVHojAAWWYfnwZXcyDKAHF6qlzCS3CNKdTadA2B4LBFtvh4ABtUAUIojPaVP4gGR+UhQJKkzUmeKGgKbWwmRAyEzCdLygC6FiAA)


examples/specs/selection_type_point_zorder.vl.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "mark": {"type": "circle", "opacity": 1},
+  "encoding": {
+    "color": {
+      "condition": {"param": "param_122", "value": "coral", "empty": false},
+      "value": "grey"
+    },
+    "order": {
+      "value": 0,
+      "condition": {"param": "param_122", "value": 1, "empty": false}
```
kanitw, line 14: For consisistency with size and to make it reads like if-then-else, better put `condition` above `value`.

site/docs/encoding.md
```diff
@@ -413,7 +413,7 @@ The key channel can enable object constancy for transitions over dynamic data. W
 
 ## Order Channel
 
-`order` channel can define a data field (or a ordered list of data fields) that are used to sorts stacking order for stacked charts (see [an example in the `stack` page](stack.html#order)) and the order of data points in line marks for connected scatterplots (see [an example in the `line` page](line.html#connected-scatter-plot)).
+`order` channel can define a data field (or a ordered list of data fields) that are used to sorts stacking order for stacked charts (see [an example in the `stack` page](stack.html#order)), the order of data points in line marks for connected scatterplots (see [an example in the `line` page](line.html#connected-scatter-plot)), and which data points are plotted on top in a chart (the "zorder", see [an example in the gallery]({{ site.baseurl }}/examples/selection_type_point_zorder.html)).
```
kanitw, line 416: (the "zorder", see [an example in the gallery]({{ site.baseurl }}/examples/selection_type_point_zorder.html)) --> I think this URL probably doesn't work unless you add it to `examples.json`.

I think you can add this example to the end of the following block in `examples.json`. :) 

```
"Interactive": {
    "Interactive Charts": [
```

kanitw:
Thanks for adding the example! :)

joelostblom:
Thanks for the quick review @kanitw ! I believe I have addressed your comment in my last couple of commits, so please have another look.

kanitw (approved):

PR #9144, opened 2023-10-19
docs: add pygwalker and gwalkr

ObservedObserver:
Add [pygwalker](https://github.com/Kanaries/pygwalker) and [GWalkR](https://github.com/Kanaries/GWalkR) to ecosystem.

I think there was a PR accepted before: https://github.com/vega/vega-lite/pull/8735 , but its target branch is not merged into master.


domoritz:
Good catch and thanks for the new pull request.

PR #9171, opened 2023-11-11
docs: update valid image path

### LINKED ISSUE #9151: Missing tooltip images in docs
At https://vega.github.io/vega-lite/docs/tooltip.html#tooltip-image the image doesn't show up. 

Screenshot 2023-10-23 at 11 47 04

ChiaLingWeng:
This is an attempt to fix #9151 ,

I found that this issue is due to invalid image path and find ` "image": "../data/ffox.png"` is the right valid for building site,
but it's weird that for the vega-lite editor, the right path is the original  `"data/ffpx.png"`, same as the image mark path,
since I'm quite new to vega-lite repo, I would like to ask for some instruction to start investigate this, thanks


site/_data/versions.yml
domoritz, line 1: This should not be included.

examples/specs/text_tooltip_image.vl.json
```diff
@@ -1,18 +1,35 @@
...
   "data": {
     "values": [
-      {"image": "data/ffox.png"},
-      {"image": "data/gimp.png"},
-      {"image": "data/7zip.png"}
+      {
+        "image": "data/ffox.png"
+      },
+      {
+        "image": "../data/gimp.png"
```
domoritz, line 13: I don't think this is enough since other places depend on the previous location. Why did this break in the first place? How do we make sure that e.g. the editor still works correctly?

domoritz (changes requested):

domoritz:
Thanks for taking a stab at this!

Issue #6624, opened 2020-06-15
Density visualizations with group by are not stacking correctly


domoritz:
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JACyUAVhDYA7EABoQAdxoATemjEAGbUvg0yWNAEZdStUkwpUoBgCdBaEJesEoSexCkz5AXyVkewBrZ084FCVMeyRZCAAzNnsENABtUETBNTSQABEaCBxBJChEOFlMRRAACSSIOBw2ZTh7EABdJSQIHIQIiAcyiqriJEEGOHaA0DVyiDoAT2cRsYmlMns2BhwmRdRUkF6ugd6htr8OkHKoNjUaWTI0UAAPR5B4mjgspdHxqsx5nATdAARwYsUwdCsNFIIACIF2GQ+X3QMziCz+AKBIFB4MhEJhSgg1igoXQAC8WmxYUproIkq93p9suhDv17INKlFMc5ZGwEHdRrC-H4gA)

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
  "width": 400,
  "height": 100,
  "data": {
    "url": "data/cars.json"
  },
  "mark": "area",
  "transform": [
    {
      "fold": ["Displacement", "Horsepower"],
      "as": ["measurement", "value"]
    },
    {
      "density": "value",
      "groupby": ["measurement"]
    }
  ],
  "encoding": {
    "x": {"field": "value", "type": "quantitative"},
    "y": {"field": "density", "type": "quantitative", "stack": "zero"},
    "color": {"field": "measurement", "type": "nominal"}
  }
}
```

![image](https://user-images.githubusercontent.com/589034/84685025-5fbea300-aeee-11ea-83b6-a4dde7b37f3b.png)

domoritz:
The issue is that we use impute to fill in missing densities as 0 but the densities are not signed.

joelostblom:
This is currently working as intended, so I'm closing the issue. I believe it was solved in https://github.com/vega/vega-lite/pull/9106

![image](https://github.com/vega/vega-lite/assets/4560057/99915472-b06b-419f-a985-6898dfb78879)

Issue #5611, opened 2019-12-02
Don't add impute for violin plot


domoritz:
```json
{
  "data": {"url": "data/iris.json"},
  "transform": [
    {
      "fold": ["sepalLength", "sepalWidth",
      "petalLength", "petalWidth"],
      "as": ["feature", "value"]
    },
    {
      "density": "value",
      "extent": [0, 8],
      "groupby": ["feature"]
    }
  ],
  "mark": {"type": "area", "orient": "horizontal"},
  "encoding": {
    "column": {"type": "nominal", "field": "feature"},
    "x": {"type": "quantitative", "field": "density", "stack": "center"},
    "y": {"type": "quantitative", "field": "value"}
  },
  "width": 60
}
```

@jheer said: I took a closer look and the culprit is not a sorting issue, but rather the auto-magical inclusion of an impute  transform that has no business being there. By default, Vega performs adaptive sampling to determine which points along the density curve to include. As this can result in different sample points for the different areas, their domain values should not be used together to perform imputation. @domoritz, @kanitw I think this needs to be fixed prior to a v4 release.

domoritz:
In the future, we will provide alignment to create violin plots. For now, we won't change anything about the spec above (I also don't know how what we would change) but we should allow disabling of imputation like this: 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
  "data": {"url": "data/iris.json"},
  "transform": [
    {
      "fold": ["sepalLength", "sepalWidth", "petalLength", "petalWidth"],
      "as": ["feature", "value"]
    },
    {"density": "value", "extent": [0, 8], "groupby": ["feature"]}
  ],
  "mark": {"type": "area", "orient": "horizontal"},
  "encoding": {
    "column": {"type": "quantitative", "field": "feature"},
    "x": {
      "type": "quantitative",
      "field": "density",
      "stack": "center",
      "impute": null
    },
    "y": {"type": "quantitative", "field": "value"}
  },
  "width": 60
}
```

jheer:
Not clear to me why imputation would be opt out rather than opt in. Also, if the same logic also applies to line marks we have the same problem for regression lines. In general it is simply not correct to assume that different groups within a groupby should have identical x/y domain values.

domoritz:
I'm looking into why we initially decided to add imputation by default. Maybe we can disable it.

domoritz:
We are currently only adding imputation when stacking is on.

[...]

PR #9172, opened 2023-11-11
feat: add explicit option to control how densities are resolved, change how densities are resolved by default

joelostblom:
Will add more description soon, just seeing if tests pass first because my local installation doesn't work.


joelostblom:
I got stuck playing trying to fix the issues with density area marks, so I thought I would summarize what I'm trying to do so that others can help: 

I believe the recent issues with density marks stem from PR https://github.com/vega/vega-lite/pull/9018, which made stacking the new default for area marks. Unfortunately stacked area marks also perform an imputation to zero by default as described here https://github.com/vega/vega-lite/issues/5611#issuecomment-561039336 by @domoritz  which leads to issues with stacked density as previously noted in https://github.com/vega/vega-lite/issues/6624 (this became more obvious after the PR mentioned above since it made stacked areas the default). Thanks @mattijn for pointing out that this imputation was happening in https://github.com/vega/vega/issues/3815.

The reason the imputation leads to jagged densities is because grouped densities value are defined only for data values existing within their own group. So when putting two grouped densities together on the same axis, but without exactly the same extent over which the density is defined will lead to a lot of imputation to zero and spiky appearance. In https://github.com/vega/vega-lite/pull/9106 @jonmmease  fixed this by guaranteeing the same extent for all individual grouped densities. However this introduced a new issue, which is that densities are no longer cut off at the min/max values of the data within each group, but extends to the min/max values of the data within all groups leading to the appearance of values where there actually are none as shown in https://github.com/vega/vega/issues/3815.

---

For density area marks to work properly, we need these two things:

1. Grouped density transforms should be limited to the extent of the data within each group to not avoid suggesting non-existing values.
    - This can be fixed if we go back to setting `"resolve": "indepedent""` in the density transform. If we do that, we need area marks used for densities to not be imputed to zero by default. I think the current zero imputation is not ideal for non-density areas either, sine it gives the impression of non-existing values as can be seen in . Maybe the suggestion in https://github.com/vega/vega-lite/issues/5611#issuecomment-561039488 from @jheer  to use Vega's `xc` channel instead of the stack could help?
    - Another way of fixing this would be to keep `"resolve": "shared"`, but we trim the densities after. Note that we can't simply trim values where the density is 0, but need to trim all values outside the original data range for each group (because we need access to each group's min/max values, I believe these needs to happen inside the KDE transform as suggested in https://github.com/vega/vega/issues/3815).
2. By default area marks used for densities should not be stacked since it makes it harder to compare the distributions with each other. 
    - I'm unsure how to solve this in a way that is compatible with the desire to stack other area marks by default as per https://github.com/vega/vega-lite/pull/9018. We can't simply always turn off stacking for density marks since that would not work together with use cases when stacking is explicitly desired. Personally, I would be in favor of rolling back #9018 since I think comparing distributions via densities is more common than the use case described for that PR, but if there is a way to have both that's of course ideal.

My hunch would be that limiting the range of each grouped density in the KDE transform in Vega and rolling back #9018, would be the most straightforward way to fix this, but I'm very open to hearing other suggestions.

joelostblom:
Here is another example of how the current behavior leads to unexpected outcomes. Take this spec that creates two densities which seem to cover the same data range:

image
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO50AhoTl4QpAE4AbFCDGEADpWQB6Q2DpQAdACtKdTOrhpV5qJkKGougLaG0mBHIBaeUVKV0oAATQARnMAJgBOczZDYLlDTwgMTEorSmgQGRAYOTBXZXg7Oi0AWTlLCFUAfQBxTChVTBBuQgBPXU7kECgITxI5TULKfrBldTkezFUUAG1QTzlVAGtlXv6tDcwlQrbIOhIkVBAAD3KsdSrBgElqgBEAIUaAJQVzrpBdgYgACOpDkRDgimIvgK3B6t0qWlsUEoELh3T6gJBYOIkIcnWk3VUYMoMAanhWoCRKN6Wme7y+PygSG4CFUFF0ACM4chliBavUmq12p0ALrcOSUFYgOkfb58ZkgKmokAimSgeDqQiLeH3Gp1BotNodP7QTAAeRgUoAgmBiPluG85mBNgACADCI0wdDhqskqrF-yJyNJqnJPPVcE12suFV1g35BqFxu4pot1ttcHtIEdJVdHs8Xp90lV3Eg6lInmRKAALNwACSUMCMTDrLQ6fRGHx+OTmBAQxikDnmTOGRvN9Zd-wBexanwAVnMUQXMWs+UkQA)

If we add another chart to the layer, such as a mean line (but really any chart), suddenly it becomes apparent that the densities don't at all cover the same region and the one to the right now appears cut off for no reason since it looked like the extent was the same in the previous chart:

image
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO50AhoTl4QpAE4AbFCDGEADpWQB6Q2DpQAdACtKdTOrhpV5qJkKGougLaG0mBHIBaeUVKV0oAATQARnMAJgBOczZDYLlDTwgMTEorSmgQGRAYOTBXZXg7Oi0AWTlLCFUAfQBxTChVTBBuQgBPXU7kECgITxI5TULKfrBldTkezFUUAG1QTzlVAGtlXv6tDcwlQrbIOhIkVBAAD3KsdSrBgElqgBEAIUaAJQVzrpBdgYgACOpDkRDgimIvgK3B6t0qWlsUEoELh3T6gJBYOIkIcnWk3VUYMoMAanhWoCRKN6Wme7y+PygSG4CFUFF0ACM4chliBavUmq12p0ALrcOSUFYgOkfb58ZkgKmokAimSgeDqQiLeH3Gp1BotNodP7QTAAeRgUoAgmBiPluG85mBNgACADCI0wdDhqskqs4aw220uAK0qlI6nx3BOEDOTOUN0ucgQrL8CkBnkOUD+FV1T1essZCtDgyx4Nx0OkfsJxNJqnJPPVcE12suuYefP1gqNnW4pot1ttcHtIEdJVdHsz3pV0n9PHUpE8yJQABZuAASShgRiYdZaHT6Iw+NPmBAQxikDnmYeGLc79bH-wBexanwAVnMUQ-MWs+UkQA)

The reason seems to be that the layering is moving the transforms around, but this is not easy to understand and the inconsistent behavior is quite confusing.

joelostblom:
Gentle ping on this @jheer, @domoritz, and @kanitw. What do you think is the best way forward to fix the density specs? The decision we take here will also have an impact on the violin plots we have been talking about.

My hunch would be that limiting the range of each grouped density in the KDE transform in Vega as suggested in https://github.com/vega/vega/issues/3815, and rolling back https://github.com/vega/vega-lite/pull/9018 would be the most straightforward way to fix this, but I'm very open to hearing other suggestions.

examples/compiled/area_density_facet.png
domoritz, line 1: This example is getting worse.

examples/compiled/area_density_facet.svg
domoritz, line 1: This example is getting worse.

examples/compiled/area_density_stacked.png
domoritz, line 1: This example is getting worse.

examples/compiled/area_density_stacked.svg
domoritz, line 1: This example is getting worse.

src/transform.ts
```diff
@@ -496,6 +496,14 @@ export interface DensityTransform {
    * __Default value:__ `["value", "density"]`
    */
   as?: [FieldName, FieldName];
+  /**
+   * Indicates how parameters for multiple densities should be resolved.
+   * If "independent", each density may have its own domain extent and dynamic number of curve sample steps.
+   * If "shared", the KDE transform will ensure that all densities are defined over a shared domain and curve steps, enabling stacking.
```
domoritz, line 502:
```suggestion
   * If `"independent"`, each density may have its own domain extent and dynamic number of curve sample steps.
   * If `"shared"`, the KDE transform will ensure that all densities are defined over a shared domain and curve steps, enabling stacking.
```

domoritz (changes requested):

domoritz:
If densities should be handled differently from other area marks when it comes to stacking, I think it would make sense to introduce enough knobs to support handling them differently and then to introduce a higher-level mark type (which could even create the density transform as well). I know @jheer was advocating for that for a long time 😅 so I'd say let's use this opportunity to do that.

Imputation by default has issues but I still think it's a reasonable choice (as discussed in https://github.com/vega/vega-lite/issues/5611). Would introducing xc as a channel as an alternative to centering a stacked area mark help as well? It would make for a more canonical representation of densities (https://vega.github.io/vega/examples/violin-plot/), I think. 

@yhoonkim and @kanitw since you introduced https://github.com/vega/vega-lite/pull/9018, what are your thoughts on the issue/reverting?

domoritz:
Outcome from meeting: let's get this option added but not change the default for now.

examples/compiled/area_density.vg.json
```diff
@@ -15,7 +15,8 @@
           "type": "kde",
           "field": "IMDB Rating",
           "bandwidth": 0.3,
-          "as": ["value", "density"]
+          "as": ["value", "density"],
+          "resolve": "shared"
```
domoritz, line 19: This used to be `independent`. https://vega.github.io/vega/docs/transforms/kde/

Why should this be shared for non-stacked transforms? Maybe a reasonable default is to have it only shared if there is stacking?

test/compile/data/density.test.ts
```diff
@@ -54,14 +55,16 @@ describe('compile/data/fold', () => {
       expect(density.assemble()).toEqual({
         type: 'kde',
         field: 'v',
+        resolve: 'shared',
         as: ['A', 'density']
       });
     });
 
-    it('should add resolve shared if we group', () => {
+    it('should only add resolve "shared" if we set it explicitly', () => {
```
domoritz, line 63: To test this, we would need a case that doesn't set `resolve` when it's not set in the transform.

domoritz (changes requested):
Let's consider the default and make sure we test different cases where resolve is set and not set in unit tests.

examples/compiled/area_density.vg.json
joelostblom: If I remember correctly from our discussion, we said it would be difficult to set the default value of the transform based on whether the encoding channel is stacked or not. I believe the main reason was that it would be hard to detect which encoding channel the density values are plotted on.

Leaving "shared" as the default would work in most situations, including most of the common ones, and if there are situations where the "independent" is needed, it would be easy to switch. To avoid that densities stack by default, we said we would turn off stacking as part of a `mark_density`, but leave the default stacking behavior for the area mark as it is now to be consistent with other marks.

test/compile/data/density.test.ts
joelostblom: Won't `resolve` always be in the transform sine the parameter has a default value? I think this test case should just be switched to "independent" instead of "shared" to reflect the updates in e3f4505d06d0f8a6c7e8d1d0900e85c1fdfdb8d0. I did that in my latest commit.

joelostblom (commented):
Ugh, I'm really sorry, I missed that you had left new comments a while ago @domoritz. I just responded to the two comments you made and pushed a new commit for the test, let me know  what you think.

domoritz:
Thank. I will take a look. Can you merge the latest main and make sure that the CI runs?

domoritz (approved):

PR #9196, opened 2023-12-07
docs: update valid image path

### LINKED ISSUE #9151: Missing tooltip images in docs

ChiaLingWeng:
this is a follow up to #9171 and will close #9151.
since the original pull request is a little messy and the workflow failed thus open a new one here, 
for this issue, the most graceful solution I think is to add a symbolic link that points to the valid data folder.

image


domoritz:
Hmm, https://vega.github.io/vega-lite/docs/image.html#examples works but https://vega.github.io/vega-lite/docs/tooltip.html#tooltip-image does not because the base url doesn't seem to be set correctly for the tooltip.

domoritz:
I think the real issue here is that the tooltip plugin doesn't use the baseurl from the loader.

domoritz:
I think the best fix is to pass the baseurl through to tooltips. Done in https://github.com/vega/vega-embed/pull/1286

ChiaLingWeng:
Good to know where is the issue from, thanks!

domoritz:
Thank you for filing the bug and looking into fixing it. Seeing your fixes helped me pinpoint the bug much faster.

PR #9202, opened 2023-12-13
chore: make link in contributing docs clickable

domoritz:


keckelt (approved):
Thanks! I went through the md file and did not see any other _real_ links that need to be adapted.

PR #9203, opened 2023-12-13
chore: update deps

domoritz:
Remove `@babel/plugin-proposal-class-properties` since it's not a proposal anymore. Also update a vscode setting.


babel.config.js
```diff
@@ -10,6 +10,5 @@ module.exports = {
       }
     ],
     '@babel/preset-typescript'
-  ],
-  plugins: ['@babel/proposal-class-properties']
```
hydrosquall, line 14: From the NPM page: https://www.npmjs.com/package/@babel/plugin-proposal-class-properties

> This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-class-properties instead.

Upon inspection of the new site ( https://babeljs.io/docs/babel-plugin-transform-class-properties ). It appears this plugin is safe to drop because it's been moved to `@babel/preset-en` as long as you support ES2022 .

.vscode/settings.json
```diff
@@ -19,7 +19,7 @@
   "typescript.referencesCodeLens.enabled": true,
   "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
hydrosquall, line 22: This happened because of a VSCode change shift from 1.84 to 1.85, when boolean values for properties was deprecated. This change just shipped this month ( https://stackoverflow.com/a/77637765 ).

hydrosquall (approved):
This change contains 3 distinct adjustments

1. Updates VSCode settings to match the defaults of the latest version of VSCode
2. Removes the `@babel/proposal-class-properties` plugin
3. Updates a battery of devDependencies

I'm usually used to updating 1 of these things at a time in other projects, but these changes seem reasonable to me both individually and in aggregate . Leaning on CI to catch any surprise consequences of the package upgrades  ✅

PR #9207, opened 2023-12-14
refactor: simplify spec loading in docs with async

domoritz:


domoritz:
Ignore the pending checks. We don't need them for this change.

arvind (approved):
LGTM. One thought is whether we want to do any error checking for the `fetch` call? But, this question doesn't block the PR since I don't think we had any error checking for bad/404 URLs before either.

domoritz:
Added some error handling. 

Screenshot 2024-01-22 at 11 33 22

PR #9208, opened 2023-12-15
refactor: small code simplificiations

domoritz:


src/compile/selection/scales.ts
```diff
@@ -76,7 +76,7 @@ const scaleBindings: SelectionCompiler<'interval'> = {
     // Nested signals need only push to top-level signals with multiview displays.
     if (model.parent && !isTopLevelLayer(model)) {
       for (const proj of selCmpt.scales) {
-        const signal: any = signals.filter(s => s.name === proj.signals.data)[0];
+        const signal: any = signals.find(s => s.name === proj.signals.data);
```
lsh, line 79: Is this guaranteed to exist?

domoritz: Find returns undefined if there is no such item. The previous code would crash so I think the new one is strictly better, no?

lsh: Isn't the following line `signal.push = ...`? If `find` returns undefined what would we do? Otherwise why will it always be defined?

domoritz: Not sure but the previous behavior was to get an element from an empty array which also returns undefined so this should be an equivalent change.

lsh: Fair enough. I was just hoping we'd be able to also add some sort of comment to increase confidence in local reasoning.

lsh (approved):

PR #9212, opened 2023-12-21
fix: mark sequential scales as deprecated

### LINKED ISSUE #9017: Deprecate sequential scales
Sequential scales are deprecated (https://vega.github.io/vega/docs/scales/#sequential) so we should deprecate them here as well.

ChiaLingWeng:
close #9017 

Update in the continuous scale section,
let me know if need to add some warning in the script.
image


domoritz (changes requested):
Thanks for the pull request. I'd say that since we don't document the sequential scale anywhere in the Vega-Lite docs, we could instead just add a warning if someone uses a sequential scale so that we can delete the types in 6.x. 

We don't need to add a note to the docs.

domoritz (approved):
Thank you.

PR #9214, opened 2023-12-27
docs: link to valid bar chart example

luis261:
## replace link leading to an invalid specification error message in the online editor

I was doing some research on which lib to use for visualizations in one of my side projects.
Stumbled upon Vega/Vega-Lite via https://github.com/altair-viz/altair and really liked what I saw in terms of the approach it takes to visualization.

When reading up on Vega-Lite, I noticed that this link in your README leads to an error message in the online editor: https://vega.github.io/editor/#/custom/vega-lite

This can be confusing, especially on mobile. I suggest swapping in a link to a simple example instead.


domoritz:
Thanks. Good idea. The current link sets the mode of the editor but I agree that it's nicer to show a working example.

domoritz (approved):

luis261:
Awesome, thank you for your swift response!

PR #9234, opened 2024-01-29
feat: use consistent quotes that don't need to be escaped

### LINKED ISSUE #4112: Unify predicates
- Use spaces between operators.

I also propose to use ' instead of " so that we don't have to escape in JSON. Thoughts?

Ge-limin:
fixes #4112

### DIFF
--- a/src/predicate.ts
+++ b/src/predicate.ts
@@ -209,7 +209,8 @@ export function fieldFilterExpression(predicate: FieldPredicate, useInRange = tr
     : rawFieldExpr;
 
   if (isFieldEqualPredicate(predicate)) {
-    return `${fieldExpr} === ${predicateValueExpr(predicate.equal, unit)}`;
+    const equal = predicate.equal;
+    return `${fieldExpr} === ${predicateValueExpr(equal, unit)}`;
   } else if (isFieldLTPredicate(predicate)) {
     const upper = predicate.lt;
     return `${fieldExpr} < ${predicateValueExpr(upper, unit)}`;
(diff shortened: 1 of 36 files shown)


src/predicate.ts
```diff
@@ -209,23 +209,26 @@ export function fieldFilterExpression(predicate: FieldPredicate, useInRange = tr
     : rawFieldExpr;
 
   if (isFieldEqualPredicate(predicate)) {
-    return `${fieldExpr}===${predicateValueExpr(predicate.equal, unit)}`;
+    const equal = predicate.equal;
+    return `${fieldExpr} === ${predicateValueExpr(equal, unit)}`;
```
domoritz, line 213: Let's change spacing in a separate pull request.

src/predicate.ts
```diff
@@ -209,23 +209,26 @@ export function fieldFilterExpression(predicate: FieldPredicate, useInRange = tr
...
-    return `${fieldExpr}>=${predicateValueExpr(lower, unit)}`;
+    return `${fieldExpr} >= ${predicateValueExpr(lower, unit)}`;
   } else if (isFieldOneOfPredicate(predicate)) {
-    return `indexof([${predicateValuesExpr(predicate.oneOf, unit).join(',')}], ${fieldExpr}) !== -1`;
+    const oneOf = predicate.oneOf;
+    return `indexof([${predicateValuesExpr(oneOf, unit).join(',')}], ${fieldExpr}) !== -1`;
   } else if (isFieldValidPredicate(predicate)) {
-    return fieldValidPredicate(fieldExpr, predicate.valid);
+    const valid = predicate.valid;
+    return fieldValidPredicate(fieldExpr, valid);
```
domoritz, line 231: Not really a need to define a variable for one use.

src/util.ts
```diff
@@ -288,13 +288,17 @@ export function accessPathWithDatum(path: string, datum = 'datum') {
...
 export function flatAccessWithDatum(path: string, datum: 'datum' | 'parent' | 'datum.datum' = 'datum') {
-  return `${datum}[${stringValue(splitAccessPath(path).join('.'))}]`;
+  return `${datum}[${doubleQuotes2Single(stringValue(splitAccessPath(path).join('.')))}]`;
 }
 
 function escapePathAccess(string: string) {
   return string.replace(/(\[|\]|\.|'|")/g, '\\$1');
 }
 
+function doubleQuotes2Single(string: string) {
```
domoritz, line 298: No, we can't do this. It's too brittle and possibly introduces errors. If we want to change the quotes, we need to extend `stringValue` to accept a quote parameter.

domoritz:
Are you still working on this?

PR #9236, opened 2024-01-30
feat: switch to ESM

domoritz:
Switching to type module to support more modern js development. 

My hope is that this will help with the editor issues and let us update to newer versions of packages.


lsh (approved):
Amazing, thanks for the change!

domoritz:
We are now esm!

PR #9246, opened 2024-02-01
fix: layer with shared encoding ignore repeater

### LINKED ISSUE #8734: Repeat spec with shared encoding does not replace repeat
In 

```json
{
  "repeat": {"column": ["US Gross", "Worldwide Gross"]},
  "spec": {
    "encoding": {
      "x": {"bin": true, "field": "IMDB Rating", "type": "quantitative"},
      "y": {
        "aggregate": "mean",
        "type": "quantitative",
        "title": "Mean of US and Worldwide Gross"
      },
      "color": {"datum": {"repeat": "column"}, "type": "nominal"}
    },
    "layer": [
      {"name": "foo", "mark": "line"},
      {"name": "bar", "mark": "point"}
    ]
  }
}
```

the `{"repeat": "column"}` is not replaced in the normalized spec.

ChiaLingWeng:
close #8734 

I noticed that the main reason for the spec failed to replace `repeat` is due to there are some problem when handling the shared layer encoding, the params.repeater will be read when running mapUnit at the first time, but since in shared layer spec, the encoding will not be at the same layer as the repeater, and the repeater will not be record since mapUnitWithParentEncodingOrProjection will return config only, and the `repeat` is thus cannot be replaced.

I am not familiar with typescript so using some redundant code, I think this can be clean up and be more concised.


src/normalize/core.ts
```diff
@@ -225,14 +230,31 @@ export class CoreNormalizer extends SpecMapper<NormalizerParams, FacetedUnitSpec
...
 
-    return this.mapUnit(
-      {
-        ...spec,
-        ...(mergedProjection ? {projection: mergedProjection} : {}),
-        ...(mergedEncoding ? {encoding: mergedEncoding} : {})
-      },
-      {config}
-    );
+    if (this.isEmpty(encoding) && !this.isEmpty(params.repeater)) {
```
domoritz, line 233: Can we love this check to the second argument of the map unit method?

domoritz (changes requested):

src/normalize/core.ts
ChiaLingWeng: ah yes, simplify and updated!

src/normalize/core.ts
```diff
@@ -212,6 +211,13 @@ export class CoreNormalizer extends SpecMapper<NormalizerParams, FacetedUnitSpec
 
     return super.mapFacet(spec, params);
   }
+  //
+  private isEmpty(obj: any) {
```
domoritz, line 215: I think this should be a helper function outside of this class, no?

domoritz (approved):

src/normalize/core.ts
ChiaLingWeng: Yes, move this outside the class

src/normalize/core.ts
```diff
@@ -398,3 +405,10 @@ function mergeProjection<ES extends ExprRef | SignalRef>(opt: {
   }
   return projection ?? parentProjection;
 }
+
+function emptyFlag(obj: any) {
```
domoritz, line 409: I was thinking this should be it utils. We already have `isEmpty` there so you could use that.

src/normalize/core.ts
```diff
@@ -224,14 +224,21 @@ export class CoreNormalizer extends SpecMapper<NormalizerParams, FacetedUnitSpec
...
     });
-
     return this.mapUnit(
       {
         ...spec,
         ...(mergedProjection ? {projection: mergedProjection} : {}),
         ...(mergedEncoding ? {encoding: mergedEncoding} : {})
       },
-      {config}
+      emptyFlag(encoding) && !emptyFlag(params.repeater)
```
domoritz, line 233:
```suggestion
      isEmpty(encoding ?? {}) && ! isEmpty(params.repeater ?? {})
```

src/normalize/core.ts
domoritz: Suggested code above.

domoritz (changes requested):

src/normalize/core.ts
ChiaLingWeng: good to know that!

Issue #8230, opened 2022-06-19
Parameter usage in compound charts


mattijn:
_Goal_: Introduce parameters in Altair to get the Altair-repo ready for VL5.
This issue is a combination of the findings of [#7812](https://github.com/vega/vega-lite/issues/7812) and [#7854](https://github.com/vega/vega-lite/issues/7854) and https://github.com/vega/vega-lite/issues/5758#issuecomment-1148085715.

In compound charts, one can define parameters at the top-level or on the nested charts.

Currently it depend on the chart if one should define parameters at the top-level or define the parameter for each nested chart.

Is there any design philosophy that one can stick to when deciding where the parameters in compound charts should be placed?

- `strategy A`: _define parameters at top-level_.
- `strategy B`: _define parameters on nested chart_.

The following spec-examples can be summarized as follow
|           | `strategy A` | `strategy B` |
|-----------|------------|------------|
| example 1 | OK         | ERROR      |
| example 2 | ERROR      | OK         |

* * *

_Example 1_
Given the following example, a layered tick chart with a rule with a range slider as parameter

1. `strategy A` (define parameters at top-level) works: 
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUATJQAMlAFYQ2AOxAAaEDiQAnJAghoA2qGV64aEAGUAMgEkAIgFEASupBMaygCZpQnnAyYFrrKZOYayAAeaGIyEZ5ocSAQmHA4aACMAL4axEiCDOaokmJZALoagkgAnnDaRqBeSJgoqKAM2oIWTS0EUDoQCkqqOSDI2gDWFpg0UFMacMpQbF6eZL4g0W0gAGY0cII+6ACyNIJwEAD6OHWXAOL5gipumNU3FgCODEjKMy0zpBAWWBakazVaoDyBQuDXKo3GU22r3e6G0DHObggNAAXkUAMyjRbLVahDZbMGYBgIDZwSI4eroGwOFxAjTIoogL4-P7NGiA4FwoA)
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.2.0.json",
  "params": [
    {
      "name": "SLIDER",
      "bind": {"input": "range", "max": 50, "min": 0, "step": 1},
      "value": 25
    }
  ],
[...]

PR #9258, opened 2024-02-08
chore: update deps and rebuild

domoritz:


keckelt:
How would I best test this? I re-built the examples with `yarn clean` and `yarn build:examples` on both the main branch and the PR branch to verify how the changes effect the produced chart. But they look the same on my machine:

`main`:
line_dynamic_axis-main

`dom/rebuild-6782`:
line_dynamic_axis-pr

domoritz:
The ci should have taken care of examples that are in the repo already. The question is whether the params should be in the first layer rather than at the outer spec level. The code looked intentional so I wanted to have someone else check.

domoritz:
I think a good check is whether a deeply nested spec breaks with this change.

mattijn:
The intention of this PR is not exactly clear, but is the following issue + comment related here? https://github.com/vega/vega-lite/issues/8230#issuecomment-1179213186

Also, when is this code invoked? I think this param lifting could be done in the section creating the extended vega-lite code.

PR #9262, opened 2024-02-15
fix: support arrays in color schemes

### LINKED ISSUE #9022: Support array of values for scheme
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtCrVJOcNZkFx+AdxP042ylAjEQAXyXJtAa31Mk2xSBwslBsyjSyZGigAB5RIJg0CHAAqrJ0+ghy9AEAZjRwgsr6qphwAZgAnjhl6GzaYbJIekoJZjUgALJZMAAEbDk9tj0VJv5uIBVxSGRk2nyl+iEMspjlVe0AjgxIK3RqNKSuSiGCdXF5BUXo1mrw-i3r+rJsCOFNAdBNNaBciDUA2iAAMQARgAHAA2AAMUCgAVBYOUACZlABOeHg2EAZghIAAuuNWuZ9AB1Gx3QaPFzUoA)

This already works but just isn't supported by the types. 

See https://github.com/vega/vega-lite/pull/9015#issuecomment-1643380017

domoritz:
fixes #9022

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -21929,6 +21929,12 @@
               "$ref": "#/definitions/ColorScheme"
             },
             {
+              "items": {
+                "type": "string"
+              },
+              "type": "array"
+            },
+            {
               "$ref": "#/definitions/SchemeParams"
             },
             {
--- a/src/scale.ts
+++ b/src/scale.ts
@@ -481,7 +481,7 @@ export type Domain =
 
 export type Scheme = string | SchemeParams;
 
-export function isExtendedScheme(scheme: Scheme | SignalRef): scheme is SchemeParams {
+export function isExtendedScheme(scheme: Scheme | SignalRef | string[]): scheme is SchemeParams {
   return !isString(scheme) && !!scheme['name'];
 }
 
@@ -611,7 +611,7 @@ export interface Scale<ES extends ExprRef | SignalRef = ExprRef | SignalRef> {
    *
    * For the full list of supported schemes, please refer to the [Vega Scheme](https://vega.github.io/vega/docs/schemes/#reference) reference.
    */
-  scheme?: ColorScheme | SchemeParams | ES;
+  scheme?: ColorScheme | string[] | SchemeParams | ES;
 
   /**
    * The alignment of the steps within the scale range.
--- a/src/compile/scale/range.ts
+++ b/src/compile/scale/range.ts
@@ -211,7 +211,7 @@ export function parseRangeForChannel(channel: ScaleChannel, model: UnitModel): E
   return makeImplicit(d);
 }
 
-function parseScheme(scheme: Scheme | SignalRef): RangeScheme {
+function parseScheme(scheme: Scheme | SignalRef | string[]): RangeScheme {
   if (isExtendedScheme(scheme)) {
     return {
       scheme: scheme.name,


src/scale.ts
```diff
@@ -611,7 +611,7 @@ export interface Scale<ES extends ExprRef | SignalRef = ExprRef | SignalRef> {
    *
```
hydrosquall, line 611: Optional: we could also advertise this capability somewhere in the docs or in the docstring? Currently it's not mentioned

https://vega.github.io/vega-lite/docs/scale.html#2-setting-the-range-property-to-an-array-of-valid-css-color-strings seems close but it targets `scale.range` rather than Scale.scheme

src/scale.ts
```diff
@@ -611,7 +611,7 @@ export interface Scale<ES extends ExprRef | SignalRef = ExprRef | SignalRef> {
...
    * For the full list of supported schemes, please refer to the [Vega Scheme](https://vega.github.io/vega/docs/schemes/#reference) reference.
```
hydrosquall, line 612: Should we advertise this capability in the docs and/or a docstring? 

https://vega.github.io/vega-lite/docs/scale.html#2-setting-the-range-property-to-an-array-of-valid-css-color-strings seems close but it targets `scale.range` rather advertising allowing `scheme` to be immediately set as an array. I'm assuming this arrays  is a faster shorthand for the same behavior.

build/vega-lite-schema.json
```diff
@@ -21928,6 +21928,12 @@
             {
               "$ref": "#/definitions/ColorScheme"
             },
+            {
```
hydrosquall, line 21931: Confirmed that this matches what we see in the underlying `Vega` types for RangeScheme 👍 

https://github.com/vega/vega/blob/b8353d57e4842c88b81fbcfbfd7ccf48b867ae2b/packages/vega-typings/types/spec/scale.d.ts#L21

hydrosquall (approved):
Thanks for making the change, I noticed this issue before but didn't know if perhaps this shorthand for setting scale.range wasn't meant to be supported. I didn't run this locally but it looks like this should work. 

For future PRs - what does the workflow look like for testing changes like this? One thought is that launching a preview build of the Vega editor using this branch's version of the library could help with QA so we can validate that specs like [this](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtCrVJOcNZkFx+AdxP042ylAjEQAXyXJtAa31Mk2xSBwslBsyjSyZGigAB5RIJg0CHAAqrJ0+ghy9AEAZjRwgsr6qphwAZgAnjhl6GzaYbJIekoJZjUgALJZMAAEbDk9tj0VJv5uIBVxSGRk2nyl+iEMspjlVe0AjgxIK3RqNKSuSiGCdXF5BUXo1mrw-i3r+rJsCOFNAdBNNaBciDUA2iAAMQARgAHAA2AAMUCgAVBYOUACZlABOeHg2EAZghIAAuuNWuZ9AB1Gx3QaPFzUoA) are schema checked happily (e.g. no yellow squiggle under the array in the JSON.).

image

domoritz:
> For future PRs - what does the workflow look like for testing changes like this? One thought is that launching a preview build of the Vega editor.

Right now, local checkout is needed. A preview of the editor would be awesome (similarly a preview of the website).

src/scale.ts
domoritz: Good point. I think we should revert this change since there ideally aren't multiple ways to do the same thing.

domoritz: I sent https://github.com/vega/vega-lite/pull/9266

hydrosquall:
> local checkout is needed. A preview of the editor would be awesome

I've made an issue for looking into this, I'll see if we can do this without any 3rd party providers: https://github.com/vega/vega-lite/issues/9276

Issue #9276, opened 2024-03-03
developer-experience: deployment previews for pull requests


hydrosquall:
## Enhancement Description

Goal: Reduce time and effort to validate change requests, and improve review quality for vega-lite change requests (PRs). By making a deployment preview easy, it makes it easier to test end-user facing behaviors.

Motivating use-cases

- When new PRs are opened, enable maintainers to review and test edge cases
- Create a test environment that shields
- See discussion w/ @domoritz here: https://github.com/vega/vega-lite/pull/9262#issuecomment-1950624437

Expected behavior

- When a new PR is opened, build a copy of the vega-lite docs site and/or vega-editor that uses this branch's copy of vega-lite.
  - TBD which is easier to do first.

## Checklist

- [x] I checked for duplicate issues. ( [preview](https://github.com/vega/vega-lite/issues?q=is%3Aissue+preview+) )

## Notes

- It's possible to set this up with a 3rd party provider (each with their own pricing model), but I'd like to first try if we can do this purely with Github actions.

hydrosquall:
Here's a copy of the setup on Cloudflare so anyone can configure this for their own account.

 For now this only builds the `vega/editor`, not the docs .

![image](https://github.com/vega/vega-lite/assets/9020979/fb07a6a9-59d3-401f-b0da-cdec1363d5b1)

- Disable automatic deployments on `main`
- Blocklist for branches: `dependabot/*` , `docs/*`
- Keeplist for builds: the `src/*` directory
- Build settings
  - run `yarn run build-editor-preview`
  - Output the result to `/editor/dist`

See the result here: https://vega-lite.pages.dev/#/

domoritz:
Sweet. I think we can close this now, right?

hydrosquall:
Yep, unless we also want previews of the docs site to be deployed in
parallel.

domoritz:
That's not as critical as the editor imo but could also be useful. How about we close this for now until we learn a bit more about how the current setup works out?

hydrosquall:
Sounds good!

hydrosquall:
Idea for future : if building previews per commit automatically  eats through the builds too quickly much, we can use the "Deploy hooks" feature to trigger deploys via specialized comments (like `/request-deploy`), or having a microsite where people can request builds by entering the PR #.

hydrosquall:
Noticed today that the current solution works for people in the Vega org. This is good for saving my CI minutes, but less good for enabling new contributors

[...]

PR #9277, opened 2024-03-03
build: add deployment previews for vega-lite documentation site

hydrosquall:
## Motivation

- A deployment preview is a hosted version of a static site created just with the contents of this branch
- If this works out, we could add a similar change to deploy the editor site down the line.
- https://github.com/vega/vega-lite/issues/9276

## Changes

> TK

Try configuring https://github.com/rossjrw/pr-preview-action

## Testing

- The comment on this PR should link to a version of the docs site that outputs a console.log every time the vega-lite `compile` function is called.

## Notes

- As far as being a part of native Github, this feature is currently only available to Github pages for GH employees ( https://github.com/orgs/community/discussions/7730 )
  - https://github.com/actions/deploy-pages/issues/180#issuecomment-1921081514

### DIFF
--- a/.github/workflows/deployment-preview.yml
+++ b/.github/workflows/deployment-preview.yml
@@ -1,4 +1,4 @@
-# .github/workflows/preview.yml
+# .github/workflows/deployment-preview.yml
 name: Deploy PR previews
 
 on:
@@ -8,6 +8,7 @@ on:
       - reopened
       - synchronize
       - closed
+      # TODO: should this be rerun on new commits?
 
 concurrency: preview-${{ github.ref }}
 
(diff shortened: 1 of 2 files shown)


hydrosquall:
It looks like this action doesn't work, perhaps as we deploy the docs site from "github actions" instead of from a fixed "github pages" branch?

image

I'm going to leave this alone until we have a chance to discuss, as we can avoid needing to configure a custom action if we're willing to use a 3rd party provider (StackBlitz, cloudflare pages, vercel, netlify, etc) to handle this for us.

hydrosquall:
After looking at the [size](https://github.com/vega/vega-lite/commit/7fb50d0e77d5caf6ac73bd49bc71096e8c2e832c#diff-41b03121fbe3203074f2a65487e9fe2fab9619a88837a6bb1476182062bb5362) of the pushes to gh-pages, I don't think deploying the preview site based on GH pages is a good idea, even if it were possible to do so inside this repo. Each new PR would generate a diff millions of lines long due to all the compiled JSON examples. Unless our maintainers and contributors remember to not download the gh-pages branch, this is a lot of unnecessary bloat in the github repository. I think this is part of why Github is trying to deprecate the branch-based deployment model, and keep the deployment artifacts _outside_ of Git.

I'm inclined to recommend that we do this _without_  native github actions, and instead choose one of the providers that offer deployment preview for free to open source projects. Let me know what you think @domoritz , and we can go from there.

nicolaskruchten:
The way I've solved this in the past is not to commit to `gh-pages` as a child commit of the existing one, but rather as a whole new parent-less commit. This way there are no diffs, just a bunch of orphaned commits that eventually go away or at least don't really bother anyone.

hydrosquall:
@nicolaskruchten thanks for taking a look. Do you have a link to a public repository where you had set this up?

I'm not sure that the GH pages route for branch deploys will work inside this repo, but maybe what you're describing is something we could try in a repo dedicated to _just_ hosting PR previews, e.g. https://github.com/vega/vega-lite-pr-previews

nicolaskruchten:
Here you go https://github.com/plotly/plotly.py/blob/master/.circleci/config.yml#L558-L567 ... I was surprised that this worked but it does! The target for this is here https://github.com/plotly/plotly.py-docs/tree/gh-pages (which renders at https://plotly.com/python-api-reference/) and it's always a single commit with no parent. Presumably the other commits are piling up somewhere in Github's infra and getting garbage-collected? I'm not sure but no one has ever complained to my knowledge :)

nicolaskruchten:
To get deployment previews per-PR is probably a lot trickier than this, though. Incredibly valuable but trickier :) We've done this in the past with Heroku's non-free features, and maybe there's a workflow with a provider with hobby plans? Fly.io doesn't collect on monthly bills under $5 so maybe that's an option since these deployment previews will get very little traffic?

domoritz:
Thanks for making a POC for this. I think hosting previews in a separate repo sounds great. Alternatively, I've seen vercel/netflify deploy PRs and that could be an alternative. That's what https://github.com/vega/react-vega does for example (@kristw set that up). I don't care either way as long as we don't pollute the main repo.

.github/workflows/deployment-preview.yml
```diff
@@ -0,0 +1,80 @@
...
+name: Deploy PR previews
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
+      # TODO: should this be rerun on new commits?
```
domoritz, line 11: Yes

.github/workflows/deployment-preview.yml
```diff
@@ -0,0 +1,80 @@
...
+
+concurrency: preview-${{ github.ref }}
+
+jobs:
+  # TODO: do we still want to run these on dependabot/non-human branches?
```
domoritz, line 16: No. That would be too much noise. I've never had to check dependabot PRs.

.github/workflows/deployment-preview.yml
```diff
@@ -0,0 +1,80 @@
...
+  deploy-preview-docs:
+    if: github.event.action != 'closed'
+    runs-on: ubuntu-20.04
```
domoritz, line 19: Why not latest?

hydrosquall: I was trying to avoid surprise breakages by pinning to a specific version.

hydrosquall:
Replaced by https://github.com/vega/vega-lite/pull/9294

.github/workflows/deployment-preview.yml
domoritz: I'm not worried about os updates.

PR #9284, opened 2024-03-15
fix: generate unescaped boxplot calculated field aliases

n8agrin:
## PR Description

This change ensures all field aliases are unescaped when generated. Before this change if a field was referenced like `"field": "one\\.two"` the boxplot code would generate calculated fields like `"as": "lower_box_one\\.two"`. This breaks downstream in the compiled Vega code which expects aliases to not include escape sequences.

As far as I could tell there is no issue open for this. The contributing.md doc explicitly calls out it is not necessary to open one, but I'm happy to do so if it helps.

### DIFF
--- a/src/compositemark/boxplot.ts
+++ b/src/compositemark/boxplot.ts
@@ -8,7 +8,7 @@ import {isMarkDef, MarkDef, MarkInvalidMixins} from '../mark';
 import {NormalizerParams} from '../normalize';
 import {GenericUnitSpec, NormalizedLayerSpec, NormalizedUnitSpec} from '../spec';
 import {AggregatedFieldDef, CalculateTransform, JoinAggregateTransform, Transform} from '../transform';
-import {isEmpty, omit} from '../util';
+import {isEmpty, omit, removePathFromField} from '../util';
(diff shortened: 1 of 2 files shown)


n8agrin:
@domoritz thanks for the quick 👍. Do you know if this is expected to ever resolve? 
image

domoritz:
It won't automatically. You either need to figure out how to run the particular job on your fork (we can't do that but I think it may work if you set the personal access token as described in https://github.com/vega/vega-lite/blob/main/CONTRIBUTING.md#contributing but I haven't been able to test this in a while). Otherwise I need to make a branch for you in the repo so we can run the checks.

n8agrin:
@domoritz got it thanks! Looks like everything has resolve. Appreciate your help!

domoritz:
I made a clone in https://github.com/vega/vega-lite/pull/9285

domoritz:
Yeah, once I made a clone, the CI runs and shows up. I'd really like to figure out how we can better support PRs from external contributors.

n8agrin:
@domoritz sounds good. I'm fine with either merging, I'll leave it up to you.

domoritz:
Thank you

PR #9305, opened 2024-04-11
fix: for interactive charts, changed cursor to pointer (also updated tests)

### LINKED ISSUE #4155: Interactive views should use cursors to provide affordance

JessamineQ:
## PR Description

Fixes #4155 (specifically for cursor over interactive bars)

### DIFF
--- a/src/compile/mark/mark.ts
+++ b/src/compile/mark/mark.ts
@@ -313,7 +313,21 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
   const key = encoding.key;
   const sort = getSort(model);
   const interactive = interactiveFlag(model);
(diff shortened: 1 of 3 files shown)


domoritz (changes requested):
I don't think this is quite right yet. For example in http://localhost:5173/editor/#/examples/vega-lite/dynamic_color_legend we show a pointer cursor but the selection is an interval selection with a brush. For brushes, the cursor should change for the whole chart (where you can draw the brush) and not change just because you hover over a mark. Think of it this way: an interval selection isn't a selection of a mark so the cursor should not be specific to the mark.

domoritz (changes requested):

domoritz:
Replaced by https://github.com/vega/vega-lite/pull/9358

PR #9328, opened 2024-05-01
feat: for interactive charts, changed cursor to pointer (also updated tests)

### LINKED ISSUE #4155: Interactive views should use cursors to provide affordance

domoritz:
Closes #4155

Updated version of https://github.com/vega/vega-lite/pull/9305.


domoritz:
new version in https://github.com/vega/vega-lite/pull/9358

Issue #9326, opened 2024-04-28
Line with invalid null broken 


kanitw:
## Bug Description

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "data": {
    "values": [
      {
        "x": -1,
        "y": null
      },
      {
        "x": 1,
        "y": 10
      },
      {
        "x": 2,
        "y": 30
      },
      {
        "x": 3,
        "y": null
      },
      {
        "x": 4,
        "y": 15
      },
      {
        "x": 5,
        "y": 30
      },
      { 
        "x": 6,
        "y": 40
      },
      {
        "x": 7,
        "y": 20
      },
      {
        "x": 10,
        "y": null
      }
    ]
  },
  "mark": {"type": "line", "invalid": null},
  "encoding": {
    "x": {"field": "x", "type": "quantitative"},
    "y": {"field": "y", "type": "quantitative"}
  }
}

```

<img width="274" alt="image" src="https://github.com/vega/vega-lite/assets/111269/c3c841eb-e1ad-45ce-8939-e388a6751dab">

All null values are interpreted as max values!!

PR #9334, opened 2024-05-02
fix: correctly place unfiltered null values for line (#9326)

kanitw:
## PR Description

Previously we forgot to add conditional logic for invalid values when filtering is disabled (mark.invalid = null)

Interestingly, given area applies stack layout by default, it doesn't suffer from the same bug that line does.


lsh (approved):
LGTM since given new examples and lack of diffs generated.

kanitw:
>given new examples and lack of diffs generated. 

to be clear, the new example shows the right fix: 

image

PR #9339, opened 2024-05-03
test: add example specs as regression test for invalid values

kanitw:
Just add more specs examples


lsh (approved):
Only adding examples, so no concerns

Issue #6417, opened 2020-04-27
Format of null data in tooltips.


PhML:
In the following example, I use a tooltip to display underlying data but I have troubles to have consistent behavior when displaying null values of different types.

Indeed, while missing quantitative data leads to display `NaN` (no `c` key in third value of values), temporal data is interpreted as zero (`Jan 01, 1970 00:00:00`). I expect to display null values for temporal data as `null` or `NaT` to be consistent with `NaN` display of quantitative data.

When values are explicitly `null`, behavior is consistent as both quantitative and temporal data are rendered as respectively `0` and `Jan 01, 1970 00:00:00`.

*But I rather like to display `null` or `N/A` in such case. So having a way to specify **how** to handle null values.* (Please, tell me how if it is possible).

```json
{
  "data": {
    "values": [
      {
        "a": "A",
        "b": 28,
        "c": null,
        "d": null
      },
      {
        "a": "B",
        "b": 55,
        "c": "2020-04-27T14:15:18.586949+00:00",
        "d": 1000
      },
      {
        "a": "C",
        "b": 43
      }
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "a",
      "type": "ordinal"
    },
    "y": {
      "field": "b",
      "type": "quantitative"
    },
    "tooltip": [
      {
        "field": "a",
        "type": "ordinal"
      },
      {
        "field": "b",
        "type": "quantitative"
      },
      {
        "field": "c",
        "type": "temporal",
        "title": "c",
        "timeUnit": "yearmonthdatehoursminutesseconds",
        "scale": {
          "type": "utc"
        }
      },
      {
        "field": "d",
        "type": "quantitative"
      }
    ]
  }
}

```
[...]

PR #9342, opened 2024-05-07
feat: more options for how marks and scales represent invalid data (e.g., nulls, NaNs)

### LINKED ISSUE #8254: mark.invalid set the invalid points on y channel at the top
[Here](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygG4QDYFcCmBneBtUAxgPYpEBO8IAQuhiADQgCe8AjAAwC+9hJ5lAcTIYMAOwbN4otChTdepCnBAAlDGAks4AJi4BdbiDHEwAS1EBzeKAAe1kADNTGFBuXFFEqEwAOdZaJEALbmqF6mUCj+IB78hlqgTi5ukozefpQAjmgQolAR0KaIdJyGQRBkANb26dEo5nRpRCT5PvBQZJiGxKJOVggg5VX25sj1btKypZxAA) is the repro.
```json
{
  "data": {
    "values": [
      {"color": "Blue", "y": 10},
      {"color": "Green", "y": null},
      {"color": "Red", "y": 20}
    ]
  },
  "encoding": {
    "x": {"field": "color", "type": "nominal", "title": "color"},
    "y": {"field": "y", "type": "quantitative"}
  },
  "mark": {"type": "line", "tooltip": true},
  "config": {"mark": {"invalid": null}}
}
```
image

It is supposed to place "Green" value at the bottom (the 0 value position (yScale(0)) not pure 0 position)

kanitw:
- Add more spec options for handling invalid data 
  - extend `mark.invalid`
  - add new `config.scale.invalid`

- See [docs's pdf file from invalid.md](https://github.com/vega/vega-lite/files/15329038/Modes.for.Handling.Invalid.Data._.Vega-Lite.pdf) or the actual doc for explanation of the behavior.  (Given the complexity of the PR, reviewers should check out the branch and test the spec locally.) 

refactor:
- Adjust how we generate rules for invalid data to handle these options, per channel type. 
    - data's invalid filter
    - `defined` encoding (for breaking paths)

- Make wrapCondition accept invalidValueRef for "include" mode.
- Remove the hidden/screen "hide" mode for invalid data (Given this is a hidden feature, it's not a breaking change.)## PR Description

bug fixes:
- fix #8254  -- correctly place invalid point as zero, when included. 

Known issues / Follow up items: 
- Tooltip doesn't show null correctly (https://github.com/vega/vega-lite/issues/6417) -- I'll fix this separately in a follow up PR
- Should Vega-lite always include color for invalid by default? (https://github.com/vega/vega-lite/issues/9351)

## Checklist

Tests:
- [x] Integration tests -- added a bunch of examples
- [x] Unit tests  (in progress) -- I'll re-review the code and add more unit tests. 

Docs: 


src/compile/data/filterinvalid.ts
```diff
@@ -20,27 +20,40 @@ export class FilterInvalidNode extends DataFlowNode {
...
+            markDef,
+            config,
+            scaleType,
+            isCountAggregate: isCountingAggregateOp(aggregate)
+          });
 
-          // While discrete domain scales can handle invalid values, continuous scales can't.
-          // Thus, for non-path marks, we have to filter null for scales with continuous domains.
-          // (For path marks, we will use "defined" property and skip these values instead.)
-          if (hasContinuousDomain(scaleType) && fieldDef.aggregate !== 'count' && !isPathMark(mark)) {
```
kanitw, line 40: These rules are now part of `getScaleInvalidDataMode`

src/compile/data/parse.ts
```diff
@@ -384,13 +399,41 @@ export function parseData(model: Model): DataComponent {
     head = StackNode.makeFromEncoding(head, model) ?? head;
   }
 
+  let preFilterInvalid: OutputNode | undefined;
```
kanitw, line 402: **TODO: file a separate ticket** 

Separate from this PR, I have a hunch that this will remain slightly incorrect for stacked chart, since null values will be "stacked", but it's probablly

src/compile/data/parse.ts
kanitw, line 1: FYI, in this file, we're creating data sources pre-/post-filter invalid values, depending on the invalid data mode.

src/compile/guide.ts
```diff
@@ -2,15 +2,20 @@ import {GuideEncodingEntry} from '../guide';
...
+import {wrapCondition} from './mark/encode/conditional';
 import {UnitModel} from './unit';
 
 export function guideEncodeEntry(encoding: GuideEncodingEntry, model: UnitModel) {
   return keys(encoding).reduce((encode, channel: VgEncodeChannel) => {
-    const valueDef = encoding[channel];
     return {
       ...encode,
-      ...wrapCondition(model, valueDef, channel, def => signalOrValueRef(def.value))
+      ...wrapCondition({
```
kanitw, line 12: FYI, no behavior change here. I just switch `wrapCondition` to use parameter argument and make it explicit that `guideEncodeEntry` won't need the new `invalidValueRef` added.

src/compile/mark/encode/aria.ts
```diff
@@ -45,7 +45,13 @@ export function description(model: UnitModel) {
   const channelDef = encoding.description;
 
   if (channelDef) {
-    return wrapCondition(model, channelDef, 'description', cDef => textRef(cDef, model.config));
+    return wrapCondition({
```
kanitw, line 48: FYI, no behavior change here. I just switch `wrapCondition` to use parameter argument and make it explicit that `aria` won't need the new `invalidValueRef` added.

src/compile/mark/encode/base.ts
kanitw, line 1: FYI, all old code in this file were for the legacy **secret/hidden** "hide" mode that I remove. 

The old hide mode hides data by adjusting fill/stroke, which still wastes rendering time to draw transparent marks. 
With the new `break-paths-keep-domain`, we're switching to rely on filters to produce this behavior.

src/compile/mark/encode/conditional.ts
kanitw, line 1: FYI, the main change in this file is to add `invalidValueRef` that caller of this method may pass in. 

However, I also switch to object argument so it's easy to make explicit when `invalidValueRef` is not used.

src/compile/mark/encode/defined.ts
```diff
@@ -1,49 +1,48 @@
...
+      markDef,
+      config,
+      scaleType,
+      isCountAggregate
+    });
+    if (shouldBreakPath(invalidDataMode)) {
       const field = model.vgField(channel, {expr: 'datum', binSuffix: model.stack?.impute ? 'mid' : undefined});
-
-      // While discrete domain scales can handle invalid values, continuous scales can't.
-      if (field && hasContinuousDomain(scaleType)) {
```
kanitw, line 36: All the continuous scale check is now in `getScaleInvalidDataMode`

src/compile/mark/encode/nonposition.ts
```diff
@@ -33,17 +34,34 @@ export function nonPosition(
...
   const channelDef = encoding[channel];
+  const commonProps = {
+    markDef,
+    config,
+    scaleName: model.scaleName(channel),
+    scale: model.getScaleComponent(channel)
+  };
 
-  return wrapCondition(model, channelDef, vgChannel ?? channel, cDef => {
+  const invalidValueRef = getConditionalValueRefForIncludingInvalidValue({
```
kanitw, line 44: FYI, the main change in this file is adding `invalidValueRef` here.

src/compile/model.ts
```diff
@@ -635,6 +635,11 @@ export abstract class Model {
     return this.parent ? this.parent.getScaleComponent(channel) : undefined;
   }
 
+  public getScaleType(channel: ScaleChannel): ScaleType {
```
kanitw, line 638: FYI, a convenience method.

src/compile/mark/encode/zindex.ts
```diff
@@ -9,7 +9,13 @@ export function zindex(model: UnitModel) {
   const order = encoding.order;
 
   if (!isPathMark(mark) && isValueDef(order)) {
-    return wrapCondition(model, order, 'zindex', cd => signalOrValueRef(cd.value));
+    return wrapCondition({
+      model,
+      channelDef: order,
+      vgChannel: 'zindex',
+      mainRefFn: cd => signalOrValueRef(cd.value),
+      invalidValueRef: undefined // zindex encoding doesn't have continuous scales and thus can't have invalid values
```
kanitw, line 17: FYI, no behavior change here. I just switch `wrapCondition` to use parameter argument and make it explicit that `zindex` won't need the new `invalidValueRef` added.

src/compile/scale/component.ts
```diff
@@ -34,6 +34,13 @@ export class ScaleComponent extends Split<ScaleComponentProps> {
    * Whether the scale definitely includes zero in the domain
    */
   public domainDefinitelyIncludesZero() {
+    if (contains([ScaleType.LOG, ScaleType.TIME, ScaleType.UTC], this.get('type'))) {
```
kanitw, line 37: A good rule that I discovered as I read relevant code.

src/compile/scale/domain.ts
```diff
@@ -249,6 +251,11 @@ function parseSingleChannelDomain(
   const {type} = fieldOrDatumDef;
   const timeUnit = fieldOrDatumDef['timeUnit'];
 
+  const dataSourceTypeForScaleDomain = getScaleDataSourceForHandlingInvalidValues({
```
kanitw, line 254: FYI, this is the main change in this file. Instead of always drawing scale domains from "main" data source, we may choose pre-/post- filtering invalid values, depending on the invalid data mode.

src/mark.ts
```diff
@@ -72,18 +77,6 @@ export interface TooltipContent {
   content: 'encoding' | 'data';
 }
 
-/** @hidden */
-export type Hide = 'hide';
-
-export interface MarkInvalidMixins {
```
kanitw, line 78: FYI, moved to `src/invalid.ts` + revamped

test/compile/mark/text.test.ts
```diff
@@ -242,16 +242,10 @@ describe('Mark: Text', () => {
...
-        {
-          test: '!isValid(datum["mean_Acceleration"]) || !isFinite(+datum["mean_Acceleration"])',
-          value: null
-        },
-        {
-          scale: 'color',
-          field: 'mean_Acceleration'
-        }
-      ]);
+      expect(props.fill).toEqual({
```
kanitw, line 245: Intended change -- we no longer relying "null" fill to hide invalid marks in the new `"break-paths-keep-domains"` that replaces in the old secret "hide" mode.

src/mark.ts
```diff
@@ -72,18 +77,6 @@ export interface TooltipContent {
   content: 'encoding' | 'data';
 }
 
-/** @hidden */
-export type Hide = 'hide';
```
kanitw, line 76: FYI, the old hide mode was a secret experimental feature (not part of public API) that I experimented with many years ago.  It's not documented and thus can be removed.

test/compile/selection/layers.test.ts
```diff
@@ -85,16 +85,10 @@ describe('Layered Selections', () => {
...
-              {
-                test: '!isValid(datum["Horsepower"]) || !isFinite(+datum["Horsepower"]) || !isValid(datum["Miles_per_Gallon"]) || !isFinite(+datum["Miles_per_Gallon"])',
-                value: null
-              },
-              {
-                scale: 'color',
-                field: 'Origin'
-              }
-            ],
+            fill: {
```
kanitw, line 88: FYI, intended change -- we no longer relying "null" fill to hide invalid marks in the new `"break-paths-keep-domains"` that replaces in the old secret "hide" mode. 

Same for the rest of this file.

src/compile/invalid/datasources.ts
kanitw, line 1: This is one of the main rules for defining how these modes affect data sources for scale / marks.

src/compile/mark/encode/conditional.ts
```diff
@@ -7,21 +7,37 @@ import {parseSelectionPredicate} from '../../selection/parse';
...
-  model: UnitModel,
-  channelDef: CD,
-  vgChannel: string,
-  refFn: (cDef: CD) => VgValueRef
-): VgEncodeEntry {
+export function wrapCondition<CD extends ChannelDef | GuideEncodingConditionalValueDef>({
+  model,
+  channelDef,
+  vgChannel,
+  invalidValueRef,
```
kanitw, line 19: FYI, the main change in this file is to add invalidValueRef that caller of this method may pass in.

However, I also switch to object argument so it's easy to make explicit when invalidValueRef is not used.

src/compile/invalid/datasources.ts
```diff
@@ -0,0 +1,63 @@
...
+export interface DataSourcesForHandlingInvalidValues {
+  marks: PreOrPostFilteringInvalidValues;
+  scales: PreOrPostFilteringInvalidValues;
+}
+interface GetDataSourcesForHandlingInvalidValuesProps {
+  invalid: MarkInvalidDataMode | null | undefined;
+  isPath: boolean;
+}
+
+export function getDataSourcesForHandlingInvalidValues({
```
kanitw, line 17: This method includes one of the main rules for defining how these modes affect data sources for scale / marks.

src/compile/mark/encode/base.ts
```diff
@@ -45,27 +39,7 @@ export function baseEncodeEntry(model: UnitModel, ignore: Ignore) {
   };
 }
 
-// TODO: mark VgValueRef[] as readonly after https://github.com/vega/vega/pull/1987
-function wrapAllFieldsInvalid(model: UnitModel, channel: Channel, valueRef: VgValueRef | VgValueRef[]): VgEncodeEntry {
```
kanitw, line 49: FYI, all old code in this file were for the legacy **secret/hidden** "hide" mode that I remove. 

The old hide mode hides data by adjusting fill/stroke, which still wastes rendering time to draw transparent marks. 
With the new `break-paths-keep-domain`, we're switching to rely on filters to produce this behavior.

examples/compiled/bar_distinct.vg.json
```diff
@@ -16,10 +16,6 @@
           "ops": ["distinct"],
           "fields": ["Name"],
           "as": ["distinct_Name"]
-        },
-        {
```
kanitw, line 20: FHI, correct change since `distinct` aggregate can't output `null`s.

examples/compiled/bar_grouped_repeated.vg.json
```diff
@@ -52,7 +52,8 @@
           "y": [
             {
               "test": "!isValid(datum[\"sum_Worldwide Gross\"]) || !isFinite(+datum[\"sum_Worldwide Gross\"])",
-              "field": {"group": "height"}
+              "scale": "y",
```
kanitw, line 55: FYI, Good change as intended in a later fix commit -- invalid values get placed as zero rather than min.

examples/compiled/point_invalid_color.vg.json
```diff
@@ -32,14 +32,16 @@
           "x": [
             {
               "test": "!isValid(datum[\"IMDB Rating\"]) || !isFinite(+datum[\"IMDB Rating\"])",
+              "scale": "x",
```
kanitw, line 35: FYI, good change as intended in a later fix commit -- invalid values get placed as zero rather than min. 

- value: 0 = min 
- scale: 'x', value: '0' -> encode same output as 0

src/channel.ts
```diff
@@ -480,7 +480,7 @@ const SCALE_CHANNEL_INDEX = {
 export const SCALE_CHANNELS = keys(SCALE_CHANNEL_INDEX);
 export type ScaleChannel = (typeof SCALE_CHANNELS)[number];
 
-export function isScaleChannel(channel: Channel): channel is ScaleChannel {
+export function isScaleChannel(channel: Channel | ExtendedChannel): channel is ScaleChannel {
```
richardliu-db, line 483: `ExtendedChannel` already covers `Channel` (`ExtendedChannel = Channel | FacetChannel;`). Just `channel: ExtendedChannel`?

src/compile/data/filterinvalid.ts
```diff
@@ -20,27 +20,40 @@ export class FilterInvalidNode extends DataFlowNode {
...
+    parent: DataFlowNode,
+    model: UnitModel,
+    dataSourcesForHandlingInvalidValues: DataSourcesForHandlingInvalidValues
+  ): FilterInvalidNode {
+    const {config, markDef} = model;
 
-    const invalid = getMarkPropOrConfig('invalid', markDef, config);
-    if (invalid !== 'filter') {
+    const {marks, scales} = dataSourcesForHandlingInvalidValues;
+    if (marks === 'pre-filter' && scales === 'pre-filter') {
```
richardliu-db, line 31: I'm not sure if `pre-filter` is a used term in VL, but I found `pre-filter` hard to understand without context, especially when we are not using enum here and it's hard for us to see the comment (if any) explaining the meaning. 

The `pre` part is not clear to me. `pre` which filter? 
IIUC, `pre-filter` means pre-null-filter. It's easier to understand for me to call it `include-nulls`.

src/invalid.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export interface MarkInvalidMixins {
+  /**
+   * Invalid data mode for marks, which defines how the visualization should represent invalid values (`null` and `NaN` in continuous scales without defined output for invalid values) in the marks and their scale domains.
+   *
+   * -  `"filter"` — *Exclude* all invalid values from the visualization's *marks* and *scales*.
+   * For path marks (for line, area, trail), this option will create paths that connect valid points, as if the points with invalid values do not exist.
+   *
+   * - `"break-paths"` —
+   * Break path marks (for line, area, trail) at invalid values.  For non-path marks, this is equivalent to `"filter"`.
```
richardliu-db, line 17: This comment doesn't say anything about the "ignore in domain" thing. Ppl have to look at `"break-paths-keep-domains"` to guess that this is actually `"break-paths-ignore-domain"`. I'd prefer being explicit -- name this as "break-paths-ignore-domain" and add corresponding comment about the domain behavior.

src/invalid.ts
```diff
@@ -0,0 +1,79 @@
...
+import {isObject} from 'vega-util';
+
+/**
+ * Mixins for Vega-Lite Spec's Mark Definiton (to add mark.invalid)
+ */
+export interface MarkInvalidMixins {
+  /**
+   * Invalid data mode for marks, which defines how the visualization should represent invalid values (`null` and `NaN` in continuous scales without defined output for invalid values) in the marks and their scale domains.
+   *
+   * -  `"filter"` — *Exclude* all invalid values from the visualization's *marks* and *scales*.
```
richardliu-db, line 13: If I want these:
* ignore domain
* connect nulls with dotted lines.
How can we do it with these modes? I wonder if we should have "ignore-domain-connect-nulls" (temp name) mode and allow the users to customize how to connect the nulls, dotted, dashed, solid lines.

src/invalid.ts
```diff
@@ -0,0 +1,79 @@
...
+   * Break path marks (for line, area, trail) at invalid values.  For non-path marks, this is equivalent to `"filter"`.
+   * All *scale* domains will *exclude* these filtered data points.
+   *
+   * - `"break-paths-keep-domains"` —
+   * Break paths (for line, area, trail) at invalid values.  Hide invalid values for non-path marks.
+   * All *scale* domains will *include* these filtered data points.
+   *
+   * - `"include"` or `null` —
+   * Include all data points in the marks and scale domains. Each scale will use the output for invalid values defined in `config.scale.invalid`
+   * or, if unspecified, by default invalid values will produce the same visual values as zero (if the scale includes zero) or the minimum value (if the scale does not include zero).
```
richardliu-db, line 26: this behavior sounds more like `replace`?

src/invalid.ts
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * - `"break-paths-and-keep-path-domains"` (default) —
+   * This is equivalent to `"break-path-keep-domains"` for path-based marks (line/area/trail)
+   * and `"filter"` for other marks.
+   *
+   * __Note__: If any scale has an output for invalid values defined in `config.scale.invalid`,
+   * all values for the scales will be considered "valid" since they can produce a reasonable output for the scales.
```
richardliu-db, line 33: > considered "valid"

What does this imply?

src/invalid.ts
kanitw: "show at default position" that's super long for a property name.   I can rename to "show" though.

src/invalid.ts
kanitw: >connect nulls with dotted lines.

As discussed earlier, this is super complicated to implement so it's out of scope.

src/compile/scale/component.ts
```diff
@@ -31,16 +31,37 @@ export class ScaleComponent extends Split<ScaleComponentProps> {
...
-    return some(
-      this.get('domains'),
-      d => isArray(d) && d.length === 2 && isNumber(d[0]) && d[0] <= 0 && isNumber(d[1]) && d[1] >= 0
-    );
+
+    const scaleZero = this.get('zero');
+    if (
+      scaleZero === true ||
+      // If zero is undefined, linear/sqrt/pow scales have zero by default.
+      (scaleType === undefined && contains([ScaleType.LINEAR, ScaleType.SQRT, ScaleType.POW], scaleType))
```
richardliu-db, line 50: Confused. If scaleType === undefined, how can it be one of linear, sqrt, pow? Did you want to write `scaleZero === undefined`?

src/compile/scale/component.ts
```diff
@@ -31,16 +31,37 @@ export class ScaleComponent extends Split<ScaleComponentProps> {
...
+      return 'definitely-not';
     }
-    return some(
-      this.get('domains'),
-      d => isArray(d) && d.length === 2 && isNumber(d[0]) && d[0] <= 0 && isNumber(d[1]) && d[1] >= 0
-    );
+
+    const scaleZero = this.get('zero');
+    if (
+      scaleZero === true ||
```
richardliu-db, line 48: what's the type of `scaleZero`? boolean | undefined? If so, I might just use `scaleZero` to avoid checking with boolean.

src/compile/scale/component.ts
```diff
@@ -31,16 +31,37 @@ export class ScaleComponent extends Split<ScaleComponentProps> {
...
+    ) {
+      return 'definitely';
+    }
+
+    const domains = this.get('domains');
+
+    if (domains.length > 0) {
+      const hasDomainWithZero = some(
+        domains,
+        d => isArray(d) && d.length === 2 && isNumber(d[0]) && d[0] <= 0 && isNumber(d[1]) && d[1] >= 0
```
richardliu-db, line 60: curious, when will `d.length != 2`?

src/compile/scale/component.ts
```diff
@@ -31,16 +31,37 @@ export class ScaleComponent extends Split<ScaleComponentProps> {
...
+      scaleZero === true ||
+      // If zero is undefined, linear/sqrt/pow scales have zero by default.
+      (scaleType === undefined && contains([ScaleType.LINEAR, ScaleType.SQRT, ScaleType.POW], scaleType))
+    ) {
+      return 'definitely';
+    }
+
+    const domains = this.get('domains');
+
+    if (domains.length > 0) {
```
richardliu-db, line 57: I might write
```
if (domains.length === 0) {
  return 'maybe';
}
// the rest
```

src/compile/invalid/ScaleInvalidDataMode.ts
```diff
@@ -0,0 +1,51 @@
...
+import {getMarkPropOrConfig} from '../common';
+import {normalizeInvalidDataMode} from './normalizeInvalidDataMode';
+
+export type ScaleInvalidDataMode =
+  // remove 'break-paths-and-keep-path-domains' from MarkInvalidDataMode
+  // because it is a macro for '"filter"' or `"break-path-keep-domains`
+  | Omit<MarkInvalidDataMode, 'break-paths-and-keep-path-domains'>
+
+  // Add always-valid because at scale level, categorical scales can handle any values and thus is always valid.
+  | 'always-valid';
```
richardliu-db, line 16: what if we have `null` and "null" in data? `null` is as "valid" as "null"`. Maybe we want to allow the users to customize null text later.

src/compile/invalid/ScaleInvalidDataMode.ts
```diff
@@ -0,0 +1,51 @@
...
+  scaleType,
+  isCountAggregate
+}: {
+  markDef: MarkDef;
+  config: Config<SignalRef>;
+  scaleChannel: C;
+  scaleType: ScaleType;
+  isCountAggregate: boolean;
+}): ScaleInvalidDataMode {
+  if (!scaleType || !hasContinuousDomain(scaleType) || isCountAggregate) {
```
richardliu-db, line 31: What's `!scaleType` checking? If scaleType === undefined, it's linear?

src/compile/invalid/datasources.ts
```diff
@@ -0,0 +1,63 @@
+import {MarkInvalidDataMode} from '../../invalid';
+
+import {DataSourceType} from '../../data';
+import {normalizeInvalidDataMode} from './normalizeInvalidDataMode';
+
+type PreOrPostFilteringInvalidValues = 'pre-filter' | 'post-filter';
```
richardliu-db, line 6: Are `include-null` and `exclude-null` easier to understand?

src/compile/invalid/normalizeInvalidDataMode.ts
```diff
@@ -0,0 +1,15 @@
+import {MarkInvalidDataMode} from '../../invalid';
+
+type NormalizedMarkInvalidDataMode = Exclude<MarkInvalidDataMode, 'break-paths-and-keep-path-domains'>;
+
+export function normalizeInvalidDataMode(
+  mode: MarkInvalidDataMode | null | undefined,
+  {isPath}: {isPath: boolean}
+): NormalizedMarkInvalidDataMode {
+  if (mode === undefined || mode === 'break-paths-and-keep-path-domains') {
```
richardliu-db, line 9: should we have a const DefaultMarkInvalidDataMode = 'break-paths-and-keep-path-domains' and use it here?

src/compile/mark/encode/defined.ts
```diff
@@ -1,49 +1,48 @@
...
-  if (invalid) {
-    const signal = allFieldsInvalidPredicate(model, {channels: POSITION_SCALE_CHANNELS});
+  // For each channel (x/y), add fields to break path to a set first.
+  const fieldsToBreakPath = new Set<string>();
 
-    if (signal) {
-      return {defined: {signal}};
+  model.forEachFieldDef((fieldDef, channel) => {
+    let scaleType;
+    if (!isScaleChannel(channel) || !(scaleType = model.getScaleType(channel))) {
```
richardliu-db, line 21: nit: I prefer not to use assignment in condition in production code.

src/compile/mark/encode/zeroOrMinOrMax.ts
```diff
@@ -0,0 +1,75 @@
...
+    case 'fillOpacity':
+    case 'strokeOpacity':
+      return {value: config.scale[mode === 'zeroOrMin' ? 'minOpacity' : 'maxOpacity']};
+    case 'strokeWidth':
+      return {value: config.scale[mode === 'zeroOrMin' ? 'minStrokeWidth' : 'maxStrokeWidth']};
+    case 'size':
+      return {value: config.scale[mode === 'zeroOrMin' ? 'minSize' : 'maxSize']};
+    case 'strokeDash':
+    case 'shape':
+      return {value: null};
```
richardliu-db, line 73: what does this `{value: null}` mean? For dash, the line won't be dashed? For `shape`, what's null shape?

src/compile/mark/encode/defined.ts
kanitw: Not sure why? I don't think it hurts readability that much in this case. It actually compact the code. 

In any case, Vega-Lite uses this quite regularly, so let's not waste time with this.

src/compile/invalid/normalizeInvalidDataMode.ts
kanitw: that seems unnecessary? 

We use these undefined check everywhere in Vega-Lite.  `undefined` normally means default already.

src/compile/invalid/ScaleInvalidDataMode.ts
kanitw: when there is no scale.

src/compile/invalid/ScaleInvalidDataMode.ts
kanitw: Yeah, I agree it's an issue, but it's a separate issue orthogonal to this PR (which focuses on continuous scales).

src/compile/scale/component.ts
kanitw: yes we can use `scaleZero` too, but I think this reads clearer.

src/compile/invalid/ScaleInvalidDataMode.ts
kanitw: Btw, one thing I hate about Vega-lite code is that it's written before Typescript has `strictNullChecks` and we haven't migrated to turn it on.

src/compile/mark/encode/zeroOrMinOrMax.ts
kanitw: forgot to add  + will add 

```
// Vega-Lite don't allow strokeDash and shape to be used with continuous scale.
// So it should never reach here. Here we just return null so Typescript doesn't complain.
```

kanitw: Ideally we should narrow the type a bit more, but it's too much work.

src/compile/invalid/datasources.ts
kanitw: will rename to 'include/exclude-invalid-values' -- it's not just Null, Infinity and NaNs are invalid too.

src/compile/scale/component.ts
kanitw: I'd say the code here isn't long enough to use the early return pattern, which sometimes make the code harder to read, since readers need to negate the logic in their head.

To me, it's easier to follow that the code directly says that do something if the length is more than zero.  

(For longer methods, I agree that early return is good.) In any case, let's not die on this hill.

src/compile/scale/component.ts
kanitw: good call. I think this is from old code. It likely doesn't matter.

kanitw: will replace with

```
      const hasDomainWithZero = some(domains, d => {
        if (isArray(d)) {
          const first = d[0];
          const last = d[d.length - 1];
          return isNumber(first) && first <= 0 && isNumber(last) && last >= 0;
        }
        return false;
      });
```

src/compile/scale/component.ts
kanitw: good catch.

src/compile/data/filterinvalid.ts
kanitw: renamed.

src/invalid.ts
kanitw: Good question. Will add 

> Thus, fields for such channels will not be filtered and will not break paths.

src/compile/mark/encode/defined.ts
richardliu-db: https://eslint.org/docs/latest/rules/no-cond-assign Ppl might easily mess the code between `foo = bar` and `foo == bar`. You can keep it if it's an existing standard.

src/compile/invalid/ScaleInvalidDataMode.ts
richardliu-db: In which case does the channel have no scale?

src/compile/invalid/ScaleInvalidDataMode.ts
richardliu-db: I didn't find one searching `null categorical `. I'll defer to when we need that in Helios, since I guess you'll be busy and won't have the chance to add it if it's not requested from Helios.

richardliu-db (approved):
LGTM. Thanks for the awesome work!

src/compile/scale/component.ts
```diff
@@ -31,16 +31,41 @@ export class ScaleComponent extends Split<ScaleComponentProps> {
   }
 
   /**
-   * Whether the scale definitely includes zero in the domain
+   * Whether the scale definitely includes or not include zero in the domain
    */
-  public domainDefinitelyIncludesZero() {
-    if (this.get('zero') !== false) {
-      return true;
+  public domainHasZero(): 'definitely' | 'definitely-not' | 'maybe' {
```
lsh, line 36: nit: I'm not super sure about the wording here in this union, but non blocking either way.

src/invalid.ts
```diff
@@ -0,0 +1,77 @@
...
+   * Thus, fields for such channels will not be filtered and will not cause path breaks.
+   */
+  invalid?: MarkInvalidDataMode | null;
+}
+
+export type MarkInvalidDataMode =
+  | 'filter'
+  | 'break-paths-filter-domains'
+  | 'break-paths-show-domains'
+  | 'break-paths-and-show-path-domains'
```
lsh, line 43: Why is this the only one that includes the word `and`? Several of these options feel like conjunctive payloads, but this one is special for some reason?

src/compile/scale/component.ts
kanitw: yeah, i like it better than `boolean | null` where null is maybe though. Given it's internal, we can always change later.

src/invalid.ts
kanitw: good call! fixed.

lsh (commented):
Ham and I talked about the design here for a while yesterday and I'm convinced it's extensible and composable. Since he addressed my main nits about syntax, I'm good with these changes.

lsh (approved):

PBI-David:
I'm not 100% sure but possible regression here related to this?

https://stackoverflow.com/questions/78879887/vega-lite-5-30-0-retaining-null-values-for-proper-conditional-formatting/78879894?noredirect=1#comment139073178_78879894

Issue #9343, opened 2024-05-08
Binned stacked area doesn't correctly filter null 


kanitw:
## Bug Description

Describe how to reproduce this bug. Provide an [minimal, reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) spec in JSON, wrapped by triple backticks like this:

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "data": {"url": "data/movies.json"},
  "mark": "area",
  "encoding": {
    "x": {
      "bin": true,
      "field": "IMDB Rating"
    },
    "y": {"aggregate": "count"},
    "color": {"field": "Major Genre"}
  }
}
```

<img width="393" alt="image" src="https://github.com/vega/vega-lite/assets/111269/1a9ba4e1-b51e-40cc-a384-897c28a56d06">

PR #9344, opened 2024-05-08
fix(#9343): binned stacked area now correctly filter null values

kanitw:
binned stacked area now correctly filter null values

See the example fixed in the 4th commit.


yhoonkim (approved):
LGTM!

domoritz:
Nice

PR #9347, opened 2024-05-09
fix(#3488): don't use zero as min for size scale by default so no points will be ever hidden

### LINKED ISSUE #3488: Small non-zero values for size scale are not visible

kanitw:
Also adjust the default  `defaultMinSize` to be `4`, so the smallest point is not too big. 
fix #3488


kanitw:
I actually like that we now show zero in legend, so it's clear what's the min size shown. 

image

lsh (approved):
This change seems sensible to me, as does eventually allowing custom min and max.

PR #9358, opened 2024-05-27
feat: for interactive charts, changed cursor to pointer

domoritz:
replaces https://github.com/vega/vega-lite/pull/9305

### DIFF
--- a/src/compile/mark/init.ts
+++ b/src/compile/mark/init.ts
@@ -61,7 +61,8 @@ export function initMarkdef(originalMarkDef: MarkDef, encoding: Encoding<string>
     markDef.opacity = opacity(markDef.type, encoding);
   }
 
-  // set cursor, which should be pointer if href channel is present unless otherwise specified
+  // Set cursor, which should be pointer if href channel is present unless otherwise specified.
+  // We will also set the cursor in parse via getMarkGroup since we need access to the selections.
   const specifiedCursor = getMarkPropOrConfig('cursor', markDef, config);
   if (specifiedCursor === undefined) {
     markDef.cursor = cursor(markDef, encoding, config);
--- a/src/compile/mark/mark.ts
+++ b/src/compile/mark/mark.ts
@@ -313,7 +313,21 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
   const key = encoding.key;
   const sort = getSort(model);
   const interactive = interactiveFlag(model);
-  interactive ? (model.markDef.cursor ??= 'pointer') : {};
+
+  let isPoint = false;
+  let isBind = false;
+  if (model.component.selection) {
+    for (const k of keys(model.component.selection)) {
+      if (model.component.selection[k].bind) {
+        isBind = true;
+      }
+      if (model.component.selection[k].type == 'point') {
+        isPoint = true;
+      }
+    }
+  }
+  interactive && isPoint && !isBind ? (model.markDef.cursor ??= 'pointer') : {};
+
   const aria = getMarkPropOrConfig('aria', markDef, config);
 
   const postEncodingTransform = markCompiler[mark].postEncodingTransform
--- a/src/compile/mark/mark.ts
+++ b/src/compile/mark/mark.ts
@@ -313,6 +313,7 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
   const key = encoding.key;
   const sort = getSort(model);
   const interactive = interactiveFlag(model);
+  interactive ? (model.markDef.cursor ??= 'pointer') : {};
   const aria = getMarkPropOrConfig('aria', markDef, config);
 
   const postEncodingTransform = markCompiler[mark].postEncodingTransform
--- a/src/compile/mark/mark.ts
+++ b/src/compile/mark/mark.ts
@@ -314,19 +314,10 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
   const sort = getSort(model);
   const interactive = interactiveFlag(model);
 
-  let isPoint = false;
-  let isBind = false;
-  if (model.component.selection) {
-    for (const k of keys(model.component.selection)) {
-      if (model.component.selection[k].bind) {
-        isBind = true;
-      }
-      if (model.component.selection[k].type == 'point') {
-        isPoint = true;
-      }
-    }
+  // set pointer cursor for point selections that are not bound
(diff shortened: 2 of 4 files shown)


joelostblom:
Exciting! Would this PR also change the cursor automatically for interactive legends in a chart?

domoritz:
Not yet. But that's probably something we should do in Vega instead.

joelostblom:
> Not yet. But that's probably something we should do in Vega instead.

Following the comment in https://github.com/vega/vega/issues/3927#issuecomment-2157871819, it seems like this is already possible in Vega. E.g. [take this VL spec](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAGyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICUJNohSZ8gL5Lk2gNb6cbGrMyKQOU0gQQ0AbVBZHzh9QXI4WWUAfShhKEclCDhwqHdNEEwATxxQ9GdXdyUAMxoU5T9UfxAAeW0aWnkAXRsQJldlMIiokCsmpUioNmVXMjRQAA9xkFLy-QAJNjM4ZwB3OG0PbNz9AEcGJDc6NRpSXqUs6dnBTvQAWRpwiGjc7WiAcSRBQTktnLyQPtDphjiCzq0hj9NhlrrdavVGn8duhZGwEK4vucQGwvFA6JcMkMonR2PIMl5tD4umRIjE4jQEh5iF8GHkAIytZmCVloAAMlAATFZhUA); the only changes that needs to be made to the underlying Vega spec to get a cursor on the legend are the following:

```diff
@@ -202,6 +202,7 @@
           "name": "Origin_legend_labels",
           "interactive": true,
           "update": {
+            "cursor": {"value": "pointer"},
             "opacity": [
               {
                 "test": "(!length(data(\"legend_click_store\")) || (legend_click[\"Origin\"] && indexof(legend_click[\"Origin\"], datum.value) >= 0))",
@@ -215,6 +216,7 @@
           "name": "Origin_legend_symbols",
           "interactive": true,
           "update": {
+            "cursor": {"value": "pointer"},
             "fill": {"value": "transparent"},
             "opacity": [
               {
```

This seems very similar to the changes made in this PR for the marks of interactive chart, so maybe this same change to the legend could also be part of this PR? Or do you prefer if I open a separate issue in the VL repo for this and close the Vega issue https://github.com/vega/vega/issues/3927? Or do you mean that Vega should somehow automatically change the cursor whenever `"interactive": true` is present without the need for `"cursor": {"value": "pointer"}`?

src/compile/mark/mark.ts
```diff
@@ -277,6 +277,21 @@ function getMarkGroup(model: UnitModel, opt: {fromPrefix: string} = {fromPrefix:
...
+
+  // set pointer cursor for point selections that are not bound
+  if (
+    interactive &&
+    Object.values(model.component.selection).some(
+      (s) =>
+        s.type === 'point' &&
+        !s.bind &&
+        // if on is a pointerover (hover) the pointer makes less sense since the mark is not clickable.
+        (s as any).on !== 'pointerover',
```
hydrosquall, line 289: Is there a type we could use instead of `any` for checking the presence of `on` ?

src/compile/mark/init.ts
```diff
@@ -66,7 +66,8 @@ export function initMarkdef(originalMarkDef: MarkDef, encoding: Encoding<string>
     markDef.opacity = opacity(markDef.type, encoding);
   }
 
-  // set cursor, which should be pointer if href channel is present unless otherwise specified
+  // Set cursor, which should be pointer if href channel is present unless otherwise specified.
+  // We will also set the cursor in parse via getMarkGroup since we need access to the selections.
```
hydrosquall, line 70: `We will also set the cursor in parse`

Is this referring to the change in `legend/parse.ts`, or some other diff?

src/compile/legend/parse.ts
```diff
@@ -167,8 +167,8 @@ export function parseLegendForChannel(model: UnitModel, channel: NonPositionScal
         ...(selections?.length && isFieldDef(fieldOrDatumDef)
           ? {name: `${varName(fieldOrDatumDef.field)}_legend_${part}`}
           : {}),
-        ...(selections?.length ? {interactive: !!selections} : {}),
-        update: value,
+        ...(selections?.length ? {interactive: true} : {}),
```
hydrosquall, line 170: Checking my understanding - Should this be `interactive: true` even if there are no `selections` ?

src/compile/legend/assemble.ts
```diff
@@ -16,8 +16,8 @@ function setLegendEncode(
   legend.encode ??= {};
   legend.encode[part] ??= {};
   legend.encode[part].update ??= {};
-  // TODO: remove as any after https://github.com/prisma/nexus-prisma/issues/291
-  (legend.encode[part].update[vgProp] as any) = vgRef;
+  // @ts-expect-error expression is too complex for typescript to understand
```
hydrosquall, line 19: `expression is too complex for typescript to understand`

Confirming understanding that this  PR was merged, but the issue still appears.

examples/specs/interactive_bar_select_highlight.vl.json
```diff
@@ -18,8 +18,7 @@
   "mark": {
     "type": "bar",
     "fill": "#4C78A8",
-    "stroke": "black",
-    "cursor": "pointer"
```
hydrosquall, line 22: Assuming this was removed because the compiler now adds this automatically.

hydrosquall (commented):
## Docs

1. Can we update the PR description to note both marks and legends will be impacted?
2. Should we mention this behavior on the any of the [docs page](https://vega.github.io/vega-lite/docs/parameter.html) ?

## Code

- The change looks reasonable for the modified examples, thanks for adding the unit tests confirming this should only impact unbound point selections.
-  I would like to cross check the output on the preview branch. Can we change the base url? It looks like the CLI argument is now called `--base` [instead](https://vite.dev/guide/build.html#public-base-path)

image

domoritz:
I'll update the docs. Good idea. You should be able to push the cli changes directly to my branch.

src/compile/mark/init.ts
domoritz: It refers to the changes in mark.ts. This is not related to legends.

examples/specs/interactive_bar_select_highlight.vl.json
domoritz: Yep

PR #9361, opened 2024-05-30
fix: use array for single item valueRefs in conditional if test exists

richardliu-db:
## PR Description

In [this change to conditional.ts](https://github.com/vega/vega-lite/pull/9342/files#diff-f26cb911cd80159bbd659afb430b97fa7227afb4cc15e43565f2ca74c3361d22R64), we use object instead of array if there is a single value. But I found that test must be in an array form otherwise Vega doesn't recognize it.

This PR is to fix this issue -- always use array form if test exists, even if there is only a single item.

**Array**
image

**Object**
image

## Checklist

- For new features:
  - [ ] Has unit tests.

Tips:


kanitw (approved):

domoritz:
Isn't this a workaround an issue in Vega?

PR #9378, opened 2024-06-25
Update CONTRIBUTING.md

kanitw:


domoritz (changes requested):
The editor should pull in the typescript sources so builds should not be needed anymore.

I'm against this change.

PR #9394, opened 2024-07-28
feat: update code to not need implicit any, add hasProperty util

domoritz:
* update deps including typescript
* fix a few issues along the way
* improve code robustness for example for type checks and index accesses
* remove need for suppressImplicitAnyIndexErrors
* introduces new utility method `hasProperty` which checks whether the object has the property and it's not undefined


lsh (approved):
Strongly in favor of these changes. This gives us a roadmap of where we can improve on types in the future.

test/example.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    {type: 'interval', encodings: ['y']},
+    {
+      x: {aggregate: 'count', type: 'quantitative'},
+      y: {bin: true},
+      color: {value: 'steelblue'}
+    }
+  );
+  expect(s).toBeDefined();
+
+  console.log(JSON.stringify(s));
```
kanitw, line 17: @lsh @domoritz Was this console.log intended? It's polluting the test output.

domoritz: Not sure what the test does here. Removed in https://github.com/vega/vega-lite/pull/9517.

PR #9395, opened 2024-07-30
fix: make tick size responsive to plot height 

kanitw:
## PR Description

Before:

Tick size doesn't respond to plot size change

image

after: 

Now it does:

image


kanitw:
- [ ] need to fix these examples:

image

ZacharyBys (approved):

PR #9400, opened 2024-07-31
Cameron.yick/check vega lite compile spec discrete legend

hydrosquall:
## Motivation

- Fixes https://github.com/vega/vega-lite/issues/9399

## Testing

- Check sandbox: the `gradientLength` property should be respected for a color legend that uses quantile, quantize, or threshold scales.

## Notes

- I temporarily patched the type in `vega-lite` so this PR can be reviewed. However, we could make the diff in this repo smaller if we instead push a change to the `LegendType` field inside of `vega` instead.


src/compile/legend/parse.ts
```diff
@@ -126,7 +126,10 @@ export function parseLegendForChannel(model: UnitModel, channel: NonPositionScal
 
   for (const property of LEGEND_COMPONENT_PROPERTIES) {
     if (
+      // Continuous and discrete gradients ignore symbol properties
       (legendType === 'gradient' && property.startsWith('symbol')) ||
+      (legendType === 'discrete' && property.startsWith('symbol')) ||
```
hydrosquall, line 131: Before I add any tests or open PRs in `vega`, I want to check if this is the behavior (a third legendType called discrete) we want as a public API, or if this was supposed to be a private/internal API.

The discrete legend type was added by @jheer in https://github.com/vega/vega/commit/ddd2bbff9897415be968fb60e2afa6c695c9eec4 , so I'll bring it up as a new issue in the vega repo.

hydrosquall: I added the new type in https://github.com/vega/vega/pull/3957 . We'll need a new release of `vega` to happen in order for vega-lite to be able to reference it.

PR #9411, opened 2024-08-22
fix: make 1D tick mark responsive to width/height 

kanitw:
## PR Description

See fixed example + relevant unit tests


ZacharyBys (approved):
LG, thanks for fixing

PR #9427, opened 2024-09-05
docs: add step size to histogram_rel_freq

domoritz:
This improves the ticks on the x axis.


joelostblom (approved):
I agree it looks better with the ticks between the bars instead of in the middle of some of them

Issue #8675, opened 2023-01-28
Facet Cannot Be Correctly Sorted


PBI-David:
I realised we don't have a VL example of the Vega calendar so I tried to create one to submit as a PR. Everything is working fine apart from years with incomplete months cannon be sorted. Observe that Jan, Feb, Mar, Apr for 2020 are all pushed to the end of the chart and listed under Sep, Oct, Nov & Dec. I have tried various documented sort operations but this seems like a bug. 

![image](https://user-images.githubusercontent.com/86846675/215265021-cb93748c-3dee-4b84-a8b7-40bea6297d20.png)

[...]

Issue #5937, opened 2020-02-19
Row and column sorting do not work for nominal and ordinal types


sorig:
Sorting of row and column encodings seems to be broken for discrete data types. Here is an unsorted specification that is working as expected:

[Unsorted spec](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoAWOgCtY3AHaQANOCgYATgGtsYUBAiQ8AT0LJ9kdui2QNAX3XGAJujyYcho5AroxrMiw+gDaGkYGkACCYITcwTiQAHLcYC6MAGYZyFrISvCWqlAx6XaJKWmZ2bn5hZAAQrHxVgAyyBl4ag2VZVAAYrYsOZBO4UbA0U0JUACyjC4uYoXFPVZJyLbDRd1x05BtHV2NpVYDWkN2oxEQEzG7VgBKjNT4XSWMvZBnF0dTj8+vVTdE6Jb4oS7Oa63P6JJ4vTpA96fdabOxAxr3WEAhHAj5rDa5CFjG6TTGzeaLOpI-Go35kyBzBZLX4gqAowkjSERaH0xmUt6rUGDcF05qJPnM9GC-rC4ZXKGksVs1LpLI5PIFAWs5Iqqrq2qivYVVXVDV1Y548q6tU1TXy7mKvYHHHUoXnEVS+nOlmWmXuuVc8aO1rtF3S5IErY7JX7UM+5GRonXSJ3GMSqnh9lRjExuGA3GfMEB4kpmFQb2I8NFtHRp1xqXa6sjMYAXUckMgGu46SU1H03mMWm4AHd+yXIBlGMgxC4rNTA8YzBYrEpuKhGEp-PYIvaoPBuAFUCovOPJ9PZ4lU8EF1Al5ZEqv15uxNujLvIAAPMfJidTmdWHNr3HO8VzXDctxvSAEH8e8DBLVw13QDdQngnxjT1W1ClQ4wK2w8kmSw5MIkgPNOlQttkwcMZ31Mb9rhMcxYMgABHVh0CUPBmHcRgqDUcd0GoahcloPAmP3VgONfCB333MRuF6AcfDPf9EgtNFgMY0Cny3aj2w0SBh3mFh9AARgABjMjsUGxUyLJABwgA)

[...]

PR #9434, opened 2024-09-16
fix: facet missing data bug fix

wirhabenzeit:
## PR Description

This is a suggested fix for https://github.com/vega/vega-lite/issues/5937 and potentially also https://github.com/vega/vega-lite/issues/8675 and https://github.com/vega/altair/issues/3588 and https://github.com/vega/altair/issues/3481

I included two examples demonstrating the bug. For instance the spec 

```json
{
  "facet": {
    "column": {"field": "Origin", "sort": ["USA", "Europe", "Japan"]},
    "row": {"field": "Cylinders", "sort": {"field": "Horsepower", "op": "mean"}}
  },
  "spec": {
      "mark": "point",
      "encoding": {
        "y": {"field": "Acceleration", "type": "quantitative"},
        "color": {"field": "Origin", "type": "nominal"},
        "shape": {"field": "Cylinders", "type": "nominal"}
      }
  }
}
```

results in 

visualization-2

while it should look like this:

facet_bug_cars

I think the bug is due to the following. The key part of the compiled Vega spec looks like 

```json 
{
      "name": "cell",
      "type": "group",
      "style": "cell",
      "from": {
        "facet": {
          "name": "facet",
          "data": "source_0",
          "groupby": ["Cylinders", "Origin"],
          "aggregate": {
            "cross": true,
            "fields": [
              "mean_Horsepower_by_Cylinders",
              "column_Origin_sort_index"
            ],
            "ops": ["max", "max"],
            "as": ["mean_Horsepower_by_Cylinders", "column_Origin_sort_index"]
          }
        }
      },
}
```

As far as I understand this performs a `groupby` according to cylinders and origin, and then for each group determines `mean_Horsepower_by_Cylinders` and `column_Origin_sort_index ` used for sorting by aggregating over the data points in the facet and taking the max (it does not matter as the column sort index is the same for all points in the facet anyways, so one could also take e.g. the min). The problem arises when a facet is empty as then the aggregation cannot determine the valid sort index. As a result, some facets are not assigned a sort-index, see this screenshot from the Vega Editor -> Data Viewer -> cell:

Screenshot 2024-09-17 at 00 32 24

For some reason the cells with missing sort-index result in some data points being placed in the wrong facet (not exactly sure what goes wrong here exactly). 

My bug fix suggestion simply replaces the above spec by 

```json
{
      "name": "cell",
      "type": "group",
      "style": "cell",
      "from": {
        "facet": {
          "name": "facet",
          "data": "source_0",
          "groupby": [
            "mean_Horsepower_by_Cylinders",
            "column_Origin_sort_index"
          ],
          "aggregate": {"cross": true}
        }
      },
}
```

i.e. it directly groups by the sort index, avoiding cells without assigned indices. This is simpler than the original code. For all examples I looked at, it worked. However, maybe I overlook some unintended side effects. 

Out of all the tests a single test in `test/compile/facet.test.ts` fails, namely line 507. As far as I can see this test directly checks the section of the compiled Vega spec outlined above, so in some sense I would expect it to fail. 

## Checklist

### DIFF
--- a/src/compile/data/facet.ts
+++ b/src/compile/data/facet.ts
@@ -1,5 +1,5 @@
 import type {AggregateOp} from 'vega';
-import {isArray} from 'vega-util';
+import {isArray, stringValue} from 'vega-util';
 import {isBinning} from '../../bin.js';
 import {COLUMN, FACET_CHANNELS, POSITION_SCALE_CHANNELS, ROW} from '../../channel.js';
 import {vgField} from '../../channeldef.js';
@@ -8,7 +8,7 @@ import {hasDiscreteDomain} from '../../scale.js';
 import {DEFAULT_SORT_OP, EncodingSortField, isSortField} from '../../sort.js';
 import {hash} from '../../util.js';
 import {isVgRangeStep, VgData} from '../../vega.schema.js';
-import {FacetModel} from '../facet.js';
+import type {FacetModel} from '../facet.js';
 import {HEADER_CHANNELS, HEADER_TYPES} from '../header/component.js';
 import {Model} from '../model.js';
 import {assembleDomain, getFieldFromDomain} from '../scale/domain.js';
@@ -18,6 +18,15 @@ import {DataFlowNode} from './dataflow.js';
 interface ChildIndependentFieldsWithStep {
   x?: string;
   y?: string;
+}
+
+function isCrossedFacetWithCustomSort(model: FacetModel) {
+  const {row, column} = model.facet;
+  return !!(row && column && (isCustomSortField(row) || isCustomSortField(column)));
+}
+
+function isCustomSortField(fieldDef?: FacetModel['facet']['row']) {
+  return !!fieldDef && (isSortField(fieldDef.sort) || Array.isArray(fieldDef.sort));
 }
 
 interface FacetChannelInfo {
@@ -61,7 +70,7 @@ export class FacetNode extends DataFlowNode {
           name: model.getName(`${channel}_domain`),
           fields: [vgField(fieldDef), ...(isBinning(bin) ? [vgField(fieldDef, {binSuffix: 'end'})] : [])],
           ...(isSortField(sort)
-            ? {sortField: sort}
+            ? {sortField: sort as EncodingSortField<string>}
             : isArray(sort)
               ? {sortIndexField: sortArrayIndexField(fieldDef, channel)}
               : {}),
@@ -128,7 +137,6 @@ export class FacetNode extends DataFlowNode {
     for (const channel of POSITION_SCALE_CHANNELS) {
       const childScaleComponent = this.childModel.component.scales[channel];
       if (childScaleComponent && !childScaleComponent.merged) {
-        // independent scale
         const type = childScaleComponent.get('type');
         const range = childScaleComponent.get('range');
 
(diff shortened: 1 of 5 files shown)


domoritz:
Thanks for the pull request. Please make sure that the test pass and if you can that the linting also passes.

wirhabenzeit:
@domoritz As I mentioned in the pull request there is one failing test, namely 
https://github.com/vega/vega-lite/blob/d5e98a8b4cff5567aecaedf2b3e8af6190256a57/test/compile/facet.test.ts#L490-L518 

Since my suggested workaround changes the facet grouping as described above, it is clear that this test should fail in 
https://github.com/vega/vega-lite/blob/d5e98a8b4cff5567aecaedf2b3e8af6190256a57/test/compile/facet.test.ts#L507-L512

I have now updated the testing file in my fork by changing the above to

```ts
expect(marks[0].from.facet.aggregate).toEqual({ 
   cross: true, 
 }); 
```

Regarding the linting, it already seems to pass?

joelostblom:
@domoritz Is this PR looking good to you? Fixing this bug in VL would help us understand if there is anything additional going wrong in Altair for https://github.com/vega/altair/issues/3588.

domoritz:
I haven't had time to review it again yet. Maybe someone else can help so we are not blocked on me.

wirhabenzeit:
Any update on this? If it helps I can look into potential conflicts with updates to main since my original MR. Would be great to have a fix for this to be able to work with categorical data reliably

domoritz:
Thanks for bubbling this up. Yes, that would be great.

wirhabenzeit:
@domoritz I have updated the pull request. I noticed that my previous fix was too simplistic and did not work in all cases. I believe the current version should cover every edge case

src/compile/data/facet.ts
```diff
@@ -61,7 +70,7 @@ export class FacetNode extends DataFlowNode {
           name: model.getName(`${channel}_domain`),
           fields: [vgField(fieldDef), ...(isBinning(bin) ? [vgField(fieldDef, {binSuffix: 'end'})] : [])],
           ...(isSortField(sort)
-            ? {sortField: sort}
+            ? {sortField: sort as EncodingSortField<string>}
```
domoritz, line 73: Why is this `as` needed? Shouldn't `isSortField` already check that?

src/compile/facet.ts
```diff
@@ -35,6 +35,15 @@ export function facetSortFieldName(
...
 }
 
+function isCrossedFacetWithCustomSort(facet: Pick<EncodingFacetMapping<string, SignalRef>, 'row' | 'column'>) {
+  const {row, column} = facet;
+  return !!(row && column && (isCustomSortField(row) || isCustomSortField(column)));
+}
+
+function isCustomSortField(fieldDef?: FacetFieldDef<string, SignalRef>) {
+  return !!fieldDef && (isSortField(fieldDef.sort) || Array.isArray(fieldDef.sort));
+}
```
domoritz, line 45: is this code redundant with the code above?

domoritz:
I created https://github.com/vega/vega-lite/pull/9818 with an internal fork so we can run the format action. Please finish the code here and I will move it over so we can merge.

fxwiegand:
@wirhabenzeit any movement on this? Still the only real fix for [#5937](https://github.com/vega/vega-lite/issues/5937), would love to see it land.

PR #9450, opened 2024-10-04
fix: links in docs

lukaw3d:
## PR Description

Fixes these links on https://vega.github.io/vega-lite/docs/:
> This documentation describes the [JSON specification language](https://vega.github.io/vega-lite/docs/#spec)
> ...
> - [Point Selection Properties](https://vega.github.io/vega-lite/docs/selection.html#point)
> - [Interval Selection Properties](https://vega.github.io/vega-lite/docs/selection.html#interval)

### DIFF
--- a/site/docs/parameter/select.md
+++ b/site/docs/parameter/select.md
@@ -56,9 +56,9 @@ A [selection's type](parameter.html#select) determines which data values fall wi
 
 ### Selection Projection with `encodings` and `fields`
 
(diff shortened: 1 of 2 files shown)


site/docs/parameter/select.md
domoritz, line 1: Why did this file change?

domoritz (changes requested):
Thanks. Looks like there are two unrelated changes in this pull request?

site/docs/parameter/select.md
domoritz: Oh, I see. We have the same ID twice. Good catch!

PR #9462, opened 2024-11-01
docs: Indicate what type of CI is made

joelostblom:
Minor clarification to the docs so that it is clear what type of CI is added. This info is available in the aggregation docs, but not in the docstring to the errorbar and errorband marks.


domoritz (approved):
Great addition.

PR #9472, opened 2024-11-15
chore: update deps and rebuild examples

domoritz:
Follow up from https://github.com/vega/vega-lite/pull/8921


domoritz:
Merging since this is blocking builds.

PR #9476, opened 2024-11-20
chore: add missing file extension

domoritz:
@jonathanzong @arvind this was missing from https://github.com/vega/vega-lite/pull/8921


domoritz:
Merging to fix the missing extension. Please let me know if there is something wrong with this @arvind @jonathanzong

PR #9479, opened 2024-11-23
apply single quote escape

yhoonkim:
## PR Description
Fixes the issue of timeunit band position transforms which don't escape the field name correctly.


domoritz:
Can you link to the issue?

yhoonkim:
> Can you link to the issue?

Done! (edited the PR description)

domoritz:
If you say "fixes" or "closes" before it, GitHub will link it and auto close the issue when the pull request is merged.

src/util.ts
```diff
@@ -297,6 +297,10 @@ function escapePathAccess(string: string) {
   return string.replace(/(\[|\]|\.|'|")/g, '\\$1');
 }
 
+export function escapeSingleQuotes(value: string) {
```
kanitw, line 300: It seems like a code smell that we need to add a new `escapeSingleQuotes` utility method.  

Shouldn't we already have another escape utility in other places? (Also, for more than just "single quotes"?)

yhoonkim: I tried other utility functions to re-escape the field name but could not find a good one.  Maybe I will try to find again and ping you if I can't find.

domoritz: There may be some similar logic in Vega (utils).

yhoonkim: Ended up with creating more generic function, `accessWithDatumToUnescapedPath`, to solve this case.
`accessWithDatumToUnescapedPath` is taking un-escaped string to build the datum-access expression.

src/compositemark/errorbar.ts
```diff
@@ -556,7 +559,7 @@ function errorBarAggregationAndCalculation<
     for (const postAggregateCalculate of postAggregateCalculates) {
       tooltipSummary.push({
         fieldPrefix: postAggregateCalculate.as.substring(0, 6),
-        titlePrefix: replaceAll(replaceAll(postAggregateCalculate.calculate, 'datum["', ''), '"]', '')
+        titlePrefix: replaceAll(replaceAll(postAggregateCalculate.calculate, "datum['", ''), "']", '')
```
domoritz, line 562: do we use accessWithDatumToUnescapedPath here?

yhoonkim: Yes. 

1) The existing logic peel off `datum["`  and `"]` from the `calculate` to create a prefix. 
E.g., `datum["yh"]`  -> `yh`. 

2) I replaced the double quote datum expression with `accessWithDatumToUnescapedPath`  using the single quote.

3) To keep the existing logic, I changed the code to peel off `datum['`  and `']`

kanitw (approved):
Approving provided that you have sufficiently addressed @domoritz's concern. 

(I haven't checked if you do, but at least you have sufficiently addressed mine.)

PR #9512, opened 2025-01-20
fix: Arc mark does not stack for `quantitative` definitions when different fields are used for `theta` and `radius` [draft]

### LINKED ISSUE #7957: Radial chart radius field unexpectedly changes arc stack behavior
Please:

- [x] Check for duplicate issues. Please file separate requests as separate issues.
- [x] Describe how to reproduce the bug.
- [x] Provide an [minimal, reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) spec in JSON, wrapped by triple backticks like this:

Using field for `radius` other than the one provided to `theta` unexpectedly breaks the `stack` behavior for `theta`.  It seems like `radius` should be able use a different scale than `theta`. For example, with the following code, the arcs do not stack on the `theta` variable as expected.

```json
{
  "data": {
    "values": [
      { "segment": 1, "value": 8 },
      { "segment": 2, "value": 7 },
      { "segment": 3, "value": 6 }
    ]
  },
  "mark": {"type": "arc", "innerRadius": 20, "stroke": "#fff", "opacity": 0.5},
  "encoding": {
    "theta": {"field": "segment", "type": "quantitative", "stack": true},
    "radius": {"field": "value", "scale": {"type": "sqrt", "zero": true, "rangeMin": 20}}
  }
}
```

- [x] If applicable, include error messages and screenshots, GIF videos (e.g. using https://www.cockos.com/licecap/), or working examples (e.g. by clicking share in the [Vega-Editor](https://vega.github.io/editor/) or https://bit.ly/vega-lite-blocks)
* [Working version](https://vega.github.io/editor/?#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoCOQRxZmNAEYl23XDQAOAL4KTZi1dQAmOzr00AHY3DzJzSzQAZn8HNAA2FwBdNxBkACcAazRQTABPHEd0JHSoRRAaWVk4dIAlJGUaBgNfAAYlCEx0tkyikABiADNh8rYcJCg6PLRWyjFUiyg2RtkyHJB6OHV1wZo4QWU0EFNwr3L8wqOARwYkSzo1GlJyzons1C69VPSGppbQXb7Q7oE4RKwdKA6Iq5Ap9CBXdLgkAALxqbDQnzgSh+qzgAFlKmgfK0XKSXEA): `theta` and `radius` use same field
image

* [Broken version](https://vega.github.io/editor/?#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoCOQRxZmNAEYl23XDQAOAL4KTZi1dQAmOzr00AHY3DzJzSzQAZn8HNAA2FwBdNxBkACcAazRQTABPHEd0JHSoRRAaWVk4dIAlJGUaBgNfAAYlCEx0tkyikABiADNh8rYcJCg6PLRWyjFUiyg2RtkyHJB6OHV1wZo4QWU0EFNwr3L8wqOARwYkSzo1GlJyzons1C69VPSGppbQXb7Q7oeyBDpQHRFXIFPoQK7pKxKABeNTYaE+cCUP1WcAAspU0D5Wi4SS4gA): `theta` and `radius` use different fields
image

hydrosquall:
## Motivation

- When using different marks for theta vs radius using an `arc` mark with stacks, change graph rendering
- Fixes https://github.com/vega/vega-lite/issues/7957

## Changes

- Do not include `groupBy` field when the `dimension` field is using a `quantitative` scale. This ensures that stacking is applied correctly.
- I've made this change only impact polar coordinate channels to keep the impact minimal, since this customization is specific to the arc mark.
- Try the VL Spec from the original bug report  
  - [preview](https://cameron-yick-fix-compile-pie.vega-lite.pages.dev/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoCOQRxZmNAEYl23XDQAOAL4KTZi1dQAmOzr00AHY3DzJzSzQAZn8HNAA2FwBdNxBkACcAazRQTABPHEd0JHSoRRAaWVk4dIAlJGUaBgNfAAYlCEx0tkyikABiADNh8rYcJCg6PLRWyjFUiyg2RtkyHJB6OHV1wZo4QWU0EFNwr3L8wqOARwYkSzo1GlJyzons1C69VPSGppbQXb7Q7oeyBDpQHRFXIFPoQK7pKxKABeNTYaE+cCUP1WcAAspU0D5Wi4SS4gA)
  - You can also check with the [colorBy case](https://cameron-yick-fix-compile-pie.vega-lite.pages.dev/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoCOQRxZmNAEYl23XDQAOAL4KTZi1dQAmOzr00AHY3DzJzSzQAZn8HNAA2FwBdNxBkACcAazRQTABPHEd0JHSoRRAaWVk4dIAlJGUaBgNfAAYlCEx0tkyikABiADNh8rYcJCg6PLRWyjFUiyg2RtkyHJB6OHV1wZo4QWU0EFNwr3L8wqOARwYkSzo1GlJyzons1C69VPSGppbQXb7Q7oeyBDpQHRFXIFPoQK7pKxKABeNTYaE+cCUP1WcAAspU0D5Wi5UktBGx0js9gcjicIlYSS4gA)
  - For comparison [this shape is cool](https://cameron-yick-fix-compile-pie.vega-lite.pages.dev/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoCOQRxZmNAEYl23XDQAOAL4KTZi1dQAmOzr00AHY3DzJzSzQAZn8HNAA2FwBdNxBkACcAazRQTABPHEd0JHSoRRAaWVk4dIAlJGUaBgNfAAYlCEx0tkyikABiADNh8rYcJCg6PLRWyjFUiyg2RtkyHJB6OHV1wZo4QWU0EFNwr3L8wqOARwYkSzo1GlIQVPSGppbQXf3D9HtAjpQHRFXIFPoQK7pKxKABeNTYaC6eiUb1WcAAspU0D5WqlOhNsqgkXAXKSgA), and shows we can still get something useful when the stack is applied on `radius` rather than `theta`

## Testing

- API design: does this approach make sense for how `stack`s should apply for `arc` marks?
- See unit test suite : Let me know if additional cases should be added to either the image snapshots or the unit test suite.
- Compare with [similar cartesian plot](https://vega.github.io/editor/?#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoCOQRxZmNAEYl23XDQAOAL4KTZi1dQAmOzr00AHY3DzJzSzQAZn8HNAA2FwBdNxBkACcAazRQTABPHEd0dLgoKyUITHS2TKKQAGIAM2bFEDYcJCg6PLQABkoxVIsoNmUaWTIckAAPKcaaOEFlNBBTcK9W-MKVgEcGJEs6NRpSVsrO7NQqvVSezRB5xeX0e0CKqB0i3IK64Vk4JDpEAuEEuIA)


src/stack.ts
```diff
@@ -176,9 +176,15 @@ export function stack(m: Mark | MarkDef, encoding: Encoding<string>): StackPrope
...
+    const hasSameDimensionAndStackedField = dimensionField && dimensionField === stackedField;
 
-    if (dimensionField && dimensionField !== stackedField) {
+    // This change breaks cartesian tests, just pushing to start the discussion. When dev'ing locally, try removing this condition.
+    // TBD: we should only enter this branch if there's also a "color" and maybe a group channel?
+    const isPolar = isPolarPositionChannel(fieldChannel) || isPolarPositionChannel(dimensionChannel);
+
+    if (!hasSameDimensionAndStackedField && (isPolar ? dimensionChannel === fieldChannel || encoding['color'] : true)) {
       // avoid grouping by the stacked field
+      // TKTK: find out why
```
hydrosquall, line 187: If this works here, we may need a similar change for the code block under `dimensionOffsetChannel` .

src/stack.ts
```diff
@@ -182,7 +182,7 @@ export function stack(m: Mark | MarkDef, encoding: Encoding<string>): StackPrope
     // TBD: we should only enter this branch if there's also a "color" and maybe a group channel?
     const isPolar = isPolarPositionChannel(fieldChannel) || isPolarPositionChannel(dimensionChannel);
 
-    if (!hasSameDimensionAndStackedField && (isPolar ? dimensionChannel === fieldChannel || encoding['color'] : true)) {
+    if (!hasSameDimensionAndStackedField && (isPolar ? dimensionChannel === fieldChannel : true)) {
```
hydrosquall, line 185: Before I removed the color encoding check, the graph rendered like this: 

image

src/stack.ts
```diff
@@ -176,8 +176,13 @@ export function stack(m: Mark | MarkDef, encoding: Encoding<string>): StackPrope
   if (encoding[dimensionChannel]) {
     const dimensionDef = encoding[dimensionChannel];
     const dimensionField = isFieldDef(dimensionDef) ? vgField(dimensionDef, {}) : undefined;
+    const hasSameDimensionAndStackedField = dimensionField && dimensionField === stackedField;
```
kanitw, line 179: just having the same input field doesn't mean that it's the same transformed field? I think we need to check if they have the same aggregate/bin/timeUnit?

src/stack.ts
```diff
@@ -176,8 +176,13 @@ export function stack(m: Mark | MarkDef, encoding: Encoding<string>): StackPrope
...
 
-    if (dimensionField && dimensionField !== stackedField) {
+    // For polar coordinates, do not set a groupBy when working with quantitative fields.
+    const isPolar = isPolarPositionChannel(fieldChannel) || isPolarPositionChannel(dimensionChannel);
+    const shouldAddPolarGroupBy = !isUnbinnedQuantitative(dimensionDef);
+
+    if (isPolar ? shouldAddPolarGroupBy : !hasSameDimensionAndStackedField) {
```
kanitw, line 185: I wonder if we really need `isPolar ? shouldAddPolarGroupBy :` clause?   If we strictly check the field above, would that be sufficient?

kanitw (commented):
Thanks for the PR. I think the test and regression case look good, but I wonder if the condition added can be more precise/simpler.

src/stack.ts
hydrosquall: Ah interesting. Do you know if we have helper utilities for checking if two encodings are "equivalent"? (I.e. matching in field name and any relevant inline transformations like bin/timeUnit/bin? 

I looked in channelDef just now and didn't see anything that matches. I can add one if it seems like it would be appropriate here.

src/stack.ts
hydrosquall: By strictly check, do you mean we'd check if `dimensionDef` and `stackedFieldDef` have the same

- `field` (the current check)
- `bin`
- `aggregator`
- `timeUnit`

for both polar and cartesian graphs?

Currently if I just remove the `isPolar ? shouldAddPolarGroupBy` clause and use the existing check, we'd introduce [this bug](https://github.com/vega/vega-lite/pull/9537/commits/c2de794c127f33c19ccec6e47adfb442c13c98a7)

hydrosquall: Made a PoC commit to see if this is what you had in mind

https://github.com/vega/vega-lite/pull/9537/commits/8b010a61aed5a455d5e60e52ec9ddb1c18323c63

kanitw: >we'd introduce this bug

can't find this link

src/stack.ts
kanitw: Reading this method, I think we check if it's the same field by converting a field def to vgField first. 

For example, see this block in this method. 

```ts
const dimensionField = isFieldDef(dimensionDef) ? vgField(dimensionDef, {}) : undefined;
if (dimensionField && dimensionField !== stackedField) {
```

src/stack.ts
hydrosquall: Fixed link: https://github.com/vega/vega-lite/blob/231ed17c2b37811701a89029ab1524750abdd80b/examples/compiled/arc_quantitative_stacked_theta.png

hydrosquall:
Shrinking the example down to see if the issue persists in both coordinate systems

- cartesian version of the “bug” (having a common x1) [here](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoCOQRxZmNAEYl23XDQAOAL4KTZi1dQAmOzr00AHY3DzJzSzQAZn8HNAA2FwBdNxBkACcAazRQTABPHEd0dLgoKyUITHS2TKKQAGIAM2bFEDYcJCg6PLQABkoxVIsoNmUaWTIckAAPKcaaOEFlNBBTcK9W-MKVgEcGJEs6NRpSVsrO7NQqvVSezRB5xeX0e0CKqB0i3IK64Vk4JDpEAuEEuIA) (pic 1)
- polar version of the bug (having a common theta1) [here](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoCOQRxZmNAEYl23XDQAOAL4KTZi1dQAmOzr00AHY3DzJzSzQAZn8HNAA2FwBdNxBkACcAazRQTABPHEd0JHSoRRAITHS2TKKQAGIAM2bythwkKDo8tAAGSjFUiyg2ZRpZMhyQejh1ScaaOEFlNArPSKV8wpWARwYkSzo1GlJyyo7s1Cq9VPSkUYYDTRB5xeX0e0ClaB0i3IK64SyOAlEAuMEuIA) (pic 2) 

| Cartesian | Polar |
| --- | --- |
| image |  image |

src/stack.ts
hydrosquall: I found it interesting to see how many of the visual tests break if I removed the check that only applies this change to polar encodings.  

https://github.com/vega/vega-lite/pull/9557

hydrosquall:
I’ve been thinking more about why the polarCoordinates check was needed in this PR (and revisiting the grammar of graphics book) as I have tried a few variations on trying to get the intended behavior without checking for polar coordinates, and I noticed something that makes me feel more OK about having a special condition for polar.

- We [write](https://github.com/vega/vega-lite/blob/2ec1bf3d2e02a5e914876e79c069a7620e991998/src/mark.ts#L59) that `arc` is analogous to `rect`
- However, `rect` does not support stacking  (even if forced), only `bar` does! (according to [STACKABLE_MARKS](https://github.com/vega/vega-lite/blob/2ec1bf3d2e02a5e914876e79c069a7620e991998/src/stack.ts#L76))
- Therefore, `arc` is more like `bar` than `rect` . If we want `arc` to have attributes of both `rect` and `bar`, we are going to have to add special logic for it in more places given that in cartesian coordinates, we aren't treating rect and bar the same way when it comes to stacking.

If we want stacking to be supported by `arc` that doesn’t have to care about whether we are in cartesian or polar, we have the choices of

- Allow stacking to be forced in `rect` too (currently not supported)
- Leave `rect` alone, and just allow forcing stacking via `arc` (as I [do in the current PR)](https://github.com/vega/vega-lite/pull/9512) or
- Introduce a new mark (`radial_bar`?) which is to `arc` as `bar` is to `rect` . We leave the current `arc` behavior alone, and we only give the new behavior of enabling force stacking to `radial_bar`.
- Something else?

domoritz:
@kanitw @hydrosquall what's left here to merge this?

hydrosquall:
I’m hoping to get @kanitw’s review on whether this change is usable.

We both wish a different solution that wasn’t polar aware might be
possible, but I haven’t found a way to do it yet.

domoritz:
@kanitw can you do a review on this so we can merge? This just came up in Altair: https://github.com/vega/vega-lite/issues/9680

hydrosquall:
(Notes from running a trio of Claude Opus 4.6 on this approach)

## Analysis: Can the polar coordinate check be simplified?

Both @kanitw and I wished for a simpler condition that doesn't need `isPolarPositionChannel`. After systematic exploration, the answer is **no**: the check is load-bearing.

<details>
<summary>Why every coordinate-agnostic simplification fails</summary>

The test at `test/stack.test.ts:373` ("should stack even for two plain quantitatives with the orient on the axes") requires that cartesian charts with two different unbinned quantitative fields (e.g. `x: 'IMDB Rating'`, `y: 'Rotten Tomatoes Rating'`) add the dimension to `groupBy`. But the polar test cases at line 451-487 require the opposite: `groupbyChannels = []` when theta and radius are different unbinned quantitative fields.

Any condition that drops the polar check either:
- Removes `x` from `groupBy` in cartesian bar/area charts (breaking stacked bars), or
- Adds `radius` to `groupBy` in polar charts (breaking pie/donut layouts)

The root cause: in cartesian stacking, the dimension axis defines groups that accumulate. In polar stacking, the dimension (radius) encodes magnitude, not a grouping category. This asymmetry is inherent to the coordinate systems.

</details>

<details>
<summary>Should arc be split into arc + radial_bar?</summary>

`arc` does dual duty as the polar analog of both `rect` (positioned shape) and `bar` (stackable interval). In cartesian coords these are separate marks. Should we split arc too?

**No.** Unlike the rect/bar split (which is justified by differences in orientation semantics, size channel meaning, and corner radius config), arc has no encoding-level differences that would distinguish the two roles. A `radial_bar` mark would still need the same polar coordinate logic in `stack.ts`, so the split adds API surface and maintenance burden (8-12 files) without removing the underlying complexity.

**Cross-framework precedent:** No major grammar-of-graphics framework has cleanly solved this. ggplot2 uses `coord_polar()` as a coordinate transform (theoretically pure but practically limiting, which led ggforce to add `geom_arc_bar()`). D3 keeps arc/pie/stack as composable primitives with no smart defaults. Observable Plot hasn't shipped arcs yet. Vega-Lite's hybrid is a pragmatic shortcut unique to its "smart defaults" philosophy. Splitting the mark wouldn't match any established precedent.

</details>

PR #9519, opened 2025-01-30
chore(install): allow install from GitHub sha

n8agrin:
## PR Description

Allows installing from github link by adding a prepare step


n8agrin:
Wrong repository

PR #9483, opened 2024-11-30
fix: correct y axis title in boxplot when mark.extent isnt set ##9223

MarcBaeuerle:
closes #9223 

<details>
  <summary><h2>Checklist</h2></summary>

- [x] This PR is atomic (i.e., it fixes one issue at a time).
- [x] The title is a concise [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties").
- [x] `yarn test` runs successfully

PR #9526, opened 2025-02-11
fix: correct y axis title in boxplot when mark.extent isnt set

domoritz:
Copy of https://github.com/vega/vega-lite/pull/9483 so we build the examples.


domoritz:
https://github.com/vega/vega-lite/pull/9483 was merged

PR #9527, opened 2025-02-11
feat!: esm only package

domoritz:
Part of https://github.com/vega/vega/issues/3990


test-runtime/animation.test.ts
```diff
@@ -15,99 +19,93 @@ const gapminderSpec: TopLevelSpec = {
...
-  });
+const sleepTime = 1000 / fps;
 
+describe('time encoding animations', () => {
   it('renders a frame for each anim_value', async () => {
-    await embed(gapminderSpec);
+    const view = await embed(gapminderSpec, false);
 
-    await page.evaluate(setSignal('is_playing', false));
+    expect(await getSignal(view, 'is_playing')).toBe(false);
```
domoritz, line 62: @arvind @jonathanzong I could use some help here. Can you explain why this should be false? 

Running this is just `yarn vitest test-runtime/animation.test.ts`

jonathanzong: The intention here is to use `page.evaluate` to call `setSignal` to set the `is_playing` signal to false at runtime, i.e. pause the animation. We're doing this so that the subsequent calls to `await page.evaluate(setSignal('anim_clock', i * 500));` inside the for loop can control the time step of the animation, rather than timer events in the page that we can't control the timing of.

It's probably a good idea to add comments to this effect, sorry it wasn't clear. But I think the change here should be reverted unless there's something else causing an issue.

domoritz: Ohhhhh, I accidentally reaplced setSignal with getSignal. 🤦

domoritz: `page.evaluate` is gone btw which is why I can't just revert this. We can now just call methods directly. No need to go through strings anymore.

domoritz: Do we need the `await sleep(100);`? That seems to be an anti pattern for tests where we really don't want to artificially slow them down. Could we manually set the timer to a specific position rather than waiting?

jonathanzong: that makes sense about page.evaluate!

i don't remember why the `sleep(100)` is there. i think if the test passes without it then we should be able to get rid of it?

domoritz: I can't run the runtime tests in main and the animation tests are the last runtime tests not passing in this branch. I'll dig more into it.

What about the `await sleep(max_range_extent)` further down? Is that needed?

domoritz: I could use another pair of eyes on the tests here. I can't get them to pass and am really confused why we need to run them for so long. I think it's an anti pattern to require tests to run for a specific time rather than to mock the time somehow. Can you help me?

jonathanzong: those last two tests are testing whether the clock updates in real time and whether it loops at the end, which is why they use sleep. since this does make the tests take artificially long i think we could just remove these tests. they're really just testing whether timer events update the clock signal correctly, which is not covered by tests where we manually set the click signal but also is unlikely to not be working / we would notice immediately because examples wouldn't run

here's a diff of what i think the change would be. lmk if you want me to PR this or if you want to just do it here.

https://github.com/vega/vega-lite/pull/9540/commits/ad08b6da6a365d5ce2ba9c57eea02e211932c758

domoritz: I see. I think removing the tests makes sense. Alternatively, we could increase the fps to the point where the tests run super fast but relying on timing might still be tricky. 

Feel free to push directly to this branch. Does `yarn vitest test-runtime/animation.test.ts` pass for you?

jonathanzong: i made a commit but don't think i have push access to this repo.

also, i got the test to run properly but it's failing because it seems that the gapminder.json dataset has changed since i generated the correct test outputs? how often is this dataset going to change? https://github.com/vega/vega-datasets/commit/76feaab74a7f5f2bee8ede28ddcf865241681e64

domoritz: It shouldn't change but we also shouldn't have a copy of it. I can regen the examples.

jonathanzong: made you a PR into this branch here https://github.com/vega/vega-lite/pull/9542

manzt:
love to see it

domoritz:
❤️  me too. Now onto Vega!

PR #9537, opened 2025-02-19
test: show effect of checking for same dimension and stacked field in polar and cartesian fields

hydrosquall:
## Motivation

- Show whether we can improve arc mark stacking with same logic for polar and cartesian coordinates


domoritz:
@hydrosquall are you still working on this?

hydrosquall:
This is just a proof of concept to compare with
https://github.com/vega/vega-lite/pull/9512#issuecomment-3031977006 . I
plan to close it when/if the upstream is resolved.

Issue #4057, opened 2018-07-19
Consider dropping support for format and title in axis and legends for Vega Lite 3.0


Saba9:
Proposed by @domoritz. Allowing users to to define format and title in two locations seems confusing. 
One example of potentially confusing behavior is that when generating tool tips we use the encoding title and format and not the axis title and format which might not be straight forward for users.

kanitw:
FYI, currently there is only `format` for `text`, but not other channels. So I don't know if we want a breaking changes there, especially that the format are only for axis and legend labels, but have no effect on the data fields itself. 

For `title`, it will be also weird if there are properties for enabling parts of axis ( `ticks`, `labels`, `domain`, `grid`), but no `title` property.  So if we gonna change, a sensible change would be changing title to be boolean (so it's more consistent with `ticks`, `labels`, `grid`, `domain`).

domoritz:
I'm not so sure about positioning titles that way but do you agree that we should change `title` to only support booleans?

domoritz:
Any objections @arvind @jheer?

arvind:
Makes sense to me!

kanitw:
I don't know if we should really introduce this breaking changes given it's probably already widely used.  In fact, we even use `axis.title` in the figures for Making Data Visual. 

While we can of course go update the online gallery for making data visual, it seems like we don't gain much from this breaking change -- maybe a better thing to do is fix the bug in tooltip encoding to read axis/legend/header title too?

domoritz:
I think it's better to have only one way to set the title, rather than four (title, axis title, legend title, header title). We only used `axis.title` because that was the only way. Yes, it's a breaking change but we could add a simple preprocessor that copies `axis.title` to `title` when it's not just a boolean. I think now is a good time to change this.

kanitw:
> Yes, it's a breaking change but we could add a simple preprocessor that copies axis.title to title when it's not just a boolean. I think now is a good time to change this.

If you're adding a pre-processor, then it's not changing the behavior then.  Just how we implement it.

domoritz:
Well, and I'd not document it officially. Just to not annoy users too much.

kanitw:
> Well, and I'd not document it officially. Just to not annoy users too much.

I don't see how it's more annoying for users.  Having document that doesn't match behavior is weird. (I think having two ways of doing things is actually more annoying for us who maintain the projects.) 

[...]

PR #9541, opened 2025-02-20
fix: display title when set in config.axisX/axisY.title

### LINKED ISSUE #9429: `"config": {"axisX": {"title": "foo"}}` removes title
When defining the x-axis title using the general config object, as such:
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.20.1.json",
  "data": {"values": [{"a": 1, "b": 4}, {"a": 2, "b": 5}, {"a": 3, "b": 6}]},
  "mark": {"type": "line"},
  "encoding": {"x": {"field": "a"}, "y": {"field": "b"}},
  "config": {"axisX": {"title": "foo"}}
}
```
It is rendered like:
image

As can be seen, the x-axis title is removed completely.

Upon inspecting the compilation process, I noticed that the general axis configuration in Vega-Lite is directly mapped to a Vega config object. But, the Vega config object is essentially a subset of the available Vega-Lite configurations for axis properties. Since title is **not** defined as one of the axis properties in the Vega config ([source](https://vega.github.io/vega/docs/config/#axes)), but it is clearly defined as a [Vega-Lite axis property](https://vega.github.io/vega-lite/docs/axis.html#properties), leading to a mismatch.

I believe there are a couple of possible solutions:

- The VL schema could be adjusted to reflect that the general axis config object in Vega-Lite is a subset of the available axis properties.
- The compilation process from Vega-Lite to Vega could be improved or extended to ensure it does not directly map the general config object to Vega but also accounts for Vega-Lite-specific properties like title.

It's odd though, that Vega removes the axis title during rendering, even though it's not part of its schema.

MarcBaeuerle:
potentially closes #9429

This fixes the compilation process from vega lite to vega by accounting for the title property inside the config.
@domoritz suggested removing support for setting the axis title inside the config entirely, but due to the consensus in #4057 I chose this solution instead.

<details>

### DIFF
--- a/src/axis.ts
+++ b/src/axis.ts
@@ -16,7 +16,7 @@ import type {
 import {ConditionalPredicate, Value, ValueDef} from './channeldef';
 import {DateTime} from './datetime';
 import {ExprRef} from './expr';
-import {Guide, GuideEncodingEntry, TitleMixins, VlOnlyGuideConfig} from './guide';
+import {Guide, GuideEncodingEntry, TitleMixins} from './guide';
 import {Flag, keys} from './util';
 import {MapExcludeValueRefAndReplaceSignalWith, VgEncodeChannel} from './vega.schema';
 import {hasOwnProperty} from 'vega-util';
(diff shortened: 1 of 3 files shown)


domoritz:
I'm maybe missing a detail in the discussions but I don't see how we decided to allow setting axis titles in the config. I don't really want to allow this as it opens up too many places to set the title and configs are also for general properties that are valid across multiple charts (think themes) and then titles wouldn't really make sense there. 

Closing this for now as not planned but might change my mind if I missed something.

PR #9554, opened 2025-03-03
fix(docs): improve header alignment for `facet_bullet` graph example

hydrosquall:
## Motivation

In current spec, the labels are not lined up. Adding header label alignment will remove the unexpected staggered header labels.

## Changes

| Before | After |
| --- | --- |
| image |  image |

From what I can see from @kanitw's [original PR](https://github.com/vega/vega-lite/pull/3717), the "after" I'm proposing is the text alignment that was originally intended.

image

## Testing

- CI passing should be enough, this doesn't change any runtime behaviors.


domoritz (approved):
It's odd that we need to align left but seems to be what's required in Vega.

PR #9566, opened 2025-03-22
feat!: change default continuous size

### LINKED ISSUE #7464: Change default sizes
@jheer proposed to change the default sizes of Vega-Lite Charts. The idea would be that libraries don't have to change the default anymore. 

One option would be to use the default sized of [Vega-Lite API on Observable](https://observablehq.com/@vega/vega-lite-api-v5) and Altair. 

```
{continuousWidth: 400, continuousHeight: 300}
```

domoritz:
fixes #7464


joelostblom:
I think it is great that the charts are getting bigger by default! What do you think about using a square aspect ratio instead for the same reasons we switched to a square ration in altair as per https://github.com/vega/altair/issues/2581?

domoritz:
I don't have strong thoughts about the aspect ratio. I feel most charts make sense to be slightly wider but as a default square sounds reasonable as well. 

I think we should make Vega-Lite bigger by default because the original design was for Voyager and in most cases you want charts to be bigger. Also, screens have gotten larger in the last decade so we should go with the times.

domoritz:
@joelostblom switched to 300x300.

PR #9584, opened 2025-04-05
feat: export types

### LINKED ISSUE #9222: Export all spec typings for better TypeScript support
## Enhancement Description

*tl;dr: Vega-Lite should export all spec typings as a stable API for TS developers to use*

Vega-Lite currently only exports the `TopLevelSpec` union type for working with specs in TypeScript. [The docs here](https://vega.github.io/vega-lite/usage/typescript.html) show how to use this to write a static spec.

However, if you use TS to construct or modify a spec, you will quickly hit issues with type narrowing throughout your spec as Vega-Lite relies heavily on discriminated unions. This makes it very difficult to use unless you are willing to resort to using `any` as a workaround for these TS errors.

[See this example sandbox](https://codesandbox.io/p/sandbox/vega-lite-type-narrowing-lcz3g3?file=%2Fsrc%2Findex.ts) where trying to turn an axis off on an existing spec throws a type narrowing error.

```typescript
import { TopLevelSpec } from "vega-lite";

const spec: TopLevelSpec = {
  $schema: "https://vega.github.io/schema/vega-lite/v5.json",
  width: 100,
  height: 100,
  layer: [],
};

spec.layer = [
  {
    mark: {
      type: "bar",
    },
    encoding: {
      x: {
        field: "x",
        type: "nominal",
      },
      y: {
        field: "y",
        type: "quantitative",
      },
    },
  },
];

// Type narrowing error, can't disable axis
spec.layer[0]!.encoding!.y!.axis = null;
```

An issue about this has been opened before: https://github.com/vega/vega-lite/issues/8900. The resolution to that issue was a recommendation to import any needed types directly from the Vega-Lite source. There are 2 issues with this:
1. Developers may be using Vega-Lite in an environment where they only have a bundled Vega-Lite package + the exported types. With no access to src in such an environment, they can't access any of these typings.
2. Importing types this way forces a developer to take an unstable path that developers of Vega-Lite could break at any time

Since the Vega-Lite spec itself is a stable API, the typings for that spec should also be treated as a stable API. Otherwise, working with a Vega-Lite spec in TS beyond just a single spec definition typed to `TopLevelSpec` is painful and hacky.

## Checklist

### LINKED ISSUE #9583: Unable to import types in v6
## Bug Description

```ts
import type { NonNormalizedSpec } from 'vega-lite/build/src/spec';
```

 Pre v6, I was able to import types like `NonNormalizedSpec` from the spec file. Since v6 (and the switch to ESM), I'm no longer able to reach these types. It would be great if the `spec` file could get exposed as a module again so that I can migrate to v6.

domoritz:
Often, people need to access internal types. This pull request makes internal types available as an export yet we don't guarantee stability. 

Fixes #9583, fixes #9222

Example:

```ts
import type {TopLevelUnitSpec, Field} from 'vega-lite/types_unstable';

const spec: TopLevelUnitSpec<Field> = {
  $schema: 'https://vega.github.io/schema/vega-lite/v5.json',
  description: 'A simple bar chart with embedded data.',
  data: {
    values: [
      {a: 'A', b: 28},
      {a: 'B', b: 55},
      {a: 'C', b: 43},
    ],
  },
  mark: 'bar',
  encoding: {
    x: {field: 'a', type: 'ordinal'},
    y: {field: 'b', type: 'quantitative'},
  },
};
```


src/type.ts
```diff
@@ -18,13 +18,6 @@ export function isType(t: any): t is Type {
   return hasOwnProperty(Type, t);
 }
 
-export function isContinuous(type: Type): type is 'quantitative' | 'temporal' {
-  return type === 'quantitative' || type === 'temporal';
-}
-export function isDiscrete(type: Type): type is 'ordinal' | 'nominal' {
-  return type === 'ordinal' || type === 'nominal';
-}
-
```
domoritz, line 27: I replaced these with isDiscrete from https://github.com/vega/vega-lite/blob/11dea5338accc1ed81a9c9a3fbc41d5c6d82cc05/src/channeldef.ts#L870. The semantics are slightly different in that we also consider whether a channeldef is discretized via bin or geojson. I think this is correct but please check carefully.

monken:
I think `package.json` exports need to be updated too.

domoritz:
Ups, forgot to push those. Just pushed the additional export.

src/type.ts
hydrosquall: Is this change in functions required to be able to export types? If not, could we save this adjustment for a separate PR ?

src/types_unstable.ts
```diff
@@ -0,0 +1,54 @@
...
+export type * from './predicate.js';
+export type * from './projection.js';
+export type * from './resolve.js';
+export type * from './scale.js';
+export type * from './selection.js';
+export type * from './sort.js';
+export type * from './spec/base.js';
+export type * from './spec/concat.js';
+export type * from './spec/facet.js';
+export type * from './spec/index.js';
```
hydrosquall, line 43: To maintain parity with `5.23.0`, I think only 2 types have to be exported

- `Config`
- `TopLevelSpec`

https://github.com/vega/vega-lite/blob/cbad08391897b835e0a27956faef7a75afdae4c9/src/index.ts#L5-L7

I recognize we're adding more in order to address the fact that some teams were importing directly from the src folder before and bypassing the index.ts route.

While we could export the rest, if we're concerned about instability, we'll have an easier time if we err on the side of exporting fewer.

If we have multiple types that have the same name from any of these files, they may end up colliding in this file. It's not a huge deal while this file is marked unstable, but it's something we'll want to guard and have linting for before we make this API stable.

src/types_unstable.ts
```diff
@@ -0,0 +1,54 @@
+export type * from './aggregate.js';
```
hydrosquall, line 1: 1. Should we put a comment in this file to the effect of "Types exported from this file are considered unstable, and may be changed/removed at any time" ? 

2. How did you determine which files to export from? (I.e. when we add new files to Vega-Lite, what would drive us to update this file)? If we wanted to, we could probably prune this list down to a smaller number (maybe just the ones that are in the root or end with index.js). If the idea is that every file under `src` except itself should be listed, we could handle that with a comment first, and look into scripting/linting it later.

src/index.ts
```diff
@@ -6,3 +6,5 @@ export type {Config} from './config.js';
 export {normalize} from './normalize/index.js';
 export type {TopLevelSpec} from './spec/index.js';
 export * from './util.js';
+
+import './types_unstable.js';
```
hydrosquall, line 10: Do we need to have this import if we already have this in the package.json?

```
    "./types_unstable": {
      "types": "./build/types_unstable.d.ts"
    },
```

hydrosquall (commented):
The exported unstable types looks reasonable.

 I would advise against changing `isAggregateEncoding`, `isDiscrete`, `isContinuous`, `TEXT_MARK`, `isContinuous` in the same PR unless it's required since unlike the exports part that may have subtle impacts on runtime behavior, I'd like to understand the reasoning for combining those changes before adding ✅ .

domoritz:
It's required because we cannot export a flat namespace if we have name collisions.

hydrosquall:
Ah ok it's good to know that typescript will flag those collisions, I was worried they might be silently overwritten.

With that said, it's strange that exporting types is also exporting function definitions!

Would it be feasible to do the function replacements in a separate PR? It would be one thing if this was a pure rename, but it's a rename plus replacement in the isContinuous / isDiscrete case.

domoritz:
I can just rename the methods. It feels weird to have two different kinds of logic so I'd rather try to unify eventually and exporting types is not urgent.

src/index.ts
domoritz: I think I needed this for rollup to produce the file. It shouldn't change the default bundle.

src/types_unstable.ts
domoritz: We do keep the exports in index.ts, though. The unstable type exports are in addition.

domoritz: Yeah, something like `find . -type f -exec echo {} \;  | sed 's/\(.*\)/export type \* from \"\1\"/'` would be nice but it's too generic and creates conflicts.

domoritz:
Here is a simpler version that doesn't put all the types into the same namespace. It's a bit harder to use but also more predictable: https://github.com/vega/vega-lite/pull/9595.

PR #9585, opened 2025-04-07
docs: add link to Lace's videos

domoritz:
@lacepadilla does this look good to you?


domoritz:
Got an ok from Lace as well.

PR #9586, opened 2025-04-07
docs: update links and materials

domoritz:
Twitter is bad so replace links, add link to curriculum, remove mention of Vega-Lite 2

### DIFF
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -6,8 +6,6 @@ description:
 
 baseurl: '/vega-lite' # the subpath of your site, e.g. /blog/
 url: 'https://vega.github.io' # the base hostname & protocol for your site
-twitter:
-  username: vega_vis
 exclude:
   - Gemfile
   - Gemfile.lock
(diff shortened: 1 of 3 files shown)


domoritz:
@arvind @jheer @kanitw I replaced your twitter links with bsky or homepages. Happy to change it to something else you prefer.

site/_config.yml
```diff
@@ -6,8 +6,6 @@ description:
 
 baseurl: '/vega-lite' # the subpath of your site, e.g. /blog/
 url: 'https://vega.github.io' # the base hostname & protocol for your site
-twitter:
```
hydrosquall, line 9: It doesn't look like the site actually checked for this slug ( https://github.com/search?q=repo%3Avega%2Fvega-lite%20twitter&type=code ), so removing it without a bsky replacement seems fine.

hydrosquall (approved):
My profile link isn't impacted by this change, but the diff looks reasonable to me. 

I'm not sure if it still makes sense to list each of the individuals on the homepage vs pointing to the Vega governance group, but revising the list of people are actively in maintaining the Vega-Lite repo might be a better topic for a separate PR.

PR #9595, opened 2025-04-20
feat: export types

### LINKED ISSUE #9222: Export all spec typings for better TypeScript support

### LINKED ISSUE #9583: Unable to import types in v6

domoritz:
Fixes #9583, fixes #9222

Simpler version of #9584

You can use this new feature to import types from arbitrary files. 

```ts
import type {TopLevelUnitSpec} from 'vega-lite/types_unstable/spec/unit.d.ts';
import type {Field} from 'vega-lite/types_unstable/channeldef.d.ts';

const spec: TopLevelUnitSpec<Field> = {
  $schema: 'https://vega.github.io/schema/vega-lite/v5.json',
  description: 'A simple bar chart with embedded data.',
  data: {
    values: [
      {a: 'A', b: 28},
      {a: 'B', b: 55},
      {a: 'C', b: 43},
    ],
  },
  mark: 'bar',
  encoding: {
    x: {field: 'a', type: 'ordinal'},
    y: {field: 'b', type: 'quantitative'},
  },
};

```

### DIFF
--- a/src/compile/format.ts
+++ b/src/compile/format.ts
@@ -21,6 +21,7 @@ import {stringify} from '../util.js';
 import {isSignalRef} from '../vega.schema.js';
 import {TimeUnit} from './../timeunit.js';
 import {datumDefToExpr} from './mark/encode/valueref.js';
+import {BIN_RANGE_DELIMITER} from './common.js';
 
 export function isCustomFormatType(formatType: string) {
   return formatType && formatType !== 'number' && formatType !== 'time';
@@ -29,8 +30,6 @@ export function isCustomFormatType(formatType: string) {
 function customFormatExpr(formatType: string, field: string, format: Format) {
   return `${formatType}(${field}${format ? `, ${stringify(format)}` : ''})`;
 }
-
-export const BIN_RANGE_DELIMITER = ' \u2013 ';
 
 export function formatSignalRef({
   fieldOrDatumDef,
--- a/src/spec/index.ts
+++ b/src/spec/index.ts
@@ -15,7 +15,7 @@ import {TopLevel} from './toplevel.js';
 import {FacetedUnitSpec, GenericUnitSpec, NormalizedUnitSpec, TopLevelUnitSpec, UnitSpecWithFrame} from './unit.js';
 
 export type {BaseSpec, LayoutSizeMixins} from './base.js';
-export type {GenericHConcatSpec, GenericVConcatSpec, NormalizedConcatSpec} from './concat.js';
+export type {GenericConcatSpec, GenericHConcatSpec, GenericVConcatSpec, NormalizedConcatSpec} from './concat.js';
 export {isAnyConcatSpec, isHConcatSpec, isVConcatSpec} from './concat.js';
 export type {GenericFacetSpec, NormalizedFacetSpec} from './facet.js';
 export {isFacetSpec} from './facet.js';
@@ -24,7 +24,14 @@ export {isLayerSpec} from './layer.js';
 export type {RepeatSpec} from './repeat.js';
 export {isRepeatSpec} from './repeat.js';
 export type {TopLevel} from './toplevel.js';
-export type {FacetedUnitSpec, GenericUnitSpec, NormalizedUnitSpec, UnitSpec} from './unit.js';
+export type {
+  FacetedUnitSpec,
+  GenericUnitSpec,
+  NormalizedUnitSpec,
+  UnitSpec,
+  TopLevelUnitSpec,
+  UnitSpecWithFrame,
+} from './unit.js';
 export {isUnitSpec} from './unit.js';
 
 /**
--- a/package.json
+++ b/package.json
@@ -29,6 +29,9 @@
       "types": "./build/index.d.ts",
       "default": "./build/index.js"
     },
+    "./types_unstable/*": {
+      "types": "./build/*"
+    },
     "./vega-lite-schema.json": "./build/vega-lite-schema.json"
   },
   "unpkg": "build/vega-lite.min.js",


src/compile/format.ts
```diff
@@ -30,8 +31,6 @@ function customFormatExpr(formatType: string, field: string, format: Format) {
   return `${formatType}(${field}${format ? `, ${stringify(format)}` : ''})`;
 }
 
-export const BIN_RANGE_DELIMITER = ' \u2013 ';
```
hydrosquall, line 33: Nice to not have to export something that isn't reused on the outside 👍

src/spec/index.ts
```diff
@@ -15,7 +15,7 @@ import {TopLevel} from './toplevel.js';
 import {FacetedUnitSpec, GenericUnitSpec, NormalizedUnitSpec, TopLevelUnitSpec, UnitSpecWithFrame} from './unit.js';
 
 export type {BaseSpec, LayoutSizeMixins} from './base.js';
-export type {GenericHConcatSpec, GenericVConcatSpec, NormalizedConcatSpec} from './concat.js';
+export type {GenericConcatSpec, GenericHConcatSpec, GenericVConcatSpec, NormalizedConcatSpec} from './concat.js';
```
hydrosquall, line 18: Noting `GenericConcatSpec` is a new export, but that seems reasonable!

hydrosquall (approved):
Thanks for the change!

This will be easier to verify than the previous change, thanks for splitting it up.

### Maintenance

Going forward, what do you think would be a good place to document this api? Since it's unstable a MD file in vega might be enough, if it were stable API we would probably want to mention this in the vega-lite docs site.

## Build

The build preview error seems unrelated to this PR, and is likely something to investigate separately. I think it may be related to how vega-lite and the editor's build switched away from yarn, but the build preview is still using yarn.

```

00:51:30.948 | Executing user command: yarn run build-editor-preview
-- | --
00:51:31.248 | ! The local project doesn't define a 'packageManager' field. Corepack will now add one referencing yarn@3.6.3+sha512.d432ab8b22b89cb8b46649ebaf196159b0c1870904224771d75fdf8b55490aaa5de4d9f028deb0425f1513bff7843e97b9d117103795a23b12d2e7c1cedcf50f.
00:51:31.248 | ! For more details about this field, consult the documentation at https://nodejs.org/api/packages.html#packagemanager
00:51:31.248 |  
00:51:31.505 | Internal Error: vega-lite@workspace:.: This package doesn't seem to be present in your lockfile; run "yarn install" to update the lockfile
00:51:31.506 | at V0.getCandidates (/opt/buildhome/.cache/node/corepack/v1/yarn/3.6.3/yarn.js:435:5145)
00:51:31.506 | at kf.getCandidates (/opt/buildhome/.cache/node/corepack/v1/yarn/3.6.3/yarn.js:391:1264)
00:51:31.506 | at /opt/buildhome/.cache/node/corepack/v1/yarn/3.6.3/yarn.js:439:8033
00:51:31.506 | at df (/opt/buildhome/.cache/node/corepack/v1/yarn/3.6.3/yarn.js:390:11070)
00:51:31.506 | at ge (/opt/buildhome/.cache/node/corepack/v1/yarn/3.6.3/yarn.js:439:8013)
00:51:31.506 | at async Promise.allSettled (index 0)
00:51:31.506 | at async io (/opt/buildhome/.cache/node/corepack/v1/yarn/3.6.3/yarn.js:390:10398)
00:51:31.506 | at async /opt/buildhome/.cache/node/corepack/v1/yarn/3.6.3/yarn.js:439:8545
00:51:31.506 | at async ti.startProgressPromise (/opt/buildhome/.cache/node/corepack/v1/yarn/3.6.3/yarn.js:390:47579)
00:51:31.507 | at async je.resolveEverything (/opt/buildhome/.cache/node/corepack/v1/yarn/3.6.3/yarn.js:439:6573)
00:51:31.586 | Failed: Error while executing user command. Exited with error code: 1
00:51:31.596 | Failed: build command exited with code: 1
00:51:32.649 | Failed: error occurred while running build command
```

src/compile/format.ts
domoritz: We had this twice.

src/spec/index.ts
domoritz: Yeah

domoritz:
> Going forward, what do you think would be a good place to document this api? Since it's unstable a MD file in vega might be enough, if it were stable API we would probably want to mention this in the vega-lite docs site.

I think the package.json for now is enough. I suspect the people who need this will know where to look.

PR #8757, opened 2023-03-04
fix: ensure temporal field propagates to time expression

marr:
Fixes https://github.com/vega/vega-lite/issues/7185

PR #9622, opened 2025-06-16
fix: ensure temporal field propagates to time expression

domoritz:
Internal version of https://github.com/vega/vega-lite/pull/8757 so that build runs.


domoritz:
Confirmed it works with the original example. Code looks reasonable as well. 

Screenshot 2025-06-16 at 10 17 19

PR #9610, opened 2025-05-16
feat: Add new ternary chart

ajcappo:
🎯 Summary
This PR introduces a custom ternary diagram built with Vega-Lite, visualizing the distribution of population by socioeconomic level (High, Medium, Low) across major cities. The visualization is fictional and designed for illustrative or educational purposes.

📊 Features & Design Highlights
Ternary layout for multivariate analysis of proportions.

Each city is positioned based on the relative proportion of High, Medium, and Low socioeconomic levels.

Color coding reflects the dominant socioeconomic group per city using custom logic:

🔴 High > 80%

🟡 Medium > 80%

🟢 Low > 80%

Mixed conditions blend colors for nuanced distributions.

Size of points scales with the total population value (fictional).

Grid lines and triangle guides for better spatial understanding.

Dynamic tooltips provide raw values and percentage for each category.

Includes annotations and stylized labels for readability.

🛠️ Technical Details
Built using Vega-Lite v5 schema.

Utilizes layered marks (line, text, rule, point) and rich calculate transforms.

Designed with a dark theme and high-contrast colors for accessibility.

Font and spacing tuned for clarity and presentation.

Includes robust handling for tooltip formatting and label positioning.

📝 Notes
Data used is entirely fictional and created for demo purposes.

Author credit added in subtitle (Adrián Bazzana – LinkedIn).
[https://www.linkedin.com/in/adrian-bazzana/](url)

Encodes visual logic within the visualization layer using calculate expressions for reproducibility.

PR #9627, opened 2025-06-27
docs: add ternary example from #9610

domoritz:


dsmedia:
> "Population Distribution by Socioeconomic Level in Cities [Data shown is fictional]"

Small suggestion: fictional city names might be better here to avoid any mix-ups with real demographic data. Thanks for the contribution!

PR #9644, opened 2025-07-22
Fix:  align stack order only for nominal color domains

basbroek:
## PR Description

Refines [PR: fix: align stack order with color domain](https://github.com/vega/vega-lite/pull/9641):

- Only for specs with nominal color encoding the stack order is aligned with a custom color domain (if one is provided).
- The internal field that is use to order the stack is no longer added to aria-labels in the SVG

### DIFF
--- a/src/compile/mark/encode/aria.ts
+++ b/src/compile/mark/encode/aria.ts
@@ -77,6 +77,7 @@ export function description(model: UnitModel) {
   return {
     description: {
       signal: entries(data)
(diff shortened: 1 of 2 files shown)


domoritz:
Thanks. Moved to https://github.com/vega/vega-lite/pull/9645 so we get a build of the examples.

domoritz:
Thanks. Merged in https://github.com/vega/vega-lite/pull/9645.

Issue #5361, opened 2019-09-01
Change default quotes to `"`


domoritz:
I'm proposing to change the default quotes to `"` instead of the current `'`. The reason is that JSON quotes are always `"` and if we use the same quotes, we can easily use `'` to quote expressions that are just JSON strings and avoid having to escape with `\`. 

My hope is that it could clean up things like this:
<img width="640" alt="Screen Shot 2019-09-01 at 10 42 58 AM" src="https://user-images.githubusercontent.com/589034/64080168-d7944780-cca5-11e9-9dde-d17dcfa557f4.png">

Single quotes are better for sentences that use quotes themselves (`He then said "Hello World"`) but double quotes are better if you want to use ' (`Don't do that`). However, for our warnings, we often use \` anyway. 

Since we use prettier, this won't really be difficult to enforce. 

I am proposing to change the default here and in all other Vega libraries. For many projects, we can easily enforce this change via https://github.com/vega/vega-lite-dev-config/. 

Thoughts?

kanitw:
‘ is easier to type, no need to press shift.

I’m voting against this change.

arvind:
My preference would be to minimize the amount of churn unless absolutely necessary.

PR #9647, opened 2025-07-24
feat: support newlines in tooltips

### LINKED ISSUE #7564: Support newline characters or max width for tooltips
There are [a few suggestions of how to insert newlines in HTML tooltips in these SO answers](https://stackoverflow.com/questions/358874/how-can-i-use-a-carriage-return-in-a-html-tooltip?rq=1). The [two last Vega tooltip examples on this page](https://vega.github.io/vega-tooltip/vega-examples.html) also seem to include newline via different mechanisms, but I can't figure out a way to create a VegaLite specification with newlines in the tooltip.

I tried both a comma separated list which works for titles and the common control sequences from the SO post above like so `"this <br> is &#10; a &#13; really &#x0a; long &#xD; text \n with \r\n many \\n many \r words \\r."` Unless there is already a way to do this in Vega-Lite, I think it would be great to add support for newlines somehow.

I most commonly run into a need for this with long tooltips, so another partial solution would be a way to limit the width of the overall tooltip box manually.

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAFgAY2AacaYsiygAlMiRoRQBmDgF9p3OgENCCvCCgKAtphQhFygLRsARgEYAHADYYmFhICcdTBwVPMYO2wVszmOqbYArJgmIHIgGgoATgDWOgAOECTi3JhQkHQkSKgghBAQADbEcaqEAJ5x2shqEBokCvkg3PCY+XQ6hJgAHuJhnSXllSAAjqQKRHDKxGjaTVitOn1hpf0VOiNjxJNw040gzfNVy7LcACSUYIyYETpihHGUyAD0j9MICgB0CBOMpEbvcBBHudLhEXpg3vp8hNMC8WO8zO8TO8AFaUaC7PQKSiYQiUVSYwymSzWWwOJyeVzuTzeXz+IIhZAAbVAHW6KEZIAAyowIJFkjkuuIALrcPrIFjcZbICxyFmC9qMOCUAAEAB4jJEAHzKpXKgBkAGITGwANzKhT6o0SM2RTD1fKlS2dTxm-LQBBOgAiZtZhGVAB0oMrsN8A5FA8qIlBHf6I1GY5Fg7y6CrY5F3rsxQFJSgAOzSIWyIA)

basbroek:
## PR Description

This PR adds support for rendering newlines in tooltips and potentially closes #7564.

A new function has been added in the tooltip encoder that transforms tooltip values that are arrays into strings joined by newline characters. This allows multi-line tooltip content to be displayed correctly.

To support this behavior visually, a small CSS update is required in the vega-tooltip package — see the accompanying PR in that repository.

### Example
With the new code, this example spec ([open in editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAdxoATemgBMABh1K1STClShZSBHDQhjTQVaXEkghnAhoA2qACecJACdtHS0AZiUoNgZZTDQARj0lTDgADxj0ehoIAAJMrMkGCEwsuTgs4VkrAF8FHz9A1F0tABZwyOig-RsUtI8bGFzcpCzMNjZBTBocRRU6GGHlNjKaCvcAXWragKCtMVaotJCErtTPEAAVfuzB4dHxyenVemGYfzhS8rdpmUssgEcGERlORkLJIbLeSJZKBIWRZCBvEDrdZKZD+ADW1iYWyUcFkETUyzIaFAyWJIAAZjQ4II1NZfFtqiBvGTKdTaegIvtpphvDgrOh-jCJsYJqQQIyRmMJlNUF4KVSaXS6uKavK2dZOe0lOS2P5kGlphs1Yr0t1xatWoJdWSnC5+SAAMR6JoANgAnG7xZVKkA)) will render a tooltip like this:

example-multiline-tooltip

### DIFF
--- a/src/compile/mark/encode/tooltip.ts
+++ b/src/compile/mark/encode/tooltip.ts
@@ -16,7 +16,7 @@ import {Config} from '../../../config.js';
 import {Encoding, forEach} from '../../../encoding.js';
 import {StackProperties} from '../../../stack.js';
 import {Dict, entries} from '../../../util.js';
-import {isSignalRef} from '../../../vega.schema.js';
+import {isSignalRef, VgValueRef} from '../../../vega.schema.js';
 import {getMarkPropOrConfig} from '../../common.js';
 import {binFormatExpression, formatSignalRef} from '../../format.js';
 import {UnitModel} from '../../unit.js';
@@ -26,13 +26,14 @@ import {textRef} from './text.js';
 export function tooltip(model: UnitModel, opt: {reactiveGeom?: boolean} = {}) {
   const {encoding, markDef, config, stack} = model;
   const channelDef = encoding.tooltip;
+
   if (isArray(channelDef)) {
     return {tooltip: tooltipRefForEncoding({tooltip: channelDef}, stack, config, opt)};
   } else {
     const datum = opt.reactiveGeom ? 'datum.datum' : 'datum';
     const mainRefFn = (cDef: Encoding<string>['tooltip']) => {
       // use valueRef based on channelDef first
-      const tooltipRefFromChannelDef = textRef(cDef, config, datum);
+      const tooltipRefFromChannelDef = addLineBreaksToTooltip(cDef, config, datum);
       if (tooltipRefFromChannelDef) {
         return tooltipRefFromChannelDef;
       }
@@ -130,7 +131,7 @@ export function tooltipData(
       }).signal;
     }
 
-    value ??= textRef(fieldDef, formatConfig, expr).signal;
+    value ??= addLineBreaksToTooltip(fieldDef, formatConfig, expr).signal;
 
     tuples.push({channel, key, value});
   }
@@ -163,4 +164,24 @@ export function tooltipRefForEncoding(
 
   const keyValues = entries(data).map(([key, value]) => `"${key}": ${value}`);
   return keyValues.length > 0 ? {signal: `{${keyValues.join(', ')}}`} : undefined;
+}
+
+/**
+ * Transforms a tooltip value that is an array to a string with line breaks
+ */
+function addLineBreaksToTooltip(
+  channelDef: Encoding<string>['text' | 'tooltip'],
+  config: Config,
+  expr: 'datum' | 'datum.datum' = 'datum',
+): VgValueRef {
+  // tooltip fields with a format property are no strings
+  const fieldDefWithFormat = channelDef as {field: string; type: string; format: string};
+  if (fieldDefWithFormat?.type === 'nominal' && !fieldDefWithFormat.format) {
+    const fieldString = `datum["${fieldDefWithFormat.field}"]`;
+    return {
+      signal: `isValid(${fieldString}) ? isArray(${fieldString}) ? join(${fieldString}, '\\n') : ${fieldString} : ""+${fieldString}`,
+    };
+  }
+
+  return textRef(channelDef, config, expr);
 }
--- a/test/compile/mark/encode/tooltip.test.ts
+++ b/test/compile/mark/encode/tooltip.test.ts
@@ -137,7 +137,8 @@ describe('compile/mark/encode/tooltip', () => {
       });
       const props = tooltip(model);
       expect(props.tooltip).toEqual({
-        signal: '{"Foobar": isValid(datum["Foobar"]) ? datum["Foobar"] : ""+datum["Foobar"]}',
+        signal:
+          '{"Foobar": isValid(datum["Foobar"]) ? isArray(datum["Foobar"]) ? join(datum["Foobar"], \'\\n\') : datum["Foobar"] : ""+datum["Foobar"]}',
       });
     });
     it('generates correct keys and values for channels with title', () => {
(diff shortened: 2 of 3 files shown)


basbroek:
@domoritz: working on this issue I noticed signal strings created by Vega-Lite all use escaped double quotes `\"` while single quotes would also be valid and might enhance readability of the compiled Vega specs. For instance

`"signal": "\"a: \" + (isValid(datum[\"a\"]) ? datum[\"a\"] : \"\"+datum[\"a\"]) + \"; b: \" + (format(datum[\"b\"], \"\"))"`

would become 

`"signal": "'a: ' + (isValid(datum['a']) ? datum['a'] : ''+datum['a']) + '; b: ' + (format(datum['b'], ''))"`

Is there a reason for not using single quotes? Maybe I'm missing some edge-cases?

domoritz:
At least one reason would be that " are slightly easier to use with the default ' in the Vega-Lite code base. I suggested changing to " in https://github.com/vega/vega-lite/issues/5361. We could do it but should merge any big pending prs first. I'm supportive of it if you would like to tackle the task (in a separate pull request of course).

basbroek:
@domoritz  I'll be offline the next two weeks, but when I'm back, I would be happy to dig into the formatting of signal strings in a future pull request (after I have finished working on the [backgrounds for text marks](https://github.com/vega/vega-lite/pull/9646)).

src/compile/mark/encode/tooltip.ts
```diff
@@ -164,3 +165,23 @@ export function tooltipRefForEncoding(
...
+/**
+ * Transforms a tooltip value that is an array to a string with line breaks
+ */
+function addLineBreaksToTooltip(
+  channelDef: Encoding<string>['text' | 'tooltip'],
+  config: Config,
+  expr: 'datum' | 'datum.datum' = 'datum',
+): VgValueRef {
+  // tooltip fields with a format property are no strings
+  const fieldDefWithFormat = channelDef as {field: string; type: string; format: string};
```
domoritz, line 178: how do you know that the type here is `{field: string; type: string; format: string}`?

src/compile/mark/encode/tooltip.ts
```diff
@@ -164,3 +165,23 @@ export function tooltipRefForEncoding(
...
+  if (fieldDefWithFormat?.type === 'nominal' && !fieldDefWithFormat.format) {
+    const fieldString = `datum["${fieldDefWithFormat.field}"]`;
+    return {
+      signal: `isValid(${fieldString}) ? isArray(${fieldString}) ? join(${fieldString}, '\\n') : ${fieldString} : ""+${fieldString}`,
```
domoritz, line 182: Why do we need to convert invalid values to strings? Can we reuse logic here?

domoritz (commented):
I made a copy in https://github.com/vega/vega-lite/pull/9678 because this pull request has different images than the CI would output. Please use a PAT and try whether the CI builds the examples rather than manually building them.

src/compile/mark/encode/tooltip.ts
basbroek: I'm still wrestling a bit with the elaborate type system in this repo 😅
I have replaced the cast with a type guard.

src/compile/mark/encode/tooltip.ts
basbroek: This is in line with the way [formatSignalRef](https://github.com/vega/vega-lite/blob/f1052b6cadf57e866a9ab1054e2d7767d1258d7a/src/compile/format.ts#L132) handles nominal fields:

```return {signal: `isValid(${field}) ? ${field} : ""+${field}`};```

I think to render `null`, `undefined`, or `NaN` as strings in tooltips and area descriptions?

basbroek:
> Please use a PAT and try whether the CI builds the examples rather than manually building them.

I'm sorry for that. I did add an action secret and tried to trigger the CI by including `[SVG]` in the last commit message, but that didn't work. Hence I built the examples locally to pass all CI tests. Won't happen again.

domoritz:
All good. Our setup for external forks is a bit wonky because of some restrictions GitHub actions had (have?). I often make an internal copy to get the build to work correctly.

domoritz:
Is it true that if there is a format, the tooltip is not a string? It's a string afterwards but I guess it's not likely to be a multi line string, right?

basbroek:
>Is it true that if there is a format, the tooltip is not a string? It's a string afterwards but I guess it's not likely to be a multi line string, right?

The comment  was unclear, I have removed it. The logic is:
- Only tooltip fields with type `nominal` or `ordinal` have to be formatted as a vg-signal that checks for arrays with multi-line strings
- If the spec doesn't contain an explicit type for a `temporal` or `quantitative` tooltip field, it gets the default type `nominal` earlier in compiling
- If the tooltip field has type `nominal` but also a  format specifier, it should fall thru so it can be formatted as a vg-signal that includes the format (e.g. `"signal":  \"count\": format(datum[\"count\"], \",\")`

I hope this answers your question?

domoritz:
I merged your changes into https://github.com/vega/vega-lite/pull/9678.

domoritz:
We should make you a maintainer. Would you like to (https://github.com/vega/.github/blob/main/project-docs/MAINTAINERS.md)?

domoritz:
Thank you. Merged in https://github.com/vega/vega-lite/pull/9678/

basbroek:
@domoritz Thanks for reviewing and merging! If you have time, maybe you can have a look at the corresponding [PR in vega-tooltip](https://github.com/vega/vega-tooltip/pull/1075) that is necessary to render the newlines in the tooltip.

Issue #7798, opened 2021-11-12
Cannot set height to 0


domoritz:
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEzjQATjRyZ289AEEABBBoIcguIaZJ1h2DcyGA7nRiHETOMtXLDypJhUiioBKKigxEiCDGpoANqgYSD6wUxoAEwAHAC+ColoIABCqWhiYrn56ADCJagALADMFSBJACK1AJwAjM1JAKK1mT15LQUAYrViTSNJAOK1XR29BQASgwDsy+gAkpPp2QC6uRhwNGRYaAAMSsjqANYF1urBcLJQbMo0smRooAAevxAADMaHBBMoCiglJgAJ44OAFWRsBBfKLBJB-GgQQGCJCeQT6b4WK7ZY4wwEgsEQ9BpaFwhHoACODCQsh0gR0pBApOyQA)

Setting height to 0 is being ignored.

horvathaa:
Hi Dominik! After our chat yesterday, I took a stab at handling this bug. Here's what I've found...

The problem seems to be in two parts of the code. The first part is in the Unit model's constructor where the `size` property is set: https://github.com/vega/vega-lite/blob/81866c0fd2a692afc6acb9c0b3a40e86074abe51/src/compile/unit.ts#L111 

Since it is just checking for `spec.height`, the 0 values are ignored. Simply switching to `spec.height !== undefined` resolves this issue. 

The second part, downstream in the compilation during parsing, is when the size property is used in `parseUnitLayoutSize` and suffers from the same issue of not checking for `undefined` as opposed to a falsey value, thus the same fix applies: https://github.com/vega/vega-lite/blob/81866c0fd2a692afc6acb9c0b3a40e86074abe51/src/compile/layoutsize/parse.ts#L100

Implementing these fixes results in the behavior I think one would expect.
![Image](https://github.com/user-attachments/assets/11dfd69d-46ed-4ad8-ae5d-8a0116296312)

[...]

PR #9658, opened 2025-08-20
Fix setting height/width to 0

yuvvantalreja:
## PR Description

This PR fixes https://github.com/vega/vega-lite/issues/7798

The problem was that in three places where truthy/falsy checks were being used instead of explicit undefined checks:

### Expected Outcome:

Now when you test in the Vega Editor with height/width: 0, it should respect that setting instead of falling back to the default height of 300.

<details>

- [ ] This PR is atomic (i.e., it fixes one issue at a time).
- [ ] The title is a concise [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties").
- [ ] `npm test` runs successfully
- For new features:
  - [ ] Has unit tests.

Tips:

</details>

### DIFF
--- a/src/compile/layoutsize/assemble.ts
+++ b/src/compile/layoutsize/assemble.ts
@@ -20,7 +20,7 @@ export function assembleLayoutSignals(model: Model): NewSignal[] {
 export function sizeSignals(model: Model, sizeType: LayoutSizeType): (NewSignal | InitSignal)[] {
   const channel = sizeType === 'width' ? 'x' : 'y';
   const size = model.component.layoutSize.get(sizeType);
-  if (!size || size === 'merged') {
+  if (size === undefined || size === null || size === 'merged') {
     return [];
   }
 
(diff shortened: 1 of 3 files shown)


src/compile/layoutsize/assemble.ts
```diff
@@ -20,7 +20,7 @@ export function assembleLayoutSignals(model: Model): NewSignal[] {
 export function sizeSignals(model: Model, sizeType: LayoutSizeType): (NewSignal | InitSignal)[] {
   const channel = sizeType === 'width' ? 'x' : 'y';
   const size = model.component.layoutSize.get(sizeType);
-  if (!size || size === 'merged') {
+  if (size === undefined || size === null || size === 'merged') {
```
domoritz, line 23:
```suggestion
  if (size == null || size === 'merged') {
```

src/compile/layoutsize/parse.ts
```diff
@@ -97,7 +97,7 @@ export function parseUnitLayoutSize(model: UnitModel) {
   for (const channel of POSITION_SCALE_CHANNELS) {
     const sizeType = getSizeChannel(channel);
 
-    if (size[sizeType]) {
+    if (size[sizeType] !== undefined) {
```
domoritz, line 100: Should this be using `size[sizeType] != null` or should we not catch null and only catch undefined? How is this different from the logic above then?

PR #9660, opened 2025-08-25
fix: unescape the field (datum access path) correctly for binned time unit

yhoonkim:
## PR Description

fix the bug that doesn't work with the fieldname having the signle quote for binned time unit


test/compile/data/timeunit.test.ts
```diff
@@ -211,6 +211,24 @@ describe('compile/data/timeunit', () => {
       ]);
     });
 
+    it('should return the proper field escaping with binnedutcyearmonth', () => {
```
kanitw, line 214: I can't tell if this unit test is truly fixing an actual issue in a spec.

Please 

1. Add an example spec where this is broken. Push a commit with the example spec alone and wait for CI to compile a broken spec and generate broken screenshot.
2. Push this fix to the same branch (after having broken screenshots), so the spec is recompiled and show that your patches addresses the issue.

PR #9707, opened 2025-11-01
fix: tooltip data "undefined" when using nearest hit testing

michaelwallabi:
# Custom tooltip handlers receive string "undefined" instead of data values in v6.4.0

## Description

In vega-lite 6.4.0, custom tooltip handlers (set via `view.tooltip()`) receive objects where some  field values are the string `"undefined"` instead of the actual data values. This regression was introduced as part of: https://github.com/vega/vega-lite/pull/9678.

This PR includes a potential fix and some tests. This is my first time digging in to vega-lite details so very possible this isn't the right fix.

## Repro
[tooltip-no-repro-6.3.1.html](https://github.com/user-attachments/files/23279954/tooltip-no-repro-6.3.1.html)
[tooltip-repro-6.4.0.html](https://github.com/user-attachments/files/23279955/tooltip-repro-6.4.0.html)

## Expected Behavior

The tooltip handler should receive an object with actual data values:

```javascript
{
  date: "2024-01-01",
  revenue: 1200,
  category: "Product A"
}
```

## Actual Behavior (v6.4.0)

The tooltip handler receives an object where some values are the string `"undefined"`:

```javascript
{
  date: "2024-01-01",
  revenue: 1200,
  category: "undefined"
}
```

## Key Details

The issue occurs specifically with:
- Custom tooltip handlers set via `view.tooltip()`
- Use of hover parameters

The default Vega tooltip rendering appears to work correctly, but custom handlers receive corrupted data.


basbroek:
@michaelwallabi Thanks for finding this bug and creating such comprehensive test cases! As the author of the PR that seems to have introduced the bug I feel responsible and am eager to look into this together. 

Although it seems to fix the problem I'm not quite sure if returning `null` mitigates the root cause as this part of the VG-signal ``` \"\"+datum[\"category\"]``` is the same in the newLine PR (formatted in [this line](https://github.com/vega/vega-lite/blob/main/src/compile/mark/encode/tooltip.ts#L181)) as before (formatted in [here](https://github.com/vega/vega-lite/blob/main/src/compile/format.ts#L132)).

As you point out, the problem (only) seems to occur when using hover parameters. I think this is because the signal for the voronoi layer should reference `datum.datum.category` instead of `datum.category` (see this [modified version](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSAbJQBWENgDsQAGhBMkUANZkATmwbiAJmhAB3GHTjSQOJBo01xZNAFYZOmhvo2ADM5nwaZLC7cgImAE8AG0N0KDggoKMNJEwUVABtUHEkBFCMNlJlAH1-NmVDGUxlJHEIADN8hDQkkECcdKg2SLgoTCMxZXbUUHKaCK10bOIyCAdskABfSYBdSalk1PSxBmVw7OcjYiQghjgIGtAYzHSAJmdTgBYAWmcARlu7owLScT20O-PfKFjyfIDtAAFVQaBhtAAEAEFnuR2JJ0AA5fJOeZHX7ac5XR63U4w17vVB3AAcrhkPxOZH+QJBYMw4IAQjDaBJtABlNQohYgY5nC43e63ADMeLgb1Cd2spJA5L+ygB6GBbFBEOhMgKzPhIAAokh-FMuTyMXzsc5LiKxWhTncpTLKXLqUraeCAMJMuHaADq+3aqO56PQmP5D2c1nNBLuAHYbb87fKQIrlXTGWrYSzEciYPq0ScjViBc4BGHxQBOaMUqkKmkqt1pkDshgouZclJpbTHJAbDpqNbLbvrTZFEplSrKaqJUD1dIjhAMIIoGRwAAeOGU2kwbAAIr8ABTHBgIBIAHT9J2PMwAlEZdW30b7J9oHKLMDQ+nBVzJr0MRmMNBM7wEGm0PoghOd8QCXFdtG3GgIC3E5d1ifcjxPOAz3PcEAB8MPBaCIAANR2BwEMwJDjx5NDwQAMko8EYIAMQsAxtwAaj3A8yN+NDz3Q6jaPwwiNGI0iQBeUU9go3j6MY+DWMQ9iRLgfFUJAC8pjmGYZDGMgUiCA5xxAFt0nUOgth2AlgHmEA01qRSnz04w2AsUCEEyQoQAYHBDXQGCABUPJCbcVDUHBt248EAH5wSCjzQvBVBwWMn0m0WVt0BgVywI8ryQGIIJWQiVpnwkAAlfZmlIbdj3SrJcnXApjykcFj2MiQGvBYo9kajq4EvX1DO0aq32yEicBCIxrNAWzxEwPTahWHttGgNhAKKAD0hwRzprfDKjGQZQFH69BiHyCRHLUmQsv9E99yomiDAQULKD2hQnqQfbJ3BABCABeb7wQAcminB-tu2jNCXNhymgk4HvPV79soVQQka-6QjIUUNH+9CAB5wWcCLwWARL4uPNrhlGcZ4uhxBHpgvCTvERyCbYyg2Li8E2PPZDyZ-bIz0mdm3kiIwR3CNBut9KaZsOPw+3SYh+h0Ix73QDanO2htzvczyrqFqJZj6pYBoy4a2DIMgxpkbZdlCcodggOAZAm8DXml-T5rF9AlpWuo1u0NWttULJdreg6jaOhmztmC6dZzdApcoCB9HKTAAGk4ABSXXdm0APflxXlb91XNtAjktcuuOQDt3TDAN5tw4yGqXPMcoASd+EbOztBc88HTjZqkaxssiv0mbl8AkqxuhryerpHBQacnXc2QgJvX2YX4b-Md+eTaXi24FX2cgnZ7rGo3veV8i8+t8Fo-evUmRnpzgyG6fwvAK9gIECYZoOkCMbEjShoGsMaGkQDqxKG0GgpBxbKE6lXVQY5szxBPB2TYllRRNA0KEUAI9u5WRMFAOg8pQDWwJM4SgEZLLrmaM+HA+CtJ93QMADip40AJUwFAOiVRYhCXkuRFSjVjwAFImDgmERoRqwiACax5zxCIUkpY88Vpy8LYshUSYozwKLkQo20-xlF8QIkEIi6jjz6LlBRSKMFITKBKBPMx0oYwGJUuhSKwhNp8OQhYgI2iAaHkPOILG7NHE+LPOzUmIBZIkXkmElSkx9RVxoMLHofgfgAOlM0fIIt+hBEGE48s9pLJvRoEgYqzQ4Abn2FAZQNAcBFXhKQsyjRgFQCHjIRcDD0npE6TIV8eSbw5ksiQtJOx0htySQMbQmj3iWSTkgD+TSbbaCISA2u0xDapRypHGg791qxEzDICB8hnwwNQKfEAJT4jVwdn0xB+D2zaDfhg8QWCcHay8r0ZJURUlkPSMUUoEATAFGmok-wqgFBwA9A4JwvzmloAoYKawczihsEhfgv5a4hxArek+RJdMdkYvhecuBcBqFsFoXUhhvcdjaBYShQxDYuE8MwHw1mclkICI0k1EAojxGSPEbIkA8ieUzOUiollbLHFir8ZEkV5jnGWPYXTASUqOUKsKb41xBMbF2KQA4uS7KYneMVVqi8BMPEWDVcajVsozUowCUE9C8UWahNNeEkmIBjzRP3Ea4ScS5hTGmIOQF04ZYq22aoRmuyOn4IgqudArrDWLkwthdBMgRnxsGb6lmARU14zBTQAAXqEOaNKfm6BhZmX0jDaVpTgJ4bwBsH6jJCM-Q6IBem+w-u5Th0Q2DIAsA82IKD2ydj6bk-JhpLJDnRjUNwPdtJ1srY4atyUX5bImRGkx4g4BvX7YOxpqDBloJyVM9AYrEmztLYuphGAG1eB9FIZwYCS2qFgfA8QNBPYS3rlspoQRsmrR7fkcwfcZAaAHUgIdqTHmJpHeOyZAywimq7F0D9ZK1SlDnShzVWskCLn2DLaAYztBdvyP0UF6Af7YAHUYFQDhWTdO0BMhjgwLnPkUM6NQVHb3LvCMk7c9hV0EEuM4XqEGoMwZuY7EAc4mARD0jJj8tTrn21k8gRcWpFwnCo74BAFhtO6e6L4TjCglPqZkEWiw2DOmoGcDW5j6AJkUbxegEIKd6O1I0Ex0j6Au1sYw0Ub9ChuPqG6HxitAmgjbg8I+0T4nEmQcPWgZTcmkAKd0qlyzlzVPZZro-AjRm3P6cMzpkrwXFAWYKyAaz4M7MOa5CRjJ5HaludkGwWj1QZCBdQGl58mAMmGhkPJiIdFdhJyC+lzLAB5LIc56EkvgWZsLvG-DlpWQ2mLwn6AJYk7VmzS4EWOb8yAFzbWqNybgJ5nr3n8u3LqHQDJl6RsZYiHNt8C2psrZ4xF9bS6otbdiw+rAe3El1ds8dsBaMMbP2AlFrJYEIBfx-kEHyRcgFrOVk9xoqGFyvKVO85H39mh6VwbHd5y15DEKJcs+zlCNnTDmEAA) of the Vega spec of your example) but the code that generates that signal is never reached because of the shortcut in ```addLineBreaksToTooltip```. 

Do you agree? If so, any ideas for an elegant solution?

michaelwallabi:
Thanks for responding @basbroek. 

Agree, the fix here is just papering over a deeper issue. Appreicate the pointers, will take a look and see if I can come up with something better.

michaelwallabi:
Found a better/simpler fix.

Can't say I understand the difference between `datum` and `datum.datum` but this seems consistent with the intent of the `expr` arg. 

My understanding is that the key nuance here is that when using "nearest" tooltips a fully transparent voronoi diagram is used as a hit target for the actual marks. It's adding this voronoi layer that requires using `datum.datum` to get at the actual value. Is this indirection, voronoi -> actual mark, the extra lookup in the tooltip implied by the additional `.datum`, so without that it's looking up a value in the voronoi which doesn't exist?

domoritz (approved):

basbroek:
@michaelwallabi nice fix. 

> It's adding this voronoi layer that requires using `datum.datum` to get at the actual value. Is this indirection, voronoi -> actual mark, the extra lookup in the tooltip implied by the additional `.datum`, so without that it's looking up a value in the voronoi which doesn't exist?

You're right, the voronoi mark takes its data from the actual mark using [reactive geometry](https://vega.github.io/vega/docs/marks/#reactive-geometry.). The original data object is stored in a field `datum`. See also the data viewer in this example ([open in editor](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSAbJQBWENgDsQAGhAATONABONHJnaT0AQQAETJBBpRdSRdtgnM2uAA8kCHABs4U7QHc6MbcSQOGcbQ5ITHAOENoQMGyu4toMOBLa8TTimHCmkaSKlNIg7rL0aABMAAzFMvA0ZFhFpTI4SLKyyWRoAKwyskiYKKgA2qDidnBoIN1MTjnevgpo-SBQXeRsigCeI5o5dmwMKUUAHAC+UqALqWTLa+gAQpsI27uora1HJ4vnqyMAwrf3mGgALABmF7zN4XEYAER+Oz+qAAnABGEGnJYfdAAUWhDz2SOOoLO4PQADEsbDWsC8Sj3pcQABxUloBFw5FgtEgAASDNQewA7AcALoCmTQHwzPoDIYjawiiYyTArHDDdB6cSyHKyNjIZJoUCdbojMaykAAMxoITV6Cp4KOIEUSHEZCVuRo+RgOXqjWaaGKlGK7Vt9wtmEUfhBgwQTpWMuGHU1SG1qF1XR6oyCRtN5pGWxhIBt4kMSuDfhkdodToqVT+QpASGsYrmyzNDxATDY2E1OWjUq7IMbcGbTmNf2FCwm6Cjo+G1eQigA1hBZhKIyM9BA4ABZEzznLyxUjRRwKDDk2KDuJuTJg1pqcyftQNjyHUgfupRRP6xPrvoaWTnIZhwWviqJrDaeQFOeX4gD+oo5CqFq4iAlygJBE4wTI-6Admuw2ishSfr+45djIUx+N6Bw2nEepKqApoOA4T4kU6ECpCE4ykeR5RsJkT60fR56MfucBquRHGgLuTqpNYx7GqeCBPnqKarhuW4Ljad4PtRICUYs748WaAFSjkbDGsaa6wv8OF6Zm465nKNiwjR+mAXqDAIJQWFViJgpAA)).

michaelwallabi:
Anything else I should be doing for this change to be complete/merged? Looks like there's a pending check but not sure hot to make that go.

basbroek:
I think no further action on your part is required, @michaelwallabi. 

@domoritz; can you merge or do you want me to do so?

michaelwallabi:
Hello all. Friendly poke on this. Would to be able to upgrade to the latest vega-lite, but can't until this is fixed.

domoritz:
Thanks for the ping. Just got back from traveling to VIS in Europe. Looks good here!

PR #9714, opened 2025-11-04
feat(validate): include field and channel in invalid field type errors

didikana:
#This PR improves the invalidFieldType error messages by including both the field and channel information in the output.
Adds optional opts parameter ({ field?, channel? }) to log.message.invalidFieldType.
Updates relevant call sites (channeldef.ts, scale/type.ts) to pass this context where available.
Adds unit test test/error-messages.type.invalid.test.ts verifying that invalid type errors mention both the field and channel names.
Example new error message:
Invalid field type "bool" for field "x" on channel "x".

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1480,6 +1480,7 @@
       "integrity": "sha512-vvmsN0r7rguA+FySiCsbaTTobSftpIDIpPW81trAmsv9TGxg3YCujAxRYp/Uy8xmDgYCzzgulG62H7KYUFmeIg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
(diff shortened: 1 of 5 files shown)


hydrosquall:
Hey @didikana , it looks like this change may have introduced a few `any` types as well as undefined variables (def). Would you be able to make the updates so that the test can pass without bypassing the typechecker? 

Thanks for your interest in improving the specificity of the error messages! Curious if you have some context about what inspired this change.

didikana:
Closing this stale PR for now because it is still unmergeable and the current diff includes unrelated lockfile churn plus unresolved review feedback about `any` usage and undefined variables. The idea can be revisited in a cleaner, targeted PR that preserves type safety and avoids package-lock changes.

PR #9735, opened 2026-01-01
fix: check tooltip for custom formatters

basbroek:
## PR Description

This PR aims to fix this bug: [Vega displays the raw values on tooltip for ordinal or nominal type](https://github.com/vega/vega/issues/4204)

If a channelDef contains any kind formatting, including custom formatters, the `addLineBreaksToTooltip` function passes it to `textRef` and `formatSignalRef`.


.github/workflows/test.yml
```diff
@@ -80,7 +80,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-13, windows-latest]
```
domoritz: I already changed this on main to be latest.

src/compile/mark/encode/tooltip.ts
```diff
@@ -175,9 +175,13 @@ function addLineBreaksToTooltip(
   config: Config,
   expr: 'datum' | 'datum.datum' = 'datum',
 ): VgValueRef {
-  if (isFieldDef(channelDef) && isDiscrete(channelDef.type) && !hasProperty(channelDef, 'format')) {
+  if (
```
domoritz: could you add a unit test?

basbroek:
Thanks for reviewing. Maybe we can do a release after merging? That would solve both this and the previous [tooltip bug](https://github.com/vega/vega-lite/pull/9707).

domoritz:
I don't have cycles to make a release right now but we will make one eventually with this patch.

PR #9779, opened 2026-03-09
docs: document multiple condition array form

### LINKED ISSUE #9454: [documentation] Document support for multiple conditions, including support for "aggregate"
## Motivation

- The "array" form of conditions was added in https://github.com/vega/vega-lite/issues/3040 . However, none of the examples on this page use this format for showing conditions in action.
- Related bug:  conditions are converted from single to array form, the JSONSchema [does not like it](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygK4CcA29zQgegMYWQGcA6AK0IHsA7EAXwBoQAHAiAW0PgG1Qr2BTDEyidGhfqn64oGAJZUo-ZADcI6WgF1GbAgGsMyKTMb8quCmHkBzeKACetkADNZEsBgDyyWVfl1GAB6OLm4YAMJ2qPJgSpwMIOaoFMiO5lSWULLU3KAsyOxCIiCMEFZWhlbQgnAJFIgKdFogqqiI1SDl-A60PUA), even though the [object form is ok](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtCrVICUJNohSZ8gL5KcppAghoA2qFkO4+nJidKIcQTgoTH0aWUw4bWIkPSsAXSVkbQBrfW0gkKU4WSg2ZTCyNFAATyKQADMaAOV9AHltGlprJQAPMsrq-QBhYuFZZUinGxBcwTZtMtz+unZ5TRA7bQcvH0UQJDIydIoI-VyGcJBh6MEGT3QtuFKrG6A)

Image

## Suggested fixes

1. Add an array example to `condition` page
2. Look into why the `aggregate` field is marked as invalid per the vega-lite schema

didikana:
Fixes #9454.

## Summary
- document that `condition` can be a single object or an ordered array of condition objects
- add examples that show condition precedence for selected and hovered marks
- fix two `fall within` typos in the condition docs

## Tests
- `npx vitest run examples/examples.test.ts -t "selection_multi_condition|interactive_bar_select_highlight"`
- `npm run lint`

### DIFF
--- a/site/docs/encoding/condition.md
+++ b/site/docs/encoding/condition.md
@@ -30,9 +30,11 @@ permalink: /docs/condition.html
 }
 ```
 
-For [mark property channels](encoding.html#mark-prop) as well as [text and tooltip channels](encoding.html#text), the `condition` property of their channel definitions can be used to determine encoding rules based on whether data values satisfy a [parameter](parameter.html) (e.g. fall withing a [selection parameter](parameter.html#select)) or satisfy a `test` predicate.
+For [mark property channels](encoding.html#mark-prop) as well as [text and tooltip channels](encoding.html#text), the `condition` property of their channel definitions can be used to determine encoding rules based on whether data values satisfy a [parameter](parameter.html) (e.g. fall within a [selection parameter](parameter.html#select)) or satisfy a `test` predicate.
 
 {:#condition}
+
+The `condition` property can be either a **single** condition object or an **array** of condition objects. When using an array, conditions are evaluated in order; the first matching condition determines the encoding for that mark.
 
 There are two ways to specify the condition:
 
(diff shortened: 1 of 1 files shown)


site/docs/encoding/condition.md
```diff
@@ -119,10 +121,24 @@ For example, in the following plot, the color of `rect` marks is driven by a con
 }
 ```
 
-Condition value definitions and conditional datum definitions use a constant value encoding when data satisfy a [parameter](parameter.html) (e.g. fall withing a [selection parameter](parameter.html#select)) or satisfy a logical predicate. Another field, datum, or value definition can be specified as the "else" case when the condition is not satisfied.
+Condition value definitions and conditional datum definitions use a constant value encoding when data satisfy a [parameter](parameter.html) (e.g. fall within a [selection parameter](parameter.html#select)) or satisfy a logical predicate. Another field, datum, or value definition can be specified as the "else" case when the condition is not satisfied.
 
 A condition value/datum definition must contain either [a `param` name or a `test` predicate](#condition) in addition to the encoded constant [`value`](encoding.html#value-def) or constant data value ([`datum`](encoding.html#datum-def)).
 
+{:#multiple-conditions}
```
domoritz, line 128: Could you move the whole {:#multiple-conditions} block down to just after the selection_type_interval_invert example (right before "Besides specifying param name, we can also specify a test condition.")? That keeps the value/datum example sequence intact and gives the array-form material its own clean home before the test-condition discussion begins. As a bonus, moving it also removes the three-in-a-row "For example" openings that the current placement creates.

domoritz (approved):

PR #9784, opened 2026-03-19
feat: add treemap mark

kanitw:
## Summary

Adds a new `"mark": "treemap"` mark type for hierarchical data visualization.

- Introduces a `"hierarchy"` encoding channel that specifies how flat data is converted into a tree (supports both `key`/`parentKey` stratify and `nest` grouping modes).
- Treemap mark compiles to Vega `rect` marks with automatic `treemap` layout transforms. Position channels (`x`, `y`, `x2`, `y2`) are computed by the layout; `size`, `color`, and `tooltip` are user-specified.
- Layout props (`method`, `padding`, `ratio`, `round`, `nodes`) are configurable via mark properties.

### Key examples

**1. [`treemap_simple`](https://github.com/vega/vega-lite/blob/feat/treemap/examples/specs/treemap_simple.vl.json)** — A treemap of the Flare class hierarchy using `key`/`parentKey` stratify. Shows the core use case: ~25 lines of Vega-Lite vs ~60+ lines of equivalent Vega.

treemap_simple

**2. [`treemap_nest`](https://github.com/vega/vega-lite/blob/feat/treemap/examples/specs/treemap_nest.vl.json)** — A treemap of the Cars dataset grouped by Origin and Cylinders using the `nest` mode. Demonstrates that ordinary tabular data (no explicit parent-child IDs) can be turned into a treemap with a single encoding.

treemap_nest

**3. [`treemap_layered`](https://github.com/vega/vega-lite/blob/feat/treemap/examples/specs/treemap_layered.vl.json)** — A layered treemap where internal nodes render as colored category backgrounds and leaf nodes render as white-stroked cells with tooltips. Shows compositional power: hierarchy encoding is specified once in the outer spec and inherited by both layers.

treemap_layered

## Test plan

- [x] Unit tests for treemap dataflow node (`test/compile/data/treemap.test.ts`)
- [x] Unit tests for treemap mark compiler (`test/compile/mark/treemap.test.ts`)
- [x] Unit tests for hierarchy inheritance in layers (`test/compile/mark/treemap_inherit.test.ts`)
- [x] Visual verification of example specs
- [x] Review compiled Vega output for correctness

### DIFF
--- a/examples/compiled/treemap_layered.vg.json
+++ b/examples/compiled/treemap_layered.vg.json
@@ -15,7 +15,6 @@
         {"type": "stratify", "key": "id", "parentKey": "parent"},
         {
           "type": "treemap",
-          "method": "squarify",
           "field": "size",
           "size": [{"signal": "width"}, {"signal": "height"}],
           "round": true
@@ -30,7 +29,6 @@
         {"type": "stratify", "key": "id", "parentKey": "parent"},
         {
           "type": "treemap",
-          "method": "squarify",
           "field": "size",
           "size": [{"signal": "width"}, {"signal": "height"}],
           "round": true
--- a/examples/compiled/treemap_nest.vg.json
+++ b/examples/compiled/treemap_nest.vg.json
@@ -15,7 +15,6 @@
         {"type": "nest", "keys": ["Origin", "Cylinders"]},
         {
           "type": "treemap",
-          "method": "squarify",
           "field": "Horsepower",
           "size": [{"signal": "width"}, {"signal": "height"}]
         },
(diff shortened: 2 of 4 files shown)


kanitw:
> Only one worried thing (?) is if the VL spec allows to layer the treemap mark and non-treemap mark. (e.g., just bar chart). Wonder if this case safely fails.


Technically I'm replicating the same design as maps. layering map with cartesian chart will similarly not working.

kanitw:
>  As far as I know, it's the first(?) macro channel. It serves two things


Not really the first.  Technically bar stacking and geoshape also outputs `"transform": [{"type": "geoshape", "projection": "projection"}]` automatically.

kanitw:
> Also I believe, we can use this "hierarchy" channel for the other marktype like network diagram in the future as well. The implementation seems decoupled enough.

hierarchy should be specific to trees, so I expect we can use this with other tree viz like sunburst  (but not arbitrarily networks).

PR #9819, opened 2026-04-09
fix: Ungroup pathmarks when used in offset channels

joelostblom:
## PR Description

Previously, pathmarks like line/trail/area, would be grouped across the y/x encoding rather than creating a separate pathmark per category in the encoding in the main channel (e.g. the y-encoding if yOffset is used).

For example, take the following spec:

```json
{
  "data": {
    "values": [
      {"a": "A", "b": 28, "c": "x", "d": "m"},
      {"a": "B", "b": 55, "c": "x", "d": "m"},
      {"a": "C", "b": 43, "c": "x", "d": "n"},
      {"a": "D", "b": 91, "c": "x", "d": "n"},
      {"a": "A", "b": 88, "c": "y", "d": "m"},
      {"a": "B", "b": 55, "c": "y", "d": "m"},
      {"a": "C", "b": 43, "c": "y", "d": "n"},
      {"a": "D", "b": 55, "c": "y", "d": "n"},


    ]
  },
  "mark": {"type": "trail"},
  "encoding": {
    "x": {"field": "a"},
    "yOffset": {"field": "b", "type": "quantitative"},
    "y": {"field": "c"},
  }
}
```

Before this PR, it would generate a nonsensical chart:
image

If `"detail": {"field": "c"}` was added to the spec above, it would instead generate this more useful chart with with "spark lines":
image

The drawback with having to use `details` is that it is difficult to discover and that it is not longer possible to group by another field in the `details` encoding. After this PR, the spark lines chart is generated without the need to use details which leaves it open for another grouping.

<details>

- For new features:
  - [x] Has unit tests.

### DIFF
--- a/src/encoding.ts
+++ b/src/encoding.ts
@@ -728,7 +728,23 @@ export function pathGroupingFields(mark: Mark, encoding: Encoding<string>): stri
       case X2:
       case Y2:
       case XOFFSET:
-      case YOFFSET:
+      case YOFFSET: {
+        if (mark === 'line' || mark === 'area' || mark === 'trail') {
+          const offsetDef = encoding[channel];
+          if (isFieldDef(offsetDef)) {
+            const mainChannel = getMainChannelFromOffsetChannel(channel as typeof XOFFSET | typeof YOFFSET);
+            const mainDef = encoding[mainChannel];
+            if (isFieldDef(mainDef) && !mainDef.aggregate && !offsetDef.aggregate) {
+              const mainField = vgField(mainDef, {});
+              const offsetField = vgField(offsetDef, {});
+              if (mainField && offsetField && mainField !== offsetField) {
+                details.push(mainField);
+              }
+            }
+          }
+        }
+        return details;
+      }
       case THETA:
       case THETA2:
       case RADIUS:
--- a/src/encoding.ts
+++ b/src/encoding.ts
@@ -727,12 +727,14 @@ export function pathGroupingFields(mark: Mark, encoding: Encoding<string>): stri
       case URL:
       case X2:
       case Y2:
+        return details;
+
       case XOFFSET:
       case YOFFSET: {
         if (mark === 'line' || mark === 'area' || mark === 'trail') {
           const offsetDef = encoding[channel];
           if (isFieldDef(offsetDef)) {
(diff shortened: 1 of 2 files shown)


domoritz (approved):
Makes sense. Another option is to use a facet encoding.

```json
{
  "data": {
    "values": [
      {"a": "A", "b": 28, "c": "x", "d": "m"},
      {"a": "B", "b": 55, "c": "x", "d": "m"},
      {"a": "C", "b": 43, "c": "x", "d": "n"},
      {"a": "D", "b": 91, "c": "x", "d": "n"},
      {"a": "A", "b": 88, "c": "y", "d": "m"},
      {"a": "B", "b": 55, "c": "y", "d": "m"},
      {"a": "C", "b": 43, "c": "y", "d": "n"},
      {"a": "D", "b": 55, "c": "y", "d": "n"},


    ]
  },
  "mark": {"type": "trail"},
  "encoding": {
    "x": {"field": "a"},
    "row": {"field": "c"}
  }
}
```

joelostblom:
Thanks for the quick review! Yes, this could definitely be done with faceting as well. My thinking here is that this offset channel approach is a more lightweight alternative to faceting when a spark-line like chart is desired, and it provides an additional dimension that frees up faceting for another level of categorical nesting if desired.

I am trying to do some general work on extending some grammatical concepts to offset encodings so that they behave similarly to other encodings, which feels satisfying  to me =) A few PRs down the road, I'm hoping this might help with visualizing multiple distributions as well, but I have yet to see if that works out!

domoritz:
Hmm, think I'm about this more, maybe the original behavior was actually better. What if you do want to draw a single line to connect marks across categorical encodings? Splitting is definitely a reasonable default but then you cannot undo it in the spec, right? That makes me think the original behavior was maybe better since it provides more flexibility. 

Iirc, you can add an array of detail encodings so it's possible to split by more than one variable.

joelostblom:
That's still possible, the new behavior only triggers if there is an offset channel in use:
image

Leaving out the offset channel still connects the line across categories, which I think is the use case you are referring to:
image

domoritz:
I'm thinking even with an offset channel, you might want to connect. Imagine a chart where we want to show some connection between two series of values. It's doesn't work across faceted charts but could be done with offset. Detail makes is exactly for this use case where you want to break a series but not add some other discrete channel like color or shape.

PR #9817, opened 2026-04-09
fix: Reverse `yOffset` directionality for quantitative scales to match `y`

joelostblom:
Close #9816, details and motivation can be found in that issue.

This PR makes the `yOffset` encoding behave the same as the regular `y` encoding for quantitative scales, which I think is the expected behavior.

PR #9823, opened 2026-04-10
fix: Make ranged marks compatible with offset channels

joelostblom:
In short, this PR changes ranged marks in offset channels so that they compile as true ranged marks with consistent baseline, grouping, and axis alignment. There are still some things to fix with the examples but it would be helpful with high level comments on the overall approach even before then. (Note that all the screenshot are including https://github.com/vega/vega-lite/pull/9817, otherwise they would be upside down)

### Details

Currently, ranged marks are not working well in offset channels. For example, a bar mark simply places a rectangle of fixed size at the location of the offset. I think that the desired behavior should mimic how bars work across other quantitative channel, ie start from a baseline and extent to the data value, which is what I have tried to implement here:

| Before this PR | After this PR |
|-|-|
| image | image |

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygG4QDYFcCmBneBtUscIAgiADQgBG8ATABwUDG8IAHueCwLYgC+ZBFgCEO1OAFZxTFuwphufASEIgAwqPgAWAMzSiszkQB2iwUQAiGuAE4AjHrYd5x08pakKYugxDMiATycFfjMQEU94SQdAuWClFXUIuB1ooJcQtwsrKN8WGMMQE14AXX4QLggAJwBreFAofwAHDBZKKsUQDCNGAHswAEsjAHM6xwQQADN+jBRnNzL-AHkJiawMKFGpmbnqCgbmlgBHNAgjKH6Yc8QWhc3p2ZZmXmegA)

Similarly, area marks display unusual behavior currently and don't show up at all. With the changes in this PR, they will show similar to how the line mark from #9819 display:

| Before this PR | After this PR | Line in #9819 |
|-|-|-|
| image | image | image |

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygG4QDYFcCmBneBtUscIAgiADQgBG8ATABwUDG8IAHueCwLYgC+ZBFgCEO1OAFZxTFuwphufASEIgAwqPgAWAMzSiszkQB2iwUQAiGuAE4AjHrYd5x08pakKYugxDMiATycFfjMQEU94SQdAuWClFXUIuB1ooJcQtwsrKN8WGMMQE14AXX4QLggAJwBreFAofwAHDBYqjFgyjCNGAHswAEsjAHM6xwQQADN+jBRnNzL-AHkJiawMKFGpmbnqCgbmlgBHNAgjKH6Yc8QWhc3p2ZZmXmegA)

I also chose to align axis/tick/label behavior with the new baseline semantics, anchoring them at the bottom of the bar/area, because I thought it looked odd when they were in the middle of the band:
image

I think that a case could be made for including a grid line by default as a baseline, but I was unsure if this was too automagical, so I left it out in this implementation. It is quite helpful, particularly when the marks have different range, but maybe it should be added manually for consistency:

|Without gridline| With gridline |
|-|-|
| image | image |

<details>

- For new features:
  - [x] Has unit tests.

Tips:

</details>


examples/compiled/bar_axis_space_saving.png
domoritz, line 1: That's a regression

domoritz:
I wonder whether you still want this pull request. offset isn't really designed for ranged marks where maybe using row and column makes more sense. Now that we don't automatically group by the range mark (for reasons explained in the other prs), I wonder what you think about this pull request. 

The zero: true default applies to all quantitative offset scales, not just ranged marks. We definitely want to be careful with that before merging. 

```json
{
  "description": "Points spread within each band by a meaningful field",
  "height": {"step": 50},
  "mark": "point",
  "encoding": {
    "y": {"field": "Cylinders", "type": "ordinal"},
    "yOffset": {"field": "Miles_per_Gallon", "type": "quantitative"}
  }
}
```

On main, the yOffset domain fits the data (around [9, 47]), so points use the full bandwidth. With this PR the domain becomes [0, 47], compressing all points into the upper ~80% of each band and leaving the bottom fifth empty — and the further the data sits from zero, the worse the compression (a field with values 100 to 200 would squeeze everything into half the band). For random() jitter the change is invisible (domain is already around [0, 1]), but offset-as-secondary-encoding regresses.

domoritz:
I changed that zero is only used for ranged marks. This should avoid regressions while still enabling the feature you aimed for. What do you think?

joelostblom:
Thanks for having a look and for improving the PR! Good call on applying `zero: true` only for ranged marks; I think that makes more sense too.

In terms of this feature in general, it does seem grammatically intuitive to me that it is possible to use ranged marks in offset channels since it works for  non-ranged marks. I think the use case for this is faster access to more "sparkline"-like charts than with facetting (even if they can look similar with styling) and that this also gives an additional grouping dimension since it can be used with facetting. The ability to use density areas in offset channels could potentially make it easier to create e.g. ridgeline charts (I believe some additional tweak was needed to make densities work in offline channel, but I don't quite remember atm). All that to say that unless we see some harm/regressions from supporting this, I think it would be a "nice to have" (but not a "must have").

One observation: the requirement to specify the `detail` encoding for grouping the marks in their offset channel only applies to areas (and lines). Bar marks (and points) look the same with and without using the `detail` channel. This is in a sense expected since those marks are not connected, but it also means that it might not be obvious why a chart looks "broken" after changing the mark e.g. from bar to area, and it might also not be immediately clear that the fix is to add `detail`. I don't have a better suggestion though, and I think it is reasonable that these charts can take a bit extra thinking to create.

examples/compiled/bar_axis_space_saving.png
joelostblom: Resolving since your commits fixed this

domoritz (approved):

PR #9822, opened 2026-04-10
fix: Ungroup pathmarks when used in offset channels (revert #9819)

domoritz:
See https://github.com/vega/vega-lite/pull/9819#issuecomment-4223518214

PR #9821, opened 2026-04-10
fix: Ungroup pathmarks in offset channels also for aggregated fields

joelostblom:
## PR Description

I missed testing with aggregated field properly in #9819, so an aggregation in the offset channel currently connects pathmarks incorrectly:
image

After this PR, pathmarks are correctly ungrouped also for aggregated fields:
image

Spec for testing:

```json
{
  "data": {
    "values": [
      {"a": "A", "b": 28, "c": "x", "d": "m"},
      {"a": "B", "b": 55, "c": "x", "d": "m"},
      {"a": "C", "b": 43, "c": "x", "d": "n"},
      {"a": "D", "b": 91, "c": "x", "d": "n"},
      {"a": "A", "b": 88, "c": "y", "d": "m"},
      {"a": "B", "b": 55, "c": "y", "d": "m"},
      {"a": "C", "b": 43, "c": "y", "d": "n"},
      {"a": "D", "b": 55, "c": "y", "d": "n"},

    ]
  },
  "mark": {"type": "line"},
  "encoding": {
    "x": {"field": "a"},
    "yOffset": {"field": "b", "aggregate": "sum", "type": "quantitative"},
    "y": {"field": "c"}
  }
}
```

<details>
  <summary><h2>Checklist</h2></summary>

- [x] This PR is atomic (i.e., it fixes one issue at a time).
- [x] The title is a concise [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties").
- [x] `npm test` runs successfully
- For new features:
  - [x] Has unit tests.
  - [ ] Has documentation under `site/docs/` + examples.

Tips:

- https://medium.com/@greenberg/writing-pull-requests-your-coworkers-might-enjoy-reading-9d0307e93da3 is a nice article about writing a nice PR.
- Use draft PR for work in progress PRs / when you want early feedback (https://github.blog/2019-02-14-introducing-draft-pull-requests/).
</details>

PR #9826, opened 2026-04-11
feat: Control grouping of pathmarks via `groupByOffset` option

joelostblom:
Details in https://github.com/vega/vega-lite/pull/9822#issuecomment-4229202440

Setting `"offsetGroup": "none"` in the spec now creates the original chart before #9819:

image

```json
{
  "data": {
    "values": [
      {"a": "A", "b": 28, "c": "x", "d": "m"},
      {"a": "B", "b": 55, "c": "x", "d": "m"},
      {"a": "C", "b": 43, "c": "x", "d": "n"},
      {"a": "D", "b": 91, "c": "x", "d": "n"},
      {"a": "A", "b": 88, "c": "y", "d": "m"},
      {"a": "B", "b": 55, "c": "y", "d": "m"},
      {"a": "C", "b": 43, "c": "y", "d": "n"},
      {"a": "D", "b": 55, "c": "y", "d": "n"},


    ]
  },
  "mark": {"type": "line", "offsetGroup": "none"},
  "encoding": {
    "x": {"field": "a"},
    "yOffset": {"field": "b", "type": "quantitative"},
    "y": {"field": "c"},
  }
}
```

We could also add an example in the docs, but just putting this up first to show what the implementation would look like.

If we like this approach, I think we should wait to merge after https://github.com/vega/vega-lite/pull/9821, since this PR also contains that fix for testing purposes (just to have a clearer commit history and not two separate fixes coming from this one PR).

I think there is already some precedence for controlling behavior like this directly in the mark setting vs the encoding, such as the default stacking of some marks but not others (e.g. areas as per https://github.com/vega/vega-lite/pull/9018).

<details>

- For new features:
  - [x] Has unit tests.

Tips:

</details>

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -17107,6 +17107,14 @@
           ],
           "description": "The minimum band size for bar and rectangle marks. __Default value:__ `0.25`"
         },
+        "offsetGroup": {
+          "description": "Controls automatic path grouping when `xOffset`/`yOffset` is used with path marks (`line`, `area`, `trail`).\n\n- `\"auto\"` (default) adds grouping by the corresponding base channel (`x` or `y`) to keep paths confined within lanes.\n- `\"none\"` disables this automatic grouping so only explicitly encoded grouping channels (`detail`, `color`, etc.) are used.",
+          "enum": [
+            "auto",
+            "none"
+          ],
+          "type": "string"
+        },
         "opacity": {
           "anyOf": [
             {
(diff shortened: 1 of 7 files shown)


build/vega-lite-schema.json
```diff
@@ -17107,6 +17107,14 @@
           ],
           "description": "The minimum band size for bar and rectangle marks. __Default value:__ `0.25`"
         },
+        "offsetGroup": {
```
joelostblom: We could discuss the naming of the option and if `true` / `false` is more natural than `'auto'` `'none'`

domoritz:
Thanks for the pull request. I think for consistency with other APIs, we should use boolean/null. I pushed a commit in https://github.com/vega/vega-lite/pull/9826/changes/8b4605529a97fd5223f3c6187980de8ed8ba302f. @joelostblom.

I'll change the pr title to feat since this adds API.

However, I still don't think this is right since we have found cases where the default behavior of not grouping makes sense. So It's hard to justify changing the default behavior and introduce new API when we already have `detail`.

domoritz (approved):

domoritz (changes requested):
See my last comment.

domoritz:
Merging https://github.com/vega/vega-lite/pull/9822 for now but we can discuss further whether we want to introduce this option.

joelostblom:
Thanks for the comments, I agree with your comment regarding this maybe not being motivated enough, so I think I'm ok with us closing this in favor of relying on the `detail` encoding. We can always revisit if that ends up being not sufficient in some way.

PR #9827, opened 2026-04-12
fix(tick): respect bandSize in tick sizing

### LINKED ISSUE #9767: Tick Config exports `bandSize` which seemingly has no effect; should it be `size`?
## Bug Description

If you try the [example in the docs](https://vega.github.io/vega-lite/docs/tick.html#example-customizing-ticks-size-and-thickness), you will see that changing `bandSize` has no effect either for vertical or horizontal ticks. Changing the argument name to `size` instead will change the length of the tick as expected.

Is the wrong parameter name exported for `tickConfig`, so it should be `size` instead of `bandSize`? (and the docstring changed to `length` instead of `width`?

https://github.com/vega/vega-lite/blob/b9f3b1480df3a85f9cdce93a70b0ebf7e72f53e3/src/mark.ts#L671-L680

Adeyinka1:
### Summary
This PR fixes an issue where `bandSize` is defined in `TickConfig` but has no effect on tick marks. close https://github.com/vega/vega-lite/issues/9767

### Problem
Currently:
- `size` controls tick thickness correctly
- `bandSize` is ignored for tick marks

This creates inconsistency between configuration and rendering behavior.

### Solution
- Added fallback logic in `getBandSize()` for tick marks
- `bandSize` is used when `size` is not explicitly provided
- `size` continues to take precedence

### Changes
- Updated `src/channeldef.ts` to include `bandSize` fallback for tick marks
- Added tests to validate:
  - `bandSize` is respected when `size` is absent
  - `size` overrides `bandSize` when both are defined

### Testing
- Ran targeted tests:

### DIFF
--- a/site/docs/mark/tick.md
+++ b/site/docs/mark/tick.md
@@ -71,7 +71,7 @@ The `tick` property of the top-level [`config`](config.html) object sets the def
 
 Besides standard [mark config properties](mark.html#config), tick config can contain the following additional properties:
 
-{% include table.html props="bandSize,thickness" source="TickConfig" %}
+{% include table.html props="thickness" source="TickConfig" %}
 
 #### Example Customizing Tick's Size and Thickness
 
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -581,13 +581,6 @@ export function getBandSize({
     return size;
   }
 
-  if (mark.type === 'tick') {
-    const tickBandSize = getMarkPropOrConfig('bandSize', mark, config);
-    if (tickBandSize !== undefined) {
-      return tickBandSize;
-    }
-  }
-
   if (isFieldDef(fieldDef)) {
     const {timeUnit, bin} = fieldDef;
 
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -581,6 +581,13 @@ export function getBandSize({
     return size;
   }
 
+  if (mark.type === 'tick') {
+    const tickBandSize = getMarkPropOrConfig('bandSize', mark, config);
+    if (tickBandSize !== undefined) {
+      return tickBandSize;
+    }
+  }
+
   if (isFieldDef(fieldDef)) {
     const {timeUnit, bin} = fieldDef;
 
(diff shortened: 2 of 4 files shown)


Adeyinka1:
Thanks for taking a look. I’ve implemented the `bandSize` fallback for tick sizing so that `size` still takes precedence, and I added targeted tests for both cases. Happy to adjust the approach if you’d prefer this handled differently.

joelostblom:
Thanks for making the PR! I think this could make sense technically to restore the effect of `bandSize`. However, I wonder if the more appropriate fix is not to change the documentation to advertise size instead and remove any relation of `bandSize` to `tickConfig`. As you discovered, it is already possible to use `size` for controlling the thick length and it seems odd to have two different knobs to control the same characteristics of the mark.

Let's wait to hear if someone else on the team with more historical context around why `bandSize` might be useful can chime in, so that we are not missing anything here.

Adeyinka1:
Thanks for the thoughtful feedback; that makes sense.

My initial assumption was that since `bandSize` is defined in `TickConfig`, it was intended to have an effect similar to other band-based marks. But I agree that having both `size` and `bandSize` control the same visual property could be confusing.

If the intent is for `size` to be the single source of control for tick thickness, then updating the documentation and removing `bandSize` from `TickConfig` might indeed be the cleaner approach.

Happy to adjust the PR in either direction depending on what’s preferred, whether that’s keeping the current fix or aligning with a documentation/cleanup change.

yhoonkim:
Thanks @Adeyinka1 for suggesting the fix! (thanks for pinging me @joelostblom )

I prefer changing the documentation (remove `bandSize` and promote to use `size`) than adding another knob to sync with bandSize.
(For me, it's natural to say tick's `size` is the length of it since it is just a line having thickness and length. It doesn't have "band" like a bar.)

Adeyinka1:
Thanks @yhoonkim for the clarification; that direction sounds good to me. I will update the PR to align with using `size` as the single control for ticks and shift this toward the documentation/config cleanup approach.

Adeyinka1:
I have updated the PR to follow the documentation/config cleanup direction:

* removed the tick `bandSize` wiring change
* removed the added `bandSize` tick tests
* removed `bandSize` from `TickConfig`
* updated the tick docs to promote `size`/`thickness` instead

I also reran the targeted tick tests locally after the change.

PR #9831, opened 2026-04-14
feat: Add comparators/operators for selection params

joelostblom:
This PR extends field predicate comparators (`equal`, `lt`, `lte`, `gt`, `gte`) to accept parameter value references and makes them practical for selection params. My original motivation came from https://github.com/vega/vega-lite/pull/9830 where I had to use a not that intuitive syntax for gt/lt comparisons when using paramaters for the new time/animation functionality. As I was working on this I also thought it was a natural fit in the grammar that already existed for variable parameters and that it had benefits when used with selection parameters not related to time encodings, so I added a couple of example to illustrate that.

*Edit: Just a note that the original motivation of enabling comparisons with animation might be superseeded by #9902 as per https://github.com/vega/vega-lite/pull/9902#issuecomment-5109527930:*

> I think https://github.com/vega/vega-lite/pull/9831 is related but distinct to this PR. Looks like you're adding parameter support to where field predicates already exist, and this is adding field predicates that can live inside selection parameters. I haven't really thought through the differences / whether both are needed in general for VL, however, the selection approach is better for animation because selections can be referenced across not just transforms but also conditional encodings, etc.

Here is how the simple new example in the docs look upon interaction:
recording-2026-04-14_16 19 54-ezgif com-video-to-gif-converter

And here is how the new gallery example looks with a `lt` comparator and empty initial condition:
recording-2026-04-14_16 16 07-ezgif com-video-to-gif-converter

### A few details

1. **Comparator param refs**
   - Adds `{param: ...}` and `{param: ..., field: ...}` support in field comparators.
   - Applies to both transform filters and condition tests via predicate compilation.
2. **Selection-param value resolution**
   - For selection params, comparator refs are resolved from the selection store (projected value index), rather than assuming object-like signal access in all contexts.
3. **Empty-selection semantics**
   - Adds optional `empty` on comparator refs to match variable params:
     - `empty: true` => empty selection passes predicate
     - `empty: false` => empty selection fails predicate
   - Defaults comparator refs for selection params to `empty: true`, avoiding init-time null coercion surprises.
4. **Cross-view composed specs**
   - Comparator refs resolve selection components across composed/sibling views (e.g. `vconcat` where selection is defined in one view and used in another).
5. **Line+point interaction robustness**
   - For point selections, ignores clicks on path marks (`line`/`trail`/`area`) for tuple capture.
   - Makes compiled point-overlay symbols (`mark.point=true`) interactive with pointer cursor so clicks are captured as expected.
   - It was quite unintuitive for lines with point=true how to select the points otherwise
6. **Schema alignment**
   - Regenerates schema to include `ParameterValueRef` (including `empty`) in comparator definitions.

<details>

- For new features:
  - [x] Has unit tests.

Tips:

</details>


src/compile/predicate.ts
```diff
@@ -1,11 +1,108 @@
...
+  ParameterValueRef,
+  Predicate,
+} from '../predicate.js';
 import {logicalExpr} from '../util.js';
+import {varName} from '../util.js';
 import {DataFlowNode} from './data/dataflow.js';
 import {Model} from './model.js';
 import {parseSelectionPredicate} from './selection/parse.js';
 
+function findSelectionComponent(model: Model, param: string) {
```
joelostblom, line 22: Comparator refs can be evaluated in a different unit model than where the selection is declared (for example, `vconcat` sibling views). Ancestor-only lookup misses that case, so this helper first tries normal lookup and then scans from the root model to find the declared selection component.

src/compile/predicate.ts
```diff
@@ -1,11 +1,108 @@
...
+    }
+    if (current?.children?.length) {
+      stack.push(...current.children);
+    }
+  }
+
+  return undefined;
+}
+
+function resolveSelectionParameterValueExpr(model: Model, v: any): string {
```
joelostblom, line 51: This resolves `{param, field}` using the selection store’s projected value index (`values[idx]`) rather than assuming `param[field]` exists everywhere.

src/compile/predicate.ts
```diff
@@ -1,11 +1,108 @@
...
+  if (!fieldProjection) {
+    return undefined;
+  }
+
+  const store = stringValue(`${v.param}_store`);
+  const idx = fieldProjection.index;
+  return `(length(data(${store})) ? data(${store})[0].values[${idx}] : null)`;
+}
+
+function getParameterValueRef(predicate: FieldPredicate): ParameterValueRef {
```
joelostblom, line 68: This extracts the comparator’s value-ref payload uniformly across `equal/lt/lte/gt/gte` so empty-handling logic is operator-agnostic and stays in one place.

examples/compiled/airport_connections.vg.json
```diff
@@ -139,7 +139,7 @@
               "markname": "layer_2_voronoi"
             }
           ],
-          "update": "datum && item().mark.marktype !== 'group' && indexof(item().mark.role, 'legend') < 0 ? {unit: \"layer_2\", fields: org_tuple_fields, values: [(item().isVoronoi ? datum.datum : datum)[\"origin\"]]} : null",
+          "update": "datum && item().mark.marktype !== 'group' && indexof(item().mark.role, 'legend') < 0 && indexof(['line', 'trail', 'area'], item().mark.marktype) < 0 ? {unit: \"layer_2\", fields: org_tuple_fields, values: [(item().isVoronoi ? datum.datum : datum)[\"origin\"]]} : null",
```
joelostblom, line 142: It's not that pretty that this change shows up in so many examples, but I didn't find another way to ensure that points received clicks/interactions instead of lines when plotted together.

src/compile/selection/assemble.ts
```diff
@@ -155,20 +155,28 @@ export function assembleUnitSelectionData(model: UnitModel, data: readonly VgDat
       const sourceName = model.lookupDataSource(model.getDataName(DataSourceType.Main));
       const sourceData = data.find((d) => d.name === sourceName);
 
-      // find the filter transform for the current selection
-      const sourceDataFilter = sourceData.transform.find(
-        (t) => t.type === 'filter' && t.expr.includes('vlSelectionTest'),
+      // find animation-related filters to be applied on the per-frame dataset
+      const timerValueSignal = `${selCmpt.name}_value`;
+      const timerObjectSignal = `${selCmpt.name}[`;
+      const timerStoreSignal = `${selCmpt.name}_store`;
```
joelostblom, line 161: Comparator refs for selection params compile to expressions using `data('<param>_store')`.  For timer selections, those filters must be moved to `*_curr` alongside other animation filters; otherwise frame/dataflow behavior diverges.

src/compile/predicate.ts
```diff
@@ -1,11 +1,108 @@
...
+
+  const selCmpt = findSelectionComponent(model, valueRef.param);
+  if (!selCmpt) {
+    return expr;
+  }
+
+  const store = stringValue(`${valueRef.param}_store`);
+  const isEmptyExpr = `!length(data(${store}))`;
+  const empty = valueRef.empty ?? true;
+  return empty ? `(${isEmptyExpr} || (${expr}))` : `(!${isEmptyExpr} && (${expr}))`;
```
joelostblom, line 103: Defaulting to `true` here matches the selection-predicate default (`empty passes`) and prevents surprising partial data at initialization before any click.

PR #9834, opened 2026-04-17
fix: Avoid cyclic reference when using frozen/unreactive domains

joelostblom:
This PR fixes unstable frozen legend domains when users define parameter expressions like `sort(domain("color"))` and then freeze with `react: false`. In some specs, this pattern effectively referenced a signal-driven scale domain from the same parameter pipeline, which could lead to initialization-order issues (empty/unsorted domains, composition fragility, facet edge cases).

Currently a domain cannot be set to a fixed sequence of values unless one knows the sequence up front, ie there is no way to retrieve the values from the data. This leads to issues when creating e.g. a legend that allows for filtering of the data, because the color domain is recomputed at each click on the legend so that the color changes for that category. We also have to click outside the legend to bring back the the full range of colors and there is no way to shift click to select multiple categories:

recording-2026-04-17_12 36 51-ezgif com-video-to-gif-converter

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygK4CcA28QAspQA4Gc4B6IgYzADsA6AK3zAFNUBLAN2SooaiItwFsirBgHMIAWkgx83fAAFWARioAmAJxUADESkQyEZPlr4A9hRABfADQhcBiP3zwA2qAoOGGVKIYUwAfVIWUgBrEBsZb1IoeFAoAE9cTzhbE2YKGJsAM2YmMCc4ZxAAYXiWPwZDEABdaxAAI3SwLx8-S2qbVlIzUmgXUH4DMIQQBKSMXDSMyxsoZAgKfCyTZH5+kBzUKErY23s1lO8RXwCg5lDLCw6QX26wdJFd7tQV3Zy8jFLyxirZxOSQBQTPx0hB0HUAB5vXKoZopAASKxkkwA7js-uMUgBHRALKDMGD44QzEDxaEfFIAWWY3nw-iSyH8AHEwS9zBiATi8QToGxPBYBVZQJgGMwRNh4IpNJobINkMM4v8MPUDCS5gslisDq4NjTtshdnZ5gcQEcToFgmEBddbiZ7hRHiMoSN3rCMIjDAxUeiGul4HNEAw6mSRhARCJkKJoADuohpgKrhYgA)

After this PR, we can use a parameter with `react: false` to initialize and sort the color domain, making sure that all the categorical levels in the original data are always visible in the legend, so that each category is constantly linked to the same color even as the data is filter and so that shift-clicking to select multiple categories work as desired:

recording-2026-04-17_15 10 02-ezgif com-video-to-gif-converter

This diff for this compared to the spec above is relatively small:

```diff
{
  "data": {
    "url": "https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/cars.json"
  },
  "params": [
    {
      "name": "legend_click",
      "select": {"type": "point", "fields": ["Cylinders"]},
      "bind": "legend"
    },
+   {"name": "frozen_domain", "expr": "sort(domain('color'))", "react": false}
  ],
  "vconcat": [
    {
      "mark": {"type": "point"},
      "transform": [{"filter": {"param": "legend_click"}}],
      "encoding": {
        "color": {
          "field": "Cylinders",
          "type": "nominal",
+         "scale": {"domain": {"expr": "frozen_domain"}}
        },
        "x": {"field": "Horsepower", "type": "quantitative"},
        "y": {"field": "Miles_per_Gallon", "type": "quantitative"}
      }
    },
    {
      "height": 100,
      "mark": {"type": "bar"},
      "transform": [
        {"filter": {"param": "legend_click"}}
      ],
      "encoding": {
        "x": {"field": "Horsepower", "bin": true},
        "y": {"aggregate": "count"}
      }
    }
  ]
}
```

An additional motivation for this PR is to unblock https://github.com/vega/altair/pull/3394. While some of the code in this fix feels a bit "patchy" (like the regex match), the desired behavior more closely matches what one would expect from `react: false`, ie it retrieves a value at chart initialization that does not later change/react to any subsequent updates to the chart. Previously,`sort(domain("color"))` + `react:false` could freeze too early or depend on fragile scale-domain evaluation order.

### Detailed changes

1. **Rewrite parameter `domain('<scale>')` refs to helper source-domain scales**  
   In compiled Vega, parameter signal expressions that reference `domain('color')` now get rewritten to a helper scale (for example `__color_domain_source`) when the referenced scale domain is parameter-driven.
   - This avoids self-referential scale-domain dependency patterns.
   - Applied at compile time in `src/compile/compile.ts`.
2. **Generalize `react: false` parameter freezing behavior**  
   `expr + react:false` parameters compile to one-time/latching updates so the first evaluated value is captured and held stable during interaction.
   - This preserves frozen legend domain behavior under filtering and selection updates.
   - Implemented in `src/parameter.ts`.
3. **Fix helper-scale source inference for composed specs (especially facet)**  
   Helper-scale data-source inference now correctly resolves through composed mark structures and facet wrappers, instead of accidentally using the synthetic `facet` placeholder source.
   - Implemented in `src/compile/compile.ts`.
4. **Expanded tests**  
   Added/updated compile + runtime-focused tests to cover:
   - sorted/frozen parameter domain rewrites,
   - non-sort domain expressions,
   - `react:false` freezing behavior,
   - layered / faceted / concatenated specs.

<details>

- For new features:
  - [x] Has unit tests.

Tips:

</details>

### DIFF
--- a/src/compile/compile.ts
+++ b/src/compile/compile.ts
@@ -6,6 +6,7 @@ import {Config, initConfig, stripAndRedirectConfig} from '../config.js';
 import * as log from '../log/index.js';
 import {normalize} from '../normalize/index.js';
 import {assembleParameterSignals} from '../parameter.js';
+import {isSelectionParameter} from '../selection.js';
 import {LayoutSizeMixins, TopLevel, TopLevelSpec} from '../spec/index.js';
 import {
   AutoSizeParams,
@@ -222,6 +223,17 @@ function assembleTopLevelModel(
   });
 
   const {params, ...otherTopLevelProps} = topLevelProperties;
+  const group = model.assembleGroup([
+    ...layoutSignals,
+    ...model.assembleSelectionTopLevelSignals([]),
+    ...assembleParameterSignals(params),
+  ]);
+
+  const parameterSignalNames = new Set(
+    (params || []).filter((param) => !isSelectionParameter(param)).map((param) => param.name),
+  );
+
+  rewriteSignalDomainRefs(group, parameterSignalNames);
 
   return {
     $schema: 'https://vega.github.io/schema/vega/v6.json',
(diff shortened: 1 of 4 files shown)


src/compile/compile.ts
```diff
@@ -232,12 +244,87 @@ function assembleTopLevelModel(
...
   };
 }
+
+function rewriteSignalDomainRefs(group: any, parameterSignalNames: Set<string>) {
+  if (!group?.signals || !group?.scales) return;
+
+  const helperScaleName = new Map<string, string>();
+
+  for (const signal of group.signals) {
+    if (!parameterSignalNames.has(signal.name)) continue;
```
joelostblom, line 259: We intentionally scope rewriting to parameter signals only, so regular signal expressions and non-parameter paths keep existing behavior

src/compile/compile.ts
```diff
@@ -232,12 +244,87 @@ function assembleTopLevelModel(
...
+
+      signal[key] = expr.replace(/domain\((['"])([^'"]+)\1\)/g, (match: string, _quote: string, scaleName: string) => {
+        const helper = helperScaleFor(scaleName, group, helperScaleName, parameterSignalNames);
+        return helper ? `domain('${helper}')` : match;
+      });
+    }
+  }
+}
+
+function helperScaleFor(
```
joelostblom, line 273: Helper scales (`__<scale>_domain_source`) are introduced to decouple parameter domain expressions from the final signal-driven scale and avoid self-reference / init-order hazards.

src/compile/compile.ts
```diff
@@ -232,12 +244,87 @@ function assembleTopLevelModel(
...
+  helperScaleName: Map<string, string>,
+  parameterSignalNames: Set<string>,
+) {
+  if (helperScaleName.has(scaleName)) {
+    return helperScaleName.get(scaleName);
+  }
+
+  const scale = (group.scales as any[]).find((s) => s.name === scaleName);
+  if (!scale || !scale.domain || !scale.domain.signal) return null;
+  if (!parameterSignalNames.has(scale.domain.signal)) return null;
```
joelostblom, line 285: We only rewrite when the referenced scale domain is itself parameter-driven (`scale.domain.signal` points to a parameter signal), which keeps the transform targeted and avoids unnecessary rewrites.

src/compile/compile.ts
```diff
@@ -232,12 +244,87 @@ function assembleTopLevelModel(
...
+    });
+  }
+
+  helperScaleName.set(scaleName, helper);
+  return helper;
+}
+
+type ScaleFieldRef = {data: string; field: string} | null;
+
+function findScaleFieldRef(marks: any[], scaleName: string, inheritedData: string): ScaleFieldRef {
```
joelostblom, line 305: Domain-source inference traverses composed mark trees to recover a concrete `{data, field}` source for helper scales, rather than relying on the final display scale

src/parameter.ts
```diff
@@ -40,21 +40,30 @@ export function assembleParameterSignals(params: (VariableParameter | TopLevelSe
...
-    if (bind && expr) {
+    if (expr && bind) {
       // Vega's InitSignal -- apply expr to "init"
       const signal: InitSignal = {
         ...rest,
         bind,
         init: expr,
       };
       signals.push(signal);
+    } else if (expr && react === false) {
```
joelostblom, line 53: `expr + react:false` is lowered to latch semantics (capture first resolved value, then hold), which matches “freeze snapshot” expectations for interactive domains.

src/parameter.ts
```diff
@@ -40,21 +40,30 @@ export function assembleParameterSignals(params: (VariableParameter | TopLevelSe
...
+      const sentinelKey = `__vl_${param.name}_sentinel`;
```
joelostblom, line 54: Sentinel object marks the uninitialized state explicitly, avoiding ambiguity with legitimate `null` / empty array expression results.

src/parameter.ts
```diff
@@ -40,21 +40,30 @@ export function assembleParameterSignals(params: (VariableParameter | TopLevelSe
...
+      const signal: NewSignal = {
+        ...rest,
+        value: {[sentinelKey]: true},
+        update: `${param.name} && ${param.name}.${sentinelKey} ? (${expr}) : ${param.name}`,
```
joelostblom, line 58: The update expression captures exactly once (`sentinel -> expr`) and then returns the stored value on subsequent pulses, preserving stability during interaction

PR #9835, opened 2026-04-18
feat: Enable scale binding (pan and zoom) for geoshape marks

### LINKED ISSUE #3306: Selection support for geo

joelostblom:
close #3306

This enables `bind: "scales"` (pan and zoom via `interval` selection) for geographic/cartographic projections, so viewers can drag to pan and scroll to zoom on projected maps just like they can with regular scatter plots.

Previously, using `params: [{name: "mapNav", select: {type: "interval"}, bind: "scales"}]` on a view with a `projection` and `longitude`/`latitude` (or `geoshape`) encodings would produce a runtime error (`Unrecognized signal name: "mapNav_projection_fit"`) because the necessary Vega signals were never generated for projected views.

This currently does not support layering of another marks or multiviews. There are also some projections that should probably have panning mapped to rotation instead of the current default. I intentionally wanted to keep this PR as simple as possible to get feedback on the behavior and the implementation; all the remaining functionality could come in follow up PRs.

One behavioral issue I see currently is that I'm not sure is that for some projections, there is a small "jump" when starting to either pan or zoom from the initial position (or from reset after doubleclick). It seems like the signals responsible for the panning are updating to the correct values, so the issue lies elsewhere and I can't figure out where it is coming from. You can see an example in the animation below when I switch to the mercator projection. When I'm dragging with the pointer upwards, there is first a small "jump" downwards:

recording-2026-04-18_12 23 40-ezgif com-video-to-gif-converter

In the animation below, you can also see that when zooming close to the edge there is sometimes an initial step of panning only with no zoom so the view gets closer to the mouse pointer. That's not as disruptive as the "pan in the wrong direction" mentioned above, but still seems like a bug and there should at least be some zoom in the initial step (and I don't see this when panning charts without geoshape marks). It happens for both projections in the clip:

recording-2026-04-18_12 26 04-ezgif com-video-to-gif-converter

```json

{
  "width": 600,
  "height": 250,
  "params": [
    {
      "name": "projection",
      "value": "equalEarth",
      "bind": {
        "input": "select",
        "options": [
          "albers",
          "albersUsa",
          "azimuthalEqualArea",
          "azimuthalEquidistant",
          "conicConformal",
          "conicEqualArea",
          "conicEquidistant",
          "equalEarth",
          "equirectangular",
          "gnomonic",
          "mercator",
          "naturalEarth1",
          "orthographic",
          "stereographic",
          "transverseMercator"
        ]
      }
    },
    { "name": "geo", "select": { "type": "interval" }, "bind": "scales" }
  ],
  "data": {
    "url": "data/world-110m.json",
    "format": {"type": "topojson", "feature": "countries"}
  },
  "projection": {"type": {"expr": "projection"}},
  "mark": {"type": "geoshape", "fill": "lightgray", "stroke": "gray"}
}
```
One limitation is that some views lend themselves best to rotation rather than panning, e.g. the orthographic/sphere, but that is outside scope for this PR. A follow up could maybe include something like this to indicate what type of panning should be executed:

```json
{
  "name": "nav",
  "select": {
    "type": "interval",
    "pan": "rotate"
  },
  "bind": "scales"
}
```

### DIFF
--- a/src/compile/selection/interval.ts
+++ b/src/compile/selection/interval.ts
@@ -31,8 +31,14 @@ const interval: SelectionCompiler<'interval'> = {
       const def: IntervalSelectionConfigWithField = {...(isObject(selDef.select) ? selDef.select : {})};
       def.fields = [SELECTION_ID];
       if (!def.encodings) {
-        // Remap default x/y projection
-        def.encodings = selDef.value ? (keys(selDef.value) as GeoPositionChannel[]) : [LONGITUDE, LATITUDE];
+        // Remap default x/y projection. Only include channels that the model
+        // actually has field definitions for (geoshape specs may have no
+        // longitude/latitude encoding, so projection-bound interactions work
+        // purely through scale/translate signals).
+        const candidates: GeoPositionChannel[] = selDef.value
+          ? (keys(selDef.value) as GeoPositionChannel[])
+          : [LONGITUDE, LATITUDE];
+        def.encodings = candidates.filter((ch) => model.fieldDef(ch));
       }
 
       selDef.select = {type: 'interval', ...def};
--- a/src/compile/projection/assemble.ts
+++ b/src/compile/projection/assemble.ts
@@ -2,6 +2,7 @@ import {Projection as VgProjection, SignalRef} from 'vega';
 import {contains} from '../../util.js';
 import {isSignalRef} from '../../vega.schema.js';
 import {isConcatModel, isLayerModel, Model} from '../model.js';
+import {projectionFitName, projectionScaleName, projectionTranslateName} from '../selection/scales.js';
 
 export function assembleProjections(model: Model): VgProjection[] {
   if (isLayerModel(model) || isConcatModel(model)) {
@@ -25,6 +26,7 @@ export function assembleProjectionForModel(model: Model): VgProjection[] {
 
   const projection = component.combine();
   const {name} = projection; // we need to extract name so that it is always present in the output and pass TS type validation
+  const selectionName = component.interactiveSelection;
 
   if (!component.data) {
     // generate custom projection, no automatic fitting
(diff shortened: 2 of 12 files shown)


src/compile/selection/interval.ts
```diff
@@ -31,8 +31,14 @@ const interval: SelectionCompiler<'interval'> = {
...
-        // Remap default x/y projection
-        def.encodings = selDef.value ? (keys(selDef.value) as GeoPositionChannel[]) : [LONGITUDE, LATITUDE];
+        // Remap default x/y projection. Only include channels that the model
+        // actually has field definitions for (geoshape specs may have no
+        // longitude/latitude encoding, so projection-bound interactions work
+        // purely through scale/translate signals).
+        const candidates: GeoPositionChannel[] = selDef.value
+          ? (keys(selDef.value) as GeoPositionChannel[])
+          : [LONGITUDE, LATITUDE];
+        def.encodings = candidates.filter((ch) => model.fieldDef(ch));
```
joelostblom, line 41: For geoshape specs that have `projection` + `bind: "scales"` but no `longitude`/`latitude` field encodings, the interval parser unconditionally remapped the default `['x', 'y']` encodings to `['longitude', 'latitude']`. Since those channels had no field definition, this produced two `"Cannot project a selection on encoding channel …"` warnings. The default-encoding candidates are now filtered to only include channels the model actually has field definitions for. This was a pre-existing issue that became visible now that `bind: "scales"` works for projected views.

joelostblom:
I think I found the solution to the initial jagged jump when panning and zooming! With the following changes in Vega https://github.com/vega/vega/pull/4313 and the recent updates in this PR, I see completely smooth zooming and panning for the VL specs above locally:

https://github.com/user-attachments/assets/0d6c60cc-a276-433d-a888-bf119d2f6f3f

Once these two PRs are reviewed and hopefully merged, I think the next step would be to make this work together with other marks in layered views (e.g. points on top of a geoshape).

PR #9840, opened 2026-05-07
feat: support custom arrays for sorting order

joelostblom:
This is in a sense a follow up on https://github.com/vega/vega-lite/pull/9641 which made it so that a sorted color domain automatically set the order of e.g. stacked bars. However, that does not work if the color encoding is sorted by an array since the `order` encoding sort does not accept an array (however, it does work to specify a domain as a custom array to color which gets picked up by order). This PR tries to bridge that gap by adding support for custom sort arrays on the `order` encoding channel.

This makes definitions such as the following valid:

```json
    "order": {
      "field": "segment",
      "type": "nominal",
      "sort": ["high", "medium", "low"]
    }
```

Custom order arrays work for:

- mark ordering, including line and trail marks,
- stack ordering,
- multiple order fields,
- aggregated charts,
- aggregated order fields,
- sort arrays inherited from matching `color` or `fill` encodings.

### Order field API

`OrderFieldDef.sort` now accepts either `SortOrder` or `SortArray`. `OrderOnlyDef.sort` remains restricted to ascending or descending because a sort array has no field to compare against in that form. Multiple order field definitions are also preserved during encoding
normalization.

### Derived sort indexes

Vega comparators accept fields and ascending/descending directions rather than custom value arrays. Each custom order array is therefore compiled into a formula field that assigns an integer index to each value.

For example:

    ["high", "medium", "low"]

produces an index equivalent to:

    high -> 0
    medium -> 1
    low -> 2
    other values -> 3

The mark or stack comparator then sorts the derived index in ascending order. Generated field names include the order definition's position, for example `order_segment_0_sort_index`, so multiple custom order fields do not collide.

### Aggregation

Custom order indexes must remain available after encoding aggregation. For non-aggregated order fields, both the original field and its derived sort index are included in the aggregate grouping. Preserving the original field is important because unspecified values share the fallback index and must not be collapsed into one group.

For aggregated order fields, the aggregate is calculated first and the custom sort index formula is inserted afterward. This ensures the formula compares the post-aggregate value and that downstream mark or stack comparators reference an existing field.

### Color and fill inheritance

When a `color` or `fill` field has a custom sort array, a matching order field without its own sort inherits that array.

Inheritance happens during encoding normalization so that:

- nominal order fields do not produce a misleading "no inherent order" warning,
- an explicit `order.sort` is never overwritten,
- inheritance occurs only when the effective field references match, including
  transformations such as `timeUnit`,
- fields using different transformations do not accidentally share ordering.

An explicit color scale domain retains precedence over `color.sort`, matching the existing scale-domain behavior.

Automatically generated stack-order formulas use field predicate expressions, so transformed time-unit fields are compared consistently.


joelostblom:
While this PR is ready for review, I realize that it has become quite large and given that it is already possibly to specify custom sort to `order` by setting `color.scale.domain` (rather than `color.sort`), this PR is really only useful in the following scenarios:

- Custom stack order different from color order.
- Custom stack order without a color encoding.
- Custom ordering for line/trail points.
- Custom ordering for other marks where order controls rendering order.
- Multiple custom order fields and tie-breaking.

As such, I will leave this as a draft. There is a lot else to review that is higher priority. If someone feels strongly about any of the functionality above, just comment here with a use case.

PR #9842, opened 2026-05-11
refactor: simplify isDiscrete and isContinuous

### LINKED ISSUE #9596: Simplify isDiscrete and isContinuous
The methods in https://github.com/vega/vega-lite/blob/1a3e501ccda41ec1ddcc9643e92e2729977fcdb7/src/type.ts#L21-L26 can probably be replaced by calls to the methods of the same names in channeled.ts. The logic is different so we need to carefully review.

1wos:
Closes #9596.

## Summary

Migrate three callsites from the type-only `isDiscrete` / `isContinuous` in `src/type.ts` to the field-aware `isDiscrete` (binning-aware) plus a new mirror `isContinuous` in `src/channeldef.ts`:

- `channelHasNestedOffsetScale` (`src/encoding.ts`)
- `initEncoding` continuous-position warning (`src/encoding.ts`)
- `getPositionStep` `offsetIsDiscrete` (`src/compile/scale/range.ts`)

`tooltipRefForEncoding` (`src/compile/mark/encode/tooltip.ts`) keeps the type-only `isDiscrete` because it wants the bin-range tooltip format for binned quantitative fields, not the discrete array-join format (an existing tooltip test caught this when a full migration was attempted).

Added `if (!def.type) return false` guard to `channeldef.isDiscrete` so a `DatumDef` without an explicit type returns `false` silently (matching the prior type-only behavior) instead of throwing.

## Notable behavior changes

For the three migrated callsites:
- **Binned quantitative** fields are now classified as discrete. This affects how `channelHasNestedOffsetScale`, the continuous-position warning, and `getPositionStep` treat e.g. `x: {type: 'quantitative', bin: true}` with an offset. Covered by new tests.
- **`geojson`** is now classified as discrete in those callsites (vs neither under the type-only version). `geojson` is not used on position/offset channels in practice, but the change is documented and covered by a `channelHasNestedOffsetScale` test.

## Test plan

- 11 progression cases for `isDiscrete` (FieldDef/DatumDef × all 5 types × binning variants) in `test/channeldef.test.ts`
- 7 cases for the new `isContinuous` mirror
- 7 direct cases for `channelHasNestedOffsetScale` (ordinal, nominal, binned-quantitative, unbinned-quantitative, missing offset, temporal+timeUnit, geojson)
- 1 case asserting `xOffset` is preserved when `x` is binned quantitative
- Full suite: 3264 unit + 3253 examples passing locally; lint + schema clean

### DIFF
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -868,6 +868,7 @@ export function vgField(
 }
 
 export function isDiscrete(def: TypedFieldDef<Field> | DatumDef<any, any>) {
+  if (!def.type) return false;
   switch (def.type) {
     case 'nominal':
     case 'ordinal':
@@ -879,6 +880,10 @@ export function isDiscrete(def: TypedFieldDef<Field> | DatumDef<any, any>) {
       return false;
   }
   throw new Error(log.message.invalidFieldType(def.type));
+}
+
+export function isContinuous(def: TypedFieldDef<Field> | DatumDef<any, any>) {
+  return !!def.type && !isDiscrete(def);
 }
 
 export function isDiscretizing(def: TypedFieldDef<Field> | DatumDef<any, any>) {
--- a/src/compile/scale/range.ts
+++ b/src/compile/scale/range.ts
@@ -31,6 +31,7 @@ import {
 import {
   getBandPosition,
   getFieldOrDatumDef,
+  isDiscrete,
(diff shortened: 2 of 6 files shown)


1wos:
Closing for now — will reopen after additional refinement.

PR #9843, opened 2026-05-11
refactor: simplify isDiscrete and isContinuous

### LINKED ISSUE #9596: Simplify isDiscrete and isContinuous

1wos:
Closes #9596.

## Summary

Migrate three callsites from the type-only `isDiscrete`/`isContinuous` in `src/type.ts` to the field-aware `isDiscrete` (binning-aware) plus a new mirror `isContinuous` in `src/channeldef.ts`:

- `channelHasNestedOffsetScale` (`src/encoding.ts`)
- `initEncoding` continuous-position warning (`src/encoding.ts`)
- `getPositionStep` `offsetIsDiscrete` (`src/compile/scale/range.ts`)

`tooltipRefForEncoding` (`src/compile/mark/encode/tooltip.ts`) keeps the type-only `isDiscrete` because it wants the bin-range tooltip format for binned quantitative fields, not the discrete array-join format (an existing tooltip test caught this when a full migration was attempted).

Added `if (!def.type) return false` guard to `channeldef.isDiscrete` so a `DatumDef` without an explicit type returns `false` silently (matching the prior type-only behavior) instead of throwing.

## Behavior change

The only semantic change in the migrated callsites is that **binned quantitative fields are now classified as discrete**. Concretely:

- `channelHasNestedOffsetScale` returns `true` for `x: {type: 'quantitative', bin: true}` with an offset sibling.
- The continuous-position warning in `initEncoding` no longer fires for binned quantitative `x`, so `xOffset` is preserved instead of being silently dropped.
- `getPositionStep`'s `offsetIsDiscrete` is `true` for a binned quantitative offset.

This matches the semantic intent: binned quantitative is effectively categorical (discrete buckets), so treating it as discrete for offset-sibling and step-calculation purposes is what users expect.

`geojson` would also be classified as discrete by the field-aware function, but TypeScript already prevents `geojson` from being assigned to position/offset channels, so the migrated callsites cannot encounter it.

## Test plan

- 11 progression cases for `isDiscrete` (FieldDef/DatumDef × all 5 types × binning variants) in `test/channeldef.test.ts`
- 7 cases for the new `isContinuous` mirror
- 6 direct cases for `channelHasNestedOffsetScale` (ordinal, nominal, binned-quantitative, unbinned-quantitative, missing offset, temporal+timeUnit)
- 1 case asserting `xOffset` is preserved when `x` is binned quantitative

Local checks: 3264 unit tests + 3253 examples passing, `tsc --noEmit` clean, lint clean, schema build clean.

### DIFF
--- a/src/channeldef.ts
+++ b/src/channeldef.ts
@@ -868,6 +868,7 @@ export function vgField(
 }
 
 export function isDiscrete(def: TypedFieldDef<Field> | DatumDef<any, any>) {
+  if (!def.type) return false;
   switch (def.type) {
     case 'nominal':
     case 'ordinal':
@@ -879,6 +880,10 @@ export function isDiscrete(def: TypedFieldDef<Field> | DatumDef<any, any>) {
       return false;
   }
   throw new Error(log.message.invalidFieldType(def.type));
+}
+
+export function isContinuous(def: TypedFieldDef<Field> | DatumDef<any, any>) {
+  return !!def.type && !isDiscrete(def);
 }
 
 export function isDiscretizing(def: TypedFieldDef<Field> | DatumDef<any, any>) {
--- a/src/compile/scale/range.ts
+++ b/src/compile/scale/range.ts
@@ -31,6 +31,7 @@ import {
 import {
   getBandPosition,
   getFieldOrDatumDef,
+  isDiscrete,
   isFieldDef,
   isFieldOrDatumDef,
   ScaleDatumDef,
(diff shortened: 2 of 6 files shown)


domoritz:
Thanks for taking this on, @1wos. This made the problem with #9596 concrete, which was useful.

All suites pass, but compile probes against main show the migration changes behavior for binned fields near offset channels:

1. Crash instead of graceful degradation. A binned continuous position + offset channel no longer drops the offset with a warning. It now throws Cannot use xOffset scale if x scale is not discrete. (also for bin: {maxbins} and bin: "binned"). That's because scaleType() never gives binned quantitative fields the band scale a nested offset needs.
2. Silent sizing change. In getPositionStep, a binned quantitative offset now counts as discrete, but its scale is linear. domain('xOffset').length is always 2, so width: {step: 20} silently becomes step 50.

So these callsites depend on type-only semantics: making them bin-aware implies band-scale support for binned fields. That's a feature, not a refactor. Closing this PR, but we're salvaging your predicate tests and doc clarifications in a follow-up, plus compile-level regression tests for binned + offset.

PR #9847, opened 2026-05-16
fix: include year in automatic date axis labels for multi-year ranges

### LINKED ISSUE #9747: Omitting years for automatically formatted date axis labels impedes interpretation of axis
### What happened?

Image

As you can see, the x tick labels on the plot show months but no year references.

Code to reproduce:

```python
import numpy as np
import polars as pl
import altair as alt
from datetime import datetime, timedelta

start_date = datetime(2015, 9, 10)
num_days = 3695
dates = pl.date_range(
    start=start_date,
    end=start_date + timedelta(days=num_days - 1),
    interval="1d",
    eager=True,
)
price = np.ones(num_days)

df = pl.DataFrame({
    "Date": dates,
    "Data": price,
})

(
    alt.Chart(df.select(["Date", "Data"]))
    .mark_line(opacity=0.5)
    .encode(x="Date", y="Data")
    .properties(height=200, width=900)
)
```

### What would you like to happen instead?

I can think of several options:
1. Display years slightly below months
2. Display years in between months
3. Display years in place of months

Any would work for me. 3. implies losing some granularity but maybe is simpler. Generally I think the most important value to show is the lowest frequency period that changes within the plotting window, but I haven't thought this through thoroughly.

### Which version of Altair are you using?

6.0.0

mvanhorn:
## Summary

When the inferred date domain spans more than one calendar year, the automatic axis-label format dropped the year and only showed month/day. The chart was ambiguous - a tick labeled "Mar 5" could be any year in the range. Detect the multi-year case and append the year to the default format.

## Why this matters

#9747 reproduced this with a two-year financial dataset where the chart was unreadable without manually setting `labelFormat`. Multi-year auto-formatting is the expected behavior in every other charting library.

## Changes

- `src/compile/axis/properties.ts` - compute domain year-span and switch the auto-format to include the year.
- `test/compile/axis/assemble.test.ts` - tests covering single-year (no year suffix), exact-year-boundary, and multi-year cases.

## Testing

New axis tests + existing axis tests pass locally.

Fixes #9747

### DIFF
--- a/src/compile/axis/properties.ts
+++ b/src/compile/axis/properties.ts
@@ -19,7 +19,7 @@ import {hasDiscreteDomain} from '../../scale.js';
 import {Sort} from '../../sort.js';
 import {durationExpr, normalizeTimeUnit} from '../../timeunit.js';
 import {NOMINAL, ORDINAL, Type} from '../../type.js';
-import {contains, normalizeAngle} from '../../util.js';
+import {contains, normalizeAngle, stringify} from '../../util.js';
 import {isSignalRef} from '../../vega.schema.js';
 import {mergeTitle, mergeTitleFieldDefs} from '../common.js';
 import {guideFormatType} from '../format.js';
@@ -65,6 +65,9 @@ export const axisRules: {
     axis.labelBaseline || defaultLabelBaseline(labelAngle, orient, channel),
 
   labelFlush: ({axis, fieldOrDatumDef, channel}) => axis.labelFlush ?? defaultLabelFlush(fieldOrDatumDef.type, channel),
+
+  labelExpr: ({axis, config, channel, fieldOrDatumDef, model, scaleType}) =>
+    axis.labelExpr ?? defaultMultiYearTemporalLabelExpr({axis, config, channel, fieldOrDatumDef, model, scaleType}),
 
   labelOverlap: ({axis, fieldOrDatumDef, scaleType}) =>
     axis.labelOverlap ??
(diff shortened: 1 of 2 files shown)


domoritz (commented):
Is this really a Vega-Lite issue or a Vega or even D3 issue? Yes, we can manually patch this somewhat in Vega-Lite but it feels like we may want a more fundamental fix for similar issues (like not having the hour when showing minutes) if they also happen. Could you dig more deeply into this? We need this investigation before accepting this pull request.

mvanhorn:
Good instinct - I dug into where the year-omission comes from, and I think you're right that it's fundamentally a Vega/D3-level behavior rather than a Vega-Lite bug.

The labels come from the default multi-scale time format (D3's multi-format, which Vega uses for time-scale axes). By design it shows only the smallest-granularity label per tick and the enclosing unit only at its boundary - so on a multi-year daily domain, ticks land on months and show "Feb", "Mar", with the year only at the January boundary. That's the same mechanism behind the "no hour when showing minutes" case you mentioned: minutes at minute ticks, hour only at the hour boundary. One root behavior, several symptoms.

So there are two levels to fix at:

1. Vega-Lite `labelExpr` (what this PR does): post-process labels to append the missing enclosing unit. I scoped it to the multi-year/year case, but the same expression generalizes cleanly to the whole class - append the next-larger enclosing unit whenever the domain spans more than one of it (year when it crosses years, day when it crosses days, hour when it crosses hours), all detectable from `domain()` and the field accessors, no Vega or D3 change needed. Happy to generalize this PR that way so it covers the "minutes without hour" family too.
2. Vega / the format layer: if you'd rather this be a default for time axes generally (not just when a Vega-Lite user hits it), it belongs in Vega's time-format defaults or a configurable "always show enclosing unit when the domain spans multiple" option, closer to the D3 multi-format design.

My inclination is (1) generalized - least invasive, keeps the fix where Vega-Lite already owns the label expression, and covers the broader class you flagged - but if you'd prefer to push it into Vega I'm glad to take it there. Which direction do you want?

domoritz:
I feel that this is something where the default should be nice in Vega as well. We have more information in Vega at compile time (in Vega-Lite, we don't know at compile time what the scale domain is). It's probably more powerful to do this in Vega. What do you think based on this argument (asking since you leaned towards Vega-Lite).

mvanhorn:
That argument settles it for me: with scale domains known at compile time, Vega can pick the default deterministically, while Vega-Lite would be guessing ahead of the domain. Happy to pursue this on the Vega side instead.

PR #9851, opened 2026-05-18
docs: fix documentation for heatmap default color scheme

tuner:
This PR fixes the incorrect claim in the docs that `"viridis"` is used for the `"heatmap"` color scheme.

### Exhibit 1:

Screenshot from https://vega.github.io/vega-lite/docs/scale.html#range-config. It doesn't look like viridis but something like `"yellowgreenblue"`.

image

### Exhibit 2:

Vega-Lite delegates the actual scheme selection to Vega, which defaults to "yellowgreenblue".

https://github.com/vega/vega/blob/f436c6727018c7bfb41dce846b5746f7a494f7cd/packages/vega-parser/src/config.js#L222-L224

### DIFF
--- a/site/docs/encoding/scale.md
+++ b/site/docs/encoding/scale.md
@@ -163,7 +163,7 @@ By default, Vega-Lite assigns different [default color schemes](#range-config) b
 
 <div class="vl-example" data-name="point_color_ordinal"></div>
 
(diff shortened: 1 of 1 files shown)


hydrosquall (approved):
Well-spotted. Thanks @tuner !

Issue #9152, opened 2023-09-18
Filtering out zero/null values in tooltip and sort the tooltip by field values 


sfc-gh-cheliu:
I'd like to add a feature to **hide/filter out zero/null values** in the tooltip, and potentially **sort the tooltip by the value of each field/column**. Here's an example:
<img width="401" alt="example" src="https://github.com/vega/vega-lite/assets/112987415/3a9f8650-5616-417c-9f4f-a613705d37d2">
Basically I want to hide `type3` and `type5` in the tooltip, and sort the remaining by descending order of Value.

Since I'm not familiar with altair/vega libraries, so this might not be the best approach, but my thought is to add a new parameter/condition in altair and passing this to the tooltip object, so that when the frontend receives this signal, it will hide the values I don't want to see in the tooltip and possibly sort the values.

I found 2 places that can be modified:
1. https://github.com/vega/vega-lite/blob/4ba3b93e575df97ffe56713287b40d6c55b18340/src/compile/mark/encode/tooltip.ts
2. https://github.com/vega/vega-tooltip/blob/5f805fb8a6e76e3e58d2541e51b6106bcc133f11/src/Handler.ts

Any guidance and suggestions are welcome. Thanks! cc @joelostblom @jakevdp as well as core vega-lite developers @kanitw @domoritz @arvind @jheer

sfc-gh-cheliu:
@domoritz I agree. I've pushed some new commits (making 'sort' to be a new parameter as part of 'encoding') in https://github.com/vega/vega-lite/pull/9148 with more details in the description. Appreciate it if you could take a look.

Yeah sorting is trickier. From my understanding, before the encoding defs are passed to vega-tooltip they remain to be some vega expressions, and their values cannot be determined. I really spent some time investigating whether there's a intermediate step to convert these expressions to actual values. But it seems that the values will only be evaluated when the mouse hover event is triggered, that's why I modify the things in vega-tooltip as well. Correct me if I'm wrong, and if you have a way to handle everything within vega-lite & vega, that'd be great:)

domoritz:
Note that Vega tooltip is only one example implementation of a tooltip for Vega. Therefore I think we should try to do as much as possible in Vega/Vega-Lite. As you develop your code, test it without Vega tooltip to see what the default tooltips (using HTML `title`) look like for example.

Can you summarize the proposed API changes here? It'll be easier to discuss it compared to the pull request. I know it's a bit more work but I would appreciate it. Thanks!

[...]

PR #9861, opened 2026-06-26
feat: filter tooltip fields

vs-li:
### Summary

Adds field-level controls for generated tooltips:

- `tooltip: false` omits a field from encoding-generated tooltips while still allowing grouping/selection/other encoding behavior.
- `tooltip: {"filter": "valid"}` omits invalid (null, undefined, NaN, non-finite) values from generated tooltip objects.
- `tooltip: {"filter": {"operator": "!=", "value": 0}}` supports explicit literal comparisons for cases like hiding zero values.

This also updates the generated schema, tooltip docs, compiler tests, and ARIA behavior.

### Motivation

Some fields are useful in an encoding but should not be shown in auto-generated tooltips. Without a Vega-Lite-level control, downstream users may have to reconstruct tooltip behavior manually or post-process generated specs, which is prone to errors and drift.

This addresses the filtering/exclusion part of [#9152](https://github.com/vega/vega-lite/issues/9152). 
Tooltip sorting is not included.

### Examples

Excluding a field from tooltip altogether:
```json
{
  "mark": {"type": "point", "tooltip": true},
  "encoding": {
    "detail": {"field": "__internal_key", "type": "nominal", "tooltip": false}
  }
}
```

Filtering for nonzero values and filtering for "valid" values (non-null, defined, not NaN, finite):
```json
{
  "encoding": {
    "tooltip": [
      {"field": "Date", "type": "nominal"},
      {"field": "type1", "type": "quantitative", "tooltip": {"filter": {"operator": "!=", "value": 0}}},
      {"field": "type2", "type": "quantitative", "tooltip": {"filter": "valid"}}
    ]
  }
}
```

2026-06-26-filter-and-disable-tooltip

<details>

- For new features:
  - [x] Has unit tests.

Tips:

</details>


vs-li:
Hi @domoritz, thanks for running the workflows before - could I get a re-run? Pushed some tests for increased code coverage. Thank you!

vs-li:
@domoritz Would appreciate another approval for the "Format, Schema, and Examples" check :)

domoritz:
Those checks don't need my approval but instead need to run on your fork (may need a token to be added). Alternatively, we need to copy your code into a branch in this repo so the action can run here.

vs-li:
@domoritz Ran the workflow on my fork and it auto-pushed the `chore: update examples [CI]` commit. Looks like I need the first set of workflows to run again - will "Format, Schema, and Examples" run/pass after these ones do?

vs-li:
> @domoritz Ran the workflow on my fork and it auto-pushed the `chore: update examples [CI]` commit. Looks like I need the first set of workflows to run again - will "Format, Schema, and Examples" run/pass after these ones do?

Bumping this!

vs-li:
@domoritz The "Check: Format, Schema, and Examples" action ran and passed on my fork: https://github.com/vs-li/vega-lite/actions/runs/28828777801/job/85497848268
But that doesn't make the check pass here. Can you copy vs-li:vl/filter-tooltip-fields to a branch here or otherwise run check.yml on its content upstream?

examples/specs/bar_tooltip_filter.vl.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "mark": {"type": "bar", "tooltip": true},
+  "encoding": {
+    "x": {"field": "category", "type": "nominal"},
+    "y": {"field": "plotValue", "type": "quantitative"},
+    "color": {"field": "nonZeroTooltip", "type": "quantitative"},
+    "tooltip": [
+      {"field": "category", "type": "nominal"},
+      {"field": "validTooltip", "type": "quantitative", "tooltip": {"filter": "valid"}},
+      {"field": "nonZeroTooltip", "type": "quantitative", "tooltip": {"filter": {"operator": "!=", "value": 0}}},
```
domoritz: This isn't really a syntax we have in Vega-Lite so far. We have lt: etc rather than an explicit operator property. Let's be consistent with existing api unless there is a specific reason for diverging.

src/channeldef.ts
```diff
@@ -243,6 +267,16 @@ export interface FieldDefBase<F, B extends Bin = Bin> extends BandMixins {
...
 
+  /**
+   * Controls whether this field appears in generated tooltips.
+   *
+   * Set to `false` to omit this field from tooltips generated from the encoding. Set to a filter object to omit this
+   * field when its unformatted value does not pass the filter.
+   *
+   * __Default value:__ `true`
+   */
+  tooltip?: boolean | TooltipFieldControl;
```
domoritz: Adding this to field def base means it's being added to encodings where it's meaningless.

site/docs/tooltip.md
```diff
@@ -64,6 +76,27 @@ To avoid that the tooltips groups the data, add an aggregate to the tooltip enco
...
+```js
+{
+  "mark": {"type": "bar", "tooltip": true},
+  "encoding": {
+    "x": {"field": "category", "type": "nominal"},
+    "y": {"field": "plotValue", "type": "quantitative"},
+    "color": {"field": "nonZeroTooltip", "type": "quantitative"},
+    "tooltip": [
+      {"field": "category", "type": "nominal"},
+      {"field": "validTooltip", "type": "quantitative", "tooltip": {"filter": "valid"}},
```
domoritz: Tooltip inside tooltip nesting is a bit odd

src/compile/mark/encode/tooltip.ts
```diff
@@ -145,26 +178,112 @@ export function tooltipData(
...
+  }
+
+  switch (filter.operator) {
+    case '==':
+      return fieldFilterExpression({...fieldPredicate, equal: filter.value}, true, expr);
+    case '!=':
+      return `!(${fieldFilterExpression({...fieldPredicate, equal: filter.value}, true, expr)})`;
+    case '<':
+      return typeof filter.value === 'boolean'
+        ? undefined
```
domoritz: Would this mean we just silently drop without a warning?

domoritz (changes requested):
Thanks for the pull request. I need to make a more thorough pass but here are my initial comments. 

I'm not sure I get why we need the filter with custom predicates. Maybe a simpler, more scoped design could work.

src/compile/mark/encode/tooltip.ts
vs-li: Yes, but will add a warning in log/message.ts!

site/docs/tooltip.md
vs-li: True, changing the structure to a "filter" field inside of tooltip instead, + will be using the established logicalExpr helpers as well

src/channeldef.ts
vs-li: Changing the filter to be on tooltip encodings only but leaving tooltip: false on channeldef so any encoding can be turned off in the tooltip. Is that okay?

examples/specs/bar_tooltip_filter.vl.json
vs-li: Changing to use field predicate filter syntax!

vs-li:
> Thanks for the pull request. I need to make a more thorough pass but here are my initial comments.
> 
> I'm not sure I get why we need the filter with custom predicates. Maybe a simpler, more scoped design could work.

The custom predicates was just a way to solve issue #9152 and for any others with custom use cases for it. I personally wanted a way to remove specific fields in the tooltip (did not want the user to read) but keep them in the encoding for other use cases, like matching a named field between multiple charts without having to reconstruct the rest of auto-tooltip's functionality. That part is just `tooltip: false`.

I figured if I was adding the `tooltip: false` part already, might as well try to solve the zero-filtering at the same time... and if there's already filtering being added, might as well make it customizable too.

site/docs/tooltip.md
vs-li: So new behavior will be that instead of putting tooltip filtering on a visual encoding:
```json
"y": {
  "field": "value",
  "type": "quantitative",
  "tooltip": {"filter": {"lt": 0}}
}
```
the filter will belong to an explicit tooltip encoding for the field:
```json
"encoding": {
  "y": {"field": "value", "type": "quantitative"},
  "tooltip": [
    {"field": "value", "type": "quantitative", "filter": {"lt": 0}}
  ]
}
```

src/compile/mark/encode/tooltip.ts
vs-li: Added warnings.

site/docs/tooltip.md
vs-li: Changed to this above format!

src/channeldef.ts
vs-li: Changed to this mentioned format.

examples/specs/bar_tooltip_filter.vl.json
vs-li: Changed!

vs-li:
@domoritz Pushed a new commit to address comments, would be good to get the checks running too 🙂

vs-li:
@domoritz Bumping for checks and another look!

vs-li:
@domoritz Hi, just another ping here for review.

PR #9868, opened 2026-07-10
fix: apply cornerRadiusEnd to the rendered value end of bars

vs-li:
### Summary

Makes `cornerRadiusEnd` apply to the actual rendered value-end of bars:

- Positive horizontal bars round the top-right & bottom-right, negative bars round the top-left & bottom-left
- Positive vertical bars round the top-left & top-right, negative bars round the bottom-left & bottom-right
- Above applies to stacked bars too.
- Ranged bars round all corners.

### Motivation

Previously, Vega-Lite compiled `cornerRadiusEnd` to hardcoded corners (right-side for horizontal and top-side for vertical). This assumes positive values, causing negative bars to round the visually incorrect side.

Downstream consumers would need to post-process generated Vega specs, which is fragile.
It's best if we make the fix here since the behavior seems unintentional/bandaid-like.

### Example
image

```json
{
  "data": {
    "values": [
      {"category": "Positive", "value": 10},
      {"category": "Negative", "value": -10}
    ]
  },
  "mark": {
    "type": "bar",
    "cornerRadiusEnd": 5
  },
  "encoding": {
    "y": {"field": "label", "type": "nominal"},
    "x": {"field": "value", "type": "quantitative"}
  }
}
```

<details>
  <summary><h2>Checklist</h2></summary>

- [x] This PR is atomic (i.e., it fixes one issue at a time).
- [x] The title is a concise [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties").
- [x] `npm test` runs successfully
- For new features:
  - [x] Has unit tests.
  - [ ] Has documentation under `site/docs/` + examples.
</details>

PR #9872, opened 2026-07-16
fix: apply cornerRadiusEnd to the rendered value end of bars

domoritz:
New version of https://github.com/vega/vega-lite/pull/9868 so we rebuild examples etc. 

### Summary

Makes `cornerRadiusEnd` apply to the actual rendered value-end of bars:

- Positive horizontal bars round the top-right & bottom-right, negative bars round the top-left & bottom-left
- Positive vertical bars round the top-left & top-right, negative bars round the bottom-left & bottom-right
- Above applies to stacked bars too.
- Ranged bars round all corners.

### Motivation

Previously, Vega-Lite compiled `cornerRadiusEnd` to hardcoded corners (right-side for horizontal and top-side for vertical). This assumes positive values, causing negative bars to round the visually incorrect side.

Downstream consumers would need to post-process generated Vega specs, which is fragile.
It's best if we make the fix here since the behavior seems unintentional/bandaid-like.

### Example
image

```json
{
  "data": {
    "values": [
      {"category": "Positive", "value": 10},
      {"category": "Negative", "value": -10}
    ]
  },
  "mark": {"type": "bar", "cornerRadiusEnd": 5},
  "encoding": {
    "y": {"field": "category", "type": "nominal"},
    "x": {"field": "value", "type": "quantitative"}
  }
}
```

### DIFF
--- a/build/vega-lite-schema.json
+++ b/build/vega-lite-schema.json
@@ -3141,7 +3141,7 @@
               "$ref": "#/definitions/ExprRef"
             }
           ],
-          "description": "- For vertical bars, top-left and top-right corner radius.\n\n- For horizontal bars, top-right and bottom-right corner radius."
+          "description": "Corner radius of the value-end side."
         },
         "cornerRadiusTopLeft": {
           "anyOf": [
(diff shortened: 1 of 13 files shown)


vs-li:
Thank you! 🙂

PR #9148, opened 2023-10-22
feat: add filter and sort support for tooltips

sfc-gh-cheliu:
**Parameter signatures**:
**Filter** - added as a parameter for each tooltip object in the tooltip array:
`{"field": "type1", "type": "quantitative", "**filter**": {"operator": ">", "literal": 0}`

**Sort** - added as a parameter for `encoding`:
`"encoding": { "opacity": { "condition": { "empty": false, "param": "param_46", "value": 0.2 }, "value": 0 }, "size": { "value": 4 }, "tooltip": [ { "field": "Date", "type": "nominal" }, { "field": "type1", "type": "quantitative" }, { "field": "type2", "type": "quantitative" }, { "field": "type3", "type": "quantitative" }, { "field": "type4", "type": "quantitative" }, { "field": "type5", "type": "quantitative" }, { "field": "type6", "type": "quantitative" }, { "field": "type7", "type": "quantitative" }, { "field": "type8", "type": "quantitative" } ], "**sort_tooltip**": {"value": "descending"}, "x": { "field": "Date", "title": null, "type": "ordinal" } }, "mark": { "type": "rule" }, "params": [ { "name": "param_46", "select": { "clear": "mouseout", "fields": [ "Date" ], "nearest": true, "on": "mouseover", "type": "point" } } ] `

Original vega-lite specification:
```json
{
    "config": {
        "view": {
            "continuousHeight": 300,
            "continuousWidth": 300
        }
    },
    "layer": [{
            "data": {
                "values": [{
                        "Date": "2021-06-29",
                        "Type": "type7",
                        "Value": 0
                    },
                    {
                        "Date": "2021-06-29",
                        "Type": "type4",
                        "Value": 0
                    },
                    {
                        "Date": "2021-06-29",
                        "Type": "type5",
                        "Value": 10687
                    },
                    {
                        "Date": "2021-06-29",
                        "Type": "type1",
                        "Value": 11501
                    },
                    {
                        "Date": "2021-06-29",
                        "Type": "type3",
                        "Value": 0
                    },
                    {
                        "Date": "2021-06-22",
                        "Type": "type7",
                        "Value": 5735
                    },
                    {
                        "Date": "2021-06-22",
                        "Type": "type8",
                        "Value": 7486
                    },
                    {
                        "Date": "2021-06-22",
                        "Type": "type1",
                        "Value": 8283
                    },
                    {
                        "Date": "2021-06-22",
                        "Type": "type4",
                        "Value": 8283
                    },
                    {
[...]

PR #9876, opened 2026-07-17
feat: filter tooltip fields

domoritz:
Closes #9861, closes #9148, closes https://github.com/vega/vega-tooltip/pull/820

Adds field-level control over encoding-generated tooltips, addressing the filtering part of #9152 (sorting is out of scope):

- **`tooltip: false`** on any encoding field def omits the field from generated tooltips while keeping it for grouping, selection, or other channels.
- **`filter`** on tooltip encoding array entries omits a field at runtime when its value fails a [predicate](https://vega.github.io/vega-lite/docs/predicate.html). Predicates are tested against the field they are defined on, so `field` and `timeUnit` are implied; logical composition (`not`/`and`/`or`) is supported.

```json
"tooltip": [
  {"field": "Date", "type": "temporal"},
  {"field": "sales", "type": "quantitative", "filter": {"not": {"equal": 0}}},
  {"field": "forecast", "type": "quantitative", "filter": {"valid": true}}
]
```

Filters compile to plain Vega expressions (`merge(...)` of conditional objects, `null` when every field is filtered out), so they work with any tooltip handler (vega-tooltip, the default HTML title tooltip, and custom handlers) without runtime support. ARIA descriptions apply the same tests so screen readers match the visible tooltip.

Unlike reusing full filter-transform predicates, binding the predicate to its field def lets the compiler resolve derived field names (`aggregate`, `bin`, `timeUnit`) against the mark's datum.


vs-li:
Thanks Dominik!! 😄

PR #9883, opened 2026-07-18
feat: specify ranged area marks with size/thickness around a center line 

joelostblom:
I had a go at addressing #3442 

## Summary

This PR adds quantitative `size` support for `area` marks, allowing an area to expand symmetrically around an `x` or `y` centerline.

This enables ranged area/ribbon visualizations without manually calculating separate `x2` or `y2` boundary fields, which means these now also work with a categorical grouping on the one axis and we can 

```json
{
  "title": "Body Mass Density Ribbons by Species",
  "data": {"url": "data/penguins.json"},
  "transform": [
    {
      "density": "Body Mass (g)",
      "groupby": ["Species"],
      "resolve": "independent"
    }
  ],
  "mark": {"type": "area"},
  "encoding": {
    "x": {
      "field": "value",
      "type": "quantitative",
      "title": "Body Mass (g)"
    },
    "y": {"field": "Species", "type": "nominal"},
    "size": {"field": "density", "type": "quantitative"}
  }
}
```

image

In this example, `x` describes the ribbon trajectory, `y` places each ribbon in a categorical lane, and `size` controls its thickness around the `y` centerline.

The compiler resolves the thickness axis from the area's orientation:

- vertical orientation emits `y` and `y2` around the `y` centerline,
- horizontal orientation emits `x` and `x2` around the `x` centerline.

Each boundary is offset by half the scaled size:

```text
center + size / 2
center - size / 2
```

Explicit `mark.orient` takes precedence when both positional axes are quantitative and inference would be ambiguous.

If `x2` or `y2` is present, the area is treated as an explicitly ranged area rather than a size-thickness ribbon. In that case, `size` is incompatible and is dropped through the existing channel-validation path instead of silently replacing an authored boundary.

## Grouping and offsets

Area ribbons are automatically grouped by categorical lane fields so separate violins or ribbons do not merge into one path. Discrete `xOffset` and `yOffset` fields are also included in path grouping. This means grouped ribbons work without requiring a redundant `color` encoding. Grouping supports field, datum, value, and conditional `size` definitions consistently. Size thickness composes with existing positional offsets rather than replacing them so that offsetting works as expected as in this example:

```json
{
  "title": "Body Mass Density Ribbons by Species and Sex",
  "width": {"step": 70},
  "height": 240,
  "data": {"url": "data/penguins.json"},
  "transform": [
    {"filter": {"field": "Sex", "oneOf": ["MALE", "FEMALE"]}},
    {
      "density": "Body Mass (g)",
      "groupby": ["Species", "Sex"],
      "resolve": "independent",
      "bandwidth": 100
    }
  ],
  "mark": {"type": "area", "opacity": 0.9},
  "encoding": {
    "y": {
      "field": "value",
      "type": "quantitative",
      "title": "Body Mass (g)"
    },
    "x": {"field": "Species", "type": "nominal"},
    "xOffset": {"field": "Sex", "type": "nominal"},
    "size": {"field": "density", "type": "quantitative"},
    "color": {"field": "Sex", "type": "nominal"}
  }
}
```

image

For ribbons split with categorical `xOffset` or `yOffset`, the default size range uses the rendered spacing of the offset subgroups:

- band offset scales use their bandwidth,
- point offset scales use the distance between neighboring rendered positions.

This accounts for scale padding and keeps each ribbon within its subgroup. Explicit `size.scale.range` values are preserved exactly and are not divided after scaling.

## Default Size Range

Without an explicit range, area thickness is relative to the available lane:

- band lanes use the lane bandwidth,
- stepped point lanes use their configured step,
- fixed-size point lanes use the rendered distance between neighboring positions.

Scale names are resolved through the model, so the generated expressions also work in layered and other composed specifications where scales may be renamed.

## Stacking

Default area stacking is disabled in size-thickness mode.

The positional encoding represents the centerline, not a stackable area boundary, so applying the normal area stack transform would move the centerline and produce incorrect geometry. Explicit ranged areas retain their existing stacking behavior.

## Legends

The default `size` legend is disabled for area ribbons because the standard size legend does not clearly communicate symmetric thickness around a centerline.

Authors can still opt in explicitly:

```json
{
  "size": {
    "field": "density",
    "type": "quantitative",
    "legend": {}
  }
}
```


src/encoding.ts
```diff
@@ -332,6 +334,10 @@ export interface Encoding<F extends Field> {
 
 export interface EncodingWithFacet<F extends Field> extends Encoding<F>, EncodingFacetMapping<F> {}
 
+export function isAreaSizeThickness(mark: Mark, encoding: Encoding<string>): boolean {
+  return mark === AREA && !!encoding.x && !!encoding.y && !!encoding.size && !encoding.x2 && !encoding.y2;
+}
```
joelostblom, line 340: This predicate intentionally defines the feature boundary in one place. Requiring no `x2`/`y2` prevents `size` from silently changing an explicitly ranged area and keeps mark encoding, stacking, scale defaults, grouping, and legends consistent.

src/compile/mark/area.ts
```diff
@@ -28,3 +72,113 @@ export const area: MarkCompiler = {
...
+    ...thickness,
+    mult: (thickness.mult ?? 1) * factor,
+  };
+}
+
+function thicknessOffsetRef(thickness: VgValueRef, factor: number): VgValueRef {
+  return halfThicknessRef(thickness, factor);
+}
+
+function withThicknessOffset(center: AreaValueRef, thickness: AreaValueRef, factor: number): AreaValueRef {
```
joelostblom: Both the center position and `size` may compile to Vega production-rule arrays. This cross-product preserves their tests and fallbacks while applying positive or negative half-thickness to each area edge.

src/compile/mark/area.ts
```diff
@@ -28,3 +72,113 @@ export const area: MarkCompiler = {
...
+        ...(test ? {test} : {}),
+        ...withOffset(centerWithoutTest, thicknessOffsetRef(thicknessWithoutTest, factor)),
+      });
+    }
+  }
+
+  return refs.length === 1 ? refs[0] : refs;
+}
+
+function withOffset(baseRef: VgValueRef, offset: VgValueRef): VgValueRef {
```
joelostblom, line 114: Thickness is added to an existing positional offset rather than replacing it. This allows `xOffset`/`yOffset` subgroup placement and centered ribbon thickness to coexist.

src/compile/mark/area.ts
```diff
@@ -28,3 +72,113 @@ export const area: MarkCompiler = {
...
+  } else if ('value' in v && v.value !== undefined) {
+    base = stringValue(v.value);
+  } else {
+    base = '0';
+  }
+
+  if ('band' in v && v.band !== undefined && 'scale' in v && v.scale) {
+    const band = typeof v.band === 'object' ? valueRefExpr(v.band) : stringValue(v.band);
+    base = `(${base}) + bandwidth(${stringValue(v.scale)}) * (${band})`;
+  }
```
joelostblom: Band-position metadata must be preserved when converting a value reference to an expression. Offset band scales use `band: 0.5` to place marks at subgroup centers; omitting it shifts ribbons to the leading band edge.

src/compile/scale/range.ts
```diff
@@ -283,6 +283,42 @@ function defaultRange(channel: ScaleChannel, model: UnitModel): VgRange {
       return getOffsetRange(channel, model, scaleType);
 
     case SIZE: {
```
joelostblom, line 285: The default range is based on rendered spacing rather than raw domain cardinality. This accounts for point/band padding and keeps explicit `size.scale.range` values untouched.

src/compile/mark/init.ts
```diff
@@ -127,6 +127,25 @@ function orient(mark: Mark, encoding: Encoding<string>, specifiedOrient: Orienta
 
     // falls through
     case AREA:
```
joelostblom, line 139: Explicit orientation is checked first because two quantitative positional channels are ambiguous. Otherwise, orientation is inferred from which axis behaves as the measure versus the categorical lane.

src/encoding.ts
```diff
@@ -778,6 +784,47 @@ export function pathGroupingFields(mark: Mark, encoding: Encoding<string>): stri
           // For trail, size should not group trail lines.
           return details;
         }
+
```
joelostblom, line 806: `size` varies along a ribbon and therefore must not itself facet the path. Categorical lane and offset fields do facet it, preventing separate violins from being connected into one area.

src/stack.ts
```diff
@@ -169,6 +169,10 @@ export function stack(m: Mark | MarkDef, encoding: Encoding<string>): StackPrope
   const stackedFieldDef = encoding[fieldChannel] as PositionFieldDef<string> | PositionDatumDef<string>;
   const stackedField = isFieldDef(stackedFieldDef) ? vgField(stackedFieldDef, {}) : undefined;
 
+  if (stackedFieldDef.stack === undefined && isAreaSizeThickness(mark, encoding)) {
+    return null;
+  }
```
joelostblom, line 193: Normal area stacking treats a positional field as a boundary. In ribbon mode that field is the centerline, so implicit stacking would move the entire ribbon and must be disabled.

src/compile/legend/parse.ts
```diff
@@ -98,7 +99,8 @@ export function parseLegendForChannel(model: UnitModel, channel: NonPositionScal
   const legendCmpt = new LegendComponent({}, getLegendDefWithScale(model, channel));
   parseInteractiveLegend(model, channel, legendCmpt);
 
-  const disable = legend !== undefined ? !legend : legendConfig.disable;
+  const disableByDefaultForAreaSize = channel === SIZE && isAreaSizeThickness(model.mark, model.encoding);
+  const disable = legend !== undefined ? !legend : disableByDefaultForAreaSize || legendConfig.disable;
```
joelostblom, line 103: The legend is disabled only for centered area thickness mode. An explicit `legend` definition still opts in through the normal precedence path.

joshpoll:
I think a streamgraph can be viewed as a stacked violin (minus the KDE transform) since they're both center-aligned areas. (Or alternatively a streamgraph can be viewed as having some custom baseline offset like in observable plot.) I wonder if there's a way to align the two specs more closely? It would also give a natural stacking semantics here I think.

domoritz (changes requested):
I like the approach overall. Gets us closer to density marks without going all the way yet so we can compose better. 

There is some overlap with #9823 so maybe we need to resolve that one first. 

Other than that, there are a few small issues we need to resolve in this pull request:

* `"vertical area with size encoding for thickness"` case looks broken when rendered. Make sure this works with time encodings. `pathGroupingFields` treats any non-quantitative x/y as a lane
* Default size range with two continuous axes falls through to `[1, 4]` px. That's inconsistent with `[0, bandwidth]` in the lane case. Needs a better default.
* If you have stack and size, you still run the stack transform and apply thickness around the boundary. Should warn and drop one.
* The line overlay (`line: true`) inherits `size`. When the normalizer builds the line layer for a thickness-mode area, delete size from that layer's encoding so the line just traces the centerline.
* thicknessOffsetRef(...) does nothing but call halfThicknessRef(...) and return the result
* `withOffset` can interpolate `undefined` into signals. This could silently cause issues.

joelostblom:
> I wonder if there's a way to align the two specs more closely? It would also give a natural stacking semantics here I think.

Interesting thought, I can explore this more. For stacking densities specifically, I think it can make them tricky to interpret, except for the special case of comparing two distribution in [a "split" violin](https://seaborn.pydata.org/examples/grouped_violinplots.html), so it would be interesting if this direction could bring us closer to that. 

> Gets us closer to density marks without going all the way yet so we can compose better.

Yes exactly! I'm still imagining that we could create a shortcut/composite mark like `mark_density` that avoids the need for an explicit transform as per https://github.com/vega/vega-lite/pull/9810. A similar shortcut could exist for violins, either via the same `mark_density` and we detect what to do depending on which encodings are specified, or via a separate `mark_violin` mark.

> There is some overlap with https://github.com/vega/vega-lite/pull/9823 so maybe we need to resolve that one first.

Sounds good, after we merge/reject that PR I can rebase this and address your other comments.

joelostblom:
@domoritz I think I have addressed all your comments now. The only additional change I did was in [b4df66a](https://github.com/vega/vega-lite/pull/9883/commits/b4df66aa238224218ff285373da6dda66e62ea1d), which ensures that area marks show up even if one axis is missing an encoding. For example, the following spec used to show up as empty because there was no categorical y-encoding, but now it shows the areas:

```json
{
  "data": {"url": "data/penguins.json"},
  "transform": [
    {"filter": {"field": "Sex", "oneOf": ["MALE", "FEMALE"]}},
    {
      "density": "Body Mass (g)",
      "groupby": [ "Sex"],
      "resolve": "independent"
    }
  ],
  "mark": {"type": "area", "opacity": 0.9, "stack": "normalize"},
  "encoding": {
    "x": {
      "field": "value",
      "type": "quantitative",
      "title": "Body Mass (g)"
    },
    // "y": {"field": "Species", "type": "nominal"},
    "size": {"field": "density", "type": "quantitative"},
    "color": {"field": "Sex"}
  }
}
```

image

This is more in line with how VL works elsewhere, such as if we were to use `yOffset` instead of `size`:

image

Or `size` with a `point` mark instead of an area:

image

With that, I think this PR is ready for another look!

---

@joshpoll I also looked into adding stacking, but I have a hard time reasoning about this in the context of density charts. If I use `stack: center` on a density plot with a quantitative y-axis, I get this stack behavior which I don't think is that useful/informative:

image

[Open the Chart in the Vega Editor](http://localhost:1234/editor/#/url/vega-lite/N4IgJghgLhIFygK4CcA29zQgegA4FMA7Ac0QEtCBnAOgCtKB7QkAXwBoQpkIqAzB5AFt4AbVC8yqKPmTxxZfKjAYAyvgAeIDk3wB5XqJABZAIIAZAKJaQAMQunLIALot2oMEUpkoATwwAhBjAfAAIjCEpKEIAKYgBKa2JkBkRcACM-OBEQNU0nDmR8RlQAN3wMCg8CQg9CKFZ8kEEIZABrOU4fAgxcBgp67VwIAGNvTIAGagBODkoYYfa4EEIBZtQyAC9y9hAiYaCKYg7NBBAJRWUlkohURHKOX26lgEdEHihvaDIy6w+oVHKS0CwTCESisQSOy8Ww65yUGFqXl8vy6gJAr3enw+Px2+1QAlhCnhS1yrFcQA)

So it seems to be that just enabling this already existing Vega-Lite behavior for the `size` encoding, would not be that helpful, at least not for density charts. It would be more useful if the result had a fixed center baseline and densities grew out either in the positive or negative direction, like in this mockup:

image

Since this would require some special handling, I wonder if it would make more sense to pursue this together with streamgraph support in a separate PR; especially since I'm not that familiar with streamgraphs and not sure what the desired behavior would be in all cases. Would you want to create an issue with some more details and a spec that you would like to be able to type together with its desired output?

joshpoll:
I should have clarified that VL already has support for streamgraphs! https://vega.github.io/vega-lite/examples/stacked_area_stream.html

It seems like the `size` channel could potentially simplify the existing streamgraph spec by removing `"axis": null` and the explicit `"stack": "center"` (if we default to center stacking for the size channel).

The split violin could be supported similar to diverging stacked bars (https://vega.github.io/vega-lite/examples/bar_diverging_stack_population_pyramid.html) using `"stack": "zero"`.

So these would be my proposed specs if we enable stacking:

**violin**
```json
{
  "transform": [{"density": "Body Mass (g)", "groupby": ["Species"], "resolve": "independent"}],
  "mark": "area",
  "encoding": {
    "x": {"field": "value", "type": "quantitative"},
    "y": {"field": "Species", "type": "nominal"},
    "size": {"field": "density", "type": "quantitative"}
  }
}
```

**streamgraph (alternative to existing example gallery spec)**
```json
{
  "mark": "area",
  "encoding": {
    "x": {"timeUnit": "yearmonth", "field": "date"},
    "size": {"aggregate": "sum", "field": "count"},
    "color": {"field": "series"}
  }
}
```

**"diverging" violin**
```json
{
  "transform": [
    {"density": "Body Mass (g)", "groupby": ["Species", "Sex"], "resolve": "independent"},
    {"calculate": "datum.Sex == 'FEMALE' ? -datum.density : datum.density", "as": "signed_density"}
  ],
  "mark": "area",
  "encoding": {
    "x": {"field": "value", "type": "quantitative"},
    "y": {"field": "Species", "type": "nominal"},
    "size": {"field": "signed_density", "type": "quantitative", "stack": "zero"},
    "color": {"field": "Sex", "type": "nominal"}
  }
}
```

joelostblom:
Ah I see, you're thinking that the split violin would have one side computed explicitly as a negative density. In that case, I wonder if the offset channel would be a better match, similar to the examples I put up in PR #9889? I'm thinking that the more useful mental model for a "split violin" maybe is just two separate density areas, which I think becomes more explicit.

For the streamgraph case, I personally have an easier time conceptualizing `stack` in the context of a position based channel like `x/y`, `xOffset/yOffset` than the `size` channel. I believe your spec should also be covered by the recent offset PRs I have been putting up. After merging https://github.com/vega/vega-lite/pull/9887 and https://github.com/vega/vega-lite/pull/9897, it would be possible to use the spec you mentioned but with `yOffset` instead of `size` and specifying the stack to be centered (but we could discuss if that should be the default in #9897):

```json
{
  "height": 300,
  "data": {"url": "https://cdn.jsdelivr.net/npm/vega-datasets@v3.2.1/data/unemployment-across-industries.json"},
  "mark": "area",
  "encoding": {
    "x": {
      "timeUnit": "yearmonth", "field": "date"
    },
    "yOffset": {
      "aggregate": "sum", "field": "count",
      "stack": "center"
    },
    "color": {"field":"series", "scale":{"scheme": "category20b"}}
  }
}
```

image

Do you think this would be a sufficient convenience for creating streamcharts?

joelostblom:
Before reviewing this PR, I think it would be beneficial to review the PRs related to offsets that I have put up in the last few days. Particularly, https://github.com/vega/vega-lite/pull/9899 could introduce a more general mechanism that also covers the case of creating violin plots, so I'm leaning towards that this PR might be redundant.

PR #9887, opened 2026-07-20
fix: anchor ranged offsets without base channels

joelostblom:
This is a minor follow up on #9823 which makes ranged offset also work when there only one axis encoding.

These changes treat a missing x or y position as one implicit full-view lane/band when a bar or area is driven by a quantitative offset. Both generated range edges are anchored at the same screen origin, using the encoded offset for one edge and offset zero for the other. Ranged xOffset and yOffset now scales the full width or reversed height so signed values remain visible and diverge in the expected directions. Centered offset ranges for point and line marks are preserved.

| Before this PR | After this PR |
|-|-|
| image | image |

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgFgpglg5mAuIBcBGADGgNCAJgQ3j2VADc8AbAVwgGdkBtUIpEAQRGwCNkBaAJgAc2AMbIQADw64xAWxABfTEzEAhKdyQ8ArFpFjJ2HLIVKQzEAGF1vACwBmPSwPSWAOxPKWAEWuaAnCiOElJGbh5mYuxcyAJCIKIsAJ4hxoqeIGrRSDpByYappuZWWfa5KWFpEd6+OfFieS4g7vIAuoogMngATgDWxCDwiQAOEGKc3SYgEK7CAPY4UK4w-ZJIoABmUBDkoRHtiQDy6+s0EIhrIJvbu9zYgyNiAI6UeK7wUITvJKPtc+SzXX6Vx2YlE8nBQA)

Note that on main you could add an undefined channel (e.g. `"y": {"field": "aa"}`) to see something closer to the correct behavior, but with this PR that workaround will not be needed:

image

### DIFF
--- a/src/compile/mark/encode/position-point.ts
+++ b/src/compile/mark/encode/position-point.ts
@@ -29,7 +29,7 @@ export function pointPosition(
     defaultPos,
     vgChannel,
   }: {
-    defaultPos: 'mid' | 'zeroOrMin' | 'zeroOrMax' | null;
+    defaultPos: 'mid' | 'zeroOrMin' | 'zeroOrMax' | 'start' | null;
     vgChannel?: 'x' | 'y' | 'xc' | 'yc';
   },
 ) {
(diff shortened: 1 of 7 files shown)


joelostblom:
This is now ready to be reviewed.

PR #9888, opened 2026-07-20
fix: align ranged offsets to closest-to-zero baseline

joelostblom:
This is a small follow up to #9823 with more intelligent placing of the axis tick and label.

Since #9823, ranged bars and areas with a quantitative offset place categorical axis ticks at a fixed band edge: `0` for `x` and `1` for `y`. This does not necessarily match the mark's offset baseline when the offset domain is asymmetric, reversed, or explicitly excludes zero. This PR positions the ranged mark baseline and its categorical axis at the offset-domain value closest to zero, which can be seen as the more natural baseline for most ranged marks.

| Offset domain | Baseline |
| --- | --- |
| Includes zero | `0` |
| Positive only | Domain minimum |
| Negative only | Domain maximum |

|  Before #9823 | After #9823 | After this PR |
|-|-|-|
| image | image | image |
| image | image | image |

[Diverging bars in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgFgpglg5mAuIBcoDO8IAdkDYAMAvgDQgAmAhvOcqAG7kA2ArhKsgNqjVIgCCIJAEbIAtACYAHCQDGyEAA8BZOQFsQxLnIBCS4UhEBOAKwy5ikqVXqimngGFdogCwBmUz3PKeAO2u2QACKO+gCMIe4KSpY+fiDcfMEhUiCyPACeUVYacdqJJilyGRZZNjn2wa4RRV4gvtnxQULIRvmpINXRtbHx-E36khGenXWl8Tp9Ii2DmTH1cg4TlQUeMyBqczyNIHqG4cuRxTzro3K928gGBlWrI-7j50hT+x1yt2UgCw8hbs+rx-5bPQhVqFP7qAC6xDW5AATgBrGggeBpTAQOSCWHWEAQbzSAD2pCg3hgiMUKBAADMoBAGJ1qFC0gB5CkU1AQRDkqk0zrCEjI1FyACOTHI3ngUCo4toaKh+IYeJhiK5tLksgZSupKp4lgIuqAA)

[Regular Bars in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgFgpglg5mAuIBcoDO8IAdkDYAMAvgDQgAmAhvOcqAG7kA2ArhKsgNqjVIgCCIJAEbIAtACYAHCQDGyEAA8BZOQFsQxLnIBCS4UhEBOAKwy5ikqVXqimngGFdogCwBmUz3PKeAO2u2QACKO+gCMIe4KSpY+fiDcfMEhUiCyPACeUVYacdqJJilyGRZZNjn2iW4F6Zkx2fFBQsgh+akgRV4gvnVy-I36khGe0Z2x8Tp9Ik4tZjUj3eUTBgDsg7Nq84HBImGrxTzrpfG9IHpJEe3DXYe5fc3ns1f+DreVrRcl-g0nyAb3eyDrAC6xAB5AATgBrGggeBpTAQOSCcHWEAQbzSAD2pCg3hg0MUKBAADMoBAGMNqCC0gB5IlE1AQRCEklk4bCEiw+FyACOTHI3ngUCogtoCJBmIYGLB0JZ5LksipMtJcp4lhI5HkUDYhKR3lIAAUMaghVAMb4kHgAHRGAi2ghAA)

I contemplated reverting to the default 0.5 bandsize and I still think we can consider that if we think these changes are too complex, but I believe the more natural position of the tick and label is this PR's "closest to 0" placement. I think this is where I'm the most likely to want a gridline to appear if I turn that on. It is also relatively straightforward to achieve the old 0.5 baseline manually with `"axis": {"bandPosition": 0.5}` whereas it is difficult to place the baseline close to 0 manually.

### A few notes

- The baseline is evaluated at runtime so it works with data-driven domains. Its scaled position is clamped to the base band, and invalid or unsupported scale calculations fall back to the middle of the band.
- Bar and area marks use the same computed baseline for their secondary edge and categorical axis.
- Axis ticks, grid lines, and labels remain aligned with the rendered baseline.
- Reversed and independently resolved offset scales are supported.
- Explicit `axis.bandPosition` values continue to take precedence.
- Point and line offsets remain centered, preserving jitter behavior.
- This PR only covers ranged offsets within an existing discrete `x` or `y` lane. Ranged offsets without a base position channel are intentionally outside its scope.

The distinction between ranged and non-ranged marks follows existing Vega-Lite geometry semantics: point and line offsets represent positional displacement, while bar and area offsets synthesize a range from an in-band baseline to the encoded offset value.

### DIFF
--- a/examples/compiled/area_ranged_offset_quantitative.vg.json
+++ b/examples/compiled/area_ranged_offset_quantitative.vg.json
@@ -71,7 +71,9 @@
               "y2": {
                 "scale": "y",
                 "field": "team",
-                "offset": {"scale": "yOffset", "value": 0}
+                "offset": {
+                  "signal": "isFinite(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1]))) && isFinite(bandwidth(\"y\")) && bandwidth(\"y\") > 0 ? clamp(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1])), 0, bandwidth(\"y\")) : (isFinite(bandwidth(\"y\")) ? (bandwidth(\"y\")) / 2 : 0)"
+                }
               },
               "defined": {
                 "signal": "isValid(datum[\"score\"]) && isFinite(+datum[\"score\"])"
@@ -122,10 +124,20 @@
       "orient": "left",
       "grid": false,
       "title": "team",
-      "bandPosition": 1,
+      "bandPosition": {
+        "signal": "isFinite(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1]))) && isFinite(bandwidth(\"y\")) && bandwidth(\"y\") > 0 ? clamp(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1])) / bandwidth(\"y\"), 0, 1) : 0.5"
+      },
       "encode": {
         "labels": {
-          "update": {"y": {"scale": "y", "signal": "datum.value", "band": 1}}
+          "update": {
+            "y": {
+              "scale": "y",
+              "signal": "datum.value",
+              "band": {
+                "signal": "isFinite(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1]))) && isFinite(bandwidth(\"y\")) && bandwidth(\"y\") > 0 ? clamp(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1])) / bandwidth(\"y\"), 0, 1) : 0.5"
+              }
+            }
+          }
         }
       },
       "zindex": 0
--- a/examples/compiled/area_ranged_offset_quantitative.vg.json
+++ b/examples/compiled/area_ranged_offset_quantitative.vg.json
@@ -131,11 +131,7 @@
         "labels": {
           "update": {
             "y": {
-              "scale": "y",
-              "signal": "datum.value",
-              "band": {
-                "signal": "isFinite(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1]))) && isFinite(bandwidth(\"y\")) && bandwidth(\"y\") > 0 ? clamp(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1])) / bandwidth(\"y\"), 0, 1) : 0.5"
-              }
+              "signal": "scale(\"y\", datum.value) + bandwidth(\"y\") * (isFinite(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1]))) && isFinite(bandwidth(\"y\")) && bandwidth(\"y\") > 0 ? clamp(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1])) / bandwidth(\"y\"), 0, 1) : 0.5)"
             }
           }
         }
--- a/src/compile/axis/properties.ts
+++ b/src/compile/axis/properties.ts
@@ -23,6 +23,7 @@ import {contains, normalizeAngle} from '../../util.js';
 import {isSignalRef} from '../../vega.schema.js';
 import {mergeTitle, mergeTitleFieldDefs} from '../common.js';
 import {guideFormatType} from '../format.js';
+import {rangedOffsetBaseline} from '../scale/rangedOffset.js';
 import {UnitModel} from '../unit.js';
 import {ScaleType} from './../../scale.js';
 import {AxisComponentProps} from './component.js';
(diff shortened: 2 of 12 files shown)


src/compile/axis/properties.ts
```diff
@@ -124,7 +125,7 @@ export function defaultGrid(scaleType: ScaleType, fieldDef: TypedFieldDef<string
 
 export function defaultBandPosition(model: UnitModel, channel: PositionScaleChannel) {
   if (model.isRangedOffset(channel)) {
-    return channel === 'x' ? 0 : 1;
+    return rangedOffsetBaseline(model, channel).bandPosition;
```
joelostblom, line 128: Applies the dynamic position only when `isRangedOffset` is true. Other offset uses retain Vega's centered axis default.

examples/compiled/area_ranged_offset_quantitative.vg.json
```diff
@@ -71,7 +71,9 @@
               "y2": {
                 "scale": "y",
                 "field": "team",
-                "offset": {"scale": "yOffset", "value": 0}
+                "offset": {
+                  "signal": "isFinite(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1]))) && isFinite(bandwidth(\"y\")) && bandwidth(\"y\") > 0 ? clamp(scale(\"yOffset\", clamp(0, extent(domain(\"yOffset\"))[0], extent(domain(\"yOffset\"))[1])), 0, bandwidth(\"y\")) : (isFinite(bandwidth(\"y\")) ? (bandwidth(\"y\")) / 2 : 0)"
```
joelostblom, line 75: These differences on the Vega level definitely look more complex than the previous more simple logic. But that also means that if we don't have this behavior as the default, it would be quite difficult to manually place the tick and label close to 0 since it involves something similar to what is shown here for the bandposition.

src/compile/mark/encode/position-range.ts
```diff
@@ -138,21 +133,12 @@ function pointPosition2OrSize(
...
-                model.scaleName(getOffsetScaleChannel(baseChannel)),
-                {},
-                {},
-              ),
-            },
-          ),
+          [vgChannel]: {
+            ...ref.valueRefForFieldOrDatumDef(channelDef, scaleName, {}, {}),
+            offset: rangedOffsetBaseline(model, baseChannel).offset,
+          },
```
joelostblom, line 141: uses the same helper for the mark's secondary edge, preventing mark and guide alignment from diverging.

src/compile/scale/rangedOffset.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (!isContinuousToContinuous(offsetScaleType)) {
+    return fallbackBaseline(positionScaleName, positionScale?.get('type'));
+  }
+
+  const offsetName = stringValue(offsetScaleName);
+  const domainExtent = `extent(domain(${offsetName}))`;
+  const closestToZero = `clamp(0, ${domainExtent}[0], ${domainExtent}[1])`;
+  const scaledBaseline = `scale(${offsetName}, ${closestToZero})`;
+
+  if (positionScale?.get('type') === 'band' && positionScaleName) {
```
joelostblom, line 35: converts that value to a pixel offset and normalized `bandPosition`, with guards for invalid scale results or zero bandwidth.

joelostblom (commented):
This is now ready for review. I updated the description above with additional PR details.

joshpoll:
This is really interesting and the baseline-aligned labels do seem cool for the diverging bars. I'm wondering about consistency of labels w.r.t. faceting labels. Seems like there is some overlap between yOffset and a vertical facet?

The use of the line on the y-axis kinda suggests they are in the same ordinal coordinate space whereas the facets will happen in free space, but this feels like a confusing edge case.

Just wanted to surface those questions around consistency and expectations about what the different spaces mean

joelostblom:
Thanks for raising those points @joshpoll, I agree that a similar visualization could be achieved using row faceting:

image

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgFgpglg5mAuIBcBWADAGhAEwIb12VADdcAbAVwgGdkBtUQpEAQRCwCNkBaAJgA4sAY2QgAHuxyiAtiAC+GRqIBCkrkm4BOFMNESs2GfMUgmIAMJqeAFgDMu5vqnMAdsaXMAIlY0BGXw7ikoau7qaibJzIvoIgIswAnsFGCh4gqlFIvjpxokkGKSZmlpm+9rmJyaGp4V4+2YH5ziBuNWaRIOp8sfFBBdVFKj7cKDm9TiEtYcXDmpqBE4Vp3pnc-gtVILJtEfXzFSBNk62DzBmd0WN5mydpJRdZ5b1HS7UgKw8NBy-M2wC6Ci2uAATgBrIggeAJAAOEFEHBBxhAEBcQgA9tgoC4YBCJEhQAAzKAQMiTQiApL4kBEkmTLhYKGw0QARwouBc8CgBE5xDhgPRZDRwIhNNJohEgOBaIA7iLiWLmIY5MqgA)

I think they are a bit different in that using the row facet for the categorical grouping variable an the y-axis for the range of the bars is busier by default and lends itself better to exact comparisons since there are gridlines and axes. Using the y-axis as for the categorical grouping variable and the yOffset for the range of the bars makes it easier to create a [sparkline chart](https://en.wikipedia.org/wiki/Sparkline) where the important information usually is a trend or size comparison rather than reading an exact number of the chart (it is of course also achievable by styling the facet but more cumbersome). I also think this lends itself well for comparing distributions such as historgrams and densities, and I will elaborate a bit on the case of the [split density](https://github.com/vega/vega-lite/pull/9889) as a reply to your other point on the violin PR.

Allowing the yOffset encoding to be used for quantitative ranges also frees up the y-axis to be used as an additional level of categorical grouping if one would like to combine a y-encoded cateorigcal with another categorical on the row facet.

And at least to me, I think it is an intuitive extension that the grammar supports these other marks in the offset channels since points have been supported for some time. It seems like something I expect "should" work, given that points (and lines) already do work.

Issue #4318, opened 2018-12-04
Mouseover selection should trigger mouse out when swapping application


kanitw:
Vega's `hover` syntax does this quite well.

![kapture 2018-12-04 at 9 36 22](https://user-images.githubusercontent.com/111269/49461259-185b4080-f7a8-11e8-8bb0-bb1ab3123600.gif)

However, selection on `"mouseover"` in Vega-Lite does not. 

![kapture 2018-12-04 at 9 41 40](https://user-images.githubusercontent.com/111269/49461556-d1ba1600-f7a8-11e8-87fb-3af55e59fb18.gif)

PR #9892, opened 2026-07-21
fix: ResizeObserver based container sizing

### LINKED ISSUE #9349: Adopt ResizeObserver for "container" width/height?
## Enhancement Description

Vega-Lite’s [responsive width and height](https://vega.github.io/vega-lite/docs/size.html#specifying-responsive-width-and-height) are currently implemented using _resize_ events on the window. If Vega-Lite used [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) instead, it would correctly handle a variety of common use cases:

- when the container is initially detached (as in Observable notebooks and Observable Framework)
- when the page is modified without resizing the window (_e.g._, opening and closing a sidebar)

The documentation specifically calls out a workaround by manually dispatching _resize_ events:

> Vega listens to the window.resize event to update plot size from container size. This should cover many use cases. However, if you change the container size programmatically (e.g., you build a custom divider view), you’ll need to trigger window.resize manually. In a modern browser, you can do: window.dispatchEvent(new Event('resize'));.

This workaround will no longer be needed if Vega-Lite adopts ResizeObserver.

ResizeObserver is [available in >96% of browsers](https://caniuse.com/resizeobserver).

No syntax changes are needed.

Thanks for your consideration! 🙏 

## Checklist

yuvvantalreja:
## PR Description

Fixes #9349 

Refs: vega/vega#4318

<details>

- [ ] This PR is atomic (i.e., it fixes one issue at a time).
- [ ] The title is a concise [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties").
- [ ] `npm test` runs successfully
- For new features:
  - [ ] Has unit tests.

Tips:

</details>


domoritz:
Heads up — the Vega side of this is being reworked, and it changes what this PR should emit.

vega/vega#4318 is being rebuilt around a first-class **`container:resize` event source**, so a spec can subscribe to container changes the same way it subscribes to `window:resize`:

```json
{"on": [{"events": "container:resize", "update": "containerSize()[0]"}]}
```

Once that lands, the `{"signal": "autosize"}` handler here should be replaced by `container:resize`. That handler is really a workaround for a Vega gap rather than a fallback: `view.resize()` sets `_autosize` and touches the `autosize` signal but never re-evaluates the `containerSize()`-derived `width`/`height` signals, so listening on `autosize` is a way to force the re-read. Two costs come with it — it fires on *any* autosize change, not just container changes, and it still needs someone to call `view.resize()` by hand, which is why the runtime test here does `await view.resize().runAsync()`. With the event source, the observer fires on its own and that test can assert a real container change instead.

Worth keeping `window:resize` alongside it for now, though — not on the merits, but for compatibility. A spec emitting only `container:resize` on an older Vega gets `Can not resolve event source: container` and then no resize behaviour at all, and since Vega-Lite output runs on whatever Vega the user has, that would be a silent regression. Same for the (now rare) case of no `ResizeObserver`. Keeping both means old runtimes degrade to today's behaviour. On the merits the two are redundant: if a window resize changes the container's box, `ResizeObserver` already fires; if it doesn't, the `window:resize` handler re-reads the same value and the signal short-circuits. So the clean end state is to drop `window:resize` in the same PR that raises the Vega peer-dependency floor.

One note on the docs change: "In browsers where Vega uses `ResizeObserver`, container changes are detected automatically" isn't true of anything in this PR yet — it describes #4318. It becomes accurate once that ships, so it's worth landing them in that order.

Thanks for pushing on this — #9349 has been open since 2024 and the container-sizing story is better for it.

PR #9902, opened 2026-07-28
feat(selection): field-comparison predicates for point selections

jonathanzong:
This PR introduces `predicate` as a property of a point selection. We prototyped this in animated vega-lite to directly specify predicates for selections. The changes needed to support this in vega-functions were merged earlier in https://github.com/vega/vega/pull/3675.

**Claude below:**

This PR adds `predicate` to a point selection's config, so the selection holds a comparison rather than the values it captured.

```json
{
  "name": "upto",
  "select": {
    "type": "point",
    "predicate": {"field": "year", "lte": {"expr": "datum.year"}}
  }
}
```

Clicking now selects every datum whose `year` is at most the clicked year.

A point selection otherwise holds captured values, and a datum falls within the selection only by matching those values exactly. Exact matching makes some ordinary interactions awkward or impossible. "Highlight every year up to the one I clicked" forces the author to restate the comparison at every place that uses the selection, and "select a window around what I clicked" has no expression at all.

### How it works

Each field predicate inside `predicate` becomes one of the selection's tuple fields, holding a comparison rather than a value to match. `vlSelectionTest` already understands `E-LT` / `E-LTE` / `E-GT` / `E-GTE` / `E-VALID` / `E-ONE` and `R`, so the selection store evaluates the comparisons directly. The change adds no predicate machinery and touches no code that consumes a selection.

Only the store's contents change, so everything downstream keeps working: filter transforms (including `empty` semantics), conditional encodings, scale domains, scale binding, legend binding, cross-view references in layers and concats, toggling, clearing, and initial values.

A comparison value is an ordinary Vega expression, evaluated where the selection captures its input, so `datum` is the datum the user interacted with. The field a predicate tests and the field its comparison value reads need not be the same, so one selection can compare two fields.

The available comparisons are the existing field predicates: `equal`, `lt`, `lte`, `gt`, `gte`, `range`, `oneOf`, `valid`. A single `lte` like the example above selects everything up to the captured value — this description calls that shape *cumulative*. Two comparisons on one field bracket a range around the captured value — called *windowed* below:

```json
"predicate": {
  "and": [
    {"field": "year", "gte": {"expr": "datum.year - 10"}},
    {"field": "year", "lte": {"expr": "datum.year"}}
  ]
}
```

### Details for review

**The compiler does not deduplicate predicate fields**, unlike ordinary projections. A windowed predicate compares the same field twice, and the store evaluates its comparisons in order, so both entries have to survive.

**Scale binding used to require an interval selection.** Scale binding needs a selection that resolves to a range, and a predicate gives a point selection a range. `bind: "scales"` now accepts `interval || predicate`, so a windowed selection can pan a view as the window moves.

**Legend binding composes with a predicate.** The legend supplies the comparison value, so clicking an entry selects every datum up to that entry.

### Validation

- The compiler accepts a single predicate or a flat `"and"`. A store tuple tests its fields conjunctively and has no representation for `"or"` or `"not"`, so `"or"` and `"not"` warn, and the compiler ignores the predicate and falls back to the ordinary projection.
- A predicate on an **interval** selection warns and is ignored, because a brush supplies its own extent. Without the guard the compiler skipped the interval's channel projection and emitted an invalid event stream.
- A predicate comparing against `datum` cannot work with **`nearest`**, whose voronoi overlay binds `datum` to a scenegraph item rather than a data tuple. That combination warns rather than silently reading `undefined`.

### Tests

- `test/compile/selection/select-predicate.test.ts` — 32 tests: tuple construction, every consumption point above, interaction with toggle/clear/init, all eight comparison types, `timeUnit` resolution, and each validation path.
- `test-runtime/select-predicate.test.ts` — 9 tests driving the compiled Vega in a browser and asserting which rows the selection holds: cumulative, strict, windowed, range, cross-field comparison, initial value, toggle accumulation, union-not-intersection semantics across store entries, and the resulting domain of a bound scale.

Docs under **Point Selection Properties**.

### Examples

`selection_predicate_cumulative` shows the click-driven case: clicking a year highlights every year up to it.

The windowed, animated case is `animated_bird_migration`, which lives in #9915 because it draws its animated points over a static basemap layer. Its selection holds a ten-day window on the animation clock:

```json
"predicate": {"and": [
  {"field": "day", "lte": {"expr": "anim_value"}},
  {"field": "day", "gt": {"expr": "anim_value - 10"}}
]}
```

### Notes

- Purely additive. No existing spec compiles differently, and the only new example output comes from the examples this PR adds.
- The animation stack builds on predicates. An animated selection's predicate compares against the animation clock rather than a clicked datum, and that comparison expresses cumulative lines and windowed panning. The animation PRs depend on this PR, and this PR stands alone.


joelostblom:
@jonathanzong  It's exciting to see these animation-related PRs! I just wanted to bring your attention to https://github.com/vega/vega-lite/pull/9830 and https://github.com/vega/vega-lite/pull/9831; if you find that there is overlap between what you are working on and either of my earlier two PRs, feel free to let me know and we can close what I put up. I have no personal attachment to them and was just trying to get the animation functionality working for more examples/use cases.

jonathanzong:
Thanks Joel! Sorry that I had missed your two PRs previously.

It looks like https://github.com/vega/vega-lite/pull/9830 will be superseded by what I'm currently working on. Your approach there was almost the same as what I intended, except our design with selection predictates (this PR) means that we don't need to expose an "ANIM_value" signal to express the predicate filter.

I think https://github.com/vega/vega-lite/pull/9831 is related but distinct to this PR. Looks like you're adding parameter support to where field predicates already exist, and this is adding field predicates that can live inside selection parameters. I haven't really thought through the differences / whether both are needed in general for VL, however, the selection approach is better for animation because selections can be referenced across not just transforms but also conditional encodings, etc.

joelostblom:
No worries and thanks for the explanations! I'm glad this get rids of the ANIM custom name, I was not entirely happy with that solution myself, so I will go ahead and close #9830

